PR #12, opened 2017-03-15
Add pub(restricted) docs

cramertj:
See rust-lang/rust#32409
cc @steveklabnik

### DIFF
--- a/src/visibility-and-privacy.md
+++ b/src/visibility-and-privacy.md
@@ -134,6 +134,41 @@ For a Rust program to pass the privacy checking pass, all paths must be valid
 accesses given the two rules above. This includes all use statements,
 expressions, types, etc.
 
+## `pub(crate)` and `pub(in path)`
+
+In addition to public and private, Rust allows users to declare an item as
+`pub(crate)` or `pub(in path)`. These restrictions make it possible to specify
+the exact scope of an item's visibility. As their names would suggest,
+`pub(crate)` makes an item public within the current crate, and `pub(in path)`
+makes an item public within the specified path.
+
+Here's an example using `pub(crate)` and `pub(in path)`:
+
+```rust
+pub mod outer_mod {
+    pub mod inner_mod {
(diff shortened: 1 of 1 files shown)


steveklabnik:
This text looks good to me! I'm going to have to think about CI....

petrochenkov:
Could you add three more details
- `path` in `pub(in path) item` must be some parent module of the item.
- `pub(super)` and `pub(self)` are shortcuts for `pub(in super)` and `pub(in self)` respectively
- `pub(self)` is the same thing as private

cramertj:
@petrochenkov Is there some reason to use `pub(self)` that I should mention? It doesn't seem well-motivate to me.

petrochenkov:
@cramertj 
Not yet (but see https://github.com/rust-lang/rust/issues/32770), but it's useful for understanding the model.

steveklabnik:
blocked on https://github.com/rust-lang-nursery/reference/issues/14

steveklabnik:
Update: this is now unblocked, but we need the feature to stabilize to land.

petrochenkov:
@steveklabnik 
Everything is stabilized already, these docs are the only missing part.

steveklabnik:
@petrochenkov running in nightly without the flag still requires the flag. Maybe my nightly was out of date?

steveklabnik:
Ah, it was. My bad!

steveklabnik:
@cramertj can you rebase this please? We can then get it in.

cramertj:
Something weird is going on with the tests-- Travis is complaining that it can't find `outer_mod` in the crate root, but it works just fine when I run it on the playground or using rustc locally. Any ideas?

matthewjasper:
You need to add `# fn main() {}` to the example otherwise everything gets wrapped in `main`.

cramertj:
@matthewjasper lol ty

steveklabnik:
Thanks!

PR #26, opened 2017-03-23
Make sure we don't talk about unstable things

### LINKED ISSUE #14: CI: stable nightly
We need to run on nightly, but not allow nightly features. This is because this stuff now lands in rust master, so we need to be able to document newly stable features, but before they actually hit stable.

I plan on writing a script we can run in CI to check this, but if anyone is interested in helping out, that'd be great!

steveklabnik:
Fixes #14 

I'm doing this to make sure this passes; i think these two features might have actually become stable, let's see with CI.


steveklabnik:
One was, one was not. This should be good to go.

r? @frewsxcv

PR #30, opened 2017-03-29
Explain block like expression statements

matthewjasper:
A small thing that I noticed while going through the expressions chapter. `unsafe` blocks also have this property, but I'm not sure if they should just be considered another kind of block.
Does this need an example?

### DIFF
--- a/src/statements.md
+++ b/src/statements.md
@@ -36,7 +36,22 @@ declaration until the end of the enclosing block scope.
 
 ## Expression statements
 
-An _expression statement_ is one that evaluates an [expression](expressions.html)
-and ignores its result. The type of an expression statement `e;` is always
-`()`, regardless of the type of `e`. As a rule, an expression statement's
-purpose is to trigger the effects of evaluating its expression.
+An _expression statement_ is one that evaluates an
+[expression](expressions.html) and ignores its result. The type of an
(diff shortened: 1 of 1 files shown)


src/statements.md
```diff
@@ -36,7 +36,12 @@ declaration until the end of the enclosing block scope.
...
-purpose is to trigger the effects of evaluating its expression.
+An _expression statement_ is one that evaluates an
+[expression](expressions.html) and ignores its result. The type of an
+expression statement `e;` is always `()`, regardless of the type of `e`. As a
+rule, an expression statement's purpose is to trigger the effects of evaluating
+its expression. [Block expressions](expressions.html#block-expressions) and
+control flow expressions, such as [`loop`
+expressions](expressions.html#infinite-loops) and [`if`
+expressions](expressions.html#if-expressions), which evaluate to `()` can be
+used as statements without a trailing semicolon.
```
frewsxcv: Should this be:

> ...can be used as expressions statements...

Or are we using 'expression statement' and 'statement' interchangeably here?

frewsxcv: > ...which evaluate to `()` can be used as statements by omitting a trailing semicolon.

Maybe this makes it clear how to accomplish an expression statement?

steveklabnik:
Looks great, thank you!

Issue #9, opened 2017-03-11
Document all features


chriskrycho:
## Overview

Tracking issue for [RFC #1636][#1636]. Let's make the [reference](https://doc.rust-lang.org/reference.html) a reliable resource for all Rustaceans! :tada: This might seem intimidating, but the reality is that if we just chip away at it together, we'll be done in no time!

I'm going to be updating this parent issue with a master list of items that need to be documented in the reference as I find them. Quoting the RFC text:

> Updating the reference should proceed stepwise:
> 
> 1. Begin by adding an appendix in the reference with links to all accepted RFCs which have been implemented but are not yet referenced in the documentation.
> 2. As the reference material is written for each of those RFC features, remove it from that appendix.

Note that step 1 should be fairly straightforward; the main issue will be assembling the list of accepted-and-implemented-but-undocumented RFCs. (Also, any RFCs accepted before RFC 1636 but not yet stabilized should presumably be documented under the rules it establishes, but if I'm wrong about that, someone should let me know and I'll include them as well.)

Also, a pre-emptive apology for the scale of this issue description. We have let things get into a *rough* spot. (I plan to create documentation issues for each of the required items below once this list is completed, so this thread doesn't become *completely* unmanageable.)

**Status:** Last updated **May 13, 2017**, now processing a [copy] of @Eh2406's wonderful [gist] based on @matthewjasper's fork [here](https://gist.github.com/matthewjasper/e9246d2fc6b8d874fd989870a2bd9610).

## "How Can I Help?"

For right now, since you can't submit updates to the issue directly, there are two major ways you can help:

1. **Write documentation.** For the RFCs already triaged and filed under the major section *C. Documentation Needed* and specifically the subsection *2. Write reference material for undocumented, implemented, stabilized RFC features* below, you can open a pull request and reference this issue in its description. I'll keep an eye out for inbound links here and add those links to the relevant items, and once they're merged, we'll mark that item as done.

2. **Triage issues.** Work the [gist][copy] where I'm tracking the triaged-state, specifically looking at the [*Not Reviewed at All*] and [*Stabilized, Not Reviewed*] sections and evaluating whether they're documented or not. You can fork the gist and I will keep an eye out, checking roughly daily, for any changes to merge back into that document. (Do take a look at what others may have forked it so you don't duplicate work needlessly!) I'll then pull those over into this master issue.

[...]

PR #35, opened 2017-04-03
Add appendix for list of undocumented features

mrhota:
Re: #9 

r? @steveklabnik 

CC @chriskrycho

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -56,3 +56,5 @@
     - [Behavior not considered unsafe](behavior-not-considered-unsafe.md)
 
 [Appendix: Influences](influences.md)
+
+[Appendix: As-yet-undocumented Features](undocumented.md)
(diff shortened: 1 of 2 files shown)


steveklabnik:
This is an interesting idea; @chriskrycho do you think this is helpful?

mrhota:
@steveklabnik I think this is actually a requirement of the RFC for #9. The quotation at the top of the issue mentions it, with another mention in section C.1.

chriskrycho:
Quite! So much so that RFC 1636 said we'd do exactly this:

> Updating the reference should proceed stepwise:
> 
> 1. Begin by adding an appendix in the reference with links to all accepted RFCs which have been implemented but are not yet referenced in the documentation.
> 2. As the reference material is written for each of those RFC features, remove it from that appendix.

chriskrycho:
I've integrated these into the tracking issue; thanks again for the help!

PR #36, opened 2017-04-05
Fix tidy on Windows

petrochenkov:
This is exactly the same issue as https://github.com/rust-lang/book/pull/549, `x.py test src/tools/tidy` is failing again.
r? @steveklabnik @carols10cents

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto eol=lf


steveklabnik:
Thanks!

steveklabnik:
> x.py test src/tools/tidy is failing again.

incidentally, how did you notice this? Did you try to update the submodule yourself?

petrochenkov:
@steveklabnik 
Default git settings on Windows is "checkout as \r\n, commit as \n", so this issue reproduced automatically when I updated the Rust repo and run tidy.
This can be solved by adding `.gitattributes` to the repo once, or changing settings to "checkout as \n" by all people testing Rust on Windows. Adding `.gitattributes` seems nicer and the main repo did this, but every time someone creates a new submodule and forgets to add `.gitattributes` the problem repeats.

PR #40, opened 2017-04-14
document default object lifetime bounds

QuietMisdreavus:
fixes rust-lang/rust#31137

also technically part of #9


steveklabnik:
This looks good to me, thanks so much!

PR #41, opened 2017-04-17
Document that parentheses are required for chained comparison operators (RFC 0).

ekse:


ekse:
This is my first attempt at a contribution to Rust (yay!), let me know if something is wrong or if you would like the text worded differently.

src/expressions.md
```diff
@@ -436,6 +436,9 @@ operators on standard types is given here.
   : Greater than or equal.
     Calls the `ge` method on the `std::cmp::PartialOrd` trait.
 
+Parentheses are required when chaining comparison operators. For example, the
+expression `a == b == c` is invalid and may be written as `((a == b) == c)`.
```
steveklabnik: could this be `(a == b) == c`, that is, without the outer parens? In `if` statements, those outer ones will cause a warning.

steveklabnik:
One tiny comment, otherwise, this looks good! Thank you! 😄

src/expressions.md
ekse: Good catch, it's fixed now.

steveklabnik:
Thank you!

PR #45, opened 2017-04-18
Discuss unambiguous function call syntax

mrhota:
And shrink the undocumented list

CC #9

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -550,16 +550,16 @@ let pi: Result<f32, _> = "3.14".parse();
 Function calls may sometimes result in ambiguities about receiver or referent.
 See [UFCS](#unambiguous-function-call-syntax) for how to resolve this.
 
-### Unambiguous Function Call Syntax
+### Unambiguous Function Call Syntax (UFCS)
 
 Several situations often occur which result in ambiguities about the receiver or
 referent of method or associated function calls. These situations may include:
 
 * Multiple in-scope traits define the same method for the same types
 * Auto-`deref` is undesirable; for example, distinguishing between methods on a
-  smart pointer itself and the pointer's referent.
-* Methods which take no arguments and return properties of a type, like
-  `SizeOf::size_of()`
+  smart pointer itself and the pointer's referent
+* Methods which take no arguments, like `default()`, and return properties of a
+  type, like `size_of()`
 
 The programmer may unambiguously refer to their desired method or function using
 Unambiguous Function Call Syntax (UFCS), specifying only as much type and path
@@ -567,15 +567,37 @@ information as necessary.
 
 For example,
 
-```rust,ignore
-// shorthand, only if `T` is a path
-T::size_of()
+```rust
+trait Foo {
+  fn quux();
+}
 
-// infer trait `SizeOf` based on traits in scope.
-<T>::size_of()
+trait Bar {
+  fn quux();
+}
 
-// completely unambiguous; works if multiple in-scope traits define `size_of()`
-<T as SizeOf>::size_of()
+struct Faz;
+impl Foo for Faz {
+    fn quux() {}
+}
+
+struct Baz;
+impl Foo for Baz {
+    fn quux() {}
+}
+impl Bar for Baz {
+    fn quux() {}
+}
+
+fn main() {
+    // shorthand, only if unambiguous
+    Faz::quux();
+
+    // Baz::quux(); // Error: multiple `quux` found
+
+    // completely unambiguous; works if multiple in-scope traits define `quux`
+    <Baz as Foo>::quux();
+}
 ```
 
 Refer to [RFC 132] for further details, motivations, and subtleties of syntax.
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -220,6 +220,9 @@ exact `self`-type of the left-hand-side is known, or dynamically dispatching if
 the left-hand-side expression is an indirect [trait
 object](types.html#trait-objects).
 
+[UFCS](#unambiguous-function-call-syntax) allows method invocation using
+function-like syntax.
+
 ## Field expressions
 
 A _field expression_ consists of an expression followed by a single dot and an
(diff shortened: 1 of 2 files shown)


src/expressions.md
```diff
@@ -544,6 +547,41 @@ let x: i32 = add(1i32, 2i32);
...
+T::size_of()
+
+// infer trait `SizeOf` based on traits in scope.
+<T>::size_of()
+
+// completely unambiguous; works if multiple in-scope traits define `size_of()`
+<T as SizeOf>::size_of()
+```
+
+Refer to [RFC 132] for further details, motivations, and subtleties of syntax.
```
matthewjasper: Is this temporary? [RFC 1636](https://github.com/rust-lang/rfcs/blob/master/text/1636-document_all_features.md) seems to say that RFCs aren't supposed to be stable documentation.

mrhota: Hm, I want to keep the link to the RFC. I need more explanation before the link, though. As that RFC says, the reference should be good enough for normal users to learn how to use a feature without referring to the RFC. Thanks for pointing that RFC out; I'd forgotten about it!

mrhota:
Note to self: Needs more and better examples

src/expressions.md
```diff
@@ -544,6 +547,41 @@ let x: i32 = add(1i32, 2i32);
...
+### Unambiguous Function Call Syntax
+
+Several situations often occur which result in ambiguities about the receiver or
+referent of method or associated function calls. These situations may include:
+
+* Multiple in-scope traits define the same method for the same types
+* Auto-`deref` is undesirable; for example, distinguishing between methods on a
+  smart pointer itself and the pointer's referent.
+* Methods which take no arguments and return properties of a type, like
+  `SizeOf::size_of()`
```
frewsxcv: What is `SizeOf`?

mrhota: I took the example from the RFC. It became out-of-date, apparently! I'll correct it.

src/expressions.md
```diff
@@ -544,6 +547,63 @@ let x: i32 = add(1i32, 2i32);
...
+### Unambiguous Function Call Syntax (UFCS)
+
+Several situations often occur which result in ambiguities about the receiver or
+referent of method or associated function calls. These situations may include:
+
+* Multiple in-scope traits define the same method for the same types
+* Auto-`deref` is undesirable; for example, distinguishing between methods on a
+  smart pointer itself and the pointer's referent
+* Methods which take no arguments, like `default()`, and return properties of a
+  type, like `size_of()`
```
frewsxcv: How could `default()` result in an ambiguity?

frewsxcv:
This seems alright to me, seems like it _could_ be a little more detailed with some of the specifics of the syntax, but I still feel comfortable landing it in its current state. Any else have thoughts? @steveklabnik @GuillaumeGomez @QuietMisdreavus

GuillaumeGomez:
Feel the same: could be more detailed.

QuietMisdreavus:
It might be worth mentioning the `<SomeType as SomeTrait>::some_function()` details in prose as well, instead of just leaving it to code examples? Even if you just add one sentence that includes that one snippet, I like seeing a generalized form somewhere.

src/expressions.md
mrhota:
```rust
fn main() {
    let a = Default::default();
    println!("{:?}", a);
}
```

You can (must) use a type annotation or UFCS, right?

mrhota:
@GuillaumeGomez @QuietMisdreavus  @frewsxcv I'd love to add more examples and detail, but I'm afraid the RFC itself seems somewhat out-of-date (`SizeOf`???) and a bit lacking in context. Seems like it sort of leaps into the topic midstream...

If y'all have any examples I could include, I'd appreciate it! I'll see what I can come up with tonight.

QuietMisdreavus:
I was just wanting another sentence or two describing the full form, and mentioning that either `SomeType` or `SomeTrait` can be inferred most of the time, but that UFCS is there when you need both.

src/expressions.md
```diff
@@ -544,6 +547,63 @@ let x: i32 = add(1i32, 2i32);
 let pi: Result<f32, _> = "3.14".parse();
 ```
 
+Function calls may sometimes result in ambiguities about receiver or referent.
+See [UFCS](#unambiguous-function-call-syntax) for how to resolve this.
+
+### Unambiguous Function Call Syntax (UFCS)
```
steveklabnik: So, it's been the opinion of the language team for a long time that "UFCS" is a terrible name and shouldn't be used. It's also not exactly a separate feature, it's that the usual way of calling trait methods is sugar for this syntax. As such, I'm not sure the framing of this PR is right, though there is no way for you to have known that, of course. ☹️ 

I am tempted to merge it as such, and then, we'll end up doing another pass later that cleans this up. This is okay with me, but I don't know how you feel about it.

mrhota: oh that's helpful. with that in mind, maybe I can have a sub-section [here](https://doc.rust-lang.org/nightly/reference/expressions.html#call-expressions) instead of the new section I created? I can describe roughly what you said: "UFCS" confuses the issue, and the RFC didn't add a separate feature. It merely generalized some syntax and turned the then-current syntax into sugar for the fully-qualified function-call syntax. FQFC.

iono @steveklabnik, I'll push some updates, and you can tell me what you think.

steveklabnik: that sounds good to me

mrhota:
@steveklabnik let me know what you think of the description now

src/expressions.md
```diff
@@ -220,6 +220,9 @@ exact `self`-type of the left-hand-side is known, or dynamically dispatching if
 the left-hand-side expression is an indirect [trait
 object](types.html#trait-objects).
 
+[UFCS](#unambiguous-function-call-syntax) allows method invocation using
```
steveklabnik: this link is still wrong, no?

mrhota:
oh yeah. bigly. thx for heads up

mrhota:
@steveklabnik @QuietMisdreavus ok, fixed that last link (and rewrote the sentence around it).

steveklabnik:
Great, thanks!

PR #48, opened 2017-04-25
Added links to descriptions of XID_start and XID_continue in identifiers section.

VinithKrishnan:


src/identifiers.md
```diff
@@ -19,6 +19,12 @@ that does _not_ occur in the set of [keywords].
 > character ranges used to form the more familiar C and Java language-family
 > identifiers.
 
+Links to Unicode descriptions:
```
mrhota: Perhaps it would be nicer to use references like the `[keywords]` and/or `[^non_ascii_idents]` lines below. You wouldn't need this block; the external links would live on the text itself.

VinithKrishnan: Thanks, for the suggestion.I have made the changes.

mrhota:
@VinithKrishnan I think it's ok to leave the backticks around the words, inside the brackets.

```
[`XID_start`] blah blah

[`XID_start`]: http://www.example.com

steveklabnik:
Thank you so much!

PR #51, opened 2017-05-02
Add anchors to split chapters on landing page

### LINKED ISSUE #50: Provide a migration page from the old single page to new split reference pages format
Since Rust Reference underwent the split by chapters, all the links in the wild, like in answers on StackOverflow, are effectively dead, because they don't lead to the relevant parts of Reference anymore. 

I realise the problems of handling url fragments, but couldn't a migration page be generated? Like http://doc.rust-lang.org/reference.html#tuple-types should go to a place in the page which says "This section has been moved to https://doc.rust-lang.org/reference/types.html#tuple-types", and so on for every anchor in the old reference page.

This would be much more helpful, as searching whatever chapters possibly contain the relevant data manually isn't quite easy.

albel727:
This is intended to fix #50, at least temporarily.

I've dug out the last non-split version of the reference and, with a help of a small Ruby script, generated a landing page, which superficially resembles the old reference, but instead of the actual contents, links to the chapters of the new reference.

This approach relies on the Markdown->HTML converter generating anchoring ids for section headers in the particular backward-compatible way, as well as on unchanging section headers in the new reference, i.e.

1. If the markdown generator changes in the future so that its header_text->anchor_id generation no longer matches the old behaviour, then the page becomes useless. This can be worked around by using a pre-generated static html page instead of Markdown.
2. If sections are renamed in the new reference, the fragment parts of redirection links on the landing page would no longer match. But at least, the links would still lead to the correct chapter pages.

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -1,4 +1,479 @@
 % The Rust Reference has moved
+We've split up the reference into chapters. Please find it at its new home [here](reference/index.html).
 
-We've split up the reference into chapters. Please find it at its new
-home [here](reference/index.html).
+# Introduction
+This section has been moved [here](reference/introduction.html#introduction).
+
+# Notation
+This section has been moved [here](reference/notation.html#notation).
+
+## Unicode productions
+This section has been moved [here](reference/unicode-productions.html#unicode-productions).
+
+## String table productions
+This section has been moved [here](reference/string-table-productions.html#string-table-productions).
+
+# Lexical structure
+This section has been moved [here](reference/lexical-structure.html#lexical-structure).
+
+## Input format
+This section has been moved [here](reference/input-format.html#input-format).
+
+## Identifiers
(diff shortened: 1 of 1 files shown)


steveklabnik:
> This approach relies on the Markdown->HTML converter generating anchoring ids for section headers in the particular backward-compatible way, as well as on unchanging section headers in the new reference, i.e.

I forget if rustdoc does this, actually. But...

> This can be worked around by using a pre-generated static html page instead of Markdown.

I wonder if this approach just isn't better, because then it will always work, and it's not like there's heavy development going on here.

> If sections are renamed in the new reference, the fragment parts of redirection links on the landing page would no longer match. But at least, the links would still lead to the correct chapter pages.

This will probably happen, but at least that's not too bad, and we can always fix it.

Anyway, let's go the HTML route. Can you run `rustdoc` on this file, and change the PR to have the HTML instead? Thanks for this.

PR #53, opened 2017-05-12
Add the loop_break_value feature to the reference

pietroalbini:
I added to the reference the changes made by the `loop_break_value` feature, which was accepted to be stabilized in rust-lang/rust#37339. As far as I can tell I added everything needed, but this is the first time I add something to the reference so it's quite possible I missed something!

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -564,7 +564,7 @@ Significantly, lambda expressions _capture their environment_, which regular
 [function definitions](items.html#functions) do not. The exact type of capture
 depends on the [function type](types.html#function-types) inferred for the
 lambda expression. In the simplest and least-expensive form (analogous to a
-```|| { }``` expression), the lambda expression captures its environment by
+`|| { }` expression), the lambda expression captures its environment by
 reference, effectively borrowing pointers to all outer variables mentioned
 inside the function.  Alternately, the compiler may infer that a lambda
 expression should copy or move values (depending on their type) from the
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -600,6 +600,11 @@ within this loop may exit out of this loop or return control to its head.
 See [break expressions](#break-expressions) and [continue
 expressions](#continue-expressions).
 
+A `loop` expression is [diverging](items.html#diverging-functions), and doesn't
+return anything. However, when it contains a [break
+expression](#break-expressions), it returns the value attached to the `break`
+that caused the loop to stop.
+
 ## `break` expressions
 
 A `break` expression has an optional _label_. If the label is absent, then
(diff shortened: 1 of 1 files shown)


pietroalbini:
The return value is `!` when there's no break in the loop. If there's a break in it though, the return value can't be `!` anymore because the loop can stop. In this case, before the return value was always `()`, and now it's the value provided to the `break`, which defaults to `()` anyway.

src/expressions.md
```diff
@@ -600,6 +600,12 @@ within this loop may exit out of this loop or return control to its head.
 See [break expressions](#break-expressions) and [continue
 expressions](#continue-expressions).
 
+The return value of a `loop` expression is `!`, which can be coerced to any
```
Havvy: Instead of saying that `!` can be coerced to any type here, perhaps just link to the section on diverging functions.

src/expressions.md
```diff
@@ -600,6 +600,12 @@ within this loop may exit out of this loop or return control to its head.
...
+type. However, if there is a `break` expression in the loop, the return value
+of the loop becomes the one of the value provided to the [break
+expression](#break-expressions) causing the loop to end, which is `()` by
```
Havvy: The comma there is improper English. I think it'd flow better with the information in its own sentence though. Perhaps "When the break expression has no return value it defaults to `()`."

src/expressions.md
```diff
@@ -600,6 +600,12 @@ within this loop may exit out of this loop or return control to its head.
 See [break expressions](#break-expressions) and [continue
 expressions](#continue-expressions).
 
+The return value of a `loop` expression is `!`, which can be coerced to any
+type. However, if there is a `break` expression in the loop, the return value
```
Havvy: Does the `break` expression have to also break for the loop it's in? For instance, I would expect `'inner` to still be considered diverging here? I'm not actually sure though.

```rust
'outer loop {
  'inner loop {
    break 'outer;
  }
}
```

pietroalbini: I didn't write the implementation for the feature so I'm not sure, but I *guess* `'inner` is not diverging, and has a return value of `()`, because it actually returns.

src/expressions.md
pietroalbini: Thanks! I'm not a native speaker so I'm still used for the italian idioms.

src/expressions.md
```diff
@@ -608,6 +613,14 @@ enclosing it. It is only permitted in the body of a loop. If the label is
 present, then `break 'foo` terminates the loop with label `'foo`, which need not
 be the innermost label enclosing the `break` expression, but must enclose it.
 
+When the `break` expression is enclosed in a `loop`, it has an optional return
```
Havvy: Both commas in this sentence can be removed.

src/expressions.md
```diff
@@ -608,6 +613,14 @@ enclosing it. It is only permitted in the body of a loop. If the label is
...
+value attached, which will be returned by the loop ended by the
+expression. When it's not provided it defaults to `()`.
+
+If both a label and a return value are present in the expression, the label
+must be put before the return value. For example `break 'label 42` breaks the
```
Havvy: Change `put` to `placed`, I think?

dhardy:
Continued in #56.

pietroalbini:
Thanks for improving this @dhardy!

PR #54, opened 2017-05-16
Add diverging loops to diverging function section.

Havvy:
I went with "corresponding break" here since you could have a
break that's inside another loop expression for the inner loop
expression, but that wouldn't change the divergence. That probably
needs to be better fleshed out, but I'd leave the fleshing out
of divergence to the loop item section of the reference.


steveklabnik:
Thanks!

PR #55, opened 2017-05-16
Document RFC 1219: `use` group `as`

matthewjasper:
Changes this section to make it clear that this is possible.
[RFC 1219 text](https://github.com/rust-lang/rfcs/blob/master/text/1219-use-group-as.md) for reference.
cc #9


steveklabnik:
Thank you!

PR #57, opened 2017-05-18
Add union items to the reference

petrochenkov:
Copy-pasted from https://github.com/rust-lang/rfcs/pull/1897

### DIFF
--- a/src/items.md
+++ b/src/items.md
@@ -21,6 +21,7 @@ There are several kinds of item:
 * [type definitions](#type-aliases)
 * [struct definitions](#structs)
 * [enumeration definitions](#enumerations)
+* [union definitions](#unions)
 * [constant items](#constant-items)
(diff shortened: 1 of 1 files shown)


steveklabnik:
Unions are not yet stable, so we'll have to wait until they are to merge.

petrochenkov:
@steveklabnik 
So, the stabilization PR (https://github.com/rust-lang/rust/pull/42068) can land before this one? Ok.

steveklabnik:
Yup! Sending in this PR is enough to satisfy the requirement of stabilizing the feature. Then it lands, then we land tihs.

frewsxcv:
Just reran Travis and this passes now. The contents of the PR look good to me. Seems good to merge?

steveklabnik:
Yup!

PR #58, opened 2017-05-19
Minor fixes for type aliases

Havvy:
Fix a link; remove the word "Currently". The Rust Reference shouldn't be documenting hypothetical changes, so everything should be "currently".


steveklabnik:
Thanks!

PR #13, opened 2017-03-15
[WIP] Expressions rewrite

matthewjasper:
So this is a WIP attempt to improve the expressions section. It documents more features, fixes inaccuracies and hopefully makes things clearer. All feedback is welcome.

## Changes

* General
    * Always wrap at 80 characters
    * Make more things links
* Structs
    * State that struct expressions can't be used in control flow expressions
* Methods
    * Add examples for methods
    * Explained auto-deref
        * <s>UFCS (or whatever it's called now) isn't in the reference!!!</s> It has a PR now.
* Fields
    * Show the syntax for calling a struct field
* Calls
    * Move before operators
    * Explain the function call syntax
    * Mention `Fn` traits
* Lambdas
    * Move lambda section
    * Lambdas have unique types
    * `move` is part of the lambda expression according to libsyntax, so I'm going with that
    * Explain that `move` can extend the lifetime of the closure
    * Use closure, except when followed by 'expression'
    * Closure traits depend on how captures are used, not on how they are captured
* Arrays
    * Explain what arrays expressions do
    * `static`s aren't constant expressions
    * Explain `Copy` requirement
* Index
    * Specify how overloading is done
    * Indices must be of type `usize`
* Unary operators
    * Explained `std::ops::Neg`
    * Hopefully made it clearer how `Deref` works
    * Explained `std::ops::Not`
    * Say that `&mut` can't be applied to everything
* Binary operators
    * Merge with unary operators
    * Built in operators don't use traits
    * Show all of the operators in examples
    * Put all of this into a table, to avoid having to repeat so many words.
* Compound assignments
    * Explain overloading of these operators
* Precedence
    * Add unary operators
    * Show the correct associativity
* Grouped expressions
    * Better example
* Loops and control flow.
    * Use 'then' instead of and for sequenced operations in `while` loops
    * State that all loops return `()` (Except `loop` but that's in another PR)
    * Add some more examples
    * Don't use 'identical' for things that aren't (would also need to look at #47 in this respect)
    * Make the definition of `if let` and `while let` clearer

Update:

* Further Changes
    * Mention array to slice coercions happening in method calls.
    * Use 'mutable lvalue (context)' to describe some things. (I have no idea if this is how it should be described, but this was the easiest way to describe the behaviour that I tested).
    * Explain when lvalues can be moved out of. (or, how not to do a move semantics tutorial)
    * Explain that most things are actually in `libcore`, in only one place.
    * Explicitly say that blocks are rvalues.
[...]

PR #65, opened 2017-06-01
Fix expr links

matthewjasper:
This should fix the remaining broken links due to #13. Also makes a reference to temporary variables point to the appropriate section.


steveklabnik:
Wonderful, thank you so much.

PR #69, opened 2017-06-06
Statements do not have types.

eddyb:
They have no value that can be used in any way, so we cannot talk about the "type of a statement".


steveklabnik:
Thanks ❤️

PR #71, opened 2017-06-21
ignore doctest of `extern "stdcall"` on non-x86

arielb1:
The "stdcall" ABI only exists on x86, so this causes tests to fail on
other architectures. This should fix rust-lang/rust#40260 next time we
update the reference.


steveklabnik:
Thank you!

PR #78, opened 2017-07-05
Clarify that breaking the Pointer Aliasing Rules requires an access

allanlw:
This clarifies that you must access a pointer to violate LLVM's Pointer Aliasing Rules.

From a discussion today on #rust with @ubsan and @rkruppe


ubsan:
This is a good change, imo.

steveklabnik:
Thanks!

PR #79, opened 2017-07-05
Clarify that address leaks aren't considered unsafe

allanlw:
This explicitly adds leaking of randomized base addresses as not being considered unsafe.

From a discussion on #rust today with @ubsan and @rkruppe


ubsan:
This doesn't feel necessary to explicitly say; it's implicit in the language with safe casts and `{:p}`.

allanlw:
@ubsan with the exception of integer overflow, I think that could be said of everything in this list. Clearly this page cannot list all possible bad code patterns that rust does not consider "unsafe", but presumably it should include ones that can be surprising.

ubsan:
@allanlw alright, fair.

PR #94, opened 2017-08-04
Document lifetime of temporaries in conditions of if and while-expr

PieterPenninckx:
Explain that temporaries created in the condition expression of an `if` or `if`/`else` expression are dropped immediately after the condition expression. Similar for `while`. See rust-lang/rust#12033

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -89,15 +89,21 @@ innermost enclosing statement; the tail expression of a block is considered
 part of the statement that encloses the block.
 
 A first exception is when a temporary value is created in the
-condition expression of an `if` or an  `if`/`else` expression.
-In this case, the lifetime ends right after the condition expression.
+condition expression of an `if` or an  `if`/`else` expression or in the
+loop conditional expression of a `while` expression.
+In this case, the lifetime ends right after the condition expression or the
+loop conditional expression.
 
-Here is an example:
+Here are some examples:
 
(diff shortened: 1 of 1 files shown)


src/expressions.md
```diff
@@ -88,9 +88,26 @@ created and used instead. The lifetime of temporary values is typically the
 innermost enclosing statement; the tail expression of a block is considered
 part of the statement that encloses the block.
 
-When a temporary rvalue is being created that is assigned into a `let`
-declaration, however, the temporary is created with the lifetime of the
-enclosing block instead, as using the enclosing statement (the `let`
+A first exception is when a temporary value is created in the
```
Havvy: The phrasing here seems odd and opens the possibility that there are more than two exceptions. Perhaps a paragraph stating "There are two exceptions to this rule." prior to the exceptions listings?

PieterPenninckx: You are right: it is odd. I don't know if there are still some exceptions missing, however.

Can someone confirm that there are no other exceptions missing? Then we can add "There are two exceptions to this rule."

ubsan: I'd argue that this is not an exception - the only exception is the `let` rule. It's just that the conditional of an `if` or `while` are counted as a single statement for the purposes of the rule.

Also, side note, you don't have to do the whole taking a reference thing in order to create a temporary object. If a pvalue is not assigned to an lvalue, or is not used to construct an object, then a temporary object is automatically created.

steveklabnik:
Thanks!

PR #100, opened 2017-08-17
Divide items and expressions into sections

brauliobz:
* [Rendered](https://brauliobz.github.io/rust-reference-sections/)
 
* To concat `items.md` for comparison purposes:

`cat src/items.md src/items/{modules,extern-crates,use-declarations,functions,type-aliases,structs,enumerations,unions,constant-items,static-items,traits,implementations,external-blocks}.md > src/items.new.md`

* To concat `expressions.md` for comparison purposes:

`cat src/expressions.md src/expressions/{literal-expr,path-expr,tuple-expr,struct-expr,enum-variant-expr,block-expr,method-call-expr,field-expr,call-expr,closure-expr,array-expr,range-expr,operator-expr,loop-expr,if-expr,match-expr,return-expr}.md > src/expressions.new.md`

* The following sections were moved inside others and won't match in the diff above: *grouped expressions, while let, if let, tuple indexing*.

* I (hopefully) fixed all affected links.


Havvy:
@steveklabnik @QuietMisdreavus @GuillaumeGomez @frewsxcv 

This is a relatively major structural change. I'm not sure how to do the checkbox thingy - so, are you all good with this change?

I think it's a good change personally.

frewsxcv:
Sounds good to me, though I have a feeling we'll need to update links in the Rust repository as well, maybe also nomicon

brauliobz:
A `rg 'reference/(items|expressions).html#'` on Rust repo (with all submodules fetched) returns only 4 results:

```text
rust/src/librustc_typeck/diagnostics.rs
793:[Expressions]: https://doc.rust-lang.org/reference/expressions.html#lvalues-rvalues-and-temporaries

rust/src/libcore/ops/deref.rs
43:/// [ref-deref-op]: ../../reference/expressions.html#the-dereference-operator
130:/// [ref-deref-op]: ../../reference/expressions.html#the-dereference-operator

rust/src/libcore/ops/mod.rs
153://! [operator precedence]: ../../reference/expressions.html#operator-precedence
```

I didn't find any links on nomicon and rust-by-example.

Havvy:
We need coordination with the stdlib docs to update things properly. I don't know how to do the coordination required here.

frewsxcv:
> We need coordination with the stdlib docs to update things properly. I don't know how to do the coordination required here.

The approach is to merge this first, then bump the git submodule in rust-lang/rust in a follow-up PR.

@brauliobz FYI, looks like there's merge conflicts here

steveklabnik:
Yup, going to do that right now. Thanks!

steveklabnik:
FWIW this also did break inter-reference stuff:

```
reference\expressions\field-expr.html:149: broken link fragment `#method-call-expressions` pointing to `reference\expres
sions\field-expr.html`
reference\expressions\path-expr.html:150: broken link fragment `#unsafe-blocks` pointing to `reference\expressions\path-
expr.html`
reference\interior-mutability.html:149: broken link fragment `#shared-references` pointing to `reference\types.html`
reference\interior-mutability.html:161: broken link - std\sync\atomic.html
```

I'm fixing it though 😄

brauliobz:
> I'm fixing it though 😄

Thanks!

steveklabnik:
Thanks for improving the reference; this kind of stuff is *so* needed :)

PR #101, opened 2017-08-26
Types chapter cleanup and new sections

matthewjasper:
Changes to existing sections (first commit):
* Avoid using 'type constructor'
* Give a definition and more complete list of primitive types
* Use 'slice' for the unsized type `[T]`
* Don't use `Vec<T>` in the slice example
* Use the term 'variant', instead of 'variant constructor'
* Remove description of  error E0073 (now removed) when describing rules for recursive types.
* Say some more about references and pointers (including RFC 1135)
* Use `compile_fail` for examples that show compiler errors
* Mention that structs and variants also create function item types
* Try to make it clearer that/why `Fn` can be implemented by `move` closures
* List object safety rules (RFC 255)
* Remove an incorrect statement about which type is the receiver.

Additional sections (second commit):
* Mutable references
* DSTs (includes documenting RFC 490, RFC 546)
* Interior mutability
* Unions

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -3,21 +3,44 @@
 Every variable, item and value in a Rust program has a type. The _type_ of a
 *value* defines the interpretation of the memory holding it.
 
-Built-in types and type-constructors are tightly integrated into the language,
-in nontrivial ways that are not possible to emulate in user-defined types.
-User-defined types have limited capabilities.
+Built-in types are tightly integrated into the language, in nontrivial ways
+that are not possible to emulate in user-defined types. User-defined types have
+limited capabilities.
 
 ## Primitive types
 
-The primitive types are the following:
+Some types are defined by the language, rather than as part of the standard
+library, these are called _primitive types_. Some of these are induvidual
+types:
 
 * The boolean type `bool` with values `true` and `false`.
-* The machine types (integer and floating-point).
-* The machine-dependent integer types.
-* Arrays
-* Tuples
-* Slices
-* Function pointers
+* The [machine types] (integer and floating-point).
+* The [machine-dependent integer types].
+* The [textual types] `char` and `str`.
+
+There are also some primitive constructs for generic types built in to the
+language:
+
+* [Tuples]
+* [Arrays]
+* [Slices]
+* [Function pointers]
+* [References]
+* [Pointers]
+
+[machine types]: #machine-types
+[machine-dependent integer types]: #machine-dependent-integer-types
+[textual types]: #textual-types
+[Tuples]: #tuple-types
+[Arrays]: #array-and-slice-types
+[Slices]: #array-and-slice-types
+[References]: #pointer-types
+[Pointers]: #raw-pointers
+[Function pointers]: #function-pointer-types
+[function]: #function-types
+[closure]: #closure-types
+
+## Numeric types
 
 ### Machine types
 
@@ -52,14 +75,14 @@ The types `char` and `str` hold textual data.
 
 A value of type `char` is a [Unicode scalar value](
 http://www.unicode.org/glossary/#unicode_scalar_value) (i.e. a code point that
-is not a surrogate) represented as a 32-bit unsigned word in the 0x0000 to
-0xD7FF or 0xE000 to 0x10FFFF range. A `[char]` array is effectively a UCS-4 /
-UTF-32 string.
+is not a surrogate), represented as a 32-bit unsigned word in the 0x0000 to
+0xD7FF or 0xE000 to 0x10FFFF range. A `[char]` is effectively a UCS-4 / UTF-32
+string.
 
 A value of type `str` is a Unicode string, represented as an array of 8-bit
-unsigned bytes holding a sequence of UTF-8 code points. Since `str` is of
-unknown size, it is not a _first-class_ type, but can only be instantiated
-through a pointer type, such as `&str`.
+unsigned bytes holding a sequence of UTF-8 code points. Since `str` is a
+[dynamically sized type], it is not a _first-class_ type, but can only be
+instantiated through a pointer type, such as `&str`.
 
 ## Tuple types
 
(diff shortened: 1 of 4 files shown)


matthewjasper:
Also, since one of the rules listed for recursive types was out of date, does anyone know what "Recursive type definitions can cross module boundaries, but not module visibility boundaries, or crate boundaries (in order to simplify the module system and type checker)." is/was supposed to mean?

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+'a)>`. If there's no such bound then the default bound is `'static`, for
+example in `std::rc::Rc<T>`.
+
+
+## Dynamically sized types
+
+Most types have a fixed size that is known at compile time and implement the
+trait [`Sized`]. A type with a size that is known only at run-time is called a
+_dynamically sized type_ (_DST_) or unsized type. [Slices] and [trait objects]
+are two of the most DSTs. Such types can only be used in certain cases:
```
Havvy: s/of the most/examples of/

If you want to say "of the most common", I'd personally suggest against that as it's an empirical fact.

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+
+* [Pointer types] to DSTs are sized but have twice the size of pointers to
+  sized types
+    * Pointers to slices also store the number of elements of the slice.
+    * Pointers to trait objects also store a pointer to a vtable.
+* Traits may be implemented for DSTs, this means that `Self` is not assumed to
```
Havvy: Should definitely be a period instead of a comma after DSTs. I think this can be shortened considerably. Consider the following, where `elsewhere` is a link to wherever talks about the Sized trait and where it is automatically inferred.

"Traits can be implemented for DSTs. Unlike elsewhere `Self: ?Sized` by default in trait definitions."

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+  implement `Sized` in trait definitions by default.
+* DSTs can be used as type parameters with a bound of `?Sized`
```
Havvy: Does markdown support &lt;abbr&gt;? If so, we should use it everywhere DST is used. I think?

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+[variables]: variables.html
+[const]: items.html#constant-items
+[static]: items.html#static-items
+
+## Interior mutability
+
+References prevent a value being both aliased and mutated: mutable references
+don't allow a value to be aliased and shared references prevent a value from
+being mutated. Sometimes this is too restrictive. The type
+[`std::cell::UnsafeCell<T>`](../std/cell/struct.UnsafeCell.html) provides an
```
Havvy: The type should be core::cell::UnsafeCell<T> linking to https://doc.rust-lang.org/nightly/core/cell/struct.UnsafeCell.html

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+unsafe API that allows a type to own a `T` that can be mutated through a shared
+reference to the `UnsafeCell` (there is no change to mutable references).
```
Havvy: The parenthetical should be its own sentence and copy verbiage from the UnsafeCell docs themselves.

"It is undefined behavior to have multiple `&mut UnsafeCell<T>` aliases."

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+
+The standard library provides a variety of different types that safely wrap
+this functionality. For example `std::cell::RefCell<T>` uses run-time borrow
```
Havvy: Nit: Link?

src/types.md
```diff
@@ -428,13 +499,60 @@ Box<Bar<'a>>
...
+checks to ensure the usual rules around multiple references. The
+`std::sync::atomic` module contains types that wrap a value that is only
```
Havvy: Nit: Link?

src/types.md
matthewjasper: The reference usually links to the re-export in std (maybe this should change).

src/types.md
matthewjasper: Link to https://doc.rust-lang.org/std/marker/trait.Sized.html?

src/types.md
steveklabnik: generally speaking, we link to `std` and not `core`.

> (maybe this should change)

Yeah possibly, it's a tough problem.

src/types.md
Havvy: How about https://doc.rust-lang.org/nightly/reference/the-sized-trait.html and a later PR, we can fill in that section?

src/types.md
Havvy: I'm the odd one out on this one. Everybody else thinks it is better to link to std, not core.

src/types.md
```diff
@@ -428,13 +499,68 @@ Box<Bar<'a>>
...
+The default for user-defined types with a trait object used as a type parameter
+is based on the type itself. If a type parameter has a lifetime bound, then
+that lifetime bound becomes the default bound for trait objects of that type.
+For example, `std::cell::Ref<'a, T>` contains a `T: 'a` bound, therefore trait
+objects of type `Ref<'a, SomeTrait>` are the same as `Ref<'a, (SomeTrait +
+'a)>`. If there's no such bound then the default bound is `'static`, for
+example in `std::rc::Rc<T>`.
+
+
+## Dynamically sized types
```
Havvy: The more I look at this section, the more I wonder if this section should be on the Sized trait itself. The other types in this chapter (except `Self`...) are non-overlapping, whereas DSTs are a facet of a type.

matthewjasper:
I've moved the interior mutability and DST sections to their own pages following @Havvy's comment above.
Also the trait object section now has more examples about default lifetimes and I've tried to make the interior mutability section clearer.

Havvy:
Thanks ❤️

PR #111, opened 2017-09-05
Non-ASCII chars in identifiers aren't supported yet

brauliobz:
It's misleading to keep XID_Start and XID_Continue on the reference, since they may never be supported in stable.

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -18,37 +18,4 @@ Or
    * The identifier is more than one character, `_` alone is not an identifier
    * The remaining characters are alphanumeric or `_`
 
-[keywords]: keywords.html
-
 [^non_ascii_idents] Non-ASCII characters in identifiers are currently feature-gated. See [issue #28979](https://github.com/rust-lang/rust/issues/28979).
-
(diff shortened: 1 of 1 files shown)


steveklabnik:
They are still part of Rust, just behind a flag. The reference currently does document unstable things as well.

src/identifiers.md
```diff
@@ -2,10 +2,34 @@
...
+
+Or
+
+   * The first character is `_`
+   * The identifier is more than one character, `_` alone is not an identifier
+   * The remaining characters are alphanumeric or `_`
+
+[keywords]: keywords.html
+
+[^non_ascii_idents] Non-ASCII characters in identifiers are currently feature-gated. See [issue #28979](https://github.com/rust-lang/rust/issues/28979).
```
Havvy: I'd snip everything below this. You can move the comment into the Unstable Book, or it's in the git history. This also involves removing the link from the text above.

brauliobz:
The corresponding [Unstable Book PR](https://github.com/rust-lang/rust/pull/45250).

Havvy:
💟 Thanks.

PR #112, opened 2017-09-05
Inner and outer docs/comments were swapped

brauliobz:


steveklabnik:
Hm, I think it works both ways: that is, `//!` is "outer" because it's attached to the enclosing thing, and `///` is "inner" because it is attached to the next thing.

brauliobz:
It's really confusing... but the adjective refers to the attribute, not the item the attribute applies to. From the parser:

```rust
    /// Parse attributes that appear before an item
    pub fn parse_outer_attributes(&mut self) -> PResult<'a, Vec<ast::Attribute>> {
```

```rust
    pub fn parse_inner_attributes(&mut self) -> PResult<'a, Vec<ast::Attribute>> {
        ...
        // Don't even try to parse if it's not an inner attribute.
        if !self.look_ahead(1, |t| t == &token::Not) {
```

Havvy:
I don't think the naming in the reference is canonical, but probably best to follow what `rustc` does.

💟 Thanks

PR #116, opened 2017-09-17
Remove example mentioning str_eq lang item

leodasvacas:
It's a useless lang item [which we should remove](https://github.com/rust-lang/rust/pull/44658) so not a good example. There is not drop in replacement because I think it was the only lang item that was just an fn.


Havvy:
Thanks 💟

Issue #84, opened 2017-07-20
Document the (complete) grammar inside the specification


brauliobz:
This issue tracks the work in documenting the grammar.

Since [the grammar](https://github.com/rust-lang/rust/blob/master/src/doc/grammar.md) is *very* incomplete and I'm willing to complete it, I thought it would be better to do this inside the reference, alongside each item of the language.

Progress:

- [x] Create a script to join the scattered grammar in one place.
- [x] Lexical productions
  - [x] Keywords PR #97
  - [x] Identifiers PR #89 
  - [x] Comments PR #85
  - [x] Literals
    - [x] Char PR #121
    - [x] String PR #121
    - [x] Raw string PR #121
    - [x] Byte PR #121
    - [x] Byte string PR #121
    - [x] Raw byte string PR #121
    - [x] Integer PR #108
    - [x] Float PR #109 
    - [x] Bools PR #110 
  - [x] Symbols _draft_
  - [x] Paths _draft_
- [x] Syntax productions
  - [x] Crates PR #87
  - [x] Items
    - [x] Visibility PR #90
    - [x] Modules PR #143 
    - [x] Extern crate PR #144 
    - [x] Use declaration PR #145 
    - [x] Static item
    - [x] Const value
    - [x] Functions _draft_
    - [x] Type aliases
    - [x] Enums
    - [x] Structs
    - [x] Unions
    - [x] Extern blocks
    - [x] Traits _draft_
    - [x] Implementations _draft_
    - [x] Const functions
    - [x] Extern functions
  - [x] Attributes PR #88
  - [x] Statements
    - [x] Item declarations
    - [x] `let` statements 
    - [x] Expression statements
  - [x] Expressions _draft_
    - [x] Literal #138 
    - [x] Path
    - [x] Tuple
    - [x] Struct
    - [x] EnumerationVariant
    - [x] Block #136 
    - [x] Method call
    - [x] Field access #158 
    - [x] Tuple indexing
    - [x] Call expression #134 
    - [x] Closure #135 
    - [x] Array expression #133 
    - [x] Indexing expression #133 
    - [x] Range expression #142 
    - [x] Operator expression #141 
      - [x] borrow
      - [x] dereference
      - [x] error propagation
      - [x] negation
      - [x] arithmetic and logical
      - [x] comparison
      - [x] lazy boolean
      - [x] typecast
      - [x] assignment
      - [x] compound assignment
    - [x] Grouped expression #141 
[...]

PR #121, opened 2017-09-24
Add grammar for char, string, byte, and raw literals

brauliobz:
Part of #84.

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -21,12 +21,23 @@ evaluated (primarily) at compile time.
 
 |                                              | Example         | `#` sets   | Characters  | Escapes             |
 |----------------------------------------------|-----------------|------------|-------------|---------------------|
-| [Character](#character-literals)             | `'H'`           | `N/A`      | All Unicode | [Quote](#quote-escapes) & [Byte](#byte-escapes) & [Unicode](#unicode-escapes) |
-| [String](#string-literals)                   | `"hello"`       | `N/A`      | All Unicode | [Quote](#quote-escapes) & [Byte](#byte-escapes) & [Unicode](#unicode-escapes) |
+| [Character](#character-literals)             | `'H'`           | `N/A`      | All Unicode | [Quote](#quote-escapes) & [ASCII](#ascii-escapes) & [Unicode](#unicode-escapes) |
+| [String](#string-literals)                   | `"hello"`       | `N/A`      | All Unicode | [Quote](#quote-escapes) & [ASCII](#ascii-escapes) & [Unicode](#unicode-escapes) |
 | [Raw](#raw-string-literals)                  | `r#"hello"#`    | `0...`     | All Unicode | `N/A`                                                      |
 | [Byte](#byte-literals)                       | `b'H'`          | `N/A`      | All ASCII   | [Quote](#quote-escapes) & [Byte](#byte-escapes)                               |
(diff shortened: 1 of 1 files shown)


ehuss:
Now that [underscores are accepted in unicode escapes](https://github.com/rust-lang/rust/pull/43716) in Rust 1.22, should the UNICODE_ESCAPE production include underscores?  Maybe something like this?

```
UNICODE_ESCAPE :
    \u{ HEX_DIGIT _* }
  | \u{ HEX_DIGIT _* HEX_DIGIT _* }
  | \u{ HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* }
  | \u{ HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* }
  | \u{ HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* }
  | \u{ HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* HEX_DIGIT _* }
```

brauliobz:
> should the UNICODE_ESCAPE production include underscores?  Maybe something like this?

Cool! I wasn't aware of that :). Yes, that would be the way I would include it in the grammar.

So... I am against including this right now. The problem in including this change is that I'm testing the grammar on **stable** to avoid having such a fast moving target and to avoid including in it things that maybe will took too long or never reach stable at all (since in my understanding lots of things on nightly are experimental).

Also, correct me if I am wrong: the Rust 1.22 you talk about is actually **nightly**, right? It doesn't mean this will be released on Rust 1.22, right? The underscore in Unicode escapes isn't available even in **beta** yet (Rust 1.21.beta3).

ehuss:
> Rust 1.22 you talk about is actually nightly, right?

Nightly, yes.  

BTW, do you want some help on this?  I'd be willing to help if you could point me to things to add/fix.  I also might have some questions about the grammar you have so far, and am curious how you're testing it.

Havvy:
The underscores in Unicode escapes is insta-stable (there's no feature gate) and the reference rides the release trains, so it should be included in this grammar. It can be done in a separate PR if that would help.

brauliobz:
@Havvy 
> LGTM, minus the underscores from other comments

I've added them. Shouldn't they already be on beta, then? Or is it too soon?

@ehuss
> BTW, do you want some help on this? I'd be willing to help if you could point me to things to add/fix.

Yeah, that would be great. But I have too many things half done... I  will need to take a look at what I have and commit it.

> I also might have some questions about the grammar you have so far

I'm on IRC sometimes, on #rust-docs .

> and am curious how you're testing it.

I'm reading the rustc's parser/lexer, making an Antlr4 grammar out of it, and making some test cases and testing them against my grammar and rustc.

brauliobz:
Now it is a little too long. I didn't want to use {n, m} notation but maybe

> `\u{` ( HEX_DIGIT `_`* )<sup>{1,6}</sup> `}`

would be a better way to do this? Or even

> `\u{` ( HEX_DIGIT `_`* )<sup>1..6</sup> `}`

Havvy:
Yeah. It's literally one character too long when I render it, and spilling onto the next line in a confusing way. 😕 I don't care whether it's <sup>`{n,m}`</sup> or <sup>`n..m`</sup>, or even <sup>`1 to 6 repetitions`</sup> just as long as it conveys what is going on. It might be useful to describe the notation used by the grammar in `lexical-structure.html` including any interesting things you use.

Havvy:
💟 Thanks.

PR #122, opened 2017-09-26
Add best-effort warning header

### LINKED ISSUE #118: Add Best-Effort Warning at top of page
Add a warning to the top of the page detailing that the reference is currently at a best-effort stage, like the warning for the second edition Rust-book (see image).

Text from Rust book 2nd edition with warning class

The text of this warning should contain the following or similar text, but with proper linking:

"For now, the Rust reference is a best-effort document. We strive for validity and completeness, but are not yet there. At some point in the future, the docs and lang team will work together to figure out how best to do this. Until then, this document is a best-effort attempt. If you find something wrong or missing, file an issue or send in a pull request."

Havvy:
Fixes #118.

Because we modify index.hbs, we should occasionally compare with
mdbook's own copy and rebase our changes onto it.

picture of rust reference with warning header

### DIFF
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -1,0 +1,227 @@
+<!DOCTYPE HTML>
+<html lang="{{ language }}">
+    <head>
+        <meta charset="UTF-8">
+        <title>{{ title }}</title>
+        <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
+        <meta name="description" content="{{ description }}">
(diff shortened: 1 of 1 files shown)


Havvy:
cc @steveklabnik Since this relates to my PR on TRPL, and you merged that.

steveklabnik:
Thanks!

I will get to your other PRs either later today or tomorrow :)

PR #124, opened 2017-09-29
The unit value is not a literal, it is a tuple expression

brauliobz:
As can be seen [on the AST](https://github.com/rust-lang/rust/blob/51cd06170eccf91a2d93018e939473a8db18da92/src/libsyntax/ast.rs#L1114).


Havvy:
💟 Thanks.

I was a bit hesitant to merge this because I wasn't sure whether or not unit is considered a literal for Rust the language &ndash; while it's easy to encode as a tuple in the compiler, it doesn't have to be a tuple in the language. But there's other wording in the reference that says it is a tuple, so this makes it consistent.

PR #126, opened 2017-10-02
Small fixes to the expressions section

### LINKED ISSUE #107: expressions precedence
Hello,

I don't find any documentation about rust expression precedence rules in the book and reference. Could someone point me to it if it exists? If not, should it be added to the reference? Related forum [post](https://users.rust-lang.org/t/precedence-in-rust/1385) and rust issue [#21799](https://github.com/rust-lang/rust/issues/21799) and [#44288](https://github.com/rust-lang/rust/issues/44288).

Explanation:
In cases where expression parsing is ambiguous, there must be some expression precedence rules. For operator expressions it is documented in the reference as _operator precedence_. But for general expressions it is unclear, in ambiguous cases, which expression type takes precedence over the other. Two examples: `0..10.some_method()` (range expr vs method call expr), `&some_struct.some_method()` (operator expr vs method call expr).

Thank you

### LINKED ISSUE #120: Improve Constant Expression and Static Item sections
The `Constant Expression` section currently contains a bulleted list of which expressions are constant and contains one interesting line:

> Paths to statics, so long as only their address, not their value, is used. This includes using their value indirectly through a complicated expression. *

That asterisk says that's only true for static items. As such, it's sort of out of place in the `Constant Expression` section. Instead, the Static Item section should say that it allows constant expressions plus referencing statics by address. The second sentence is also hard to read, and should have an example with it. It should also say "complex" not "complicated".

matthewjasper:
A number of small changes to the expressions section.

* Extend the operator precedence section to be expression precedence (Fixes #107)
* Improve constant expressions (Fixes #120)
* Don't (incorrectly) imply that one can move out of the fields of a boxed struct separately.
* Mention that methods that aren't visible aren't used in type resolution.
* Expressions bound to patterns are always lvalues (so can be unsized) even when moved from.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -15,6 +15,35 @@ In this way, the structure of expressions dictates the structure of execution.
 Blocks are just another kind of expression, so blocks, statements, expressions,
 and blocks again can recursively nest inside each other to an arbitrary depth.
 
+## Expression precedence
+
+The precedence of Rust operators and expressions is ordered as follows, going
+from strong to weak. Binary Operators at the same precedence level are
+evaluated in the order given by their associativity.
+
+| Operator/Expression         | Associativity       |
+|-----------------------------|---------------------|
+| Paths                       |                     |
+| Method calls                |                     |
+| Field expressions           | left to right       |
+| Function calls, array indexing |                  |
+| `?`                         |                     |
+| Unary `-` `*` `!` `&` `&mut` |                    |
+| `as` `:`                    | left to right       |
+| `*` `/` `%`                 | left to right       |
+| `+` `-`                     | left to right       |
+| `<<` `>>`                   | left to right       |
+| `&`                         | left to right       |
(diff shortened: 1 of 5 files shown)


src/expressions.md
```diff
@@ -23,25 +52,24 @@ or _rvalue context_. The evaluation of an expression depends both on its own
 category and the context it occurs within.
 
 An lvalue is an expression that represents a memory location. These expressions
-are paths which refer to local variables, function and method arguments, or
-static variables, [dereferences]&nbsp;(`*expr`), [array indexing] expressions
-(`expr[expr]`), [field] references (`expr.f`) and parenthesized lvalue
-expressions. All other expressions are rvalues.
+are [paths](#path-expressions) which refer to local variables, static variables
```
Havvy: Missing comma after `static variables`.

Havvy:
💟 Thanks.

PR #127, opened 2017-10-03
Rename glossory to glossary (o->a)

Havvy:


QuietMisdreavus:
Should we check for links that may point here, or would it prudent to do that checking when it's merged upstream?

Havvy:
It's a relatively new addition to the reference, so I doubt anybody has linked to it from the stdlib and definitely not the other books.

Havvy:
This is going to cause merge conflicts with #128 one way or another, and also conflicts with a PR I want to submit, but am waiting on this one for.

steveklabnik:
@Havvy ping on this!

Havvy:
Rebased.

PR #128, opened 2017-10-08
Document destructors/drop.

matthewjasper:
cc rust-lang/rust#43034

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -1,0 +1,93 @@
+# Destructors
+
+When an [initialized]&#32;[variable] in Rust goes out of scope or a [temporary]
+is no longer needed its _destructor_ is run. [Assignment] also runs the
+destructor of its left-hand operand, unless it's an unitialized variable. If a
+[struct] variable has been partially initialized, only its initialized fields
+are dropped.
+
+The destrutor of a type consists of
+
+1. Calling its [`std::ops::Drop::drop`] method, if it has one.
+2. Recursively running the destructor of all of its fields.
+    * The fields of a [struct], [tuple] or [enum variant] are dropped in
+      declaration order. \*
+    * The elements of an [array] or owned [slice][array] are dropped from the
+      first element to the last. \*
+    * The captured values of a [closure] are dropped in an unspecified order.
+    * [Trait objects] run the destructor of the underlying type.
+    * Other types don't result in any further drops.
+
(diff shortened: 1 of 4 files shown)


src/destructors.md
```diff
@@ -0,0 +1,95 @@
...
+    let uninitialized: ShowOnDrop;
+    // Only first element drops
+    let mut partially_initialized: (ShowOnDrop, ShowOnDrop);
+    partially_initialized.0 = ShowOnDrop("Partial tuple first");
+}
+```
+
+## Not running destructors
+
+Not running destructors in Rust is safe, so `unsafe` code cannot assume a
```
Havvy: The entire `unsafe` middle clause feels unneeded. There's a section devoted to behaviour considered safe that includes leaking memory (though it can be better documented there).

src/destructors.md
```diff
@@ -0,0 +1,95 @@
...
+variable will be dropped, even if it has a type that isn't `'static`.
+[`std::mem::forget`] can be used to avoid running the destructor of a local
+variable. [`std::mem::ManuallyDrop`] provides a wrapper to prevent a field
```
Havvy: I'd rather this just say that ManuallyDrop provides a convenient wrapper to prevent a variable or field from being dropped automatically. While it's used for avoiding recursive drops, it's used for other reasons as well. E.g. `forget` actually just puts the value into a `ManuallyDrop` and then lets Rust drop it without calling the `drop` method.

src/destructors.md
```diff
@@ -0,0 +1,95 @@
...
+      declaration order. \*
+    * The elements of an [array] or (owned) [slice][array] are dropped from the
+      first element to the last. \*
+    * The captured values of a [closure] are dropped in an unspecified order.
+    * [Trait objects] run the destructor of the underlying type.
+    * Other types don't result in any further drops.
+
+\* This order was stabilized in [RFC 1857].
+
+Variables are dropped in reverse order of declaration. Variables declared in
```
Havvy: I think this should be a part of `block expressions`, but I'm not going to worry about that for this PR.

Havvy:
Apparently I don't get a notification when there are new commits. I'll look at this when I get home.

Havvy:
💟 Thanks! And 🙁 sorry for not noticing the second commit sooner.

Issue #129, opened 2017-10-11
Items must show what paths they make available.


Havvy:
Every item should declare what paths it adds. We can say that the canonical path for an item is its *path prefix* followed by `::` followed by the item's *path component*. Modules and implementations change the path prefix, while everything except implementations define new path components.

- [ ] constants
- [ ] enumerations
- [ ] extern crates
- [ ] external blocks
- [ ] functions
- [ ] implementations
- [ ] modules
- [ ] statics
- [ ] structs
- [ ] traits
- [ ] type aliases
- [ ] unions
- [ ] use declarations

ehuss:
Posted #1052 with what I think should close this out.

PR #132, opened 2017-10-13
Canonical Paths

Havvy:
Defines canonical paths, as part of #129.

Note to reviewer(s): There's also two commits that just change the location of links and change where newlines are, so just focus on the giant green blob in the paths file.

### DIFF
--- a/src/paths.md
+++ b/src/paths.md
@@ -5,10 +5,6 @@ a namespace qualifier (`::`). If a path consists of only one component, it may
 refer to either an [item] or a [variable] in a local control
 scope. If a path has multiple components, it refers to an item.
 
-Every item has a _canonical path_ within its crate, but the path naming an item
-is only meaningful within a given crate. There is no global namespace across
-crates; an item's canonical path merely identifies it within the crate.
-
 Two examples of simple paths consisting of only identifier components:
 
 ```rust,ignore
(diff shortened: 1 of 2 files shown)


src/paths.md
```diff
@@ -103,3 +92,79 @@ mod a {
...
+the crate.
+
+```rust
+// Comments show the canonical path of the item.
+
+mod a { // ::a
+    pub struct Struct; // ::a::Struct
+
+    pub trait Trait { // ::a::Trait
+        fn f(&self); // <::a::Trait>::f
```
petrochenkov: `<::a::Trait>::f` -> `::a::Trait::f`.
`<::a::Trait>::f` has another meaning - inherent associated item on the trait object type `::a::Trait`.

Havvy: I'm going with the assumption you are right, but I thought `::a::Trait::f` was syntactic sugar for `<::a::Trait>::f`.

src/paths.md
```diff
@@ -103,3 +92,79 @@ mod a {
...
+
+## Canonical paths
+
+Items defined in a module or implementation have a *canonical path* that
+corresponds to where within its crate it is defined. All other paths to these
+items are aliases. The canonical path is defined as a *path prefix* appended by
+the path component the item itself defines.
+
+[Implementations] and [use declarations] do not have canonical paths, although
+the associated items that implementations define do have them. Items defined in
```
steveklabnik: i wonder if "associated items" is too overloaded a term here, it made me do a double-take at first.

src/paths.md
```diff
@@ -103,3 +92,79 @@ mod a {
...
+block expressions do not have canonical paths. Items defined in a module that
+does not have a canonical path do not have a canonical path. Associated items
+defined in an implementation that refers to an item without a canonical path,
+e.g. as the implementing type, the trait being implemented, a type parameter or
+bound on a type parameter, do not have canonical paths.
+
+The path prefix for modules is the canonical path to that module. For bare
+implementations, it is the canonical path of the item being implemented
+surrounded by angle (`<>`) brackets. For trait implementations, it is the
+cananonical path of the item being implemented followed by `as` followed by the
```
brauliobz: cananonical -> canonical

Havvy: Fixed.

src/paths.md
Havvy: It's technically extraneous, as implementations only define associated items. I removed it; does it flow better that way?

PR #133, opened 2017-10-14
Array and indexing grammar

brauliobz:
Also, add the information that the indexing expressions can be made available in other types by implementing `Index` and `IndexMut`.

Part of #84.

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -54,6 +54,9 @@ _panicked state_ if it fails.
 ```rust,should_panic
 ([1, 2, 3, 4])[2];        // Evaluates to 3
 
+let b = [[1, 0, 0], [0, 1, 0], [0, 0, 1]];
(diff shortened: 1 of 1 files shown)


src/expressions/array-expr.md
```diff
@@ -50,4 +61,14 @@ let y = (["a", "b"])[n];  // panics
 
 let arr = ["a", "b"];
 arr[10];                  // panics
+
+# let b = [[1, 0, 0], [0, 1, 0], [0, 0, 1]];
```
Havvy: Since we show that `arr[10]` panics, we should probably show this array being defined or else somebody may wonder why `b[1][2]` doesn't panic.

brauliobz: Done. Also moved it up so it isn't grouped alongside the panic examples.

Havvy:
💟 Thanks!

PR #137, opened 2017-10-14
If and if let grammar

brauliobz:
Part of #84.

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -2,6 +2,14 @@
 
 ## `if` expressions
 
+> **<sup>Syntax</sup>**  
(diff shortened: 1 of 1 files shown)


src/expressions/if-expr.md
```diff
@@ -2,6 +2,14 @@
 
 ## `if` expressions
 
+> **<sup>Syntax</sup>**  
+> _IfExpression_ :  
+> &nbsp;&nbsp; `if` [_Expression_]<sub>_except struct expression_</sub> [_BlockExpression_]
```
Havvy: Can we link to the struct expression here?

Havvy:
💟 Thanks!

PR #139, opened 2017-10-14
Loops grammar

brauliobz:
Part of #84.

### DIFF
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -5,21 +5,21 @@
 > &nbsp;&nbsp; [_LoopLabel_]<sup>?</sup> (  
 > &nbsp;&nbsp; &nbsp;&nbsp; &nbsp;&nbsp; [_InfiniteLoopExpression_]  
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_PredicateLoopExpression_]  
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_PredicatePatternLoopExpression_]  
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_IteratorLoopExpression_]  
-> &nbsp;&nbsp; &nbsp;&nbsp; | [_WhileLetExpression_]  
 > &nbsp;&nbsp; )  
 
 [_LoopLabel_]: #loop-labels
 [_InfiniteLoopExpression_]: #infinite-loops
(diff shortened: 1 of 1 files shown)


src/expressions/loop-expr.md
```diff
@@ -44,9 +67,14 @@ while i < 10 {
...
 
+> **<sup>Syntax</sup>**  
+> [_WhileLetExpression_] :  
+> &nbsp;&nbsp; `while` `let` _Pattern_ `=` [_Expression_]<sub>except struct expression</sub>
+>              [_BlockExpression_]  
+
 A `while let` loop is semantically similar to a `while` loop but in place of a
 condition expression it expects the keyword `let` followed by a refutable
-pattern, an `=` and an expression. If the value of the expression on the right
+pattern, an `=` and a block expression. If the value of the expression on the right
```
matthewjasper: ...an `=`, _an expression_ and a block expression.

src/expressions/loop-expr.md
```diff
@@ -44,10 +67,15 @@ while i < 10 {
 
 ## `while let` loops
 
+> **<sup>Syntax</sup>**  
+> [_WhileLetExpression_] :  
+> &nbsp;&nbsp; `while` `let` _Pattern_ `=` [_Expression_]<sub>except struct expression</sub>
```
Havvy: All of the other loop expressions are named after what they do. So shouldn't this one be called `PredicatePatternLoopExpression`?

src/expressions/loop-expr.md
```diff
@@ -61,6 +89,11 @@ while let Some(y) = x.pop() {
 
 ## Iterator loops
 
+> **<sup>Syntax</sup>**  
+> _IteratorLoopExpression_ :  
+> &nbsp;&nbsp; `for` _Pattern_ `in` [_Expression_]<sub>except struct expression</sub>
```
Havvy: Like `if`, we probably want to link to struct expressions here.

src/expressions/loop-expr.md
brauliobz: Done

src/expressions/loop-expr.md
brauliobz: Still deciding how I'll do this (the "no struct expression" thing) to make it clear. But yeah, I agree a link to it will be needed.

Havvy:
💟 Thanks!

PR #141, opened 2017-10-14
Operator expressions grammar

brauliobz:
Part of #84.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -85,8 +85,8 @@ let mut array = [-2, 3, 9];
 }
 ```
 
-For ease of typing, consecutive borrow expressions can be expressed by the
-single token `&&` as if they were separated:
+Even though `&&` is a single token ([the lazy 'and' operator](#lazy-boolean-operators)),
+when used in the context of borrow expressions it works as two borrows:
 
 ```rust
 // same meanings:
@@ -94,8 +94,9 @@ let a = &&  10;
 let a = & & 10;
 
 // same meanings:
-let a = &&  mut 10;
-let a = & & mut 10;
+let a = &&&&  mut 10;
+let a = && && mut 10;
+let a = & & & & mut 10;
 ```
 
 ## The dereference operator
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -1,5 +1,18 @@
 # Operator expressions
 
+> **<sup>Syntax</sup>**  
+> _OperatorExpression_ :  
+> &nbsp;&nbsp; &nbsp;&nbsp; [_BorrowExpression_]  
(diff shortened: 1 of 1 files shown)


src/expressions/operator-expr.md
```diff
@@ -1,5 +1,18 @@
 # Operator expressions
 
+> **<sup>Syntax</sup>**  
+> _OperatorExpression_ :
```
Havvy: Missing GroupedExpression.

src/expressions/operator-expr.md
```diff
@@ -36,6 +53,11 @@ assert_eq!(y, 20);
 
 ## Borrow operators
 
+> **<sup>Syntax</sup>**  
+> _BorrowExpression_ :   
+> &nbsp;&nbsp; &nbsp;&nbsp; (`&`|`&&`) [_Expression_]
```
Havvy: Why the `&&` variant here?

brauliobz: Because the lexer will give us `&&` when it finds two contiguous ampersands, since `&&` is a valid symbol. The same problem occurs with `<<` and `>>` in generics/paths, and `||` in closures. Example:

```rust
let & & mut a = & & mut 10;
let &&  mut a = &&  mut 10;
```

I'm currently very busy, but I'll put some explanation in the next days, since it isn't really obvious why `&&` is accepted.

brauliobz: Done

src/expressions/operator-expr.md
brauliobz: Grouped expression will be located directly on the _Expression_ production. I think it would be out of place if put together with the OperatorExpressions -- even though it is in the same section. I put it in the operator expression section mainly because I think it is too small to be on a separate section by itself.

What do you think?

Havvy: I think grammar should be associated with its prose. GroupedExpression is on this page, so it should show up here. If it feels out of place in the grammar, it should feel out of place on the prose and moved to its own page, albeit the shortness is concerning. Although some examples could help with that.

src/expressions/operator-expr.md
```diff
@@ -60,8 +82,25 @@ let mut array = [-2, 3, 9];
 }
 ```
 
+For ease of typing, consecutive borrow expressions can be expressed by the
```
Havvy: Ease of typing compared to what? To most people reading the reference, it would seem obvious that `&&` is allowed and can be chained just like e.g. `--1` works (though is useless). Though I'm guessing that the `x && y` operator giving an `&&` token meaning is to blame here?

mrhota:
@brauliobz ping! what's the status?

brauliobz:
@mrhota Hey, I'm back! I'm rebasing my branches and after that I'll deal with this PR.

Havvy:
💟 Thanks!

PR #142, opened 2017-10-14
Range expressions grammar

brauliobz:
Part of #84.

### DIFF
--- a/src/expressions/range-expr.md
+++ b/src/expressions/range-expr.md
@@ -1,7 +1,35 @@
 # Range expressions
 
+> **<sup>Syntax</sup>**  
+> _RangeExpression_ :  
+> &nbsp;&nbsp; &nbsp;&nbsp; _RangeExpr_  
+> &nbsp;&nbsp; | _RangeFromExpr_  
+> &nbsp;&nbsp; | _RangeToExpr_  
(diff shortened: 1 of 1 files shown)


src/expressions/range-expr.md
```diff
@@ -1,7 +1,35 @@
...
+> _RangeFullExpr_ :  
+> &nbsp;&nbsp; `..`  
+
 The `..` operator will construct an object of one of the `std::ops::Range` (or
-`core::ops::Range`) variants.
+`core::ops::Range`) variants, according to the following table:
+
+| Production             | Syntax        | Type                         | Range                 |
+|------------------------|---------------|------------------------------|-----------------------|
+| _RangeExpr_            | start`..`end  | [std::ops::Range]            | start &le; x &lt; end |
```
Havvy: Might be worth using &lt;pre&gt; tags to preserve the formatting in the Range column?

Havvy:
💟 Thanks!

src/expressions/range-expr.md
brauliobz: Do you mean to keep the &le;, x and &lt; aligned (like they are in the source code)?

Havvy: Yes.

PR #146, opened 2017-10-26
Implementations improvements part 1

### LINKED ISSUE #5: Document Orphan Rules
Because, uh, they aren't documented.

Havvy:
Trying to not focus on what it means to associate a type, nor deal
too much with generic impls at this point. Those will come in future parts.

Right now everything about associated types are defined in `items/traits.md`. I didn't touch that here.

Fixes #5. Albeit, I'm not a fan of the wording which I just pulled directly from the RFC.

cc @scottmcm for review.

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -7,8 +7,9 @@ the structure of the program when the compiler is compiling it.
 
 ### Arity
 
-Arity refers to the number of arguments a function or operation takes.
-For example, `f(2, 3)` and `g(4, 6)` have arity 2, while `h(8, 2, 6)` has arity 3.
+Arity refers to the number of arguments a function or operator takes.
+For some examples, `f(2, 3)` and `g(4, 6)` have arity 2, while `h(8, 2, 6)`
+has arity 3. The `!` operator has arity 1.
 
 ### Array
 
@@ -19,9 +20,8 @@ at run time by the program. It occupies a contiguous region of memory.
 ### Associated Item
 
 An associated item is an item that is associated with another item. Associated
-items are defined in [implementations] and [traits]. Only functions, constants,
-and types can be associated.
-
+items are defined in [implementations] and declared in [traits]. Only functions,
+constants, and type aliases can be associated.
 
 ### Bound
 
@@ -61,8 +61,8 @@ can create such an lvalue without initializing it.
 
 ### Nominal Types
 
-Types that can be named directly. Specifically [enums], [structs], [unions],
-and [trait objects].
+Types that can be referred to by a path directly. Specifically [enums],
+[structs], [unions], and [trait objects].
 
 ### Prelude
 
@@ -100,9 +100,9 @@ Strings slices are always valid UTF-8.
 ### Trait
 
 A trait is a language item that is used for describing the functionalities a type must provide.
-It allow a type to make certain promises about its behavior.
+It allows a type to make certain promises about its behavior.
 
-Generic functions and generic structs can exploit traits to constrain, or bound, the types they accept.
+Generic functions and generic structs can use traits to constrain, or bound, the types they accept.
 
 [enums]: items/enumerations.html
 [structs]: items/structs.html
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -8,13 +8,20 @@ the structure of the program when the compiler is compiling it.
 ### Arity
 
 Arity refers to the number of arguments a function or operation takes.
-For example, `(2, 3)` and `(4, 6)` have arity 2, and`(8, 2, 6)` has arity 3.
+For example, `f(2, 3)` and `g(4, 6)` have arity 2, while `h(8, 2, 6)` has arity 3.
 
 ### Array
 
 An array, sometimes also called a fixed-size array or an inline array, is a value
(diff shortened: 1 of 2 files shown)


src/items/implementations.md
```diff
@@ -37,33 +91,43 @@ impl Shape for Circle {
...
+
+   ```
+    T = C
+      | [T]
+      | [T, ..n]
+      | &T
+      | &mut T
+      | Box<T>
+      | (..., T, ...)
+      | X<..., T, ...> where X is not bivariant with respect to T
```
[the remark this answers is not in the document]
Havvy: This was changed from the RFC then? I was wondering how `(..., T, ...)` would have worked...

src/glossary.md
```diff
@@ -8,14 +8,21 @@ the structure of the program when the compiler is compiling it.
...
 
 An array, sometimes also called a fixed-size array or an inline array, is a value
 describing a collection of elements, each selected by an index that can be computed
 at run time by the program. It occupies a contiguous region of memory.
 
+### Associated Item
+
+An associated item is an item that is associated with another item. Associated
+items are defined in [implementations] and [traits]. Only functions, constants,
+and types can be associated.
```
scottmcm: Hmm, are types associated, or are type aliases associated?  You can't put an `enum Foo {}` in an `impl` to satisfy a `type Foo;` in the trait, right?

Is it worth distinguishing between associated items are "declared in traits" and "defined in impls"?  The declared/defined words are used in trait impls below.

src/items/implementations.md
```diff
@@ -1,10 +1,62 @@
...
+type declarations, a path to a nomial tyupe, a where clause, and a bracketed
+set of associable items.
+
+The nominal type is called the *implementing type* and the associable items are
+the *associated items* to the implementing type.
+
+Bare implementations associate the associated items to the implementing type.
+
+The associated item has a path of a path to the implementing type followed by
+the associate item's path component.
```
scottmcm: Maybe note that a bare impl cannot have associated type( aliase)s.

src/glossary.md
```diff
@@ -52,6 +59,11 @@ A variable is initialized if it has been assigned a value and hasn't since been
 moved from. All other lvalues are assumed to be initialized. Only unsafe Rust
 can create such an lvalue without initializing it.
 
+### Nominal Types
+
+Types that can be named directly. Specifically [enums], [structs], [unions],
+and [trait objects].
```
scottmcm: Is it worth saying that "named" is "referred to by a path", or similar?

Perhaps also add Structural Types, and a cross-ref between them for contrast?

src/items/implementations.md
```diff
@@ -1,10 +1,62 @@
...
 Implementations are defined with the keyword `impl`.
 
+## Bare Implementations
+
+A bare implementation is defined as the sequence of the `impl` keyword, generic
+type declarations, a path to a nomial tyupe, a where clause, and a bracketed
+set of associable items.
+
+The nominal type is called the *implementing type* and the associable items are
+the *associated items* to the implementing type.
```
scottmcm: Hmm, libcore has `impl<T> [T] {`, where that slice isn't a nominal type.  Users cannot define structural types, though, so they can't do something like that...

Havvy: That definitely seems like a special case. Not entirely sure how to document that without messing up the flow for everything else.

src/items/implementations.md
Havvy: 👍 I had told myself to do that a million times, yet I did not do it.

src/glossary.md
Havvy: Worth doing, but in its own PR.

src/glossary.md
```diff
@@ -7,15 +7,22 @@ the structure of the program when the compiler is compiling it.
...
 
 ### Array
 
 An array, sometimes also called a fixed-size array or an inline array, is a value
 describing a collection of elements, each selected by an index that can be computed
 at run time by the program. It occupies a contiguous region of memory.
 
+### Associated Item
+
+An associated item is an item that is associated with another item. Associated
```
steveklabnik: I'm okay with merging this, but this definition *is* a bit circular; we should figure out a better way to describe this.

src/items/implementations.md
```diff
@@ -1,10 +1,64 @@
...
-An _implementation_ is an item that can implement a [trait](items/traits.html) for a
-specific type.
+An _implementation_ is an item that associates items with an *implementing type*.
+
+There are two types of implementations: Bare implementations and [trait] 
+implementations.
 
 Implementations are defined with the keyword `impl`.
 
+## Bare Implementations
```
steveklabnik: Aren't these called "inherent implementations"?

src/items/implementations.md
steveklabnik: (review note, this is resolved below)

src/items/implementations.md
Havvy: I've seen them called both. Inherent is probably a better name though. Will fix tonight.

src/glossary.md
Havvy: I'll do this as a part of Part 2.

steveklabnik:
Thanks!

PR #147, opened 2017-10-27
Fix 'Glossory' typo

Rantanen:
Done in two commits, in case you don't want to rename the file for link breakage reasons.

Yes, it's a very high effort PR... happy Hacktoberfest, I guess? :)


Havvy:
This is literally PR #127

Rantanen:
Oh oops, that's a really good point.

Made the assumption that at least the page title would have been fixed already if pointed out so didn't even check. Sorry!

PR #153, opened 2017-11-05
Remove unstable and 'in the future' things.

### LINKED ISSUE #119: Remove Unstable Documentation
As per #111, we don't yet have a story on whether we want to document unstable aspects of Rust, and if we do, how do we want to do so?

Havvy:
The `Compiler Features` section should be moved to the Unstable book.

I've currently commented out the section on Type Coercions because I cannot parse what it's saying right now to check if it's up to date. (Yay, sleepiness.) If anybody else wants to tell me what that is saying, I'd be grateful.

There were two other changes made in this PR, just because I was there. The first is updating what the fake type shown by rustc is for function item types. The second is that I added attributes on unions and added repr to it. This is mainly to make me not forget about unions when I get to making a representations section.

Fixes #119.

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -142,8 +142,8 @@ Coercion is allowed between the following types:
     - coerce_inner(`T`) = `U` where `T` is a concrete type which implements the
     trait `U`.
 
-    In the future, coerce_inner will be recursively extended to tuples and
+    <!--In the future, coerce_inner will be recursively extended to tuples and
     structs. In addition, coercions from sub-traits to super-traits will be
-    added. See [RFC 401] for more details.
+    added. See [RFC 401] for more details.-->
 
 * Non capturing closures to `fn` pointers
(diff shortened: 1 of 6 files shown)


src/type-coercions.md
```diff
@@ -142,8 +142,8 @@ Coercion is allowed between the following types:
     - coerce_inner(`T`) = `U` where `T` is a concrete type which implements the
     trait `U`.
 
-    In the future, coerce_inner will be recursively extended to tuples and
+    <!--In the future, coerce_inner will be recursively extended to tuples and
     structs. In addition, coercions from sub-traits to super-traits will be
-    added. See [RFC 401] for more details.
+    added. See [RFC 401] for more details.-->
```
matthewjasper: Should add the third point from [Unsize](https://doc.rust-lang.org/std/marker/trait.Unsize.html).

PR #157, opened 2017-11-22
Rename C-like enum to Field-less enum

notriddle:
There is no need to reference the C programming language to explain this concept.

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -36,12 +36,14 @@ In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
 called an enum variant. Each enum instance has a _discriminant_ which is an
 integer associated to it that is used to determine which variant it holds.
 
-## C-like Enumerations
+## Custom Discriminant Values for Field-Less Enumerations
 
-If there is no data attached to *any* of the variants of an enumeration it is
-called a *c-like enumeration*. If a discriminant isn't specified, they start at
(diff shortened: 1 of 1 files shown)


steveklabnik:
As I said in the book PR, for now, this is the official name. Please get involved over here: https://internals.rust-lang.org/t/proposal-bikeshed-rename-c-like-enumeration/6209

steveklabnik:
@notriddle so sorry here, i completely forgot that we decided in the meeting to send in PRs, and then make the final decision that way. Gah. I'm sorry.

@rust-lang/lang, we need a call here. Thoughts?

src/items/enumerations.md
```diff
@@ -36,12 +36,14 @@ In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
...
 
-## C-like Enumerations
+## Custom Discriminant Values for Field-Less Enumerations
 
-If there is no data attached to *any* of the variants of an enumeration it is
-called a *c-like enumeration*. If a discriminant isn't specified, they start at
-zero, and add one for each variant, in order. Each enum value is just its
-discriminant which you can specify explicitly:
+If there is no data attached to *any* of the variants of an enumeration,
+then the discriminant can directly chosen and accessed.
```
Havvy: "can directly" -> "can be directly"

Havvy:
💟 Thanks!

PR #166, opened 2017-12-01
Document RFC 1440.

Havvy:
AFAIK, this was stabilized in what is currently beta, so the reference will be out of date for six weeks.

I'm not entirely happy with what I've written, but I'm also not altogether happy with how the static and const sections are currently written, and will go back at some point to organize them a bit better. Likewise with destructors.

### DIFF
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -5,29 +5,22 @@
 > &nbsp;&nbsp; `static` `mut`<sup>?</sup> [IDENTIFIER] `:` [_Type_]
 >              `=` [_Expression_] `;`
 
-A *static item* is similar to a *constant*, except that it represents a precise
+A *static item* is similar to a [constant], except that it represents a precise
 memory location in the program. A static is never "inlined" at the usage site,
 and all references to it refer to the same memory location. Static items have
 the `static` lifetime, which outlives all other lifetimes in a Rust program.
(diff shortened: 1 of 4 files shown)


src/items/static-items.md
```diff
@@ -5,30 +5,23 @@
...
 
-A *static item* is similar to a *constant*, except that it represents a precise
+A *static item* is similar to a [constant], except that it represents a precise
 memory location in the program. A static is never "inlined" at the usage site,
 and all references to it refer to the same memory location. Static items have
 the `static` lifetime, which outlives all other lifetimes in a Rust program.
-Static items may be placed in read-only memory if they do not contain any
-interior mutability.
+Static items may be placed in read-only memory if the type is not [interior 
+mutable]. Static items do not call `drop` at the end of the program.
```
QuietMisdreavus: Have we ever used "interior mutable" as an adjective before? I always thought the term was always "interior mutability", as a noun.

Havvy: I don't think we do, but it is a natural analog to *interior mutability*.

QuietMisdreavus: It's a shame "interiorally" isn't a word, otherwise that's what i would suggest. With our conversation on IRC, i'm willing to call this okay. Farther bikeshedding of how to decline "interior mutability" can occur elsewhere.

PR #167, opened 2017-12-02
Rename lvalue and rvalue to place and value expressions

Havvy:
This effort is already underway in the compiler as well, and this is the only other place we talk about these two kinds of expressions.

I also moved a lot of links down to the bottom of the page. I already painstakingly checked that every link was correct.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -44,76 +44,84 @@ evaluated in the order given by their associativity.
 | `=` `+=` `-=` `*=` `/=` `%=` <br> `&=` <code>&#124;=</code> `^=` `<<=` `>>=` | right to left |
 | `return` `break` closures   |                     |
 
-## Lvalues and rvalues
+## Place Expressions and Value Expressions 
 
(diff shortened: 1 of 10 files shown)


matthewjasper:
The files in `src/expressions` and the glossary (definition of initialized) also use 'lvalue' a lot.

Havvy:
Alright, I've updated those pages as well. I even did a grep to see if there were more uses, and didn't find any. And once again, I've painstakingly checked every link I modified to make sure they are correct.

Havvy:
Oh, right, almost forgot. Some usages of `lvalue` were to describe the actual *memory location* or *place* the expression refers to. I used both terms, rather haphazardly, in this case and I'm not sure which I should be using. Though really, `a placement expression evaluates to a place` sort of doesn't define `place` either.

We could continue to be wishy-washy about this and do something about it in another PR. Probably in the memory model section.

steveklabnik:
This looks good to me! Gonna need a rebase though; you're doing too much amazing work too quickly!

Havvy:
Rebased.

PR #168, opened 2017-12-09
Type parameters have slightly different method resolution

matthewjasper:
For example, this is allowed:
```rust
fn f<T: ::std::ops::Deref>(t: T) {
     t.deref();
}
```


Havvy:
💟 Thanks.

PR #169, opened 2017-12-10
Remove unused link reference.

Havvy:
This was added by the PR I just approved, since I didn't want to wait another cycle between editing and reviewing when this PR would be faster to make.


Havvy:
cc @QuietMisdreavus or @steveklabnik for what should be a 5-second review.

PR #173, opened 2017-12-13
Clean up the theme files now that header.hbs is a thing.

Havvy:
No more having to track index.hbs between updates. \o/

There was an extra book.css in src/theme as far as I could tell.

There's no more .has_warning class, so we need to make sure to remove the first style in reference.css when we remove that warning.


Havvy:
Oh yes, this also requires that Rust is using mdbook 0.0.28 or later.

matthewjasper:
Looks like Rust is currently using 0.0.26, so this will need an upgrade there first. Otherwise this lgtm, but maybe someone more familiar with mdbook should have a look.

https://github.com/rust-lang/rust/blob/0b90e4e8cd068910f604f3e1fb5d03cc01f1658f/src/tools/rustbook/Cargo.toml#L11

steveklabnik:
merged this in manually; doing the update to 0.1 now!

PR #177, opened 2017-12-17
Better describe the scope of items defined in item declaration statements

### LINKED ISSUE #25: Clarify visibility of items in functions
Migrated from https://github.com/rust-lang/rust/issues/38193

---------------------------------------------------------------------

In Rust it is possible to define items inside functions. It is natural to ask whether it's possible to import these items with a `use` statement out of the function, like so:

```
pub fn func() {
    pub struct Struct;
}

use func::Struct;
```

The Rust Reference does not give any explicit answer on whether such a construct is valid. In particular, it says "Every item has a canonical path within its crate," which suggests the path of `Struct` in the example above is `func::Struct`, and it doesn't indicate such items are private. Only by running the code I could tell that this construct is not valid.

I suggest the Rust Reference clarify either in its discussion of paths, functions, or item statements that it's impossible to import an item from an item statement.

Havvy:
Fixes #25.

Also, given https://github.com/rust-lang/rust/issues/45251, I removed the part where whether outer-level control flow expressions are expression statements without semicolons depends on what the expression evaluates into.

### DIFF
--- a/src/paths.md
+++ b/src/paths.md
@@ -1,9 +1,9 @@
 # Paths
 
-A _path_ is a sequence of one or more path components _logically_ separated by
-a namespace qualifier (`::`). If a path consists of only one component, it may
-refer to either an [item] or a [variable] in a local control
-scope. If a path has multiple components, it refers to an item.
+A *path* is a sequence of one or more path components _logically_ separated by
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
Thanks! :sparkles:

PR #178, opened 2017-12-17
Add method, inherent {impl, fn, method, const}, trait impl to glossary.

Havvy:
I'm told that it's hard to find anywhere that actually defines inherent method by @sheeldotme on IRC, so let's add it to the glossary.

r? @sheeldotme (not that your r+ allows a merge on its own)

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -59,11 +59,26 @@ and functions that evaluate to a single value, with or without side-effects.
 
 For example, `2 + (3 * 4)` is an expression that returns the value 14.
 
+### Inherent Implementation
+
+An implementation that associates items directly to a type. These associated
+items are also called inherent: *inherent method*, *inherent function*,
+*inherent constant*.
+
+[More][inherent implementations].
+
+Contrast with [*trait implementation*][#trait-implementation].
+
 ### Initialized
 
 A variable is initialized if it has been assigned a value and hasn't since been
 moved from. All other memory locations are assumed to be initialized. Only
 unsafe Rust can create such a memory without initializing it.
+
+### Method
+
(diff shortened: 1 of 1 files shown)


src/glossary.md
```diff
@@ -125,17 +140,32 @@ Strings slices are always valid UTF-8.
...
+A trait is a language item that is used for describing the functionalities a
+type must provide. It allows a type to make certain promises about its behavior.
 
-Generic functions and generic structs can use traits to constrain, or bound, the types they accept.
+Generic functions and generic structs can use traits to constrain, or bound,
+the types they accept.
+
+[More][traits].
+
+### Trait Implementation
```
sheeldotme: The wording here can be challenging for newcomers.  The confusion comes primarily(in my case), from the following two places: "associates items through a trait" and "the trait bound of the trait".  With multiple reads, I do understand what was intended.  

In the trait definition "promises" and "behaviour" are referenced, so something along the lines of the following may be easier to understand, with minimal information loss:

An implementation that associates items that will fulfill the promises a given trait makes.

An implementation that associates the items that fulfill the the behaviour requirements of a given trait.

src/glossary.md
```diff
@@ -59,12 +59,27 @@ and functions that evaluate to a single value, with or without side-effects.
...
+
 ### Initialized
 
 A variable is initialized if it has been assigned a value and hasn't since been
 moved from. All other memory locations are assumed to be initialized. Only
 unsafe Rust can create such a memory without initializing it.
 
+### Method
+
+An associated function that takes `Self`, `&Self`, `&mut Self` or `Box<Self>` as
```
matthewjasper: Methods are defined by the first parameter being called `self`, not just the type of the first parameter.

PR #182, opened 2017-12-26
Documented the `#[repr(align(x))]` attribute.

bitshifter:
Updated reference for stabilization of the `#[repr(align(x))]` attribute. See tracking issue comments https://github.com/rust-lang/rust/issues/33626#issuecomment-348467804 for discussion.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -264,12 +264,28 @@ For all other enumerations, the layout is unspecified.
 
 Likewise, combining two primitive representations together is unspecified.
 
+### The `align` Representation
+
+The `align` representation can be used on `struct`s and `union`s to raise the
+alignment of the type to a given value.
+
+Alignment is specified as a parameter in the form of `#[repr(align(x))]`. The
+alignment value must be a power of two of type `u32`. The `align` representation
+can raise the alignment of a type to be greater than it's primitive alignment,
+it cannot lower the alignment of a type.
+
+The `align` and `packed` representations cannot be applied on the same type and
+a `packed` type cannot transitively contain another `align`ed type.
+
 ### The `packed` Representation
 
 The `packed` representation can only be used on `struct`s and `union`s.
 
 It modifies the representation (either the default or `C`) by removing any
 padding bytes and forcing the alignment of the type to `1`.
+
+The `align` and `packed` representations cannot be applied on the same type and
+a `packed` type cannot transitively contain another `align`ed type.
 
 > Warning: Dereferencing an unaligned pointer is [undefined behaviour] and it is
 > possible to [safely create unaligned pointers to `packed` fields][27060].
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -13,6 +13,7 @@
 > _MetaItem_ :  
 > &nbsp;&nbsp; &nbsp;&nbsp; IDENTIFIER  
 > &nbsp;&nbsp; | IDENTIFIER `=` LITERAL  
+> &nbsp;&nbsp; | IDENTIFIER `(` LITERAL `)`  
 > &nbsp;&nbsp; | IDENTIFIER `(` _MetaSeq_ `)`  
 > &nbsp;&nbsp; | IDENTIFIER `(` _MetaSeq_ `,` `)`  
 >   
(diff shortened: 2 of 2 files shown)


Havvy:
I thought I removed most of the verbiage of the `repr` attribute from the attributes page. Whoops.

You'll instead want to add a section ```The `align` representation` to `type-layout.md`.

bitshifter:
Ah OK, should I leave the text I added to `attributes.md` since it's already talking about the other `repr` attributes?

Havvy:
It's better to just link to the full description of the representation attribute from the attributes.md page, so as to avoid duplicating information too much. I just forgot to clean up the attributes.md page when the type layout page landed.

src/type-layout.md
```diff
@@ -264,13 +264,23 @@ For all other enumerations, the layout is unspecified.
 
 Likewise, combining two primitive representations together is unspecified.
 
-### The `packed` Representation
+### `align(x)` Representation
```
Havvy: This should be called ```The `align` Representation```.

src/type-layout.md
```diff
@@ -264,13 +264,23 @@ For all other enumerations, the layout is unspecified.
...
+
+The `align(x)` representation can be used on `struct`s and `union`s to raise the
+alignment of the type to the given value. The alignment value must be a power of
+two of type `u32`. The `align` representation cannot lower the alignment of a
+type.
+
+### `packed` Representation
```
Havvy: You got rid of the `The` here.

src/type-layout.md
```diff
@@ -264,13 +264,23 @@ For all other enumerations, the layout is unspecified.
...
 
 The `packed` representation can only be used on `struct`s and `union`s.
 
 It modifies the representation (either the default or `C`) by removing any
 padding bytes and forcing the alignment of the type to `1`.
 
+The `align` and `packed` representations cannot be applied on the same type and
```
Havvy: This section should also be included in the align representation section.

src/type-layout.md
```diff
@@ -264,13 +264,23 @@ For all other enumerations, the layout is unspecified.
 
 Likewise, combining two primitive representations together is unspecified.
 
-### The `packed` Representation
+### `align(x)` Representation
+
+The `align(x)` representation can be used on `struct`s and `union`s to raise the
```
Havvy: The paragraph structure here should mirror the packed representation section. So which categories of types it can be applied to should be one paragraph, and what it does should be another.

Havvy: Remove the `(x)` from the name. Instead, describe how it takes the alignment as a parameter in the paragraph describing how it works.

bitshifter:
Thanks @Havvy I think I've addressed your feedback. This stuff hasn't been stabilized quite yet so you might want to wait until that happens before accepting the PR I guess.

Havvy:
Github shows that you did a force push, but I don't actually see any changes in the new commit. I even checked your branch and there's no difference there either.

bitshifter:
Whoops, I did a `git commit --amend` but forgot to add first. Should be updated now.

Havvy:
> I have not updated the reference Attribute syntax grammar documentation as the new syntax is only used by #[repr(align)] and I thought it might be pretty confusing to include a grammar which is only used by one kind of attribute for now. &mdash @bitshifter; [rust-lang/rust #33626](https://github.com/rust-lang/rust/issues/33626#issuecomment-353904791)

Even if it's a change used by one attribute, it should still be in the attribute grammar. Do you want to add it? If not, I'll do so in a separate PR.

bitshifter:
Sorry, I didn't see your comment about the grammar. I can look at adding it.

bitshifter:
Updated the grammar, hopefully I got it right! It looks like my stabilization PR is almost in.

Havvy:
@bitshifter There's a merge conflict. 😢

bitshifter:
@Havvy the conflict was on a section on repr attributes that I removed as part of this review, had recently been updated by @yellow-apple. Thus I've removed the edit with the rest of the section, but perhaps @yellow-apple should check the `type-layout.md` section on `#[repr(C)] Enums` to make sure it contains the same information (it's worded differently but seems semantically similar to me).

Havvy:
💟 Thanks!

PR #183, opened 2017-12-29
Clarify implicit semicolon ambiguity resolution.

alercah:
See https://play.rust-lang.org/?gist=59090cd71dfbd84cbbdc611b94a0a9c2&version=stable for an example that shows that the rule explained in the current docs is incorrect. It implies that the parse ambiguity is resolved depending on the type of the control flow expression, but that's not true.


Havvy:
💟 Thanks!

I was going to get around to this, but forgot all about it. So 👍 for getting to it before me.

Havvy:
Rendered: https://havvy.net/rust-reference/statements.html

PR #186, opened 2017-12-30
Document variance

matthewjasper:
First of a few PRs to improve the documentation of lifetimes in the reference. Based on the documentation in the Nomicon.

cc #9 (rust-lang/rfcs#738)

### DIFF
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -1,4 +1,4 @@
-# Subtyping
+# Subtyping and Variance
 
 Subtyping is implicit and can occur at any stage in type checking or
 inference. Subtyping in Rust is very restricted and occurs only due to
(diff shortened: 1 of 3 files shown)


Havvy:
I don't know nearly enough about this to be reviewing it.

matthewjasper:
cc @Gankro @nikomatsakis @arielb1 ?

src/subtyping.md
```diff
@@ -15,5 +15,71 @@ fn bar<'a>() {
     let t: &'a str = s;
 }
 ```
-Since `'static` "lives longer" than `'a`, `&'static str` is a subtype of
-`&'a str`.
+
+Since `'static` outlives `'a`, `&'static str` is a subtype of `&'a str`.
+
+Subtyping also exists for [higher-ranked types]. Replacing a higher ranked
+lifetime with a concrete lifetime produces a subtype.
```
Gankro: Nit: I don't believe the documentation you linked specifies what a "higher ranked type" is. Not entirely sure what you mean here. Literally just function pointers?

src/subtyping.md
```diff
@@ -15,5 +15,71 @@ fn bar<'a>() {
...
+The subtype must be a valid type:
+
+```rust,compile_fail
+fn bar<'a>() {
+    let h: for<'b, 'c> fn(&'c &'b i32) -> &'b i32 = |x| &**x;
+    let j: fn(&'static &'a i32) -> &'a i32 = h;
+    // Error: in type `&'static &'a i32`, reference has a longer lifetime than
+    // the data it references
+}
+```
```
Gankro: Unclear to me why this merits discussion (is this meaningfully different from adding a note that "the expression must be parse"?)

src/subtyping.md
```diff
@@ -15,5 +15,71 @@ fn bar<'a>() {
...
+* `&'a T` is covariant in `'a` and `T`.
+* `&'a mut T` is covariant in `'a`, invariant in `T`.
+* `[T]` and `[T; n]` are covariant in `T`.
+* `*const T` is covariant in `T`.
+* `*mut T` is invariant in `T`.
+* `fn(T) -> U` is covariant in `U` and contravariant in `T`.
+* `std::cell::UnsafeCell<T>` is invariant in `T`.
+* `std::marker::PhantomData<T>` is covariant in `T`.
+* Trait objects are invariant in their type parameters and associated types and
+  covariant in their [lifetime bound].
```
Gankro: I think this would be better written as a table, but idk

src/subtyping.md
```diff
@@ -15,5 +15,71 @@ fn bar<'a>() {
...
+
+The variance of `struct`, `enum`, `union` and tuple types is decided by looking
+at the variance of the types of their fields. If a type parameter is used more
+than once then the more restrictive variance is used. For example the following
+struct is covariant in `'a` and `T` and invariant in `'b` and `U`.
```
Gankro: This is a bit unclear (and either very subtle or just actually wrong for co+contra). There's basically two interesting cases for multi-occurence:

* If everything agrees on the variance, then that variance is used
* If two different variances would apply, it's invariant

src/subtyping.md
matthewjasper: Function pointers and trait objects

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
...
+| `&'a T`                       | covariant         | covariant         |
+| `&'a mut T`                   | covariant         | invariant         |
+| `*const T`                    |                   | covariant         |
+| `*mut T`                      |                   | invariant         |
+| `[T]` and `[T; n]`            |                   | covariant         |
+| `fn() -> T`                   |                   | covariant         |
+| `fn(T) -> ()`                 |                   | contravariant     |
+| `std::cell::UnsafeCell<T>`    |                   | invariant         |
+| `std::marker::PhantomData<T>` |                   | covariant         |
+| `Trait<T> + 'a`               | covariant         | invariant         |
```
Gankro: Thing I don't know but someone should check: is `&'a mut Trait<T> + 'a` actually invariant in `'a`? (this is implied my the entries for `&mut T` and `Trait<T> + 'a`, but I would almost expect the compiler to magic this away.

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
     let t: &'a str = s;
 }
 ```
-Since `'static` "lives longer" than `'a`, `&'static str` is a subtype of
-`&'a str`.
+
+Since `'static` outlives `'a`, `&'static str` is a subtype of `&'a str`.
```
Gankro: If it's not declared above, could we clarify that `'a` here is a standin for "any generic lifetime in scope" (or maybe just "some lifetime"?)

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
...
+
+Subtyping also exists for [higher-ranked] function pointer and trait object
+types. Replacing a higher ranked lifetime with a concrete lifetime produces a
```
Gankro: I would rewrite this to "function pointers and trait objects" (current way reads awkward)

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
...
+subtype.
```
Gankro: This is an interesting phrasing, I'm not sure if it makes sense to most.

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
...
+    // Explicitly f: &(for<'a> fn(&'a i32) -> &'a i32).
+    let f: &(fn(&i32) -> &i32) = &((|x| x) as fn(&i32) -> &i32);
+    let g: &(fn(&'static i32) -> &'static i32) = f;
+}
+
+fn fun_trait() {
+    // Explicitly f: &(for<'a> Fn(&'a i32) -> &'a i32).
+    let f: &(Fn(&i32) -> &i32) = &(|x| x);
+    let g: &(Fn(&'static i32) -> &'static i32) = f;
+}
```
Gankro: Unless I'm mistaken, doctest implicitly wraps the example in `main`, so you can get rid of these distracting functions.

Also consider renaming f and g to "base" and "subtype" or something more self-descriptive.

src/subtyping.md
Gankro: Is the notion of a concrete lifetime really well-defined and/or clear here?

src/subtyping.md
```diff
@@ -15,5 +15,70 @@ fn bar<'a>() {
...
+and `U`.
+
+```rust
+use std::cell::UnsafeCell;
+struct Variance<'a, 'b, T, U: 'a> {
+    x: &'a U,
+    y: *const T,
+    z: UnsafeCell<&'b f64>,
+    w: *mut U,
+}
```
Gankro: This example would benefit from inline comments or outline discussion (I originally missed that U was used twice).

src/subtyping.md
matthewjasper: It looks like it is invariant, but there is a coercion from `&mut (Trait + 'a)` to `&mut Trait + 'b` where `'a: 'b`.  The same as `&mut (Trait + Send)` and `&mut Trait`.
```rust
trait A {}
// This compiles (also with `*mut`, or `Box<RefCell>`)
fn coercion<'a, 'b>(x: &'b mut (A + 'static)) -> &'b mut (A + 'a) {
    x
}
// This does not
fn not_subtype<'a, 'b, 'c>(x: &'c &'b mut (A + 'static)) -> &'c &'b mut (A + 'a) {
    x
}
```

Havvy:
@Gankro Wanna give this a re-review? (You're allowed to say no)

src/subtyping.md
```diff
@@ -15,5 +15,75 @@ fn bar<'a>() {
     let t: &'a str = s;
 }
 ```
-Since `'static` "lives longer" than `'a`, `&'static str` is a subtype of
-`&'a str`.
+
+Since `'static` outlives the lifetime parameter `'a`, `&'static str` is a
+subtype of `&'a str`.
+
+[Higher-ranked]&#32;[function pointers] and [trait objects] have another
```
Gankro: why is this escaped space necessary? otherwise markdown parses it as a reference?

Havvy: Markdown would read it as a link with text of "[Higher ranked]" and a link reference to "function pointers" instead of as two separate links.

Havvy:
Rebase this on master and leave a note saying you've done so and I'll get it merged.

matthewjasper:
rebased, but mdbook isn't building on travis.

PR #187, opened 2017-12-30
Lifetime elision

matthewjasper:
Second PR to add documentation for lifetimes (next will probably moving the relevant parts of the NLL RFC text into the reference when it gets stabilised).

This is mostly existing documentation from the nomicon/reference. Significant changes from that documentation:

* Don't talk about possible future changes (elision in `impl` headers)
* Be more concise in the description of function lifetime elision
* Add an example of function pointer and function trait elision
* Note that `'static` is the preferred bound in trait objects with multiple bounds:
```rust
trait A<'x, 'y>: 'x + 'y {}
impl A<'a, 'static> {} // A<'a, 'static> + 'static
```
* Don't say anything about the case in rust-lang/rust#47078
* `const RESOLVED_STATIC: &Fn(&Foo, &Bar) -> &Baz` is invalid, not inferred with `'static` lifetimes. (This matches: http://rust-lang.github.io/rfcs/1623-static.html, so I'm guessing that it's just a docs issue)

cc #9 (rust-lang/rfcs#141)

### DIFF
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -59,50 +59,6 @@ unsafe fn bump_levels_unsafe2() -> u32 {
 Mutable statics have the same restrictions as normal statics, except that the
 type does not have to implement the `Sync` trait.
 
-## `'static` lifetime elision
-
-Both constant and static declarations of reference types have *implicit*
-`'static` lifetimes unless an explicit lifetime is specified. As such, the
-constant declarations involving `'static` above may be written without the
-lifetimes. Returning to our previous example:
-
-```rust
-const BIT1: u32 = 1 << 0;
-const BIT2: u32 = 1 << 1;
-
-const BITS: [u32; 2] = [BIT1, BIT2];
-const STRING: &str = "bitstring";
-
-struct BitsNStrings<'a> {
-    mybits: [u32; 2],
-    mystring: &'a str,
-}
-
-const BITS_N_STRINGS: BitsNStrings = BitsNStrings {
-    mybits: BITS,
-    mystring: STRING,
-};
-```
-
-Note that if the `static` or `const` items include function or closure
-references, which themselves include references, the compiler will first try
-the standard elision rules ([see discussion in the nomicon][elision-nomicon]).
-If it is unable to resolve the lifetimes by its usual rules, it will default to
-using the `'static` lifetime. By way of example:
-
-```rust,ignore
-// Resolved as `fn<'a>(&'a str) -> &'a str`.
-const RESOLVED_SINGLE: fn(&str) -> &str = ..
-
-// Resolved as `Fn<'a, 'b, 'c>(&'a Foo, &'b Bar, &'c Baz) -> usize`.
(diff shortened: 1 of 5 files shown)


src/items/static-items.md
```diff
@@ -59,50 +59,6 @@ unsafe fn bump_levels_unsafe2() -> u32 {
 Mutable statics have the same restrictions as normal statics, except that the
 type does not have to implement the `Sync` trait.
 
-## `'static` lifetime elision
```
Havvy: Not sure how, but it'd be nice for this page to say it's possible and link to the rules for it.

matthewjasper: There is at least a mention of this on the constant items page with a link.

src/lifetime-elision.md
```diff
@@ -0,0 +1,169 @@
...
+type FunTrait = for<'a> Fn(&'a str) -> &'a str;         // expanded
+```
+
+## Defaults trait object lifetimes
+
+The assumed lifetime of references held by a [trait object] is called its
+_default object lifetime bound_. These were defined in [RFC 599] and amended in
+[RFC 1156].
+
+If the trait object is used as for a type argument of a generic type then the
```
aturon: "used as for a type"

src/lifetime-elision.md
```diff
@@ -0,0 +1,169 @@
...
+containing type is first used to try to infer a bound.
+* If there is a unique bound from the containing type then that is the default
+* If there is more than one bound from the containing type then an explicit
+  bound must be specified
+
+Then the bounds on the trait are used:
```
aturon: The text doesn't make clear the relationship between these bullets and the previous ones.

matthewjasper:
changes made

src/lifetime-elision.md
```diff
@@ -0,0 +1,171 @@
...
+
+Rust has rules that allow lifetimes to be elided in various places where the
+compiler can infer a sensible default choice.
+
+## Lifetime elision in functions
+
+In order to make common patterns more ergonomic, Rust allows lifetimes to be
+*elided* in [function item], [function pointer] and [closure trait] signatures.
+The following rules are used to infer lifetime parameters for elided lifetimes.
+It is an error to elide lifetime parameters that cannot be infered.
```
alercah: inferred

src/lifetime-elision.md
```diff
@@ -0,0 +1,171 @@
...
+fn new<'a>(buf: &'a mut [u8]) -> BufWriter<'a>;         // expanded
+
+type FunPtr = fn(&str) -> &str;                         // elided
+type FunPtr = for<'a> fn(&'a str) -> &'a str;           // expanded
+
+type FunTrait = Fn(&str) -> &str;                       // elided
+type FunTrait = for<'a> Fn(&'a str) -> &'a str;         // expanded
+```
+
+## Defaults trait object lifetimes
```
alercah: Default

matthewjasper:
Thanks, fixed now.

Havvy:
cc @alercah

alercah:
I don't see anything that looks off to me, but I also don't have the time for a detailed read.

PR #195, opened 2018-01-10
Add contributing document

Havvy:


CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+[Document all features][9] or a [missing feature] tag, and write about it.
+
+While writing, you may find it handy to have a [playpen] open to test out
+
+Feel free to take information from the standard library and Rustonomicon as
+appropriate.
+
+Note that we don't write documentation for purely library features such as
+threads and IO and we don't write about Rust in the future. Documentation is
+written as if the current stable release of Rust is the last release. If you
```
alercah: Possibly add a note that this isn't entirely true when parts of a feature are stabilized? I'm thinking in particular about `CoerceUnsized`, which isn't stabilized, but where stable code can take advantage of the instances already in `std`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+Likewise, the reference has a glossary. It doesn't need to explain every facet
+of every feature nor contain every definition, but it does need to be expanded
+upon. Ideally entries in the glossary link to the associated documentation.
+
+## Adding Documentation
+
+There are a lot of features that are not documented at all or are documented
+poorly. This is the hardest, but definitely most valuable. Pick something from
+[Document all features][9] or a [missing feature] tag, and write about it.
```
alercah: nit: Capitalization seems off.

CONTRIBUTING.md
Havvy: Even in situations like this, I'd rather not mention the not yet stabilized parts, unless it really does help the flow. Your PR does technically break this, and I've been trying to find the right words to say about it. I think it's fine for now, as it is strictly better than what was there before, and at the end of the day, these are all guidelines, not strict policies.

CONTRIBUTING.md
Havvy: Went with adding quotes around it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+of every feature nor contain every definition, but it does need to be expanded
+upon. Ideally entries in the glossary link to the associated documentation.
+
+## Adding Documentation
+
+There are a lot of features that are not documented at all or are documented
+poorly. This is the hardest, but definitely most valuable. Pick something from
+["Document all features"][9] or a [missing feature] tag, and write about it.
+
+While writing, you may find it handy to have a [playpen] open to test out
```
QuietMisdreavus: Feels like this paragraph got started but not finished. `>_>`

Havvy:
💟

PR #202, opened 2018-01-16
Interent -> Inherent

Havvy:


marioidival:
💟

QuietMisdreavus:
Given that travis is having issues with backlogged builds, i'm going ahead and merging this anyway since it didn't touch any code blocks. Links will be another factor, but IIRC those won't be checked until the submodule back in rust-lang/rust gets updated.

PR #209, opened 2018-01-21
Start documenting rustc

mrhota:
As far as I know, we don't have any reference documentation on the actual tools in the rust tool suite.

### DIFF
--- a/src/tools/rustc.md
+++ b/src/tools/rustc.md
@@ -40,15 +40,10 @@ Providing the `-g` option is equivalent to `-C debuginfo=2`. If both `-g` and
 ### Optimization
 
 To produce optimized output, use the `-C opt-level=val` option, where `val`
-may be one of `0`, `1`, `2`, `3`. The nightly compiler will also accept `s`,
-or `z`. The default is `0`.
-  - `0-3` direct `rustc` to optimize for execution speed with `0` meaning no
-    optimizations, and `3` meaning aggressive optimization.
-  - `s` and `z` direct `rustc` to optimize for output size, with `s` meaning
-    typical size optimizations and `z` meaning aggressive size optimizations.
+may be one of `0`, `1`, `2`, `3`. The default is `0`.
+
+The numbers indicate increasing levels of optimization for execution speed,
+with `0` meaning no optimizations, and `3` meaning aggressive optimization.
 
 Providing the option `-O` is equivalent to `-C opt-level=2`. If both `-O` and
 `-C opt-level` are provided, the compiler will complain.
-
-If `s` or `z` are provided on a non-nightly compiler, the compiler will
-complain.
--- a/src/tools/rustc.md
+++ b/src/tools/rustc.md
@@ -1,0 +1,54 @@
+# `rustc`
+
+The Rust compiler has many options and can accept a wide variety of arguments,
+and its behavior can vary depending on the values of several environment
+variables.
+
+We document the compiler's command-line options, arguments, and operative
+environment variables here.
+
+Some discussions of environment variables exists in the [Linkage](linkage.html)
+chapter and the [Operator expressions](expressions/operator-expr.html#overflow)
+chapter.
+
+## Lint options
+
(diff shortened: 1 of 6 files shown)


src/undocumented.md
```diff
@@ -14,20 +14,14 @@ to shrink!
   specified.
 - [Flexible target specification] - Some---but not all---flags are documented
   in [Conditional compilation]
-- [Require parentheses for chained comparisons]
```
matthewjasper: Is this supposed to be part of a separate PR?

src/tools/rustdoc.md
```diff
@@ -0,0 +1,8 @@
+# `rustdoc`
+
+`rustdoc` has many options and can accept a wide variety of arguments,
+and its behavior can vary depending on the values of several environment
+variables.
+
+We document its command-line options, arguments, and operative environment
```
matthewjasper: You can just link to `../rustdoc/index.html` for documenting rustdoc (and `../cargo/index.html` for cargo)

src/undocumented.md
mrhota: no: that feature is already documented [here](https://doc.rust-lang.org/reference/expressions/operator-expr.html#comparison-operators), so I'm just cleaning this list up.

src/tools/rustdoc.md
mrhota: oh dang it, I didn't even know about those. I'll link to the references.

src/tools/rustc.md
```diff
@@ -0,0 +1,54 @@
...
+or `z`. The default is `0`.
+  - `0-3` direct `rustc` to optimize for execution speed with `0` meaning no
+    optimizations, and `3` meaning aggressive optimization.
+  - `s` and `z` direct `rustc` to optimize for output size, with `s` meaning
+    typical size optimizations and `z` meaning aggressive size optimizations.
+
+Providing the option `-O` is equivalent to `-C opt-level=2`. If both `-O` and
+`-C opt-level` are provided, the compiler will complain.
+
+If `s` or `z` are provided on a non-nightly compiler, the compiler will
```
Havvy: We don't document nightly-only features in the Reference, to avoid having too large of a moving target to follow. Instead, said nightly features should be documented in the Unsafe Book inside the rust-lang/rust repo.

mrhota: I know, but the problem is `rustup run stable rustc -C help` says you can use `s` and `z` opt-levels. I thought it was worth mentioning you can't use those on non-nightlies.

I'm not married to the wording, obviously 😃

mrhota: I created rust-lang/rust#47651 for making non-nightlies stop reporting `s` and `z` options.

Havvy:
This is an extension to the scope of the Reference, and as such, needs to be discussed during a documentation meeting. As such, I've added it to the agenda for the next one on Tuesday, January 23rd. Anybody interested is free to attend (like all documentation meetings). They're in #rust-docs on irc.mozilla.org

steveklabnik:
Rustdoc has an entire book in `src/doc`, the compiler does not yet.

steveklabnik:
So! @mrhota i want to start off by saying that doing this is awesome, however, i don't think it's appropriate for the reference. Right now, we have two out of the three, but elsewhere:

* https://doc.rust-lang.org/cargo / https://github.com/rust-lang/cargo/blob/master/src/doc (included in the rust repo via submodule)
* https://doc.rust-lang.org/rustdoc/ / https://github.com/rust-lang/rust/tree/master/src/doc/rustdoc

As for `rustc`, there's *also* an effort to make a book that's not about using, but contributing to, the compiler. I've opened a comment there to ask about this: https://internals.rust-lang.org/t/so-you-want-to-hack-on-the-rust-compiler-a-plan-for-a-book/6497/37

So, let's:

1. see what they say on internals
2. create a new book in the rust repo for rustc usage
3. improve those existing books instead of doing it here

Thank you! Let me know if you want to help with 2, we can start that in a PR before the full discussion gets going.

PR #210, opened 2018-01-22
Reorganize appendices slightly

mrhota:
i recommend building locally to see how it looks. I think it's nicer to have appendices under one chapter. What do y'all think?


Havvy:
If a file doesn't exist for the link in SUMMARY.md, mdbook tries to create a file by that name. It creates the filename presumably by taking the project path and appending the filepath we provide. It is empty for "Appendices". This means it's trying to create a file with the same name as the directory, and that fails.

[Rendered](https://havvy.net/rust-reference/introduction.html) (for now)

Havvy:
I'm not actually sure why it's working locally. Rust currently uses v.0.0.28, but Travis is using the latest master, I think.

mrhota:
This isn't important, and it appears that the builder croaks on it. Unless you want this, @Havvy, I'm cool with just closing it

Havvy:
I'm neutral on it, since it breaks links.

mrhota:
eh, maybe later. closing for now!

PR #215, opened 2018-01-24
Document inclusive ranges.

alercah:
Provides the reference documentation for rust-lang/rust#28237.

### DIFF
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -1,27 +1,27 @@
 # `match` expressions
 
-> **<sup>Syntax</sup>**  
-> _MatchExpression_ :  
-> &nbsp;&nbsp; `match` [_Expression_]<sub>_except struct expression_</sub> `{`  
-> &nbsp;&nbsp; &nbsp;&nbsp; [_InnerAttribute_]<sup>\*</sup>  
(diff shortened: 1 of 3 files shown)


src/expressions/match-expr.md
```diff
@@ -95,18 +95,23 @@ match x {
...
-    2 ... 9 => "a few",
+    2 ..= 9 => "a few",
     _      => "lots"
 };
+
+assert_eq!(message, "a few");
 ```
 
+Other forms of [range] \(`..` for an exclusive range, or any range with one or
+both endpoints left unspecified) are not currently supported in matches.
```
matthewjasper: I don't think that 'currently' is needed, the reference only talks about Rust at this point in time.

alercah: Done. I also remarked that `...` is still accepted for backwards compatibility.

alercah:
According to rust-lang/rust#28237, all that's left is docs, so hopefully once all my doc PRs are good it will be ready to go.

matthewjasper:
`..=` is stable now, can you fix the merge conflict so that this can be merged?

alercah:
Done.

steveklabnik:
This just landed as stable in nightly. Let's restart travis!

steveklabnik:
🎊

PR #232, opened 2018-02-02
Remove chained comparisons from the undocumented page.

ekse:
Small edit, this feature was documented #41, I forgot to remove it from undocumented.md.


Havvy:
Doesn't touch code blocks, so merging ahead of TravisCI.

💟 Thanks @ekse for this and the prior work.

PR #234, opened 2018-02-07
Clarify which types range patterns work on

equal-l2:
According to the Rust book, the word "scalar types" means ["integers, floating-point numbers, booleans, and characters."](https://doc.rust-lang.org/book/second-edition/ch03-02-data-types.html#scalar-types)

However, range patterns don't work on booleans.
(No one would actually use range patterns on booleans, of course.)


Havvy:
💟 Thanks!

PR #160, opened 2017-11-30
Struct grammar

brauliobz:
Part of #84

PR #237, opened 2018-02-07
Struct grammar

Havvy:
This is #160 except I removed the `?`s from visibility as they are extraneous.

### DIFF
--- a/src/items/structs.md
+++ b/src/items/structs.md
@@ -1,5 +1,41 @@
 # Structs
 
+> **<sup>Syntax</sup>**  
+> _Struct_ :  
+> &nbsp;&nbsp; &nbsp;&nbsp; _StructStruct_  
+> &nbsp;&nbsp; | _TupleStruct_  
+>  
+> _StructStruct_ :  
+> &nbsp;&nbsp; `struct`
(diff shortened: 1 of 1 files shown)


src/items/structs.md
```diff
@@ -1,5 +1,41 @@
...
+> _Struct_ :  
+> &nbsp;&nbsp; &nbsp;&nbsp; _StructStruct_  
+> &nbsp;&nbsp; | _TupleStruct_  
+>  
+> _StructStruct_ :  
+> &nbsp;&nbsp; `struct`
+>   [IDENTIFIER]&nbsp;
+>   [_Generics_]<sup>?</sup>
+>   [_WhereClause_]<sup>?</sup>
+>   ( `{` _StructFields_<sup>?</sup> `}` | `;` )
```
matthewjasper: Would it help to have this match the text and split this into _StructStruct_ and _UnitStruct_?

Havvy: I think I considered that when I first reviewed @brauliobz 's grammar and found a reason against it. Not sure anymore though. I'd rather not mess with the PR anymore than needed right now.

matthewjasper:
Thanks!

PR #214, opened 2018-01-23
Move associated items to their own page.

Havvy:
Because they are a cutting concern between impls and traits, they don't really fit in either.

If anybody has concrete suggestions for how to make this better, I'm all ears. I'm not a fan of a lot of my writing in this PR. Especially the definitions.

PR #242, opened 2018-02-12
Hide more lines in code example

matthewjasper:
Hides lines to avoid unmatched close brace and extra blank lines.

Follow up to #214


Havvy:
😳 Thanks!

PR #247, opened 2018-02-19
Document RFC 1940 (must_use on fns)

Havvy:
* [Rendered](https://havvy.net/rust-reference/attributes.html)
* [Tracking Issue](https://github.com/rust-lang/rust/issues/43302)
* [RFC](http://rust-lang.github.io/rfcs/1940-must-use-functions.html)
* [Unstable Book Docs](https://doc.rust-lang.org/nightly/unstable-book/language-features/fn-must-use.html)

This is the reference side of documenting this attribute &mdash; both the updates for the RFC and the original functionality. I do not think that this PR is sufficient documentation for stabilizing the feature, as I think the API guidelines should also be updated.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -400,7 +400,9 @@ fn main() {
 
 When used on a function, if the [expression] of an
 [expression statement] is a [call expression] to that function, then the
-`unused_must_use` lint is violated.
+`unused_must_use` lint is violated. The exceptions to this is if the return type
+of the function is `()`, `!`, or a [zero-variant enum], in which case the
+attribute does nothing.
 
 ```rust
 #[must_use]
@@ -514,3 +516,4 @@ You can implement `derive` for your own type through [procedural macros].
 [`Drop`]: special-types-and-traits.html#drop
 [let statement]: statements.html#let-statements
 [unstable book plugin]: ../unstable-book/language-features/plugin.html#lint-plugins
+[zero-variant enum]: enumerations.html#zero-variant-enums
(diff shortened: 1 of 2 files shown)


src/attributes.md
```diff
@@ -376,6 +373,84 @@ pub mod m3 {
 }
 ```
 
+#### Must Use Attribute
```
alercah: I'd change this to "must_use Attribute"

src/attributes.md
```diff
@@ -376,6 +373,84 @@ pub mod m3 {
...
+```rust
+#[must_use]
+struct MustUse {
+  // some fields
+}
+
+# impl MustUse {
+#   fn new() -> MustUse { MustUse {} }
+# }
+
```
alercah: Add a # to the start of one of these blank lines to avoid two in the rendered text.

Havvy:
Updated for behaviour with trait objects.

Travis failure is in building mdbook?

steveklabnik:
> Travis failure is in building mdbook?

That's... odd

zackmdavis:
[Stabilization PR has landed](https://github.com/rust-lang/rust/pull/48925).

Havvy:
I added a commit that just added a space and then had Travis test it. It passed. I then deleted that commit and it went back to the previous status of not passing because of whatever mysterious error we had last time. So if anybody wants to merge this, it's ready for merging.

zackmdavis:
Travis will probably rebuild if you rebase on master?

Havvy:
You are correct, and it now shows as passing.

PR #249, opened 2018-02-20
Improve documentation on closure types.

### LINKED ISSUE #219: Document more precisely the rules for selecting closure traits.
Currently, the reference is vague about when a closure is `Fn` or `FnMut` (all closures are `FnOnce`). Currently, [the closures page](https://doc.rust-lang.org/reference/expressions/closure-expr.html) merely says that the compiler will work it out, but we need to be more specific about precisely what the semantics are.

It may be worth rewriting this page in general to be more authoritative as well.

### LINKED ISSUE #229: Document RFC 2132 (copy-closures)
[RFC link](https://github.com/rust-lang/rfcs/tree/master/text/2132-copy-closures.md)

The RFC is nearing stabilization and now seems like a good time to document it. I think this is a good opportunity to explain the desugaring of closures and the like, to also fix #219.

alercah:
This PR more fully documents closure types, including the mechanics of captures and the traits implemented by them. I intend to work on closure expressions as well, either in this PR or a follow-up, but felt this was a good place to put them.

I didn't fully explain capturing here, as I feel like that's better for closure expressions. The bit about no field captures might make more sense there as well, but I think the mechanics of how captures are typed and the implications on the traits belong here.

This PR is written for a post-[RFC 2132](https://github.com/rust-lang/rfcs/blob/master/text/2132-copy-closures.md) world, where closures have `Clone` and `Copy`, and intended as part of the stabilization docs for that RFC. The tracking issue is rust-lang/rust#44490. That bit could easily be split out though.

Fixes #219, fixes #229.

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -365,30 +365,99 @@ x = bo(5,7);
 ## Closure types
 
 A [closure expression] produces a closure value with a unique, anonymous type
-that cannot be written out.
+that cannot be written out. A closure type is approximately equivalent to a
+struct which contains the captured variables. For instance, the following
+closure:
 
-Depending on the requirements of the closure, its type implements one or
-more of the closure traits:
+```rust
+fn f<F : FnOnce() -> String> (g: F) {
+    println!("{}", g());
+}
 
-* `FnOnce`
-  : The closure can be called once. A closure called as `FnOnce` can move out
-    of its captured values.
+let mut s = String::from("foo");
+let t = String::from("bar");
 
-* `FnMut`
-  : The closure can be called multiple times as mutable. A closure called as
-    `FnMut` can mutate values from its environment. `FnMut` inherits from
-    `FnOnce` (i.e. anything implementing `FnMut` also implements `FnOnce`).
+f(|| {
+    s += t;
+    s
+});
+// Prints "foobar".
+```
 
-* `Fn` : The closure can be called multiple times through a shared reference. A
-    closure called as `Fn` can neither move out from nor mutate captured
-    variables, but read-only access to such values is allowed. Using `move` to
-    capture variables by value is allowed so long as they aren't mutated or
-    moved in the body of the closure. `Fn` inherits from `FnMut`, which itself
-    inherits from `FnOnce`.
+generates a closure type roughly like the following:
 
-Closures that don't use anything from their environment, called *non-capturing
-closures*, can be coerced to function pointers (`fn`) with the matching
-signature. To adopt the example from the section above:
+```rust
+struct Closure<'a> {
+    s : String
+    t : &'a String
+}
+
+impl<'a> FnOnce() -> String for Closure<'a> {
+    fn call_once(self) -> String {
+        self.s += self.t;
+        self.s
+    }
+}
+```
+
+so that the call to `f` works as if it were:
+
+```rust,ignore
+f(Closure{s: s, t: &t});
+```
+
+The compiler prefers to capture a closed-over variable by immutable borrow,
+followed by mutable borrow and finally by move (or copy, for [`Copy`] types). It
+will pick the first choice of these that allows the closure to compile. If the
+`move` keyword is used, then all captures are by move or copy, regardless of
+whether a borrow would work. The `move` keyword is usually used to allow the
+closure to outlive the captured values, such as if the closure is being returned
+or used to spawn a new thread.
+
+Structs and tuples are always captured entirely, not by individual fields. It
+may be necessary to borrow into a local variable in order to capture a single
+field:
+
+```rust
+struct SetVec {
+    set: HashSet<u32>,
+    vec: Vec<u32>
+}
+
+impl Pair {
+    fn populate(&mut self) {
+        let vec = &mut self.vec;
+        self.set.iter().for_each(|&n| {
+            vec.push(n);
+        })
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/types.md
```diff
@@ -365,30 +365,101 @@ x = bo(5,7);
...
-    inherits from `FnOnce`.
+```rust,ignore
+f(Closure{s: s, t: &t});
+```
 
-Closures that don't use anything from their environment, called *non-capturing
-closures*, can be coerced to function pointers (`fn`) with the matching
-signature. To adopt the example from the section above:
+The compiler prefers to capture a closed-over variable by immutable borrow,
+followed by mutable borrow and finally by move (or copy, for [`Copy`] types). It
```
Havvy: Need a comma before `and`.

Havvy: Might read better if the list was four long and `by copy` was in there before `by move`.

Havvy: Need a comma before the `and`.

src/types.md
```diff
@@ -365,30 +365,101 @@ x = bo(5,7);
...
+will pick the first choice of these that allows the closure to compile. If the
+`move` keyword is used, then all captures are by move or copy, regardless of
+whether a borrow would work. The `move` keyword is usually used to allow the
+closure to outlive the captured values, such as if the closure is being returned
+or used to spawn a new thread.
+
+Structs and tuples are always captured entirely, not by individual fields. It
```
Havvy: "Composite types, such as structs or tuples, are"? Or are enums somehow different?

src/types.md
```diff
@@ -365,30 +365,101 @@ x = bo(5,7);
...
+        self.set.iter().for_each(|&n| {
+            vec.push(n);
+        })
+    }
+}
+```
+
+If, instead, the closure were to use `self.vec` directly, then it would attempt
+to capture `self` by mutable reference. But since `self.set` is already
+borrowed to iterate over, the closure would not compile.
```
Havvy: the code would not compile

src/types.md
```diff
@@ -365,30 +365,101 @@ x = bo(5,7);
...
+by consuming ownership of the closure. Additionally, some closures implement
+more specific call traits:
+
+* A closure which does not move out of any captured variables implements
+  `[FnMut]`, indicating that it can be called by mutable reference.
+
+* A closure which does not mutate or move out of any captured variables
+  implements `[Fn]`, indicating that it can be called by shared reference.
+
+> Note that `move` closures may still implement `[Fn]` or `[FnMut]`, even
```
Havvy: Note needs to have a colon after it to remain consistent with the rest of the notes in the reference.

src/types.md
```diff
@@ -365,30 +365,101 @@ x = bo(5,7);
...
+> though they capture variables by move. This is because the traits
+> implemented by a closure type are determined by what the closure does with
+> captured values, not how it captures them.
+
+In addition to the call traits, *non-capturing closures*---those that don't
+capture anything from their environment---can be coerced to function pointers
```
Havvy: Define the term in its own sentence, and then describe how it's supposed to be used.

`*Non-capturing closures* are closures that don't capture any variables from their environment. They can be coereced...`

src/types.md
```diff
@@ -400,6 +471,33 @@ let bo: Binop = add;
 x = bo(5,7);
 ```
 
+### Other traits
+
+Closure types implement the following traits, if allowed to do so by the
```
Havvy: The comma is invalid.

Havvy: The actual values captured don't matter, but the types of the variables do. I'd personally write it "Closures implement the following traits conditionally on the types of the captured variables."

src/types.md
```diff
@@ -400,6 +471,33 @@ let bo: Binop = add;
...
+captured values:
+
+* `[Sized]`
```
Havvy: It feels weird that we're talking about conditionally implemented traits and then the first one on the list is always implemented.

src/types.md
```diff
@@ -400,6 +471,33 @@ let bo: Binop = add;
...
+* `[Send]`
+* `[Sync]`
+* `[Clone]`
```
Havvy: Given how Clone and Copy are lang items and Send and Sync aren't (I think? Maybe one of them is?), I'd rather see Clone and Copy listed first.

src/types.md
```diff
@@ -400,6 +471,33 @@ let bo: Binop = add;
...
+* `[Copy]`
+
+`[Sized]` is always implemented (local variables are all sized, so all captured
```
Havvy: More of a meta-point, but I really don't like having parentheticals like this. They should generally just be another sentence.

src/types.md
alercah: `Sync` is, `Send` isn't. I think it would look better reordered though, so I will do that.

src/types.md
alercah: That's a fair point. But it also doesn't necessarily depend on the type of the variable captured, but the type of the capturing variable in the closure. For instance, capturing a non-`Copy` type by reference is fine for `Copy`, since the reference is itself `Copy`.

alercah: Found a way to make this work, please take a look once I push.

src/types.md
alercah: Fixed it.

src/types.md
alercah: Removed that sentence anyway.

PR #251, opened 2018-02-21
fixes #250

### LINKED ISSUE #250: [Compiler plugins][unstable book plugin] is not an url
here: https://doc.rust-lang.org/reference/attributes.html#lint-check-attributes

> The lint checks supported by the compiler can be found via rustc -W help, along with their default settings. [Compiler plugins][unstable book plugin] can provide additional lint checks.

seems like `[Compiler plugins][unstable book plugin]` should be [an url](https://github.com/rust-lang-nursery/reference/commit/337c28ba306359fc8a209c1fe6208b4e22e31876#diff-feef7790e64d1a3e5c829b86054e3919L500) or [two](https://github.com/rust-lang-nursery/reference/commit/337c28ba306359fc8a209c1fe6208b4e22e31876#diff-b96bbd6121a0758948a83bf736162e69R14)...

xftroxgpx:
`[Compiler plugins][unstable book plugin]` is now a link again. See #250

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -430,3 +430,4 @@ You can implement `derive` for your own type through [procedural macros].
 [Doc comments]: comments.html#doc-comments
 [The Rustdoc Book]: ../rustdoc/the-doc-attribute.html
 [procedural macros]: procedural-macros.html
+[unstable book plugin]: ../unstable-book/language-features/plugin.html#lint-plugins


xftroxgpx:
fix from https://github.com/rust-lang-nursery/reference/commit/337c28ba306359fc8a209c1fe6208b4e22e31876#diff-feef7790e64d1a3e5c829b86054e3919L500

Havvy:
I wonder if we should just remove the sentence that contained the link instead (as unstable features aren't in scope), but having the link work while it's there is good for now.

PR #253, opened 2018-02-25
Split identifiers from keywords or identifiers

matthewjasper:
Separate identifiers, and keywords including identifiers in the grammar.

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -1,19 +1,25 @@
 # Identifiers
 
 > **<sup>Lexer:<sup>**  
-> IDENTIFIER :  
+> IDENTIFIER_OR_KEYWORD :  
 > &nbsp;&nbsp; &nbsp;&nbsp; [`a`-`z` `A`-`Z`]&nbsp;[`a`-`z` `A`-`Z` `0`-`9` `_`]<sup>\*</sup>  
 > &nbsp;&nbsp; | `_` [`a`-`z` `A`-`Z` `0`-`9` `_`]<sup>+</sup>  
+>  
+> IDENTIFIER :  
+> IDENTIFIER_OR_KEYWORD <sub>*Except a [strict] or [reserved] keyword*</sub>
 
 An identifier is any nonempty ASCII string of the following form:
 
 Either
 
-   * The first character is a letter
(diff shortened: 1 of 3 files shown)


src/identifiers.md
```diff
@@ -1,19 +1,25 @@
...
+* The first character is a letter
+* The remaining characters are alphanumeric or `_`
 
 Or
 
-   * The first character is `_`
-   * The identifier is more than one character, `_` alone is not an identifier
-   * The remaining characters are alphanumeric or `_`
+* The first character is `_`
+* The identifier is more than one character, `_` alone is not an identifier
```
Havvy: I know this was here previously, but can you change the comma to a period and otherwise add periods to these sentences?

src/tokens.md
```diff
@@ -486,10 +486,15 @@ The two values of the boolean type are written `true` and `false`.
...
+> LIFETIME_TOKEN
+> &nbsp;&nbsp; &nbsp;&nbsp; `'` [IDENTIFIER_OR_KEYWORD][identifier]  
+> &nbsp;&nbsp; | `'_`  
+>  
 > LIFETIME_OR_LABEL:  
 > &nbsp;&nbsp; &nbsp;&nbsp; `'` [IDENTIFIER][identifier]
 
-Lifetime parameters and [loop labels] both use this syntax.
+Lifetime parameters and [loop labels] use LIFETIME_OR_LABEL tokens. Slightly
+more tokens are accepted by the lexer, and can be used in macros, for example.
```
Havvy: I think it would read slightly better if the "for example" was after the "and". I also think the "and" would read better as a "which". It'd be nice to state or list exactly which other tokens are accepted by the lexer and where else other than macros they can be used.

src/tokens.md
```diff
@@ -486,10 +486,15 @@ The two values of the boolean type are written `true` and `false`.
 ## Lifetimes and loop labels
 
 > **<sup>Lexer</sup>**  
+> LIFETIME_TOKEN
```
Havvy: Where does this new token get used?

Havvy: Ah, in macros, as described in the prose.

Havvy:
💟 Thanks!

PR #255, opened 2018-02-26
Trait object syntax + trailing plus

Havvy:


src/types.md
```diff
@@ -506,10 +510,10 @@ Trait objects implement the base trait, its auto traits, and any super traits
...
 
 Trait objects are written as the path to the base trait followed by the list
-of auto traits followed optionally by a lifetime bound all separated by `+`. For
-example, given a trait `Trait`, the following are all trait objects: `Trait`,
-`Trait + Send`, `Trait + Send + Sync`, `Trait + 'static`,
-`Trait + Send + 'static`.
+of auto traits followed optionally by a lifetime bound all separated by and
+optionally terminated by  `+`. For example, given a trait `Trait`, the following
+are all trait objects: `Trait`, `Trait + Send`, `Trait + Send + Sync`,
+`Trait + 'static`, `Trait + Send + 'static`, `Trait +`.
```
petrochenkov: `'static + Send + Trait` also works.
In general, trait objects has exactly the same syntax as bounds after `T` in `fn f<T: BOUNDS>() { ... }`, it would be great to unify them in the reference as well.
Trait objects just have some additional semantic restrictions (n(lifetimes) <=1, n(non-auto-traits) <=1).

Havvy: `'static + Send + Trait` explicitly doesn't work giving me `error[E0225]: only auto traits can be used as additional traits in a trait object`. `'static + Trait` does work.

Generic bounds don't seem to be in the reference at all right now. 🙁

matthewjasper: I'm working on it. :shades:

matthewjasper: I'm working on adding some documentation for them soon.

Havvy:
Added lifetimes to the grammar and said they are written the same as trait bounds.

@matthewjasper Feel free to move the grammar over to describing type bounds when you get to it, and then change the grammar for trait objects to just be "TypeBounds".

src/types.md
```diff
@@ -498,18 +498,25 @@ Because captures are often by reference, the following general rules arise:
...
 of the base trait.
 
-Trait objects are written as the path to the base trait followed by the list
-of auto traits followed optionally by a lifetime bound all separated by `+`. For
-example, given a trait `Trait`, the following are all trait objects: `Trait`,
-`Trait + Send`, `Trait + Send + Sync`, `Trait + 'static`,
-`Trait + Send + 'static`.
+Trait objects are written the same as trait bounds with the exception that all
+traits except the first trait must be auto traits and there may not be more than
+one lifetime. For example, given a trait `Trait`, the following
```
matthewjasper: and `?` bounds ("opt out bounds"?) can't be used.

src/types.md
```diff
@@ -498,18 +498,25 @@ Because captures are often by reference, the following general rules arise:
...
 
 Trait objects implement the base trait, its auto traits, and any super traits
 of the base trait.
 
-Trait objects are written as the path to the base trait followed by the list
-of auto traits followed optionally by a lifetime bound all separated by `+`. For
-example, given a trait `Trait`, the following are all trait objects: `Trait`,
-`Trait + Send`, `Trait + Send + Sync`, `Trait + 'static`,
-`Trait + Send + 'static`.
+Trait objects are written the same as trait bounds with the exception that all
```
matthewjasper: I think this would be clearer as separate sentences instead of "with the exception that".

Havvy:
I've addressed the two points.

matthewjasper:
Thanks!

PR #261, opened 2018-03-10
Moved grouped expressions to its own page

brauliobz:
and added another example of its use.

Should I remove this bit ```This operator cannot be overloaded.```? It is not really an operator.


Havvy:
LGTM. I'm hesitant to merge because I'm not sure why @matthewjasper didn't earlier.

matthewjasper:
@Havvy I didn't merge to give you a chance to give feedback if you wanted.

PR #262, opened 2018-03-10
Fixes for the match grammar

brauliobz:
- added MatchArms to make things clearer
- added attributes to both the match block and match arms
- removed pipes before the patterns, since they're experimental

### DIFF
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -2,13 +2,26 @@
 
 > **<sup>Syntax</sup>**  
 > _MatchExpression_ :  
-> &nbsp;&nbsp; `match` [_Expression_]<sub>_except struct expression_</sub> _MatchBlock_  
+> &nbsp;&nbsp; `match` [_Expression_]<sub>_except struct expression_</sub> `{`  
+> &nbsp;&nbsp; &nbsp;&nbsp; [_InnerAttribute_]<sup>\*</sup>  
+> &nbsp;&nbsp; &nbsp;&nbsp; _MatchArms_<sup>?</sup>  
+> &nbsp;&nbsp; `}`  
 >  
-> _MatchBlock_ :  
-> &nbsp;&nbsp; &nbsp;&nbsp; `{` `}`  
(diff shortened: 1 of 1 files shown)


goodmanjonathan:
Pipes before the patterns are being stabilized in 1.25.

brauliobz:
> Pipes before the patterns are being stabilized in 1.25.

Are they really? Where can I follow that? Here:

https://github.com/rust-lang/rust/milestone/44 ?

I would prefer to only consider things stable when the stable build is published.

goodmanjonathan:
[https://github.com/rust-lang/rust/pull/47947](https://github.com/rust-lang/rust/pull/47947)
[https://github.com/rust-lang/rust/pull/48374](https://github.com/rust-lang/rust/pull/48374)

I added the pipes as a prerequisite for the stabilization, so please keep them.

brauliobz:
> I added the pipes as a prerequisite for the stabilization, so please keep them.

Ok, that makes all sense :+1: .

src/expressions/match-expr.md
```diff
@@ -2,13 +2,26 @@
...
-> &nbsp;&nbsp; &nbsp;&nbsp; `}`  
+> _MatchArms_ :  
+> &nbsp;&nbsp; ( _MatchArm_ `=>` 
+>                             ( [_BlockExpression_] `,`<sup>?</sup>
+>                             | [_Expression_] `,` ) 
+>                           )<sup>\*</sup>  
+> &nbsp;&nbsp; _MatchArm_ `=>` ( [_BlockExpression_] | [_Expression_] ) `,`<sup>?</sup>  
+>  
+> _MatchArm_ :  
+> &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> _MatchArmPatterns_ _MatchArmGuard_
```
Havvy: Can you give an example of the OuterAttribute here?

brauliobz: I can't think of a useful example...

brauliobz: I tried something like this, but the attribute doesn't apply to the use of deprecated `USBType::A` and rustc warns about both things instead (unused attribute and use of deprecated item):

```
enum USBType {
    #[deprecated] A,
    B,
    C,
}

fn describe(ty: USBType) -> &'static str {
    use USBType::*;
    match ty {
        #[allow(deprecated)] A => "",
        B => "",
        C => ""
    }
}
```

Havvy: So it's allowed by the grammar, but doesn't seem to have a use case? Weird, but as long as it's legal...

Havvy:
💟 Thanks!

PR #273, opened 2018-03-18
Add 128-bit to types

mark-i-m:
cc rust-lang/rust#35118

Since this is being stabilized...


matthewjasper:
Is `size_of::<[iu]128>() == 16` is guaranteed? If so could you add them to this table
https://github.com/rust-lang-nursery/reference/blob/3f5c9c9eb4cdb1d4fd5378a7f24afa477d9fc770/src/type-layout.md#L37-L50

matthewjasper:
Thanks!

PR #323, opened 2018-05-01
Statics may sometimes inline.

Havvy:
Removed inlining guarantees as they are false e.g. in the case of using a non-mutable static integer. Added "Non-`mut`" to the statement about read-only memory. Removed the parenthesis about binding `C` statics. Added a link to destructors where it says "drop".

### DIFF
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -6,18 +6,17 @@
 >              `=` [_Expression_] `;`
 
 A *static item* is similar to a [constant], except that it represents a precise
-memory location in the program. A static is never "inlined" at the usage site,
-and all references to it refer to the same memory location. Static items have
-the `static` lifetime, which outlives all other lifetimes in a Rust program.
-Static items may be placed in read-only memory if the type is not [interior 
-mutable]. Static items do not call `drop` at the end of the program.
+memory location in the program. All references to the static refer to the same
+memory location. Static items have the `static` lifetime, which outlives all
+other lifetimes in a Rust program. Non-`mut` static items that contain a type
+that is not [interior mutable] may be placed in read-only memory. Static items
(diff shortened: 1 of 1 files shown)


src/items/static-items.md
```diff
@@ -6,18 +6,17 @@
 >              `=` [_Expression_] `;`
 
 A *static item* is similar to a [constant], except that it represents a precise
-memory location in the program. A static is never "inlined" at the usage site,
-and all references to it refer to the same memory location. Static items have
-the `static` lifetime, which outlives all other lifetimes in a Rust program.
-Static items may be placed in read-only memory if the type is not [interior 
-mutable]. Static items do not call `drop` at the end of the program.
+memory location in the program. All references to the static refer to the same
+memory location. Static items have the `static` lifetime, which outlives all
```
alercah: Should this be `'static` lifetime instead?

src/items/static-items.md
```diff
@@ -6,18 +6,17 @@
 >              `=` [_Expression_] `;`
 
 A *static item* is similar to a [constant], except that it represents a precise
```
alercah: This paragraph is just a list of properties, which makes me feel like it does not flow very well. I'm not sure the best way to address that; perhaps splitting the things about lifetimes into a separate paragraph would help, or adding more connective prose?

src/items/static-items.md
Havvy: `'static` is the `static` lifetime. I think both could work there.

src/items/static-items.md
Havvy: It definitely doesn't flow well. It reads almost like a list but without the bullet points. Should probably be addressed, but not as part of this issue.

PR #329, opened 2018-05-05
Fix link on method call page

Havvy:


Havvy:
Do you want to merge this without the Travis check? I'm not touching code, so there's no reason it would fail.

Havvy:
@alercah Waiting on merge?

PR #330, opened 2018-05-06
Fix the definition of MetaItem and MetaSeq

mjbshaw:
The current definitions mean that the following are valid attributes:

- `#[foo(,)]`
- `#[foo(, , , , ,)]`

These aren't (intended to be) valid. This fixes the definitions to only allow an EMPTY at the end of a comma-separated list.


mjbshaw:
@Havvy Good catch. Done.

Havvy:
💟 Thanks!

PR #333, opened 2018-05-12
Trying to fix the CI

Havvy:


Havvy:
@steveklabnik Is 0.1.15 close enough to the version we're using in rust-lang/rust?

Issue #279, opened 2018-03-26
`dyn` Trait (RFC 2113)


matthewjasper:
Looking to stabilise in 1.27.

- [ ] Add as (contextual) keyword
- [ ] Add to trait object documentation
- [ ] Use in examples

Havvy:
We should keep this issue open for when `dyn` becomes mandatory in the Rust 2018 edition.

ehuss:
> We should keep this issue open for when `dyn` becomes mandatory in the Rust 2018 edition.

If this is referring to the keyword becoming strict, that should be done now.  If this is something else, I'm curious what it means.

Issue #278, opened 2018-03-26
`'_` lifetime (RFC 2115)


matthewjasper:
This is stabilized in 1.26.

- [x] Add `'_` to _Lifetime_ grammar
- [x] Explain how it works in lifetime-elision.md
- [x] Use it in examples

Havvy:
Is this complete? Or are we keeping open for if it becomes required in the 2018 edition as well?

matthewjasper:
I've check for any other places where `'_` could be used and I think that there aren't any.

PR #334, opened 2018-05-12
Document underscore lifetime elision

matthewjasper:
cc #278 There don't seem to be any cases where `'_` could be used outside of this chapter.

`'_` can also be used for inferred lifetimes, but I don't think there is any documentation for them yet.

Also uses dyn trait in the elision chapter (cc #279)

### DIFF
--- a/src/lifetime-elision.md
+++ b/src/lifetime-elision.md
@@ -53,8 +53,8 @@ fn new<'a>(buf: &'a mut [u8]) -> BufWriter<'a>;         // expanded
 type FunPtr = fn(&str) -> &str;                         // elided
 type FunPtr = for<'a> fn(&'a str) -> &'a str;           // expanded
 
-type FunTrait = Fn(&str) -> &str;                       // elided
-type FunTrait = for<'a> Fn(&'a str) -> &'a str;         // expanded
+type FunTrait = dyn Fn(&str) -> &str;                   // elided
+type FunTrait = dyn for<'a> Fn(&'a str) -> &'a str;     // expanded
 ```
 
 ## Default trait object lifetimes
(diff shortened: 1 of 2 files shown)


src/lifetime-elision.md
```diff
@@ -7,8 +7,10 @@ compiler can infer a sensible default choice.
 
 In order to make common patterns more ergonomic, Rust allows lifetimes to be
 *elided* in [function item], [function pointer] and [closure trait] signatures.
-The following rules are used to infer lifetime parameters for elided lifetimes.
-It is an error to elide lifetime parameters that cannot be inferred.
+An elided lifetime can be specified either by omitting it, or by using `'_`.
+For lifetimes in paths, using `'_` is preferred. The following rules are used
```
alercah: A few things here:

- What is a lifetime in a path? Do you mean as a parameter?
- Should probably mention that omitting a lifetime in a trait object type is equivalent to `'static`, not `'_`.

alercah: Also, is `'_` not preferred for references as well? I thought omitting lifetimes was going to be deprecated.

matthewjasper:
Comments addressed.

> Also, is '_ not preferred for references as well? I thought omitting lifetimes was going to be deprecated.

The [RFC](http://rust-lang.github.io/rfcs/2115-argument-lifetimes.html#the-wildcard-lifetime) says

> When referring to a type (other than `&`/`&mut`) that requires lifetime arguments, it is deprecated to leave off those parameters.

alercah:
Ah, my bad, excellent.

src/lifetime-elision.md
```diff
@@ -41,22 +46,27 @@ fn get_mut<'a>(&'a mut self) -> &'a mut T;              // expanded
...
 
 ## Default trait object lifetimes
 
 The assumed lifetime of references held by a [trait object] is called its
 _default object lifetime bound_. These were defined in [RFC 599] and amended in
-[RFC 1156]. Default object lifetime bounds are used instead of the lifetime
-parameter elision rules defined above.
+[RFC 1156].
+
+> These default object lifetime bounds are used instead of the lifetime
```
Havvy: Is this supposed to be a Note?

matthewjasper: Well, it was intentional. I've changed it back to normal prose now.

src/lifetime-elision.md
```diff
@@ -5,10 +5,14 @@ compiler can infer a sensible default choice.
 
 ## Lifetime elision in functions
 
-In order to make common patterns more ergonomic, Rust allows lifetimes to be
-*elided* in [function item], [function pointer] and [closure trait] signatures.
-The following rules are used to infer lifetime parameters for elided lifetimes.
-It is an error to elide lifetime parameters that cannot be inferred.
+In order to make common patterns more ergonomic, Rust allows lifetime argument
```
Havvy: Can you remove the "Rust allows" here?

PR #335, opened 2018-05-15
Fix #331

### LINKED ISSUE #331: [site] "file an issue" link is (sometimes) unclickable
Depending on the browser window width, if the "file an issue" link is positioned on the left, it gets obscured by the "go back" button(?) which makes filing the issue from the page impossible.

See the screenshot for potentially better explanation: 
firefox_2018-05-11_10-29-02

Havvy:
Fixes #331

There's a small amount of jank in that without the sidebar, once you scroll down, there's some area you would sort of expect to be able to click the button, but you cannot. But given most people would aim for the arrow, I don't think this is too much of an issue.


matthewjasper:
Thanks!

PR #339, opened 2018-05-17
Add u128 and i128 tokens

markmmm:
Additionally all other tokens were sorted `u..` first, then `i..` - so I re-ordered `usize` & `isize` to conform


markmmm:
Somebody beat this commit to add u128 & i128 - after merging from master - there is only the order change of isize & usize to match the other types.

Havvy:
💟 That was bstrie. Both of your PRs were pretty close to each other, but their's were the one I was looking at first...

PR #341, opened 2018-05-19
Improve supertraits documentation

Havvy:
Some of it feels like it was from the guide. Should all be referency now. And also more cross linking.

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -9,8 +9,8 @@ interface consists of [associated items], which come in three varieties:
 
 All traits define an implicit type parameter `Self` that refers to "the type
 that is implementing this interface". Traits may also contain additional type
-parameters. These type parameters (including `Self`) may be constrained by
-other traits and so forth [as usual].
+parameters. These type parameters, including `Self`, may be constrained by
+other traits and so forth [as usual][generics].
 
 Traits are implemented for specific types through separate [implementations].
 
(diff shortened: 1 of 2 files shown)


src/items/traits.md
```diff
@@ -10,7 +10,7 @@ interface consists of [associated items], which come in three varieties:
 All traits define an implicit type parameter `Self` that refers to "the type
 that is implementing this interface". Traits may also contain additional type
 parameters. These type parameters (including `Self`) may be constrained by
-other traits and so forth as usual.
+other traits and so forth [as usual].
```
alercah: Purely a style nit so feel free to disagree: would it make sense to have [as usual][generics] so it's more clear to someone reading the source where pothole links are going?

alercah:
The number of examples still feels a bit guide-y to me? Perhaps consider collapsing them down into one or two examples.

Havvy:
I disagree. The examples each showcase a different facet of supertraits, not teach how they work. I did end up removing an example of a struct implementing both traits since it didn't showcase anything specific.

src/items/traits.md
Havvy: I'm heading to work now, so I'll change it when I get home.

Havvy:
Made the "as usual" link easier to follow in markdown source, added a definition for subtract, and pared down the writing for one example.

PR #342, opened 2018-05-21
Remove incorrect coercion. Define type coercions.

Havvy:


src/type-coercions.md
```diff
@@ -53,7 +53,7 @@ sites are:
   }
   ```
 
-* Function results, either the final line of a block if it is not
+* Function results &ndash; either the final line of a block if it is not
```
alercah: Style nitpick: shouldn't this be an em dash without surrounding spaces?

src/type-coercions.md
```diff
@@ -91,7 +91,7 @@ the block has a known type.
 
 Coercion is allowed between the following types:
 
-* `T` to `U` if `T` is a subtype of `U` (*reflexive case*)
+* `T` to `U` if `T` is a [subtype] of `U` (*reflexive case*)
```
alercah: Does the "reflexive case" annotation mean anything here? T to U when they are distinct is not reflexivity, although this does imply reflexivity because T is a subtype of T

src/type-coercions.md
```diff
@@ -164,8 +164,7 @@ the compiler will provide an implementation of `Unsize<U>` for `T`:
 
 * `[T; n]` to `[T]`.
 
-* `T` to `U`, when `U` is a trait object type and either `T` implements `U` or
-  `T` is a trait object for a subtrait of `U`.
+* `T` to `dyn U`, when `T` implements `U + Sized`
```
alercah: I don't think `+ Sized` is accurate here; without it, coercions of references won't always work because going from `&T` to `&U` would require that `T` be sized. Unless that's actually how it works?

alercah:
@Havvy Can you make the changes for this one?

Havvy:
This is outdated.

Issue #344, opened 2018-05-22
Never type was unstabilized


Havvy:
We need to back out #206 until it's re-stabilized.

ehuss:
@Centril I'm guessing that the reference should probably include more detail on the never type (https://github.com/rust-lang/reference/blob/master/src/types/never.md is pretty slim), would you agree?  Is that something you'd like to do?  If not, I can probably try to figure out what could be added, but an outline of what would be useful to add would be helpful.

(And congratulations on getting it stabilized. 😄 )

Centril:
@ehuss It seems roughly in good shape, but I would probably extend that with:

- An explanation of that it is equivalent to `enum Void {}` except for the implicit coercion behavior. -- contrast & compare.
- A note wrt. the inference fallback being `()` (for now, with a link to https://github.com/rust-lang/rust/issues/65992).

If you have a sec to file the PR that would be neat!

> (And congratulations on getting it stabilized. 😄 )

Thanks! (I wonder how long it lasts... 😅)

ehuss:
Closing this as it has been re-stabilized. https://github.com/rust-lang/rust/pull/65355

PR #347, opened 2018-05-26
Fix never type link

MajorBreakfast:
This PR fixes the inconsistent spelling that currently breaks the link.

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -32,7 +32,7 @@ language:
 [machine types]: #machine-types
 [machine-dependent integer types]: #machine-dependent-integer-types
 [textual types]: #textual-types
-[never-type]: #never-type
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
This accidentally also contains stuff from my other PR. Fixing now...

MajorBreakfast:
Ok now it contains just the one change I wanted to make.

alercah:
I'd almost rather the never type link be broken, because of it being unstabilized. Hm.

MajorBreakfast:
Maybe remove the never type section entirely?

alercah:
If you want to take on #344, that would be awesome! But don't feel like you have to. :)

MajorBreakfast:
👍

PR #350, opened 2018-06-01
Attributes in generics position and general attributes improvements

Havvy:
Attributes in generics is stabilized in 1.27 as per https://github.com/rust-lang/rust/pull/48851/.

I moved the defining italics for attributes to the actual definition and gave inner/outer attributes actual definition lines instead of leaving it implied.

There's also a few newlines for 80 character style and removing trailing spaces in here.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -21,11 +21,11 @@
 > &nbsp;&nbsp; | _MetaItem_  
 > &nbsp;&nbsp; | _MetaItem_ `,` _MetaSeq_
 
-Any item declaration may have an _attribute_ applied to it. Attributes in Rust
-are modeled on Attributes in ECMA-335, with the syntax coming from ECMA-334
-(C#). An attribute is a general, free-form metadatum that is interpreted
-according to name, convention, and language and compiler version. Attributes
-may appear as any of:
+Any [item declaration] or [generic lifetime or type parameter][generics] may
+have an attribute applied to it. Attributes are modeled on Attributes in
+[ECMA-335], with the syntax coming from [ECMA-334] \(C#). An _attribute_ is a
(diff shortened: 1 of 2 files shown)


src/items/generics.md
```diff
@@ -40,21 +40,21 @@ referred to with path syntax.
 
 ## Where clauses
 
-> **<sup>Syntax</sup>**  
-> _WhereClause_ :  
+> **<sup>Syntax</sup>**
+> _WhereClause_ :
```
matthewjasper: These spaces are needed for line breaks. They could be replaced with `<br>`s if you prefer.

src/attributes.md
```diff
@@ -21,11 +21,11 @@
 > &nbsp;&nbsp; | _MetaItem_  
 > &nbsp;&nbsp; | _MetaItem_ `,` _MetaSeq_
 
-Any item declaration may have an _attribute_ applied to it. Attributes in Rust
-are modeled on Attributes in ECMA-335, with the syntax coming from ECMA-334
-(C#). An attribute is a general, free-form metadatum that is interpreted
-according to name, convention, and language and compiler version. Attributes
-may appear as any of:
+Any [item declaration] or [generic lifetime or type parameter][generics] may
```
matthewjasper: or statement

Havvy: Or block expression

Havvy:
I _may_ have made more changes than just the nits.

src/attributes.md
Havvy: Actually, this is a lot more complicated, because also enum variants, struct and union fields, match arms...

matthewjasper:
Thanks!

PR #352, opened 2018-06-04
General attribute docs improvements

Havvy:
Each commit does its own small thing.

This builds on top of #350, so merge that before taking a look at this.

### DIFF
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -136,11 +136,17 @@ impl Seq<bool> for u32 {
 }
 ```
 
-## Attributes in Implementations
+## Attributes on Implementations
 
-Implementations may contain inner [attributes] inside the brackets that contain
-the associated items. They must come before the associated items.
-
+Implementations may contain outer [attributes] before the `impl` keyword and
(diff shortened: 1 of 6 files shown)


src/items/implementations.md
```diff
@@ -136,5 +136,11 @@ impl Seq<bool> for u32 {
 }
 ```
 
+## Attributes in Implementations
+
+Implementations may contain inner [attributes] inside the brackets that contain
+the associated items. They must come before the associated items.
```
matthewjasper: Could there be a sentence on what attributes are allowed on impls?

Havvy:
Addressed comments. Also added which attributes are allowed on block expressions. And apparently, as per the example I included on the block expressions page, block expressions also allow outer attributes and the only place you'd be able to tell that isn't overlapping with statements taking attributes is the return expression of a block expression. Go figure.

I'll add which attributes are allowed on functions / modules / the main module/ other items / variants and fields later.

Havvy:
Rebased against master due to merge conflict.

Havvy:
Rebased again for another merge conflict.

PR #353, opened 2018-06-06
Import #[repr(transparent)] docs from the Unstable book

SimonSapin:
… as the feature is ready for stabilization: https://github.com/rust-lang/rust/issues/43036#issuecomment-394094318

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -294,6 +294,166 @@ a `packed` type cannot transitively contain another `align`ed type.
 > possible to [safely create unaligned pointers to `packed` fields][27060].
 > Like all ways to create undefined behavior in safe Rust, this is a bug.
 
+### The `transparent` representation
+
+#### Background
+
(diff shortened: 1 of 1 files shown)


rkruppe:
I don't think this text really fits into the reference, as it explains a lot of motivation and background rather than precisely and concisely stating what the attribute does. Certainly when I wrote it I had something like The Book or RBE in mind, not the reference.

SimonSapin:
Feature stabilization PR: https://github.com/rust-lang/rust/pull/51395

I agree that the subsections do not necessarily belong there as-is but I wasn’t sure what else to do, this PR is sort of a straw man.

alercah:
I agree that this is not ready for integration into the reference. Adding more/confusing text is not necessarily better, even if it improves coverage.

Havvy:
Closing in favor of #389 since the RFC's contents don't translate to being in the reference directly that well.

PR #356, opened 2018-06-10
Update array expressions

matthewjasper:
* const_err is now deny-by-default (not sure how this didn't get noticed by Rust's CI)
* Indexes are checked on non-const arrays.
* Fix some typos


Havvy:
💟 Thanks!

PR #358, opened 2018-06-12
Placeholder docs for impl Trait

Havvy:
r? @matthewjasper 

While we wait for the comprehensive docs, it'd be nice to have a placeholder so that we at least acknowledge it exists.

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -654,17 +654,31 @@ fn to_vec<A: Clone>(xs: &[A]) -> Vec<A> {
 Here, `first` has type `A`, referring to `to_vec`'s `A` type parameter; and
 `rest` has type `Vec<A>`, a vector with element type `A`.
 
-## Abstract types
+## Anonymous type parameters
 
 > Note: This section is a placeholder for more comprehensive reference
 > material.
 
-Abstract types are types that stand in for another concrete type where the
+> Note: This is often called "impl Trait in argument position".
+
+Functions can declare an argument to be an anonymous type parameter where the
+callee must provide a type that has the bounds declared by the anonymous type
+parameter and the function can only use the methods available by the trait
+bounds of the anonymous type parameter.
+
+They are written as `impl` followed by a set of trait bounds.
+
+## Abstract return types
+
+> Note: This section is a placeholder for more comprehensive reference
+> material.
+
+> Note: This is often called "impl Trait in return position".
+
+Functions, except for associated trait functions, can return an abstract
+return type. These  types stand in for another concrete type where the
 use-site may only use the trait methods declared by the trait bounds of the
 type.
-
-Abstract types are only allowed as the type of a parameter or return type of a
-function except for associated functions on traits.
 
 They are written as `impl` followed by a set of trait bounds.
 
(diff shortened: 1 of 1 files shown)


Havvy:
Modified this a bit due to discussion on IRC:

```
<isHavvy> What kind of type is "impl Trait"?
<est31> it is an universal type if in argument position
<est31> it is an existential type if in return position
<isHavvy> est31: I ultimately looked at the RFC and they called it an "Abstract type".
<est31> you mean the named existentials RFC?
<est31> or which RFC do you mean
<isHavvy> Conservative impl trait RFC
<isHavvy> "Abstract return type" specifically is the wording they use.
<est31> IIRC it didnt have any argument position impl trait
<est31> that's the thing the aturon RFC added
<rkruppe> the "return" is an important bit IMO, since existential impl trait isn't really any more abstract than generic parameters (which is to say, not entirely b/c we don't have parametricity)
<est31> we do have variance
<est31> it's just made to be variant on *all* input types of the function
<est31> even those on which aren't parameters of the actual return type
<est31> but yeah we don't have -> impl Trait<A,B,C>
<est31> that would run against the goals of impl Trait in argument position though
<est31> named existentials / abstract types require explicit variance
<isHavvy> So are they called "universal type" and "existential type" in type theory then? I figured that was just a property they had. After all, aren't generics also existential types?
<isHavvy> Err, universal*
<est31> idk about type theory, it's just the words being thrown around in the discussions :p
* est31 like a parrot
<est31> http://cultofthepartyparrot.com/parrots/hd/parrot.gif
<rkruppe> isHavvy: generics in general are universal types, impl Trait in argument position is just universal because it's another way to spell generics
<rkruppe> isHavvy: likewise there's plenty more to existential types than the little "abstract return types" that are impl Trait in return position, though unlike with universal types rust right now only has impl Trait in this space
<isHavvy> Alright. That all makes sense. So, is there an actual name to "impl Trait in argument position" that doesn't say "impl Trait"?
<rkruppe> also, although i've parrotted "universal type" here i am pretty unsure whether one can rightfully call "impl Trait" (or plain old T) on its own a type in the same sense e.g. char or Vec<i32> or &[&str] is a type. it's always situated in a function signature and pretty meaningless outside of it. and when type checking the function body there's just another free type *variable* corresponding to it
<isHavvy> It's a type in the sense that it's a predicate on what values are allowed, even if it's masking another type like e.g. generics do.
<isHavvy> "Anonymous generic type"?
<est31> yeah something like that
```

src/types.md
```diff
@@ -654,6 +654,34 @@ fn to_vec<A: Clone>(xs: &[A]) -> Vec<A> {
...
+
+> Note: This section is a placeholder for more comprehensive reference
+> material.
+
+> Note: This is often called "impl Trait in argument position".
+
+Functions can declare an argument to be an anonymous type parameter where the
+callee must provide a type that has the bounds declared by the anonymous type
+parameter and the function can only use the methods available by the trait
+bounds of the anonymous type parameter.
```
matthewjasper: Long sentence without punctuation.

Havvy: I'm not too worried, since it's just there to be a placeholder.

matthewjasper:
Thanks!

PR #360, opened 2018-06-18
Fix grammar in the "Visibility and Privacy" section

s3rvac:
_This equivalent_ -> _This is equivalent_

I have re-wrapped the line as it was too long after the added "is".


Havvy:
💟 Thanks!

PR #362, opened 2018-06-20
document #[panic_handler]

japaric:
r? @steveklabnik 

Required by rust-lang/rust#51366


src/beneath-std.md
```diff
@@ -0,0 +1,10 @@
+# Beneath `std`
+
+This section documents (or will document) features that are provided by the standard library and
+that `#![no_std]` developers have to deal with (i.e. provide) to build `#![no_std]` binary crates. A
+(likely incomplete) list of such features is shown below:
+
+- #[lang = "eh_personality"]
```
Havvy: We explicitly do not document lang items and anything unstable. Listing them goes against that.

src/panic-implementation.md
```diff
@@ -0,0 +1,35 @@
+## #[panic_implementation]
+
+The `#[panic_implementation]` attribute can only be applied to a function with signature
```
Havvy: The attribute name is written without the `#[]` surrounding it.

src/panic-implementation.md
```diff
@@ -0,0 +1,35 @@
...
+`fn(&PanicInfo) -> !`. The function marked with this attribute defines the behavior of `panic!` in
+`#![no_std]` applications. There must be a *single* `#[panic_implementation]` function in the
+dependency graph of a binary / dylib / cdylib crate.
```
Havvy: Use commas, not `/`s.

src/panic-implementation.md
```diff
@@ -0,0 +1,35 @@
...
+
+The `PanicInfo` struct contains information about the location of the panic. The API of `PanicInfo`
```
Havvy: Put the link around the first `PanicInfo`. E.g. ```[`PanicInfo`]```.

src/panic-implementation.md
```diff
@@ -0,0 +1,35 @@
...
+can be found in the [API docs]. As of 1.28-beta (2018-06-21) there's a difference between
```
Havvy: We don't discuss anything that's not stable. This includes beta versions. Nor do we discuss older versions.

japaric:
@Havvy Thanks for the review. I think I have addressed all your comments in the last commit. Let me know if I missed something.

src/beneath-std.md
```diff
@@ -0,0 +1,39 @@
+# Beneath `std`
+
+This section documents features that are provided by the standard library and that `#![no_std]`
+developers have to deal with (i.e. provide) to build `#![no_std]` binary crates. A list of such
```
Havvy: "deal with (i.e. provide)" should just be "provide".

src/SUMMARY.md
```diff
@@ -87,6 +87,8 @@
     - [Behavior considered undefined](behavior-considered-undefined.md)
     - [Behavior not considered unsafe](behavior-not-considered-unsafe.md)
 
+- [Beneath `std`](beneath-std.md)
```
Havvy: I don't like this name. At all. It implies this is information only useful for `std`, since it's stuff beneath it. And it also implies the things that `std` can do that other crates cannot. Perhaps something like "Runtime", since the things you have to implement are all going to be what's effectively the Rust Runtime. Then we would also be able to document what the standard behaviour does.

japaric:
Hey, @Havvy. Sorry for the radio silence. `panic_handler` (previously `panic_implementation`) is in FCP now so I have picked up the work here.

I have addressed your comment about renaming the section to "the Rust runtime" and have documented the standard behavior.

src/runtime.md
```diff
@@ -0,0 +1,43 @@
...
+    let _ = writeln!(sink, "{}", info);
+
+    loop {}
+}
+```
+
+### Standard behavior
+
+The standard library provides an implementation of `panic_handler` than can be
+statically customized using the `-C panic` flag. `-C panic=abort` makes panics
```
japaric: I'm wasn't 100% sure if we can refer to `rustc` flags in the reference.

src/runtime.md
```diff
@@ -0,0 +1,43 @@
...
+abort the process, and `-C panic=unwind` makes panics unwind the panicking
+thread. If no panicking behavior is specified using `-C panic` one of these two
+behaviors is chosen according to the compilation target.
```
japaric: I could expand here and mention that the default behavior is in the specification of the compilation target but target specifications are unstable. (See `rustc -Z unstable-options --print target-spec-json`).

src/runtime.md
japaric: The standard library also lets you customize the panicking behavior at runtime via the [`panic::set_hook`](https://doc.rust-lang.org/std/panic/fn.set_hook.html) but I wasn't sure if we should document standard API here.

japaric: Ok, I have simplified this to point to the book.

Havvy: The fact that panic behavior can be changed is a language level detail and should be documented. Linking to the function that shows how to change it would be useful.

src/runtime.md
Havvy: A table that contains the information would perhaps be useful. Probably don't want to do that in this PR though, since this one is shaping up to be mergable.

Centril:
Ping from triage, @japaric what's the status of this PR?

ehuss:
I have rebased this PR and updated for the review comments.

I noticed nothing really describes "panics" or panic behavior in the reference, seems like something that would be good for the future.

I was thinking maybe in a future PR that `global_allocator` could be moved to the runtime page. However, I'm not really confident that the allocator is part of the "runtime".  What do people consider is part of the Rust "runtime"?  Where is the line drawn?

Centril:
@ehuss This is a good start; merge when you feel like it :)

PR #363, opened 2018-06-24
Fix hard breaks for weak keywords.

ehuss:
The two trailing spaces to indicate a hard break were missing for
the list of weak keywords.

### DIFF
--- a/src/keywords.md
+++ b/src/keywords.md
@@ -96,9 +96,9 @@ is possible to declare a variable or method with the name `union`.
   followed by a path that does not start with `::`.
 
 > **<sup>Lexer</sup>**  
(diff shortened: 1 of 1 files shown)


ehuss:
Note: Using a trailing backslash might be safer than relying on invisible markup, but using spaces follows the style of the rest of the project.

Havvy:
LGTM. If you want to switch to trailing backslash, I'd be okay with that too, given both me and another person have inadvertently tried deleting them already...

Havvy:
@alercah @matthewjasper on question of trailing backslashes

Havvy:
Closing in favor of #365

PR #365, opened 2018-06-24
Use backslash instead of whitespace for hard breaks.

ehuss:
The only formatting change is to add breaks on the weak keywords in
`keywords.md`.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -1,24 +1,24 @@
 # Attributes
 
-> **<sup>Syntax</sup>**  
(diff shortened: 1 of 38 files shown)


ehuss:
Alternative solution to #363. Arguably not as pretty when looking at the raw markdown, but possibly less likely to make mistakes.

Havvy:
Soo much whitespace. 💟 Thanks for changing to backslashes and normalizing ending space in other places too.

PR #366, opened 2018-06-27
Improve verbiage around let statements and variable initialization

Havvy:
This gets rid of the `and/or` in the "`let` statements" section and gives an example showing that Rust tracks initialization of variables for uses.


matthewjasper:
Thanks!

PR #367, opened 2018-07-01
Mention the global_allocator trait

brson:
I don't know much about the organization of this book, but this seems an ok place to put it.

cc https://github.com/rust-lang/rust/pull/51974

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -161,6 +161,8 @@ information on macro scope.
 
 - `export_name` - on statics and functions, this determines the name of the
   exported symbol.
+- `global_allocator` - when applied to a static item implementing the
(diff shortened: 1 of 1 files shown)


brson:
I didn't know how to link `GlobalAlloc`.

steveklabnik:
This should be on hold until https://github.com/rust-lang/rust/pull/51974#issuecomment-401678445

Havvy:
I don't see how that comment relates to the contents of this PR. It doesn't mention what the default allocator is either.

Havvy:
Given that `global_alloc` is stable, better to have at least a stub of documentation vs. no documentation at all. We want the list to at least be complete.

PR #369, opened 2018-07-15
[RFC 231] Capture mode and closure trait rules

crlf0710:
I always have to open RFC 231 to check when i need to explain these rules to others 
So i copied these text (with minimal editorial changes) from RFC 231 to The Rust Reference so people can read them more easily.

### DIFF
--- a/src/expressions/closure-expr.md
+++ b/src/expressions/closure-expr.md
@@ -24,9 +24,9 @@ Closure expressions are most useful when passing functions as arguments to other
 functions, as an abbreviation for defining and capturing a separate function.
 
 Significantly, closure expressions _capture their environment_, which regular
-[function definitions] do not. Without the `move`
-keyword, the closure expression infers how it captures each variable from its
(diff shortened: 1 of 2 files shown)


Havvy:
I'm too tired to properly review this, but this is definitely the kind of improvements I like to see. :smile:

alercah:
Thanks! There's a fair bit of this content already on the [closure types](https://doc.rust-lang.org/nightly/reference/types.html#closure-types) page, can you please make sure that it isn't redundant?

crlf0710:
@alercah Oops, i somehow overlooked those text. So I replaced my patch with just adding a heading there and linked to it. How does that look?

alercah:
Looks good! I had some WIP on fixing up this section to be more consistent overall with the types section, but I lost the computer it was on, and this is a definite improvment! If you want to try to work more in this space and get the two meshing better, go for it. :+1: 

Thanks for the PR! :heart:

PR #376, opened 2018-07-27
Expand block expression section

Havvy:
1. Don't consider semicolons part of the statement. As per the
   `stmt` macro matcher, they're not a part of it. They are a part
   of the block expression though, so the grammar is updated to
   show that.

2. Try to better describe what a block expression is. By saying
   it's like a module, we're probably confusing people.

3. Show an example of using a block expression to force moving a
   value out of a variable. I personally didn't understand what
   that fully entailed until I wrote the example.

4. Link to the unsafety section instead of redescribing what
   unsafe is allowed to do in the unsafe block section.


matthewjasper:
Thanks!

PR #380, opened 2018-07-30
Add impl trait and remove others from undocumented list

Havvy:
About removals:

* Attributes on match arms was documented by me last time I looked at attributes.
* UFCS a.k.a. diambiguating function calls is documented on the call expression page.
* unaligned access appears to just be library things

### DIFF
--- a/src/undocumented.md
+++ b/src/undocumented.md
@@ -11,7 +11,6 @@ to shrink!
   not for everything: e.g. coherence and orphan rules.
 - [Flexible target specification] - Some---but not all---flags are documented
   in [Conditional compilation]
-- Require parentheses for chained comparisons
 - [`dllimport`] - one element mentioned but not explained at [FFI attributes]
 - [define `crt_link`]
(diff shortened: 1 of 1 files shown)


src/undocumented.md
```diff
@@ -9,24 +9,20 @@ to shrink!
 - [`libstd` facade]
 - [Trait reform] – some partial documentation exists (the use of `Self`), but
   not for everything: e.g. coherence and orphan rules.
-- [Attributes on `match` arms] – the underlying idea is documented in the
-  [Attributes] section, but the applicability to internal items is never
-  specified.
 - [Flexible target specification] - Some---but not all---flags are documented
   in [Conditional compilation]
-- [Require parentheses for chained comparisons]
+- Require parentheses for chained comparisons
```
matthewjasper: This is documented [here](https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#comparison-operators)

Havvy:
Added in a new commit.

PR #381, opened 2018-08-01
Add optimization hints section to attributes

### LINKED ISSUE #349: `#[inline]` section could use additional details
In trying to choose whether to use `#[inline]` on a function, I found the [section](https://doc.rust-lang.org/reference/attributes.html#inline-attribute) in this book. However, here are several questions I had after reading it: 

- Does `#[inline]` apply to all implementations of a function defined by a trait, if the trait's declaration of that function is annotated?
- Does `#[inline]` apply to an implementation of a trait function if the trait's declaration does not have an `#[inline]` annotation?
- Will rustc ever inline a function from another crate without `#[inline]`?
- Will rustc inline a function from another crate if it is annotated with `#[inline]`?
- In what types of situations would incorrectly using this annotation cause programs to be slower? What makes those uses of inline slower? I'm guessing it's related to keeping the instruction size small and cache-friendly, but I don't know for sure.

I volunteer to write the answers to these questions in polished form for the book and submit a pull request if someone can answer the substance of them here.

Havvy:
Fixes #349 

Went and did this because it's been two months without a response.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -126,8 +126,6 @@ names have meaning.
   - `ignore` - indicates that this test function is disabled.
 - `should_panic` - indicates that this test function should panic, inverting the
   success condition.
-- `cold` - The function is unlikely to be executed, so optimize it (and calls
-  to it) differently.
 
 ## FFI attributes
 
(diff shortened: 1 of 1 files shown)


Havvy:
Do not merge both this and the next PR at the same time until I've edited the other PR to remove the "function-only attributes" section since the two PRs remove all of the attributes from that section.

Havvy:
Rebased

PR #382, opened 2018-08-01
Test attributes deserve their own section

Havvy:
cc @QuietMisdreavus and @steveklabnik for possibly out-of-scope documentation here.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -243,8 +243,8 @@ type must be one of the following:
 > Note: The test harness is ran by passing the `--test` argument to `rustc` or
 > using `cargo test`.
 
-Tests that return `()` pass as long as they terminate do not panic. Tests that
-return a `Result` pass as long as they return `Ok(())`. Tests that do not
+Tests that return `()` pass as long as they terminate and do not panic. Tests
+that return a `Result` pass as long as they return `Ok(())`. Tests that do not
 terminate neither pass nor fail.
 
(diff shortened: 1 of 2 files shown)


Havvy:
Do not merge both this and the previous PR at the same time until I've edited the other PR to remove the "function-only attributes" section since the two PRs remove all of the attributes from that section.

Havvy:
With the other PR merged, this one now also handles removing the entirety of the "Function only items" section and can be merged.

Hmm, I probably should have done that in a new commit, but it's 2am and I'm not thinking straight. But wanted to get this unblocked before I head to bed. Sorry. The only change since the previous comment is just the removal of that header.

src/attributes.md
```diff
@@ -228,6 +220,42 @@ are transformed into `doc` attributes.
...
+> Note: The implementation of which return types are allowed is determined by
+> the unstable [`Termination`] trait.
+
+<!-- If the previous section needs updating (from "must take no arguments"
+  onwards, also update it in the crates-and-source-files.md file -->
+
+> Note: The test harness is ran by passing the `--test` argument to `rustc` or
+> using `cargo test`.
+
+Tests that return `()` pass as long as they terminate do not panic. Tests that
```
matthewjasper: missing 'and' after 'terminate'

Havvy: Fixed.

matthewjasper:
Thanks!

PR #385, opened 2018-08-02
Fix some broken links.

ehuss:


Havvy:
💟 Thanks for doing this. You're aweomse!

PR #386, opened 2018-08-03
Fix more broken links.

ehuss:
Sorry, missed one in my last commit.  Also, #380 recently removed the link, but
not the actual entry.


Havvy:
💟 Thanks! We really need a tool for this.

PR #387, opened 2018-08-03
Attributes on functions

Havvy:
This is blocked on #381 and #382 for correct links.

1. The main task of this PR is to show which attributes are allowed on functions.
2. I renamed the "`must_use` Attribute" section to `must_use` to match the "`derive`" section further on in the page.
3. Updated links to the `cfg` attribute to the new conditional compilation page. Probably should have done this when moving `cfg` to its own page. But doing it here where I'm adding a new link to it.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -331,7 +331,7 @@ pub mod m3 {
 }
 ```
 
-#### `must_use` Attribute
+#### `must_use`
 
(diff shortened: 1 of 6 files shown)


Havvy:
Both are good calls. I've improved the sentence flow around both.

matthewjasper:
Thanks!

PR #395, opened 2018-08-09
Updated section on identifiers for raw identifiers

alexreg:


src/identifiers.md
```diff
@@ -5,10 +5,12 @@
...
 > &nbsp;&nbsp; | `_` [`a`-`z` `A`-`Z` `0`-`9` `_`]<sup>+</sup>
 >
+> RAW_IDENTIFIER : `r#` IDENTIFIER_OR_KEYWORD
+>
 > IDENTIFIER :\
-> IDENTIFIER_OR_KEYWORD <sub>*Except a [strict] or [reserved] keyword*</sub>
+> IDENTIFIER_OR_KEYWORD <sub>*Except a [strict] or [reserved] keyword*</sub> | RAW_IDENTIFIER
 
-An identifier is any nonempty ASCII string of the following form:
+A normal (non-raw) identifier is any nonempty ASCII string of the following form:
```
Havvy: I don't think this line needs to change. I feel like it draws attention away from basic identifiers in a way that could be confusing.

src/identifiers.md
```diff
@@ -21,5 +23,7 @@ Or
 * The identifier is more than one character. `_` alone is not an identifier.
 * The remaining characters are alphanumeric or `_`.
 
+A raw identifier is like a normal identifier, but prefixed by `r#`, which is not included as part of the identifier. Unlike a normal identifier, a raw identifier may be any strict or reserved keyword.
```
Havvy: Nit: Comma splice. Trade the last `,` for a `.` and replace `which` with ```The `r#` ```

matthewjasper: "...any strict or reserved keyword" except `crate`, `extern`, `self`, `Self` or `super`.

src/identifiers.md
alexreg: Yeah, I thought about this... you definitely have a point. I just wasn't sure if it would be misleading to "go back" on the initial definition of an identifier slightly. But fair enough.

src/identifiers.md
alexreg: Hmm I think you misread perhaps? It wouldn't make sense like that grammatically.

alexreg: Okay, I thought there might be exceptions, thanks.

Havvy: ```A raw identifier is like a normal identifier, but prefixed by `r#`. The `r#` is not included as part of the identifier.```

alexreg: I'm not sure that improves clarity, but if you really want, I'll do it. 😉

Havvy: Actually, looking at it closer, another option is to change ", but" to "except". So ```A raw identifier is like a normal identifier except prefixed by `r#`, which is not included as part of the identifier.``` The main thing is to get rid of one of the commas as it's currently grammatically murky with both of them.

alexreg:
The grammar is technically okay as-is, but if it made you misread it, then it could plainly be clearer! I don’t want to confuse others after all. I think I’ll go with your first suggestion if that’s alright. (You still need the comma after “identifier” in your second suggestion.)

alexreg:
Can we get this merged now that the actual Rust feature has been stabilised? :-)

Havvy:
Yes. 💟 Thanks for the contribution!

PR #396, opened 2018-08-10
`derive` implements traits, not types

scooter-dangle:
It's a pretty small thing, and I don't have any examples of anyone being led astray, but it's technically incorrect to say that one would implement `derive` for a 'type'.

- - -

Note: Didn't intend to add in the final line's terminator at the end of the file—Github in-browser editor artifact. I can remove it if it's objectionable.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -454,7 +454,7 @@ impl<T: PartialEq> PartialEq for Foo<T> {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


scooter-dangle:
Just saw this fix is already pending in https://github.com/rust-lang-nursery/reference/pull/381, closing this PR in favor of that one.

PR #397, opened 2018-08-10
admit we don't have a memory model

steveklabnik:
I think this is better than the previous page, what do you all think?


Havvy:
I have no qualms with getting rid of what was there before since while it was originally useful information before the Book, it's sort of lost usefulness as reference material.

Issue #372, opened 2018-07-25
Implementations page is hard to read.


alercah:
The [implementations](https://doc.rust-lang.org/nightly/reference/items/implementations.html) page is difficult to read because it has many one-sentence paragraphs without elaboration or examples.

JoshBrudnak:
@alercah I would like to help out with this. :+1:

alercah:
Awesome, please do!

alercah:
Let me know if you need any help.

JoshBrudnak:
This issue was fixed in #401 it can probably be closed. :+1:

troysgit:
@alercah @JoshBrudnak 
Requesting this be closed to keep the 'Easy' label tasks up to date.

PR #401, opened 2018-08-15
Combine related paragraphs and add more details for implementations

JoshBrudnak:


JoshBrudnak:
Fixes: #372

src/items/implementations.md
```diff
@@ -1,32 +1,31 @@
...
-the associate item's path component.
-
-Inherent implementations cannot contain associated type aliases.
+the associate item's path component. Inherent implementations cannot contain
+associated type aliases.
 
-A type can have multiple inherent implementations.
-
-The implementing type must be defined within the same crate.
+A type can also have multiple inherent implementations. Although the
```
Havvy: The "Although" used to join here feels very artificial.

Havvy:
LGTM. I didn't realize when I did all the changes around here I made the paragraphs so separated like that. My only quibble is that the "Although" joiner really is joining two disparate pieces of information. But that's ultimately a quibble.

Since I've touched this page so much, I'd rather the r+ come from another person before merging.

src/items/implementations.md
JoshBrudnak: Yes, what if I use "but" instead of "although"

Havvy: The "but" implies an exception on the previous statement. An "and" would do well if it cannot read well as two sentences.

JoshBrudnak: @Havvy  Does this sound better: "An implementing type must be defined within the same crate as the original type definition."

JoshBrudnak:
@Havvy I separated that second clause into "An implementing type must be defined within the same crate as the original type definition.".

JoshBrudnak:
@alercah can you review my changes? Thanks :100:

Havvy:
@alercah has been busy, and @matthewjasper has given their seal of approval. So gonna merge it.

💟 Thanks!

PR #405, opened 2018-08-23
Expand documentation on procedural macros

alexcrichton:
In preparation for the 1.30 stabilization I figured I'd get started and help
write some documentation!

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -1,24 +1,549 @@
 ## Procedural Macros
 
 *Procedural macros* allow creating syntax extensions as execution of a function.
-Procedural macros can be used to implement custom [derive] on your own
-types. See [the book][procedural macros] for a tutorial.
+Procedural macros currently come in one of three flavors:
 
-Procedural macros involve a few different parts of the language and its
-standard libraries. First is the `proc_macro` crate, included with Rust,
-that defines an interface for building a procedural macro. The
-`#[proc_macro_derive(Foo)]` attribute is used to mark the deriving
-function. This function must have the type signature:
+* Bang macros - `my_macro!(...)`
+* Derive macros - `#[derive(MyTrait)]`
+* Attribute macros - `#[my_attribute]`
 
-```rust,ignore
-use proc_macro::TokenStream;
+Procedural macros allow you to run code at compile time that operates over Rust
+syntax, both consuming and producing Rust syntax. You can sort of think of
+procedural macros as Rust functions from an AST to another AST.
 
-#[proc_macro_derive(Hello)]
-pub fn hello_world(input: TokenStream) -> TokenStream
+### Crates and procedural macros
+
+All procedural macros in Rust are compiled as a crate. Procedural macro crates
+are defined with Cargo via the `proc-macro` key in your manfiest:
+
+```toml
+[lib]
+proc-macro = true
 ```
 
-Finally, procedural macros must be in their own crate, with the `proc-macro`
-crate type.
+Procedural macros are always compiled with the same target as the compiler
+itself. For example if you execute `cargo build --target
+arm-unknown-linux-gnueabi` then procedural macros will not be compiled for ARM,
+but rather your build computer (for example `x86_64-unknown-linux-gnu`).
 
-[derive]: attributes.html#derive
+Procedural macro crates are not currently allowed to export any items except
+procedural macros (we'll see how to define these in a bit). For example this
+crate is not allowed:
+
+```rust
+pub fn foo() {}
+```
+
+because the `foo` function is not a procedural macro. Procedural macros are
+loaded dynamically by the compiler when they are needed during compilation.
+Cargo will naturally make procedural macro crates available to crates which
+depend on them, or you can use the `--extern` argument.
+
+### The `proc_macro` crate
+
+Procedural macro crates almost always will link to the in-tree `proc_macro`
+crate. The `proc_macro` crate is a compiler-provided crate which provides
+facilities to working with the types of each procedural macro function. You can
+learn more about this crate by exploring [the documentation][pm-dox].
+
+[pm-dox]: https://doc.rust-lang.org/stable/proc_macro/
+
+Linking to the `proc_macro` crate can currently be done with:
+
+```rust,ignore
+extern crate proc_macro;
+```
+
+In the 2018 edition, however, this statement will not be necessary.
+
+### The `TokenStream` Type
+
+One aspect you may notice about the `proc_macro` crate is that it doesn't
+contain any AST items! Instead, it primarily contains a `TokenStream` type.
+Procedural macros in Rust operate over *token streams* instead of AST nodes,
+which is a far more stable interface over time for both the compiler and for
+procedural macros to target.
+
+A *token stream* is roughly equivalent to `Vec<TokenTree>` where a `TokenTree`
+can roughly be thought of as lexical token. For example `foo` is an `Ident`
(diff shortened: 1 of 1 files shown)


alexcrichton:
cc @dtolnay

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+
+### The `TokenStream` Type
+
+One aspect you may notice about the `proc_macro` crate is that it doesn't
+contain any AST items! Instead, it primarily contains a `TokenStream` type.
+Procedural macros in Rust operate over *token streams* instead of AST nodes,
+which is a far more stable interface over time for both the compiler and for
+procedural macros to target.
+
+A *token stream* is roughly equivalent to `Vec<TokenTree>` where a `TokenTree`
```
dtolnay: Somewhere it would be good to mention the key difference between `TokenStream` and `Vec<TokenTree>` -- that `TokenStream` is cheap to clone. Though maybe this belongs in rustdoc of the `TokenStream` type.

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+flavor of procedural macro is like writing `macro_rules!` only you get to
+execute arbitrary code over the input tokens instead of being limited to
+`macro_rules!` syntax.
+
+Procedural bang macros are defined with the `#[proc_macro]` attribute and have
+the following signature:
+
+```rust,ignore
+#[proc_macro]
+pub fn foo(x: TokenStream) -> TokenStream {
```
dtolnay: Instead of `x` I would call it `input`.

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+`Serialize` and `Deserialize`. The `syn` crate also has a [number of
+examples][synex] of defining custom derives.
+
+[procedural macros]: ../book/first-edition/procedural-macros.html
+[synex]: https://github.com/dtolnay/syn/tree/master/examples
+
+#### Custom attributes on Custom derive
+
+An additional feature of custom derive macros is that they can whitelist names
+of attributes which are considered not part of normal attribute macro expansion.
```
dtolnay: We should pick a clear name for such attributes to distinguish them from attributes that are procedural macros. This is one of the most confusing aspects of the procedural macro API (practically everybody believes `#[serde(...)]` is itself a macro).

I have been using the terms "attribute macro" and "inert attribute" to distinguish the two.

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+Procedural macros currently come in one of three flavors:
 
-Procedural macros involve a few different parts of the language and its
-standard libraries. First is the `proc_macro` crate, included with Rust,
-that defines an interface for building a procedural macro. The
-`#[proc_macro_derive(Foo)]` attribute is used to mark the deriving
-function. This function must have the type signature:
+* Custom macros - `my_macro!(...)`
+* Custom derive - `#[derive(MyTrait)]`
+* Custom attributes - `#[my_attribute]`
```
dtolnay: Let's kill the word "custom". It means nothing. The word should not appear on this page. We don't call things "custom crates" and "custom functions" and "custom structs".

I would write:

- Function-like macros / bang macros
- Derive macros
- Attribute macros

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+
+### Custom attributes
+
+The third and final form of procedural macros is custom attributes. Custom
+attributes allow you to define a custom `#[attr]`-style attribute which can be
+attached to items and generate wrappers and such. These macros are defined like:
+
+```rust,ignore
+#[proc_macro_attribute]
+pub fn foo(attr: TokenStream, item: TokenStream) -> TokenStream {
```
dtolnay: The second argument is not necessarily an item if this attribute is on a statement or expression. I would name the two `args` and `input`. Their role is fully analogous to command-line arguments and command-line input such as in:

```bash
$ input | ./foo a b c
```

```rust
#[foo(a, b, c)]
struct Input;
```

where `a b c` are arguments and separately there is input (the "in" in "stdin").

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+
+* Procedural bang macros can only be invoked in *item* contexts. For example you
+  can't define your own `format!` yet because that's only ever invoked in an
+  expression context. Put another way, procedural bang macros can only expand to
+  items (things like functions, impls, etc), not expressions.
+
+  This is primarily due to the lack of hygiene with procedural macros. Once a
+  better hygiene story is stabilized this restriction will likely be lifted.
+
+* Procedural macros cannot expand to definitions of `macro_rules!` macros (none
```
dtolnay: Hmm, serde_derive definitely does this now and it's stable. Maybe I misunderstand what this bullet means. Our two derives each include a macro_rules helper macro in the generated code that is called from elsewhere in the generated code.

Havvy:
I wish I knew you were working on this. I too have been working on it since Rustconf ended.

src/procedural-macros.md
Havvy: SGTM.

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
-Procedural macros can be used to implement custom [derive] on your own
-types. See [the book][procedural macros] for a tutorial.
+Procedural macros currently come in one of three flavors:
 
-Procedural macros involve a few different parts of the language and its
-standard libraries. First is the `proc_macro` crate, included with Rust,
-that defines an interface for building a procedural macro. The
-`#[proc_macro_derive(Foo)]` attribute is used to mark the deriving
-function. This function must have the type signature:
+* Custom macros - `my_macro!(...)`
```
Havvy: I thought function-like macros weren't being stabilized.

src/procedural-macros.md
```diff
@@ -1,24 +1,546 @@
...
+* Custom derive - `#[derive(MyTrait)]`
+* Custom attributes - `#[my_attribute]`
+
+Procedural macros allow you to run code at compile time that operates over Rust
+syntax, both consuming and producing Rust syntax. You can sort of think of
+procedural macros as Rust functions from an AST to another AST.
+
+### Crates and procedural macros
+
+All procedural macros in Rust are compiled as a crate. Procedural macro crates
```
Havvy: `in Rust` is never necessary. Everything about the reference is `in Rust` unless otherwise specified.

src/procedural-macros.md
petrochenkov: I recently had to come up with a name for this in the compiler and used "derive helper attribute".

"Inert attribute", if it means non-macro attribute, an attribute that doesn't expand the item it's applied to, can currently refer to a built-in attribute (`#[inline]`), a tool attribute (`#[rustfmt::skip]`), a "derive helper" (`#[serde(...)]`), or a "custom" attribute on nightly (`feature(custom_attribute)` + `#[my_attr]`).

src/procedural-macros.md
petrochenkov: The restriction was implemented much later than derives were stabilized (https://github.com/rust-lang/rust/pull/50820), so it was applied to fn-like macros and attribute macros to avoid breakage.

So macro_rules generated from derives do not produce immediate errors, but they can still be buggy and perhaps go through some breakage later with hygiene changes.

alexcrichton:
Updated!

alexcrichton:
Sorry @Havvy I've been really busy recently and haven't had a chance to come back to your comments here, but it looks like you've definitely got things in hand with https://github.com/rust-lang-nursery/reference/pull/412, thanks for picking this up!

PR #411, opened 2018-09-01
Added bool type section

### LINKED ISSUE #238: Add a section on `bool` in the types chapter
There's not a lot to say here, but there should be a heading in this section.

JoshBrudnak:
I added a section about the `bool` type in  [types.md](https://github.com/rust-lang-nursery/reference/blob/master/src/types.md). This resolves #238.

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -13,7 +13,7 @@ Some types are defined by the language, rather than as part of the standard
 library, these are called _primitive types_. Some of these are individual
 types:
 
-* The boolean type `bool` with values `true` and `false`.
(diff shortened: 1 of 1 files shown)


src/types.md
```diff
@@ -44,16 +45,34 @@ language:
 
 ## Numeric types
 
+### Boolean type
```
Havvy: Boolean isn't a numeric type, so it shouldn't be inside the "Numeric types" section. Move it, and it'll be good.

Havvy:
💟 Thanks!

PR #412, opened 2018-09-02
Expand Procedural Macro Docs

Havvy:
This merges my WIP stuff, @alexcrichton's mostly guide level stuff, and some new material into what I hope will be the final status for the procedural macro page.

I'm not super happy with the `attribute resolution` section personally, but I feel like we can fix that up later. I want procedural macros to be in the reference before the next stable release, since I feel people are going to have questions about it and the reference should be a good place to find answers.

Opening a new pull request instead of pushing to @alexcrichton's because so much has changed that conversation already there is entirely outdated.


src/procedural-macros.md
```diff
@@ -1,24 +1,288 @@
...
+
+These macros are only invokable in [modules]. They cannot even be invoked to
+make [item declaration statements]. Furthermore, they must either be invoked
+with curly braces and no semicolon or a different delimiter followed by a
+semicolon. For example, `make_answer` from the previous example can be invoked
+as `make_answer!{}`, `make_answer!();` or `make_answer![];`.
+
+These macros cannot expand to syntax that defines new `macro_rule` style macros.
+
+### Derive mode macros
```
petrochenkov: What's the point of "mode" in "derive mode"?
It's like a random unrelated word just inserted there.
I've seen the wording "derive mode" in a couple of compiler errors maybe, but that's all.

The phrasing "define new modes for the `derive` attribute" doesn't make sense to me.
Ok, I'm not a native speaker, let's look into [a dictionary](https://en.oxforddictionaries.com/definition/mode) - "A way or manner" / "A fashion or style" - still doesn't make any sense.
In the first approximation, `derive` just applies arbitrary macros passed to it (`#[derive(A, B, C)] ITEM` -> `#[A] #[B] #[C] ITEM`), how are those macros "modes"?

\</rant>

src/attributes.md
```diff
@@ -82,6 +81,47 @@ fn some_unused_variables() {
...
+
+An attribute is either dynamic or inert. During attribute processing, *dynamic
+attributes* remove themselves from the thing they are on while *inert attriutes*
+stay on.
+
+The `cfg` and `cfg_attr` attributes are dynamic. The `test` attribute is inert
+when compiling for tests and dynamic otherwise. Attribute macros are dynamic.
+All other attributes are inert. 
+
+## Attribute resolution
```
petrochenkov: I'd prefer to avoid specifying any details here for now, the rules are not settled yet.
Macro modularization was stabilized, but there's still a number of incomplete items from https://github.com/rust-lang/rust/pull/50911#issuecomment-411605393 that need to be done until beta, stable or maybe even later if crater allows.
(My suggestion would be to stash this section somewhere until 1.30 stable is released.)

src/attributes.md
```diff
@@ -82,6 +81,47 @@ fn some_unused_variables() {
...
+* Macro attributes
+* Derive mode helper attributes
+
+## Dynamic and inert attributes
+
+An attribute is either dynamic or inert. During attribute processing, *dynamic
+attributes* remove themselves from the thing they are on while *inert attriutes*
+stay on.
+
+The `cfg` and `cfg_attr` attributes are dynamic. The `test` attribute is inert
```
petrochenkov: The full list is `cfg`, `cfg_attr`, `test`, `bench`, `derive` currently, if I'm not missing anything.

`test` and `bench` are being migrated to macros right now though (https://github.com/rust-lang/rust/pull/53410).
(Also, `test`/`bench` are always "dynamic", they tweak their target items in test mode as well - add `pub`s, reexports.)

src/procedural-macros.md
Havvy: The main point, as I see it, is that it lets us talk about the `A`, `B`, `C` inside `#[derive(A, B, C)]` without needing to always call them macros. If we want to talk about `A` without caring that its a macro, and we just call them derive macros, then what do we call it? We could call it a "derive", but that does not jive well. "The derive attribute takes derives" is also weird. So we need some name to distinguish them. My original thought was to go with "deriver", but then I saw said compiler error that calls them a derive mode. And mode makes sense to me.

When I think of modes, I think of distinct ways of operating something. My oven has modes for "bake", "boil", some other one I never use, and "clean". Firefox has safe mode and private browsing mode. Businesses operate in stealth mode or growth mode or whatever fancy mode they want. That said, if you want to get it changed to "deriver" or something, file an issue against rust-lang/rust. I defer terminology to what it uses when I can.

petrochenkov: >I saw said compiler error that calls them a derive mode.

Yeah, I've seen those too and always thought they were bad, and wanted to get rid of them.
That's why when I'd seen that terminology reused I immediately thought it wasn't a good idea.
<sub><sup>(Please, excuse my past tenses if I got them all wrong.)</sup></sub>

I suspect the terminology came from the time when only a limited set of built-in derives existed.
"Mode" kinda implies that you have several of them and need to choose one, i.e. something `enum`-like.

>without needing to always call them macros

Why? They are macros.
If `#[proc_macro_attribute]` defines an "attribute macro", `#[proc_macro_derive]` defines a "derive macro", then we have a symmetry between language and docs.
"`derive` attribute takes derive macros as arguments" - sounds ok to me.

src/procedural-macros.md
```diff
@@ -1,24 +1,288 @@
...
+
+*Derive mode macros* define new modes for the `derive` attribute. These macros
+define new items given the token stream of a [struct], [enum], or [union]. They
+also define derive mode helper attributes.
+
+Custom deriver modes are defined by a [public] [function] with the
+`proc_maco_derive` attribute and a signature of `(TokenStream) -> TokenStream`.
```
michaelwu: `proc_macro_derive`

src/procedural-macros.md
```diff
@@ -1,24 +1,288 @@
...
+as possible given this limitation. This often includes using absolute paths to
+items in libraries (for example, `::std::option::Option` instead of `Option`) or
+by ensuring that generated functions have names that are unlikely to clash with
+other functions (like `__internal_foo` instead of `foo`).
+
+### Function-like procedural macros
+
+Function-like procedural macros define new invokable macros.
+
+These macros are defined by a [public] [function] with the `proc_maco`
```
michaelwu: `proc_macro`

src/linkage.md
```diff
@@ -205,3 +209,7 @@ a statically linked binary on MSVC you would execute:
 ```ignore,notrust
 RUSTFLAGS='-C target-feature=+crt-static' cargo build --target x86_64-pc-windows-msvc
 ```
+
+
```
steveklabnik: ✂️

Havvy:
I would like to see this section in the reference before the deadline. People are going to want to look this up.

PR #414, opened 2018-09-09
Document desugarings that happen when creating hir

### LINKED ISSUE #46: Document desugaring equivalence for `for` and `if let` and `while let`
Since any Rust impl is going to need to be semantically equivalent to the desugaring that rustc does, we should document what the desugarings are in the reference.

The `for` loop one is documented already in the [std lib](https://doc.rust-lang.org/std/iter/index.html#for-loops-and-intoiterator).

### LINKED ISSUE #47: `if let` semantically equivalence to `if` is wrong w.r.t. scope
The equivalence is wrong because the scope of the expression in the predicate has to live as long as the consequent because of the ability to `ref` in the pattern.

Or something like that.

matthewjasper:
Closes  #46
Closes #47

### DIFF
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -85,6 +85,26 @@ while let Some(y) = x.pop() {
 }
 ```
 
+A `while let` loop is equivalent to a `loop` expression containing a `match`
+expression as follows.
+
+```rust,ignore
+'label: while let PAT = EXPR {
+    /* loop body */
+}
+```
+
+is equivalent to
+
+```rust,ignore
+'label: loop {
+    match EXPR {
+        PAT => { /* loop body */ },
+        _ => break,
+    }
+}
+```
+
 ## Iterator loops
 
 > **<sup>Syntax</sup>**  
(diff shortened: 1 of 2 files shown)


src/expressions/loop-expr.md
```diff
@@ -118,6 +138,37 @@ for n in 1..11 {
...
+'label: for PATTERN in iter_expr {
+    /* loop body */
+}
+```
+
+is equivalent to
+
+```rust
+{
+    let result = match IntoIterator::into_iter(iter_expr) {
```
alercah: IIRC there should be some note here about how `into_iter` is looked up? It's *not* a lang item.

src/expressions/loop-expr.md
```diff
@@ -118,6 +138,37 @@ for n in 1..11 {
...
+        mut iter => 'label: loop {
+            let next;
```
alercah: Could we simplify this by `let PAT = match ...`?

matthewjasper: I can't work out how `let mut next = ...` would be different, but `let PAT = match ` would have a different drop order when `PAT` is something like `(x, _)`.

matthewjasper: [This](https://play.rust-lang.org/?gist=aa45f97595faf63a6e6a48e6adfefb83&version=stable&mode=debug&edition=2015) is why it can't be simplified.

matthewjasper:
Comments addressed

PR #415, opened 2018-09-12
Document preludes

Havvy:
I set out to document `no_std` and ended up documenting preludes in general. This also led to the removal of the "Module-only attributes" section as one of its attributes was placed in the prelude attributes section while the other attribute is already documented pretty well on the modules page, so I just linked to it from the attributes page. I wanted to get rid of that section anyways. This means the only two sections in the attribute sections that need to be removed are "crate only attributes" and "misc. attributes". Though having all the sections subsumed under misc attributes was a mistake, and they need to be brought a level higher. I'll do that after this lands, while also sorting them alphabetically.

I am generally happy with the new prose I've written in this PR. I specifically added why `no_std` is useful because I see it as a frequently asked question.

### DIFF
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -85,16 +85,16 @@ fn main() {
 
 ## Preludes and `no_std`
 
-All crates have a *prelude* that automatically inserts a [use declaration] into
-each [module] and an [`extern crate]` into the crate root module. By default,
-the *standard prelude* is used. The linked crate is [`std`] and the [use
-delcaration] uses [`std::prelude::v1::*`].
+All crates have a *prelude* that automatically inserts paths of a specific
+module, the *prelude module*, into scope of each [module] and an [`extern
+crate]` into the crate root module. By default, the *standard prelude* is used.
+The linked crate is [`std`] and the prelude module is [`std::prelude::v1`].
 
 The prelude can be changed to the *core prelude* by using the `no_std`
 [attribute] on the root crate module. The linked crate is [`core`] and the
-[`use` declaration] uses [`core::prelude::v1::*`]. Using the core prelude over
-the standard prelude is useful when either the crate is targeting a platform
-that does not support the standard library or is purposefully not using the
+prelude module is [`core::prelude::v1`]. Using the core prelude over the
+standard prelude is useful when either the crate is targeting a platform that
+does not support the standard library or is purposefully not using the
 capabilities of the standard library. Those capabilities are mainly dynamic
 memory allocation (e.g. `Box` and `Vec`) and file and network capabilities (e.g.
 `std::fs` and `std::io`).
@@ -139,9 +139,9 @@ type must  be one of the following:
 [_utf8 byte order mark_]: https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8
 [`Termination`]: ../std/process/trait.Termination.html
 [`core`]: ../core/index.html
-[`core::prelude::v1::*`]: ../core/preludce.index.html
+[`core::prelude::v1`]: ../core/preludce.index.html
 [`std`]: ../std/index.html
-[`std::prelude::v1::*`]: ../std/prelude/index.html
+[`std::prelude::v1`]: ../std/prelude/index.html
 [`use` declaration]: items/use-declarations.html
 [attribute]: attributes.html
 [attributes]: attributes.html
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -20,9 +20,8 @@
 Rust's semantics obey a *phase distinction* between compile-time and
 run-time.[^phase-distinction] Semantic rules that have a *static
 interpretation* govern the success or failure of compilation, while
-semantic rules
-that have a *dynamic interpretation* govern the behavior of the program at
-run-time.
+semantic rules that have a *dynamic interpretation* govern the behavior of the
+program at run-time.
 
 The compilation model centers on artifacts called _crates_. Each compilation
 processes a single crate in source form, and if successful, produces a single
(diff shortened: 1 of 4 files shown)


src/crates-and-source-files.md
```diff
@@ -100,6 +83,48 @@ fn main() {
 }
 ```
 
+## Preludes and `no_std`
+
+All crates have a *prelude* that automatically inserts a [use declaration] into
+each [module] and an [`extern crate]` into the crate root module. By default,
```
petrochenkov: This is not how it works and there are observable differences.
(Will explain later, need to go now.)

Havvy: I'm still blocked on this explanation.

petrochenkov: Sorry, I completely forgot about this.

petrochenkov: Prelude names are not placed in any modules with `use` declarations, they are just added into scope.
So you can write `Vec<T>` in any module, but not `self::Vec<T>` - ther's no `Vec` in any module.

petrochenkov: If we look at some other names that are placed into scope rather than into modules
```rust
{
    let x = 10;
    {
        let x = 11;
        {
            let x = 12;
            {
                x
            }
        }
    }
}
```
then we can extend this hierarchy with one more outer level and place prelude names there, that's the effect of std prelude injection.

A number of other names behaves in the same way, e.g. built-in types, or macros imported with `#[macro_use] extern crate ...`, or extern crates with `feature(extern_prelude)` (that's currently being stabilized in https://github.com/rust-lang/rust/pull/54404),

Havvy:
The last two weeks of work have been hard on me, but I finally got the energy to properly update this today. @petrochenkov, does this accurately describe the prelude now?

Paths and path resolution need a new look at in the future, so I'm not too worried with the writing about "in scope" or whatnot being technically specific at this moment.

src/crates-and-source-files.md
```diff
@@ -85,16 +85,16 @@ fn main() {
 
 ## Preludes and `no_std`
 
-All crates have a *prelude* that automatically inserts a [use declaration] into
-each [module] and an [`extern crate]` into the crate root module. By default,
-the *standard prelude* is used. The linked crate is [`std`] and the [use
-delcaration] uses [`std::prelude::v1::*`].
+All crates have a *prelude* that automatically inserts paths of a specific
```
petrochenkov: "of" -> "from" would be better?

src/items/modules.md
```diff
@@ -68,8 +68,12 @@ mod thread {
 }
 ```
 
-Modules implicitly have a [use declaration] specified by crate's [prelude]. The
-[use declaration] can be removed by using the `no_implicit_prelude` [attribute].
+Modules implicitly have some paths in scope. These paths are to built-in types,
```
petrochenkov: "paths" -> "names", names are in scope.

petrochenkov: Path is a higher level thing in general, the first segment is searched in the current scope, the following segments are searched in modules (or in impls if they are associated items). Path as a whole being in scope doesn't make much sense.

Havvy:
Both changes have been made.

Havvy:
Rebased against master.

r? @matthewjasper or @alercah

src/crates-and-source-files.md
```diff
@@ -100,6 +83,48 @@ fn main() {
 }
 ```
 
+## Preludes and `no_std`
+
+All crates have a *prelude* that automatically inserts names from a specific
+module, the *prelude module*, into scope of each [module] and an [`extern
+crate]` into the crate root module. By default, the *standard prelude* is used.
+The linked crate is [`std`] and the prelude module is [`std::prelude::v1`].
```
ehuss: I had a question.  By default, aren't both `std` and `core` crates included in the extern prelude?  I also think `std` is in the prelude for `no_std`, but that might change soon: https://github.com/rust-lang/rust/pull/54658

Also, would it be helpful to describe the different preludes and scope layers?  It may not be super relevant, but it might help clarify precedence when there are name conflicts (additions to the standard prelude will not shadow extern crates).  IIUC, the layers are:
1. local variables
2. items in unnamed blocks
3. items in the current module
4. crate names in the **"extern prelude"**
5. standard library prelude
6. language prelude (built-in types)

(The reason I ask is that I was looking at updating `extern crate` for the edition, and I'm trying to wrap my head around it.)

src/items/functions.md
```diff
@@ -157,8 +157,8 @@ fn test_only() {
 The attributes that have meaning on a function are [`cfg`], [`deprecated`],
 [`doc`], `export_name`, `link_section`, `no_mangle`, [the lint check
 attributes], [`must_use`], [the procedural macro attributes], [the testing
-attributes], and [the optimization hint
-attributes].
+attributes], and [the optimization hint attributes]. Functions also accept
+macro attributes.
```
matthewjasper: attribute macros?

src/crates-and-source-files.md
```diff
@@ -20,9 +20,8 @@
 Rust's semantics obey a *phase distinction* between compile-time and
 run-time.[^phase-distinction] Semantic rules that have a *static
 interpretation* govern the success or failure of compilation, while
-semantic rules
-that have a *dynamic interpretation* govern the behavior of the program at
-run-time.
+semantic rules that have a *dynamic interpretation* govern the behavior of the
+program at run-time.
```
matthewjasper: Unrelated, but does this distinction ever appear again?

src/crates-and-source-files.md
Havvy: Yes, all that would be helpful. I didn't know what all the layers were, but if this is the correct list, it's definitely something that should be included.

For `core` being included with `std` in the extern prelude, I think that's all supposed to change with https://github.com/rust-lang/rust/issues/53128 but I'm not entirely sure.

src/crates-and-source-files.md
Havvy: Const functions perhaps? It's a paragraph that's survived since the reference was originally the guide.

matthewjasper:
Thanks, I think I was just waiting to see if you wanted to address @ehuss' comment

Issue #301, opened 2018-04-03
Document RFC 0179 (and-mut-patterns)


alercah:
[RFC Link](https://github.com/rust-lang/rfcs/tree/master/text/0179-and-mut-patterns.md)

Issue #300, opened 2018-04-03
Document RFC 1492 (dotdot-in-patterns)


alercah:
[RFC link](https://github.com/rust-lang/rfcs/tree/master/text/1492-dotdot-in-patterns.md)

Issue #280, opened 2018-03-26
`match` ergonomics (RFC 2005)


matthewjasper:
Looking to stabilise in 1.27

- [ ] Add to match documentation
- [ ] Use in examples?

PR #419, opened 2018-09-16
Patterns chapter

### LINKED ISSUE #83: Patterns aren't documented
As a guide, it may have these sections:

- [x] Patterns (introduction)
- [x] Wildcard pattern `_`
- [ ] Reference patterns `& ...`
- [ ] Tuple patterns `(a, b, ...)`
- [x] Slice patterns `[a, b, ..i, y, z]]`
- [ ] Identifier patterns `binding @ subpattern`
- [x] Literal patterns `1`, `2`, etc
- [x] Range patterns `0..10`
- [ ] Struct patterns `Variant {a, b, ...}`
- [ ] Tuple struct patterns `Variant (a, b, ...)`
- [ ] Box pattern
- [x] Refutability
- [x] Destructuring

ehuss:
This is a repost of #284, rebased, updated, most review comments addressed, and some additional information added.

Closes #83, #300, #301.

This adds a small description of binding modes which covers #280, but I'm not confident it is detailed enough to consider it complete.  There are a lot of details not covered in depth (like const references).

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -63,6 +63,8 @@
         - [Match expressions](expressions/match-expr.md)
         - [Return expressions](expressions/return-expr.md)
 
+- [Patterns](patterns.md)
+
 - [Type system](type-system.md)
     - [Types](types.md)
     - [Dynamically Sized Types](dynamically-sized-types.md)
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -81,8 +81,6 @@ Please notice that the `2..=9` is a [Range Pattern], not a [Range Expression]
 and, thus, only those types of ranges supported by range patterns can be used
 in match arms.
 
-A range pattern may not be a sub-range of another range pattern inside the same `match`.
-
 Match arms can accept _match guards_ to further refine the
 criteria for matching a case. Pattern guards appear after the pattern and
 consist of a bool-typed expression following the `if` keyword. A pattern guard
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -18,18 +18,15 @@
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> _MatchArmPatterns_ _MatchArmGuard_<sup>?</sup>
 >
 > _MatchArmPatterns_ :\
-> &nbsp;&nbsp; `|`<sup>?</sup> _Pattern_ ( `|` _Pattern_ )<sup>\*</sup>
+> &nbsp;&nbsp; `|`<sup>?</sup> [_Pattern_] ( `|` [_Pattern_] )<sup>\*</sup>
 >
 > _MatchArmGuard_ :\
 > &nbsp;&nbsp; `if` [_Expression_]
 
 A *`match` expression* branches on a pattern. The exact form of matching that
-occurs depends on the pattern. *Patterns* consist of some combination of
-literals, destructured arrays or enum constructors, structs and tuples,
-variable binding specifications, wildcards (`..`), and placeholders (`_`). A
-`match` expression has a *head expression*, which is the value to compare to
-the patterns. The type of the patterns must equal the type of the head
-expression.
+occurs depends on the pattern. See [Patterns] for more details. A `match`
+expression has a *head expression*, which is the value to compare to the
+patterns. The type of the patterns must equal the type of the head expression.
 
 A `match` behaves differently depending on whether or not the head expression
 is a [place expression or value expression][place expression].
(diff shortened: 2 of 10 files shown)


src/SUMMARY.md
```diff
@@ -63,6 +63,8 @@
         - [Match expressions](expressions/match-expr.md)
         - [Return expressions](expressions/return-expr.md)
 
+- [Patterns](patterns.md)
```
ehuss: Let me know if you want to move the patterns chapter elsewhere (such as underneath `Statements and Expressions`).

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+}
+```
+
+## Identifier patterns
+
+> **<sup>Syntax</sup>**\
+> _IdentifierPattern_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `ref`<sup>?</sup> `mut`<sup>?</sup> [IDENTIFIER] (`@` [_Pattern_] ) <sup>?</sup>
+
+_Identifier patterns_ bind the value they match to a **previously undeclared** variable.
```
ehuss: The phrase "previously undeclared" here is confusing to me. I'm not certain if the original intent is to emphasize that the identifier must be unique within a pattern (i.e. `(x, x)` is an invalid pattern), or if it is trying to express something else. The identifier can shadow bindings of the same name from outside the pattern, so that's why I'm confused by the "previously undeclared" part. What do you think of changing it to something like this?

>    _Identifier patterns_ bind the value they match to a variable. The identifier must be unique within the pattern. The variable will shadow any variables of the same name in scope.

src/SUMMARY.md
Havvy: I want to eventually to split all the "X and Y" chapters eventually, so having this top-level is good to me.

Havvy: To expand on this, I'd like to see the order be "Items, Attributes, Expressions, Patterns, Statements", so the current location LGTM as well.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+> &nbsp;&nbsp; | [_ReferencePattern_]\
+> &nbsp;&nbsp; | [_StructPattern_]\
+> &nbsp;&nbsp; | [_TupleStructPattern_]\
+> &nbsp;&nbsp; | [_TuplePattern_]\
+> &nbsp;&nbsp; | [_SlicePattern_]\
+> &nbsp;&nbsp; | [_PathPattern_]
+
+Patterns in Rust are used to match values against structures and to,
+optionally, bind variables to values inside these structures. They are also
+used in variable declarations and function/closure parameters, though in these
```
Havvy: General style right now is to not use `x/y`. Could rewrite as "parameters for functions and closures".

Havvy: The clause after the comma could be its own sentence, or maybe even removed. If not removed, "though in these cases most of the time" -> "though in most of these cases,".

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+Patterns are used in:
+
+* [`let` declarations](statements.html#let-statements)
+* [Function](items/functions.html) and [closure](expressions/closure-expr.html)
+  parameters
+* [`match` expressions](expressions/match-expr.html)
+* [`if let` expressions](expressions/if-expr.html)
+* [`while let` expressions](expressions/loop-expr.html#predicate-pattern-loops)
+* [`for` expressions](expressions/loop-expr.html#iterator-loops)
+* Inside other patterns
```
Havvy: This is a weird way of saying that patterns are recursive. I wouldn't really call a subpattern a pattern on its own right personally.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+
+## Destructuring
+
+Patterns can be used to *destructure* structs, enums, and tuples. Destructuring
```
Havvy: Not required, but would be nice to make "structs", "enums", and "tuples" be links.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+breaks a value up into its component pieces. The syntax used is almost the same as
```
Havvy: "breaks up a value"?

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+when creating such values. When destructing a data structure with named (but
```
Havvy: "destructing" -> "destructuring"

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+not numbered) fields, it is allowed to write `fieldname` as a shorthand for
+`fieldname: fieldname`. In a pattern whose head expression has a `struct`,
```
Havvy: I feel the sentence about `_` and `..` should precede the one about field (de)structuring shorthand. Also "stands" -> "stands in"

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+    Message::Move{ .. } => println!("other move"),
+    Message::ChangeColor { 0: red, 1: green, 2: _ } => {
+        println!("color change, red: {}, green: {}", red, green);
+    }
+};
+```
+
+## Refutability
+
+A pattern is said to be *Refutable* when it **has the possibility of not being matched**
```
Havvy: "Refutable" -> "refutable" (casing). I'm iffy about bolding the definition part. especially since the irrefutable definition part isn't bolded. If you want to propose changing our style to actually bold them like that, file an issue.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+[BYTE_STRING_LITERAL]: tokens.html#byte-string-literals
+[RAW_BYTE_STRING_LITERAL]: tokens.html#raw-byte-string-literals
+[INTEGER_LITERAL]: tokens.html#integer-literals
+[FLOAT_LITERAL]: tokens.html#floating-point-literals
+
+_Literal patterns_ match exactly the value they represent. Since negative numbers are
+not literals in Rust, literal patterns also accept an optional minus sign before the
+literal.
+
+Floating-point literals are currently accepted, but due to the complexity of comparing
```
Havvy: We generally don't talk about the future of Rust, but this seems a worthwhile exception. But it should be placed in a warning box (you can find the syntax for it in introduction.md)

src/patterns.md
Havvy: SGTM. I would also add that the scope of the variable is dependent upon the context of where the pattern is used.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+
+```rust
+let mut variable = 10;
+fn sum(x: i32, y: i32) -> i32 {
+#    x + y
+# }
+```
+
+match any value and bind it to that identifier. This is the most commonly
+used pattern in variable declarations and function/closure parameters.
```
Havvy: Same "x/y" as earlier.

src/patterns.md
```diff
@@ -0,0 +1,646 @@
...
+match x {
+    e @ 1 ..= 5 => println!("got a range element {}", e),
+    _ => println!("anything"),
+}
+```
+
+binds to `e` the value 2 (not the entire range: the range here is a range subpattern).
+
+By default, identifier patterns bind a variable to a copy of or move from the
+matched value (depending whether the matched value implements the [Copy trait]).
```
Havvy: * The parenthesis are superfluous. The words inside are crucial to the meaning of the sentence. 
* "depending" -> "depending on".
* `the [Copy trait]` -> ``[`Copy`]``

src/expressions/match-expr.md
```diff
@@ -65,62 +62,9 @@ match x {
 Patterns that bind variables default to binding to a copy or move of the
 matched value (depending on the matched value's type). This can be changed to
 bind to a reference by using the `ref` keyword, or to a mutable reference using
-`ref mut`.
+`ref mut`. See [Identifier Patterns].
```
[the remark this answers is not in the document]
Havvy: This entire section seems duped from the Identifier Patterns" section?

src/tokens.md
```diff
@@ -611,3 +611,8 @@ them are referred to as "token trees" in [macros].  The three types of brackets
 [extern]: items/external-blocks.html
 [struct expressions]: expressions/struct-expr.html
 [tuple index]: expressions/tuple-expr.html#tuple-indexing-expressions
+[Wildcard patterns]: patterns.html#wildcard-pattern
```
Havvy: Can you alphabetize these? You don't need to alphabetize all of the ones that are already here, but at least from `extern` downwards (since the three already at the end are alphabetized by coincidence). Uppercase sorts before all lowercase.

Havvy:
Overall this looks good. I'm not done nitpicking it, and you don't have to listen to my nitpicks (especially the ones with "?"s attached. Just respond to the nitpick saying so if you don't.

src/expressions/match-expr.md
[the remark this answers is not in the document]
ehuss: Yea, the paragraph is mostly a dupe, so I rewrote it to say something relevant.

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+
+Identifier patterns bind the value they match to a variable. The identifier
+must be unique within the pattern. The variable will shadow any variables of
+the same name in scope. The scope of the new binding depends on the context of
+where the pattern is used (such as a `let` binding or a `match` arm).
+
+Patterns that consist of only an identifier, possibly with a `mut`, like
+`variable`, `x`, and `y` below:
+
+```rust
```
Havvy: This example coming in between the definition makes it really hard to read the definition.

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+let x = 2;
+
+match x {
+    e @ 1 ..= 5 => println!("got a range element {}", e),
+    _ => println!("anything"),
+}
+```
+
+binds to `e` the value 2 (not the entire range: the range here is a range subpattern).
```
Havvy: This should be on the same line as the "For example".

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+let mut variable = 10;
+fn sum(x: i32, y: i32) -> i32 {
+#    x + y
+# }
+```
+
+match any value and bind it to that identifier. This is the most commonly
+used pattern in variable declarations and parameters for functions and closures.
+
+To bind non-trivial patterns to a variable, the use of the syntax `variable @
```
Havvy: The passive voice here is an inversion on what is normally written. I also don't like the word "non-trivial", but don't have a better word to replace it personally.

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+```rust,compile_fail
+# struct Person {
+#    name: String,
+#    age: u8,
+# }
+# let value = Person{ name: String::from("John"), age: 23 };
+if let Person{name: &person_name, age: 18..=150} = value { }
+```
+
+is not valid. What we must do is:
```
Havvy: "What we must do is:" -> "To make it valid, we can write the following."

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+#    c: bool,
+# }
+# let struct_value = Struct{a: 10, b: 'X', c: false};
+#
+let Struct{a: x, b: y, c: z} = struct_value;          // destructure all fields
+```
+
+A struct pattern is refutable when one of its subpatterns is refutable.
+
+## TupleStruct patterns
```
Havvy: Do we want "TupleStruct" to be one word or two?

I generally don't like smashing words together.

src/patterns.md
```diff
@@ -0,0 +1,656 @@
...
+TupleStruct patterns match tuple struct and enum values that match all criteria defined
+by its subpatterns. They are also used to [destructure](#destructuring) a tuple struct or
+enum value.
+
+A TupleStruct pattern is refutable when one of its subpatterns is refutable.
+
+## Tuple patterns
+
+> **<sup>Syntax</sup>**\
+> _TuplePattern_ :<a name="tuple-pattern-syntax"></a>\
```
Havvy: "name" attribute is deprecated for "id" in HTML, IIRC.

Havvy:
Alright, I've finished looking over all of the changes.

ehuss:
I rewrote all the situations where a sentence was split across the example code. In the very first example, I'm not sure if it is clearer if the code sample should appear before or after the text talking about it. It's fairly small, so it should fit on one screen on most devices, so it probably doesn't matter too much?

Havvy:
Just need to rebase against master (conflict with `tokens.md`) and this should be good to merge.

ehuss:
Rebased!

Havvy:
And merged! Many thanks for taking this over and getting it finished! (Imagine twenty part hat emojis here)

PR #420, opened 2018-09-17
No more anonymous trait fn parameters

mark-i-m:
cc https://github.com/rust-lang/rust/issues/41686

These were the only ones I could find from a quick search, but please let me know if there are others.

cc @Centril

### DIFF
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -99,7 +99,7 @@ Consider the following trait:
 # type Surface = i32;
 # type BoundingBox = i32;
 trait Shape {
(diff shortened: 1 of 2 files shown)


Centril:
We should probably also document the change itself in the reference, i.e. that in Rust 2015 it is like this, and in Rust 2018 it is like that.

LGTM otherwise -- I defer to you on coverage =)

mark-i-m:
@Havvy are you telling me "LGTM, except you should document the RFC" or "LGTM, please don't document the RFC"?

PR #424, opened 2018-09-22
Add some notes on  repetition

mark-i-m:
cc rust-lang/rust#51934

The new Kleene op is not stable yet, but the removal of `?` as a separator is. I can split this up into two PRs if you want.

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -51,17 +51,23 @@ designator is already known, and so only the name of a matched nonterminal comes
 after the dollar sign.
 
 In both the matcher and transcriber, the Kleene star-like operator indicates
-repetition. The Kleene star operator consists of `$` and parentheses, optionally
-followed by a separator token, followed by `*` or `+`. `*` means zero or more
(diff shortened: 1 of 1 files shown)


src/macros-by-example.md
```diff
@@ -59,6 +59,10 @@ matches, in a structure that mimics the structure of the repetition encountered
 on a successful match. The job of the transcriber is to sort that structure
 out.
 
+> **Edition Differences**: In the 2018 edition, a new Kleene operator was
```
Havvy: What `?` means should be defined outside of the edition differences section. And then the edition differences section should just state that it's not available in the 2015 edition.

mark-i-m:
@Havvy better now?

Havvy:
💟 Thanks!

Looking at the tracking issue, I *think* this is already stable in nightly Rust. If it's not, we can revert the two commits.

mark-i-m:
@Havvy Thanks!

But it is actually unstable still: https://play.rust-lang.org/?gist=af7cee08e596c9d7b61fa4f276134205&version=nightly&mode=debug&edition=2018

PR #428, opened 2018-09-23
Grammar: Add enum expression.

ehuss:


Havvy:
LGTM when awake as well.

💟 Thanks!

PR #429, opened 2018-09-23
Grammar: Add struct expression.

ehuss:
This includes some rework and expansion of the contents.

### DIFF
--- a/src/expressions/struct-expr.md
+++ b/src/expressions/struct-expr.md
@@ -1,26 +1,33 @@
 # Struct expressions
 
-There are several forms of struct expressions. A _struct expression_ consists
-of the [path] of a [struct item](items/structs.html), followed by a
-brace-enclosed list of zero or more comma-separated name-value pairs, providing
-the field values of a new instance of the struct. A field name can be any
-[identifier], and is separated from its value expression by a
-colon. In the case of a tuple struct the field names are numbers corresponding
-to the position of the field. The numbers must be written in decimal,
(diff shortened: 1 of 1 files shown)


src/expressions/struct-expr.md
```diff
@@ -74,9 +100,43 @@ Point3d { x: x, y: y_value, z: z };
 Point3d { x, y: y_value, z };
 ```
 
+## Tuple struct expression
+
+A struct expression with fields enclosed in parentheses constructs a tuple struct. Though
+it is listed here as a specific expression for completeness, it is equivalent to a [call
+expression] to the tuple struct's constructor. Example:
```
Havvy: "For example:" please. I'm trying to keep all examples with that exact phrase.

src/expressions/struct-expr.md
```diff
@@ -74,9 +100,43 @@ Point3d { x: x, y: y_value, z: z };
...
+Position(0, 0, 0);  // Typical way of creating a tuple struct.
+let c = Position;  // `c` is a function that takes 3 arguments.
+let pos = c(8, 6, 7);  // Creates a `Position` value.
+```
+
+## Unit struct expression
+
+A unit struct expression is just the path to a unit struct item. This refers to the unit
+struct's implicit constant of its value. The unit struct value can also be constructed
+with fieldless struct expression. Example:
```
Havvy: "with a"

Havvy:
💟 Thanks!

PR #430, opened 2018-09-23
Grammar: Functions, methods, impl, traits, extern.

ehuss:
These all are intertwined with the "function" definition. Updates grammar for:
- Functions and methods
- Extern blocks
- Implementation
- Traits

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -3,7 +3,7 @@
 > **<sup>Syntax</sup>**\
 > _Trait_ :\
 > &nbsp;&nbsp; `unsafe`<sup>?</sup> `trait` [IDENTIFIER]&nbsp;
->              [_GenericsDecl_]<sup>?</sup>
+>              [_Generics_]<sup>?</sup>
 >              [_WhereClause_]<sup>?</sup> `{`\
 > &nbsp;&nbsp;&nbsp;&nbsp; _TraitItem_<sup>\*</sup>\
 > &nbsp;&nbsp; `}`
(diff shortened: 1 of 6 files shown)


ehuss:
Question: Would it be clearer if `Method` was split into two (`Method | Function`) and make `Method` something that strictly has a self parameter?  I'm thinking it would be clearer.

Also, I missed the 2015/2018 difference of trait functions taking optional named parameters.  I'm thinking of documenting the 2015 behavior with a dagger (or something) that links to the section that describes that it is no longer allowed.  Thoughts?

src/items/traits.md
```diff
@@ -115,6 +135,23 @@ let circle = Box::new(circle) as Box<dyn Circle>;
 let nonsense = circle.radius() * circle.area();
 ```
 
+## Unsafe traits
+
+Traits that begin with the `unsafe` keyword indicate that *implementing* the
+trait may be [unsafe]. It is safe to use a correctly implemented unsafe trait.
+The [trait implementation] must also include the `unsafe` keyword.
```
matthewjasper: "include" should be "start with" or "begin with"

matthewjasper:
> Question: Would it be clearer if `Method` was split into two (`Method | Function`) and make `Method` something that strictly has a self parameter? I'm thinking it would be clearer.
> EDIT: Although it could be complicated in the future due to `arbitrary_self_types`.

I'm not sure how arbitrary self types changes anything here. It would probably be slightly better for methods to only include things with `self` parameters, since I think it is the more common meaning in Rust.

ehuss:
I attempted to explain the 2018 edition changes for trait parameter patterns.

matthewjasper:
Thanks!

PR #435, opened 2018-09-27
Add grouped patterns.

ehuss:
Stabilization PR: https://github.com/rust-lang/rust/pull/54497
Tracking issue: https://github.com/rust-lang/rust/issues/51087

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -10,6 +10,7 @@
 > &nbsp;&nbsp; | [_StructPattern_]\
 > &nbsp;&nbsp; | [_TupleStructPattern_]\
 > &nbsp;&nbsp; | [_TuplePattern_]\
+> &nbsp;&nbsp; | [_GroupedPattern_]\
 > &nbsp;&nbsp; | [_SlicePattern_]\
 > &nbsp;&nbsp; | [_PathPattern_]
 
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -582,6 +583,25 @@ They are also used to [destructure](#destructuring) a tuple.
...
 
+## Grouped patterns
+
+> **<sup>Syntax</sup>**\
+> _GroupedPattern_ :\
+> &nbsp;&nbsp; `(` [_Pattern_] `)`
+
+Enclosing a pattern in parentheses can be used to explicitly control the
+precedence of compound patterns. For example, a reference pattern next to a
+range pattern such as `&0..=5` is ambiguous and is not allowed, but can be
```
Havvy: It's not ambiguous, and it *is* allowed. It means `(&0)..=5`. I wonder if we need a precedence table for patterns. But that'd probably be best done in its own PR.

ehuss: I'm a little confused.  Latest nightly says it is ambiguous:

```
error: the range pattern here has ambiguous interpretation
  --> src/main.rs:28:10
   |
28 |         &0..=30 => { println!("small") }
   |          ^^^^^^ help: add parentheses to clarify the precedence: `(0 ..=30)`

```

I originally started writing a section on precedence, but I deleted it because I wasn't really clear what it could say (and I agree, it should probably be separate).  The only precedence rules I know of are this and paths vs identifiers (and that's not really a precedence but a shadowing rule).

PR #436, opened 2018-09-28
Better link to repr attribute from structs page

phansch:
Before it was linking to `attributes.html#ffi-attributes`, which links
to `type-layout.html#representations`. This saves a click for people
that want to go straight to the `repr` documentation.


Havvy:
Good catch! 💟 Thanks!

PR #437, opened 2018-09-30
Mention syntax ambiguity for eRFC 2947.

davidtwco:
Part of rust-lang/rust#53668.

cc @centril

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -44,7 +44,7 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
 >              [_BlockExpression_]\
 > &nbsp;&nbsp; (`else` (
 >   [_BlockExpression_]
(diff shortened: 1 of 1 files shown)


src/expressions/if-expr.md
```diff
@@ -111,6 +111,12 @@ match EXPR {
 }
 ```
 
+If the expression is a [lazy boolean operator expression][_LazyBooleanOperatorExpression_] then
+this is a error as it is ambiguous which is problematic for the implementation
+of if-let chains (see [eRFC 2947][_eRFCIfLetChain_]).
```
Centril: Perhaps add a note about wrapping in parenthesis as a solution?

src/expressions/if-expr.md
```diff
@@ -111,6 +111,27 @@ match EXPR {
 }
 ```
 
+If the expression is a [lazy boolean operator expression][_LazyBooleanOperatorExpression_] then
```
Havvy: This is a very roundabout way of saying "The expression cannot be a lazy boolean operator expression."

src/expressions/if-expr.md
```diff
@@ -111,6 +111,27 @@ match EXPR {
...
+this is a error as it is ambiguous which is problematic for the implementation
```
Havvy: I'd leave out the "which is problematic". Just state it's ambiguous with a planned future change in the language.

src/expressions/if-expr.md
davidtwco: Changed this.

src/expressions/if-expr.md
davidtwco: Changed this.

davidtwco:
Oops, let me remove the `fixup!` commit.

Centril:
Thanks!

PR #440, opened 2018-10-05
Document const functions

oli-obk:
tracking issue: https://github.com/rust-lang/rust/issues/53555

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -37,7 +37,8 @@ to be ran.
 * The [dereference operator].
 * [Grouped] expressions.
 * [Cast] expressions, except pointer to address and
-  function pointer to address casts.
+* function pointer to address casts.
+* function and method calls
 
 ## Const context
 
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -37,8 +37,8 @@ to be ran.
 * The [dereference operator].
 * [Grouped] expressions.
 * [Cast] expressions, except pointer to address and
-* function pointer to address casts.
-* function and method calls
+  function pointer to address casts.
+* calls of const functions and const methods
 
 ## Const context
 
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -1,7 +1,43 @@
 # Constant evaluation
 
 Constant evaluation is the process of computing the result of
-[expressions] during compilation.
+[expressions] during compilation. Only a subset of all expressions
+can be evaluated at compile-time.
+
+## Constant expressions
+
+Certain types of expressions can be evaluated at compile time. These are called
+_constant expressions_ and are required in const contexts. In
+other places, such as in [`let` statements](statements.html#let-statements),
+constant expressions may be evaluated at compile time. If errors, such as out
+of bounds [array indexing] or [overflow] occurs,
+then it is a compiler error if the value must be evaluated at compile time,
+otherwise it is just a warning, but the code will most likely panic when run.
+
+The following expressions are constant expressions, so long as any operands are
+also constant expressions and do not cause any [`Drop::drop`][destructors] calls
+to be ran.
+
+* [Literals].
+* [Paths] to [functions](items/functions.html) and constants.
+  Recursively defining constants is not allowed.
+* [Tuple expressions].
+* [Array expressions].
+* [Struct] expressions.
+* [Enum variant] expressions.
+* [Block expressions], including `unsafe` blocks, which only contain items and
+  possibly a constant tail expression.
+* [Field] expressions.
+* Index expressions, [array indexing] or [slice] with a `usize`.
+* [Range expressions].
+* [Closure expressions] which don't capture variables from the environment.
+* Built in [negation], [arithmetic, logical], [comparison] or [lazy boolean]
+  operators used on integer and floating point types, `bool` and `char`.
+* Shared [borrow]s, except if applied to a type with [interior mutability].
+* The [dereference operator].
+* [Grouped] expressions.
+* [Cast] expressions, except pointer to address and
+  function pointer to address casts.
 
 ## Const context
 
@@ -18,4 +54,28 @@ A _const context_ is one of the following:
 [enum discriminants]: items/enumerations.html#custom-discriminant-values-for-field-less-enumerations
 [constants]: items/constant-items.html
 [statics]: items/static-items.html
-[expressions]: expressions.html#constant-expressions
+[expressions]: expressions.html
+[array indexing]:       expressions/array-expr.html#array-and-slice-indexing-expressions
+[overflow]:             expressions/operator-expr.html#overflow
+[destructors]:          destructors.html
+[literals]:             expressions/literal-expr.html
+[paths]:                expressions/path-expr.html
+[tuple expressions]:    expressions/tuple-expr.html
+[array expressions]:    expressions/array-expr.html
+[struct]:               expressions/struct-expr.html
+[enum variant]:         expressions/enum-variant-expr.html
+[block expressions]:    expressions/block-expr.html
+[field]:                expressions/field-expr.html
+[array indexing]:       expressions/array-expr.html#array-and-slice-indexing-expressions
+[slice]:                types.html#array-and-slice-types
+[range expressions]:    expressions/range-expr.html
+[closure expressions]:  expressions/closure-expr.html
+[negation]:             expressions/operator-expr.html#negation-operators
+[arithmetic, logical]:  expressions/operator-expr.html#arithmetic-and-logical-binary-operators
+[comparison]:           expressions/operator-expr.html#comparison-operators
+[lazy boolean]:         expressions/operator-expr.html#lazy-boolean-operators
+[borrow]:               expressions/operator-expr.html#borrow-operators
+[interior mutability]:  interior-mutability.html
+[dereference operator]: expressions/operator-expr.html#the-dereference-operator
+[grouped]:              expressions/grouped-expr.html
+[cast]:                 expressions/operator-expr.html#type-cast-expressions
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -1,0 +1,21 @@
+# Constant evaluation
+
+Constant evaluation is the process of computing the result of
+[expressions] during compilation.
+
+## Const context
+
+A _const context_ is one of the following:
+
+* [array type length expressions]
+* repeat expression length expessions
+* the initializer of
+  * [constants]
+  * [statics]
+  * [enum discriminants]
+
+[array type length expressions]: types.html#array-and-slice-types
+[enum discriminants]: items/enumerations.html#custom-discriminant-values-for-field-less-enumerations
+[constants]: items/constant-items.html
+[statics]: items/static-items.html
+[expressions]: expressions.html#constant-expressions
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -187,7 +187,7 @@ on floating point computations.
 Exhaustive list of permitted structures in `const fn`:
 
 > **Note**: this list is more restrictive than what you can write in
-regular constants
+> regular constants
 
 * type parameters where the parameters only have any [trait bounds]
   of the following kind:
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -8,7 +8,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]
 >
 > _FunctionFront_ :\
-> &nbsp;&nbsp; `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
+> &nbsp;&nbsp; `const`<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
 >
 > _Abi_ :\
 > &nbsp;&nbsp; [STRING_LITERAL] | [RAW_STRING_LITERAL]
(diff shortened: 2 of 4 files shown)


src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+2. arithmetic operators on integers
+3. boolean operators (except for `&&` and `||` which are banned since they are short-circuiting).
+4. any kind of aggregate constructor (array, `struct`, `enum`, tuple, ...)
+5. calls to other *safe* `const fn`s (methods and functions)
+6. index operations on arrays and slices
+7. field accesses on structs and tuples
+8. reading from constants (but not statics, not even taking a reference to a static)
+9. `&` and `*` (only dereferencing of references, not raw pointers)
+10. casts except for raw pointer to `usize` casts
```
matthewjasper: any integer type.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+`const fn`s have various restrictions to makes sure that you cannot define a `const fn` that can't be evaluated at compile-time. You will, for example, never be able to write a random number generator as a const fn. Calling a const fn at compile-time will always yield the same result as calling it at runtime, even if you call it multiple times. There's one exception to this rule: if you are doing complex floating point operations in extreme situations, then you might get (very slightly) different results. It is adviseable to not make array lengths and enum discriminants depend on floating point computations.
+
+Exhaustive list of permitted structures in `const fn`:
+
+1. type parameters where the parameters have any of the following as part of their bounds (either on `where` or directly on the parameters):
+    1. lifetimes
+    2. `Sized`
+
+    This means that `<T: 'a + ?Sized>` and `<T: 'b + Sized>` + `<T>` are all permitted.
```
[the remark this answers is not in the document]
Centril: The generic parameters on functions, impls, etc.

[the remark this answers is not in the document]
matthewjasper: Sorry, I guess I wasn't clear there. I really mean, '+' should not appear in prose.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
 attributes], and [the optimization hint
 attributes].
 
+## Const functions
+
+Functions can be `const`, meaning they can be called from within array length expressions and the initializer of constants, statics and enum discriminants. When called from such a so-called "const context", the function is interpreted by the compiler at compile time. The interpretation happens in the environment of the compilation target and not the host. So `usize` is `32` bits if you are compiling against a `32` bit system, irrelevant of whether you are building on a `64` bit or a `32` bit system.
```
Havvy: Let's actually give a name and definition with this. You informally call it `` `const fn` `` in future paragraphs, but the name should be prose, not code, so I think "const function" or "constant function" (or both?) works. We should also be more explicit about how they are defined. I propose the following, but feel to tweak it, using "constant function" and using the length of this comment box for where to put in newlines instead of 80 chars...

```markdown
*Constant functions* are functions that can be called in [const context]. They
are defined by putting the `const` keyword before the `fn` keyword.
```

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+If a `const fn` is called outside a "const context", it is indistinguishable from any other function. You can freely do anything with a `const fn` that you can do with a regular fn.
```
Havvy: nit: Last "fn" -> "function".

src/items/functions.md
Centril: I want to avoid calling these "constant functions" because it gives the impression that the functions are somehow constant; they aren't. They are *pure* functions (at least currently and besides some weird business with floating points). They also can't always be called in `const ITEM = the_const_fn(...);` contexts if given runtime arguments and they can always be called in runtime contexts.

If you must have a name in prose (which strikes me as less formal) then I would call it "const functions".

Havvy: Your `"const context" is already defined as a [constant expression](https://doc.rust-lang.org/reference/expressions.html#constant-expressions). That said, the bit about interpretation using the target's environment is useful to add there. And the explicit list of where constant expressions must be evaluated at compile-time is :100:. 

I like the example as well, but it should probably say what the target's pointer size is before saying the size of usize. E.g. "For example, if you are compiling on a 32-bit OS, then `usize` has a size of `32` bits irrelevant of ..."

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+`const fn`s have various restrictions to makes sure that you cannot define a `const fn` that can't be evaluated at compile-time. You will, for example, never be able to write a random number generator as a const fn. Calling a const fn at compile-time will always yield the same result as calling it at runtime, even if you call it multiple times. There's one exception to this rule: if you are doing complex floating point operations in extreme situations, then you might get (very slightly) different results. It is adviseable to not make array lengths and enum discriminants depend on floating point computations.
```
Havvy: concision nit: "that you cannot define a `const fn` that can't" -> "that they can".

tone nit: Let's not blame the reader for having the audacity of trying to write an RNG as a constant function. Perhaps they're merely curious if somebody could, as it'd be something important to keep in mind if it was possible. "For example, a random number generator cannot be written as a constant function."

concision note: "if you call" -> "when called" (although this one is more syllabic concision?)

Since floating point is not permitted currently, the floating point section can be removed until it is decided what we do with it. We don't document the future.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+Exhaustive list of permitted structures in `const fn`:
```
Havvy: This list should be a bulleted list. The ordering is inconsequential.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+1. type parameters where the parameters only have any of the following as part of their bounds (either on `where` or directly on the parameters):
```
Havvy: The parenthetical can be replaced by just making "bounds" a link to `trait-bounds.html`.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+    1. lifetimes
+    2. `Sized`
+
+    This means that `<T: 'a + ?Sized>`, `<T: 'b + Sized>` and `<T>` are all permitted.
+    Note that `?Sized` is the absence of a constraint when bounds have been fully elaborated
```
Havvy: This note can replaced by putting ``or [`?Sized`]`` on the second bullet point in the sublist. Link to `trait-bounds.html#sized`

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+    which includes adding implicit `Sized` bounds.
+    This entails that permitting `Sized` + lifetimes allows the above examples.
+
+    This rule also applies to type parameters of items that contain `const fn`s.
```
Havvy: By "contain", I presume you mean implementations and traits. But do you also mean e.g. functions that contain a block expression that contains a constant function? Listing them out would help readability here.

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+
+2. arithmetic and comparison operators on integers
+3. all boolean operators except for `&&` and `||` which are banned since they are short-circuiting.
+4. any kind of aggregate constructor (array, `struct`, `enum`, tuple, ...)
+5. calls to other *safe* `const fn`s (methods and functions)
+6. index operations on arrays and slices
```
Havvy: Index expressions?

src/items/functions.md
Havvy: It would be useful to have a note here stating that this list is **more** restrictive than the list of valid constant expressions. Notes are written as a blockquote starting with bolded "Note": ` > **Note**: Content of note`

src/items/functions.md
```diff
@@ -160,6 +160,38 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
...
+    Note that `?Sized` is the absence of a constraint when bounds have been fully elaborated
+    which includes adding implicit `Sized` bounds.
+    This entails that permitting `Sized` + lifetimes allows the above examples.
+
+    This rule also applies to type parameters of items that contain `const fn`s.
+
+2. arithmetic and comparison operators on integers
+3. all boolean operators except for `&&` and `||` which are banned since they are short-circuiting.
+4. any kind of aggregate constructor (array, `struct`, `enum`, tuple, ...)
+5. calls to other *safe* `const fn`s (methods and functions)
```
Havvy: nit: "(methods and functions)" -> "whether by function call or method call

src/items/functions.md
oli-obk: only floating point math is forbidden. You can define `const fn foo(f: f32) -> f32 { f }`. Depending on the registers used at runtime, your value might actually change.

src/items/functions.md
oli-obk: Not quite. "const context" requires a constant expression. The definition in the documentation you linked is also incomplete. I'll add an entry and make both documentations point to it.

Havvy: > They also can't always be called in const ITEM = the_const_fn(...); contexts if given runtime arguments and they can always be called in runtime contexts.

The `constant expression` section currently states "The following expressions are constant expressions, so long as any operands are also constant expressions and do not cause any `Drop::drop` calls to be ran.". It's not the most precise wording, but it's supposed to handle that.

> they can always be called in runtime contexts

All constant expressions can be.

> If you must have a name in prose (which strikes me as less formal)

We want to have the name for other documentation none-the-less. And we want the name so that we can then discuss it in the prose further on.

src/items/functions.md
```diff
@@ -162,35 +162,57 @@ attributes].
...
-4. any kind of aggregate constructor (array, `struct`, `enum`, tuple, ...)
-5. calls to other *safe* `const fn`s (methods and functions)
-6. index operations on arrays and slices
-7. field accesses on structs and tuples
-8. reading from constants (but not statics, not even taking a reference to a static)
-9. `&` and `*` (only dereferencing of references, not raw pointers)
-10. casts except for raw pointer to integer casts
-11. `const unsafe fn` is allowed, but the body must consist of safe operations only and you won't be able to call the `const unsafe fn` from within another `const fn` even if you use `unsafe`
+> **Note**: this list is more restrictive than what you can write in
+regular constants
```
Havvy: Missing `>`.

oli-obk:
I think I addressed everything

src/items/functions.md
Havvy: This one isn't addressed.

Havvy:
Almost. My comment about giving a name to a function with `const` on it so that the prose can call it that is unaddressed.

Furthermore, new pages need to be added to `SUMMARY.md` or they will not be built.

oli-obk:
> My comment about giving a name to a function with const on it so that the prose can call it that is unaddressed.

I felt it was addressed by @Centril . I use `const fn` as a name for it. I'd be fine with calling them "const functions". I'm probably biased by attrition. I'll accept whatever the bikeshed results in. I'd just rather not have it here.

src/items/functions.md
```diff
@@ -160,6 +160,59 @@ attributes], [`must_use`], [the procedural macro attributes], [the testing
 attributes], and [the optimization hint
 attributes].
 
+## Const functions
+
+Functions can be `const`, meaning they can be called from within
```
Centril: I would somehow explain here that this applies so long as provided arguments (expressions) to the functions explicit parameters (e.g. those within `(...)` but not those within `<...>`) are also valid in const contexts.

oli-obk: We don't do that for operators. I don't see how `add(x, y)` differs from `x + y` in the "explain constness"-aspect

Centril: I would say we should do that for operators in that case. This is the reference, so better be as unambiguous as we can be.

src/const_eval.md
```diff
@@ -0,0 +1,81 @@
...
+* Index expressions, [array indexing] or [slice] with a `usize`.
+* [Range expressions].
+* [Closure expressions] which don't capture variables from the environment.
+* Built in [negation], [arithmetic, logical], [comparison] or [lazy boolean]
+  operators used on integer and floating point types, `bool` and `char`.
+* Shared [borrow]s, except if applied to a type with [interior mutability].
+* The [dereference operator].
+* [Grouped] expressions.
+* [Cast] expressions, except pointer to address and
+  function pointer to address casts.
```
Havvy: Missing a bullet point for calling const fns.

Havvy:
> I felt it was addressed by @Centril . I use const fn as a name for it. I'd be fine with calling them "const functions". I'm probably biased by attrition. I'll accept whatever the bikeshed results in. I'd just rather not have it here.

It's less a bikeshed and more that having spoken language names for concepts is important, and that we should introduce them explicitly. I'm perfectly fine with "const function". A lot of our spoken language names are just the keywords, but not in code blocks. E.g. "enums", "structs", "traits". And we can talk about these concepts without putting them in code blocks. It's easier to talk about "const functions" than it is to talk about "`const fn`s" since one is a prose language and the other is not.

Centril:
@Havvy that fair :) My worry was that by calling it "constant function" we might conflate it with `const FOO: fn() = <def>;`.

oli-obk:
> It's easier to talk about "const functions" than it is to talk about "const fns" since one is a prose language and the other is not.

I've been calling them "consteffens" as one word since basically forever. But I have no problem with yielding to "const functions".

Centril:
I think this is good for now; we can always revisit with improvements later on.

Havvy:
Yeah, there are changes I want to make (linkify everything) that I don't want to put on somebody who's not familiar with all the pages in the reference, bu what we have here is definitely good enough.

PR #441, opened 2018-10-06
 Document type layout of `str` #410 

### LINKED ISSUE #410: Document type layout of `str`
Totes missed this during the first pass.

It's the same as `[u8]`.

Maxgy:
I documented the size behavior of type `str` to the 9.3.Type Layout section.
Fixes #410

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -86,6 +86,9 @@ Slices have the same layout as the section of the array they slice.
 > Note: This is about the raw `[T]` type, not pointers (`&[T]`, `Box<[T]>`,
 > etc.) to slices.
 
+## `str` Layout
+String slices are a UTF-8 representation of characters that have the same layout as slices of type `[u8]`.
+
 ## Tuple Layout
 
 Tuples do not have any guarantees about their layout.


Maxgy:
Can someone please review my pull?

src/type-layout.md
```diff
@@ -73,6 +73,12 @@ at least equal to the size and alignment of a pointer.
 > title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of
 > the size of `usize` and have the same alignment.
 
+## `str` Layout
+
+Strings are laid out so that the `nth` character of the string is offset from the
```
Centril: We call these "string slices" in the standard library documentation to differentiate from the owned `String` type.

src/type-layout.md
```diff
@@ -73,6 +73,12 @@ at least equal to the size and alignment of a pointer.
...
+start of the string by `n * the size of a character` bytes. A string has a size of
```
Centril: Why not write `n * size_of::<char>()` here?

src/type-layout.md
```diff
@@ -73,6 +73,12 @@ at least equal to the size and alignment of a pointer.
...
+`size_of::<char>() * n` and the same alignment of `char`.
```
Centril:
```rust
#![feature(dbg_macro)]

fn main() {
    use std::mem::*;

    let foo: &'static str = "foo";

    dbg!(size_of_val(foo));
    dbg!(size_of_val(&'f') + size_of_val(&'o') + size_of_val(&'o'));

    dbg!(align_of::<char>());
    dbg!(align_of_val(&'f'));
    dbg!(align_of_val(foo));
}
```

==>

```
[src/main.rs:8] size_of_val(foo) = 3
[src/main.rs:9] size_of_val(&'f') + size_of_val(&'o') + size_of_val(&'o') = 12
[src/main.rs:11] align_of::<char>() = 4
[src/main.rs:12] align_of_val(&'f') = 4
[src/main.rs:13] align_of_val(foo) = 1
```

This phrasing doesn't seem accurate?

Maxgy: I see my mistakes. Thank you!

src/type-layout.md
Maxgy: Do string slices have a length equal to the number of characters?

alercah: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=592619fcf323ee9576fa83ce6ee90b76

Maxgy: So UTF-8 encoded characters can be anywhere from 1 to 4 bytes?

Maxgy:
Is this any better?

alercah:
Hm. I think this is ok, but we can do better. Since `str` is a slice type, I think that mentioning it in the slice section might make more sense, and then call out that it's a UTF-8 representation of its characters. As is, "each character" is kind of vague, and I don't think we need to duplicate the array section which goes into more detail. Really, `str` is just `[u8]` with an additional guarantee that it's valid UTF-8.

Maxgy:
restarting travis

Maxgy:
It should be clean now, @alercah .

PR #444, opened 2018-10-09
Referenceify conditional compilation

### LINKED ISSUE #384: Rewrite Conditional Compilation to be Referency
It hasn't been modified from when it was a part of The Guide, and needs to be updated to be referencey.

Havvy:
[Rendered](https://havvy.net/rust-reference/conditional-compilation.html)

Fixes #384 

Partial r? @ehuss for grammar additions.

I put a couple questions into comments into the source itself.

I also wonder if making each configuration option set by the compiler its own section like I did is too spacey. I could bring it back down into a list if y'all agree with that.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -267,7 +267,7 @@ mod os;
 > **Note**: The `cfg_attr` can expand to another `cfg_attr`. For example,
 > `#[cfg_attr(linux, cfg_attr(feature = "multithreaded", some_other_attribute))`
 > is valid. This example would be equivalent to
-> `#[cfg_attr(and(linux, feaure ="multithreaded"), some_other_attribute)]`.
+> `#[cfg_attr(all(linux, feature ="multithreaded"), some_other_attribute)]`.
 
 The `cfg_attr` attribute is allowed anywhere attributes are allowed except on
 generic parameters.
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -1,18 +1,213 @@
-## Conditional compilation
+# Conditional compilation
 
-Sometimes one wants to have different compiler outputs from the same code,
-depending on build target, such as targeted operating system, or to enable
-release builds.
+> **<sup>Syntax</sup>**\
+> _ConfigurationPredicate_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _ConfigurationOption_\
+> &nbsp;&nbsp; | _ConfigurationAll_\
+> &nbsp;&nbsp; | _ConfigurationAny_\
+> &nbsp;&nbsp; | _ConfigurationNot_
+>
+> _ConfigurationOption_ :\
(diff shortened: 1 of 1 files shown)


src/conditional-compilation.md
```diff
@@ -1,18 +1,213 @@
...
+
+### `target_arch`
+
+Key-value option set once with the target's CPU architecture. The value is
+similar to the first element of the platform's target triple, but not
+identical.
+
+Example values:
+
+* `"x86"`,
```
ehuss: There are some random trailing commas here.

src/conditional-compilation.md
```diff
@@ -1,18 +1,213 @@
...
+> &nbsp;&nbsp; &nbsp;&nbsp; _ConfigurationOption_\
+> &nbsp;&nbsp; | _ConfigurationAll_\
+> &nbsp;&nbsp; | _ConfigurationAny_\
+> &nbsp;&nbsp; | _ConfigurationNot_
+>
+> _ConfigurationOption_ :\
+> &nbsp;&nbsp; [IDENTIFIER]&nbsp;(`=` ([STRING_LITERAL] | [RAW_STRING_LITERAL]))<sup>?</sup>
+>
+> _ConfigurationAll_\
+> &nbsp;&nbsp; `all` `(` _ConfigurationPredicate_ (`,` _ConfigurationPredicate_)<sup>\*</sup> `,`<sup>?</sup> `)`
```
ehuss: These don't seem to capture that they can be an empty list.  Not sure if that is important, since that is an odd usage.

src/conditional-compilation.md
```diff
@@ -41,62 +239,72 @@ fn needs_not_foo() {
...
+#[cfg_attr(linux, path = "linux.rs")]
+#[cfg_attr(windows, path = "windows.rs")]
+mod os;
 ```
 
-This is the same as `#[b]` if `a` is set by `cfg`, and nothing otherwise.
+> **Note**: The `cfg_attr` can expand to another `cfg_attr`. For example,
+> `#[cfg_attr(linux, cfg_attr(feature = "multithreaded", some_other_attribute))`
+> is valid. This example would be equivalent to
+> `#[cfg_attr(and(linux, feaure ="multithreaded"), some_other_attribute)]`.
```
ehuss: feaure → feature

Havvy:
I don't like having them all the way on the bottom either, and the page is describing things in a bottom-up way. Perhaps we should invert it, and explain the attributes and macro first?

src/conditional-compilation.md
```diff
@@ -1,18 +1,216 @@
...
+Key-value option set once with either a value of "little" or "big" depending
+on the endianness of the target's CPU.
+
+### `target_pointer_width`
+
+Key-value option set once with the target's pointer width in bits. For example,
+for targets with 32-bit pointers, this is set to `"32"`. Likewise, it is set
+to `"64"` for targets with 64-bit pointers.
+
+<!-- Are there targets that have a different bit number? -->
```
matthewjasper: There's a 16 bit platform listed in tier 3 [here](https://forge.rust-lang.org/platform-support.html)

matthewjasper:
Thanks!

PR #446, opened 2018-10-09
Mention tool lint attributes

Manishearth:
Being stabilized in https://github.com/rust-lang/rust/pull/54870 , wanted to preserve some documentation. Clippy has more docs on this.

r? @steveklabnik

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -357,6 +357,35 @@ pub mod m3 {
 }
 ```
 
+#### Tool lint attributes
+
+Tool lints let you use scoped lints, to `allow`, `warn`, `deny` or `forbid` lints of
+certain tools.
+
+Currently `clippy` is the only available lint tool.
+
+They only get checked when the associated tool is active, so if you try to use an `allow` attribute for a nonexistant tool lint, the compiler will not warn about the nonexistant lint until you use the tool.
+
+Otherwise, they work just like regular lint attributes:
+
+
+```rust,ignore
+// set the entire `pedantic` clippy lint group to warn
+#![warn(clippy::pedantic)]
+// silence warnings from the `filter_map` clippy lint
+#![allow(clippy::filter_map)]
+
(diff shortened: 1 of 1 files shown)


src/attributes.md
```diff
@@ -357,6 +357,35 @@ pub mod m3 {
 }
 ```
 
+#### Tool lint attributes (`#[allow(clippy::foo)]`)
+
+Tool lints let you use scoped lints, to `allow`, `warn`, `deny` or `forbid` lints of
+certain tools.
+
+
```
matthewjasper: extra new line

Manishearth: fixed

Havvy:
This is okay as is. Don't take the rest of what I'm saying as a requirement to make changes, although do tell me if you want to make them. Otherwise, we can merge this as is (though it'd be nice to know Travis will pass...).

With that out of the way, I think that tool lints don't need their own section. Instead, we need a more general section on lint check values that says both identifiers and `ident:ident` is valid. We should also probably have a grammar for these attributes, like I added for `cfg_attr` and `cfg` in #444. That we don't say that the lint names must be valid identifiers is an oversight I did not notice until you posted this PR.

Manishearth:
Slightly feel they should be in their own section, since the conditional checking of tool lints is important, prefer to merge as-is.

Added an `ignore` to the test since this is nightly-only right now (stabilized, but hasn't yet reached stable)

src/attributes.md
```diff
@@ -357,6 +357,35 @@ pub mod m3 {
 }
 ```
 
+#### Tool lint attributes (`#[allow(clippy::foo)]`)
```
Havvy: One nit before I merge: Everything in the parenthesis should not be in the header.

Manishearth:
done
-Manish Goregaokar

Havvy:
💟 Thanks!

PR #447, opened 2018-10-13
Const function

Havvy:
This PR does 7 things:

1. Capitalize lists.
2. Define const function with italics.
3. Call extern and const qualifiers. We may want to look at calling pub a qualifier as well.
4. Fix enum discriminants to say they take a constant expression.
5. Explain that const functions have restrictions more tersely.
6. Fix link to constant expressions on array literal page.
7. Move attributes on functions back down to the bottom of the page.
8. Rename `FunctionFront` to `FunctionQualifier` in the grammar as well.

The first six are in the first commit. The last two are in their own commit.

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -114,7 +114,7 @@ opposite functionality to [external blocks]. Whereas external
 blocks allow Rust code to call foreign code, extern functions with bodies
 defined in Rust code _can be called by foreign code_. They are defined in the
 same way as any other Rust function, except that they have the `extern`
-modifier.
+qualifier.
 
 ```rust
 // Declares an extern fn, the ABI defaults to "C"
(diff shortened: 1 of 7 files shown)


src/items/functions.md
```diff
@@ -2,12 +2,12 @@
 
 > **<sup>Syntax</sup>**\
 > _Function_ :\
-> &nbsp;&nbsp; _FunctionFront_ `fn` [IDENTIFIER]&nbsp;[_Generics_]<sup>?</sup>\
+> &nbsp;&nbsp; _FunctionQualifiers_ `fn` [IDENTIFIER]&nbsp;[_Generics_]<sup>?</sup>\
 > &nbsp;&nbsp; &nbsp;&nbsp; `(` _FunctionParameters_<sup>?</sup> `)`\
 > &nbsp;&nbsp; &nbsp;&nbsp; _FunctionReturnType_<sup>?</sup> [_WhereClause_]<sup>?</sup>\
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]
 >
-> _FunctionFront_ :\
```
ehuss: This will need to be updated in a few other places (types, associated-items, traits).

Havvy: Done and rebased. My bad for not going a grep.

ehuss: Thanks, I think it's much better than my original.

The new update uses `_FunctionQualifier_` singular, whereas the definition is `_FunctionQualifiers_` plural.  I think they all should probably be the plural one.

Havvy: That was just me being tired in the morning. Fixed and rebased with just that change.

matthewjasper:
Thanks!

PR #450, opened 2018-10-14
Document mod.rs changes.

ehuss:
Stabilization PR: https://github.com/rust-lang/rust/pull/54072
Tracking issue: https://github.com/rust-lang/rust/issues/53125
Part of [RFC 2126](https://github.com/rust-lang/rfcs/blob/master/text/2126-path-clarity.md).

### DIFF
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -42,10 +42,10 @@ extension `.rs`.
 
 A Rust source file describes a module, the name and location of which &mdash;
 in the module tree of the current crate &mdash; are defined from outside the
-source file: either by an explicit [`mod` item][module] in a referencing source file, or
-by the name of the crate itself. Every source file is a module, but not every
-module needs its own source file: [module definitions][module] can be nested
-within one file.
+source file: either by an explicit [`mod` item][module] in a referencing
+source file, or by the name of the crate itself. Every source file is a
+module, but not every module needs its own source file: [module
+definitions][module] can be nested within one file.
 
 Each source file contains a sequence of zero or more `item` definitions, and
 may optionally begin with any number of [attributes]
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -14,7 +14,7 @@
 
 > Note: Although Rust, like any other language, can be implemented by an
 > interpreter as well as a compiler, the only existing implementation is a
-> compiler,and the language has always been designed to be compiled. For these
+> compiler, and the language has always been designed to be compiled. For these
 > reasons, this section assumes a compiler.
 
 Rust's semantics obey a *phase distinction* between compile-time and
@@ -42,7 +42,7 @@ extension `.rs`.
 
 A Rust source file describes a module, the name and location of which &mdash;
 in the module tree of the current crate &mdash; are defined from outside the
-source file: either by an explicit `mod_item` in a referencing source file, or
+source file: either by an explicit [`mod` item][module] in a referencing source file, or
 by the name of the crate itself. Every source file is a module, but not every
 module needs its own source file: [module definitions][module] can be nested
 within one file.
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -120,8 +120,6 @@ mod thread {
 }
 ```
 
-
-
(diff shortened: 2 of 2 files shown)


src/items/modules.md
```diff
@@ -40,21 +40,26 @@ struct, enumeration, union, type parameter or crate can't shadow the name of a
...
+`crate::util`         | `util.rs`        | `mod config;`
+`crate::util::config` | `util/config.rs` |
 
-mod thread {
-    // Load the `local_data` module from `thread/local_data.rs`
-    // or `thread/local_data/mod.rs`.
-    mod local_data;
-}
-```
+> **Note**: Module filenames may also be the name of the module as a directory
```
Havvy: That `util/mod.rs` is also valid is a part of the language, and not just a note. The previous to Rust 1.30 should stay as a note (though I'd rather it say `rustc`, since I don't think Rust is versioned right now). We should also explicitly say it is deprecated.

---

Also, is this allowed?

```rust
#[path = "util.rs"] mod util_1;
#[path = "util/mod.rs"] mod util_2;
```

ehuss: I'll move it out of the note, though that may complicate it some.

> We should also explicitly say it is deprecated.

I wouldn't say it is deprecated.  The RFC explicitly says "The use of mod.rs continues to be allowed without any deprecation."  To me, deprecated usually implies a lint that triggers, but in this case not even the edition-idiom lints recommend this. I carefully chose to say it is "encouraged" though I don't think anyone has officially established even that.

> Also, is this allowed?

Haha, yes.

Havvy: Alright. Scratch the deprecation part then.

Though, given the interaction exists, I'm thinking it might it make sense to describe the chosen file as an algorithm? I want to say probably not, since we can statically state that the path attribute changes the resolution. But in that case, we should state that this is only the case when the path attribute is not applied.

ehuss: > this is only the case when the path attribute is not applied.

Yea, I tried to imply that by starting with "By default", but that is a little too subtle.  I'll try to think of a way to make it clear the distinction between "default" behavior and overridden behavior.

Havvy: Given that the only non-default is when there's a path attribute, you can explicitly say "When the module has no path attribute"

ehuss:
Note: I'm still waiting for the resolution for rust-lang/rust#55560 and rust-lang/rust#55094 before finishing this, since things are still in flux.

cramertj:
Clarifying: the current implemented behavior is that using `#[path]` in a non-mod.rs file will be relative to the current file's directory *unless* it is in an inline `mod x { ... }`, in which case it is relative to the current module name (usually the name of the file) plus any inline modules. So `#[path = "foo.rs"] mod y;` in `x.rs` will look for `foo.rs` in the same directory as `x.rs`, but `mod y { #[path = "foo.rs"] mod z; }` in `x.rs` will look for ./x/y/foo.rs`.

ehuss:
I have pushed a new update. I tried to balance being precise and clear with being too verbose. I don't feel I hit the target, though. It's very difficult to describe path attributes succinctly.

src/items/modules.md
```diff
@@ -40,34 +40,90 @@ struct, enumeration, union, type parameter or crate can't shadow the name of a
...
-    // Load the `local_data` module from `thread_files/tls.rs`
+    // Load the `local_data` module from `thread_files/tls.rs` relative to
+    // this source file's directory.
     #[path = "tls.rs"]
     mod local_data;
 }
 ```
 
+
+
```
matthewjasper: Extra newlines

src/crates-and-source-files.md
```diff
@@ -42,7 +42,7 @@ extension `.rs`.
 
 A Rust source file describes a module, the name and location of which &mdash;
 in the module tree of the current crate &mdash; are defined from outside the
-source file: either by an explicit `mod_item` in a referencing source file, or
+source file: either by an explicit [`mod` item][module] in a referencing source file, or
```
matthewjasper: line should be wrapped

matthewjasper:
Thanks!

PR #455, opened 2018-10-17
`extern crate` changes.

ehuss:
Stabilization PR: https://github.com/rust-lang/rust/pull/54403

### DIFF
--- a/src/items/extern-crates.md
+++ b/src/items/extern-crates.md
@@ -5,8 +5,8 @@
 > &nbsp;&nbsp; `extern` `crate` [IDENTIFIER]&nbsp;(`as` [IDENTIFIER])<sup>?</sup> `;`
 
 An _`extern crate` declaration_ specifies a dependency on an external crate.
-The external crate is then bound into the declaring scope as the `ident`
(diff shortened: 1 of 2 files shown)


src/items/extern-crates.md
```diff
@@ -38,6 +38,50 @@ Here is an example:
...
+> // Instead, you must reference the item in scope from the `extern crate`
+> // declaration.
+> use self::proc_macro::TokenStream;
+> ```
+
+<!--
+Possible upcoming changes that will change this:
+
+`extern crate` items will automatically be added to the extern prelude.
+    https://github.com/rust-lang/rust/pull/54658
```
matthewjasper: This is merged now

ehuss: The changes are behind the `extern_crate_item_prelude` feature flag, so there's not much difference.  I don't know what the future holds, there are no tracking issues. I tweaked the wording slightly, but it still seems a little awkward to me.  Is the distinction between the `--extern` added to extern prelude and `extern crate` being placed as an item in the current scope clear enough?  Should it emphasize this distinction more?

Issue #456, opened 2018-10-21
? kleene macro operator is unstable


ehuss:
As noted in PR #424, the ? operator is still unstable.  I'm wondering if it should stay (since unstable things usually aren't document).  It looks like it may get stabilized soonish, so maybe it doesn't matter? (tracking issue: https://github.com/rust-lang/rust/issues/48075) 

I've written the grammar for macros which includes this, but I'm unsure if I should continue the trend of keeping this unstable syntax.

ehuss:
This is now stabilized in rust-lang/rust#56245.

PR #458, opened 2018-10-23
Grammar: macros

ehuss:


ehuss:
- I included the unstable `?` kleene operator, see #456.
- This doesn't add any details about how substitution works, since that is a big and complex topic, not directly related to the grammar rules.
- The "novis" footnotes seem a little awkward. Not sure of the best way to express that.

src/items.md
```diff
@@ -17,8 +17,8 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_Trait_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_Implementation_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ExternBlock_]\
-> &nbsp;&nbsp; &nbsp;&nbsp; | _Macro_\
-> &nbsp;&nbsp; &nbsp;&nbsp; | _MacroDefinition_\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_MacroInvocationSemi_]&nbsp;[^novis]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_MacroRulesDefinition_]&nbsp;[^novis]\
```
matthewjasper: Was this deliberately not written like this?
_OuterAttribute_ ( _Visibility_ <sup>?</sup> ( _Module_ | ... | _ExternBlock_ ) | _MacroInvocationSemi_ | _MacroRulesDefinition_ )

ehuss: I was trying to avoid making it too nested.  In that example, there would need to be another set of parenthesis, like: _OuterAttribute_<sup>\*</sup> ( ( _Visibility_<sup>?</sup> ( _Module_ | ... | _ExternBlock_ ) ) | _MacroInvocationSemi_ | _MacroRulesDefinition_ ) which seems a little unwieldy.  I pushed a variation that removes the footnote, let me know if it looks any better.

matthewjasper: I think I prefer this version because the macro can expand to an item with a visibility.

matthewjasper:
Thanks!

src/macros-by-example.md
```diff
@@ -94,9 +135,36 @@ Rust syntax is restricted in two ways:
    a macro definition like `$i:expr [ , ]` is not legal, because `[` could be part
    of an expression. A macro definition like `$i:expr,` or `$i:expr;` would be legal,
    however, because `,` and `;` are legal separators. See [RFC 550] for more information.
+   Specifically:
+
+   * `expr` and `stmt` may only be followed by one of `=>`, `,`, or `;`.
+   * `pat` may only be followed by one of `=>`, `,`, `=`, `|`, `if`, or `in`.
+   * `path` and `ty` may only be followed by one of `=>`, `,`, `=`, `|`, `;`,
+     `:`, `>`, `[`, `{`, `as`, `where`, or a macro variable of `block`
```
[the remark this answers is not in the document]
estebank: Might want to add a note that `>>` will not work in compilers prior to 1.31. (Didn't think about backporting it to 1.30, which we could have.)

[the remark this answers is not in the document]
Havvy: The reference documents as it is today, not what it was yesterday. We don't care about older releases.

PR #460, opened 2018-10-26
Add a missing link.

yangby-cryptape:
The 3rd paragraph in [Derive mode macros](https://doc.rust-lang.org/nightly/reference/procedural-macros.html#derive-mode-macros) missed a link.


Havvy:
Thanks for catching that. Although we are trying to keep the list of links alphabetized, so can you move the link to after "attributes"?

yangby-cryptape:
Ok, done!

Havvy:
💟 Thanks!

PR #464, opened 2018-11-16
Fix some grammar issues.

ehuss:
- `trait` was missing bounds.
    `trait T: Clone {}`
- Type for `const` in trait is not optional.
    ```rust
    trait T {
        const C = 1;  // INVALID
        const C2: i32 = 1;
    }
    ```
- Associated type in trait can have a bare `:`:
    `trait T {
        type X:;
    }`
- Visibility is not needed on UseDeclaration, it is part of Item.
- `:` in generic lifetime param was missing parenthesis grouping to make it optional.
    `fn f<'a>() {}`


matthewjasper:
Thanks!

PR #469, opened 2018-11-19
Mention that #[cold] and #[inline] work on closures

jonas-schievink:
Both make their way to LLVM without problems today. Putting attributes on expressions is mostly still unstable, but can be worked around thanks to https://github.com/rust-lang/rust/issues/32796, so it's possible to rely on this in stable code.


matthewjasper:
Thanks!

PR #471, opened 2018-11-29
Add existential as a weak keyword

dlrobertson:


Centril:
> I believe this is an unstable feature, correct? 

Correct. The surface syntax will be changed before stabilization.

>  I don't think existential is reserved, or affects stable code in any way, correct?

Correct.

dlrobertson:
Ah, my bad

PR #478, opened 2018-12-09
unions have no active field

RalfJung:
Cc @ubsan @nikomatsakis

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -40,7 +40,7 @@ let f = u.f1;
 Unions have no notion of an "active field".  Instead, every union access just
 interprets the storage at the type of the field used for the access.  Reading a
 union field is equivalent to a [`transmute`]: The data in the union, no matter
-how it was stored there, is transmuted to the type if the field.  Reading data
+how it was stored there, is transmuted to the type of the field.  Reading data
 at a bad type results in undefined behavior (for example, reading the value `3`
 at type `bool`).  For this reason, all reads of union fields have to be placed
 in `unsafe` blocks:
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -145,9 +145,6 @@ aspects of Rust language (such as privacy, name resolution, type inference,
 generics, trait implementations, inherent implementations, coherence, pattern
 checking, etc etc etc).
 
-More detailed specification for unions, including unstable bits, can be found
-in [RFC 1897 "Unions v1.2"](https://github.com/rust-lang/rfcs/pull/1897).
-
 [IDENTIFIER]: identifiers.html
 [_Generics_]: items/generics.html
 [_WhereClause_]: items/generics.html#where-clauses
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -41,10 +41,10 @@ Unions have no notion of an "active field". Instead, every union access just
 interprets the storage at the type of the field used for the access. Reading a
 union field reads the bits of the union at the field's type. It is the
 programmer's responsibility to make sure that the data is valid at that
-type. Failing to do so results in undefined behavior (for example, reading the
-value `3` at type `bool` is undefined behavior). Effectively, writing to and
-then reading from a union is equivalent to a [`transmute`] from the type used
-for writing to the type used for reading.
+type. Failing to do so results in undefined behavior. For example, reading the
+value `3` at type `bool` is undefined behavior. Effectively, writing to and then
+reading from a union is equivalent to a [`transmute`] from the type used for
+writing to the type used for reading.
 
 Consequently, all reads of union fields have to be placed in `unsafe` blocks:
 
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -40,9 +40,9 @@ let f = u.f1;
 Unions have no notion of an "active field". Instead, every union access just
 interprets the storage at the type of the field used for the access. Reading a
 union field is equivalent to a [`transmute`]: The data in the union, no matter
-how it was stored there, is transmuted to the type of the field  Just like with
+how it was stored there, is transmuted to the type of the field. Just like with
 any other transmute, it is the programmer's responsibility to make sure that the
-data is valid for the field type. Failing to do so results in undefined
+data is valid at the field's type. Failing to do so results in undefined
 behavior (for example, reading the value `3` at type `bool` is undefined
 behavior). For this reason, all reads of union fields have to be placed in
 `unsafe` blocks:
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -30,21 +30,21 @@ let u = MyUnion { f1: 1 };
 ```
 
 The expression above creates a value of type `MyUnion` and initializes the
-storage using field `f1`.  The union can be accessed using the same syntax as
+storage using field `f1`. The union can be accessed using the same syntax as
 struct fields:
 
 ```rust,ignore
 let f = u.f1;
 ```
 
-Unions have no notion of an "active field".  Instead, every union access just
-interprets the storage at the type of the field used for the access.  Reading a
+Unions have no notion of an "active field". Instead, every union access just
+interprets the storage at the type of the field used for the access. Reading a
 union field is equivalent to a [`transmute`]: The data in the union, no matter
-how it was stored there, is transmuted to the type of the field.  Just like with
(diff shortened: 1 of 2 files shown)


src/items/unions.md
```diff
@@ -29,18 +29,22 @@ struct types, except that it must specify exactly one field:
...
 ```
 
-Inactive fields can be accessed as well (using the same syntax) if they are
-sufficiently layout compatible with the current value kept by the union.
-Reading incompatible fields results in undefined behavior. However, the active
-field is not generally known statically, so all reads of union fields have to
-be placed in `unsafe` blocks.
+Unions have no notion of an "active field".  Instead, every union access just
+interprets the storage at the type of the field used for the access.  The effect
+of reading a union with a different field than it was written to is that of
```
petrochenkov: "field ... it was written to" here looks suspiciously similar to the "active field" that's intended to be removed.

petrochenkov: The change compared to the previous semantics, I think, is that the "active field"/"field it was written to" doesn't necessarily exist with the new rules.
(While using the "active field" or some other wording for the field in question is a matter of terminology.)

RalfJung: I tried to improve the wording to make it more clear that there is nothing remembering which field was written to, just re-interpreting storage at different types.  Does that help?

briansmith:
The bigger problem with the current text is that it says "More detailed specification for unions, including unstable bits, can be found in RFC 1897," i.e. it defers to the RFC(s) that say that type punning is UB. I think the RFC reference should be removed too to clarify that the reference is now more correct than the RFC.

RalfJung:
This is not a conclusion from a UCG discussion.  It is, IMO, a better description of the current behavior of Rust: Rust never disallowed type punning, and describing the behavior of Rust never required the concept of an "active field".  "Active field" is a notion coming from C where it is required to enable type-based alias analysis, which was a definite non-goal for Rust for many years.  I know of no other purpose of this notion.

This does not change the semantics.  The RFC this text refers to (and I will remove that reference) did not get accepted.

RalfJung:
> I dislike the "bad type" sentence. The problem is not the type, it is the value being interpreted as that type.

Agreed, I tried to improve the wording. What do you think?

src/items/unions.md
```diff
@@ -29,18 +29,23 @@ struct types, except that it must specify exactly one field:
 let u = MyUnion { f1: 1 };
 ```
 
-The expression above creates a value of type `MyUnion` with active field `f1`.
-Active field of a union can be accessed using the same syntax as struct fields:
+The expression above creates a value of type `MyUnion` and initializes the
+storage using field `f1`.  The union can be accessed using the same syntax as
```
Centril:
```suggestion
storage using field `f1`. The union can be accessed using the same syntax as
```

src/items/unions.md
```diff
@@ -29,18 +29,23 @@ struct types, except that it must specify exactly one field:
...
-sufficiently layout compatible with the current value kept by the union.
-Reading incompatible fields results in undefined behavior. However, the active
-field is not generally known statically, so all reads of union fields have to
-be placed in `unsafe` blocks.
+Unions have no notion of an "active field".  Instead, every union access just
+interprets the storage at the type of the field used for the access.  Reading a
+union field is equivalent to a [`transmute`]: The data in the union, no matter
+how it was stored there, is transmuted to the type of the field.  Just like with
+any other transmute, it is the programmer's responsibility to make sure that the
+data is valid for the field type.  Failing to do so results in undefined
```
Centril:
```suggestion
data is valid at the field's type. Failing to do so results in undefined
```

Centril:
(While the reference isn't normative and may contain `rustc` specifics, we should *strive* to avoid that and be as spec-y as we can...)

That said; the wording "storage" carries connotations of `union`s being bags of bits that hasn't been agreed to yet. Is there a more neutral phrasing perhaps?

While in general, unions may not have an active field, that may still be part of an API invariant for a *specific* `union` data-type.

petrochenkov:
>"More detailed specification for unions, including unstable bits, can be found in RFC 1897," i.e. it defers to the RFC(s) that say that type punning is UB. 

Where did you find that?
RFC 1897 [explicitly said](https://github.com/petrochenkov/rfcs/blob/e5266bd105f592f7408b8592c5c3deaccba7f1ec/text/1444-union.md#type-punning) that supporting type punning is the most important requirement for unions.
(Also it's still the more detailed specification for unions as they are implemented now, despite the "active field" doesn't necessarily exist change.)

src/items/unions.md
RalfJung: Ah, my emacs and also my brain automatically use double-space between sentences.  I'll see if I can remove them all, but they might appear again as I edit: a double-space after a full stop is pretty deep in my muscle memory at this point. ;)

Centril:
@RalfJung I suppose "bag of bits with restriction on the bits" still makes sense if you say "storage" heh ;)

briansmith:
> Where did you find that?
> RFC 1897 [explicitly said](https://github.com/petrochenkov/rfcs/blob/e5266bd105f592f7408b8592c5c3deaccba7f1ec/text/1444-union.md#type-punning) that supporting type punning is the most important requirement for unions.

I followed the link to RFC 1897. Then I noticed that it was rejected (postponed), so I went and found the latest *accepted* RFC for unions, which was RFC 1444, IIUC. That says "In addition, since a union declared without #[repr(C)] uses an unspecified binary layout, code reading fields of such a union or pattern-matching such a union must not read from a field other than the one written to. This includes pattern-matching a specific value in a union field."

> > (Also it's still the more detailed specification for unions as they are implemented now, despite the "active field doesn't necessarily exist" change.)

I think the best way to fix that is to copy whatever is good from RFC 1897 into the reference. End-users like me have no way to interpret the authority of a postponed RFC, so using it as the reference is not good.

ubsan:
@RalfJung they're not transmutes though - they're reading bits at a type. `transmute` implies that they were initially a value of one type, that is then read as another type - in this case, you're simply doing the latter half. `transmute` is an extra step.

RalfJung:
@ubsan The first half happened when writing to a union field (including initializing the union).

Any suggestions for how to word this?

matthewjasper:
Isn't `transmute_copy` closer. The union may be larger than some of its fields.

RalfJung:
`transmute_copy` is about bypassing `!Copy`, what does it have to do with being larger?

matthewjasper:
It also doesn't require that its two type parameters are the same size, which transmute does.

ubsan:
@RalfJung I would rather word it similarly to the transmute documentation - "reads the bits of the union as a value of the field type"

RalfJung:
@ubsan better like this?

ubsan:
@RalfJung definitely! I like the new wording a lot more :)

src/items/unions.md
```diff
@@ -29,18 +29,24 @@ struct types, except that it must specify exactly one field:
...
-Inactive fields can be accessed as well (using the same syntax) if they are
-sufficiently layout compatible with the current value kept by the union.
-Reading incompatible fields results in undefined behavior. However, the active
-field is not generally known statically, so all reads of union fields have to
-be placed in `unsafe` blocks.
+Unions have no notion of an "active field". Instead, every union access just
+interprets the storage at the type of the field used for the access. Reading a
+union field reads the bits of the union at the field's type. It is the
+programmer's responsibility to make sure that the data is valid at that
+type. Failing to do so results in undefined behavior (for example, reading the
```
Havvy: The parenthetical should just be promoted to its own sentence. It can stay in the same location.

ubsan:
@Havvy that would be reading uninitialized memory, and at best would result in an unspecified value.

RalfJung:
@Havvy it'll just read whatever you put into those bytes. If you never put anything into them, as @ubsan said you get uninitialized data.

Havvy:
The PR doesn't state that that is what happens.

RalfJung:
It says that the bytes get read at the given type.  I cannot see any ambiguity in there. The comparison with transmute is just to connect to a mental model the reader might already have formed.

Do you have suggestions for wording that you think you improve clarity here?

Havvy:
I think I'm quibbling on the word "equivalent". "analogous" might be a better choice here?

matthewjasper:
Thanks

PR #480, opened 2018-12-10
Update pub(in path) for 2018.

ehuss:
When updating paths, I neglected to add a note that visibility paths must
be absolute.

### DIFF
--- a/src/visibility-and-privacy.md
+++ b/src/visibility-and-privacy.md
@@ -155,7 +155,8 @@ be a parent module of the item whose visibility is being declared.
 to `pub(in self)`.
 
 > **Edition Differences**: Starting with the 2018 edition, paths for
-> `pub(in path)` must start with `crate`, `self`, or `super`.
+> `pub(in path)` must start with `crate`, `self`, or `super`. The 2015 edition
+> may also use paths starting with `::` or modules from the crate root.
 
 Here's an example:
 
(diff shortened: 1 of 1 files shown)


src/visibility-and-privacy.md
```diff
@@ -154,13 +154,18 @@ be a parent module of the item whose visibility is being declared.
 - `pub(self)` makes an item visible to the current module. This is equivalent
 to `pub(in self)`.
 
+> **Edition Differences**: Starting with the 2018 edition, paths for
+> `pub(in path)` must start with `crate`, `self`, or `super`.
```
matthewjasper: Isn't this just a consequence of the path having to be absolute?

ehuss: Yes, but there is nothing else that explicitly calls that out for visibility paths.  The only other part is the "use declarations", which only talk about the requirements in use declarations.  Or are you saying it should be worded differently?

matthewjasper: I didn't realise how the reference described paths in use items. The wording here is probably fine, but I think that there should be some wording on how paths that start with an identifier are resolved in the 2015 edition.

ehuss: Oops, good catch!

matthewjasper:
Thanks!

PR #484, opened 2018-12-23
Update `unsafe` in const functions documentation

### LINKED ISSUE #482: Document min_const_unsafe_fn
In https://github.com/rust-lang/rust/pull/57067, the feature gate `min_const_unsafe_fn` is stabilized; this is a subset of RFC 911; Before 1.33 ships we need to update reference noting that:
+ `const unsafe fn`s may call other `const unsafe fn`s
+ `unsafe { .. }` blocks in `const unsafe? fn`s may call `const unsafe fn`s.

oli-obk:
fixes #482 

cc @Centril

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -185,9 +185,9 @@ Exhaustive list of permitted structures in const functions:
 * Reading from constants (but not statics, not even taking a reference to a static)
 * `&` and `*` (only dereferencing of references, not raw pointers)
 * Casts except for raw pointer to integer casts
-* `const unsafe fn` is allowed, but the body must consist of safe operations
-  only and you won't be able to call the `const unsafe fn` from within another
-  const function even if you use `unsafe`
+* `unsafe` blocks and `const unsafe fn` is allowed, but the body/block must consist
+  of safe operations or calls to other const functions only. Further unsafe operations
+  may get allowed in const functions in the future.
 
 ## Attributes on functions
 
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -185,9 +185,9 @@ Exhaustive list of permitted structures in const functions:
 * Reading from constants (but not statics, not even taking a reference to a static)
 * `&` and `*` (only dereferencing of references, not raw pointers)
 * Casts except for raw pointer to integer casts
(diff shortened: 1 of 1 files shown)


src/items/functions.md
```diff
@@ -185,9 +185,9 @@ Exhaustive list of permitted structures in const functions:
 * Reading from constants (but not statics, not even taking a reference to a static)
 * `&` and `*` (only dereferencing of references, not raw pointers)
 * Casts except for raw pointer to integer casts
-* `const unsafe fn` is allowed, but the body must consist of safe operations
-  only and you won't be able to call the `const unsafe fn` from within another
-  const function even if you use `unsafe`
+* `unsafe` blocks and `const unsafe fn` is allowed, but the body/block must consist
```
Centril: This feels unclear to me; I'd rephrase this as:

> * `unsafe` blocks and `const unsafe fn` in which,
>    in addition to performing safe `const` operations and calls, it is legal to:
>    - Call `const unsafe fn`s.
>
>    Further unsafe operations may be permitted the body/block in the future.

BatmanAoD:
Just for my own understanding: it looks like this _doesn't_ actually permit anything truly "unsafe" inside `const fn`, except maybe inside the standard library?

Is there ongoing discussion somewhere about additional `unsafe` features we might want to permit inside `const fn`? I'm interested in what the definition of undefined behavior would be inside a `const fn`. (Come to think of it, I'm now also wondering what the implications of undefined behavior inside a C++ `constexpr` function are, and whether the standard even addresses that...)

oli-obk:
A few truly unsafe things are tracked in https://github.com/rust-lang/rust/issues/57563 but you are right, we need to figure out our story around unsafe and const safety. This is tracked somewhat in https://github.com/rust-rfcs/const-eval/issues/14

Centril:
> Just for my own understanding: it looks like this _doesn't_ actually permit anything truly "unsafe" inside `const fn`, except maybe inside the standard library?

https://doc.rust-lang.org/nightly/std/num/struct.NonZeroU8.html#method.new_unchecked

BatmanAoD:
@Centril Ah, okay, I can see how calling that in user code would be truly "unsafe". Thanks.

PR #490, opened 2018-12-28
Note that UB is program-global

### LINKED ISSUE #485: Is UB in FFI also UB in Rust?
Link to ref: https://doc.rust-lang.org/reference/behavior-considered-undefined.html

> Warning: The following list is not exhaustive. There is no formal model of Rust's semantics for what is and is not allowed in unsafe code, so there may be more behavior considered unsafe.

My question is:
Does undefined behavior inside an extern FFI function also cause an UB in Rust?

------

There is an item inside of the list of **Behavior considered undefined**:
> * Invoking undefined behavior via compiler intrinsics:

I consider adding one more item:
> * Invoking undefined behavior via FFI:

Havvy:
Fixes #485 

I just wrote this up in a few seconds, so not sure how good the writing is. Figured a note would be good for undefined behavior crossing language barriers.

r? @RalfJung or @ubsan

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -40,10 +40,10 @@ code.
   * Non-UTF-8 byte sequences in a `str`.
 
 > **Note**: Undefined behavior affects the entire program. For example, calling
-> a function in `C` that exhibits undefined behavior of `C` means your entire
-> program contains undefined behaviour that can also affect the `Rust` code.
-> And vice versa, undefined behavior in `Rust` can cause adverse affects on
-> code executed by any FFI calls to other languages.
+> a function in C that exhibits undefined behavior of C means your entire
+> program contains undefined behaviour that can also affect the Rust code. And
+> vice versa, undefined behavior in Rust can cause adverse affects on code
+> executed by any FFI calls to other languages.
 
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
 [pointer aliasing rules]: http://llvm.org/docs/LangRef.html#pointer-aliasing-rules
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -39,6 +39,12 @@ code.
   * A value in a `char` which is a surrogate or above `char::MAX`.
   * Non-UTF-8 byte sequences in a `str`.
 
+> **Note**: Undefined behavior affects the entire program. For example, calling
+> a function in `C` that exhibits undefined behavior of `C` means your entire
+> program contains undefined behaviour that can also affect the `Rust` code.
+> And vice versa, undefined behavior in `Rust` can cause adverse affects on
+> code executed by any FFI calls to other languages.
+
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
 [pointer aliasing rules]: http://llvm.org/docs/LangRef.html#pointer-aliasing-rules
 [undef]: http://llvm.org/docs/LangRef.html#undefined-values


Centril:
I think suggesting that UB is bounded at the process level is too nice as it could cause UB in your entire system depending on OS?

Otherwise it looks good, I just don't know if it belongs in the reference or in the nomicon. It seems the latter is more appropriate?

Havvy:
Anything about unsafety/UB in the reference should be taught in the nomicon. The reference is the place to go for referring to what you've already learned from TRPL or the Nomicon. So, in that sense, it should be placed there.

As for UB affecting the whole OS, I'm not sure how that works unless the process is compiled as part of the OS. Unless you mean the general sense of having a process that could do anything means you have a system that can do anything?

Centril:
> Anything about unsafety/UB in the reference should be taught in the nomicon. The reference is the place to go for referring to what you've already learned from TRPL or the Nomicon. So, in that sense, it should be placed there.

I'm thinking of the reference more as an informal specification of Rust's syntax and static/dynamic semantics; but not as a place where you get more general advanced advice, which the proposed change strikes me as.

> As for UB affecting the whole OS, I'm not sure how that works unless the process is compiled as part of the OS. Unless you mean the general sense of having a process that could do anything means you have a system that can do anything?

I'm not sure (totally not an OS expert by any means...). I'm thinking in these terms: Consider a process that runs in ring 0... If UB in one ring-0 process causes it to corrupt memory of other processes, could it not cause UB in those?

Havvy:
If all processes share the same address space, then memory corruption can be cross-process. But at that point, you could consider all programs part of the same process? But also, by the point somebody runs across this condition, they've hopefully been warned about it in the docs on writing for such an OS. And explaining it here would just muddy everyone else's understanding of it?

It is more general semantics of inter-programming language stuff than explanation of the semantics of the Rust language. Thus I put it in a *note*. For what's a quick list of UB things to reference, reminding people that UB is not necessarily just in Rust seems like a useful thing to do. Plus the person who opened the issue wondering after looking at the list means that saying nothing leaves it potentially confusing for others as well.

Centril:
> If all processes share the same address space, then memory corruption can be cross-process. But at that point, you could consider all programs part of the same process? But also, by the point somebody runs across this condition, they've hopefully been warned about it in the docs on writing for such an OS. And explaining it here would just muddy everyone else's understanding of it?

Yeah I suppose.

> It is more general semantics of inter-programming language stuff than explanation of the semantics of the Rust language. Thus I put it in a _note_. For what's a quick list of UB things to reference, reminding people that UB is not necessarily just in Rust seems like a useful thing to do. Plus the person who opened the issue wondering after looking at the list means that saying nothing leaves it potentially confusing for others as well.

Alright, let's :shipit: .

ubsan:
@centril that's along the lines of "no way to reasonably do anything about that". The way language implementations work is that they assume that the hardware and the OS is not faulty. Anything outside of that and there's no way to do anything about it.

src/behavior-considered-undefined.md
```diff
@@ -39,6 +39,12 @@ code.
   * A value in a `char` which is a surrogate or above `char::MAX`.
   * Non-UTF-8 byte sequences in a `str`.
 
+> **Note**: Undefined behavior affects the entire program. For example, calling
+> a function in `C` that exhibits undefined behavior of `C` means your entire
```
RalfJung: I don't think we usually put C/Rust in `code font`, do we?

RalfJung:
> I think suggesting that UB is bounded at the process level is too nice as it could cause UB in your entire system depending on OS?

Well if your OS doesn't do process separation, sure -- but would we even still call that an OS nowadays?^^

This PR seems fine to me.

src/behavior-considered-undefined.md
Centril: Ping from triage: @Havvy any thoughts on Ralf's question?

Havvy: This is correct. I've de-coded them.

Centril:
Thanks @Havvy!

kpp:
Thanks!

PR #492, opened 2018-12-30
Document underscore imports.

### LINKED ISSUE #470: Document `use path as _;` and `extern crate name as _;`
This is a documentation issue for https://github.com/rust-lang/rust/pull/56303

ehuss:
Closes #470.

This includes a semi-unrelated change to `use` documentation to clarify some changes regarding `use cratename`. I'm not super comfortable with these changes, but I don't see anything obviously wrong.

### DIFF
--- a/src/items/extern-crates.md
+++ b/src/items/extern-crates.md
@@ -83,11 +83,12 @@ by using an underscore with the form `extern crate foo as _`. This may be
 useful for crates that only need to be linked, but are never referenced, and
 will avoid being reported as unused.
 
-The `#[macro_use]` attribute will work as usual and import the macro names
+The [`#[macro_use]` attribute] will work as usual and import the macro names
 into the macro-use prelude.
 
 [IDENTIFIER]: identifiers.html
 [RFC 940]: https://github.com/rust-lang/rfcs/blob/master/text/0940-hyphens-considered-harmful.md
+[`#[macro_use]` attribute]: attributes.html#macro-related-attributes
 [`alloc`]: https://doc.rust-lang.org/alloc/
 [`crate::`]: paths.html#crate
 [`proc_macro`]: https://doc.rust-lang.org/proc_macro/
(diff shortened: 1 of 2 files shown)


src/items/extern-crates.md
```diff
@@ -68,16 +68,24 @@ need for specifying `extern crate`.
...
 -->
 
+## Underscore Imports
+
+An external crate dependency can be declared without binding its name in scope
+by using an underscore with the form `extern crate foo as _`. This may be
+useful for crates that only need to be linked, but are never referenced, and
+will avoid being reported as unused.
+
+The `#[macro_use]` attribute will work as usual and import the macro names
```
Centril: Is there somewhere to hyperlink "as usual" to?

src/items/use-declarations.md
```diff
@@ -149,7 +151,37 @@ fn main() {}
...
+}
+
+use self::foo::Zoo as _;
+struct Zoo;  // Underscore import avoids name conflict with this item.
+
+fn main() {
+    let z = Zoo;
+    z.zoo();
+}
+```
```
Centril: Should we capture this comment somehow? https://github.com/rust-lang/rust/pull/56303#issuecomment-442281975

src/items/extern-crates.md
ehuss: Not really. I added a link to where `macro_use` is described, but it is very brief.  I actually just today learned about the macro-use prelude, and I'm not sure if it is wise to mention it here since it is not described anywhere.  

In general I think it would be amazing to have good documentation on name resolution — something that explained the different namespaces, the different preludes, the precedence and order, macro resolution, maybe ribs?, etc.  I added docs on "extern prelude", but I don't really understand these things.  I think if I did understand them, my mental model of how Rust works would be substantially improved.  I'm not even sure where to start learning it (is everything in librustc_resolve? and macros are mostly in libsyntax?). The current [rustc-guide docs](https://rust-lang.github.io/rustc-guide/name-resolution.html) are kinda incomplete.

Centril:
Looks good; Let's :shipit:

src/items/extern-crates.md
Centril: You should have a Q&A session with @petrochenkov... =P

PR #493, opened 2018-12-30
Clarifications

RadicalZephyr:
These are a few edits to phrases that I found hard to understand on my first reading.  I believe I've preserved the same meaning (to the best of my understanding!).

### DIFF
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -51,7 +51,7 @@ The nominal type is called the _implementing type_ and the associable items are
 the _associated items_ to the implementing type.
 
 Inherent implementations associate the contained items to the implementing type.
-The associated item has a path of a path to the implementing type followed by
+The path of an associated item consists of a path to the implementing type followed by
 the associate item's path component. Inherent implementations cannot contain
 associated type aliases.
 
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -82,10 +82,17 @@ mod values {
 
 pub use self::color::Color;
 fn main() {
-    color::Color::WHITE; // actual path to the implementing type and impl in the same module
-    color::Color::red();  // impl blocks in different modules are still accessed through a path to the type
-    Color::red();  // rexported paths to the implementing type also work
-    // values::Color::red();  // Does not work, because use in `values` is not pub
+    // Actual path to the implementing type and impl in the same module.
+    color::Color::WHITE;
+
+    // Impl blocks in different modules are still accessed through a path to the type.
+    color::Color::red();
+
+    // Re-exported paths to the implementing type also work.
+    Color::red();
+
+    // Does not work, because use in `values` is not pub.
+    // values::Color::red();
 }
 ```
 
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -55,9 +55,9 @@ implementing type.  Inherent implementations can contain [associated
 functions] (including methods) and [associated constants]. They cannot
 contain associated type aliases.
 
-The [path] to an associated item has the following structure: any path
-to the implementing type, followed by the type path to the associate item
-within the inherent implementation.
+The [path] to an associated item is any path to the implementing type,
+followed by the associated item's identifier as the final path
+component.
 
 A type can also have multiple inherent implementations. An implementing type
 must be defined within the same crate as the original type definition.
@@ -209,8 +209,8 @@ attributes].
 [_Visibility_]: visibility-and-privacy.html
 [_WhereClause_]: items/generics.html#where-clauses
 [trait]: items/traits.html
-[associated functions]: associated-items.html#associated-functions-and-methods
-[associated constants]: associated-items.html#associated-constants
+[associated functions]: items/associated-items.html#associated-functions-and-methods
+[associated constants]: items/associated-items.html#associated-constants
 [attributes]: attributes.html
 [`cfg`]: conditional-compilation.html
 [`deprecated`]: attributes.html#deprecation
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -50,10 +50,18 @@ bracketed set of associable items.
 The nominal type is called the _implementing type_ and the associable items are
 the _associated items_ to the implementing type.
 
-Inherent implementations associate the contained items to the implementing type.
-The path to an associated item is: any path to the implementing type followed by
-the associate item's path component. Inherent implementations cannot contain
-associated type aliases.
+Inherent implementations associate the contained items to the
+implementing type.  Inherent implementations can contain [associated
+functions] (including methods), [associated types], and [associated
+constants]. They cannot contain associated type aliases.
+
+[associated functions]: associated-items.html#associated-functions-and-methods
+[associated types]: associated-items.html#associated-types
+[associated constants]: associated-items.html#associated-constants
+
+The path to an associated item is: any path to the implementing type,
+followed by the path to the associate item within the inherent
+implementation.
 
 Note that the path to the module containing the inherent
 implementation does not allow access to the associate item, unless the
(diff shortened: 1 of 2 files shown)


src/items/implementations.md
```diff
@@ -51,7 +51,7 @@ The nominal type is called the _implementing type_ and the associable items are
 the _associated items_ to the implementing type.
 
 Inherent implementations associate the contained items to the implementing type.
-The associated item has a path of a path to the implementing type followed by
+The path of an associated item consists of a path to the implementing type followed by
```
Centril: Hmm... I don't find the current nor the changed wording to be great. In particular "path" gets repeated twice in succession; can we find an alternate wording?

RadicalZephyr: I don't think we can really get away from repeating path here, since it is the proper descriptive word.  This sentence is defining how to construct the path to an item associated with a trait, and part of that path is the path to the implementing type.  IMO using a synonym for path just to avoid the repetition would be even less clear.

That being said, if you have a concrete suggestion on how to improve it I'm all for it :)

Centril: I got nothing... sorry :)

src/items/implementations.md
```diff
@@ -50,10 +50,44 @@ bracketed set of associable items.
 The nominal type is called the _implementing type_ and the associable items are
 the _associated items_ to the implementing type.
 
-Inherent implementations associate the contained items to the implementing type.
-The associated item has a path of a path to the implementing type followed by
-the associate item's path component. Inherent implementations cannot contain
-associated type aliases.
+Inherent implementations associate the contained items to the
+implementing type.  Inherent implementations can contain [associated
+functions] (including methods), [associated types], and [associated
```
Havvy: Associated types are type aliases, and as per the next statement, inherent impls cannot have them.

src/items/implementations.md
```diff
@@ -50,10 +50,44 @@ bracketed set of associable items.
...
+constants]. They cannot contain associated type aliases.
+
+[associated functions]: associated-items.html#associated-functions-and-methods
+[associated types]: associated-items.html#associated-types
+[associated constants]: associated-items.html#associated-constants
+
+The path to an associated item is: any path to the implementing type,
```
Havvy: It would be nice if the first "path" was a link to the chapter on paths. I wouldn't block the PR on it though.

src/items/implementations.md
```diff
@@ -50,10 +50,44 @@ bracketed set of associable items.
...
-Inherent implementations associate the contained items to the implementing type.
-The associated item has a path of a path to the implementing type followed by
-the associate item's path component. Inherent implementations cannot contain
-associated type aliases.
+Inherent implementations associate the contained items to the
+implementing type.  Inherent implementations can contain [associated
+functions] (including methods), [associated types], and [associated
+constants]. They cannot contain associated type aliases.
+
+[associated functions]: associated-items.html#associated-functions-and-methods
```
Havvy: The link refs should be added to the bottom of the page in alphabetical (by ASCII) order.

src/items/implementations.md
```diff
@@ -50,10 +50,44 @@ bracketed set of associable items.
...
+[associated types]: associated-items.html#associated-types
+[associated constants]: associated-items.html#associated-constants
+
+The path to an associated item is: any path to the implementing type,
+followed by the path to the associate item within the inherent
+implementation.
+
+Note that the path to the module containing the inherent
```
Havvy: I'm not quite sure we want to have this note here. It's a restatement of the (visibility chapter)[https://doc.rust-lang.org/nightly/reference/visibility-and-privacy.html].

src/items/implementations.md
```diff
@@ -50,10 +50,44 @@ bracketed set of associable items.
...
+            Color(255, 0, 0)
+        }
+    }
+}
+
+pub use self::color::Color;
+fn main() {
+    color::Color::red();  // actual path to the implementing type
+    Color::red();  // rexported paths to the implementing type also work
+    // bright_theme::Color::red();  // Does not work, because use in `theme` is not pub
```
Havvy: `bright_theme` doesn't exist. Did you mean `values` or did you want to rename module?

src/items/implementations.md
RadicalZephyr: That's fair.  I haven't read the whole reference yet, but it occurred to me as a subtle consequence that was implied by what was previously there.  Maybe a link to the visibility chapter would be more appropriate than restating?

src/items/implementations.md
RadicalZephyr: whoops.  Changed the code sample too many times.  I was going to add another module with a pub use in it, but it ended up feeling like belaboring the point.

src/items/implementations.md
RadicalZephyr: done I think.  Unless I linked to the wrong place.

src/items/implementations.md
RadicalZephyr: Removed.

src/items/implementations.md
RadicalZephyr: I removed the note paragraph and instead called it out in the example by showing side by side that impl blocks in different modules still use a path to the type.

src/items/implementations.md
RadicalZephyr: fixed. and fixed the comment which mentioned yet another previous name for the secondary module.

src/items/implementations.md
```diff
@@ -50,25 +50,43 @@ bracketed set of associable items.
...
 
-```rust
-struct Point {x: i32, y: i32}
+``` rust
+pub mod color {
+    pub struct Color(pub u8, pub u8, pub u8);
 
-impl Point {
-    fn log(&self) {
-        println!("Point is at ({}, {})", self.x, self.y);
```
RadicalZephyr: Reading the whole page again, I decided to remove the old example and just use the larger example I've written since (IMHO) it more clearly demonstrates what the prose above it is discussing.

RadicalZephyr:
@ehuss wondering if you have any feedback on this PR?

src/items/implementations.md
```diff
@@ -50,25 +50,43 @@ bracketed set of associable items.
...
+    impl Color {
+        pub fn red() -> Color {
+            Color(255, 0, 0)
+        }
+    }
+}
+
+pub use self::color::Color;
+fn main() {
+    color::Color::WHITE; // actual path to the implementing type and impl in the same module
```
ehuss: These comments end up requiring one to horizontally scroll to read them.  Can you move each one up to a line on its own?  Also, I would make them sentences (capitalize the first word).

src/items/implementations.md
```diff
@@ -50,25 +50,43 @@ bracketed set of associable items.
...
+    color::Color::red();  // impl blocks in different modules are still accessed through a path to the type
+    Color::red();  // rexported paths to the implementing type also work
```
ehuss: "Re-exported"

src/items/implementations.md
```diff
@@ -50,25 +50,43 @@ bracketed set of associable items.
...
-Inherent implementations associate the contained items to the implementing type.
-The associated item has a path of a path to the implementing type followed by
-the associate item's path component. Inherent implementations cannot contain
-associated type aliases.
+Inherent implementations associate the contained items to the
+implementing type.  Inherent implementations can contain [associated
+functions] (including methods) and [associated constants]. They cannot
+contain associated type aliases.
+
+The [path] to an associated item has the following structure: any path
```
ehuss:
```suggestion
The [path] to an associated item is any path
```

src/items/implementations.md
```diff
@@ -50,25 +50,43 @@ bracketed set of associable items.
...
+to the implementing type, followed by the type path to the associate item
```
ehuss:
```suggestion
to the implementing type, followed by the associated item's identifier as the final path component.
```

Type paths are a specific concept that I don't think is correct here.

Centril:
Looks good; thanks!

PR #494, opened 2018-12-31
Document Rc/Arc method receivers.

ehuss:
Stabilization PR: https://github.com/rust-lang/rust/pull/56805
Stabilization issue: https://github.com/rust-lang/rust/issues/55786

It sounds like this may still change (https://github.com/rust-lang/rust/issues/55786#issuecomment-450254059), but I figured it wouldn't hurt to get the docs started. Since `Receiver` and `Pin` are still unstable, they are not mentioned.

### DIFF
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -93,9 +93,34 @@ Associated functions whose first parameter is named `self` are called *methods*
 and may be invoked using the [method call operator], for example, `x.foo()`, as
 well as the usual function call notation.
 
-If the type of the `self` parameter is specified, it is limited to the type
-being implemented (or `Self`), or a reference or mutable reference to the
-type, or a boxed value of the type being implemented (such as `Box<Self>`).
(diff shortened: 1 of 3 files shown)


matthewjasper:
`Pin` *is* stable though: https://doc.rust-lang.org/nightly/src/core/pin.rs.html#117

src/items/associated-items.md
```diff
@@ -93,9 +93,34 @@ Associated functions whose first parameter is named `self` are called *methods*
...
+
+- `Self`
+- `&Self`
+- `&mut Self`
+- [`Box<Self>`]
+- [`Rc<Self>`]
+- [`Arc<Self>`]
+
+The `Self` portion of the type may be replaced with the type being
+implemented.
```
matthewjasper: Hmm, this feels a bit of an awkward way to say "the way type of the receiver is written doesn't matter", but I'm not sure that's any better.

Centril:
Ping from triage, @Havvy do you think you'll be able to review this?

PR #495, opened 2018-12-31
Use mdbook release binary

RadicalZephyr:
Travis has pretty good support for caching rust builds now by store the .cargo directory.  This would enable this build to rebuild mdbook every time.

From [my test], it looks like the average build time will go from ~10 minutes to ~1 minute.

The one potential downside to this is mdbook won't be rebuilt with new nightly versions of Rust.  Presumably this doesn't matter though since the version of mdbook is pinned, and upgrading it would require running one build with `cargo install --force --version =<new-version>`, and then removing the `--force` again (otherwise it would always rebuild and thus remove the point of caching it). It doesn't seem like the version of mdbook is getting updated regularly though so this seems like it shouldn't be too big of a burden.

[my test]: https://travis-ci.com/RadicalZephyr/reference/builds/95976952

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: rust
 rust:
   - nightly
 
-before_script:
-  - (cargo install mdbook --version =0.1.7)
+install:
+  - travis_retry curl -Lf https://github.com/rust-lang-nursery/mdBook/releases/download/v0.1.7/mdbook-v0.1.7-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=$HOME/.cargo/bin
 
-script: 
+script:
   - export PATH=$PATH:/home/travis/.cargo/bin && mdbook test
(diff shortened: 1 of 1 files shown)


RadicalZephyr:
I just realized that it might be really important to build a new version of mdbook with each nightly to ensure that the code examples in the reference are being run against the new nightly, I'm not familiar with it's necessary to rebuild mdbook to make that happen.  If that's the case feel free to just close this :)

ehuss:
I recently swapped cargo to use a pre-built package instead: https://github.com/rust-lang/cargo/pull/6443.  I think that works better than caching (much easier to change versions).  mdbook shells out to rustdoc to run tests, so a single mdbook works with different toolchains.

RadicalZephyr:
Ah, that makes a lot of sense. Definitely easier to change :)

ehuss:
Thanks!

PR #498, opened 2019-01-03
"Improve language in 'Constant Expressions'.

### LINKED ISSUE #497: Language in constant evaluation section is unclear
This text from [**Constant Evaluation**: Constant Expressions][ce-ce] is a bit murky:

[ce-ce]: https://doc.rust-lang.org/reference/const_eval.html#constant-expressions

> Certain types of expressions can be evaluated at compile time. These are called constant expressions and are required in const contexts. In other places, such as in [`let` statements], constant expressions may be evaluated at compile time. If errors, such as out of bounds [array indexing] or [overflow] occurs, then it is a compiler error if the value must be evaluated at compile time, otherwise it is just a warning, but the code will most likely panic when run.
> 
> [`let` statements]: 
https://doc.rust-lang.org/reference/statements.html#let-statements
> [array indexing]: 
https://doc.rust-lang.org/reference/expressions/array-expr.html#array-and-slice-indexing-expressions
> [overflow]: 
https://doc.rust-lang.org/reference/expressions/operator-expr.html#overflow

Suggested rephrasing:

> Certain types of expressions, called constant expressions, can be evaluated at compile time. In [const contexts], these are the only allowed expressions, and are always evaluated at compile time. In other places, such as in [`let` statements], constant expressions are allowed but not required, and *may* be evaluated at compile time. Errors such as out of bounds [array indexing] or [overflow] are compiler errors if the value must be evaluated at compile time (i.e. in const contexts); otherwise they are just warnings, but the code will most likely panic when run.
>
> [const contexts]: https://doc.rust-lang.org/reference/const_eval.html#const-context
> [`let` statements]: 
https://doc.rust-lang.org/reference/statements.html#let-statements
> [array indexing]: 
https://doc.rust-lang.org/reference/expressions/array-expr.html#array-and-slice-indexing-expressions
> [overflow]: 
https://doc.rust-lang.org/reference/expressions/operator-expr.html#overflow

Happy to submit a PR to that effect or similar!

chriskrycho:
Fixes #497.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -10,7 +10,7 @@ Certain forms of expressions, called constant expressions, can be evaluated at
 compile time. In [const contexts](#const-context), these are the only allowed
 expressions, and are always evaluated at compile time. In other places, such as
 [`let` statements](statements.html#let-statements), constant expressions are
-Errors such as out of bounds [array indexing] or [overflow] are compiler errors
+Behaviors such as out of bounds [array indexing] or [overflow] are compiler errors
 if the value must be evaluated at compile time (i.e. in const contexts).
 Otherwise, these behaviors are warnings, but will likely panic at run-time.
 
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -12,7 +12,7 @@ expressions, and are always evaluated at compile time. In other places, such as
 [`let` statements](statements.html#let-statements), constant expressions are
 Errors such as out of bounds [array indexing] or [overflow] are compiler errors
 if the value must be evaluated at compile time (i.e. in const contexts);
-otherwise they are just warnings, but the code will most likely panic when run.
+Otherwise, these behaviors are warnings, but will likely panic at run-time.
 
 The following expressions are constant expressions, so long as any operands are
 also constant expressions and do not cause any [`Drop::drop`][destructors] calls
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -6,7 +6,7 @@ can be evaluated at compile-time.
 
 ## Constant expressions
 
-Certain types of expressions, called constant expressions, can be evaluated at
+Certain forms of expressions, called constant expressions, can be evaluated at
 compile time. In [const contexts](#const-context), these are the only allowed
 expressions, and are always evaluated at compile time. In other places, such as
 in [`let` statements](statements.html#let-statements), constant expressions are
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
 
 ## Constant expressions
 
-Certain types of expressions can be evaluated at compile time. These are called
-_constant expressions_ and are required in const contexts. In
-other places, such as in [`let` statements](statements.html#let-statements),
-constant expressions may be evaluated at compile time. If errors, such as out
-of bounds [array indexing] or [overflow] occurs,
-then it is a compiler error if the value must be evaluated at compile time,
-otherwise it is just a warning, but the code will most likely panic when run.
+Certain types of expressions, called constant expressions, can be evaluated at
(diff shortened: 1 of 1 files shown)


src/const_eval.md
```diff
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
...
 ## Constant expressions
 
-Certain types of expressions can be evaluated at compile time. These are called
-_constant expressions_ and are required in const contexts. In
-other places, such as in [`let` statements](statements.html#let-statements),
-constant expressions may be evaluated at compile time. If errors, such as out
-of bounds [array indexing] or [overflow] occurs,
-then it is a compiler error if the value must be evaluated at compile time,
-otherwise it is just a warning, but the code will most likely panic when run.
+Certain types of expressions, called constant expressions, can be evaluated at
```
Centril:
```suggestion
Certain forms of expressions, called constant expressions, can be evaluated at
```
to not confuse with expression typing...

src/const_eval.md
```diff
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
...
+compile time. In [const contexts](#const-context), these are the only allowed
+expressions, and are always evaluated at compile time. In other places, such as
+in [`let` statements](statements.html#let-statements), constant expressions are
```
Centril:
```suggestion
[`let` statements](statements.html#let-statements), constant expressions are
```

src/const_eval.md
```diff
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
...
+Errors such as out of bounds [array indexing] or [overflow] are compiler errors
```
Centril: I think:
> allowed but not required, and may be evaluated at compile time.

accidentally fell off?

src/const_eval.md
```diff
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
...
+if the value must be evaluated at compile time (i.e. in const contexts);
+otherwise they are just warnings, but the code will most likely panic when run.
```
Centril:
```suggestion
Otherwise, these behaviors are warnings, but will likely panic when run.
```

src/const_eval.md
```diff
@@ -6,13 +6,13 @@ can be evaluated at compile-time.
...
-constant expressions may be evaluated at compile time. If errors, such as out
-of bounds [array indexing] or [overflow] occurs,
-then it is a compiler error if the value must be evaluated at compile time,
-otherwise it is just a warning, but the code will most likely panic when run.
+Certain types of expressions, called constant expressions, can be evaluated at
+compile time. In [const contexts](#const-context), these are the only allowed
+expressions, and are always evaluated at compile time. In other places, such as
+in [`let` statements](statements.html#let-statements), constant expressions are
+Errors such as out of bounds [array indexing] or [overflow] are compiler errors
+if the value must be evaluated at compile time (i.e. in const contexts);
```
Centril:
```suggestion
if the value must be evaluated at compile time (i.e. in const contexts).
```

src/const_eval.md
Centril: Also, given that constant expressions are a strict subset of all expressions, is "allowed but not required" not implied?

Centril:
```suggestion
Behaviors such as out of bounds [array indexing] or [overflow] are compiler errors
```

chriskrycho: Ah, good catch on the missing bit; I'll push a change with that and also clarifying the other language.

chriskrycho: And, there we go.

chriskrycho:
Yeah, I realized I was more motivated to push this than to clean the dishes! 😆

Centril:
Haha; I sympathize with not cleaning the dishes ;)

This turned out great; thanks!

chriskrycho:
My pleasure!

PR #502, opened 2019-01-09
Update conditional-compilation.md

jethrogb:


nagisa:
IMHO you could just replace "possible values" with "Example values" to match the other target cfgs. This is the more appropriate way to do it *either* way.

nagisa:
Erhm, I wrote that comment above before double-checking the PR. It seems you did exactly that!

PR #511, opened 2019-01-15
Expand docs on Macros By Example.

alercah:
The primary motivation here was to increase clarity and fully address the
scoping and naming details. The inclusion of RFC 550's formal specification is
to move it to the reference where it can be updated. I made several changes,
motivated by accommodating `?` and new fragment specifiers, but there are some
other things which need highlighting so that they can be double-checked for
correctness.

  * Permit the empty string to follow on in the first invariant; this is a
    technical oversight in the definition I believe.
  * Added a requirement that repetitions obey the follow rules; this was an
    oversight in the original RFC and currently planned for fix.
  * Rewrote the definition of FIRST for complex NTs to be more clear.
  * Added a case to LAST for `?` repetitions
  * Removed the last example of LAST, because it is wrong.
  * Rearranged the definition of FOLLOW to be more clear
  * Added Shl to FOLLOW(ty) and FOLLOW(path), as documented in the Reference
    already.
  * Added missing follow sets for newer fragment specifiers.

The scoping text is probably not completely accurate, but it's certainly much
better than what was there before (i.e. basically nothing).

### DIFF
--- a/src/macro-ambiguity.md
+++ b/src/macro-ambiguity.md
@@ -118,7 +118,7 @@ legal macro definition will continue to assign the same determination as to
 where `... tt` ends and `uu ...` begins, even as new syntactic forms are added
 to the language.
 
-The second invariant says that a separated complex NT must use a seperator token
+The second invariant says that a separated complex NT must use a separator token
 that is part of the predetermined follow set for the internal contents of the
 NT. This ensures that a legal macro definition will continue to parse an input
 fragment into the same delimited sequence of `tt ...`'s, even as new syntactic
@@ -273,7 +273,7 @@ LAST(M), defined by case analysis on M itself (a sequence of token-trees):
 
 Below are some examples of FIRST and LAST.
 (Note in particular how the special ε element is introduced and
-eliminated based on the interation between the pieces of the input.)
+eliminated based on the interaction between the pieces of the input.)
 
 Our first example is presented in a tree structure to elaborate on how
 the analysis of the matcher composes. (Some of the simpler subtrees
@@ -365,7 +365,7 @@ why particular matchers are legal and others are not.
 
  * `( $($a:tt $b:tt)* ; )` : legal, because FIRST(`$b:tt`) = { `$b:tt` } is ⊆ FOLLOW(`tt`) = ANYTOKEN, as is FIRST(`;`) = { `;` }.
 
- * `( $($t:tt),* , $(t:tt),* )` : legal,  (though any attempt to actually use this macro will signal a local ambguity error during expansion).
+ * `( $($t:tt),* , $(t:tt),* )` : legal,  (though any attempt to actually use this macro will signal a local ambiguity error during expansion).
 
  * `($ty:ty $(; not sep)* -)` : illegal, because FIRST(`$(; not sep)* -`) = { `;`, `-` } is not in FOLLOW(`ty`).
 
--- a/src/macro-ambiguity.md
+++ b/src/macro-ambiguity.md
@@ -1,0 +1,378 @@
+# Appendix: Macro Follow-Set Ambiguity Formal Specification
+
+This page documents the formal specification of the follow rules for [Macros
+By Example]. They were originally specified in [RFC 550], from which the bulk
+of this text is copied, and expanded upon in subsequent RFCs.
+
+## Definitions & Conventions
+
+  - `macro`: anything invokable as `foo!(...)` in source code.
+  - `MBE`: macro-by-example, a macro defined by `macro_rules`.
+  - `matcher`: the left-hand-side of a rule in a `macro_rules` invocation, or a
+    subportion thereof.
+  - `macro parser`: the bit of code in the Rust parser that will parse the
+    input using a grammar derived from all of the matchers.
+  - `fragment`: The class of Rust syntax that a given matcher will accept (or
+    "match").
+  - `repetition` : a fragment that follows a regular repeating pattern
+  - `NT`: non-terminal, the various "meta-variables" or repetition matchers
+	that can appear in a matcher, specified in MBE syntax with a leading `$`
+	character.
+  - `simple NT`: a "meta-variable" non-terminal (further discussion below).
+  - `complex NT`: a repetition matching non-terminal, specified via repetition
+    operators (`\*`, `+`, `?`).
+  - `token`: an atomic element of a matcher; i.e. identifiers, operators,
+    open/close delimiters, *and* simple NT's.
+  - `token tree`: a tree structure formed from tokens (the leaves), complex
+    NT's, and finite sequences of token trees.
+  - `delimiter token`: a token that is meant to divide the end of one fragment
+    and the start of the next fragment.
+  - `separator token`: an optional delimiter token in an complex NT that
+    separates each pair of elements in the matched repetition.
+  - `separated complex NT`: a complex NT that has its own separator token.
+  - `delimited sequence`: a sequence of token trees with appropriate open- and
+    close-delimiters at the start and end of the sequence.
+  - `empty fragment`: The class of invisible Rust syntax that separates tokens,
+    i.e. whitespace, or (in some lexical contexts), the empty token sequence.
+  - `fragment specifier`: The identifier in a simple NT that specifies which
+    fragment the NT accepts.
+  - `language`: a context-free language.
+
+Example:
+
+```rust,compile_fail
+macro_rules! i_am_an_mbe {
+    (start $foo:expr $($i:ident),* end) => ($foo)
+}
+```
+
+`(start $foo:expr $($i:ident),\* end)` is a matcher. The whole matcher is a
+delimited sequence (with open- and close-delimiters `(` and `)`), and `$foo`
+and `$i` are simple NT's with `expr` and `ident` as their respective fragment
+specifiers.
+
+`$(i:ident),\*` is *also* an NT; it is a complex NT that matches a
+comma-separated repetition of identifiers. The `,` is the separator token for
+the complex NT; it occurs in between each pair of elements (if any) of the
+matched fragment.
+
+Another example of a complex NT is `$(hi $e:expr ;)+`, which matches any
+fragment of the form `hi <expr>; hi <expr>; ...` where `hi <expr>;` occurs at
+least once. Note that this complex NT does not have a dedicated separator
+token.
+
+(Note that Rust's parser ensures that delimited sequences always occur with
+proper nesting of token tree structure and correct matching of open- and
+close-delimiters.)
+
+We will tend to use the variable "M" to stand for a matcher, variables "t" and
+"u" for arbitrary individual tokens, and the variables "tt" and "uu" for
+arbitrary token trees. (The use of "tt" does present potential ambiguity with
+its additional role as a fragment specifier; but it will be clear from context
+which interpretation is meant.)
+
+"SEP" will range over separator tokens, "OP" over the repetition operators
+`\*`, `+`, and `?`, "OPEN"/"CLOSE" over matching token pairs surrounding a
+delimited sequence (e.g. `[` and `]`).
+
+Greek letters "α" "β" "γ" "δ"  stand for potentially empty token-tree sequences.
+(However, the Greek letter "ε" (epsilon) has a special role in the presentation
+and does not stand for a token-tree sequence.)
+
+  * This Greek letter convention is usually just employed when the presence of
+	a sequence is a technical detail; in particular, when we wish to *emphasize*
+	that we are operating on a sequence of token-trees, we will use the notation
+	"tt ..." for the sequence, not a Greek letter.
+
+Note that a matcher is merely a token tree. A "simple NT", as mentioned above,
+is an meta-variable NT; thus it is a non-repetition. For example, `$foo:ty` is
+a simple NT but `$($foo:ty)+` is a complex NT.
+
+Note also that in the context of this formalism, the term "token" generally
+*includes* simple NTs.
+
+Finally, it is useful for the reader to keep in mind that according to the
+definitions of this formalism, no simple NT matches the empty fragment, and
+likewise no token matches the empty fragment of Rust syntax. (Thus, the *only*
+NT that can match the empty fragment is a complex NT.) This is not actually
+true, because the `vis` matcher can match an empty fragment. Thus, for the
+purposes of the formalism, we will treat `$v:vis` as actually being
+`$($v:vis)?`, with a requirement that the matcher match an empty fragment.
+
+### The Matcher Invariants
+
+To be valid, a matcher must meet the following three invariants. The definitions
+of FIRST and FOLLOW are described later.
+
+1.  For any two successive token tree sequences in a matcher `M` (i.e. `M = ...
+    tt uu ...`) with `uu ...` nonempty, we must have FOLLOW(`... tt`) ∪ {ε} ⊇
+    FIRST(`uu ...`).
+1.  For any separated complex NT in a matcher, `M = ... $(tt ...) SEP OP ...`,
+	we must have `SEP` ∈ FOLLOW(`tt ...`).
+1.  For an unseparated complex NT in a matcher, `M = ... $(tt ...) OP ...`, if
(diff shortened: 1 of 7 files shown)


src/macro-ambiguity.md
```diff
@@ -0,0 +1,375 @@
...
+
+Note that a matcher is merely a token tree. A "simple NT", as mentioned above,
+is an meta-variable NT; thus it is a non-repetition. For example, `$foo:ty` is
+a simple NT but `$($foo:ty)+` is a complex NT.
+
+Note also that in the context of this formalism, the term "token" generally
+*includes* simple NTs.
+
+Finally, it is useful for the reader to keep in mind that according to the
+definitions of this formalism, no simple NT matches the empty fragment, and
```
petrochenkov: `vis` matches the empty fragment.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
-* `expr`: an [_Expression_]
-* `ty`: a [_Type_]
-* `ident`: an [IDENTIFIER_OR_KEYWORD]
-* `path`: a [_TypePath_] style path
-* `tt`: a [_TokenTree_]&nbsp;(a single [token] or tokens in matching delimiters `()`, `[]`, or `{}`)
-* `meta`: a [_MetaItem_], the contents of an attribute
-* `lifetime`: a [LIFETIME_TOKEN]
-* `vis`: a [_Visibility_] qualifier
-* `literal`: matches `-`<sup>?</sup>[_LiteralExpression_]
+Each macro by example has a name, and one or more _rules_. Each ruleh as two
```
petrochenkov: Typo: "ruleh as" -> "rule has"

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+parts: a _matcher_, describing the syntax that it matches, and a _transcriber_,
+describing the syntax that will replace a successfully matched invocation. Both
+the matcher and the transcriber must be surrounded by delimiters. Macros can
+expand to expressions, statements, items, types, or patterns.
```
petrochenkov: Also trait/impl/foreign items.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+
+When a macro is invoked, the macro expander looks up macro invocations by name,
+and tries each macro rule in turn. It transcribes the first successful match; if
+this results in an error, then future matches are not tried. When matching, no
+lookahead is performed; if the compiler cannot umambiguously determine how to
```
petrochenkov: Typo: "umambiguously"

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+parse the macro invocation one token at a time, then it is an error. In the
```
petrochenkov: I'm not sure whether "one token at time" includes https://github.com/rust-lang/rust/pull/42913 or not.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+
+## Metavariables
+
+In the matcher, `$` _name_ `:` _fragment-specifier_ matches a Rust syntax
+fragment of the kind specified and binds it to the metavariable `$`_name_. Valid
+fragment specifiers are:
+
+  * `item`: an [_Item_]
+  * `block`: a [_BlockExpression_]
+  * `stmt`: a [_Statement_] without the trailing semicolon
```
petrochenkov: Unless it's an item statement and the trailing semicolon is a part of the item syntax.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+  * `vis`: a [_Visibility_] qualifier
+  * `literal`: matches `-`<sup>?</sup>[_LiteralExpression_]
+
+In the transcriber, metavariables are referred to simply by `$`_name_`, since
+the fragment kind is specified in the matcher. Metavariables are replaced with
+the syntax element that matched them. The keyword metavariable `$crate` can be
+used to refer to the current crate; see [Hygiene] below. Metavariables can be
+transcribed more than once or not at all.
+
+## Reptitions
```
petrochenkov: Typo: "Reptitions" :snake:

petrochenkov: There are few instances of the same typo below.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+does not have the same number. This requirement applies to every layer of nested
+repetitions.
+
+> **Edition Differences**: The `?` repetition operator did not exist before the
+> 2018 edition. Prior to the 2018 Edition, `?` was an allowed
+> separator token, rather than a repetition operator.
+
+## Scoping, Exporting, and Importing
+
+For historical reasons, the scoping of macros by example does not work entirely like
```
petrochenkov: Or rather works almost entirely not like items.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+    (lazy) => {};
+}
+
+lazy_static!{lazy} // Textual lookup finds our macro first.
+self::lazy_static!{} // Path-based lookup ignores our macro, finds imported one.
+```
+
+### Textual Scope
+
+Textual scope is based largely on the order that things appear in source files.
```
petrochenkov: IMO, It's important to draw an analogy with `let` since most of users have intuition about how `let` works, and `macro_rules` behave like "`let` at module level".

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+For historical reasons, the scoping of macros by example does not work entirely like
+items. Macros have two forms of scope: textual scope, and path-based scope.
+Textual scope is based on the order that things appear in source files, or even
+across multiple files, and is the default scoping. It's explained further below.
+Path-based scope works exactly the same way that item scoping does. The scoping,
+exporting, and importing of macros is controlled largely by attributes.
+
+When a macro is invoked by an unqualified identifier (not part of a multi-part
+path), it's first looked up in textual scoping. If this does not yield any
+results, then it is looked up in path-based scoping. If the macro's name is
```
petrochenkov: This is not entirely true, "items" in inner scopes can shadow `macro_rules` in outer scopes, similarly to `let`:
```rust
fn f() {
    let S = 10;
    {
        struct S;

        let s = S; // refers to the struct
    }
}
```
This is not fully implemented for `macro_rules` though, because it's not high priority and I didn't have enough time, so for `macro_rules` you mostly get ambiguity errors in these situations.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+        () => {};
+    }
+}
+
+m!();
+```
+
+Second, it can be used to import macros from another crate, by attaching it to
+an  `extern crate` declaration. Macros imported this way are imported textually
+only. Optionally, a list of macros to import can be specified; this is not
```
petrochenkov: `#[macro_use]` on `extern crate` doesn't work textually (e.g. you can use the macro textually before the `extern crate` item).
It adds the macro into prelude instead.
(Also the `extern crate` item must be in the crate root, otherwise `#[macro_use]` is an error.)

The only "textual" aspect here is that in
```rust
#[macro_use]
extern crate a;
#[macro_use]
extern crate b;
```
prelude entries introduced by `b` will override entries previously introduced by `a`.

This is quite unfortunate, but understandable from backward-compatibility point of view - before Rust 1.29 if you had and ambiguity between macros provided by `a` and `b` there was no way to explicitly disambiguate, so we couldn't report ambiguity errors in this situation.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+macro_rules! call_foo {
+    () => { $crate::foo() };
+}
+
+fn foo() {}
+```
+
+> **Version & Edition Differences**: Prior to Rust 1.30, `$crate` and
+> `local_inner_macros` (below) were unsupported. They were added in
+> preparation for Rust 2018's path-based imports of macros (described above), in
```
petrochenkov: Path-based imports are not specific to Rust 2018, it's just "before Rust 1.30" and "after Rust 1.30".

petrochenkov:
Reviewed.
Great work.

nikomatsakis:
Not sure what they are up to these days, but @paulstansifer was the original author of the algorithm we use and I don't think the approach has changed much. There is at least some chance they have the time to review this PR =)

src/macro-ambiguity.md
nikomatsakis: I don't think we discussed that in particular, no.

nikomatsakis: Er, sorry, I misread as "implications on the running time" -- in any case, I'm not sure precisely what implications @alercah has in mind. (I haven't read this document.) That said, the behavior of the `vis` matcher is stable, I believe? (And quite useful.)

alercah: I am not sure what, if any, there are either. But the algorithm pretty explicitly assumes that fragment specifiers can't match empty token sequences, so that makes me think that it ought to have been considered before stabilization. ;)

(I had a related follow-up on the tracking issue about `crate`; I don't know if you saw that? `vis` has possibly too broad of a follow set)

Centril: > That said, the behavior of the `vis` matcher is stable, I believe? (And quite useful.)

Yup.

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+  * If the reptition can match zero times (`*` or `?`), then whatever comes
+    after must be able to follow whatever comes before.
+
+
+For more detail, see the [formal specification].
 
 [IDENTIFIER]: identifiers.html
 [IDENTIFIER_OR_KEYWORD]: identifiers.html
 [LIFETIME_TOKEN]: tokens.html#lifetimes-and-loop-labels
+[formal specification]: macro-ambiguity.md
```
ehuss: This should have the `.html` extension.

src/macro-ambiguity.md
```diff
@@ -0,0 +1,375 @@
...
+
+ * `( $($t:tt),* , $(t:tt),* )` : legal,  (though any attempt to actually use this macro will signal a local ambguity error during expansion).
+
+ * `($ty:ty $(; not sep)* -)` : illegal, because FIRST(`$(; not sep)* -`) = { `;`, `-` } is not in FOLLOW(`ty`).
+
+ * `($($ty:ty)-+)` : illegal, because separator `-` is not in FOLLOW(`ty`).
+
+ * `($($e:expr)*)` : illegal, because expr NTs are not in FOLLOW(expr NT).
+
+[Macros by Example]: macros-by-example.md
```
ehuss: This should have a `.html` extension.

alexreg:
It looks like the macro guru @petrochenkov has already reviewed, so my input probably isn't needed -- but in brief, good work. :-)

src/macros-by-example.md
alercah: If it's not fully implemented, I'm not going to document it I think

src/macros-by-example.md
```diff
@@ -24,149 +24,404 @@
...
+across multiple files, and is the default scoping. It's explained further below.
+Path-based scope works exactly the same way that item scoping does. The scoping,
+exporting, and importing of macros is controlled largely by attributes.
+
+When a macro is invoked by an unqualified identifier (not part of a multi-part
+path), it's first looked up in textual scoping. If this does not yield any
+results, then it is looked up in path-based scoping. If the macro's name is
+qualified with a path, then it is only looked up in path-based scoping.
+
+```ignore
```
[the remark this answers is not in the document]
alercah: mdbook and rustdoc don't need `rust` I thought? They're definitely tested.

src/macro-ambiguity.md
alercah: Okay, I realized that this works fine as long as we just define the algorithm to treat it like `$($v:vis)?`---now that semantics for `?` are defined, possibly-empty matches for metavariables should pose no problem. So I put that in as such.

src/macros-by-example.md
alercah: Almost, but not quite, entirely unlike... ;)

They do work sort of like items for path-based scoping, though, and I'd like to see us move in that direction. So I won't include this. ;)

src/macros-by-example.md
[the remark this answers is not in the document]
ehuss: Oh, that's interesting. It's not highlighted, but it is tested.  It looks like mdbook just shells out to `rustdoc` and passes it the raw markdown file.  But mdbook's renderer treats code blocks differently from rustdoc.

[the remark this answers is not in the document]
alercah: Huh. Fixed.

src/macros-by-example.md
```diff
@@ -24,149 +24,428 @@
...
+    }
+}
+
+m!();
+```
+
+Second, it can be used to import macros from another crate, by attaching it to
+an  `extern crate` declaration appearing in the crate's root module. Macros
+imported this way are imported into the prelude of the crate, not textually,
+which means that they can be uare shadowed by any other name. While macros
```
petrochenkov: Typo: uare -> ???

src/macros-by-example.md
```diff
@@ -24,149 +24,428 @@
...
+imported by `#[macro_use]` can be used before the import statement, in case of a
+conflict, the last macro imported wins. Optionally, a list of macros to import
+can be specified; this is not supported when `#[macro_use]` is applied to a
```
Centril: We're documenting the rules as-is here so this is irrelevant for the reference, but is there a reason for this restriction? cc @petrochenkov

jethrogb:
Can you add something along these lines?

> When forwarding a matched fragment to another macro-by-example, matchers in the second macro will see an opaque AST of the fragment type. The second macro can't use literal tokens to match this in the match, only a fragment specifier of the same type. The `ident`, `lifetime`, and `tt` fragment types are an exception, and *can* be matched by literal tokens.

src/macros-by-example.md
petrochenkov: I wish `#[macro_use]` on `mod` was still named `#[macro_escape]` to avoid unnecessary analogies.
`#[macro_escape(name)]` would be a new feature rather than lifted restriction.
I suspect the most common answer to the question "why is feature X not in the language" applies here as well - nobody proposed and implemented it.

Centril: Ah; I thought there was some technical reason rather than a slip... since there doesn't seem to be, I think it would be good to simply fix this as an "omission" so that we can remove this line from the reference and simplify the language...

Centril:
Ping @alercah, there are some unaddressed review comments but otherwise this looks real good :)

alercah:
I'm unable to work on this at the moment, someone else is welcome to bring it home or else I will get to it when I can.

ehuss:
I'd be happy to address the final bits of the review. I think this is great content, and would like to see it published.  I have rebased and pushed a small update.

src/macros-by-example.md
```diff
@@ -24,149 +24,475 @@
...
+transcribed literally, with one exception. The exception is that the outer
+delimiters for the matcher will match any pair of delimiters. Thus, for
+instance, the matcher `(())` will match `{()}` but not `{{}}`. The character
+`$` cannot be matched or transcribed literally.
+
+When forwarding a matched fragment to another macro-by-example, matchers in
+the second macro will see an opaque AST of the fragment type. The second macro
+can't use literal tokens to match the fragments in the matcher, only a
+fragment specifier of the same type. The `ident`, `lifetime`, and `tt`
+fragment types are an exception, and can be matched by literal tokens. The
```
jethrogb: Just wanted to make sure you intentionally removed the emphasis on “can” here.

paulstansifer:
@nikomatsakis Thanks for thinking of me! I'm too busy to do anything other than skim this and say that it looks like a great resource. I also couldn't vet the technical stuff too well; the stuff with `FIRST` and `FOLLOW` is from after my time; I was pleasantly surprised when the compiler sprung it on me.

src/macros-by-example.md
ehuss: Nope, was a copy/paste mistake.  Wish it was easier to grab the raw markdown in github.

ehuss:
Merging so I can use it, thanks everyone!

PR #512, opened 2019-01-15
Let bindings are now available in const contexts

### LINKED ISSUE #506: Document `const_let`
Stabilization PR: https://github.com/rust-lang/rust/pull/57175
RFC: https://github.com/rust-lang/rfcs/pull/2341
Tracking: https://github.com/rust-lang/rust/issues/48821
Target = 1.33

oli-obk:
fixes #506

r? @Centril

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -9,7 +9,7 @@ can be evaluated at compile-time.
 Certain forms of expressions, called constant expressions, can be evaluated at
 compile time. In [const contexts](#const-context), these are the only allowed
 expressions, and are always evaluated at compile time. In other places, such as
-[`let` statements](statements.html#let-statements), constant expressions *may*
+[let statement]s, constant expressions *may*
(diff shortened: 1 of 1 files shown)


src/const_eval.md
```diff
@@ -40,6 +40,7 @@ to be ran.
 * [Cast] expressions, except pointer to address and
   function pointer to address casts.
 * Calls of const functions and const methods
+* [let statement]s and thus irrefutable [patterns]
```
Centril: We should mention the restriction wrt. `&&` here. (also end sentence with punctuation for consistency, and the bullet above...).

Centril: Also is this in the right place? `let` is not an expression...

Centril:
I think this also misses:

* assignment (e.g. `x = y`) and assignment operator (e.g. `x += y`) expressions, even where the assignment target is a projection (e.g. a struct field or index operation like `x[3] = 42`)
* expression statements (e.g. `3;`)

PR #514, opened 2019-01-19
Document drop scopes

### LINKED ISSUE #452: Explicitly describe temporary drop order difference between `return expr;` and final block expression
Due to temporaries in the final expression of a block living longer than the block expression itself, there's actually a difference between a return expression and the final expression in a block. This should be explicitly documented so that we can link to people about it.

Associated Rust issue: https://github.com/rust-lang/rust/issues/21114

matthewjasper:
Closes #452

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -278,7 +278,8 @@ scope of the initializer expression is extended.
 
 #### Extending based on expressions
 
-For a let statement with an initializer, an *extending expression* if it is
+For a let statement with an initializer, an *extending expression* is an
+expression which is one of the following:
 
 * The initializer expression.
 * The operand of a extending [borrow expression].
@@ -342,8 +343,9 @@ variable or field from being dropped automatically.
 [initialized]: glossary.md#initialized
 [interior mutability]: interior-mutability.md
 [lazy boolean expression]: expressions/operator-expr.md#lazy-boolean-operators
+[place context]: expressions.md#place-expressions-and-value-expressions
 [statement]: statements.md
-[temporary]: expressions.md#temporary-lifetimes
+[temporary]: expressions.md#temporaries
 [variable]: variables.md
 
 [array]: types/array.md
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -10,8 +10,9 @@ The destructor of a type `T` consists of:
 
 1. If `T: Drop`, calling [`<T as std::ops::Drop>::drop`]
 2. Recursively running the destructor of all of its fields.
-    * The fields of a [struct] or [tuple] are dropped in declaration order.
+    * The fields of a [struct] are dropped in declaration order.
     * The fields of the active [enum variant] are dropped in declaration order.
+    * The fields of a [tuple] are dropped in order.
     * The elements of an [array] or owned [slice] are dropped from the
       first element to the last.
     * The variables that a [closure] captures by move are dropped in an
@@ -71,7 +72,8 @@ Given a function, or closure, there are drop scopes for:
 * Each [statement]
 * Each [expression]
 * Each block, including the function body
-    * [Block expressions] the expression and block scopes are the same scope.
+    * In the case of a [block expression], the scope for the block and the
+      expression are the same scope.
 * Each arm of a `match` expression
 
 Drop scopes are nested within one another as follows. When multiple scopes are
@@ -82,13 +84,14 @@ from the inside outwards.
 * The function body block is contained within the scope of the entire function.
 * The parent of the expression in an expression statement is the scope of the
   statement.
-* The parent of the initializer of a `let` statement is the `let`
-  statement's scope.
+* The parent of the initializer of a [`let` statement] is the `let` statement's
+  scope.
 * The parent of a statement scope is the scope of the block that contains the
   statement.
 * The parent of the expression for a `match` guard is the scope of the arm that
   the guard is for.
-* The parent of the expression for a given `match` arm is that arm's scope.
+* The parent of the expression after the `=>` in a `match` expression is the
+  scope of the arm that it's in.
 * The parent of the arm scope is the scope of the `match` expression that it
   belongs to.
 * The parent of all other scopes is the scope of the immediately enclosing
@@ -97,8 +100,8 @@ from the inside outwards.
 ### Scopes of function parameters
 
 All function parameters are in the scope of the entire function body, so are
-dropped last when evaluating the function. Actual function parameters are
-dropped after any named parameters that are bound to parts of it.
+dropped last when evaluating the function. Each actual function parameter is
+dropped after any bindings introduced in that parameter's pattern.
 
 ```rust
 # struct PrintOnDrop(&'static str);
@@ -124,8 +127,8 @@ patterns_in_parameters(
 
 Local variables declared in a `let` statement are associated to the scope of
 the block that contains the `let` statement. Local variables declared in a
-`match` expression are associated to the arm scope of the `match` arm that they are declared
-in.
+`match` expression are associated to the arm scope of the `match` arm that they
+are declared in.
 
 ```rust
 # struct PrintOnDrop(&'static str);
@@ -141,14 +144,14 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
 let declared_last = PrintOnDrop("Dropped first in outer scope");
 ```
 
-If multiple patterns are used in the same arm for a `match` expression, then an unspecified
-pattern will be used to determine the drop order.
+If multiple patterns are used in the same arm for a `match` expression, then an
+unspecified pattern will be used to determine the drop order.
 
 ### Temporary scopes
 
 The *temporary scope* of an expression is the scope that is used for the
 temporary variable that holds the result of that expression when used in a
-[place context], unless it is promoted to a `static`.
+[place context], unless it is [promoted].
 
 Apart from lifetime extension, the temporary scope of an expression is the
 smallest scope that contains the expression and is for one of the following:
@@ -162,9 +165,16 @@ smallest scope that contains the expression and is for one of the following:
 * The expression for a match arm.
 * The second operand of a [lazy boolean expression].
 
-> Note: Temporaries that are created in the final expression of a function body
-> are dropped *after* any named variables bound in the function body, as there
-> is no smaller enclosing temporary scope.
+> **Notes**:
+> 
+> Temporaries that are created in the final expression of a function
+> body are dropped *after* any named variables bound in the function body, as
+> there is no smaller enclosing temporary scope.
+>
+> The [scrutinee] of a `match` expression is not a temporary scope, so
+> temporaries in the scrutinee can be dropped after the `match` expression. For
+> example, the temporary for `1` in `match 1 { ref mut z => z };` lives until
+> the end of the statement.
 
 Some examples:
 
@@ -245,6 +255,9 @@ always has the type `&'static Option<_>`, as it contains nothing disallowed).
 
 ### Temporary lifetime extension
 
+> **Note**: The exact rules for temporary lifetime extension are subject to
+> change. This is describing the current behavior only.
+
 The temporary scopes for expressions in `let` statements are sometimes
 *extended* to the scope of the block containing the `let` statement. This is
 done when the usual temporary scope would be too small, based on certain
@@ -287,11 +300,11 @@ expression which is one of the following:
 * The operand(s) of an extending [array][array expression], [cast][cast
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
-* The final expression of any extending [block expression][block expressions].
+* The final expression of any extending [block expression].
 
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some { 0: &mut 3 }`
-are all extending expressions, while the borrows in `&0 + &1` and
-`Some(&mut 0)` are not.
+are all extending expressions. The borrows in `&0 + &1` and `Some(&mut 0)` are
+not: the latter is syntactically a function call expression.
 
 The operand of any extending borrow expression has its temporary scope
 extended.
@@ -345,6 +358,8 @@ variable or field from being dropped automatically.
 [interior mutability]: interior-mutability.md
 [lazy boolean expression]: expressions/operator-expr.md#lazy-boolean-operators
 [place context]: expressions.md#place-expressions-and-value-expressions
+[promoted]: destructors.md#constant-promotion
+[scrutinee]: glossary.md#scrutinee
 [statement]: statements.md
 [temporary]: expressions.md#temporaries
 [variable]: variables.md
@@ -362,7 +377,7 @@ variable or field from being dropped automatically.
 [tuple struct pattern]: patterns.md#tuple-struct-patterns
 
 [array expression]: expressions/array-expr.md#array-expressions
-[block expressions]: expressions/block-expr.md
+[block expression]: expressions/block-expr.md
 [borrow expression]: expressions/operator-expr.md#borrow-operators
 [cast expression]: expressions/operator-expr.md#type-cast-expressions
 [struct expression]: expressions/struct-expr.md
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -1,27 +1,23 @@
 # Destructors
 
-When an [initialized]&#32;[variable] in Rust goes out of scope or a [temporary]
-is no longer needed its _destructor_ is run. [Assignment] also runs the
-destructor of its left-hand operand, unless it's an uninitialized variable. If a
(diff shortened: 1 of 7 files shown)


src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
     }
 }
 
+let mut overwritten = ShowOnDrop("drops when overwritten");
```
Centril: I'd rename `ShowOnDrop` => `PrintOnDrop` since that's what it actually does.

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+overwritten = ShowOnDrop("drops when scope ends");
+
+let tuple = (ShowOnDrop("Tuple first"), ShowOnDrop("Tuple second"));
+
+let moved;
+// No destructor run on assignment.
+moved = ShowOnDrop("Drops when moved");
+// drops now, but is then uninitialized
```
Centril:
```suggestion
// Drops now, but is then uninitialized.
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+moved;
+let uninitialized: ShowOnDrop;
+// Only the first element is dropped
```
Centril:
```suggestion
// Only the first element is dropped.
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+overwritten = ShowOnDrop("drops when scope ends");
+
+let tuple = (ShowOnDrop("Tuple first"), ShowOnDrop("Tuple second"));
+
+let moved;
+// No destructor run on assignment.
+moved = ShowOnDrop("Drops when moved");
+// drops now, but is then uninitialized
+moved;
+let uninitialized: ShowOnDrop;
```
Centril:
```suggestion

let uninitialized: ShowOnDrop;
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+partially_initialized.0 = ShowOnDrop("Partial tuple first");
+```
+
+## Drop scopes
+
+Each variable or temporary is associated to a *drop scope*. When control flow
+leaves a drop scope all variables associated to that scope are dropped in
+reverse order of declaration (for variables) or creation (for temporaries).
+
+Drop scopes are determined after replacing [`for`], [`if let`] and
```
Centril:
```suggestion
Drop scopes are determined after replacing [`for`], [`if let`], and
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+moved = ShowOnDrop("Drops when moved");
+// drops now, but is then uninitialized
+moved;
+let uninitialized: ShowOnDrop;
+// Only the first element is dropped
+let mut partially_initialized: (ShowOnDrop, ShowOnDrop);
+partially_initialized.0 = ShowOnDrop("Partial tuple first");
+```
+
+## Drop scopes
```
Centril: Is all of this behavior specified/decided somewhere.. is some of it documenting current rustc behavior or are all parts decided upon by e.g. the lang team or e.g. by necessity due to NLL?

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+Overloaded operators are not distinguished from built-in operators and [binding
+modes] are not considered.
+
+Given a function, or closure, there are drop scopes for:
+
+* The entire function
+* Each [statement]
+* Each [expression]
+* Each block, including the function body
+    * [Block expressions] the expression and block scopes are the same scope.
```
Centril: This doesn't read well; separate "block expressions" and "the expression" somehow?

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+* The entire function scope is the outer most scope.
+* The function body block is contained within the scope of the entire function.
+* The parent of the expression in an expression statement is the scope of the
+  statement.
+* The parent of the initializer of a `let` statement is the `let`
+  statement's scope.
+* The parent of a statement scope is the scope of the block that contains the
+  statement.
+* The parent of the expression for a `match` guard is the scope of the arm that
+  it's for.
```
Centril:
```suggestion
  the guard is for.
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+* The parent of the expression for a given `match` arm is that arm's scope.
```
Centril: It isn't clear that this refers to the expr on RHS of `=>`.

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+* The parent of the arm scope is the scope of the match expression that it
```
Centril:
```suggestion
* The parent of the arm scope is the scope of the `match` expression that it
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+  belongs to.
+* The parent of all other scopes is the scope of the immediately enclosing
+  expression.
+
+### Scopes of function parameters
+
+All function parameters are in the scope of the entire function body, so are
+dropped last when evaluating the function. The actual function parameter is
```
Centril:
```suggestion
dropped last when evaluating the function. Actual function parameters are
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+dropped after any named parameters that are bound to parts of it.
```
Centril:
```suggestion
dropped after any bindings introduced in each parameter's pattern.
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+    (ShowOnDrop("0"), ShowOnDrop("1")),
+    (ShowOnDrop("2"), ShowOnDrop("3")),
+);
+```
+
+### Scopes of local variables
+
+Local variables declared in a `let` statement are associated to the scope of
+the block that contains the `let` statement. Local variables declared in a
+match are associated to the arm scope of the `match` arm that they are declared
```
Centril:
```suggestion
`match` expression are associated to the arm scope of the `match` arm that they are declared
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+    let declared_in_block = ShowOnDrop("Dropped in inner scope");
 }
-# println!("");
-{
-    // Tuple elements drop in forwards order
-    let tuple = (ShowOnDrop("Tuple first"), ShowOnDrop("Tuple second"));
+let declared_last = ShowOnDrop("Dropped first in outer scope");
+```
+
+If multiple patterns are used in the same arm for a match, then an unspecified
```
Centril:
```suggestion
If multiple patterns are used in the same arm for a `match` expression, then an unspecified
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+pattern will be used to determine the drop order.
+
+### Temporary scopes
+
+The *temporary scope* of an expression is the scope that is used for the
+temporary variable that holds the result of that expression when used in a
+[place context], unless it is promoted to a `static`.
```
Centril: "a static" meaning?

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+
+Apart from lifetime extension, the temporary scope of an expression is the
+smallest scope that contains the expression and is for one of the following:
+
+* The entire function body.
+* A statement.
+* The body of a [`if`], [`while`] or [`loop`] expression.
+* The `else` block of an `if` expression.
+* The condition expression of an `if` or `while` expression, or a `match` guard.
```
Centril: What about the scrutinee of `match` / `if let` / `while let` ?

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+syntactic rules. For example:
+
+```rust
+let x = &mut 0;
+// Usually a temporary would be dropped by now, but the temporary for `0` lives
+// to the end of the block.
+println!("{}", x);
+```
+
+If a borrow, dereference, field or tuple indexing expression has an extended
```
Centril:
```suggestion
If a borrow, dereference, field, or tuple indexing expression has an extended
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+extended temporary scope then the indexed expression also has an extended
+temporary scope.
+
+#### Extending based on patterns
+
+An *extending pattern* is either
+
+* An [identifier pattern] that binds by reference or mutable reference.
+* A [struct][struct pattern], [tuple][tuple pattern], [tuple struct][tuple
+  struct pattern] or [slice][slice pattern] pattern where at least one of the
```
Centril:
```suggestion
  struct pattern], or [slice][slice pattern] pattern where at least one of the
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+If the pattern in a `let` statement is an extending pattern then the temporary
+scope of the initializer expression is extended.
+
+#### Extending based on expressions
+
+For a let statement with an initializer, an *extending expression* is an
+expression which is one of the following:
+
+* The initializer expression.
+* The operand of a extending [borrow expression].
```
Centril:
```suggestion
* The operand of an extending [borrow expression].
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+* The operand(s) of an extending [array][array expression], [cast][cast
+  expression], [braced struct][struct expression] or [tuple][tuple expression]
```
Centril:
```suggestion
  expression], [braced struct][struct expression], or [tuple][tuple expression]
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+  expression.
+* The final expression of any extending [block expression][block expressions].
+
+So the borrow expressions in `&mut 0`, `(&1, &mut 2)` and `Some { 0: &mut 3 }`
```
Centril:
```suggestion
So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some { 0: &mut 3 }`
```

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+are all extending expressions, while the borrows in `&0 + &1` and
+`Some(&mut 0)` are not.
```
Centril: `Some(&mut 0)` behaves this way because it is a function call?

src/destructors.md
```diff
@@ -37,57 +33,344 @@ impl Drop for ShowOnDrop {
...
+# x;
+```
+
+Here are some examples where expressions don't have extended temporary scopes:
+
+```rust,compile_fail
+# fn temp() {}
+# trait Use { fn use_temp(&self) -> &Self { self } }
+# impl Use for () {}
+// The temporary that stores the result of `temp()` lives only lives until the
```
Centril:
```suggestion
// The temporary that stores the result of `temp()` only lives until the
```

src/destructors.md
matthewjasper: They're not a temporary scopes. For example in `match 1 { ref mut z => (), };` the temporary for `1` lives until the end of the statement

src/destructors.md
matthewjasper: It's a description of the current behaviour, which is mostly observable. That said, some of it could be changed: RFC 66 proposes this, but it is still unimplemented.

src/destructors.md
Centril: Right; can you add a note explaining why `Some(&mut 0)` is different from `Some { 0: &mut 3 }` since it may not be so clear for anyone reading this?

src/destructors.md
Centril: That seems like noteworthy in the reference.

src/destructors.md
Centril: OK; my interest is mainly in not additionally tying the language team's hand due to what we add here; if the text could be clarified to note what are stable guarantees and what are not that'd be helpful.

src/destructors.md
```diff
@@ -1,28 +1,24 @@
...
+1. If `T: Drop`, calling [`<T as std::ops::Drop>::drop`]
 2. Recursively running the destructor of all of its fields.
-    * The fields of a [struct], [tuple] or [enum variant] are dropped in
-      declaration order. \*
-    * The elements of an [array] or owned [slice][array] are dropped from the
-      first element to the last. \*
-    * The captured values of a [closure] are dropped in an unspecified order.
+    * The fields of a [struct] or [tuple] are dropped in declaration order.
+    * The fields of the active [enum variant] are dropped in declaration order.
+    * The elements of an [array] or owned [slice] are dropped from the
```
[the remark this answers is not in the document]
matthewjasper: I don't think that it's specified anywhere authoritative, but judging from the comments on that RFC I think that the RFC left it out by accident.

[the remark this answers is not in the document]
Centril: cc @nikomatsakis -- I think this looks quite reasonable and justified; can you confirm that this was intended?

pnkfelix:
@matthewjasper sorry I had not noticed this until now. I'll try to provide my review soon

src/destructors.md
```diff
@@ -1,92 +1,392 @@
...
+the expression was originally written, without changing the runtime behavior.
+That is, the promoted expression can be evaluated at compile-time and the
+resulting value does not contain [interior mutability] or [destructors] (these
+properties are determined based on the value where possible, e.g. `&None`
+always has the type `&'static Option<_>`, as it contains nothing disallowed).
+
+### Temporary lifetime extension
+
+> **Note**: The exact rules for temporary lifetime extension are subject to
+> change. This is describes the current behavior only.
```
jonas-schievink:
```suggestion
> change. This is describing the current behavior only.
```

Centril:
ping @pnkfelix :)
cc also @nikomatsakis

ehuss:
I rebased this on the latest master and resolved the conflicts.

src/destructors.md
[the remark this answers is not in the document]
nikomatsakis: I believe it was intended, yes. But more generally, if it matches the current behavior, I think we can confirm it, because the meta-intent of RFC #1857 was to specify that built-in types will not change their drop order.

src/destructors.md
```diff
@@ -1,97 +1,396 @@
...
 
-1. Calling its [`std::ops::Drop::drop`] method, if it has one.
+1. If `T: Drop`, calling [`<T as std::ops::Drop>::drop`]
 2. Recursively running the destructor of all of its fields.
-    * The fields of a [struct], [tuple] or [enum variant] are dropped in
-      declaration order. \*
-    * The elements of an [array] or owned [slice][array] are dropped from the
-      first element to the last. \*
-    * The captured values of a [closure] are dropped in an unspecified order.
+    * The fields of a [struct] or [tuple] are dropped in declaration order.
```
nikomatsakis, line 13: Nit: tuples are not declared, right? I think I would say "The fields of a tuple are dropped in order." or something

src/destructors.md
nikomatsakis: The documentation is "in the code", as they predate RFCs. Thanks @matthewjasper for writing it up, I too had an "in progress" write-up somewhere (part of a larger amendment to RFC 66...) but I'm happy to have someone else take a crack at it.

Some relevant links:

* [terminating scopes](https://github.com/rust-lang/rust/blob/df768c5c8fcb361c4dc94b4c776d6a78c12862e1/src/librustc_passes/region.rs#L58-L77)
* [the rules for temporary scopes and lets](https://github.com/rust-lang/rust/blob/df768c5c8fcb361c4dc94b4c776d6a78c12862e1/src/librustc_passes/region.rs#L438-L492)

src/introduction.md
```diff
@@ -139,9 +139,9 @@ attention to making those sections the best that they can be.
 [_Expression_]: expressions.md
 [cargo book]: ../cargo/index.html
 [cargo reference]: ../cargo/reference/index.html
-[expressions chapter]: expressions.md
-[lifetime of temporaries]: expressions.md#temporary-lifetimes
-[linkage]: linkage.md
+[expressions chapter]: expressions.html
```
lzutao, line 142: These md -> html changes may not really useful as mdbook handles that. cc @ehuss

ehuss: @lzutao can you send a PR to fix it?

PR #516, opened 2019-01-27
Document if_while_or_patterns

### LINKED ISSUE #505: Document if_while_or_patterns
RFC: https://github.com/rust-lang/rfcs/pull/2175 + https://github.com/rust-lang/rfcs/pull/2530
Proposal: https://github.com/rust-lang/rust/issues/56212
PR: https://github.com/rust-lang/rust/pull/57532
Tracking: https://github.com/rust-lang/rust/issues/48215
Target = 1.33

ehuss:
Closes #505

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -44,7 +44,7 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Patterns_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_MatchArmPatterns_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
 >              [_BlockExpression_]\
 > &nbsp;&nbsp; (`else` (
 >   [_BlockExpression_]
@@ -152,7 +152,7 @@ if let PAT = ( EXPR || EXPR ) { .. }
 [_BlockExpression_]: expressions/block-expr.html
 [_Expression_]: expressions.html
 [_LazyBooleanOperatorExpression_]: expressions/operator-expr.html#lazy-boolean-operators
-[_Patterns_]: expressions/match-expr.html
+[_MatchArmPatterns_]: expressions/match-expr.html
 [_eRFCIfLetChain_]: https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#rollout-plan-and-transitioning-to-rust-2018
 [`match` expression]: expressions/match-expr.html
 [scrutinee]: glossary.html#scrutinee
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -44,7 +44,7 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_Patterns_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
 >              [_BlockExpression_]\
 > &nbsp;&nbsp; (`else` (
 >   [_BlockExpression_]
@@ -95,10 +95,10 @@ let a = if let Some(1) = x {
 assert_eq!(a, 3);
 ```
 
-An `if let` expression is equivalent to a `match` expression as follows:
+An `if let` expression is equivalent to a [`match` expression] as follows:
 
 ```rust,ignore
-if let PAT = EXPR {
+if let PATS = EXPR {
     /* body */
 } else {
     /*else */
@@ -109,8 +109,23 @@ is equivalent to
 
 ```rust,ignore
 match EXPR {
-    PAT => { /* body */ },
+    PATS => { /* body */ },
     _ => { /* else */ },    // () if there is no else
+}
+```
+
+Multiple patterns may be specified with the `|` operator. This has the same semantics
+as with `|` in `match` expressions:
+
+```rust
+enum E {
+    X(u8),
+    Y(u8),
+    Z(u8),
+}
+let v = E::Y(12);
+if let E::X(n) | E::Y(n) = v {
+    assert_eq!(n, 12);
 }
 ```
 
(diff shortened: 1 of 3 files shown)


src/expressions/if-expr.md
```diff
@@ -44,7 +44,7 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_Patterns_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
```
Centril:
```suggestion
> &nbsp;&nbsp; `if` `let` [_MatchArmPatterns_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
```
(see https://doc.rust-lang.org/nightly/reference/expressions/match-expr.html)

src/expressions/if-expr.md
```diff
@@ -106,11 +106,26 @@ is equivalent to
 
 ```rust,ignore
 match EXPR {
-    PAT => { /* body */ },
+    PATS => { /* body */ },
     _ => { /* else */ },    // () if there is no else
 }
 ```
 
+Multiple patterns may be specified with the `|` operator. This has the same semantics
```
Centril:
```suggestion
Multiple patterns may be specified with the `|` operator. A leading `|` is also allowed. This has the same semantics
```

src/expressions/loop-expr.md
```diff
@@ -67,7 +67,7 @@ while i < 10 {
 
 > **<sup>Syntax</sup>**\
 > [_PredicatePatternLoopExpression_] :\
-> &nbsp;&nbsp; `while` `let` [_Pattern_] `=` [_Expression_]<sub>except struct expression</sub>
+> &nbsp;&nbsp; `while` `let` [_Patterns_] `=` [_Expression_]<sub>except struct expression</sub>
```
Centril:
```suggestion
> &nbsp;&nbsp; `while` `let` [_MatchArmPatterns_] `=` [_Expression_]<sub>except struct expression</sub>
```

src/expressions/loop-expr.md
```diff
@@ -99,12 +99,23 @@ is equivalent to
...
 'label: loop {
     match EXPR {
-        PAT => { /* loop body */ },
+        PATS => { /* loop body */ },
         _ => break,
     }
 }
 ```
 
+Multiple patterns may be specified with the `|` operator. This has the same semantics
```
Centril:
```suggestion
Multiple patterns may be specified with the `|` operator. A leading `|` is also allowed. This has the same semantics
```

src/expressions/match-expr.md
```diff
@@ -62,7 +62,8 @@ match x {
 Variables bound within the pattern are scoped to the match guard and the arm's
 expression. The [binding mode] (move, copy, or reference) depends on the pattern.
 
-Multiple match patterns may be joined with the `|` operator:
+Multiple match patterns may be joined with the `|` operator. Each pattern will be
+tested in left-to-right sequence until a successful match is found.
```
Centril: Is the LTR sequence actually semantically observable atm due to some side-effect?

src/expressions/match-expr.md
```diff
@@ -75,17 +76,21 @@ let message = match x {
 assert_eq!(message, "a few");
 ```
 
-Please notice that the `2..=9` is a [Range Pattern], not a [Range Expression]
-and, thus, only those types of ranges supported by range patterns can be used
-in match arms.
+> Note: The `2..=9` is a [Range Pattern], not a [Range Expression] and, thus,
```
Centril:
```suggestion
> Note: The `2..=9` is a [Range Pattern], not a [Range Expression]. Thus,
```

src/expressions/match-expr.md
```diff
@@ -75,17 +76,21 @@ let message = match x {
...
+> only those types of ranges supported by range patterns can be used in match
+> arms.
+
+Every binding in each `|` separated pattern must appear in all of the patterns
+in the arm. Every binding of the same name must have the same type, and have
```
Centril: This is a bit vague since it doesn't say whether "same type" accounts for subtyping or not, but it's serviceable until we have a more formal semantics.

src/expressions/match-expr.md
ehuss: Yea. In the example at the bottom of the page shows how guards will be called for each pattern, and the result would be different if it was some other order.

Another example:
```rust
struct S {
    f1: i32,
    f2: i32,
}
let s = S { f1: 1, f2: 2 };
match s {
    S { f1: z @ 1, .. } | S { f2: z @ 2, .. } => assert_eq!(z, 1),
    _ => {}
}
```

Again, the answer for `z` would be different if the order was right-to-left.

BTW, I'm not absolutely certain it is always left-to-right.  I perused the code, and it looks like it visits the patterns in the order they are parsed, and of course tests bear that out. But that is part of the rustc code that I know almost nothing about.  I could imagine there could be some weird optimizations somewhere that rearrange pattern tests.  I can take it out unless someone knows for certain.

src/expressions/if-expr.md
ehuss: I'm not sure I agree with this change. I think it is clear from the grammar that a leading `|` is allowed.  In general I'd like to reduce the amount of "describe the grammar in prose" unless it is necessary.  I can add it if you feel strongly about it, but I would lean towards leaving this out.

Centril: Hah. I actually agree with your sentiment but wanted to be sure you didn't miss anything... I also lean towards leaving it out.

src/expressions/match-expr.md
Centril: > Yea. In the example at the bottom of the page shows how guards will be called for each pattern, and the result would be different if it was some other order.

This bit isn't relevant here tho since the guard isn't part of the last pattern in the list of or-patterns.

> Another example:
> [...]
> Again, the answer for `z` would be different if the order was right-to-left.

Oh of course...! That's a good example to include since it's not so obvious... Simplified:
```rust
struct S(i32, i32);

let s = S(1, 2);

match s {
    S(z @ 1, _) | S(_, z @ 2) => assert_eq!(z, 1),
    _ => {}
}
```

> I could imagine there could be some weird optimizations somewhere that rearrange pattern tests. I can take it out unless someone knows for certain.

Those optimizations would only be valid if they preserved the dynamic semantics (which is what you've stated / defined) so I'd leave it in.

PR #517, opened 2019-01-27
Document extern_crate_self

### LINKED ISSUE #501: Document #56409 (`extern crate self as ALIAS`)
+ Tracking issue: https://github.com/rust-lang/rust/issues/56409
+ Expected version: 1.34
+ Implementation: https://github.com/rust-lang/rust/pull/55275
+ Stabilization PR: https://github.com/rust-lang/rust/pull/57407

ehuss:
Closes #501.

### DIFF
--- a/src/items/extern-crates.md
+++ b/src/items/extern-crates.md
@@ -2,7 +2,13 @@
 
 > **<sup>Syntax:<sup>**\
 > _ExternCrate_ :\
-> &nbsp;&nbsp; `extern` `crate` ( [IDENTIFIER] | `self` ) (`as` ( [IDENTIFIER] | `_` ) )<sup>?</sup> `;`
+> &nbsp;&nbsp; `extern` `crate` _CrateRef_ _AsClause_<sup>?</sup> `;`
+>
+> _CrateRef_ :\
+> &nbsp;&nbsp; [IDENTIFIER] | `self`
+>
+> _AsClause_ :\
+> &nbsp;&nbsp; `as` ( [IDENTIFIER] | `_` )
 
 An _`extern crate` declaration_ specifies a dependency on an external crate.
 The external crate is then bound into the declaring scope as the [identifier]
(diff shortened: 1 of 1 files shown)


src/items/extern-crates.md
```diff
@@ -2,11 +2,12 @@
 
 > **<sup>Syntax:<sup>**\
 > _ExternCrate_ :\
-> &nbsp;&nbsp; `extern` `crate` [IDENTIFIER]&nbsp;(`as` ( [IDENTIFIER] | `_` ) )<sup>?</sup> `;`
+> &nbsp;&nbsp; `extern` `crate` ( [IDENTIFIER] | `self` ) (`as` ( [IDENTIFIER] | `_` ) )<sup>?</sup> `;`
```
Centril: Since you've referenced the "as clause" below, I'd pull that bit out to a separate production:
```
AsClause :
  `as` ( [IDENTIFIER] | `_` )
```
I'd also pull `( [IDENTIFIER] | self )` out into a production `CrateRef`.

src/items/extern-crates.md
```diff
@@ -16,6 +17,9 @@ the `crateid` attributes that were declared on the external crate when it was
 compiled. If no `crateid` is provided, a default `name` attribute is assumed,
 equal to the [identifier] given in the `extern crate` declaration.
 
+The `self` crate may be imported which creates a binding to the current crate.
```
Centril: In the case of macros, I'm not sure whether the "current crate" always means the crate for the call-site or the def-site of the macro... I guess until you can use `Span::def_site` on stable it doesn't matter?

ehuss: I believe it is always call-site. I think the new macro-rule documentation explains this in the hygiene section pretty well ("all identifiers referred to in a macro are expanded as-is, and are looked up at the macro's invocation site."). Do you think it needs extra clarification here?

Centril: > Do you think it needs extra clarification here?

No. =P

PR #518, opened 2019-01-27
Clarify what access struct updates do

### LINKED ISSUE #315: Explain struct partial initiation/update syntax in more detail
When reading about struct update syntax in The Book 2nd Edition, I wondered what would happen with Copy field members versus Clone field members versus neither.  Googling around didn't help (if there's an RFC I couldn't find it) and eventually I just used the playground to figure out the behavior: Copy fields are copied, other fields are moved, possibly leading to a "partially moved" struct.

It would be nice if the book and/or reference spelled this out.

Citations:
  https://doc.rust-lang.org/reference/expressions/struct-expr.html
  "with [...] the values in the base expression for all other fields", 'with' is vague

  https://doc.rust-lang.org/stable/book/second-edition/ch05-01-defining-structs.html
  "use the rest of the values from the fields of the instance", 'use' is vague

  https://doc.rust-lang.org/nightly/book/first-edition/structs.html#update-syntax
  "it will copy the values", technically not true for non-Copy types

  https://play.rust-lang.org/?gist=cb23fb3f9f9c631370b2b631367b8b16&version=stable
  Example of a "partially moved value"

matthewjasper:
Closes #315

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -98,11 +98,16 @@ expressions are value expressions.
 
 A *value expression* is an expression that represents an actual value.
 
-The left operand of an [assignment][assign] or [compound assignment] expression
-is a place expression context, as is the single operand of a unary [borrow], and
-the operand of any [implicit borrow]. The discriminant or subject of a
-[match expression][match] and right side of a [let statement] is also a place
-expression context. All other expression contexts are value expression contexts.
+The following contexts are *place expression* contexts:
+
+* The left operand of an [assignment][assign] or [compound assignment] expression.
+* The operand of a unary [borrow] or [dereference] operator.
+* The operand of a field expression.
+* The indexed operand of an array indexing expression.
+* The operand of any [implicit borrow].
+* The initializer of a [let statement].
+* The [scrutinee] of a [match expression][match].
+* The base of a [functional update] struct expression.
 
 > Note: Historically, place expressions were called *lvalues* and value
 > expressions were called *rvalues*.
@@ -279,6 +284,7 @@ They are never allowed before:
 [closure expressions]:  expressions/closure-expr.html
 [enum variant]:         expressions/enum-variant-expr.html
 [field]:                expressions/field-expr.html
+[functional update]:    expressions/struct-expr.html#functional-update-syntax
 [grouped]:              expressions/grouped-expr.html
 [literals]:             expressions/literal-expr.html
 [match]:                expressions/match-expr.html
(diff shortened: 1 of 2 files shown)


src/expressions.md
```diff
@@ -98,11 +98,16 @@ expressions are value expressions.
...
 
-The left operand of an [assignment][assign] or [compound assignment] expression
-is a place expression context, as is the single operand of a unary [borrow], and
-the operand of any [implicit borrow]. The discriminant or subject of a
-[match expression][match] and right side of a [let statement] is also a place
-expression context. All other expression contexts are value expression contexts.
+The following contexts are *place expression* contexts:
+
+* The left operand of an [assignment][assign] or [compound assignment] expression.
+* The operand of a unary [borrow] or [dereference] operator.
```
Centril: [dereference] is a dead link when I click "View file".

src/expressions.md
```diff
@@ -98,11 +98,16 @@ expressions are value expressions.
...
+* The operand of a field expression.
+* The indexed operand of an array indexing expression.
+* The operand of any [implicit borrow].
+* The initializer of a [let statement].
+* The [scrutinee] of a [match expression][match].
```
Centril: Surely for `if let` and `while let` also since they de-sugar to `match`?

src/expressions/struct-expr.md
```diff
@@ -56,19 +56,23 @@ colon.
...
-base) must have the same struct type as the new struct type being formed. The
-entire expression denotes the result of constructing a new struct (with the
-same type as the base expression) with the given values for the fields that
-were explicitly specified and the values in the base expression for all other
-fields. Just as with all struct expressions, all of the fields of the struct
-must be [visible], even those not explicitly named.
+base) must have the same struct type as the new struct type being formed.
+
+The entire expression uses the given values for the fields that were specified
+and moves or copies the remaining fields from the base expression. Just as with
```
Centril:
```suggestion
and moves or copies the remaining fields from the base expression. As with
```

src/expressions.md
```diff
@@ -92,17 +92,24 @@ depends both on its own category and the context it occurs within.
...
+The following contexts are *place expression* contexts:
+
+* The left operand of an [assignment][assign] or [compound assignment]
+  expression.
+* The operand of a unary [borrow] or [dereference][deref] operator.
+* The operand of a field expression.
+* The indexed operand of an array indexing expression.
+* The operand of any [implicit borrow].
+* The initializer of a [let statement].
+* The [scrutinee] of a [`if let`], [`match`][match] or [`while let`]
```
Centril:
```suggestion
* The [scrutinee] of a [`if let`], [`match`][match], or [`while let`]
```

PR #521, opened 2019-01-27
Add linkchecker.

ehuss:
This is a little gumby, but it works.

It:
- Downloads linkchecker from rust-lang/rust.
- Copies the HTML pages from nightly rustup, which should be a close reflection of how they look on doc.rust-lang.org.
- Build and copies the reference.
- Runs the linkchecker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ install:
 
 script:
   - export PATH=$PATH:/home/travis/.cargo/bin && mdbook test
-  - cd stable-check && cargo run -- ../src
+  - (cd stable-check && cargo run -- ../src)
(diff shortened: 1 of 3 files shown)


Havvy:
@Mark-Simulacrum Is this in your review queue?

Mark-Simulacrum:
It wasn't, but I will try to review it this week. Sorry for missing it earlier!

Mark-Simulacrum:
This looks reasonable to me, though I would expect spurious failures somewhat regularly.

Havvy:
I don't think anybody else is going to review this, nor are there any known stoppers, so let's try it and see what happens.

PR #522, opened 2019-02-01
Fix an incorrect note

upsuper:
Closures that capture local variables shouldn't be able to cast to function pointers, should they?


matthewjasper:
Thanks!

PR #528, opened 2019-03-01
Update attribute documentation.

### LINKED ISSUE #500: Document RFC 2539 (`#[cfg_attr(p, attr1, attr2)]`)
- Version: 1.33
- PR: https://github.com/rust-lang/rust/pull/57332
- Tracking issue: https://github.com/rust-lang/rust/issues/54881
- RFC: https://github.com/rust-lang/rfcs/pull/2539

### LINKED ISSUE #525: Document `unrestricted_attribute_tokens`
- Version: 1.34
- Stabilization: https://github.com/rust-lang/rust/pull/57367
- Tracking: https://github.com/rust-lang/rust/issues/55208

### LINKED ISSUE #526: Document RFC 2103 `tool_attributes`
- Version: 1.30
- Stabilization: https://github.com/rust-lang/rust/pull/53459/
- Tracking: https://github.com/rust-lang/rust/issues/44690
- RFC: https://github.com/rust-lang/rfcs/blob/master/text/2103-tool-attributes.md

Tool lint attributes are documented, but general tool attributes are not.

ehuss:
- Document `unrestricted_attribute_tokens`, closes #525
- Document `tool_attributes`, closes #526
- Document `cfg_attr_multi`, closes #500
- Remove `linked_from`, it was removed in https://github.com/rust-lang/rust/pull/37973/
- Remove `macro_reexport`, it was removed in https://github.com/rust-lang/rust/pull/49982/
- Remove unstable attributes: `linkage`, `link_args`

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -84,8 +84,9 @@ fn some_unused_variables() {
 
 ## Meta Item Attribute Syntax
 
-A "meta item" is the syntax used for the _Attr_ rule by built-in attributes
-and the [`meta` macro fragment specifier]. It has the following grammar:
+A "meta item" is the syntax used for the _Attr_ rule by most built-in
+attributes and the [`meta` macro fragment specifier]. It has the following
+grammar:
 
 > **<sup>Syntax</sup>**\
 > _MetaItem_ :\
@@ -576,10 +577,12 @@ used.
 struct S {
 }
 
-// Controls the "cyclomatic complexity" threshold for the cLippy tool.
+// Controls the "cyclomatic complexity" threshold for the clippy tool.
 #[clippy::cyclomatic_complexity = "100"]
 pub fn f() {}
 ```
+
+> Note: `rustc` currently recognizes the tools "clippy" and "rustfmt".
 
 [_DelimTokenTree_]: macros.html
 [_LiteralExpression_]: expressions/literal-expr.html
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -1,44 +1,39 @@
 # Attributes
 
 > **<sup>Syntax</sup>**\
-> _Attribute_ :\
-> &nbsp;&nbsp; _InnerAttribute_ | _OuterAttribute_
->
 > _InnerAttribute_ :\
-> &nbsp;&nbsp; `#![` MetaItem `]`
+> &nbsp;&nbsp; `#` `!` `[` _Attr_ `]`
 >
 > _OuterAttribute_ :\
-> &nbsp;&nbsp; `#[` MetaItem `]`
+> &nbsp;&nbsp; `#` `[` _Attr_ `]`
 >
-> _MetaItem_ :\
-> &nbsp;&nbsp; &nbsp;&nbsp; [_SimplePath_]\
-> &nbsp;&nbsp; | [_SimplePath_] `=` [_LiteralExpression_]<sub>_without suffix_</sub>\
-> &nbsp;&nbsp; | [_SimplePath_] `(` _MetaSeq_<sup>?</sup> `)`
+> _Attr_ :\
+> &nbsp;&nbsp; [_SimplePath_] _AttrInput_<sup>?</sup>
 >
-> _MetaSeq_ :\
-> &nbsp;&nbsp; _MetaItemInner_ ( `,` MetaItemInner )<sup>\*</sup> `,`<sup>?</sup>
->
-> _MetaItemInner_ :\
-> &nbsp;&nbsp; &nbsp;&nbsp; _MetaItem_\
-> &nbsp;&nbsp; | [_LiteralExpression_]<sub>_without suffix_</sub>
+> _AttrInput_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_DelimTokenTree_]\
+> &nbsp;&nbsp; | `=` [_LiteralExpression_]<sub>_without suffix_</sub>
 
 An _attribute_ is a general, free-form metadatum that is interpreted according
 to name, convention, and language and compiler version. Attributes are modeled
 on Attributes in [ECMA-335], with the syntax coming from [ECMA-334] \(C#).
 
-Attributes may appear as any of:
-
-* A single identifier, the _attribute name_
-* An identifier followed by the equals sign '=' and a literal, providing a
-  key/value pair
-* An identifier followed by a parenthesized list of sub-attribute arguments
-  which include literals
-
-Literal values must not include integer or float type suffixes.
-
-_Inner attributes_, written with a bang ("!") after the hash ("#"), apply to the
+_Inner attributes_, written with a bang (`!`) after the hash (`#`), apply to the
 item that the attribute is declared within. _Outer attributes_, written without
 the bang after the hash, apply to the thing that follows the attribute.
+
+The attribute consists of a path to the attribute, followed by an optional
+delimited token tree whose interpretation is defined by the attribute.
+Attributes other than macro attributes also allow the input to be an equals
+sign (`=`) followed by a literal expression. See the [meta item
+syntax](#meta-item-attribute-syntax) below for more details.
+
+Attributes can be classified into the following kinds:
+
+* Built-in attributes
+* [Macro attributes][attribute macro]
+* [Derive macro helper attributes]
+* [Tool attributes](#tool-attributes)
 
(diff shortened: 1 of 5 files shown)


ehuss:
This explicitly does not talk about attribute name scoping and resolution. That is a much bigger topic that I might tackle separately.

I'm also looking at documenting the syntax for every built-in attribute, better incorporating _MetaItem_. I think this should be done separately since it will likely be a large change.

Finally, I've noted a number of built-in attributes that are not documented that probably should be. I'll try to follow up on those later.

Can someone clarify the distinction of inert/active attributes? I can't find anything in rustc that tracks that. In what way is that difference ever observable by the user?

src/attributes.md
```diff
@@ -1,44 +1,39 @@
...
 
-Literal values must not include integer or float type suffixes.
+Attributes can be classified into the following kinds:
 
-_Inner attributes_, written with a bang ("!") after the hash ("#"), apply to the
-item that the attribute is declared within. _Outer attributes_, written without
-the bang after the hash, apply to the thing that follows the attribute.
+* Built-in attributes
+* [Macro attributes][attribute macro]
+* [Derive mode helper attributes]
```
petrochenkov: https://github.com/rust-lang-nursery/reference/pull/412#discussion_r214538427
I still think "mode" should not be used here.

src/attributes.md
```diff
@@ -87,11 +82,33 @@ fn some_unused_variables() {
 }
 ```
 
-There are three kinds of attributes:
+## Meta Item Attribute Syntax
 
-* Built-in attributes
-* Macro attributes
-* Derive mode helper attributes
+A "meta item" is the syntax used for the _Attr_ rule by built-in attributes
```
petrochenkov: "built-in attributes" -> "most of built-in attributes", this is not a requirement but rather a "tradition" and implementation detail.
`cfg_attr` for example, doesn't fit into the meta item syntax and may contain attributes with "unrestricted" input inside it.

Also, `meta` macro fragment not accepting attributes with the unrestricted input is generally a bug / something under-implemented, but I guess the reference needs to describe the actual compiler behavior rather than the intent.

src/attributes.md
```diff
@@ -552,6 +556,32 @@ impl<T: PartialEq> PartialEq for Foo<T> {
...
+
+```rust
+// Tells the rustfmt tool to not format the following element.
+#[rustfmt::skip]
+struct S {
+}
+
+// Controls the "cyclomatic complexity" threshold for the cLippy tool.
+#[clippy::cyclomatic_complexity = "100"]
+pub fn f() {}
```
petrochenkov: It probably makes sense to say that only two "tool modules" are currently available - `rustfmt` and `clippy`, and both are hard-coded into the compiler, and there's no way to introduce a custom tool module right now.

petrochenkov:
>Can someone clarify the distinction of inert/active attributes? I can't find anything in rustc that tracks that. In what way is that difference ever observable by the user?

An active attribute (a macro attribute, `cfg`, `cfg_attr` or `derive`) transforms its input and disappears in the process.
An inert attribute just sits there and waits for being interpreted (by the compiler, or a tool, or a proc macro).

ehuss:
Thanks for the speedy review!

From a language perspective, it seems like the only scenario where the presence/absence of an attribute matters is for attribute and derive macros, right? I've been doing some tests, and it doesn't seem to match my expectations. Attribute macros see `cfg` attributes if the cfg predicate passes, the `cfg` isn't removed. If the cfg predicate fails, then everything is pruned before the proc macro gets a chance to run, so that implies that `cfg` is processed before proc-macros and it should have been removed when it passed. When does a `cfg` with a passing predicate get removed? How can one write code such that the removal can be observed?

The docs also mention `test` being either inert or active, but proc-macros see `#[test]` regardless of whether or not `--test` is used. Perhaps `test` runs after proc-macros? In which case, it seems like an unobservable distinction, irrelevant to the language.

I do see that previously processed attribute macros are removed when you have multiple of them.

I want to be clear if Tool Attributes are always inert. I can't find the code where this removal happens, and I can't find a way to observe it (other than the multiple attribute macro situation).

src/attributes.md
ehuss: I agree, "mode" has always seemed strange to me.

src/attributes.md
ehuss: Ah, good point. If it changes, I'll try to make sure this gets updated. 

My plan is to expand this section in the future to include some of the common templates you added (like _MetaNameValueStr_) so that other places in the documentation can just say something like "`crate_name` uses the _MetaNameValueStr_ syntax to …". When I do that, I might tweak the wording here to make the relationship to built-in attributes clearer.

src/attributes.md
ehuss: I thought about that, but I was reluctant to enshrine "clippy" and "rustfmt" as part of the language. I don't want to be too pedantic, but maybe the added note is ok? To me it's a question of the balance of the reference targeting the language vs the rustc implementation.

src/conditional-compilation.md
```diff
@@ -151,7 +151,7 @@ Key-value option set once with the target's pointer width in bits. For example,
 for targets with 32-bit pointers, this is set to `"32"`. Likewise, it is set
 to `"64"` for targets with 64-bit pointers.
 
-<!-- Are there targets that have a different bit number? --> 
+<!-- Are there targets that have a different bit number? -->
 
 ### `target_has_atomic`
```
Centril: Note: this one is unstable.

src/procedural-macros.md
```diff
@@ -195,14 +195,14 @@ struct Struct {
...
 Attribute macros are defined by a [public]&#32;[function] with the
-`proc_macro_attribute` [attribute] that a signature of
-`(TokenStream, TokenStream) -> TokenStream`. The first [`TokenStream`] is the
-attribute's metaitems, not including the delimiters. If the attribute is written
-without a metaitem, the attribute [`TokenStream`] is empty. The second
-[`TokenStream`] is of the rest of the [item] including other [attributes] on the
-[item]. The returned [`TokenStream`] replaces the [item] with an arbitrary
-number of [items]. These macros cannot expand to syntax that defines new
-`macro_rule` style macros.
+`proc_macro_attribute` [attribute] that a signature of `(TokenStream,
```
Centril:
```suggestion
`proc_macro_attribute` [attribute] that has a signature of `(TokenStream,
```

src/attributes.md
```diff
@@ -552,6 +557,34 @@ impl<T: PartialEq> PartialEq for Foo<T> {
...
+#[rustfmt::skip]
+struct S {
+}
+
+// Controls the "cyclomatic complexity" threshold for the clippy tool.
+#[clippy::cyclomatic_complexity = "100"]
+pub fn f() {}
+```
+
+> Note: `rustc` currently recognizes the tools "clippy" and "rustfmt".
```
Havvy:
```suggestion
> Note: `rustc` currently only recognizes the tools "clippy" and "rustfmt".
```

petrochenkov:
@ehuss 
>From a language perspective, it seems like the only scenario where the presence/absence of an attribute matters is for attribute and derive macros, right?

I don't understand what this means.

>`cfg` is processed before proc-macros

Generally, yes, but in details it's a mess that needs fixing (individual pieces were implemented in different time by different people without understanding the whole picture and having a common "vision").
I have a couple of PRs in the pipeline that'll hopefully land in March, then we'll be able to document how `cfg`/`cfg_attr` works relatively to attribute macros and derives.

>Attribute macros see cfg attributes if the cfg predicate passes, the cfg isn't removed.

This is an implementation defect, it causes cfg predicates being evaluated multiple times.
At least the predicate result is the same every time (I think) and `cfg` application is idempotent.
`cfg` is effectively a macro and it should behave like macro, with exception of the general "early expansion" aspect which we cannot change.

>I want to be clear if Tool Attributes are always inert.

The compiler always treats tool attributes as inert.
Tools of course may do anything with them, or with code in general.

petrochenkov:
Meta: looks like I'm now assigned, but I cannot merge PRs in this repo.

Centril:
@petrochenkov Oh; uhmm, just leave a note when you think this is in a good state and someone will merge it. :)

ehuss:
Thanks everyone for taking a look!

@petrochenkov thanks, I think that makes it clearer to me.

I'll follow up with separate PR's for some of the other things noted here.

PR #529, opened 2019-03-02
Remove target_has_atomic.

ehuss:
`target_has_atomic` is not stable, and it's not clear to me if it ever will be. See https://github.com/rust-lang/rust/issues/56753 for more context.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -153,20 +153,6 @@ to `"64"` for targets with 64-bit pointers.
 
 <!-- Are there targets that have a different bit number? --> 
 
-### `target_has_atomic`
-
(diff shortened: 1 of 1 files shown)


Havvy:
The history here seems to me to be that it was merged (see https://github.com/rust-lang/rust/pull/57425) and then partially destablized in https://github.com/rust-lang/rust/pull/58616/files except I cannot tell if the cfgs were destabilized themselves without going into the autogenerator code which I don't have the energy do do right now. So I can't actually see if this is destabilized or not.

Centril:
@Havvy It's unstable (I made sure of that...): https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=550f1b52e55d13a818979fe2fa0be95f

PR #530, opened 2019-03-02
Remove "mode" from derive macro terminology.

ehuss:
This is just my personal preference, which seems to read more easily to me. The concept of "mode" wasn't really clear, and I think "derive macro" is sufficient.

For context, see: https://github.com/rust-lang-nursery/reference/pull/412#discussion_r214538427 and https://github.com/rust-lang-nursery/reference/pull/528#pullrequestreview-209855431

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -91,7 +91,7 @@ There are three kinds of attributes:
 
 * Built-in attributes
 * Macro attributes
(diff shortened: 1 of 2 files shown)


Havvy:
Just a quick check (e.g. didn't check if links work), so not going to r+, but everything appears fine to me.

PR #537, opened 2019-03-15
Reorg and update attributes

### LINKED ISSUE #534: Attribute cleanup
I would like to regroup and update the attributes page, and I would like some feedback on how to best improve it.  Some of my goals are:

- Every attribute should have a dedicated section *somewhere* in the reference, making it easy to link to them.
- Every attribute should explain the format of its syntax for its meta input (this is finished in my local branch).
- Every attribute should be documented well enough to understand what it does and how it works.
- Remove the "Crate-only" and "Misc" sections (or at least shrink "misc").
- Ensure all the "Attributes on" sections exist and are complete (#355).
- Maybe avoid using attributes.md as a dumping ground for random attributes?

I think the last point may be the most debatable.  Right now, to me, it appears as just a giant page that is difficult to navigate.  I think it would be good to move the details off to separate pages, and just have an index in attributes.md that links to where to find each one.  This is already done for about half the attributes.  Unfortunately it's not really clear where some of the other attributes should live.

Here is a preliminary outline of how I think such an index would look, and where the attributes link to.

- Conditional Compilation
    - `cfg` (conditional-compilation.md)
    - `cfg_attr` (conditional-compilation.md)
- Testing
    - `test`
    - `ignore`
    - `should_panic`
- Derive
    - `derive`
    - `automatically_derived` (currently undocumented)
- Macro-related
    - `macro_use` (FUTURE macros-by-example.md)
    - `macro_export` (FUTURE macros-by-example.md)
    - `proc_macro` (procedural-macros.md)
    - `proc_macro_derive` (procedural-macros.md)
    - `proc_macro_attribute` (procedural-macros.md)
- Lint Check  
    - `allow`
    - `deny`
    - `forbid`
    - `warn`
    - `must_use`
    - Tool lints
- Linking, Symbols, and FFI  
    - `link` (external-blocks.md)
    - `link_name` (external-blocks.md???)
    - `no_link` (MAYBE extern-crates.md?)
    - `repr` (type-layout.md)
    - `crate_type` (linkage.md)
    - `no_main`
    - `no_start`
    - `export_name` (abi.md)
    - `link_section` (abi.md)
    - `no_mangle` (abi.md)
    - `used` (abi.md)
    - `crate_name`
    - `windows_subsystem`
- Code Generation  
    - `inline`
    - `cold`
    - `no_builtins`
    - `target_feature` (undocumented)
- Compiler Settings  
    - `feature` (undocumented)
    - `recursion_limit`
    - `type_length_limit` (undocumented)
- Documentation  
     - `doc` (rustdoc book)
- Preludes  
    - `no_std` (crates-and-source-files.md)
    - `no_implicit_prelude` (modules.md)
- Deprecation  
    - `deprecated`  (maybe move this to lints or documentation?)
- `path`  (modules.md)
- Runtime  
    - `panic_handler` (runtime.md)
    - `global_allocator` (??? is this runtime?)
- Tool

I'm thinking maybe just moving some of these to dedicated pages (testing, lints, code generation, etc.)  

Any thoughts?  Ideas on how to better organize?  Does this sound terrible?

ehuss:
- Move "Attributes" out of the "Items" chapter.
- Move each attribute out of `attributes.md`, scattered to various locations throughout the reference, including a few new dedicated pages. Each attribute gets a dedicated section to make it easier to link to, and enough room to fully document.
- See `attributes.md` for the consolidated built-in index.
- Document the syntax for the inputs of every built-in attribute.
- Minor updates and additions to various attributes.
- Removed `no_start`, it is no longer used. Removed when runtime/green threads were removed rust-lang/rust#18967.

Things reserved for **future PRs**:
- Expanding individual attributes. Some are too brief or slightly wrong.
- Documenting the remaining undocumented attributes.
- Finish the "Attributes on" sections to specify where each attribute may be used.

Some of the new locations seem a little awkward or arbitrary, but I couldn't think of better choices. I wanted to open this up to start getting feedback.

This is written assuming https://github.com/rust-lang/rust/pull/58899 will get merged. Also, some attributes are not very strict about validating their inputs (for example, `link` ignores any unknown field). I documented with what I think they are intended, not what they actually accept.

Closes #534


src/attributes.md
```diff
@@ -582,52 +172,133 @@ pub fn f() {}
 
 > Note: `rustc` currently recognizes the tools "clippy" and "rustfmt".
 
+## Built-in attributes
+
+The following is an index of all built-in attributes.
+
+- Conditional Compilation
```
Havvy: We should follow the style of headings on the page, and lowercase the non-first word.

src/codegen.md
```diff
@@ -0,0 +1,64 @@
...
+compile your code in a way that may be faster than what it would do without
+the hint. The attributes are only suggestions, and the compiler may choose to
+ignore it.
+
+Both attributes can be used on closures and [functions]. When applied to a
+function in a [trait], they apply only to that function when used as a default
+function for a trait implementation and not to all trait implementations. The
+attributes have no effect on a trait function without a body.
+
+<!-- TODO: I believe it is currently not possible to use these on closures
```
Havvy: I can't find a way to get the cold attribute onto a closure. So removing it seems fine. Do it in it its own PR after this one though (or file an issue)

src/codegen.md
```diff
@@ -0,0 +1,64 @@
...
+       call does not break any live ranges in the caller side."
+-->
+
+## The `no_builtins` attribute
+
+The *`no_builtins` [attribute]* may be applied at the crate level to disable
+optimizing certain code patterns to invocations of library functions that are
+assumed to exist.
+
+<!-- TODO: This needs more information.
```
Havvy: If this is usable in stable Rust, its effects should be documented here. Where rustc uses it, the rustc book should point here / discuss how that effects things there.

src/compiler-attrs.md
```diff
@@ -0,0 +1,18 @@
+# Compiler attributes
```
Havvy: Once again going to say I don't like the name of this section. I think `crate_name` could be added here, but need to know more about what it actually does to be fully certain. If anything "Compile setting attributes" would be a better name.

src/compiler-attrs.md
```diff
@@ -0,0 +1,18 @@
...
+
+The following [attributes] affect internal settings of the compiler.
+
+<!-- TODO: feature -->
```
Havvy: I've been thinking about `feature` while sleeping, and think that while we should include it in the index, we should do so as a note and point to the rustc book. We should otherwise not discuss it since it will never be a part of stable Rust.

src/derive.md
```diff
@@ -0,0 +1,43 @@
+# Derive
+
+<!-- TODO: Maybe link to https://doc.rust-lang.org/book/appendix-03-derivable-traits.html ?
+Should the reference say more about which traits are derivable?
```
Havvy: This TODO should be its own issue. It's separate from the rest of this. The list of derivable traits should indeed be in the reference standalone from the book.

src/derive.md
```diff
@@ -0,0 +1,43 @@
...
+
+    fn ne(&self, other: &Foo<T>) -> bool {
+        self.a != other.a || self.b != other.b
+    }
+}
+```
+
+You can implement `derive` for your own traits through [procedural macros].
+
+<!-- TODO: automatically_derived -->
```
Havvy: These are documented [here](https://doc.rust-lang.org/nightly/reference/special-types-and-traits.html#auto-traits), and not part of the derive macro system.

src/diagnostics.md
```diff
@@ -0,0 +1,271 @@
+# Diagnostics
+
```
Havvy: A paragraph summarizing the contents of this page would be helpful. If not part of this PR, then please open an issue for it.

src/items/external-blocks.md
```diff
@@ -42,16 +42,16 @@ Functions within external blocks may be called by Rust code, just like
...
+extern "abi" for<'l1, ..., 'lm> fn(A1, ..., An) -> R`, where `'l1`, ... `'lm`
+are its lifetime parameters, `A1`, ..., `An` are the declared types of its
+parameters and `R` is the declared return type.
 
-```rust,ignore
-extern {
-    fn foo(x: i32, ...);
-}
-```
+It is `unsafe` to access a static item declared in an extern block, whether or
```
Havvy: "whether it's mutable or not" is what I expected to read. Not sure if that's just me or if this is incorrect grammar.

src/runtime.md
```diff
@@ -51,7 +51,28 @@ defaults to unwinding the stack but that can be [changed to abort the
 process][abort]. The standard library's panic behavior can be modified at
 runtime with the [set_hook] function.
 
+## The `global_allocator` attribute
+
+The *`global_allocator` attribute* is used on a static item implementing the
```
Havvy: Link for "static item"

src/runtime.md
```diff
@@ -51,7 +51,28 @@ defaults to unwinding the stack but that can be [changed to abort the
...
+[`GlobalAlloc`] trait to set the global allocator.
+
+## The `windows_subsystem` attribute
+
+The *`windows_subsystem` attribute* may be applied at the crate level to
+indicate that when this crate is linked for a Windows target it will configure
+the resulting binary's [subsystem] via the linker. It uses the
+[_MetaNameValueStr_] syntax specify the subsystem with a value of either
+`console` or `windows`. More subsystems may be allowed in the future, and this
```
Havvy: This last sentence talks about the future in the first part and can be removed. The second part is an inverse of part of the first sentence. It can be removed.

src/testing.md
```diff
@@ -0,0 +1,88 @@
+# Testing
+
```
Havvy: Again, another paragraph for what to expect on this page would be nice. Again, doesn't need to be in this PR.

src/unsafety.md
```diff
@@ -14,7 +14,7 @@ Rust:
 
 [`Copy`]: special-types-and-traits.html#copy
 [`union`]: items/unions.html
-[mutable static variable]: items/static-items.html#mutable-statics
-[external static variable]: items/external-blocks.html
+[mutable]: items/static-items.html#mutable-statics
```
Havvy: Can you move the change to this page to its own PR so we can merge it now?

src/items/external-blocks.md
ehuss: [This](https://www.dictionary.com/browse/whether-or-not) suggests that it can go either way. I lean towards keeping it together as a single phrase "whether or not", since that is a common idiom.

src/derive.md
ehuss: This refers to the `#[automatically_derived]` attribute which is added to derived impls.  It may have been accidentally leaked to stable, but now that it is in use by several crates (see https://github.com/rust-lang-nursery/reference/issues/532#issuecomment-471086171) it may need to be documented.

src/compiler-attrs.md
Centril: I'd suggest dropping "compiler" altogether and just have "Recursion limits" as the title until we have more attributes that fit the more broader "settings" title. Otherwise I'd move `type_length_limit` into the type system sections and the other to the macro sections.

src/abi.md
```diff
@@ -56,7 +56,40 @@ $ nm -C foo.o
 0000000000000000 T foo::quux
 ```
 
+## The `no_mangle` attribute
```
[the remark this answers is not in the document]
ehuss: I like the idea.  However, we currently don't have complete documentation on the "allowed on" for every attribute.  Havvy will be working on it, so maybe we can consider this once we have all the information?

I think the "allowed on" is particularly difficult to document (like you noticed with `link_section`).  Some attributes are allowed everywhere, including places where they are silently ignored or don't make sense.  Should the documentation only specify where they are intended to be used?  Some of it is complicated by the `unused_attributes` lint not always working, and many attributes don't validate where they are located.

Some attributes like `no_mangle` apply in places where the effects are not clear. For example, at the crate level, it seems to affect the resulting object file, but I can't tell exactly how.

src/attributes.md
```diff
@@ -582,52 +172,133 @@ pub fn f() {}
...
+  - [`macro_use`] — Expands macro visibility, or imports macros from other
+    crates.
+  - [`proc_macro`] — Defines a function-like macro.
+  - [`proc_macro_derive`] — Defines a derive macro.
+  - [`proc_macro_attribute`] — Defines an attribute macro.
+- Diagnostics
+  - [`allow`], [`warn`], [`deny`], [`forbid`] — Alters the default lint level.
+  - [`deprecated`] — Generates deprecation notices.
+  - [`must_use`] — Generates a lint for unused values.
+- ABI, Linking, Symbols, and FFI
```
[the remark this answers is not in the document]
ehuss: :) I was wondering if anyone would notice that.  The sections are ordered based on how many crates on crates.io use the attributes.  It's somewhat arbitrary, and not really intended as a formal ordering (mostly just "important things towards the top").  I liked the idea of grouping them, but alphabetizing the groups didn't seem satisfying.  I'm up for any ordering, though.

src/codegen.md
```diff
@@ -0,0 +1,64 @@
+# Code generation
+
+<!-- TODO: target_feature -->
+
+## Optimization hints
+
+The `cold` and `inline` [attributes] give suggestions to the compiler to
```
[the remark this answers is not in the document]
ehuss: Absolutely should be in second person. I mostly tried to avoid editing the existing documentation, since I wanted to focus on just reorganization and adding the built-in syntax.  I also agree with the compiler/interpreter neutrality.

src/abi.md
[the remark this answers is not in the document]
Centril: > Havvy will be working on it, so maybe we can consider this once we have all the information?

👍 I'm very happy to iterate :) Perhaps open an issue about this?

> I think the "allowed on" is particularly difficult to document (like you noticed with `link_section`).

Agreed! But I think we should try nonetheless ;)

> Should the documentation only specify where they are intended to be used?

I would document both where it *has an effect* and where it is allowed as well. The latter is more a question of static semantics while the former varies wrt. static / dynamic semantics.

> Some of it is complicated by the `unused_attributes` lint not always working, and many attributes don't validate where they are located.

Bugs are bugs; we can document them with notes and links to issues imo. Having a fixme here or there is fine, we even note that this reference is a work in progress so it's all good.

> Some attributes like `no_mangle` apply in places where the effects are not clear.

I think again best effort applies here. We can even note that it is a best effort when we believe so.

src/attributes.md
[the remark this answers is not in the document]
Centril: =P

>  The sections are ordered based on how many crates on crates.io use the attributes.

This was not a possibility I had considered 😂 

> I liked the idea of grouping them, but alphabetizing the groups didn't seem satisfying.

Yeah I alphabetized in my head but it felt unsatisfying to me also.

>  I'm up for any ordering, though.

Cool, let's perhaps file as an issue to consider later?

src/conditional-compilation.md
```diff
@@ -167,7 +167,7 @@ Example values:
 ### `test`
```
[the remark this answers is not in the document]
ehuss: It is [built in](https://github.com/rust-lang/rust/blame/52e885628e4317aa3f158622435927eb29b812e9/src/librustc/session/config.rs#L1512-L1514).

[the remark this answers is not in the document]
Centril: Oh I see, it's injected into the user cfgs but not directly part of libsyntax cfg machinery. Thanks

src/diagnostics.md
```diff
@@ -0,0 +1,271 @@
...
+    pub fn undocumented_too() -> i32 { 2 }
+}
+```
+
+### Tool lint attributes
+
+Tool lints let you use scoped lints, to `allow`, `warn`, `deny` or `forbid` lints of
+certain tools.
+
+Currently `clippy` is the only available lint tool.
```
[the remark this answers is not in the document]
ehuss: Yep https://github.com/rust-lang/rust/blame/52e885628e4317aa3f158622435927eb29b812e9/src/libsyntax/attr/mod.rs#L62-L64

src/diagnostics.md
```diff
@@ -0,0 +1,271 @@
...
+#[forbid(missing_docs)]
+pub mod m3 {
+    // Attempting to toggle warning signals an error here
+    #[allow(missing_docs)]
+    /// Returns 2.
+    pub fn undocumented_too() -> i32 { 2 }
+}
+```
+
+### Tool lint attributes
```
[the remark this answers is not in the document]
ehuss: No, resolution is not really documented at all.  I'm still intending to make progress on it, but it's a big project.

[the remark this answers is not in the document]
Centril: Cool, looking forward to it :)

src/diagnostics.md
```diff
@@ -0,0 +1,271 @@
...
+
+The [RFC][1270-deprecation.md] contains motivations and more details.
+
+[1270-deprecation.md]: https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md
+
+
+## The `must_use` attribute
+
+The *`must_use` attribute* can be used on user-defined composite types
+([`struct`s][struct], [`enum`s][enum], and [`union`s][union]) and [functions].
```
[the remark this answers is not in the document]
ehuss: I can't quite figure out how to get the must_use lint to trigger for a trait object (it works fine for impl trait).  rust-lang/rust#55506 implies something like `Box<dyn Trait>` would work, but I can't get it to trigger (I think the code that handles it is [here](https://github.com/rust-lang/rust/blame/52e885628e4317aa3f158622435927eb29b812e9/src/librustc_lint/unused.rs#L81-L91)). Can you make an example that triggers for a trait object?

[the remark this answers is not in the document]
Centril: That is curious... I cannot produce a warning even with `unsized_locals`.. cc @varkor.

src/attributes/limits.md
```diff
@@ -1,6 +1,6 @@
-# Compiler setting attributes
+# Limits
```
Centril: 👍

src/attributes.md
```diff
@@ -229,10 +224,16 @@ The following is an index of all built-in attributes.
...
+- Limits
+  - [`recursion_limit`] — Sets the maximum recursion limit for certain
+    compile-time operations.
 - Runtime
   - [`panic_handler`] — Sets the function to handle panics.
   - [`global_allocator`] — Sets the global memory allocator.
   - [`windows_subsystem`] — Specifies the windows subsystem to link with.
+- Features
+  - `feature` — Used to enable unstable or experimental compiler features. See
+    [The Unstable Book] for features implemented in `rustc`.
```
Centril: I think perhaps `B-unstable` is more accurate. Consider linking to it as well?

src/diagnostics.md
```diff
@@ -0,0 +1,271 @@
+# Diagnostics
+
+## Lint check attributes
+
+A lint check names a potentially undesirable coding pattern, such as
+unreachable code or omitted documentation. The lint attributes `allow`,
+`warn`, `deny`, and `forbid` use the [_MetaListPaths_] syntax to specify a
+list of lint names to change the lint level for the entity to which the
+attribute applies.
```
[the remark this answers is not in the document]
ehuss: I think it should be documented.  However, I'm not getting the behavior you describe.  Lints don't seem to fire at all.  For example:
```rust
// Crate bar
#[macro_export]
macro_rules! thing {
    () => {
        #[no_mangle]
        const FOO: i32 = 5;
    }
}

// Crate A
bar::thing!{}  // No errors or warnings.
```

ehuss:
I think I got most of Centril's points. I renamed the section with no name to "Limits".  Still can't think of anything better.

The main points still not addressed:
- `must_use` with traits needs an explanation (waiting for clarification on trait objects)
- lint controls should document how they behave with macros.  

I'll file some issues for some of the todo items soon.

src/attributes/diagnostics.md
```diff
@@ -193,21 +190,21 @@ when the call expression is a function from an implementation of the trait.
...
 fn main() {
-  // Violates the `unused_must_use` lint.
-  5i32.use_me();
+    // Violates the `unused_must_use` lint.
+    5i32.use_me();
 }
 ```
 
-When used on a function in an implementation, the attribute does nothing.
+When used on a function in a trait implementation, the attribute does nothing.
```
Centril: (The below comment is not relevant for this PR)

@varkor @oli-obk We could reasonably make this do something by having the `#[must_use]` trigger when the concrete type is known? E.g.

```rust
trait A { fn b() -> usize; }

struct C;

impl A for C {
    #[must_use]
    fn b() -> usize { 0 }
}

fn main() {
    C::b();
//  ^--- We know statically that `C::b` references the function
//       on `impl A for C { ... }` and so we may trigger the lint...?
}
```

src/diagnostics.md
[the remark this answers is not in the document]
Centril: Here's a reproducer:
```rust
#![feature(unsized_locals)]

#[must_use] trait Foo {}

impl<T> Foo for T {}

fn _mk() -> Box<dyn Foo> { unimplemented!() }

fn main() {
    *_mk();
}
```

I would be surprised if it could be triggered without `unsized_locals`.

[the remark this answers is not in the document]
ehuss: I added documentation just for impl trait.  I tried a few times to rewrite the entire section, because the repetition of "expression of an expression statement" was bothering me.  However, each time I tried it seemed to end up worse.  If someone wants to tackle simplifying it, I think that would be good.  Generally it is always an "expression of an expression statement", with different interpretations based on type vs call result.

There are also a lot of other things the `must_use` checks handle, such as uninhabited types and binary operators.  However, that seems to be mostly applicable to std implementations or unstable features.  There is likely more to be said, but I couldn't figure out what should be said.

https://github.com/rust-lang/rust/blob/2c8bbf50db0ef90a33f986ba8fc2e1fe129197ff/src/librustc_lint/unused.rs#L49-L197

src/diagnostics.md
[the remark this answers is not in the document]
ehuss: Filed #544 to address this later.

src/codegen.md
```diff
@@ -0,0 +1,64 @@
...
+* `#[inline]` hints the compiler to perform an inline expansion.
+* `#[inline(always)]` asks the compiler to always perform an inline expansion.
+* `#[inline(never)]` asks the compiler to never perform an inline expansion.
+
+### The `cold` attribute
+
+The *`cold` [attribute]* suggests to the compiler that the attributed function or
+closure is unlikely to be called.
+<!-- TODO: Expand this section.
+    Should the exact semantics be documented here, or in rustc book?
```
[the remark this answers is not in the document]
ehuss: Filed #543 to address this later.

src/abi.md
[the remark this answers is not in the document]
ehuss: Noted at https://github.com/rust-lang-nursery/reference/issues/355#issuecomment-473576828

src/derive.md
ehuss: Filed #540

src/codegen.md
ehuss: Filed #542.

src/attributes.md
```diff
@@ -104,456 +105,45 @@ grammar:
...
-}
-```
+An attribute is either active or inert. During attribute processing, *active
+attributes* remove themselves from the thing they are on while *inert attributes*
+stay on.
 
-You can implement `derive` for your own traits through [procedural macros].
+The [`cfg`] and [`cfg_attr`] attributes are active. The [`test`] attribute is
+inert when compiling for tests and active otherwise. [Attribute macros] are
+active. All other attributes are inert.
```
petrochenkov: `test` is always active, when not compiling for tests it transforms its input into nothing and disappears itself too.
Also, `derive` is an active attribute too.

PR #539, opened 2019-03-15
Remove undocumented page

### LINKED ISSUE #536: Remove undocumented page?
I feel like the [undocumented page](https://github.com/rust-lang-nursery/reference/blob/master/src/undocumented.md) is not particularly useful. It is very incomplete, and not kept up-to-date. I'd like to propose removing it, and just adding a link to the issue tracker in the introduction. Taking a close look at the current items:

- [`libstd` facade] — Looks like this RFC never hit its potential, I consider it not relevant.
- [Trait reform] – Somewhat documented (coherence, orphan rules, method resolution order, Self types are all at least partially covered).
- [Flexible target specification] — I believe conditional compilation is now complete (except target_feature).
- [`dllimport`] — I have expanded slightly docs on `link`, but it could explain more.
- [define `crt_link`] — Don't think this is relevant.
- `impl Trait` — Tracked in #276.

[`libstd` facade]: https://github.com/rust-lang/rfcs/pull/40
[Trait reform]: https://github.com/rust-lang/rfcs/pull/48
[Attributes]: attributes.html
[Flexible target specification]: https://github.com/rust-lang/rfcs/pull/131
[Conditional compilation]: attributes.html#conditional-compilation
[`dllimport`]: https://github.com/rust-lang/rfcs/pull/1717
[FFI attributes]: attributes.html#ffi-attributes
[define `crt_link`]: https://github.com/rust-lang/rfcs/pull/1721

Agree/disagree?

Havvy:
We're not keeping this page up to date, so it's only a small
portion of undocumented content. This is misleading because it
makes readers think that this is the only information not
documented.

Fixes #536


ehuss:
Yay for linkchecker 😜 .  I was thinking the intro could just link to https://github.com/rust-lang-nursery/reference/issues.  I also noticed another stray link "big issue" at the bottom of the intro page that should be removed. I wish pulldown-cmark would warn about those.

src/introduction.md
```diff
@@ -12,7 +12,7 @@ provides three kinds of material:
 <div class="warning">
 
 Warning: This book is incomplete. Documenting everything takes a while. See
-the [undocumented] page for what is not documented in this book.
+the [github issues] for what is not documented in this book.
```
ehuss:
```suggestion
the [GitHub issues] for what is not documented in this book.
```

PR #545, opened 2019-03-18
Document `target_feature` and `cfg_target_feature`.

### LINKED ISSUE #286: target_feature (RFC 2045, RFC 2325)
Stabiliing soon (rust-lang/rust#48556)

- [ ] Document these attributes (with examples)
- [ ] Document list of stable target features ?

ehuss:
- Stabilization PR: https://github.com/rust-lang/rust/pull/49664
- [RFC 2045](https://github.com/rust-lang/rfcs/blob/master/text/2045-target-feature.md) `target_feature` and `cfg_target_feature`
    - Tracking issue: https://github.com/rust-lang/rust/issues/44839
    - Tracking issue `cfg_target_feature`: https://github.com/rust-lang/rust/issues/29717
- [RFC 2325](https://github.com/rust-lang/rfcs/blob/master/text/2325-stable-simd.md) Stable SIMD
    - Tracking issue: https://github.com/rust-lang/rust/issues/48556

Closes #286

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -86,7 +86,7 @@ Feature     | Implicitly Enables | Description
 `rdrand`    |          | [`rdrand`] — Read random number
 `rdseed`    |          | [`rdseed`] — Read random seed
 `sha`       | `sse2`   | [SHA] — Secure Hash Algorithm
-`sse`       |          | [SSE] — Streaming SIMD Extensions
+`sse`       |          | [SSE] — Streaming <abbr title="Single Instruction Multiple Data">SIMD</abbr> Extensions
 `sse2`      | `sse`    | [SSE2] — Streaming SIMD Extensions 2
 `sse3`      | `sse2`   | [SSE3] — Streaming SIMD Extensions 3
 `sse4.1`    | `sse3`   | [SSE4.1] — Streaming SIMD Extensions 4.1
@@ -96,6 +96,8 @@ Feature     | Implicitly Enables | Description
 `xsavec`    |          | [`xsavec`] — Save processor extended states with compaction
 `xsaveopt`  |          | [`xsaveopt`] — Save processor extended states optimized
 `xsaves`    |          | [`xsaves`] — Save processor extended states supervisor
+
+<!-- Keep links near each table to make it easier to move and update. -->
 
 [AES]: https://en.wikipedia.org/wiki/AES_instruction_set
 [AVX]: https://en.wikipedia.org/wiki/Advanced_Vector_Extensions
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -42,7 +42,110 @@ The *`no_builtins` [attribute]* may be applied at the crate level to disable
 optimizing certain code patterns to invocations of library functions that are
 assumed to exist.
 
+## The `target_feature` attribute
+
+The *`target_feature` [attribute]* may be applied to an [unsafe function] to
+enable code generation of that function for specific platform architecture
+features. It uses the [_MetaListNameValueStr_] syntax with a single key of
+`enable` whose value is a string of comma-separated feature names to enable.
+
+```rust,ignore
+#[target_feature(enable = "avx2")]
+unsafe fn foo_avx2() {}
+```
+
+Each [target architecture] has a set of features that may be enabled. It is an
+error to specify a feature for a target architecture that the crate is not
+being compiled for.
+
+It is [undefined behavior] to call a function that is compiled with a feature
+that is not supported on the current platform the code is running on.
+
+Functions marked with `target_feature` are not inlined into a context that
+does not support the given features. The `#[inline(always)]` attribute may not
+be used with a `target_feature` attribute.
+
+### Available features
+
+The following is a list of the available feature names.
+
+#### `x86` or `x86_64`
+
+Feature     | Implicitly Enables | Description
+------------|--------------------|-------------------
+`aes`       | `sse2`   | [AES] — Advanced Encryption Standard
+`avx`       | `sse4.2` | [AVX] — Advanced Vector Extensions
+`avx2`      | `avx`    | [AVX2] — Advanced Vector Extensions 2
+`bmi1`      |          | [BMI1] — Bit Manipulation Instruction Sets
+`bmi2`      |          | [BMI2] — Bit Manipulation Instruction Sets 2
+`fma`       | `avx`    | [FMA3] — Three-operand fused multiply-add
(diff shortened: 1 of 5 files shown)


src/attributes/codegen.md
```diff
@@ -42,7 +42,110 @@ The *`no_builtins` [attribute]* may be applied at the crate level to disable
...
+`bmi2`      |          | [BMI2] — Bit Manipulation Instruction Sets 2
+`fma`       | `avx`    | [FMA3] — Three-operand fused multiply-add
+`fxsr`      |          | [`fxsave`] and [`fxrstor`] — Save and restore x87 FPU, MMX Technology, and SSE State
+`lzcnt`     |          | [`lzcnt`] — Leading zeros count
+`pclmulqdq` | `sse2`   | [`pclmulqdq`] — Packed carry-less multiplication quadword
+`popcnt`    |          | [`popcnt`] — Count of bits set to 1
+`rdrand`    |          | [`rdrand`] — Read random number
+`rdseed`    |          | [`rdseed`] — Read random seed
+`sha`       | `sse2`   | [SHA] — Secure Hash Algorithm
+`sse`       |          | [SSE] — Streaming SIMD Extensions
```
Havvy:
```suggestion
`sse`       |          | [SSE] — Streaming <attr title="Single Instruction Multiple Data">SIMD</attr> Extensions
```

src/attributes/codegen.md
```diff
@@ -42,7 +42,110 @@ The *`no_builtins` [attribute]* may be applied at the crate level to disable
...
+`sse3`      | `sse2`   | [SSE3] — Streaming SIMD Extensions 3
+`sse4.1`    | `sse3`   | [SSE4.1] — Streaming SIMD Extensions 4.1
+`sse4.2`    | `sse4.1` | [SSE4.2] — Streaming SIMD Extensions 4.2
+`ssse3`     | `sse3`   | [SSSE3] — Supplemental Streaming SIMD Extensions 3
+`xsave`     |          | [`xsave`] — Save processor extended states
+`xsavec`    |          | [`xsavec`] — Save processor extended states with compaction
+`xsaveopt`  |          | [`xsaveopt`] — Save processor extended states optimized
+`xsaves`    |          | [`xsaves`] — Save processor extended states supervisor
+
+[AES]: https://en.wikipedia.org/wiki/AES_instruction_set
```
Havvy: I'm okay with leaving these links here instead of at the bottom, but please leave a comment with rationale.

src/behavior-considered-undefined.md
```diff
@@ -38,6 +38,8 @@ code.
   * A discriminant in an `enum` not included in the type definition.
   * A value in a `char` which is a surrogate or above `char::MAX`.
   * Non-UTF-8 byte sequences in a `str`.
+* Executing code compiled with platform features that the current platform
```
Havvy:
```suggestion
  * Executing code compiled with platform features that the current platform
```

Missing leading spaces.

src/conditional-compilation.md
```diff
@@ -90,6 +90,26 @@ Example values:
...
+* `"crt-static"`
+* `"rdrand"`
+* `"sse"`
+* `"sse2"`
+* `"sse4.1"`
+
+See the [`target_feature` attribute] for more details on the available
+features. An additional feature of `crt-static` is available to the
+`target_feature` option to indicate that a static C runtime is available, see
+[static and dynamic C runtimes] for more information.
```
Havvy: Just make "static C runtime" the link instead of this "See more".

src/attributes/codegen.md
ehuss: `abbr` I assume you mean?

src/behavior-considered-undefined.md
ehuss: I'm not sure I understand. The indented list is of the previous thing ("Invalid values in primitive types").

ehuss:
> I think that platform features should be discussed outside

I wouldn't mind rearranging things, though I don't quite see what it would look like.  I am concerned that the tables will get quite large as more features are stabilized.  I'm not sure what else can be said about platform features, though.

ehuss:
Oh, I was also wondering about pinging some of the people for review who know this stuff well (like alex or gnzlbg).  The "Implicitly Enables" was particularly difficult to figure out. I think it is correct, but they should know more.

src/attributes/codegen.md
```diff
@@ -42,7 +42,112 @@ The *`no_builtins` [attribute]* may be applied at the crate level to disable
...
+It is [undefined behavior] to call a function that is compiled with a feature
+that is not supported on the current platform the code is running on.
+
+Functions marked with `target_feature` are not inlined into a context that
+does not support the given features. The `#[inline(always)]` attribute may not
+be used with a `target_feature` attribute.
+
+### Available features
+
+The following is a list of the available feature names.
```
gnzlbg: Is the table below only intended to document the stable target features? 

I know it wouldn't be great to link to the compiler implementation as the source of documentation, and I'll try to keep this in sync as new PRs are merged, but particularly for those who want to learn about the unstable features, the compiler source might be useful (e.g. here: https://github.com/rust-lang/rust/blob/master/src/librustc_codegen_llvm/llvm_util.rs#L96).

src/attributes/codegen.md
```diff
@@ -42,7 +42,112 @@ The *`no_builtins` [attribute]* may be applied at the crate level to disable
...
+#[target_feature(enable = "avx2")]
+unsafe fn foo_avx2() {}
+```
+
+Each [target architecture] has a set of features that may be enabled. It is an
+error to specify a feature for a target architecture that the crate is not
+being compiled for.
+
+It is [undefined behavior] to call a function that is compiled with a feature
+that is not supported on the current platform the code is running on.
```
gnzlbg: I don't know whether unsafe superpowers are documented in the reference, but being able to call a target-feature function without proving that the platform the code is running on supports the feature is an unsafe super power.

gnzlbg: So forget about this, the the unsafe super-powers are documented here: https://doc.rust-lang.org/reference/unsafety.html#unsafety but they already contain "calling unsafe functions" which is what applies for these.

We'll only need to update this list if the target-feature 1.1 RFC gets merged.

src/behavior-considered-undefined.md
Havvy: Ah. That's what happens when I don't look at the whole context.

src/attributes/codegen.md
Havvy: Everything in the reference documents what is stable (as of the current nightly; the reference updates on the website alongside rustc releases)

For unstable things, there's the unstable book (https://github.com/rust-lang/rust/tree/master/src/doc/unstable-book).

Havvy:
💟 Thanks!

Issue #532, opened 2019-03-08
Undocumented Attributes


ehuss:
There are a few attributes that are currently not documented. I wanted to get feedback if these *should* be documented.  I've included my inclination, but I'd like to hear what others think.

- [x] [`automatically_derived`](https://github.com/rust-lang/rust/pull/12929) — I vote "probably"?  It has some use in the wild. I think the only reason to use it is for rustdoc. Are there other reasons to use it?
- [x] `feature` — I vote "maybe" 🤔?  Although this is a nightly-oriented attribute, it is very widely used.
- [x] [`type_length_limit`](https://github.com/rust-lang/rust/pull/37789) — I vote "yes".  I think it is analogous to `recursion_limit`, and I suspect it just fell through the cracks.
- `bench` — I vote "no". It requires the unstable `Bencher` to be used. (Now destabilized via https://github.com/rust-lang/rust/pull/62507)
- [`reexport_test_harness_main`](https://github.com/rust-lang/rust/pull/15964) — I vote "no". Even though it is usable on stable, I think it is an internal implementation detail, and I do not see anyone using it.

These are just waiting to be finished:
- [x] `panic_handler` (#362)
- [x] `used` (#361)
- [x] `target_feature` (#286)

petrochenkov:
`automatically_derived` is also an implementation detail, I don't know why it's not feature gated.

Centril:
> [`automatically_derived`](https://github.com/rust-lang/rust/pull/12929) — I vote "probably"? It has some use in the wild. I think the only reason to use it is for rustdoc. Are there other reasons to use it?

What does the attribute do?

@petrochenkov Could you try a PR + crater to "destabilization"?

> `bench` — I vote "no". It requires the unstable `Bencher` to be used.

👍 

> `feature` — I vote "maybe" 🤔? Although this is a nightly-oriented attribute, it is very widely used.

👍 + note that it is unstable.

> [`reexport_test_harness_main`](https://github.com/rust-lang/rust/pull/15964) — I vote "no". Even though it is usable on stable, I think it is an internal implementation detail, and I do not see anyone using it.

What... @petrochenkov I'd also attempt destabilization for this...

> [`type_length_limit`](https://github.com/rust-lang/rust/pull/37789) — I vote "yes". I think it is analogous to `recursion_limit`, and I suspect it just fell through the cracks.

Sounds good.

> These are just waiting to be finished:

[...]

PR #546, opened 2019-03-18
Document `type_length_limit`.

ehuss:
cc #532

### DIFF
--- a/src/attributes/limits.md
+++ b/src/attributes/limits.md
@@ -4,10 +4,12 @@ The following [attributes] affect compile-time limits.
 
 ## The `recursion_limit` attribute
 
-The *`recursion_limit` attribute* may be applied at the crate level to set the
+The *`recursion_limit` attribute* may be applied at the [crate] level to set the
 maximum depth for potentially infinitely-recursive compile-time operations
 like auto-dereference or macro expansion. It uses the [_MetaNameValueStr_]
-syntax to specify the recursion depth. The default is 64.
+syntax to specify the recursion depth.
+
+> Note: The default in `rustc` is 64.
 
 ```rust,ignore
 #![recursion_limit = "4"]
@@ -26,10 +28,12 @@ a!{}
 
 ## The `type_length_limit` attribute
 
-The *`type_length_limit` attribute* limits the maximum size of a type
-constructed during monomorphization. It is applied at the crate level, and
-uses the [_MetaNameValueStr_] syntax to set the limit based on the number of
-type substitutions within the type. The default value is 1048576.
+The *`type_length_limit` attribute* limits the maximum number of type
+substitutions made when constructing a concrete type during monomorphization.
+It is applied at the [crate] level, and uses the [_MetaNameValueStr_] syntax
+to set the limit based on the number of type substitutions.
+
+> Note: The default in `rustc` is 1048576.
 
 ```rust,ignore
 #![type_length_limit = "8"]
@@ -44,5 +48,6 @@ struct C;
 drop::<Option<A>>(None);
 ```
 
-[attributes]: attributes.html
 [_MetaNameValueStr_]: attributes.html#meta-item-attribute-syntax
+[attributes]: attributes.html
+[crate]: crates-and-source-files.html
--- a/src/attributes/limits.md
+++ b/src/attributes/limits.md
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
 The *`recursion_limit` attribute* may be applied at the crate level to set the
 maximum depth for potentially infinitely-recursive compile-time operations
 like auto-dereference or macro expansion. It uses the [_MetaNameValueStr_]
-syntax to specify the recursion depth. The default is
-`#![recursion_limit="64"]`.
+syntax to specify the recursion depth. The default is 64.
+
+```rust,ignore
+#![recursion_limit = "4"]
+
+macro_rules! a {
+    () => { a!(1) };
+    (1) => { a!(2) };
+    (2) => { a!(3) };
+    (3) => { a!(4) };
+    (4) => { };
+}
+
+// This fails to expand because it requires a recursion depth greater than 4.
+a!{}
+```
+
+## The `type_length_limit` attribute
+
+The *`type_length_limit` attribute* limits the maximum size of a type
+constructed during monomorphization. It is applied at the crate level, and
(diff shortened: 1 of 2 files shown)


src/attributes/limits.md
```diff
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
 The *`recursion_limit` attribute* may be applied at the crate level to set the
 maximum depth for potentially infinitely-recursive compile-time operations
 like auto-dereference or macro expansion. It uses the [_MetaNameValueStr_]
-syntax to specify the recursion depth. The default is
-`#![recursion_limit="64"]`.
+syntax to specify the recursion depth. The default is 64.
```
Havvy: That default is rustc specific, and either be in a note or we should link to a corresponding page on the rustc book.

src/attributes/limits.md
```diff
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
...
+}
+
+// This fails to expand because it requires a recursion depth greater than 4.
+a!{}
+```
+
+## The `type_length_limit` attribute
+
+The *`type_length_limit` attribute* limits the maximum size of a type
+constructed during monomorphization. It is applied at the crate level, and
```
Havvy: "crate" should link to the crate and source files page.

I think we should say "concrete type" instead of "type constructed during monomorphization" and we should add "concrete type" to the glossary/make sure its defined in one of the types chapters. I'll file an issue for that one.

src/attributes/limits.md
```diff
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
...
+uses the [_MetaNameValueStr_] syntax to set the limit based on the number of
+type substitutions within the type. The default value is 1048576.
```
Havvy: Same comment as before on rustc defaults.

src/attributes/limits.md
```diff
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
...
+
+```rust,ignore
+#![type_length_limit = "8"]
+
+type A = (B, B, B);
+type B = (C, C, C);
+struct C;
```
Havvy: Can we use `u8` instead of `C` to show that it counts primitive built-in types as well?

src/attributes/limits.md
```diff
@@ -7,8 +7,42 @@ The following [attributes] affect compile-time limits.
...
+
+// This fails to compile because monomorphizing to
+// `drop::<Option<((C, C, C), (C, C, C), (C, C, C))>>` requires more than 8
+// type elements.
+drop::<Option<A>>(None);
```
Havvy: Needing Turbofish syntax and `drop` here makes this example a bit more confusing. I'd show an identity function or something equivalent.

src/attributes/limits.md
ehuss: I've reworded it in a way that is maybe a little clearer. I think it is important to make the distinction that it is the instantiation of an abstract type to a concrete one that triggers this check.

(Also, this is far beyond my knowledge of the compiler, so this is the best to my understanding of reading the code and writing tests.)

src/attributes/limits.md
ehuss: I'm not sure I fully understand. Do you mean something like this?

```rust
#![type_length_limit = "8"]

fn f<T>(x: T) {}

// This fails to compile because monomorphizing to
// `f::<((i32, i32, i32), (i32, i32, i32), (i32, i32, i32))>>` requires more
// than 8 type elements.
f(((1, 2, 3), (4, 5, 6), (7, 8, 9)));
```

As an aside, I was uneasy about adding any examples at all, since they don't work with rustdoc, and it's relatively obscure, and could change in the future.

src/attributes/limits.md
```diff
@@ -4,10 +4,12 @@ The following [attributes] affect compile-time limits.
 
 ## The `recursion_limit` attribute
 
-The *`recursion_limit` attribute* may be applied at the crate level to set the
+The *`recursion_limit` attribute* may be applied at the [crate] level to set the
 maximum depth for potentially infinitely-recursive compile-time operations
 like auto-dereference or macro expansion. It uses the [_MetaNameValueStr_]
```
Centril: Add an example (after macros) for auto-dereference since you reference it, e.g.
```rust
#![recursion_limit = "1"]

fn main() {
    (|_: &u8| {})(&&1);
}
```

Centril:
```suggestion
like macro expansion or auto-dereference. It uses the [_MetaNameValueStr_]
```
since macro expansion is much more common; I didn't even know auto deref was subject to this.

src/attributes/limits.md
Havvy: That would work as an example.

src/attributes/limits.md
ehuss: There are many processes which check the limit:

https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc/traits/project.rs#L349-L358
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc/traits/project.rs#L864-L868
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc/traits/select.rs#L1121-L1131
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc/traits/query/normalize.rs#L100-L108
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc/ty/layout.rs#L168-L174
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc_mir/monomorphize/collector.rs#L452-L460
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc_traits/dropck_outlives.rs#L162-L168
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/librustc_typeck/check/autoderef.rs#L52-L58
https://github.com/rust-lang/rust/blob/ef4d1c419525e1510aa650e2bec0d8f7449a2130/src/libsyntax/ext/expand.rs#L525-L537

PR #547, opened 2019-03-18
Add some terms to the glossary.

ehuss:


ehuss:
I'm not really comfortable writing these, but I think they would be useful, so I figured I'd give it a shot.

src/glossary.md
```diff
@@ -59,18 +59,32 @@ and functions that evaluate to a single value, with or without side-effects.
 
 For example, `2 + (3 * 4)` is an expression that returns the value 14.
 
+### Free function
```
Centril: This can be generalized to "free item", e.g. for a free const (not associated) or free static (we don't have associated statics). This should also be moved to after "associated item" and be worded *in opposition* to associated items.

src/glossary.md
```diff
@@ -59,18 +59,32 @@ and functions that evaluate to a single value, with or without side-effects.
...
+
+A [function] that is not a member of an [implementation].
+
+### Inherent implementation
+
+An [implementation] that applies to a nominal type, not to a trait.
```
Centril:
```suggestion
An [implementation] that applies to a nominal type, not to a trait-type pair.
```
(to disassociate from `impl dyn? Debug { ... }`)

src/glossary.md
Havvy: Unalphabetized glossaries are confusing.

Centril: Yes fair enough. Maybe link back and forth instead?

PR #549, opened 2019-03-19
Define free parameters and concrete

### LINKED ISSUE #548: Define "concrete"
We use it twice (soon to be thrice) but don't define it anywhere. It should be defined on the generics page as an item where all generic parameters have been instantiated with a non-generic type. The uses of "concrete" should link to this definition, and it should also be added to the glossary, again linking to the full definition.

We should also have examples that show that scalar types are concrete, types that don't have any generics are concrete, things like `Option<u8>` are concrete, and probably some counterexamples as well.

Havvy:
Fixes #548 

I'm dissatisfied by just having two definitions and then mostly showing examples. But still better than nothing.

And while they're called "free variables" elsewhere, while we say they're generic parameters, I'd also like to call them "free parameters". I'd be willing to change `generic parameter` -> `generic variable`.

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -22,11 +22,11 @@
 
 Functions, type aliases, structs, enumerations, unions, traits and
 implementations may be *parameterized* by types and lifetimes. These parameters
-are listed in angle <span class="parenthetical">brackets (`<...>`)</span>,
+are declared in angle <span class="parenthetical">brackets (`<...>`)</span>,
 usually immediately after the name of the item and before its definition. For
-implementations, which don't have a name, they come directly after `impl`.
-Lifetime parameters must be declared before type parameters. Some examples of
-items with type and lifetime parameters:
+implementations, which don't have a name, they are declared directly after the
+`impl` keyword. Lifetime parameters must be declared before type parameters.
+Some examples of items with type and lifetime parameters:
 
 ```rust
 fn foo<'a, T>() {}
@@ -82,6 +82,54 @@ where
     f: T,
 }
 ```
+
+## Free parameters and concrete items and types
+
+The *free parameters* of an item or type are the type and lifetime paremeters
+that refer to a generically declared parameter. A *concrete item* or *concrete
+type* is one that does not have any free parameters.
+
+For example, consider this trait:
+
+```rust
+# struct SomeStruct;
+#
+trait Example<A> {
+    fn foo(a: A)
+
+    fn bar<'b'>(a: A, b: &'b SomeStruct)
+
+    fn baz()
+
+    fn quux(a: i32, b: Option<i32>)
+}
+```
+
+In it, the trait `Example` has a free variable `A` because it declares it
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
+type of its first argument. `baz` has both `A` and `B` as free type parameters,
+getting `A` from the trait's parameters and defining `B` itself. Both `baz` and
+`quux` have no free type parameters and are thusly concrete.
+
+For another example, the function `for<'a> fn foo(a: &'a i32>` is concrete.
+
+For another example, the following table shows the free type parameters of
+various types. Types without free parmaeters show "concrete" instead of "none".
+Assume `A`, `E`, and `'a` are defined as generic parameters.
+
+| Type | Free Parameters |
+| `i32` | concrete | 
+| `&'a i32` | `'a` |
+| `&'static i32` | concrete |
+| `UserDefinedType` | concrete |
+| `GenericDefinedType<A>` | `A` |
+| `GenericDefinedType<i32>` | concrete |
+| `GenericDefinedType<Option<A>>` | `A` |
+| `GenericDefinedType<Result<&'a A, E>` | `'a`, `A`, `E` |
+| `GenericDefinedType<Option<UserDefinedType>` | concrete |
+
+> Note: In some programming languages and in type theory, concrete types are
+> called *ground types*.
 
 ## Attributes
 
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -87,7 +87,8 @@ where
 
 The *free variables* of an item or type are the type and lifetime variables
 that refer to a generically declared parameter. A *concrete item* or *concrete
-type* is one that does not have any free variables.
+type* is one that does not have any free variables. See [Wikipeida][concrete
+wikipedia] for more.
 
 For example, consider this trait:
 
@@ -159,6 +160,7 @@ generic parameter.
 [_TypeParamBounds_]: trait-bounds.html
 
 [arrays]: types/array.html
+[concrete wikipedia]: https://en.wikipedia.org/wiki/Free_variables_and_bound_variables
 [function pointers]: types/function-pointer.html
 [references]: types/pointer.html#shared-references-
 [raw pointers]: types/pointer.html#raw-pointers-const-and-mut
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -83,11 +83,11 @@ where
 }
 ```
 
-## Free parameters and concrete items and types
+## Free variables and concrete items and types
 
-The *free parameters* of an item or type are the type and lifetime paremeters
+The *free variables* of an item or type are the type and lifetime variables
 that refer to a generically declared parameter. A *concrete item* or *concrete
-type* is one that does not have any free parameters.
+type* is one that does not have any free variables.
 
 For example, consider this trait:
 
@@ -107,17 +107,18 @@ trait Example<A> {
 
 In it, the trait `Example` has a free variable `A` because it declares it
 itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
-type of its first argument. `baz` has both `A` and `B` as free type parameters,
-getting `A` from the trait's parameters and defining `B` itself. Both `baz` and
-`quux` have no free type parameters and are thusly concrete.
+type of its first argument. `baz` has both `A` and `B` as free type variables,
+getting `A` from the trait's generic parameters and defining `B` as a generic
+parmaeter itself. Both `baz` and `quux` have no free type variables and are
+thusly concrete.
 
 For another example, the function `for<'a> fn foo(a: &'a i32>` is concrete.
 
-For another example, the following table shows the free type parameters of
+For another example, the following table shows the free type variables of
 various types. Types without free parmaeters show "concrete" instead of "none".
 Assume `A`, `E`, and `'a` are defined as generic parameters.
 
-| Type | Free Parameters |
+| Type | Free Variables |
 | `i32` | concrete | 
 | `&'a i32` | `'a` |
 | `&'static i32` | concrete |
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -86,8 +86,8 @@ where
 ## Free variables and concrete items and types
 
 The *free variables* of an item or type are the type and lifetime variables
-that refer to a generically declared parameter. A *concrete item* or *concrete
-type* is one that does not have any free variables. See [Wikipeida][concrete
+not bound therein. These variables may be substituted for other types or lifetimes. A *concrete item* or *concrete
+type* is one without free variables. See [Wikipeida][concrete
 wikipedia] for more.
 
 For example, consider this trait:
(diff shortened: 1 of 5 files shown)


src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
 }
 ```
 
+## Free variables and concrete items and types
+
+The *free variables* of an item or type are the type and lifetime variables
+that refer to a generically declared parameter. A *concrete item* or *concrete
+type* is one that does not have any free variables. See [Wikipeida][concrete
+wikipedia] for more.
```
Centril: Some general notes about this paragraph that I think are not well captured:
+ Bounds/Constraints can also have free variables, e.g. `u8: 'a` has `'a` as a free variable.
+ A lifetime can be concrete: `'static`.

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
 }
 ```
 
+## Free variables and concrete items and types
+
+The *free variables* of an item or type are the type and lifetime variables
```
Centril:
```suggestion
In a given context, the *free variables* of an item or type are the type and lifetime variables
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+that refer to a generically declared parameter. A *concrete item* or *concrete
```
Centril:
```suggestion
not bound therein. These variables may be substituted for other types or lifetimes. A *concrete item* or *concrete
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+
+For example, consider this trait:
+
+```rust
+# struct SomeStruct;
+#
+trait Example<A> {
+    fn foo(a: A);
+
+    fn bar<'b'>(a: A, b: &'b SomeStruct);
```
Centril:
```suggestion
    fn bar<'b>(a: A, b: &'b SomeStruct);
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+
+    fn baz();
+
+    fn quux(a: i32, b: Option<i32>);
+}
+```
+
+In it, the trait `Example` has a free variable `A` because it declares it
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
+type of its first argument. `baz` has both `A` and `B` as free type variables,
```
Centril:
```suggestion
type of its first argument. The function `bar` has `A` as a free type variable,
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+getting `A` from the trait's generic parameters and defining `B` as a generic
```
Centril:
```suggestion
getting `A` from the trait's generic parameters. However, the lifetime `'b` is bound in `baz` but is free in `&'b SomeStruct`.
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+parmaeter itself. Both `baz` and `quux` have no free type variables and are
```
Centril:
```suggestion
Both `baz` and `quux` have no free type variables and are
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+thusly concrete.
```
Centril:
```suggestion
thus concrete.
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+
+For another example, the function `for<'a> fn foo(a: &'a i32>` is concrete.
```
Centril:
```suggestion
For another example, the function `for<'a> fn foo(a: &'a i32>` is concrete but in `&'a i32`, `'a` is a free variable.
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+
+For another example, the following table shows the free type variables of
+various types. Types without free parmaeters show "concrete" instead of "none".
+Assume `A`, `E`, and `'a` are defined as generic parameters.
```
Centril:
```suggestion
Assume that `A`, `E`, and `'a` are defined as generic parameters.
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
+type of its first argument. `baz` has both `A` and `B` as free type variables,
+getting `A` from the trait's generic parameters and defining `B` as a generic
+parmaeter itself. Both `baz` and `quux` have no free type variables and are
+thusly concrete.
+
+For another example, the function `for<'a> fn foo(a: &'a i32>` is concrete.
+
+For another example, the following table shows the free type variables of
+various types. Types without free parmaeters show "concrete" instead of "none".
```
Centril:
```suggestion
various types. Types without free parameters show "concrete" instead of "none".
```

src/items/generics.md
```diff
@@ -83,6 +83,56 @@ where
...
+Assume `A`, `E`, and `'a` are defined as generic parameters.
+
+| Type | Free Variables |
+| `i32` | concrete | 
+| `&'a i32` | `'a` |
+| `&'static i32` | concrete |
```
Centril: Would be good to have `&i32` and say that it has an (anonymous) free (inference) variable.

Havvy:
Rebased against master because of conflicts in `glossary.md`

src/items/generics.md
Havvy: How's that?

Centril: Well `&i32` really means `&'_ i32` and `_` is a placeholder for something that may be substituted or unified with something else, possibly concrete.

src/items/generics.md
```diff
@@ -83,6 +83,58 @@ where
...
+
+    fn quux(a: i32, b: Option<i32>);
+}
+```
+
+In it, the trait `Example` has a free variable `A` because it declares it
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
+type of its first argument. The function `bar` has `A` as a free type variable,
+getting `A` from the trait's generic parameters. However, the lifetime `'b` is
+bound in `baz` but is free in tye type `&'b SomeStruct`. Both `baz` and `quux`
```
Centril:
```suggestion
bound in `baz` but is free in the type `&'b SomeStruct`. Both `baz` and `quux`
```

src/items/generics.md
```diff
@@ -83,6 +83,58 @@ where
...
+#
+trait Example<A> {
+    fn foo(a: A);
+
+    fn bar<'b>(a: A, b: &'b SomeStruct);
+
+    fn baz();
+
+    fn quux(a: i32, b: Option<i32>);
+}
```
Centril:
```suggestion
}
```

src/items/generics.md
```diff
@@ -83,6 +83,58 @@ where
...
+```
+
+In it, the trait `Example` has a free variable `A` because it declares it
```
Centril:
```suggestion
In it, the trait `Example` has a bound variable `A` because it declares it
```

src/items/generics.md
```diff
@@ -83,6 +83,58 @@ where
...
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
+type of its first argument. The function `bar` has `A` as a free type variable,
+getting `A` from the trait's generic parameters. However, the lifetime `'b` is
+bound in `baz` but is free in tye type `&'b SomeStruct`. Both `baz` and `quux`
+have no free type variables and are thus concrete.
```
Centril:
```suggestion
```

src/items/generics.md
```diff
@@ -83,6 +83,58 @@ where
...
+    fn bar<'b>(a: A, b: &'b SomeStruct);
+
+    fn baz();
+
+    fn quux(a: i32, b: Option<i32>);
+}
+```
+
+In it, the trait `Example` has a free variable `A` because it declares it
+itself. Furthermore, `foo` has `A` as a free variable because it uses it as the
```
Centril:
```suggestion
itself. The trait also has `Self` as an implicitly bound variable. Furthermore, `foo` has `A` as a free variable because it uses it as the
```

src/items/generics.md
matthewjasper: Bound in `bar`

Centril: I don't follow...

Havvy:
The trait example needs to be rewritten, but that can be done as another PR. We also need an actual section on type and item constructors, but that should be its own PR. Further exposition on binders (and how unbinding through item declarations works) should also be done; also as its own PR? Perhaps next time I look at this, since the build just failed?

src/items/generics.md
```diff
@@ -83,6 +83,98 @@ where
...
+A formulae is *concrete*, or *closed*, if there are no free variables in it.
+See [Wikipedia][wikipedia free variables] for more.
+
+Examples of generic binders for type and lifetime variables are:
+
+* `for<'a>` in `for<'a> fn(&'a u8)`
+* `<T>` in `fn identity<T>(x: T) -> T { x }` for item constructors
+* [Traits] and [implementations] implicitly bind `Self`
+
+All type and lifetimes are unbound at the boundaries of items.
```
Centril: Also not sure what this sentence is saying...

src/items/generics.md
```diff
@@ -83,6 +83,98 @@ where
...
+
+Formulae nest. For example, in a function item, the entire function item is a
+formulae but so are the types of the parameters of the function. A type or
+lifetime variable may be free in one formulae but bound in a containing
+formulae. For example, in the function prototype `for<'a> fn foo(a: &'a i32>`,
```
Centril:
```suggestion
formulae. For example, in the function prototype `for<'a> fn foo(a: &'a i32)`,
```
"prototype"?

src/items/generics.md
```diff
@@ -83,6 +83,98 @@ where
...
+`'a` is bound by the `for<'a>` but in the formulae of the type
```
Centril:
```suggestion
`'a` is bound by the binder `for<'a>` but in the type
```

src/items/generics.md
```diff
@@ -83,6 +83,98 @@ where
...
+`&'a i32`, `'a` is a free variable.
+
+Furthermore, the same syntax may have free variables when looked at as one
+formulae while having no free variables when looked at as another formulae. Most
+commonly, the type or item will have a free variable when the type or item
+constructor does not. For example, `Option<T>` as a type has `T` as a free
+variable but as a type constructor, has no free variables.
```
Centril:
```suggestion
variable but as a type constructor, `Option` has no free variables.
```
(it's not the same type expression so it's not a valid example)

src/items/generics.md
Havvy: T'was the last thing I wrote before my brain turned off. I need to expand it, but it's the fact that in the following example, `CONST` has no free variables, whether implicitly `Self` from traits or the generic parameter `T` of the function itself:

```rust
fn generic<T>(t: T) {
    const CONST: () = ();
}
```

src/items/generics.md
Havvy: Name for everything of a function except the block expression. The part you can put in a trait. Do you have a better name?

src/items/generics.md
Centril: Ah yes; that's a good point, we don't have scoped type variables. Be careful not to imply that this applies to `bar` in `trait Foo { fn bar(); }` because `Foo` and `bar` are both in the syntactic category of items here.

src/items/generics.md
```diff
@@ -83,6 +83,98 @@ where
...
 ```
 
+## Formulae and Free Variables
+
+The syntatic locations a *variable* is allowed is a *formulae*. Formulae for
+type and lifetime variables are:
+
+* [Types]
+* Type constructors (for example, `Option` in `enum Option<T> { ... }`)
+* Lifetimes (only lifetime variables)
```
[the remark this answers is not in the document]
Centril: Ah; can you expand to "(only lifetime variables are allowed here)"?

src/items/generics.md
Centril: I'd call this "function signature" or "function declaration".

Centril:
cc @Havvy -- this is waiting for you if you want to pursue this further :)

Havvy:
Every time I get back to this, I always lead myself into an unhappy path. So closing.

PR #553, opened 2019-03-25
Document repr packed(N).

### LINKED ISSUE #483: Document `#[repr(packed(N))]`
- Version: 1.33
- PR: https://github.com/rust-lang/rust/pull/57049
- Tracking issue: https://github.com/rust-lang/rust/issues/33158
- RFC: https://github.com/rust-lang/rfcs/pull/1399

The implementation diverged from the RFC a bit...

cc @cramertj

ehuss:
Closes #483

I feel like this is lacking, but I'm not sure what else to add.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -110,10 +110,13 @@ Closures have no layout guarantees.
 ## Representations
 
 All user-defined composite types (`struct`s, `enum`s, and `union`s) have a
-*representation* that specifies what the layout is for the type.
+*representation* that specifies what the layout is for the type. The possible
+representations for a type are:
 
-The possible representations for a type are the default representation, `C`,
-the primitive representations, and `transparent`.
+- [Default]
+- [`C`]
+- The [primitive representations]
+- [`transparent`]
 
 The representation of a type can be changed by applying the `repr` attribute
 to it. The following example shows a struct with a `C` representation.
@@ -296,42 +299,28 @@ For all other enumerations, the layout is unspecified.
 
 Likewise, combining two primitive representations together is unspecified.
 
-### The `align` modifier
+### The alignment modifiers
 
-The `align` modifier can be used on `struct`s and `union`s to raise the
-alignment of the type to a given value.
+The `align` and `packed` modifiers can be used to respectively raise or lower
+the alignment of `struct`s and `union`s. `packed` may also alter the padding
+between fields.
 
 The alignment is specified as an integer parameter in the form of
-`#[repr(align(x))]`. The alignment value must be a power of two from 1 up to
-2<sup>29</sup>.
+`#[repr(align(x))]` or `#[repr(packed(x))]`. The alignment value must be a
+power of two from 1 up to 2<sup>29</sup>. For `packed`, if no value is given,
+as in `#[repr(packed)]`, then the value is 1.
 
-The `align` modifier raises the type's alignment to the specified alignment.
-If the specified alignment is less than the alignment of the type without the
-`align` modifier, then the alignment is unaffected.
+For `align`, if the specified alignment is less than the alignment of the type
+without the `align` modifier, then the alignment is unaffected.
 
-The `align` and `packed` modifiers cannot be applied on the same type and a
-`packed` type cannot transitively contain another `align`ed type. `align` may
-only be applied to the default and `C` representations.
-
-### The `packed` modifier
-
-The `packed` modifier can be used on `struct`s and `union`s to lower the
-alignment of the type to a given value.
-
-The packing value (hence: "value") is specified as an integer parameter in the form of
-`#[repr(packed(x))]`. If no value is given, as in `#[repr(packed)]`, then the
-value is 1. The value must be a power of two from 1 up to
-2<sup>29</sup>.
-
-The `packed` modifier lowers the type's alignment to the specified value. If
-the specified value is greater than the type's alignment without the
-`packed` modifier, then the alignment and layout is unaffected. The alignments
-of each field, for the purpose of positioning fields, is the smaller of the
-specified packing and the alignment of the field's type.
+For `packed`, if the specified alignment is greater than the type's alignment
+without the `packed` modifier, then the alignment and layout is unaffected.
+The alignments of each field, for the purpose of positioning fields, is the
+smaller of the specified alignment and the alignment of the field's type.
 
 The `align` and `packed` modifiers cannot be applied on the same type and a
-`packed` type cannot transitively contain another `align`ed type. `packed` may
-only be applied to the default and `C` representations.
+`packed` type cannot transitively contain another `align`ed type. `align` and
+`packed` may only be applied to the [default] and [`C`] representations.
 
 <div class="warning">
 
@@ -369,3 +358,7 @@ used with any other representation.
 [undefined behavior]: behavior-considered-undefined.html
 [27060]: https://github.com/rust-lang/rust/issues/27060
 [`PhantomData<T>`]: special-types-and-traits.html#phantomdatat
+[Default]: #the-default-representation
+[`C`]: #the-c-representation
+[primitive representations]: #primitive-representations
+[`transparent`]: #the-transparent-representation
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -297,10 +297,10 @@ The packing value is specified as a parameter in the form of
 packing value is 1. The packing value must be a power of two up to
 2<sup>29</sup>.
 
-The `packed` representation sets the alignment to the smaller of the specified
+The `packed` representation sets the type's alignment to the smaller of the specified
 packing and the current alignment (either default or `C`). The alignments of
-each field for the purpose of positioning fields would also be the smaller of
-the specified packing and the alignment of the type of that field. If the
+each field, for the purpose of positioning fields, is the smaller of
+the specified packing and the alignment of the field's type. If the
 specified packing is greater than or equal to the default alignment of the
 type, then the alignment and layout is unaffected.
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -113,8 +113,7 @@ All user-defined composite types (`struct`s, `enum`s, and `union`s) have a
 *representation* that specifies what the layout is for the type.
 
 The possible representations for a type are the default representation, `C`,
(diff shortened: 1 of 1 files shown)


src/type-layout.md
```diff
@@ -280,19 +280,29 @@ The `align` representation can be used on `struct`s and `union`s to raise the
...
+alignment and padding of the type.
+
+The packing value is specified as a parameter in the form of
+`#[repr(packed(x))]`. If no value is given, as in `#[repr(packed)]`, then the
+packing value is 1. The packing value must be a power of two up to
+2<sup>29</sup>.
 
-It modifies the representation (either the default or `C`) by removing any
-padding bytes and forcing the alignment of the type to `1`.
+The `packed` representation sets the alignment to the smaller of the specified
```
Centril:
```suggestion
The `packed` representation sets the type's alignment to the smaller of the specified
```

src/type-layout.md
```diff
@@ -280,19 +280,29 @@ The `align` representation can be used on `struct`s and `union`s to raise the
...
+packing and the current alignment (either default or `C`). The alignments of
+each field for the purpose of positioning fields would also be the smaller of
```
Centril:
```suggestion
each field, for the purpose of positioning fields, is the smaller of
```

src/type-layout.md
```diff
@@ -280,19 +280,29 @@ The `align` representation can be used on `struct`s and `union`s to raise the
...
+the specified packing and the alignment of the type of that field. If the
```
Centril:
```suggestion
the specified packing and the alignment of the field's type. If the
```

src/type-layout.md
```diff
@@ -280,19 +280,29 @@ The `align` representation can be used on `struct`s and `union`s to raise the
...
+
+The packing value is specified as a parameter in the form of
+`#[repr(packed(x))]`. If no value is given, as in `#[repr(packed)]`, then the
+packing value is 1. The packing value must be a power of two up to
+2<sup>29</sup>.
 
-It modifies the representation (either the default or `C`) by removing any
-padding bytes and forcing the alignment of the type to `1`.
+The `packed` representation sets the alignment to the smaller of the specified
+packing and the current alignment (either default or `C`). The alignments of
```
Centril: Here the "default" alignment is referenced but above "primitive alignment" is used... we should pick one phrasing.

Centril: "or `C`" is not super clear wrt. referring to C-the-language

Havvy: Get rid of both "current' and "primitive". Get rid of the parenthetical. Instead spell it out explicitly that if the type has the default or C representation, and the alignment from that representation is smaller, it will use that alignment.

ehuss:
I added an experimental commit to propose a different way of documenting these `repr` instructions. I think it more accurately captures how it works. That is, there are some [basic "representations"](https://github.com/rust-lang/rust/blob/4c27fb19ba15a2e45485e601a79914c6280196b0/src/librustc/ty/mod.rs#L2024-L2035), and `packed` and `align` just alter them. In the relevant cases (on stable) there are only two — `C` and default. I think it makes it clearer about what can be "combined".

I do have some concerns and doubts:
- This introduces new terminology ("modifier") that I made up.
- The descriptions between `align` and `packed` have a lot of duplication.
- I think that `align` and `packed` *force* the alignment to the given value. However, I'm not certain. The [parsing code](https://github.com/rust-lang/rust/blob/4c27fb19ba15a2e45485e601a79914c6280196b0/src/libsyntax/attr/builtin.rs#L732) calls them "hints" in the error messages, but that terminology is not used elsewhere. I'm fairly sure that align is forced [here](https://github.com/rust-lang/rust/blob/4c27fb19ba15a2e45485e601a79914c6280196b0/src/librustc/ty/layout.rs#L347) and packed is forced [here](https://github.com/rust-lang/rust/blob/4c27fb19ba15a2e45485e601a79914c6280196b0/src/librustc/ty/layout.rs#L331).  This 1000 line function is a bit hard to follow.
    - Related: I think that `packed` has a floor of the i8 alignment [here](https://github.com/rust-lang/rust/blob/4c27fb19ba15a2e45485e601a79914c6280196b0/src/librustc/ty/layout.rs#L260). This is always 8-bits on all platforms rustc builds on that I can find. I tried building with a custom target where it was 16-bits, but libcore fails to build with `i8:16:32` alignment (and I need the `align_of` function to test my hypothesis). I'm not sure if this floor should be documented, or if Rust should always assume this to be "1"

src/type-layout.md
```diff
@@ -128,14 +127,36 @@ struct ThreeInts {
 }
 ```
 
+The alignment may be raised or lowered with the `align` and `packed` modifiers
+respectively. They alter the representation specified in the attribute:
```
Centril:
```suggestion
respectively. They alter the representation specified in the attribute.
If no representation is specified, the default one is altered.
```

src/type-layout.md
```diff
@@ -113,13 +113,12 @@ All user-defined composite types (`struct`s, `enum`s, and `union`s) have a
 *representation* that specifies what the layout is for the type.
 
 The possible representations for a type are the default representation, `C`,
-the primitive representations, `packed`, and `transparent`. Multiple
-representations can be applied to a single type.
+the primitive representations, and `transparent`.
```
Centril: Itemize ^

Also make it clear what the "primitive representations" entail (by itemizing all the primitive reprs)

src/type-layout.md
```diff
@@ -128,14 +127,36 @@ struct ThreeInts {
...
 > the same name have the same representation. For example, `Foo<Bar>` and
 > `Foo<Baz>` both have the same representation.
 
-The representation of a type does not change the layout of its fields. For
-example, a struct with a `C` representation that contains a struct `Inner` with
-the default representation will not change the layout of Inner.
+The representation of a type can change the padding between fields, but does
+not change the layout of the fields themselves. For example, a struct with a
+`C` representation that contains a struct `Inner` with the default
+representation will not change the layout of Inner.
```
Centril:
```suggestion
representation will not change the layout of `Inner`.
```

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
 
-Alignment is specified as a parameter in the form of `#[repr(align(x))]`. The
-alignment value must be a power of two of type `u32`. The `align` representation
-can raise the alignment of a type to be greater than it's primitive alignment,
-it cannot lower the alignment of a type.
+The alignment is specified as an integer parameter in the form of
+`#[repr(align(x))]`. The alignment value must be a power of two from 1 up to
+2<sup>29</sup>.
+
+The `align` modifier raises the type's alignment to the specified alignment.
```
Centril: This is a partial repetition of what you said in the leading paragraph.

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
+If the specified alignment is less than the alignment of the type without the
```
Centril: I believe this concept of "without the ** modifier" is referred to as the "natural **", for example "natural alignment". I think it would be good to introduce such terms earlier (when we first start to speak about alignment) and then use them.

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
+### The `packed` modifier
+
+The `packed` modifier can be used on `struct`s and `union`s to lower the
+alignment of the type to a given value.
 
-The `packed` representation can only be used on `struct`s and `union`s.
+The packing value is specified as an integer parameter in the form of
+`#[repr(packed(x))]`. If no value is given, as in `#[repr(packed)]`, then the
+packing value is 1. The packing value must be a power of two from 1 up to
+2<sup>29</sup>.
```
Centril: If you extract out the wording about the default it is the same as the paragraph for `align`. So extract these two out into a common notion `repr_value` or something and then reference that one.

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
 
-It modifies the representation (either the default or `C`) by removing any
-padding bytes and forcing the alignment of the type to `1`.
+The `packed` modifier lowers the type's alignment to the specified packing. If
```
Centril:
```suggestion
The `packed` modifier lowers the type's alignment to the specified value. If
```

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
+the specified packing is greater to the alignment of the type without the
```
Centril:
```suggestion
the specified value is greater than the type's alignment without the
```

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
+only be applied to the default and `C` representations.
 
-### The `packed` Representation
+### The `packed` modifier
+
+The `packed` modifier can be used on `struct`s and `union`s to lower the
+alignment of the type to a given value.
 
-The `packed` representation can only be used on `struct`s and `union`s.
+The packing value is specified as an integer parameter in the form of
```
Centril:
```suggestion
The packing value (hence: "value") is specified as an integer parameter in the form of
```

src/type-layout.md
```diff
@@ -274,28 +295,42 @@ For all other enumerations, the layout is unspecified.
...
+`#[repr(packed(x))]`. If no value is given, as in `#[repr(packed)]`, then the
+packing value is 1. The packing value must be a power of two from 1 up to
```
Centril:
```suggestion
value is 1. The value must be a power of two from 1 up to
```

ehuss:
I attempted to reduce some of the duplication.  It still seems a little awkward, though.

Centril:
I think this is strict improvement for now. Thanks @ehuss.
Feel free to merge at will :)

retep998:
Technically it would be more accurate to say `packed` lowers the alignment of the *fields* in the struct, rather than the struct itself. But since the alignment of the struct itself is determined from the maximum alignment of the fields, the struct alignment effectively decreases as well. `align` meanwhile increases the alignment of the struct but *not* its fields.

Centril:
@retep998 Could you file a clarification PR?

PR #554, opened 2019-03-25
Document wasm_import_module for #[link].

ehuss:
Stabilization PR: https://github.com/rust-lang/rust/pull/52445
Tracking issue: https://github.com/rust-lang/rust/issues/52090

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -110,16 +110,23 @@ The following [attributes] control the behavior of external blocks.
 ### The `link` attribute
 
 The *`link` attribute* specifies the name of a native library that the
-compiler should link with. It uses the [_MetaListNameValueStr_] syntax to
-specify its inputs. The `name` key is the name of the native library to link.
-The `kind` key is an optional value which specifies the kind of library with
-the following possible values:
+compiler should link with for the items within an `extern` block. It uses the
+[_MetaListNameValueStr_] syntax to specify its inputs. The `name` key is the
+name of the native library to link. The `kind` key is an optional value which
+specifies the kind of library with the following possible values:
 
 - `dylib` — Indicates a dynamic library. This is the default if `kind` is not
   specified.
 - `static` — Indicates a static library.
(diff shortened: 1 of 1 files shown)


src/items/external-blocks.md
```diff
@@ -131,6 +138,11 @@ extern {
 extern {
     // …
 }
+
+#[link(wasm_import_module = "foo")]
```
Havvy: Is there a common wasm import module we could put here?

src/items/external-blocks.md
```diff
@@ -152,6 +164,7 @@ extern {
 ```
 
 [IDENTIFIER]: identifiers.html
+[WebAssembly]: https://webassembly.org/
```
Centril:
```suggestion
[WebAssembly module]: https://webassembly.github.io/spec/core/syntax/modules.html
```

src/items/external-blocks.md
```diff
@@ -110,17 +110,24 @@ The following [attributes] control the behavior of external blocks.
...
 
 - `dylib` — Indicates a dynamic library. This is the default if `kind` is not
   specified.
 - `static` — Indicates a static library.
 - `framework` — Indicates a macOS framework. This is only valid for macOS
   targets.
 
+The `name` key must be included if `kind` is specified.
+
+The `wasm_import_module` key may be used to specify the [WebAssembly] module
```
Centril:
```suggestion
The `wasm_import_module` key may be used to specify the [WebAssembly module]
```

src/items/external-blocks.md
ehuss: Hehe, I spent about 10 minutes when I wrote it trying to find that exact same answer.  I don't like foo/bar examples, but I couldn't find anything better.  I don't think there is anything "common".  It can be any UTF-8 string, something that you would typically name a module.  If you can think of something different, I'd be happy to change it.  I think any word here will be arbitrary.

Centril: Let's revisit later if need be.

PR #555, opened 2019-03-25
Document the `automatically_derived` attribute.

ehuss:
It's not clear if this *should* be documented. Alternatively, it could be deprecated if usage in stable is really not desired. Not sure who can make this call.

Implemented in: https://github.com/rust-lang/rust/pull/12929

CC #532

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -185,6 +185,8 @@ The following is an index of all built-in attributes.
   - [`should_panic`] — Indicates a test should generate a panic.
 - Derive
   - [`derive`] — Automatic trait implementations.
+  - [`automatically_derived`] — Marker for implementations created by
+    `derive`.
 - Macros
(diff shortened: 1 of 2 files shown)


Havvy:
cc @QuietMisdreavus

Centril:
> It's not clear if this _should_ be documented.

If it's to remain on stable (deprecated or otherwise) I would prefer it to be documented.

> Alternatively, it could be deprecated if usage in stable is really not desired. Not sure who can make this call.

The language team would be the reviewing team.

QuietMisdreavus:
As someone who implemented [a tool reacting to this attribute](https://github.com/rust-lang/rust/pull/58626), i would actually prefer if we started pushing derive-macro authors to use this attribute. Having it in the reference like this would help raise awareness that it can be used and reacted to.

Centril:
@QuietMisdreavus Could you give examples of all the ways the attribute is used so we can list them in the reference?

PR #556, opened 2019-03-28
Typo fixes

u32i64:
Fixes various typos.

### DIFF
--- a/src/special-types-and-traits.md
+++ b/src/special-types-and-traits.md
@@ -110,7 +110,7 @@ according to the following rules:
   closure that captures a `T` by shared reference and a `U` by value implements
   any auto traits that both `&T` and `U` do.
 
-For generic types (counting the built-in types above as generic over `T`), if an
+For generic types (counting the built-in types above as generic over `T`), if a
 generic implementation is available, then the compiler does not automatically
 implement it for types that could use the implementation except that they do not
 meet the requisite trait bounds. For instance, the standard library implements
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -96,7 +96,7 @@ Coercion is allowed between the following types:
 * `T_1` to `T_3` where `T_1` coerces to `T_2` and `T_2` coerces to `T_3`
 (*transitive case*)
 
-    Note that this is not fully supported yet
+    Note that this is not fully supported yet.
 
 * `&mut T` to `&T`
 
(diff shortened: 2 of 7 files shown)


src/type-coercions.md
```diff
@@ -158,8 +158,8 @@ cases where other coercions are not, as described above. They can still happen
 anywhere else a coercion can occur.
 
 Two traits, [`Unsize`] and [`CoerceUnsized`], are used
-to assist in this process and expose it for library use. The compiler following
-coercions are built-in and, if `T` can be coerced to `U` with one of the, then
+to assist in this process and expose it for library use. The following
+coercions are compiler built-ins and, if `T` can be coerced to `U` with one of them, then
```
Centril:
```suggestion
coercions are built-ins and, if `T` can be coerced to `U` with one of them, then
```

src/type-coercions.md
```diff
@@ -158,8 +158,8 @@ cases where other coercions are not, as described above. They can still happen
...
 the compiler will provide an implementation of `Unsize<U>` for `T`:
```
Centril:
```suggestion
an implementation of `Unsize<U>` for `T` will be provided:
```

src/special-types-and-traits.md
```diff
@@ -110,7 +110,7 @@ according to the following rules:
   closure that captures a `T` by shared reference and a `U` by value implements
   any auto traits that both `&T` and `U` do.
 
-For generic types (counting the built-in types above as generic over `T`), if an
+For generic types (counting the built-in types above as generic over `T`), if a
 generic implementation is available, then the compiler does not automatically
 implement it for types that could use the implementation except that they do not
```
u32i64: @Centril Perhaps this can be re-phrased? I can't seem to understand what the `except that they do not meet...` part means. Does that mean `except if they` or `except then they`...

Centril: Consider:
```rust
struct A<T>(T);

unsafe impl<T: Copy> Send for A<T> {}

fn assert_send<T: Send>() {}

fn main() {
    assert_send::<A<u8>>();
    assert_send::<A<String>>();
}
```

u32i64: So the compiler would have done this  (and does so if we comment out our `unsafe impl`):
```rust
unsafe impl Send for A<String> {}
// as part of:
unsafe impl Send for A<T> where T: Send {}
```

But our `unsafe impl` tells the compiler to disregard its `auto trait` rules and instead use our "rule" (`impl`), i.e. we tell the compiler that it is not safe to `Send` values of type `A<T>` unless `T: Copy`?

Centril: Yep, pretty much.

u32i64:
@Centril I think this is ready for merging now.

Centril:
Yep; Thanks!

PR #562, opened 2019-04-12
Fix attribute redirects.

ehuss:
When these pages were moved during the PR changes, I forgot to update the redirects.


Havvy:
Too bad the linkchecker couldn't see those links; but still, they're fixed for now.

PR #600, opened 2019-04-23
Mention that unicode escapes allow _

matklad:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=11bfffcbf19def9813810c71a820519f

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -67,6 +67,8 @@ evaluated (primarily) at compile time.
 |---|------|
 | `\u{7FFF}` | 24-bit Unicode character code (up to 6 digits) |
 
+`*` Escape allows `_` as a visual separator: `\u{7F_FF}`
+
 #### Quote escapes
 
 |   | Name |


ehuss:
Thanks! I think the underscores are captured in the `UNICODE_ESCAPE` rule, but it is never mentioned in prose. I would maybe do things slightly differently:

- The `*` is intended to be a footnote, but in this case there is no backreference. I would not use a footnote, and just add a second row to the example table (these are examples only).
- I would include a brief mention of underscores in the section that starts "A 24-bit code point escape...", perhaps worded similarly to the way it is in integer literals. I would also probably mention (although I'm not sure where) that these are *visual* separators only and are ignored . I don't see anywhere where it describes what the underscores are, it just assumes everyone knows they are ignored.

matklad:
>I think the underscores are captured in the UNICODE_ESCAPE rule, but it is never mentioned in prose. I would maybe do things slightly differently:

Aha, haven't seen that! Though, the rule there is not entirely correct:

* leasing underscore is forbidden
* there should be at most six digits, but there may be more character b/c of underscores.

I'll try to fix this soon

ehuss:
It looks correct to me?

( HEX_DIGIT `_`<sup>\*</sup> )<sup>1..6</sup>

That is: must start with a digit, there must be between 1 and 6 digits, and each digit may have 0 or more underscores following it.

matklad:
Ah, sorry, you are totally right of course!

PR #601, opened 2019-04-28
Remove unneeded | from AssignmentExpression

MemPage:
Remove unneeded | from AssignmentExpression


lzutao:
Hm, Centril doesn't merge it yet.

PR #609, opened 2019-05-12
Add `non_exhaustive` to reference.

davidtwco:
This PR adds documentation for [RFC 2008](https://github.com/rust-lang/rfcs/pull/2008) (tracking issue rust-lang/rust#44109) to the reference.

### DIFF
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -1,0 +1,140 @@
+# Type system attributes
+
+The following [attributes] are used for changing how a type can be used.
+
+## The `non_exhaustive` attribute
+
+The *`non_exhaustive` attribute* indicates that a type or variant may have
+more fields or variants added in the future. It can be applied to
+[`struct`s][struct], [`enum`s][enum], and `enum` variants.
+
+The `non_exhaustive` attribute uses the [_MetaWord_] syntax and thus does not
+take any inputs.
+
+Within the defining crate, `non_exhaustive` has no effect.
+
+```rust
+#[non_exhaustive]
+pub struct Config {
+    pub window_width: u16,
+    pub window_height: u16,
+}
+
+#[non_exhaustive]
+pub enum Error {
+    Message(String),
+    Other,
+}
+
+pub enum Message {
+    #[non_exhaustive] Send { from: u32, to: u32, contents: String },
+    #[non_exhaustive] Reaction(u32),
+    #[non_exhaustive] Quit,
+}
+
+// Non-exhaustive structs can be constructed as normal within the defining crate.
+let config = Config { window_width: 640, window_height: 480 };
+
+// Non-exhaustive structs can be matched on exhaustively within the defining crate.
+if let Config { window_width, window_height } = config {
+    // ...
+}
+
+let error = Error::Other;
+let message = Message::Reaction(3);
+
+// Non-exhaustive enums can be matched on exhaustively within the defining crate.
+match error {
+    Error::Message(ref s) => { },
+    Error::Other => { },
+}
+
+match message {
+    // Non-exhaustive variants can be matched on exhaustively within the defining crate.
+    Message::Send { from, to, contents } => { },
+    Message::Reaction(id) => { },
+    Message::Quit => { },
+}
+```
+
+Outside of the defining crate, types annotated with `non_exhaustive` have limitations that
+preserve backwards compatibility when new fields or variants are added.
+
+Non-exhaustive types cannot be constructed outside of the defining crate:
+
+- Non-exhaustive variants ([`struct`][struct] or [`enum` variant][enum]) cannot be constructed
+  with a [_StructExpression_] \(including with [functional update syntax]).
+- [`enum`][enum] instances can be constructed in an [_EnumerationVariantExpression_].
+
+```rust,ignore (requires multiple crates)
+// `Config`, `Error`, and `Message` are types defined in an upstream crate that have been
+// annotated as `#[non_exhaustive]`.
+use upstream::{Config, Error, Message};
+
+// Cannot construct an instance of `Config`, if new fields were added in
+// a new version of `upstream` then this would fail to compile, so it is
+// disallowed.
+let config = Config { window_width: 640, window_height: 480 };
+
+// Can construct an instance of `Error`, new variants being introduced would
+// not result in this failing to compile.
+let error = Error::Message("foo".to_string());
+
+// Cannot construct an instance of `Message::Send` or `Message::Reaction`,
+// if new fields were added in a new version of `upstream` then this would
+// fail to compile, so it is disallowed.
+let message = Message::Send { from: 0, to: 1, contents: "foo".to_string(), };
+let message = Message::Reaction(0);
+
+// Cannot construct an instance of `Message::Quit`, if this were converted to
+// a tuple-variant `upstream` then this would fail to compile.
+let message = Message::Quit;
+```
+
+There are limitations when matching on non-exhaustive types outside of the defining crate:
+
+- When pattern matching on a non-exhaustive variant ([`struct`][struct] or [`enum` variant][enum]),
+  a [_StructPattern_] must be used which must include a `..`. Tuple variant constructor visibility
+  is lowered to `min($vis, pub(crate))`.
+- When pattern matching on a non-exhaustive [`enum`][enum], matching on a variant does not
+  contribute towards the exhaustiveness of the arms.
+
+```rust, ignore (requires multiple crates)
+// `Config`, `Error`, and `Message` are types defined in an upstream crate that have been
+// annotated as `#[non_exhaustive]`.
+use upstream::{Config, Error, Message};
+
+// Cannot match on a non-exhaustive enum without including a wildcard arm.
+match error {
+  Error::Message(ref s) => { },
+  Error::Other => { },
+  // would compile with: `_ => {},`
+}
+
+// Cannot match on a non-exhaustive struct without a wildcard.
+if let Ok(Config { window_width, window_height }) = config {
+    // would compile with: `..`
+}
+
+match message {
+  // Cannot match on a non-exhaustive struct enum variant without including a wildcard.
+  Message::Send { from, to, contents } => { },
+  // Cannot match on a non-exhaustive tuple or unit enum variant.
+  Message::Reaction(type) => { },
+  Message::Quit => { },
+}
+```
+
+Non-exhaustive types are always considered inhabited in downstream crates.
+
+[_EnumerationVariantExpression_]: ../expressions/enum-variant-expr.md
+[_MetaWord_]: ../attributes.md#meta-item-attribute-syntax
+[_StructExpression_]: ../expressions/struct-expr.md
+[_StructPattern_]: ../patterns.md#struct-patterns
+[_TupleStructPattern_]: ../patterns.md#tuple-struct-patterns
+[`if let`]: ../expressions/if-expr.md#if-let-expressions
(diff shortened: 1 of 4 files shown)


ehuss:
Nice! I'm looking forward to this being stabilized. The stabilization report says 1.35, but that's in about 10 days, will it maybe be in a future release?

As for categorization, that's a tough one. I had originally intended that section to be just for lints, but I can't think of a better place to put it.  I don't feel strongly where it goes, so unless someone else has a suggestion, I don't have a preference.

davidtwco:
> The stabilization report says 1.35, but that's in about 10 days, will it maybe be in a future release?

I just wrote the next release when I was drafting that, I have no idea what release it will be. I've changed it to TBD in that document.

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
 > let _ = five();
 > ```
 
+# The `non_exhaustive` attribute
+
+The *`non_exhaustive` attribute* is used to indicate that a type will have
+more fields/variants added in the future. It can be applied to
```
Havvy:
```suggestion
more fields or variants added in the future. It can be applied to
```

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+match message {
+    // Non-exhaustive variants can be matched on exhaustively within the defining crate.
+    Message::Send { from, to, contents } => { },
+    Message::Reaction(type) => { },
+    Message::Quit => { },
+}
+```
+
+In downstream crates, types annotated with `non_exhaustive` have limitations that
+preserve backwards compatibility when new fields/variants are added.
```
Havvy:
```suggestion
preserve backwards compatibility when new fields or variants are added.
```

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+// fail to compile, so it is disallowed.
+let message = Message::Send { from: 0, to: 1, contents: "foo".to_string(), };
+let message = Message::Reaction(0);
+
+// Cannot construct an instance of `Message::Quit`, if this were converted to
+// a tuple-variant `upstream` then this would fail to compile.
+let message = Message::Quit;
+```
+
+Non-exhaustive types cannot be used in a `match`/`if let` expression without a
```
Havvy:
```suggestion
Non-exhaustive types cannot be used in a `match` or `if let` expression without a
```

It'd be nice if these were links as well.

ehuss:
@davidtwco Just a heads up, I have rebased this PR on latest master to resolve the conflicts, so be careful updating locally.

The reference has migrated to mdbook 0.3.  This means that links should use the `.md` extension, and are relative to the page they are on (previously they were relative to the root of the book). Please don't hesitate to ask if you have any questions or any trouble updating. Sorry for the interruption!

davidtwco:
@ehuss thanks a lot for doing that, I hadn’t been checking to see if there were any conflicts.

Centril:
The stabilization PR (https://github.com/rust-lang/rust/pull/64639) just went through.

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+    Message::Send { from, to, contents } => { },
+    Message::Reaction(type) => { },
+    Message::Quit => { },
+}
+```
+
+In downstream crates, types annotated with `non_exhaustive` have limitations that
+preserve backwards compatibility when new fields or variants are added.
+
+Non-exhaustive types cannot be constructed in downstream crates:
```
Centril: It would be good to note what "constructed" means here to attempt something a bit more formal by talking about expression forms (e.g. path expressions, struct expressions, call expressions, etc.) as already defined in the reference and what effects `#[non_exhaustive]` has on those. As written, the examples below is largely guide-level and motivational.

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+// fail to compile, so it is disallowed.
+let message = Message::Send { from: 0, to: 1, contents: "foo".to_string(), };
+let message = Message::Reaction(0);
+
+// Cannot construct an instance of `Message::Quit`, if this were converted to
+// a tuple-variant `upstream` then this would fail to compile.
+let message = Message::Quit;
+```
+
+Non-exhaustive types cannot be used in a [`match`] or [`if let`] expression without a
```
Centril: And `while let`, `for`, and `let` as well as function parameters?
Also, talking about adding a wildcard arm to `if let` seems strange; it already has one implicitly.

I think we can describe things as (modulo rewording -- and do check my logic):

- When pattern matching on a non_exhaustive `enum`, matching on a variant does not contribute towards exhaustiveness of the arms. (`_ =>` falls out from this)
- When pattern matching on a non_exhaustive variant (struct, enum variant), struct patterns must be used and  `..` must be included in in them.

(Ideally we reference entities from the pattern grammar.)

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
 > let _ = five();
 > ```
 
+# The `non_exhaustive` attribute
+
+The *`non_exhaustive` attribute* is used to indicate that a type will have
```
Centril: In general, the usage of "type" does not cover variants.

src/attributes/diagnostics.md
davidtwco: I've updated the wording based on what you provided here, and added some references to the patterns section.

src/attributes/diagnostics.md
davidtwco: I've updated the wording here, including references to the expression forms.

src/attributes/diagnostics.md
```diff
@@ -252,10 +252,125 @@ When used on a function in a trait implementation, the attribute does nothing.
 > let _ = five();
 > ```
 
+# The `non_exhaustive` attribute
```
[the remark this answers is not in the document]
davidtwco: Instead of doing this, I've added a new sub-section to the attributes section and added this there. I think it makes sense for this to be included alongside the other attributes, but it didn't quite fit in any of the existing categories.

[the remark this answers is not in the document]
Centril: Looks good, thanks.

src/attributes/type_system.md
```diff
@@ -0,0 +1,137 @@
...
+
+// Cannot construct an instance of `Message::Quit`, if this were converted to
+// a tuple-variant `upstream` then this would fail to compile.
+let message = Message::Quit;
+```
+
+There are limitations when matching on non-exhaustive types outside of the defining crate:
+
+- When pattern matching on a non-exhaustive variant ([`struct`][struct] or [`enum` variant][enum]),
+a [_StructPattern_] or [_TupleStructPattern_] must be used which must include a `..`.
```
Centril: Noteworthy: tuples and tuple variants cannot be matched with `Tuple(...)` - we could describe this with "the constructor's visibility is lowered to `MIN($vis, pub(crate))` or some such? (The example above only covers tuple variants, not regular tuples.)

src/attributes/type_system.md
```diff
@@ -0,0 +1,137 @@
...
+Outside of the defining crate, types annotated with `non_exhaustive` have limitations that
+preserve backwards compatibility when new fields or variants are added.
+
+Non-exhaustive types cannot be constructed outside of the defining crate:
+
+- [`struct`s][struct] cannot be constructed with a [struct expression].
+- [`enum`][enum] instances can be constructed in an [enumeration variant expression].
+
+```rust,ignore (requires multiple crates)
+// `Config`, `Error` and `Message` are types defined in an upstream crate that have been
```
Centril:
```suggestion
// `Config`, `Error`, and `Message` are types defined in an upstream crate that have been
```

src/attributes/type_system.md
```diff
@@ -0,0 +1,137 @@
...
+
+There are limitations when matching on non-exhaustive types outside of the defining crate:
+
+- When pattern matching on a non-exhaustive variant ([`struct`][struct] or [`enum` variant][enum]),
+a [_StructPattern_] or [_TupleStructPattern_] must be used which must include a `..`.
+- When pattern matching on a non-exhaustive [`enum`][enum], matching on a variant does not
+  contribute towards the exhaustiveness of the arms.
+
+```rust, ignore (requires multiple crates)
+// `Config`, `Error` and `Message` are types defined in an upstream crate that have been
```
Centril:
```suggestion
// `Config`, `Error`, and `Message` are types defined in an upstream crate that have been
```

src/attributes/type_system.md
```diff
@@ -0,0 +1,137 @@
...
+    Message::Send { from, to, contents } => { },
+    Message::Reaction(id) => { },
+    Message::Quit => { },
+}
+```
+
+Outside of the defining crate, types annotated with `non_exhaustive` have limitations that
+preserve backwards compatibility when new fields or variants are added.
+
+Non-exhaustive types cannot be constructed outside of the defining crate:
```
Centril: Seems record updates are not covered either atm?

davidtwco: Added some wording for this.

src/attributes/type_system.md
```diff
@@ -0,0 +1,137 @@
...
+
+- [`struct`s][struct] cannot be constructed with a [struct expression].
+- [`enum`][enum] instances can be constructed in an [enumeration variant expression].
```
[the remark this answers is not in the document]
davidtwco: Changed to the `EnumerationVariantExpression` style. Re-worded the first point to apply to enum variants as well as structs.

src/attributes/type_system.md
davidtwco: I've added a sentence on tuple variant constructor visibility.

Centril:
LGTM; but maybe @ehuss and @petrochenkov wants to take a look also.

ehuss:
I was wondering if you would be willing to make a PR to define "inhabited" and "uninhabited" in the glossary?  It is not defined anywhere, and this adds a second reference to the term, which (I think) is unclear for anyone not well versed in type theory (e.g. me).

davidtwco:
> 
> 
> I was wondering if you would be willing to make a PR to define "inhabited" and "uninhabited" in the glossary? It is not defined anywhere, and this adds a second reference to the term, which (I think) is unclear for anyone not well versed in type theory (e.g. me).

I've added a commit which adds inhabited/uninhabited to the glossary.

ehuss:
Thanks! ❤️

src/glossary.md
```diff
@@ -65,6 +65,12 @@ For example, `2 + (3 * 4)` is an expression that returns the value 14.
 An [item] that is not a member of an [implementation], such as a *free
 function* or a *free const*. Contrast to an [associated item].
 
+### Inhabited
+
+A type is inhabited if it has constructors and therefore can be instantiated. An inhabited type is
```
Centril: Is your notion of "has constructors" intended to e.g. take visibility into account? That's the current definition of inhabitedness *in the type system* but *not the abstract machine*. You are describing this from the POV of the type system, right?

davidtwco: Uh, yes..? I just borrowed the explanation from [RFC 1216](https://github.com/rust-lang/rfcs/blob/master/text/1216-bang-type.md).

Centril: I see. :) Well, this is good.

PR #611, opened 2019-05-29
Fix example code of derive macro

yvt:
There is an example in `procedural-macros.md` that doesn’t compile:
```
   Compiling proctest v0.1.0 (/nowhere/proctest)
error[E0308]: mismatched types
  --> src/lib.rs:42:50
   |
42 | pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
   |        ------------------                        ^^^^^^^^^^^ expected struct `proc_macro::TokenStream`, found ()
   |        |
   |        this function's body doesn't return
43 |     TokenStream::new();
   |                       - help: consider removing this semicolon
   |
   = note: expected type `proc_macro::TokenStream`
              found type `()`
```
This PR fixes the example so that it compiles.


Centril:
Thanks!

PR #612, opened 2019-05-31
Document that literals with any suffixes are valid as tokens

petrochenkov:
Closes https://github.com/rust-lang/rust/issues/60494

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -88,6 +88,9 @@ evaluated (primarily) at compile time.
 
 #### Suffixes
 
+Suffix is a non-raw identifier immediately (without whitespace)
+following the primary part of a literal.
+
 Any kind of literal (string, integer, etc) with any suffix is valid as a token,
 and can be passed to a macro without producing an error.  
 The macro itself will decide how to interpret such a token and whether to produce an error or not.
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -88,7 +88,7 @@ evaluated (primarily) at compile time.
 
 #### Suffixes
 
-Suffix is a non-raw identifier immediately (without whitespace)
+A suffix is a non-raw identifier immediately (without whitespace)
 following the primary part of a literal.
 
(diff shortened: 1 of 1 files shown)


src/tokens.md
```diff
@@ -88,6 +88,20 @@ evaluated (primarily) at compile time.
 
 #### Suffixes
 
+Any kind of literal (string, integer, etc) with any suffix is valid as a token,
```
Centril: The suffix must be an IDENT, right?

e.g. `blackhole!("string"4);` is not OK even tho `4` is a suffix of `"string"`

petrochenkov: >The suffix must be an IDENT, right?

Yes, added a commit with the definition.

petrochenkov: `4` is not a suffix, `"string"4` is lexed into two literal tokens.

src/tokens.md
```diff
@@ -88,6 +88,23 @@ evaluated (primarily) at compile time.
 
 #### Suffixes
 
+Suffix is a non-raw identifier immediately (without whitespace)
```
Centril:
```suggestion
A suffix is a non-raw identifier immediately (without whitespace)
```

ehuss:
That section isn't really intended to be the definition, but is rather just a list of examples.  Perhaps it should be reorganized since others have been a bit confused by it.

I would think there also needs to be some updates to the `Lexer` definitions to cover this.  It's unfortunate that this makes essentially two sets of literals, but perhaps just a consolidated section could be created to cover "macro tokens".

I can work on rearranging this some time soonish, but I wouldn't consider this resolved.

Centril:
@ehuss Could you file an issue meanwhile?

petrochenkov:
Yeah, a new reference issue, not reopening that lang issue.
I wasn't too happy about the structure of `src/tokens.md` either, but I also wasn't ready for doing a larger rewrite.

PR #619, opened 2019-06-10
Describe `#[repr(align(x))]` semantics on `enum`s

Centril:
Closes https://github.com/rust-lang/rust/issues/57996

Stabilization report: https://github.com/rust-lang/rust/pull/61229#issuecomment-496092471

r? @ehuss

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -322,6 +322,10 @@ The `align` and `packed` modifiers cannot be applied on the same type and a
 `packed` type cannot transitively contain another `align`ed type. `align` and
 `packed` may only be applied to the [default] and [`C`] representations.
 
+The `align` modifier can also be applied on an `enum`.
+When it is, the effect on the `enum`'s alignment is the same as if the `enum`
+was wrapped in a newtype `struct` with the same `align` modifier.
+
 <div class="warning">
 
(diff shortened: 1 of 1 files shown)


ehuss:
Can you explain why this is described with a level of indirection by wrapping in a newtype struct?

Why can't it just say something like "`align` may be applied on an `enum` to specify the alignment of the `enum`'s values"?

Centril:
> Can you explain why this is described with a level of indirection by wrapping in a newtype struct?

This gives an unambiguous definition for the semantics... I guess I can "unwrap" the indirection and describe it in more native terms.

> Why can't it just say something like "`align` may be applied on an `enum` to specify the alignment of the `enum`'s values"?

That doesn't seem like a specification...

ehuss:
Ok, makes sense. I was just feeling like it was trying to say something more, or that enums are somehow special.  But I guess any other way would end up being wordy and duplicating. If you want to "unwrap" it to make it more direct, go ahead. Otherwise, click the merge button if you're happy.  It can always be updated later if someone feels compelled to.

Centril:
Thanks. :)

I'll merge for now cause I don't want to devote too much time to it and it feels serviceable in the interim but I do think it should/can be improved.

PR #621, opened 2019-06-15
Remove extra parenthesis

nasso:
There was an extra left parenthesis on the demo code for the `type_length_limit` attribute.
The code snippet isn't supposed to compile but produced the wrong error.

Edit: The snippet now _compiles_ when it isn't supposed to; I think we should open a separate issue?

### DIFF
--- a/src/attributes/limits.md
+++ b/src/attributes/limits.md
@@ -42,7 +42,11 @@ to set the limit based on the number of type substitutions.
 
 > Note: The default in `rustc` is 1048576.
 
-```rust,compile_fail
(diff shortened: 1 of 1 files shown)


ehuss:
Oops!  

As for the pass/fail, it looks like `compile_fail` implies `no_run`.  `no_run` causes rustdoc to only run the analysis phase.  However, this error seems to only trigger at a later stage (code generation? seems weird).  At any rate, it looks like this is one of those tests that won't work with rustdoc, so I guess just adding `ignore` will be required.

ehuss:
I went ahead and added the `ignore` attribute to get tests to pass.

PR #627, opened 2019-06-29
unions: call out field offset issues

RalfJung:
@gnzlbg we *do* guarantee offset 0 for `repr(C)`, right? That just follows from being C-compatible?

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -39,11 +39,13 @@ let f = u.f1;
 
 Unions have no notion of an "active field". Instead, every union access just
 interprets the storage at the type of the field used for the access. Reading a
-union field reads the bits of the union at the field's type. It is the
-programmer's responsibility to make sure that the data is valid at that
-type. Failing to do so results in undefined behavior. For example, reading the
-value `3` at type `bool` is undefined behavior. Effectively, writing to and then
-reading from a union is analogous to a [`transmute`] from the type used for
+union field reads the bits of the union at the field's type. Fields might have a
+non-zero offset (except when `#[repr(C)]` is used); in that case the bits
+starting at the offset of the fields are read. It is the programmer's
+responsibility to make sure that the data is valid at that type. Failing to do
+so results in undefined behavior. For example, reading the value `3` at type
+`bool` is undefined behavior. Effectively, writing to and then reading from a
+`#[repr(C)]` union is analogous to a [`transmute`] from the type used for
 writing to the type used for reading.
 
 Consequently, all reads of union fields have to be placed in `unsafe` blocks:
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -20,6 +20,8 @@ The key property of unions is that all fields of a union share common storage.
 As a result writes to one field of a union can overwrite its other fields, and
 size of a union is determined by the size of its largest field.
 
+## Initialization of a union
+
 A value of a union type can be created using the same syntax that is used for
 struct types, except that it must specify exactly one field:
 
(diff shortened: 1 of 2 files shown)


src/items/unions.md
```diff
@@ -39,11 +39,13 @@ let f = u.f1;
...
 interprets the storage at the type of the field used for the access. Reading a
-union field reads the bits of the union at the field's type. It is the
-programmer's responsibility to make sure that the data is valid at that
-type. Failing to do so results in undefined behavior. For example, reading the
-value `3` at type `bool` is undefined behavior. Effectively, writing to and then
-reading from a union is analogous to a [`transmute`] from the type used for
+union field reads the bits of the union at the field's type. Fields might have a
+non-zero offset (except when `#[repr(C)]` is used); in that case the bits
+starting at the offset of the fields are read. It is the programmer's
+responsibility to make sure that the data is valid at that type. Failing to do
```
Centril: which type? (the field's presumably, but spell it out...)

gnzlbg:
> we do guarantee offset 0 for repr(C), right? That just follows from being C-compatible?

Yes (https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/unions.md#c-compatible-layout-repr-c):

> For unions tagged #[repr(C)], the compiler will apply the C layout scheme. Per sections 6.5.8.5 and 6.7.2.1.16 of the C11 specification, this means that the offset of every field is 0. Unsafe code can cast a pointer to the union to a field type to obtain a pointer to any field, and vice versa.

src/items/unions.md
RalfJung: Done.

src/items/unions.md
```diff
@@ -37,13 +39,17 @@ struct fields:
...
 interprets the storage at the type of the field used for the access. Reading a
-union field reads the bits of the union at the field's type. It is the
-programmer's responsibility to make sure that the data is valid at that
-type. Failing to do so results in undefined behavior. For example, reading the
-value `3` at type `bool` is undefined behavior. Effectively, writing to and then
-reading from a union is analogous to a [`transmute`] from the type used for
+union field reads the bits of the union at the field's type. Fields might have a
+non-zero offset (except when `#[repr(C)]` is used); in that case the bits
+starting at the offset of the fields are read. It is the programmer's
+responsibility to make sure that the data is valid at the field's type. Failing
```
Centril: It's not really the role of the reference to say "it is the programmer's responsibility" or "this is how you use X" but it's fine here imo.

RalfJung: Do you have a suggestion for wording it differently? I think it is useful to think in terms of responsibilities/obligations when talking about unsafe code, and hence I think it is useful for the reference to explain it as such.

I also have 0 experience writing standard-esque text.^^ Given my experience *reading* it though, it seems often so useless that I am not sure I am sad about that...

Centril: Not really; the text itself is good but it's probably more aimed at users (e.g. the nomicon) whereas the reference is intended as the seeds of a spec with just definitions of the language but not the "why" and "this is how I use it". That is, should be more like the details of one of your academic papers with operational semantics and judgements... it isn't now, but it should be. :)

RalfJung: You mean the details that only 2 people on this planet read because they are incomprehensible? ;)

Centril: That's not the papers I read :)

I think something like the WASM spec has a nice structure:
- https://webassembly.github.io/spec/core/index.html

Issue #634, opened 2019-07-10
document the stabilized surface of async-await


nikomatsakis:
This issue is part of async-await stabilization (see https://github.com/rust-lang/rust/issues/62149). We need to document the following things:

- [ ] grammar for `async fn` and describe their effects
- [ ] grammar for `async move { }` blocks
- [ ] grammar for `await` expressions, and describe the "rough" desugaring

nikomatsakis:
I am taking a first pass at this.

alercah:
Async closures are missing.

nikomatsakis:
Async closures are not stabilized, and this issue was specific to stabilized surface area -- actually, I think we can probably close this? I'm not sure what I meant by 'document the drop order'.

nikomatsakis:
@cramertj or @withoutboats do you have any idea what I might have meant by "link to drop order when that is stabilized"?

I do know that @matthewjasper recently opened a PR about https://github.com/rust-lang/reference/pull/514 drop scopes which I think mentioned something about drop order here...

alercah:
They're not stabilized? Why does
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3795067a2d54b2d90a5d807eed51a93e
work then?

cramertj:
I would guess the drop order bit has to do with the changes we made around when `_` arguments to async functions get dropped.

kpreid:
Another thing that is currently undocumented is async block/fn future types, as types. I filed #1555 for that.

PR #635, opened 2019-07-10
async-await initial reference material

nikomatsakis:
I'm not sure what the best reviewing process is here.

cc @cramertj  @withoutboats @centril @joshtriplett @pnkfelix @scottmcm @pnkfelix @eddyb (I think that's the lang team, did I miss anybody?)

cc https://github.com/rust-lang-nursery/reference/issues/634 https://github.com/rust-lang/rust/issues/62566

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -13,6 +13,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_OperatorExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_GroupedExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ArrayExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_AwaitExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_IndexExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_TupleExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_TupleIndexingExpression_]\
@@ -33,6 +34,7 @@
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup>[†](#expression-attributes)\
 > &nbsp;&nbsp; (\
 > &nbsp;&nbsp; &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_AsyncBlockExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_UnsafeBlockExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_LoopExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_IfExpression_]\
@@ -324,6 +326,8 @@ They are never allowed before:
 
 [_ArithmeticOrLogicalExpression_]: expressions/operator-expr.md#arithmetic-and-logical-binary-operators
 [_ArrayExpression_]:              expressions/array-expr.md
+[_AsyncBlockExpression_]:         expressions/block-expr.md#async-blocks
+[_AwaitExpression_]:              expressions/await-expr.md
 [_AssignmentExpression_]:         expressions/operator-expr.md#assignment-expressions
 [_BlockExpression_]:              expressions/block-expr.md
 [_BreakExpression_]:              expressions/loop-expr.md#break-expressions
--- a/src/expressions/await-expr.md
+++ b/src/expressions/await-expr.md
@@ -1,0 +1,68 @@
+# Await expressions
+
+> **<sup>Syntax</sup>**\
+> _AwaitExpression_ :\
+> &nbsp;&nbsp; [_Expression_] `.` `await`
+
+Await expressions are legal only within an [async context], like an
+[`async fn`] or an [`async` block]. They operate on a [future]. Their effect
+is to suspend the current computation until the given future is ready
+to produce a value.
+
+More specifically, an `<expr>.await` expression has the following effect.
+
+1. Evaluate `<expr>` to a [future] `tmp`;
+2. Pin `tmp` using [`Pin::new_unchecked`];
+3. This pinned future is then polled by calling the [`Future::poll`] method and
+   passing it the current [task context](#task-context);
+3. If the call to `poll` returns [`Poll::Pending`], then the future
+   returns `Poll::Pending`, suspending its state so that, when the
+   surrounding async context is re-polled, execution returns to step
+   2;
+4. Otherwise the call to `poll` must have returned [`Poll::Ready`], in which case the
+   value contained in the [`Poll::Ready`] variant is used as the result
+   of the `await` expression itself.
+
+[`async fn`]: ../items/functions.md#async-functions
+[`async` block]: block-expr.md#async-blocks
+[future]: ../../std/future/trait.Future.html
+[_Expression_]: ../expressions.md
+[`Future::poll`]: ../../std/future/trait.Future.html#tymethod.poll
+[`Context`]: ../../std/task/struct.Context.html
+[`Pin::new_unchecked`]: ../../std/pin/struct.Pin.html#method.new_unchecked
+[`Poll::Pending`]: ../../std/task/enum.Poll.html#variant.Pending
+[`Poll::Ready`]: ../../std/task/enum.Poll.html#variant.Ready
+
+> **Edition differences**: Await expressions are only available beginning with
+> Rust 2018.
+
+## Task context
+
+The task context refers to the [`Context`] which was supplied to the
+current [async context] when the async context itself was
+polled. Because `await` expressions are only legal in an async
+context, there must be some task context available.
+
+[`Context`]: ../../std/task/struct.Context.html
+[async context]: ../expressions/block-expr.md#async-context
+
+## Approximate desugaring
+
+Effectively, an `<expr>.await` expression is roughly
+equivalent to the following (this desugaring is not normative):
+
+```rust,ignore
+let future = /* <expr> */;
+loop {
+    let mut pin = unsafe { Pin::new_unchecked(&mut future) };
+    match Pin::future::poll(Pin::borrow(&mut pin), &mut current_context) {
+        Poll::Ready(r) => break r,
+        Poll::Pending => yield Poll::Pending,
+    }
+}
+```
+
+where the `yield` pseudo-code returns `Poll::Pending` and, when
+re-invoked, resumes execution from that point. The variable
+`current_context` refers to the context taken from the async
+environment.
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -80,6 +80,74 @@ fn move_by_block_expression() {
 }
 ```
 
+## `async` blocks
+
+> **<sup>Syntax</sup>**\
+> _AsyncBlockExpression_ :\
+> &nbsp;&nbsp; `async` `move`<sup>?</sup> _BlockExpression_
+
+An *async block* is a variant of a block expression which evaluates to
+a *future*. The final expression of the block, if present, determines
+the result value of the future.
+
+Executing an async block is similar to executing a closure expression:
+its immediate effect is to produce and return an anonymous type.
+Whereas closures return a type that implements one or more of the
+[`std::ops::Fn`] traits, however, the type returned for an async block
+implements the [`std::future::Future`] trait. The actual data format for
+this type is unspecified.
+
+> **Note:** The future type that rustc generates is roughly equivalent
+> to an enum with one variant per `await` point, where each variant
+> stores the data needed to resume from its corresponding point.
+
+> **Edition differences**: Async blocks are only available beginning with Rust 2018.
+
+[`std::ops::Fn`]: ../../std/ops/trait.Fn.html
+[`std::future::Future`]: ../../std/future/trait.Future.html
+
+### Capture modes
+
+Async blocks capture variables from their environment using the same
+[capture modes] as closures. Like closures, when written `async {
+.. }` the capture mode for each variable will be inferred from the
+content of the block. `async move { .. }` blocks however will move all
+referenced variables into the resulting future.
+
+[capture modes]: ../types/closure.md#capture-modes
+[shared references]: ../types/pointer.md#shared-references-
+[mutable reference]: ../types/pointer.md#mutables-references-
+
+### Async context
+
+Because async blocks construct a future, they define an **async
+context** which can in turn contain [`await` expressions].  Async
+contexts are established by async blocks as well as the bodies of
+async functions, whose semantics are defined in terms of async blocks.
+
+[`await` expressions]: await-expr.md
+
+### Control-flow operators
+
+Async blocks act like a function boundary, much like
+closures. Therefore, the `?` operator and `return` expressions both
+affect the output of the future, not the enclosing function or other
+context. That is, `return <expr>` from within a closure will return
+the result of `<expr>` as the output of the future. Similarly, if
+`<expr>?` propagates an error, that error is propagated as the result
+of the future.
+
+Finally, the `break` and `continue` keywords cannot be used to branch
+out from an async block. Therefore the following is illegal:
+
+```rust,edition2018,compile_fail
+loop {
+    async move {
+        break; // This would break out of the loop.
+    }
+}
+```
+
 ## `unsafe` blocks
 
 > **<sup>Syntax</sup>**\
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -8,7 +8,10 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]
 >
 > _FunctionQualifiers_ :\
-> &nbsp;&nbsp; `const`<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
+> &nbsp;&nbsp; _AsyncConstQualifiers_<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
+>
+> _AsyncConstQualifiers_ :\
+> &nbsp;&nbsp; `async` | `const`
 >
 > _Abi_ :\
 > &nbsp;&nbsp; [STRING_LITERAL] | [RAW_STRING_LITERAL]
@@ -188,6 +191,104 @@ Exhaustive list of permitted structures in const functions:
 * `unsafe` blocks and `const unsafe fn` are allowed, but the body/block may only do
   the following unsafe operations:
     * calls to const unsafe functions
+
+## Async functions
+
+Functions may be qualified as async, and this can also be combined with the
+`unsafe` qualifier:
+
+```rust,edition2018
+async fn regular_example() { }
+async unsafe fn unsafe_example() { }
+```
+
+Async functions do no work when called: instead, they
+capture their arguments into a future. When polled, that future will
+execute the function's body.
+
+An async function is roughly equivalent to a function
+that returns [`impl Future`] and with an [`async move` block][async-blocks] as
+its body:
+
+```rust,edition2018
+// Source
+async fn example(x: &str) -> usize {
+    x.len()
+}
+```
+
+is roughly equivalent to:
+
+```rust,edition2018
+# use std::future::Future;
+// Desugared
+fn example<'a>(x: &'a str) -> impl Future<Output = usize> + 'a {
+    async move { x.len() }
+}
+```
+
+The actual desugaring is more complex:
+
+- The return type in the desugaring is assumed to capture all lifetime
+  parameters from the `async fn` declaration. This can be seen in the
+  desugared example above, which explicitly outlives, and hence
+  captures, `'a`.
+- The [`async move` block][async-blocks] in the body captures all function
+  parameters, including those that are unused or bound to a `_`
+  pattern. This ensures that function parameters are dropped in the
+  same order as they would be if the function were not async, except
+  that the drop occurs when the returned future has been fully
+  awaited.
+
+For more information on the effect of async, see [`async` blocks][async-blocks].
+
+[async-blocks]: ../expressions/block-expr.md#async-blocks
+[`impl Future`]: ../types/impl-trait.md
+
+> **Edition differences**: Async functions are only available beginning with
+> Rust 2018.
+
+### Combining `async` and `unsafe`
+
+It is legal to declare a function that is both async and unsafe. The
+resulting function is unsafe to call and (like any async function)
+returns a future. This future is just an ordinary future and thus an
+`unsafe` context is not required to "await" it:
+
+```rust,edition2018
+// Returns a future that, when awaited, dereferences `x`.
+//
+// Soundness condition: `x` must be safe to dereference until
+// the resulting future is complete.
+async unsafe fn unsafe_example(x: *const i32) -> i32 {
+  *x
+}
+
+async fn safe_example() {
+    // An `unsafe` block is required to invoke the function initially:
+    let p = 22;
+    let future = unsafe { unsafe_example(&p) };
+    
+    // But no `unsafe` block required here. This will
+    // read the value of `p`:
+    let q = future.await;
+}
+```
+
+Note that this behavior is a consequence of the desugaring to a
+function that returns an `impl Future` -- in this case, the function
+we desugar to is an `unsafe` function, but the return value remains
+the same.
+
+Unsafe is used on an async function in precisely the same way that it
+is used on other functions: it indicates that the function imposes
+some additional obligations on its caller to ensure soundness. As in any
+other unsafe function, these conditions may extend beyond the initial
+call itself -- in the snippet above, for example, the `unsafe_example`
+function took a pointer `x` as argument, and then (when awaited)
+dereferenced that pointer. This implies that `x` would have to be
+valid until the future is finished executing, and it is the callers
+responsibility to ensure that.
 
 ## Attributes on functions
 
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -67,6 +67,7 @@
         - [If and if let expressions](expressions/if-expr.md)
         - [Match expressions](expressions/match-expr.md)
         - [Return expressions](expressions/return-expr.md)
+        - [Await expressions](expressions/await-expr.md)
 
 - [Patterns](patterns.md)
 


src/items/functions.md
```diff
@@ -189,6 +189,50 @@ Exhaustive list of permitted structures in const functions:
...
+  parameters from the `async fn` declaration. This can be seen in the
+  desugared example above, which explicitly outlives (and hence
+  captures) `'a`.
+- The `async move` in the body includes some code to capture otherwise
+  unused function parameters. This code ensures that function
+  parameters are dropped in the same order as they would be if the
+  function were not async, except that the drop occurs when the
+  returned future has been fully awaited.
+  
+For more information on the effect of async, see async blocks.
```
Centril: A link here to async blocks.

src/items/functions.md
```diff
@@ -189,6 +189,50 @@ Exhaustive list of permitted structures in const functions:
...
+The actual desugaring is performed internally to the compiler and is mildly
+more complex:
+
+- The `impl Future` which returns is assumed to capture all lifetime
+  parameters from the `async fn` declaration. This can be seen in the
+  desugared example above, which explicitly outlives (and hence
+  captures) `'a`.
+- The `async move` in the body includes some code to capture otherwise
+  unused function parameters. This code ensures that function
+  parameters are dropped in the same order as they would be if the
```
Centril: A link to drop order would be good; However, @matthewjasper's PR hasn't landed so please make a follow-up issue.

src/items/functions.md
```diff
@@ -8,7 +8,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]
 >
 > _FunctionQualifiers_ :\
-> &nbsp;&nbsp; `const`<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
+> &nbsp;&nbsp; `async`<sup>?</sup> `const`<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
```
Centril: This grammar is wrong; `async const fn` will not parse.

src/expressions/block-expr.md
```diff
@@ -80,6 +80,57 @@ fn move_by_block_expression() {
...
 ```
 
+## `async` blocks
+
+> **<sup>Syntax</sup>**\
+> _AsyncBlockExpression_ :\
+> &nbsp;&nbsp; `async` `move`<sup>?</sup> _BlockExpression_
+
+An *async block* is a variant of a block expression which evalutes to
+a *future*. The final expression of the block (if present) determines
```
Centril:
```suggestion
a *future*. The final expression of the block, if present, determines
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+the result value of the future.
+
+Executing an async block is thus similar to executing a closure
+expression: its immediate effect is to produce and return a struct.
+Whereas closures return a struct that implements one or more of the
+`std::ops::Fn` traits, however, the type returned for an async block
+implements the `std::future::Future` trait. The actual data format for
+this type is undefined, but it is roughly equivalent to an enum with
```
Centril:
```suggestion
this type is unspecified, but it is roughly equivalent to an enum with
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+
+> **<sup>Syntax</sup>**\
+> _AsyncBlockExpression_ :\
+> &nbsp;&nbsp; `async` `move`<sup>?</sup> _BlockExpression_
+
+An *async block* is a variant of a block expression which evalutes to
+a *future*. The final expression of the block (if present) determines
+the result value of the future.
+
+Executing an async block is thus similar to executing a closure
```
Centril:
```suggestion
Executing an async block is similar to executing a closure
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+expression: its immediate effect is to produce and return a struct.
```
Centril: Closures are not described in this way in the reference -- it would be sufficient to say "anonymous type" here.

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+Whereas closures return a struct that implements one or more of the
+`std::ops::Fn` traits, however, the type returned for an async block
+implements the `std::future::Future` trait. The actual data format for
+this type is undefined, but it is roughly equivalent to an enum with
+one variant per `await` point (such that we store the data needed to
```
Centril:
```suggestion
one variant per `await` point such that the data needed to
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+resume from that point).
```
Centril:
```suggestion
resume from that point is stored.
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+
+Async blocks capture variables from their environment in a way similar
+to closures. That is, when the async block is first executed, any
+variables referenced in the block will be "captured" into the
+resulting future. In the absence of the `move` keyword, async blocks
+will capture in the weakest way possible, depending on how the
+variable is used (just as closures do). That is, if the variable is
+only used to produce shared references, or is taken by value but is of
```
Centril: Link to shared references (https://doc.rust-lang.org/nightly/reference/types/pointer.html#shared-references-)

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+`Copy` type, they will capture an `&`-reference to the variable. If
```
Centril:
```suggestion
`Copy` type, async blocks will capture a shared reference to the variable. If
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+the variable is mutated, they will capture an `&mut`-reference.
```
Centril:
```suggestion
the variable is mutated, they will capture a mutable reference.
```
(and link to mutable references)

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+Finally, if the variable is moved, they will capture the variable "by
```
Centril:
```suggestion
Finally, if the variable is moved, async blocks will capture the variable "by
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+move". If the async block has the move keyword (`async move { ... })`,
+however, then all referenced variables are moved into the resulting
+future, no matter how they are used. This is the most common form.
```
Centril:
```suggestion
future, no matter how they are used.
```
It's not the reference's job to say what is and isn't common or what the use cases are; it should just define the language.

Centril: Overall it feels like this duplicates a lot of https://doc.rust-lang.org/nightly/reference/types/closure.html#capture-modes.

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+
+Executing an async block is thus similar to executing a closure
+expression: its immediate effect is to produce and return a struct.
+Whereas closures return a struct that implements one or more of the
+`std::ops::Fn` traits, however, the type returned for an async block
+implements the `std::future::Future` trait. The actual data format for
+this type is undefined, but it is roughly equivalent to an enum with
+one variant per `await` point (such that we store the data needed to
+resume from that point).
+
```
Centril:
```suggestion

### Capture modes

```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+will capture in the weakest way possible, depending on how the
+variable is used (just as closures do). That is, if the variable is
+only used to produce shared references, or is taken by value but is of
+`Copy` type, they will capture an `&`-reference to the variable. If
+the variable is mutated, they will capture an `&mut`-reference.
+Finally, if the variable is moved, they will capture the variable "by
+move". If the async block has the move keyword (`async move { ... })`,
+however, then all referenced variables are moved into the resulting
+future, no matter how they are used. This is the most common form.
+
```
Centril:
```suggestion
### <Invent some title>
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+The bodies of async blocks differ from ordinary blocks in a few ways.
```
Centril: This leading sentence isn't necessary I think.

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+The most obvious is that they can contain `await` expressions. These
```
Centril:
```suggestion
The bodies of async blocks can contain `await` expressions. These
```
The word "obvious" is obviously banned ;)

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+expressions are applied to futures and cause the async block to poll
+the future and return control to the caller if the data is not yet
+ready (see the section on [await expressions](./await-expr.html) for more details).
```
Centril:
```suggestion
ready. See the section on [await expressions] for more details.
```
Please avoid inline links.

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+
+The other difference is that async blocks are considered a sort of
```
Centril:
```suggestion
Async blocks are considered a sort of
```

src/expressions/block-expr.md
```diff
@@ -80,6 +80,64 @@ fn move_by_block_expression() {
...
+the future. Similarly, if `<expr>?` propagates an error, that error is
+propagated as the result of the future.
+
+Finally, the `break` and `continue` keywords cannot be used to branch
+out from an async block. Therefore the following is illegal:
+
+```rust
+loop {
+    async move {
+        break; // would break out of the loop
```
Centril:
```suggestion
        break; // This would break out of the loop.
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
+# Await expressions
+
+> **<sup>Syntax</sup>**\
+> _AwaitExpression_ :\
+> &nbsp;&nbsp; [_Expression_] `.` `await`
+
+Await expressions are legal only within `async` contexts, like an
```
Centril: The reference should separately define "async contexts" (also you use backticks inconsistently, either use it everywhere or nowhere) and then refer to it.

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+`async fn` or an `async` block. They operate on a future. Their effect
+is to "pause" the current computation until the given future is ready
```
Centril:
```suggestion
is to "suspend" the current computation until the given future is ready
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+to produce a value.
+
+More specifically, a `f.await` expression has the following effect.
```
Centril:
```suggestion
More specifically, an `<expr>.await` expression has the following effect.
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. evaluate `f` to a future `tmp` using the `IntoFuture:into_future` method;
```
Centril:
```suggestion
1. Evaluate `f` to a future `tmp` using the `IntoFuture::into_future` method;
```

Also there's no such trait...?

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+2. `tmp` is pinned using `Pin::new_unchecked`;
```
Centril:
```suggestion
2. Pin `tmp` using `Pin::new_unchecked`;
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+3. This pinned future is then polled by invoking the [`Future::poll`] method and
```
Centril:
```suggestion
3. This pinned future is then polled by calling the [`Future::poll`] method and
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+   passing it the current [task context] (see below);
+3. if the result returns `Poll::Pending`, then the future returns
```
Centril:
```suggestion
3. If the result evaluates to `Poll::Pending`, then the future returns
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+   `Poll::Pending`, suspending its state so that, when re-invoked, we will return
```
Centril:
```suggestion
   `Poll::Pending`, suspending its state so that, when re-invoked,
    execution resumes at step 2;
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+   to step 2;
+4. if the result returns `Poll::Ready`, then the result is used as the result
```
Centril:
```suggestion
4. If the result evaluates to `Poll::Ready`, then the result is used as the result
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+   of the await expression itself.
```
Centril:
```suggestion
   of the `await` expression itself.
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+
+[_Expression_]: expressions.html
+[`Future::poll`]: https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll
+[task context]: https://doc.rust-lang.org/std/task/struct.Context.html
+
```
Centril:
```suggestion

## Task context

```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+The [task context] refers to the context which was supplied to the
+current async block/fn when it itself was polled. Because `await`
```
Centril:
```suggestion
current async context when it itself was polled. Because `await`
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,50 @@
...
+expressions are only legal in an async context, there must be some
+task context available.
+
+Effectively, an `f.await` expression desugars into something roughly
```
Centril:
```suggestion
Effectively, an `f.await` expression is roughly
```

src/expressions/async-block-expr.md
```diff
@@ -0,0 +1,152 @@
+# Block expressions
```
Centril: Unused duplicated file?

nikomatsakis: yeah whoops!

nikomatsakis:
So I get a bunch of errors from the examples like

```
error[E0670]: `async fn` is not permitted in the 2015 edition
 --> /tmp/mdbookaDvS38/items/functions.md:198:1
  |
3 | async fn regular_example() { }
  | ^^^^^
error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `unsafe`
 --> /tmp/mdbookaDvS38/items/functions.md:199:7
  |
4 | async unsafe fn unsafe_example() { }
  |       ^^^^^^ expected one of 8 possible tokens here
error[E0425]: cannot find value `async` in this scope
 --> /tmp/mdbookaDvS38/items/functions.md:199:1
  |
4 | async unsafe fn unsafe_example() { }
  | ^^^^^ not found in this scope
error[E0658]: async fn is unstable
 --> /tmp/mdbookaDvS38/items/functions.md:198:1
  |
3 | async fn regular_example() { }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: for more information, see https://github.com/rust-lang/rust/issues/50547
  = help: add #![feature(async_await)] to the crate attributes to enable
```

Any tips on how to address these? :)

nikomatsakis:
(Especially the edition flag)

Centril:
> I'm not sure what the best reviewing process is here.

Not sure either, but the current process is basically: "@Centril and @ehuss reviews, we approve, and then we merge".

Centril:
> (Especially the edition flag)

I think you can follow these examples: https://github.com/rust-lang-nursery/reference/search?q=edition2018&unscoped_q=edition2018

Centril:
Speaking of editions; I think it was not mentioned in the PR that the new stuff only works in edition 2018. Would be good to work that in (see https://github.com/rust-lang-nursery/reference/search?q=edition&unscoped_q=edition for examples).

src/expressions/await-expr.md
Havvy:
```suggestion
1. evaluate `f` to a future temporary variable `tmp` using the `IntoFuture:into_future` method;
```

I don't like "future" there, but don't really see a better way to succinctly convey the information. It should probably link to the Future trait as well.

src/expressions/await-expr.md
Havvy: "re-invoked" to "re-polled"?

src/expressions/await-expr.md
Havvy: "it itself" should be changed to "the async context" as well. It took me a minute to figure out what "it" was.

src/expressions/await-expr.md
Centril: Sure, that might be better -- I think both work.

src/expressions/block-expr.md
Havvy: The second clause (after the "but") should be in a Note also stating that it is what rustc does.

src/expressions/block-expr.md
Havvy: Also, can't we get rid of the "See also" line by putting the link on the first instance of "`await` expressions" in the paragraph?

Havvy:
Given that the semantics of `move` and `?` are affected the same in a closure as in an async block, we could get a name for these psuedo-function boundaries and put information on capturing groups in there, the propagation boundary for `?` can refer to it, and we can just say closures and async blocks are both whatever we call these psuedo-function boundaries, linking to that.

If we don't want to do that, then we should say `move` behaves the same way as on closures, linking to that, and in the capturing groups section on the closures page, say that the section applies to both closures and async blocks, linking back to async blocks.

eddyb:
Nit: the title doesn't mention async/await.

src/items/functions.md
nikomatsakis: oh yeah I meant to check on that, thanks

nikomatsakis: ugh, it's going to be a pain then. It feels like we *should* parse this (and reject later).

Centril: Maybe... but that also entails deciding on an order, which I would prefer not to do at this juncture.

For now you can add a new production `effect_qualifier ::= "const" | "async"` and then use that one. That doesn't seem too cumbersome.

(The extern bit is also wrong and should be added to `effect_qualifier`)

Centril: Tho `unsafe extern "C" fn f() {}` is valid syntax which does add a bit of pain.

src/expressions/await-expr.md
nikomatsakis: hmm yes I guess this changed since the RFC :)

src/items/functions.md
nikomatsakis: > but that also entails deciding on an order, which I would prefer not to do at this juncture.

I won't debate the point, but since it's an error if they occur together, it doesn't really. =) 

(I think I personally would prefer to accept qualifiers in any order, and then sort them in some canonical order with rustfmt, but anyway this PR just aims to document what's there.)

I introduced `AsyncConstQualifier` for now.

src/items/functions.md
nikomatsakis: I've added a note to #634

nikomatsakis:
> Given that the semantics of move and ? are affected the same in a closure as in an async block, we could get a name for these psuedo-function boundaries and put information on capturing groups in there, the propagation boundary for ? can refer to it, and we can just say closures and async blocks are both whatever we call these psuedo-function boundaries, linking to that.

I did link to closures to explain capture modes. I did not do so for `?` but I agree that there is *some* concept that is similar between them.

src/items/functions.md
Centril: > I won't debate the point, but since it's an error if they occur together, it doesn't really. =)

(It matters wrt. `#[cfg(FALSE)]` which in turn creates a distinction between syntax and semantics (AST and HIR) that matters wrt. stability guarantees)

> (I think I personally would prefer to accept qualifiers in any order, and then sort them in some canonical order with rustfmt, but anyway this PR just aims to document what's there.)

(Yeah maybe; It ties into the question of effect polymorphism and such... but for now I am mainly in favor of procrastination ^^)

> I introduced `AsyncConstQualifier` for now.

Thanks!

nikomatsakis:
Note to self: document `async unsafe` behavior

src/items/functions.md
```diff
@@ -189,6 +192,57 @@ Exhaustive list of permitted structures in const functions:
...
+
+Async functions do no work when called: instead, they
+capture their arguments into a future. When polled, that future will
+execute the function's body.
+
+An async function is roughly equivalent to a function
+that returns [`impl Future`] and with an [`async move` block][async-blocks] as
+its body:
+
+```rust,edition2018
```
ehuss: I think we'll need to add `,ignore` here.  This won't compile because the identifier `example` is used twice.  I can't really think of a nice way to make it work better. I don't like using `ignore` because examples often end up broken, but I think changing the identifier would make it less clear.

nikomatsakis: Maybe just two blocks, like so:

---

```rust
// Source
async fn example(x: &str) -> usize {
    x.len()
}
```

is roughly equivalent to 

```rust
// Desugared
fn example<'a>(x: &'a str) -> impl Future<Output = usize> + 'a {
    async move { x.len() }
}
```

ehuss: I think that would be good, too.  If you do that, be sure to add:
`# use std::future::Future;` so it can compile.

src/expressions/await-expr.md
```diff
@@ -0,0 +1,68 @@
...
+[async context]: expressions/block-expr.html#async-context
+
+## Approximate desugaring
+
+Effectively, an `<expr>.await` expression is roughly
+equivalent to the following (this desugaring is not normative):
+
+```rust,ignore
+loop {
+    let tmp = /* <expr> */;
```
Havvy: Shouldn't this line be outside of the loop? `tmp` isn't re-evaluated every single time, right? Also, should probably call it `awaited` and not `tmp`.

ehuss:
Just a heads up, I have rebased this PR on latest master to resolve the conflicts, so be careful updating locally.

The reference has migrated to mdbook 0.3.  This means that links should use the `.md` extension, and are relative to the page they are on (previously they were relative to the root of the book). Please don't hesitate to ask if you have any questions or any trouble updating. Sorry for the interruption!

cramertj:
Are there remaining blockers to landing this?

ehuss:
> Are there remaining blockers to landing this?

The reference only includes stable features, so normally we wait until it is stabilized on nightly.  IIUC, it is currently waiting on FCP in #62149?  

There are also some unresolved comments above.  I'll take care of the formatting issue, but I don't know enough about async to do much else.  Is there someone who can pick it up since Niko isn't around?

There's an unresolved issue about "unsafe" being undocumented, but I personally think it would be ok to land that separately.  Others may disagree, though.

nikomatsakis:
I pushed two commits -- one addresses @Havvy's nit, the other documents `async unsafe`. I'm not sure what remaining things are left to be done.

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+// Returns a future that, when awaited, dereferences `x`.
+//
+// Unsafe condition: `x` must be safe to dereference until
+// the resulting future is complete.
+async unsafe fn unsafe_example(x: *const i32) -> i32 {
+  *x
+}
+
+async fn safe_example() {
+    // Unsafe keyword required to invoke the function
```
Centril:
```suggestion
    // An `unsafe` block is required to invoke the function initially:
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+    // initially:
```
Centril:
```suggestion
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+    let p = 22;
+    let future = unsafe { unsafe_example(&p) };
+    
+    // But no unsafe keyword required here. This will
```
Centril:
```suggestion
    // But no `unsafe` block required here. This will
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+[`impl Future`]: ../types/impl-trait.md
+
+> **Edition differences**: Async functions are only available beginning with
+> Rust 2018.
+
+### Combining async and unsafe
+
+It is legal to declare a function that is both async and unsafe. The
+resulting function is unsafe to call and (like any async function)
+returns a future. This future is just an ordinary future and thus the
```
Centril:
```suggestion
returns a future. This future is just an ordinary future and thus an
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+unsafe keyword keyword is not required to "await" it:
+
+```rust,edition2018
+// Returns a future that, when awaited, dereferences `x`.
+//
+// Unsafe condition: `x` must be safe to dereference until
```
Centril:
```suggestion
// Soundness condition: `x` must be safe to dereference until
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+```
+
+Note that this behavior is a consequence of the desugaring to a
+function that returns an `impl Future` -- in this case, the function
+we desugar to is an `unsafe` function, but the return value remains
+the same.
+
+Unsafe is used on an async function in precisely the same way that it
+is used on other functions: it indicates that the function imposes
+some additional obligations on its caller to ensure safety. As in any
```
Centril:
```suggestion
some additional obligations on its caller to ensure soundness. As in any
```
(feels a bit more precise in a reference setting since "safe" feels more ambiguous)

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+other unsafe function, these conditions may extend beyond the initial
+call itself -- in the snippet above, for example, the `unsafe_example`
+function took a pointer `x` as argument, and then (when awaited)
+dereferenced that pointer. This implies that `x` would have to be
+valid until the future is finished execution, and it is the callers
```
Centril:
```suggestion
valid until the future is finished executing, and it is the callers
```

src/items/functions.md
```diff
@@ -189,6 +192,105 @@ Exhaustive list of permitted structures in const functions:
...
+job to ensure that.
```
Centril:
```suggestion
responsibility to ensure that.
```

src/expressions/await-expr.md
```diff
@@ -0,0 +1,68 @@
+# Await expressions
+
+> **<sup>Syntax</sup>**\
+> _AwaitExpression_ :\
+> &nbsp;&nbsp; [_Expression_] `.` `await`
+
+Await expressions are legal only within `async` contexts, like an
```
Centril: Can you also link to async context from here?

src/items/functions.md
```diff
@@ -8,7 +8,10 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]
 >
 > _FunctionQualifiers_ :\
-> &nbsp;&nbsp; `const`<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
+> &nbsp;&nbsp; _AsyncConstQualifiers_ <sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
```
Centril:
```suggestion
> &nbsp;&nbsp; _AsyncConstQualifiers_<sup>?</sup> `unsafe`<sup>?</sup> (`extern` _Abi_<sup>?</sup>)<sup>?</sup>
```

src/expressions.md
```diff
@@ -13,6 +13,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_OperatorExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_GroupedExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ArrayExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_Awaitxpression_]\
```
F001:
```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; | [_Awaitxpression_]\
&nbsp;&nbsp; &nbsp;&nbsp; | [_AwaitExpression_]\
```

Centril:
@nikomatsakis Btw, you can apply suggestions in bulk as a single commit if you go to the "Files changed" tab. Use the ["Add suggestion to batch"](https://help.github.com/en/articles/incorporating-feedback-in-your-pull-request) button when you are there.

Centril:
@nikomatsakis Since the stabilization PR landed, want to make the finishing touches and then squash? Let's merge after that.

Centril:
Yep, thanks! We'll have to remember to adjust this if we adjust the lowering wrt. the temporary lifetime stuff.

Issue #244, opened 2018-02-15
Replace remaining "C-like enumerations" with "fieldless enumerations"


alercah:
See rust-lang/rust#46348. Looks like a few snuck through: https://doc.rust-lang.org/nightly/reference/type-layout.html#reprc-enums, for instance.

Havvy:
Already done in my branch for the `#[repr(C, Int)]` on enums with fields RFC.

PR #639, opened 2019-07-14
Update documentation for arbitrary_enum_discriminant feature

jswrenn:
Updates documentation to reflect the **still unstable** `arbitrary_enum_discriminant`feature (tracking issue: rust-lang/rust#60553).

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -13,8 +13,8 @@
 >
 > _EnumItem_ :\
 > &nbsp;&nbsp; _OuterAttribute_<sup>\*</sup>\
-> &nbsp;&nbsp; [IDENTIFIER]&nbsp;( _EnumItemTuple_ | _EnumItemStruct_
->                                | _EnumItemDiscriminant_ )<sup>?</sup>
+> &nbsp;&nbsp; [IDENTIFIER]&nbsp;( _EnumItemTuple_ | _EnumItemStruct_ )<sup>?</sup>
+>                                _EnumItemDiscriminant_<sup>?</sup>
 >
 > _EnumItemTuple_ :\
 > &nbsp;&nbsp; `(` [_TupleFields_]<sup>?</sup> `)`
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
 ```
 
 In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
-called an enum variant. Each enum instance has a _discriminant_ which is an
-integer associated to it that is used to determine which variant it holds. An
-opaque reference to this discriminant can be obtained with the
-[`mem::discriminant`] function.
+called an enum variant.
 
-## Custom Discriminant Values for Field-Less Enumerations
+## Discriminants
 
-If there is no data attached to *any* of the variants of an enumeration,
-then the discriminant can be directly chosen and accessed.
+Each enum instance has a _discriminant_: an integer logically associated to it
+that is used to determine which variant it holds. An opaque reference to this
+discriminant can be obtained with the [`mem::discriminant`] function.
 
-These enumerations can be cast to integer types with the `as` operator by a
-[numeric cast]. The enumeration can optionally specify which integer each
-discriminant gets by following the variant name with `=` followed by a [constant
-expression]. If the first variant in the declaration is unspecified, then it is
-set to zero. For every other unspecified discriminant, it is set to one higher
-than the previous variant in the declaration.
+Under the [default representation], the discriminant is interpreted as
+an `isize` value. However, the compiler is allowed to use a smaller type (or
+another means of distinguishing variants) in its actual memory layout.
+
+If the [primitive representation] or the [`C` representation] is used, the
+leading bytes of a variant (e.g., two bytes if `#[repr(u16)]` is used), will
+correspond exactly to the discriminant.
+
+### Assigning Discriminant Values
+
+#### Explicit Discriminants
+
+In two circumstances, the discriminant of a variant may be explicitly set by
+following the variant name with `=` and a [constant expression]:
+
+<ol>
+<li>
+
+if the enumeration is "C-like" (i.e., it has no tuple or struct variants); e.g.:
+
+```rust
+# #![feature(arbitrary_enum_discriminant)]
+enum Enum {
+    Foo = 3,
+    Bar = 2,
+    Baz = 1,
+}
+```
+</li>
+<li>
+
+if a [primitive representation] is used; e.g.:
+
+```rust
+# #![feature(arbitrary_enum_discriminant)]
+#[repr(u8)]
+enum Enum {
+    Unit = 3,
+    Tuple(u16),
+    Struct {
+        a: u8,
+        b: u16,
+    } = 1,
+}
+```
+</li>
+</ol>
+
+#### Implicit Discriminants
+
+If a discriminant for a variant is not specified, then it is set to one higher
+than the discriminant of the previous variant in the declaration. If the
+discriminant of the first variant in the declaration is unspecified, then
+it is set to zero.
 
 ```rust
 enum Foo {
@@ -84,14 +130,11 @@ let baz_discriminant = Foo::Baz as u32;
 assert_eq!(baz_discriminant, 123);
 ```
 
-Under the [default representation], the specified discriminant is interpreted as
-an `isize` value although the compiler is allowed to use a smaller type in the
-actual memory layout. The size and thus acceptable values can be changed by
-using a [primitive representation] or the [`C` representation].
+#### Restrictions
 
 It is an error when two variants share the same discriminant.
 
-```rust,ignore
+```rust,compile_fail
 enum SharedDiscriminantError {
     SharedA = 1,
     SharedB = 1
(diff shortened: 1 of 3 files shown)


src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+
+#### Explicit Discriminants
+
+In two circumstances, the discriminant of a variant may be explicitly set by
+following the variant name with `=` and a [constant expression]:
+
+<ol>
+<li>
+
+if the enumeration is "C-like" (i.e., it has no tuple or struct variants); e.g.:
```
Havvy:
```suggestion
if the enumeration is field-less. For example:
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+
+```rust
+# #![feature(arbitrary_enum_discriminant)]
```
Havvy: Feature flags aren't needed; the prose will only be merged after the feature is stable.

```suggestion
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+enum Enum {
+    Foo = 3,
+    Bar = 2,
+    Baz = 1,
+}
+```
+</li>
+<li>
+
+if a [primitive representation] is used; e.g.:
```
Havvy:
```suggestion
if a [primitive representation] is used. For example:
```

src/items/enumerations.md
```diff
@@ -122,6 +165,53 @@ enum OverflowingDiscriminantError2 {
 }
 ```
 
+### Accessing Discriminant Values
+
+#### Casting
+
+If there is no data attached to *any* of the variants of an enumeration, then
```
Havvy: There's no need to describe a field-less enum here. Just use the term.

```suggestion
If the enum is a field-less enum, then
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
-[`mem::discriminant`] function.
+called an enum variant.
 
-## Custom Discriminant Values for Field-Less Enumerations
+## Discriminants
 
-If there is no data attached to *any* of the variants of an enumeration,
-then the discriminant can be directly chosen and accessed.
+Each enum instance has a _discriminant_: an integer logically associated to it
+that is used to determine which variant it holds. An opaque reference to this
```
Havvy: Using `mem::discriminant` should be in the accessing the discriminant section.

src/type-layout.md
```diff
@@ -357,7 +357,7 @@ used with any other representation.
 [`size_of`]: ../std/mem/fn.size_of.html
 [`Sized`]: ../std/marker/trait.Sized.html
 [dynamically sized types]: dynamically-sized-types.html
-[C-like enumerations]:  items/enumerations.html#custom-discriminant-values-for-field-less-enumerations
```
Havvy: This link change is wrong. We aren't properly defining a field-less enum anywhere after this PR.

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
 ```
 
 In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
-called an enum variant. Each enum instance has a _discriminant_ which is an
-integer associated to it that is used to determine which variant it holds. An
-opaque reference to this discriminant can be obtained with the
-[`mem::discriminant`] function.
+called an enum variant.
```
Havvy: The re-organization looks fine, but we lose our definition of a field-less enum. This suggestion re-adds it. I don't actually know if the anchor actually works. It should just create an anchor that can be linked too without changing any styles or making it clickable.

```suggestion
called an enum variant.

An enum where no constructors contain fields are called a *<a name="field-less-enum">field-less enum</a>*.
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+    Baz = 1,
+}
+```
+</li>
+<li>
+
+if a [primitive representation] is used; e.g.:
+
+```rust
+# #![feature(arbitrary_enum_discriminant)]
```
Havvy: Feature flag not needed.

```suggestion
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
-discriminant gets by following the variant name with `=` followed by a [constant
-expression]. If the first variant in the declaration is unspecified, then it is
-set to zero. For every other unspecified discriminant, it is set to one higher
-than the previous variant in the declaration.
+Under the [default representation], the discriminant is interpreted as
+an `isize` value. However, the compiler is allowed to use a smaller type (or
+another means of distinguishing variants) in its actual memory layout.
+
+If the [primitive representation] or the [`C` representation] is used, the
+leading bytes of a variant (e.g., two bytes if `#[repr(u16)]` is used), will
```
Havvy:
```suggestion
leading bytes of a variant (for example, two bytes if `#[repr(u16)]` is used), will
```

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
-[numeric cast]. The enumeration can optionally specify which integer each
-discriminant gets by following the variant name with `=` followed by a [constant
-expression]. If the first variant in the declaration is unspecified, then it is
-set to zero. For every other unspecified discriminant, it is set to one higher
-than the previous variant in the declaration.
+Under the [default representation], the discriminant is interpreted as
+an `isize` value. However, the compiler is allowed to use a smaller type (or
+another means of distinguishing variants) in its actual memory layout.
+
+If the [primitive representation] or the [`C` representation] is used, the
```
Havvy: I don't understand what this paragraph is trying to say.

src/items/enumerations.md
```diff
@@ -56,22 +56,68 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+leading bytes of a variant (e.g., two bytes if `#[repr(u16)]` is used), will
+correspond exactly to the discriminant.
+
+### Assigning Discriminant Values
```
Havvy: For this entire section, it only makes sense for field-less enums or enums with a primitive representation (or the `C` representation?). The way it is written, it seems like explicit discriminants are only for those and implicit applies to all. Instead, it should state that having a non-opaque (feel free to pick a different descriptor) discriminant applies in those situations and in all others, it is opaque and must only be accessed via `mem::discriminant`.

jswrenn:
Thank you so much for your review, @Havvy! If it's alright, I'm going to defer removing the feature flags until this gets a little closer to being merged; otherwise, I can't run `mdbook tests`.

## Terminology

I have a lot of uncertainty about terminology. I only just saw that there's prior agreement (rust-lang/rust#46348, #244) to move away from the term "C-like enumeration" (which I believe is a very poor term) and towards "field-less enumeration". I am thrilled to strike the phrase "C-like" from this PR.

However, using "field-less enumeration" as a __synonym__ for "C-like" isn't quite appropriate.

A **C-like enumeration** is one in which every variant is unit-like; for example:
```rust
enum CLike {
  VariantA,
  VariantB,
  VariantC,
}
```

All C-like enumerations are field-less, because they cannot include tuple-like or struct-like variants. **However, not all field-less enums are are C-like.** For instance:
```rust
enum Fieldless {
    Unit,
    Tuple(),
    Struct{},
}
```

This distinction unfortunately (ugh) matters:
 1. _Only_ **C-like** enums may specify explicit determinants without specifying a repr.
 2. _Only_ **field-less** enums (regardless of whether they're also C-like) may be `as`-casted to their discriminant values.

It would be nice to have terms to describe both of these sets. **My opinion:** 
- "Field-less" is the right term to describe the set of enums which can be `as`-casted
-  However it cannot then _also_ be used as a synonym for "C-like", so **we should identify an alternative term** to "C-like" to describe enums in which all variant are unit-like.

src/items/enumerations.md
jswrenn: Per https://github.com/rust-lang-nursery/reference/pull/639#issuecomment-511515290, these terms may not be truly interchangeable.

src/items/enumerations.md
jswrenn: Per https://github.com/rust-lang-nursery/reference/pull/639#issuecomment-511515290, these terms may not be completely interchangeable. If "field-less" is meant to be synonymous to "C-like", then this change would be incorrect.

Havvy:
Well, that is quite unfortunate of an edge case. I'm failing to think of any good names.

ehuss:
@jswrenn Just letting you know that the reference has migrated to mdbook 0.3 (from 0.1). This means that the style of links are slightly different. They should use the `.md` extension, and are relative to the page they are on (previously they were relative to the root of the book). It looks like this PR should be fairly easy to rebase to resolve the conflicts. Please don't hesitate to ask if you have any questions or any trouble rebasing.

fee1-dead:
@jswrenn Are you still working on this? If not, I would like to work on it.

JohnTitor:
According to the label, this is waiting for stabilization.

jswrenn:
> @jswrenn Are you still working on this? If not, I would like to work on it.

I'm not! Go for it! 

(Sorry for the delayed reply; I thought I had already replied.)

ehuss:
Closing as this has been continued by #1055.

Issue #638, opened 2019-07-13
#[link] attribute should better describe platform specific handling


ralfbiedert:
When working cross platform, in particular when producing a dylib from Rust and then again consuming it, the current resolution of `#[link(name = "mylib"]` is somewhat confusing. 

- If you start developing on Windows, Rust will produce a `mylib.dll` and `mylib.dll.lib`. To use this lib again from Rust you will have to specify `#[link(name = "mylib.dll")]`, thus giving the impression that the full file name has to be specified. On Mac, however, `#[link(name = "libmylib.dylib"]` will fail (likewise Linux).

- If you start developing on Mac and Linux, `#[link(name = "mylib")]` just works, giving you the impression Rust handles the name resolution (fully) automatically like other platforms that just require the base name. 

In fact, the correct way to cross platform link against a dylib produced by Rust seems to be:

```rust
#[cfg_attr(target_os = "windows", link(name = "mylib.dll"))]
#[cfg_attr(not(target_os = "windows"), link(name = "mylib"))]
extern "C" {}
```

Since according to [this issue](https://github.com/rust-lang/rust/issues/62588) the current behavior can't be fixed and is "stable", I believe this should be documented somewhere. For me, the `#[link]` attribute was where I started my debug journey originally.

The documentation could be something like:

> Note that on Mac and Linux `name` is the base name of the actual library (e.g., `#[link(name = "mylib")]` if you want to link against `libmylib.so`). On Windows, the base name of the `.lib` file has to be provided. 
>
>For 3rd party libraries such as `mylib.dll` and `mylib.lib`, this still equals `#[link(name = "mylib")]`, but for Rust produced library pairs `mylib.dll` and `mylib.dll.lib` a `#[link(name = "mylib.dll")]` is needed instead.

Havvy:
The documentation you provided is written in a more guide-like style than (to quote Centril on Discord) a "definition-style spec text", but the information is there. Do *you* want to take your hand at improving that section? You definitely know more about the link attribute than I do.

ralfbiedert:
Sure, I can try a PR.

rossy62:
Is there any information on *why* Rust produces mylib.dll then mylib.dll.lib and mylib.dll.exp in windows instead of mylib.exp and mylib.exp which is what the windows linker expects?  At the moment I'm renaming these outputs to mylib.exp and mylib.lib
[...]

PR #640, opened 2019-07-15
Better describe how #[link(name="")] works.

ralfbiedert:
Based on #638.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -145,8 +145,8 @@ not specified.
 
 <!-- ignore: requires extern linking -->
 ```rust,ignore
-// Links to `libcrypto.so` on Linux, `libcrypto.dylib` on Mac, and the
-// (`crypto.lib`, `crypto.dll`) pair on windows-msvc.
+// Links against `libcrypto.so` on Linux, `libcrypto.dylib` on Mac, and `crypto.lib`
+// on windows-msvc (resulting in a runtime dependency on `crypto.dll`).
 #[link(name = "crypto")]
 extern {
     // …
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -145,11 +145,17 @@ not specified.
 
 <!-- ignore: requires extern linking -->
 ```rust,ignore
-// Links to `libcrypto.so` on Linux and the (`foo.lib`, `foo.dll`) pair
-// on windows-msvc. Note that if you wanted to import a `cdylib` produced by Rust
-// on `x86_64-pc-windows-msvc` you would have to link against `name = "crypto.dll"`
-// as Rust produces a (`foo.dll.lib`, `foo.dll`) pair.
+// Links to `libcrypto.so` on Linux, `libcrypto.dylib` on Mac, and the
+// (`crypto.lib`, `crypto.dll`) pair on windows-msvc.
 #[link(name = "crypto")]
+extern {
+    // …
+}
+
+// Note that if you wanted to import a `cdylib` produced by Rust on
+// `x86_64-pc-windows-msvc` you would have to link against `crypto.dll`
+// as Rust produces a (`crypto.dll.lib`, `crypto.dll`) pair on that target.
+#[link(name = "crypto.dll")]
 extern {
     // …
 }
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -131,7 +131,7 @@ specifies the kind of library with the following possible values:
 - `framework` — Indicates a macOS framework. This is only valid for macOS
   targets.
 
-The `name` key must be included if `kind` is specified. Depending on the target `name`
+The `name` key must be included if `kind` is specified. Depending on the target, `name`
 is interpreted as follows:
 
 - `*-pc-windows-msvc` — Refers to the base name of the static library `name.lib`,
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -134,11 +134,9 @@ specifies the kind of library with the following possible values:
 The `name` key must be included if `kind` is specified. Depending on the target `name`
 is interpreted as follows:
 
-- `x86_64-pc-windows-msvc` — Refers to the base name of the `.lib` file (i.e., if you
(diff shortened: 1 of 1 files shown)


src/items/external-blocks.md
```diff
@@ -121,14 +121,23 @@ specifies the kind of library with the following possible values:
 - `framework` — Indicates a macOS framework. This is only valid for macOS
   targets.
 
-The `name` key must be included if `kind` is specified.
+The `name` key must be included if `kind` is specified. Depending on the target `name`
```
Havvy:
```suggestion
The `name` key must be included if `kind` is specified. Depending on the target, `name`
```

src/items/external-blocks.md
```diff
@@ -121,14 +121,23 @@ specifies the kind of library with the following possible values:
...
+- All other targets — Refers to the base name of the shared library.
 
 The `wasm_import_module` key may be used to specify the [WebAssembly module]
 name for the items within an `extern` block when importing symbols from the
 host environment. The default module name is `env` if `wasm_import_module` is
 not specified.
 
 ```rust,ignore
+// Links to `libcrypto.so` on Linux and the (`foo.lib`, `foo.dll`) pair
+// on windows-msvc. Note that if you wanted to import a `cdylib` produced by Rust
```
Havvy: The Windows-MSVC part should be part of its own example. Perhaps a second extern block after the first.

src/items/external-blocks.md
```diff
@@ -121,14 +121,23 @@ specifies the kind of library with the following possible values:
...
+or the base name of the import library `name.lib` accompanying a dynamic-link library `name.dll`.
+- All other targets — Refers to the base name of the shared library.
 
 The `wasm_import_module` key may be used to specify the [WebAssembly module]
 name for the items within an `extern` block when importing symbols from the
 host environment. The default module name is `env` if `wasm_import_module` is
 not specified.
 
 ```rust,ignore
+// Links to `libcrypto.so` on Linux and the (`foo.lib`, `foo.dll`) pair
```
Havvy: `foo.lib` should be `crypto.lib`? Same with the `foo.dll`?

ralfbiedert: Yes ... I could swear I did this ...

src/items/external-blocks.md
ralfbiedert: Done.

src/items/external-blocks.md
```diff
@@ -121,19 +121,34 @@ specifies the kind of library with the following possible values:
...
+- All other targets — Refers to the base name of the shared library.
 
 The `wasm_import_module` key may be used to specify the [WebAssembly module]
 name for the items within an `extern` block when importing symbols from the
 host environment. The default module name is `env` if `wasm_import_module` is
 not specified.
 
 ```rust,ignore
+// Links to `libcrypto.so` on Linux, `libcrypto.dylib` on Mac, and the
+// (`crypto.lib`, `crypto.dll`) pair on windows-msvc.
```
retep998: Rust isn't linking to `crypto.dll` at all on `windows-msvc`. The linker never looks for that file. Linking is strictly done through the `.lib` where some symbols resolve to a runtime dependency on a `.dll`.

`windows-gnu` is the odd one out as the MinGW linker is able to link against both import libraries and dlls directly so `-lfoo` looks for both `libfoo.a` and `foo.dll`. Of course if you're using the LLVM linker for `windows-gnu` then you're back to only linking against import libraries, so you're better off just always linking against the import library on windows except for the weird fact that Rust chooses not to emit import libraries for `cdylib`s on `windows-gnu` targets...

ralfbiedert: Sure. What text would you propose instead?

Centril: cc @retep998 -- any thought on ^--- (in general, you're the expert on this issue, and we, the reference maintainers are... not. ^^)

retep998: Better phrasing might be "links against `crypto.lib` resulting in a runtime dependency on `crypto.dll`"

ralfbiedert: Updated PR. Slightly changed your wording to fit into the sentence.

Centril:
@ralfbiedert Can you rebase the PR?
And @retep998 can you take another look?

src/items/external-blocks.md
```diff
@@ -131,7 +131,12 @@ specifies the kind of library with the following possible values:
 - `framework` — Indicates a macOS framework. This is only valid for macOS
   targets.
 
-The `name` key must be included if `kind` is specified.
+The `name` key must be included if `kind` is specified. Depending on the target, `name`
+is interpreted as follows:
+
+- `*-pc-windows-msvc` — Refers to the base name of the static library `name.lib`,
+or the base name of the import library `name.lib` accompanying a dynamic-link library `name.dll`.
+- All other targets — Refers to the base name of the shared library.
```
ehuss, line 139: I had a bit of a hard time following this. The term "base name" isn't defined, and doesn't match my normal understanding of "base name" when it comes to paths. It also seems a little vague (doesn't mention prefixes at all). 

Perhaps it could be worded something like this?

 The `name` is the library name without the platform's usual filename prefix
 and suffix.

> **Note:** For example, a link name of `foo` would search for a dynamic
> library named `libfoo.so` on Linux targets. On Windows MSVC targets it would
> search for the `foo.lib` import library.
>
> On Windows MSVC targets, to avoid ambiguity with staticlib filenames,
> `rustc` produces an import library named `foo.dll.lib` when generating a
> cdylib crate named `foo`. This means that if another crate wants to use that
> import library as-is, then the link name will need to be `foo.dll` to match
> that import library naming convention.

ralfbiedert: Thanks for the suggestion. I can change "base name" to something else. 

However, I don't quite understand the 2nd part of your comment, are you suggesting to add both quoted paragraphs as well?

ehuss: Yea, instead of using the list, just provide a side-note that provides some hints about platform-specific concerns.

PR #645, opened 2019-08-02
Fix a type in the modules section: functions => modules

mb64:
The last paragraph in the modules section talks about attributes on modules.  It then says, "The built-in attributes that have meaning on a function are ...", and lists attributes such as `#[path]` that work on modules.

This changes it to "The built-in attributes that have meaning on a module are ..."


ehuss:
Thanks!

PR #646, opened 2019-08-03
Add doc-alias documentation

GuillaumeGomez:
Part of https://github.com/rust-lang/rust/pull/63236

### DIFF
--- a/src/attributes/doc-alias.md
+++ b/src/attributes/doc-alias.md
@@ -1,0 +1,23 @@
+# `doc_alias`
+
+The tracking issue for this feature is: [#50146]
+
+[#50146]: https://github.com/rust-lang/rust/issues/50146
+
+------------------------
+
+You can add alias(es) to an item when using the `rustdoc` search through the
+`doc(alias)` attribute. Example:
+
+```rust,no_run
+#![feature(doc_alias)]
+
+#[doc(alias = "x")]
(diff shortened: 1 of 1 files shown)


Centril:
Why is this being added here?

It's both unstable and AFAIK not part of Rust the language. The only part that is part of the language is that the compiler is aware that a `doc` attribute exists and not what form `$tts` may take in `#[doc($tts)]`.

GuillaumeGomez:
It's part of the stabilization process as far as I can tell. Also, it's an attribute so it seems to be at its place or did I miss something?

Centril:
The reference is meant to specify Rust the language, not rustdoc specific things. I would prefer to just mention that `#[doc]` exists and then link to some other source, e.g. rustdoc specific documentation.

cc @ehuss

ehuss:
Yea, we only briefly mention the `doc` attribute in https://doc.rust-lang.org/nightly/reference/attributes.html and defer to https://doc.rust-lang.org/rustdoc/the-doc-attribute.html for the actual documentation.  It looks like your rustdoc PR already updates that page, so I would say this is not necessary.

GuillaumeGomez:
Fine by me. Just close the PR if it's your final decision. :smiley:

Centril:
Alright; thanks!

PR #647, opened 2019-08-05
Define sound and unsound

RalfJung:
Cc https://github.com/rust-lang/unsafe-code-guidelines/pull/190

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -1,9 +1,14 @@
 ## Behavior considered undefined
 
-Rust code, including within `unsafe` blocks and `unsafe` functions is incorrect
-if it exhibits any of the behaviors in the following list. It is the
-programmer's responsibility when writing `unsafe` code that it is not possible
-to let `safe` code exhibit these behaviors.
+Rust code is incorrect if it exhibits any of the behaviors in the following
+list. This includes code within `unsafe` blocks and `unsafe` functions!
+`unsafe` is not an excuse to cause undefined behavior.
+
+It is the programmer's responsibility when writing `unsafe` code that it is not
+possible to let `safe` code exhibit these behaviors. `unsafe` code that
+satisfies this property for any `safe` client is called *sound*; if `unsafe`
+code can be misused by `safe` code to exhibit undefined behavior, it is
+*unsound*.
 
 <div class="warning">
 
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -1,13 +1,13 @@
 ## Behavior considered undefined
 
 Rust code is incorrect if it exhibits any of the behaviors in the following
-list. This includes code within `unsafe` blocks and `unsafe` functions!
+list. This includes code within `unsafe` blocks and `unsafe` functions.
 `unsafe` is not an excuse to cause undefined behavior.
 
-It is the programmer's responsibility when writing `unsafe` code that it is not
+It is the programmer's responsibility when writing `unsafe` code to ensure that it is not
 possible to let `safe` code exhibit these behaviors. `unsafe` code that
-satisfies this property for any `safe` client is called *sound*; if `unsafe`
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
 ## Behavior considered undefined
 
-Rust code, including within `unsafe` blocks and `unsafe` functions is incorrect
-if it exhibits any of the behaviors in the following list. It is the
-programmer's responsibility when writing `unsafe` code that it is not possible
-to let `safe` code exhibit these behaviors.
+Rust code is incorrect if it exhibits any of the behaviors in the following
+list. This includes code within `unsafe` blocks and `unsafe` functions!
```
Centril:
```suggestion
list. This includes code within `unsafe` blocks and `unsafe` functions.
```

src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
...
+`unsafe` is not an excuse to cause undefined behavior.
```
Centril:
```suggestion
Use of `unsafe` should never result in undefined behavior.
```

src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
...
+
+It is the programmer's responsibility when writing `unsafe` code that it is not
```
Centril:
```suggestion
It is the programmer's responsibility when writing `unsafe` code to ensure that it is not
```

src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
...
+possible to let `safe` code exhibit these behaviors. `unsafe` code that
```
Centril:
```suggestion
possible for safe code exhibit these behaviors for any input, return value, or program state. `unsafe` code that
```
- there's no `safe` in code

src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
...
+satisfies this property for any `safe` client is called *sound*; if `unsafe`
```
Centril:
```suggestion
satisfies this property for any safe client is called *sound*; if `unsafe`
```
Do we also need to define "client"?

src/behavior-considered-undefined.md
```diff
@@ -1,9 +1,14 @@
...
+code can be misused by `safe` code to exhibit undefined behavior, it is
```
Centril:
```suggestion
code can be misused by safe code to exhibit undefined behavior, it is
```

src/behavior-considered-undefined.md
RalfJung: "should" is way too weak IMO. If you don't like "excuse", what about

> Even with `unsafe`, Rust code must never cause undefined behavior.

I also considered something like

> `unsafe` only means that avoiding undefined behavior is on the programmer; it does not change anything about the fact that Rust programs must never cause undefined behavior.

src/behavior-considered-undefined.md
RalfJung: True, I just stuck to existing terminology there.

The enumeration "input, return value or program state" is very odd though. "For any safe code interacting with it" or so would be better IMO?

src/behavior-considered-undefined.md
Centril: I like the second version  better although "must" is definitely better than "should", I agree.

src/behavior-considered-undefined.md
Centril: The main thing I wanted to get at is that e.g.
```rust
fn foo(x: u8) {
    if x < 5 { unsafe { trigger_ub(); } }
}
```
is not sound because it can cause UB for some inputs.

That is, I wanted to point at the "universal quantifier" here.

> "For any safe code interacting with it" or so would be better IMO?

Yeah I suppose that is mostly the same thing. I think that's good.

RalfJung:
@Centril okay I think I resolved your concerns.

PR #648, opened 2019-08-06
Update partially initialized values in drop documentation.

ehuss:
As part of rust-lang/rust#63230, it does not appear to be possible to create a partially initialized value.

cc @RalfJung @tmandry  Please correct me if I'm wrong, I am unfamiliar with this. I was planning to do an update today, but can't with breakage.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -63,11 +63,15 @@ loop {
     moved = ShowOnDrop("Drops when moved");
     // drops now, but is then uninitialized
     moved;
+
     // Uninitialized does not drop.
     let uninitialized: ShowOnDrop;
-    // Only first element drops
-    let mut partially_initialized = (ShowOnDrop("one"), ShowOnDrop("two"));
-    core::mem::forget(partially_initialized.1);
+
+    // After a partial move, only the remaining fields are dropped.
+    let mut partial_move = (ShowOnDrop("first"), ShowOnDrop("forgotten"));
+    // Perform a partial move, leaving only `partial_move.0` initialized.
+    core::mem::forget(partial_move.1);
+    // When partial_move's scope ends, only the first field is dropped.
 }
 ```
 
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -63,10 +63,11 @@ loop {
     moved = ShowOnDrop("Drops when moved");
     // drops now, but is then uninitialized
     moved;
+    // Uninitialized does not drop.
     let uninitialized: ShowOnDrop;
     // Only first element drops
-    let mut partially_initialized: (ShowOnDrop, ShowOnDrop);
-    partially_initialized.0 = ShowOnDrop("Partial tuple first");
+    let mut partially_initialized = (ShowOnDrop("one"), ShowOnDrop("two"));
+    core::mem::forget(partially_initialized.1);
 }
 ```
 


src/destructors.md
```diff
@@ -2,9 +2,7 @@
 
 When an [initialized]&#32;[variable] in Rust goes out of scope or a [temporary]
 is no longer needed its _destructor_ is run. [Assignment] also runs the
-destructor of its left-hand operand, unless it's an uninitialized variable. If a
-[struct] variable has been partially initialized, only its initialized fields
-are dropped.
```
Centril: We can still do:
```rust
struct ND(u8);

impl Drop for ND {
    fn drop(&mut self) {
        dbg!(self.0);
    }
}

struct Prod(ND, ND, ND);

fn main() {
    let mut prod = Prod(ND(0), ND(1), ND(2));
    core::mem::forget(prod.0);
    // prod.0 is not initialized here.
}
```

===>

```rust
[src/main.rs:5] self.0 = 1
[src/main.rs:5] self.0 = 2
```

ehuss: OK.  So how about just updating the example to do that? I just pushed a change, with the wording from #514.

ehuss: Or is a "partial move" different from a partially initialized value?

Centril: It seems to me that "partial move" is something that happens and "partially initialized" is a state of being that can result from a "partial move".

cc @RalfJung @matthewjasper

RalfJung: The corresponding adjective would be "partially moved out of", I guess.

But as a starter, I don't see how removing `struct` here helps with anything. If we (re)spin "partially initialized" to mean "partially moved out of", this is still about structs (and tuples), like before.

However, I think it would be better to change the wording. or do we consistently call moved-out-of variables "uninitialized"?

Centril: > However, I think it would be better to change the wording. or do we consistently call moved-out-of variables "uninitialized"?

I'm not sure. I'm not too keen on a particular wording as long as the content is there (which it is).

ehuss: I only removed the word "struct" to avoid confusion (and that's what #514 did), since it could be a struct, enum, tuple, or whatever.  Since the example uses a tuple, it seemed better to not be so specific.

I think it would be good to be clearer about this, but I also think #514 would be the place to discuss and debate the specifics. I'd like to keep this PR focused on the minimum to get the tests to pass.  Does the new example make sense? Should it use a different wording? Or should it just be removed?

RalfJung: > I'd like to keep this PR focused on the minimum to get the tests to pass.

Then please don't make unrelated wording changes. :)

src/destructors.md
```diff
@@ -63,10 +63,11 @@ loop {
     moved = ShowOnDrop("Drops when moved");
     // drops now, but is then uninitialized
     moved;
+    // Uninitialized does not drop.
     let uninitialized: ShowOnDrop;
     // Only first element drops
-    let mut partially_initialized: (ShowOnDrop, ShowOnDrop);
-    partially_initialized.0 = ShowOnDrop("Partial tuple first");
+    let mut partially_initialized = (ShowOnDrop("one"), ShowOnDrop("two"));
+    core::mem::forget(partially_initialized.1);
```
RalfJung: Just above there is a move example that actually drops early by moving, so maybe the same would make sense here?
```
// drops now, but second field (.1) is then uninitialized
partially_initialized.1;
// first field (.0) implicitly dropped.
```

src/destructors.md
ehuss: Fair, I removed it.

src/destructors.md
Centril: Yeah that's probably better stylistically.

ehuss: I like the explicitness of calling `forget`, since if you run the example you can actually see how it only drops the remaining fields.  What about the following text to be more explicit (and to match the wording in rustc):

```rust
    // After a partial move, only the remaining fields are dropped.
    let mut partial_move = (ShowOnDrop("first"), ShowOnDrop("forgotten"));
    // Perform a partial move, leaving `partial_move.0` initialized.
    core::mem::forget(partial_move.1);
    // When `partial_move` scope ends, only the first field is dropped.
```

Centril: Some wording nits: "leaving" -> "leaving only"

"`partial_move`" -> "`partial_move`'s"

Aside from that I think it is good.
Imo we can land this and iterate on style :)

PR #650, opened 2019-08-12
our closure syntax comes from Ruby

RalfJung:
At least I think it does? @steveklabnik can probably confirm.

### DIFF
--- a/src/influences.md
+++ b/src/influences.md
@@ -16,7 +16,7 @@ that have since been removed):
 * Swift: optional bindings
 * Scheme: hygienic macros
 * C#: attributes
-* Ruby: <strike>block syntax</strike>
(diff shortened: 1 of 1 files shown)


steveklabnik:
It was created before I started, but smalltalk/ruby are clear influences, yeah:

```
// Rust, redundant {}s added for clarity
let sum = |a, b| { a + b };

// Ruby
sum = {|a, b| a + b };

// Smalltalk
sum := [ :a :b | a + b ]
```

RalfJung:
Well I mostly wrote this because (a) I wondered before "why this strange closure syntax in Rust", and (b) Ruby looked a little sad in that list with its only contributed stroked through. ;)

PR #652, opened 2019-08-14
Correct errors in the reference of extern functions definitions and declarations

gnzlbg:
The current reference of extern functions definitions is incorrect. It doesn't take into account that all "normal" Rust functions are `extern "Rust" fn`.

The section about external blocks specifies that only "non-Rust functions" (whatever that means) can be declared in external blocks. This is incorrect, since both an `extern "C"` Rust function, and a function with the `"Rust"` ABI, can be declared in external blocks.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -29,8 +29,12 @@
 > _NamedFunctionParametersWithVariadics_ :\
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
-External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external libraries.
+External blocks provide _declarations_ of items that are not _defined_ in the
+current crate and are the basis of Rust's foreign function interface. Using
+items declared in external blocks is `unsafe`.
+
+Two kind of item _declarations_ are allowed in external blocks:
+[functions](function.md) and [statics](static-items.md).
 
 Functions within external blocks are declared in the same way as other Rust
 functions, with the exception that they may not have a body and are instead
@@ -47,6 +51,8 @@ extern "abi" for<'l1, ..., 'lm> fn(A1, ..., An) -> R`, where `'l1`, ... `'lm`
 are its lifetime parameters, `A1`, ..., `An` are the declared types of its
 parameters and `R` is the declared return type.
 
+Statics within external blocks are declared in the same way as other Rust statics,
+with the exception that they may not have an expression initializing their value.
 It is `unsafe` to access a static item declared in an extern block, whether or
 not it's mutable.
 
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -30,11 +30,12 @@
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
 External blocks provide _declarations_ of items that are not _defined_ in the
-current crate and are the basis of Rust's foreign function interface. Using
-items declared in external blocks is only allowed in an `unsafe` context.
+current crate and are the basis of Rust's foreign function interface. These are
+sort of like unchecked imports. 
 
-Two kind of item _declarations_ are allowed in external blocks:
-[functions] and [statics].
+Two kind of item _declarations_ are allowed in external blocks: [functions] and
+[statics]. Calling functions or accessing statics that are declared in external
+blocks is only allowed in an `unsafe` context.
 
 Functions within external blocks are declared in the same way as other Rust
 functions, with the exception that they may not have a body and are instead
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -90,13 +90,6 @@ There are also some platform-specific ABI strings:
 * `extern "vectorcall"` -- The `vectorcall` ABI -- corresponds to MSVC's
   `__vectorcall` and clang's `__attribute__((vectorcall))`
 
-Finally, there are some unstable rustc-specific ABI strings:
-
-* `extern "rust-intrinsic"` -- The ABI of rustc intrinsics, equivalent to `"Rust"`.
-* `extern "rust-call"` -- The ABI of the Fn::call trait functions, equivalent to `"Rust"`.
-* `extern "platform-intrinsic"` -- The ABI of Specific platform intrinsics, equivalent to `"Rust"` -- like, for
-  example, `sqrt` -- equivalent to `"Rust"`. You should never have to deal with it.
-
 ## Variadic functions
 
 Functions within external blocks may be variadic by specifying `...` after one
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -31,7 +31,7 @@
 
 External blocks provide _declarations_ of items that are not _defined_ in the
 current crate and are the basis of Rust's foreign function interface. Using
-items declared in external blocks is `unsafe`.
+items declared in external blocks is only allowed in an `unsafe` context.
 
 Two kind of item _declarations_ are allowed in external blocks:
 [functions] and [statics].
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -30,8 +30,7 @@
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
 External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external, non-Rust
-libraries.
+Declarations in an external block describe symbols in external libraries.
 
 Functions within external blocks are declared in the same way as other Rust
 functions, with the exception that they may not have a body and are instead
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -31,7 +31,7 @@
 
 External blocks provide _declarations_ of items that are not _defined_ in the
 current crate and are the basis of Rust's foreign function interface. These are
-sort of like unchecked imports. 
+akin to unchecked imports. 
 
 Two kind of item _declarations_ are allowed in external blocks: [functions] and
 [statics]. Calling functions or accessing statics that are declared in external
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -85,12 +85,12 @@ There are also some platform-specific ABI strings:
 * `extern "vectorcall"` -- The `vectorcall` ABI -- corresponds to MSVC's
   `__vectorcall` and clang's `__attribute__((vectorcall))`
 
-Finally, there are some rustc-specific ABI strings:
+Finally, there are some unstable rustc-specific ABI strings:
 
-* `extern "rust-intrinsic"` -- The ABI of rustc intrinsics.
-* `extern "rust-call"` -- The ABI of the Fn::call trait functions.
-* `extern "platform-intrinsic"` -- Specific platform intrinsics -- like, for
-  example, `sqrt` -- have this ABI. You should never have to deal with it.
+* `extern "rust-intrinsic"` -- The ABI of rustc intrinsics, equivalent to `"Rust"`.
+* `extern "rust-call"` -- The ABI of the Fn::call trait functions, equivalent to `"Rust"`.
+* `extern "platform-intrinsic"` -- The ABI of Specific platform intrinsics, equivalent to `"Rust"` -- like, for
+  example, `sqrt` -- equivalent to `"Rust"`. You should never have to deal with it.
 
 ## Variadic functions
 
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -53,7 +53,7 @@ are its lifetime parameters, `A1`, ..., `An` are the declared types of its
 parameters and `R` is the declared return type.
 
 Statics within external blocks are declared in the same way as statics outside of external blocks,
-with the exception that they may not have an expression initializing their value.
+except that they do not have an expression initializing their value.
 It is `unsafe` to access a static item declared in an extern block, whether or
 not it's mutable.
 
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -52,7 +52,7 @@ extern "abi" for<'l1, ..., 'lm> fn(A1, ..., An) -> R`, where `'l1`, ... `'lm`
 are its lifetime parameters, `A1`, ..., `An` are the declared types of its
 parameters and `R` is the declared return type.
 
-Statics within external blocks are declared in the same way as other Rust statics,
+Statics within external blocks are declared in the same way as statics outside of external blocks,
 with the exception that they may not have an expression initializing their value.
 It is `unsafe` to access a static item declared in an extern block, whether or
 not it's mutable.
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -34,7 +34,7 @@ current crate and are the basis of Rust's foreign function interface. Using
 items declared in external blocks is `unsafe`.
 
 Two kind of item _declarations_ are allowed in external blocks:
-[functions](function.md) and [statics](static-items.md).
+[functions] and [statics].
 
 Functions within external blocks are declared in the same way as other Rust
 functions, with the exception that they may not have a body and are instead
@@ -170,6 +170,8 @@ extern {
 
 [IDENTIFIER]: ../identifiers.md
 [WebAssembly module]: https://webassembly.github.io/spec/core/syntax/modules.html
+[functions]: functions.md
+[statics]: static-items.md
 [_Abi_]: functions.md
 [_FunctionReturnType_]: functions.md
 [_Generics_]: generics.md
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -107,21 +107,28 @@ component after the function name. This might be necessary if there is not
 sufficient context to determine the type parameters. For example,
 `mem::size_of::<u32>() == 4`.
 
-## Extern functions
+## Extern function qualifier
 
-Extern function _definitions_ allow defining functions that can be called 
-with a particular ABI:
+The `extern` function qualifier allows providing function _definitions_ that can
+be called with a particular ABI:
 
 ```rust,ignore
 extern "ABI" fn foo() { ... }
 ```
 
-An extern function _declaration_ via an [external block] can be used to
-provide an item for these functions that can be called by Rust code without
-providing their definition. 
+These are often used in combination with [external block] items which provide
+function _declarations_ that can be used to call functions without providing
+their _definition_:
 
-When `"extern" Abi?*` is omitted from `FunctionQualifiers`, the ABI `"Rust"` is
-assigned. For example: 
+```rust,ignore
+extern "ABI" {
+  fn foo(); /* no body */
+}
+unsafe { foo() }
+```
+
+When `"extern" Abi?*` is omitted from `FunctionQualifiers` in function items,
+the ABI `"Rust"` is assigned. For example:
 
 ```rust
 fn foo() {}
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -112,7 +112,7 @@ sufficient context to determine the type parameters. For example,
 Extern function _definitions_ allow defining functions that can be called 
 with a particular ABI:
 
-```rust,no_run
+```rust,ignore
 extern "ABI" fn foo() { ... }
 ```
 
@@ -173,8 +173,9 @@ equivalent to the `"Rust"` ABI, e.g.,
 [`"rust-intrinsic"`](https://doc.rust-lang.org/unstable-book/language-features/intrinsics.html?highlight=rust-intrin#intrinsics)
 or
 [`"platform-intrinsic"`](https://doc.rust-lang.org/unstable-book/language-features/platform-intrinsics.html).
-Refer to the [Unstable Book](https://doc.rust-lang.org/unstable-book) for more
-information about these.
+For more information about these refer to the [ABI section of external block
+items][external-blocks.md#ABI] and the [Unstable
+Book](https://doc.rust-lang.org/unstable-book).
 
 ## Const functions
 
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -176,13 +176,10 @@ with such ABIs causes the process to abort.
 aborts the process by executing an illegal instruction.
 
 **Non-normative note**: There are other ABIs available in unstable Rust that are
-equivalent to the `"Rust"` ABI, e.g.,
-[`"rust-intrinsic"`](https://doc.rust-lang.org/unstable-book/language-features/intrinsics.html?highlight=rust-intrin#intrinsics)
-or
-[`"platform-intrinsic"`](https://doc.rust-lang.org/unstable-book/language-features/platform-intrinsics.html).
-For more information about these refer to the [ABI section of external block
-items][external-blocks.md#ABI] and the [Unstable
-Book](https://doc.rust-lang.org/unstable-book).
+equivalent to the `"Rust"` ABI, e.g., [`"rust-intrinsic"`][rust_intrinsic] or
+[`"platform-intrinsic"`][platform_intrinsic]. For more information about these
+refer to the [ABI section of external block items][external_block_abi] and the
+[Unstable Book].
 
 ## Const functions
 
@@ -290,3 +287,7 @@ attributes macros.
 [`export_name`]: ../abi.md#the-export_name-attribute
 [`link_section`]: ../abi.md#the-link_section-attribute
 [`no_mangle`]: ../abi.md#the-no_mangle-attribute
+[external_block_abi]: external-blocks.md#abi
+[Unstable Book]: https://doc.rust-lang.org/unstable-book
+[rust_intrinsic]: https://doc.rust-lang.org/unstable-book/language-features/intrinsics.html?highlight=rust-intrin#intrinsics
+[platform_intrinsic]: https://doc.rust-lang.org/unstable-book/language-features/platform-intrinsics.html
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -112,7 +112,7 @@ sufficient context to determine the type parameters. For example,
 Extern function _definitions_ allow defining functions that can be called 
 with a particular ABI:
 
-```rust,norun
+```rust,no_run
 extern "ABI" fn foo() { ... }
 ```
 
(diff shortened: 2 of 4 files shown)


src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
+abbreviate the type of Rust functions like `foo` as `fn()`, this is actually a 
+synonym for `extern "Rust" fn()`. That is, this:
 
 ```rust
-// Declares an extern fn, the ABI defaults to "C"
-extern fn new_i32() -> i32 { 0 }
+fn foo() {}
+```
+
+is identical to
```
Centril:
```suggestion
is equivalent to:
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
-same type as the functions declared in an extern block.
+Just as with [external block], when the `extern` keyword is used and the `"ABI` 
+is omitted, the ABI used defaults to `"C"`. That is, this
+
+```rust
+extern fn new_i32() -> i32 { 0 }
+let fptr: extern fn() -> i32 = new_i32;
+```
+
+is identical to
```
Centril:
```suggestion
is equivalent to:
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
+
+// Declares a function with the "stdcall" ABI
 # #[cfg(target_arch = "x86_64")]
 extern "stdcall" fn new_i32_stdcall() -> i32 { 0 }
 ```
 
-Unlike normal functions, extern fns have type `extern "ABI" fn()`. This is the
-same type as the functions declared in an extern block.
+Just as with [external block], when the `extern` keyword is used and the `"ABI` 
+is omitted, the ABI used defaults to `"C"`. That is, this
```
Centril:
```suggestion
is omitted, the ABI used defaults to `"C"`. That is, this:
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
 
 ```rust
-# extern fn new_i32() -> i32 { 0 }
+extern "C" fn new_i32() -> i32 { 0 }
 let fptr: extern "C" fn() -> i32 = new_i32;
 ```
 
-As non-Rust calling conventions do not support unwinding, unwinding past the end
-of an extern function will cause the process to abort. In LLVM, this is
+Since functions with an ABI that differs from `"Rust"` do not support
```
Centril:
```suggestion
Functions with an ABI that differs from `"Rust"` do not support 
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
 ```rust
-# extern fn new_i32() -> i32 { 0 }
+extern "C" fn new_i32() -> i32 { 0 }
 let fptr: extern "C" fn() -> i32 = new_i32;
 ```
 
-As non-Rust calling conventions do not support unwinding, unwinding past the end
-of an extern function will cause the process to abort. In LLVM, this is
+Since functions with an ABI that differs from `"Rust"` do not support 
+unwinding in the exact same way that Rust does, unwinding past the end
```
Centril:
```suggestion
unwinding in the exact same way that Rust does. Therefore, unwinding past the end 
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
-# extern fn new_i32() -> i32 { 0 }
+extern "C" fn new_i32() -> i32 { 0 }
 let fptr: extern "C" fn() -> i32 = new_i32;
 ```
 
-As non-Rust calling conventions do not support unwinding, unwinding past the end
-of an extern function will cause the process to abort. In LLVM, this is
+Since functions with an ABI that differs from `"Rust"` do not support 
+unwinding in the exact same way that Rust does, unwinding past the end 
+of functions with such ABIs causes the process to abort. In LLVM, this is
```
Centril: What LLVM does is not normative and should at most be left as a note.

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
 implemented by executing an illegal instruction.
 
+Some ABIs that are identical to `"Rust"` are:
```
Centril: Remove this bit; these are all unstable.

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
-opposite functionality to [external blocks]. Whereas external
-blocks allow Rust code to call foreign code, extern functions with bodies
-defined in Rust code _can be called by foreign code_. They are defined in the
-same way as any other Rust function, except that they have the `extern`
-qualifier.
+Extern function _definitions_ allow defining functions that can be called 
+with a particular ABI:
+
+```rust,norun
+extern "ABI" fn foo() { ... }
```
ehuss: There's a CI error here.  I would just show a simple example like:

```suggestion
extern "C" fn foo() { /*...*/ }
```

src/items/functions.md
```diff
@@ -109,34 +109,70 @@ sufficient context to determine the type parameters. For example,
...
-Extern functions are part of Rust's foreign function interface, providing the
-opposite functionality to [external blocks]. Whereas external
-blocks allow Rust code to call foreign code, extern functions with bodies
-defined in Rust code _can be called by foreign code_. They are defined in the
-same way as any other Rust function, except that they have the `extern`
-qualifier.
+Extern function _definitions_ allow defining functions that can be called 
+with a particular ABI:
+
+```rust,norun
```
ehuss: `norun` is not a valid attribute, it is `no_run`. But it should not be necessary if it is made into valid syntax.

src/items/functions.md
gnzlbg: @ehuss I think using `"ABI"` here is better for exposition purposes. I tried to avoid running this with `norun`, but that did not work out. I'll fix that.

src/items/functions.md
gnzlbg: @Centril where would be the right place to document that these are all equivalent to the `"Rust"` ABI ? The "unstable" Reference? Maybe I can add that there. 

I don't know if it would be worth it to hyperlink that from here, or to just have a small mention here that there are some unstable ABIs that are also equivalent to `"Rust"`. 

Thoughts?

src/items/functions.md
gnzlbg: This should be fixed in the latest commit, so I'm resolving this issue. 

Please feel free to re-open if you disagree with using `"ABI"` for exposition purposes.

src/items/functions.md
gnzlbg: @Centril I've added a "Non-normative note" that mentions that there are some unstable ABIs that are equivalent to `"Rust"`, and that refers readers to the ABI section of the reference - which for some reason mentions these ABIs as well without saying that these are unstable - and the "Unstable Book".

gnzlbg:
@eddyb would you be able to review these changes as well ? Particularly, I've added that some ABIs are equivalent to `"Rust"` and documented this for some of them. Would like to triple-check that this is indeed actually the case.

src/items/functions.md
gnzlbg: Moved this to a non-normative note.

gnzlbg:
@Centril I also have the feeling that talking about function definitions/declarations is maybe too "C/C++ speak". 

Is there a better and clearer way to talk about these in "Rust speak" ? The only thing that comes to mind is to consistently use "function item" for "definitions" (`fn foo() {}`), "external function item" for "declarations" (`extern { fn foo(); }`), and "extern-qualified function item" or "function item with an extern qualifier" for a "definition" that uses the extern qualifier (`extern fn foo() {}`). However, I feel that it is too easy to confuse "external function item" with "extern-qualified function item" when discussing this part of the language.

src/items/functions.md
Havvy: We want to make zero reference to unstable parts of the language. Not even notes.

Those ABIs should be documented in the rustc book.

src/items/external-blocks.md
```diff
@@ -29,9 +29,12 @@
 > _NamedFunctionParametersWithVariadics_ :\
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
-External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external, non-Rust
-libraries.
+External blocks provide _declarations_ of items that are not _defined_ in the
+current crate and are the basis of Rust's foreign function interface. Using
+items declared in external blocks is `unsafe`.
```
Centril:
```suggestion
items declared in external blocks is only allowed in an `unsafe` context.
```

src/items/external-blocks.md
```diff
@@ -29,9 +29,12 @@
 > _NamedFunctionParametersWithVariadics_ :\
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
-External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external, non-Rust
-libraries.
+External blocks provide _declarations_ of items that are not _defined_ in the
+current crate and are the basis of Rust's foreign function interface. Using
```
Centril: "Using" is ambiguous; for example:
```rust
extern "C" {
    fn foo();
}

fn bar() { let _ = foo; }
```
does compile just fine and is one form of "using".

src/items/external-blocks.md
```diff
@@ -85,12 +90,12 @@ There are also some platform-specific ABI strings:
 * `extern "vectorcall"` -- The `vectorcall` ABI -- corresponds to MSVC's
   `__vectorcall` and clang's `__attribute__((vectorcall))`
 
-Finally, there are some rustc-specific ABI strings:
+Finally, there are some unstable rustc-specific ABI strings:
```
Centril:
```suggestion
Finally, there are some unstable `rustc`-specific ABI strings:
```

src/items/functions.md
```diff
@@ -107,35 +107,79 @@ component after the function name. This might be necessary if there is not
...
 ```
 
-As non-Rust calling conventions do not support unwinding, unwinding past the end
-of an extern function will cause the process to abort. In LLVM, this is
-implemented by executing an illegal instruction.
+Functions with an ABI that differs from `"Rust"` do not support unwinding in the
+exact same way that Rust does. Therefore, unwinding past the end of functions
+with such ABIs causes the process to abort.
+
+**Non-normative note**: The LLVM backend of the current Rust implementation
```
Centril:
```suggestion
> **Note**: The LLVM backend of the `rustc` implementation
```

Centril:
> @Centril I also have the feeling that talking about function definitions/declarations is maybe too "C/C++ speak".

See https://doc.rust-lang.org/nightly/reference/items/traits.html#traits
You'll note:
```
TraitFunc :
      TraitFunctionDecl ( ; | BlockExpression )
```

I think it's fine to use "declaration" for function signatures and definitions for declarations enhanced with a block expression.

src/items/functions.md
gnzlbg: Shall I remove mentions to these both from this section and the extern block section then?

src/items/external-blocks.md
gnzlbg: Nice catch. I would suggest replacing "using" with "calling" for functions and "accessing" for statics, but I'm not sure if those are comprehensive enough.

src/items/functions.md
Centril: Seems like a good idea yeah.

src/items/external-blocks.md
Centril: I agree that "calling" and "accessing" are better. :)

gnzlbg:
> I think it's fine to use "declaration" for function signatures and definitions for declarations enhanced with a block expression.

Makes sense, I'll keep the text that uses definition/declarations then. 

Maybe we could open an issue about updating the extern-block section to use "_ExternalFunctionDeclaration_" instead of "_ExternalFunctionItem_". I am not sure if such a change makes sense, or whether there are other places where we might want to take this into account. Probably should be consistent with the grammar as well, although maybe the grammar calls these something different, no idea.

Centril:
Let's do that in a follow up if so; we'll need to make sure that we have no stale references to/in the grammar if we change.

src/items/external-blocks.md
```diff
@@ -85,13 +90,16 @@ There are also some platform-specific ABI strings:
 * `extern "vectorcall"` -- The `vectorcall` ABI -- corresponds to MSVC's
   `__vectorcall` and clang's `__attribute__((vectorcall))`
 
-Finally, there are some rustc-specific ABI strings:
+<<<<<<< HEAD
```
Centril: oops? ;)

src/items/external-blocks.md
eddyb: Note that for functions this can be explained by the type: they are `unsafe fn`s, whereas I think even `let _ = FFI_STATIC;` requires `unsafe`.

src/items/functions.md
```diff
@@ -107,35 +107,73 @@ component after the function name. This might be necessary if there is not
...
+## Extern function qualifier
 
-Extern functions are part of Rust's foreign function interface, providing the
-opposite functionality to [external blocks]. Whereas external
-blocks allow Rust code to call foreign code, extern functions with bodies
-defined in Rust code _can be called by foreign code_. They are defined in the
-same way as any other Rust function, except that they have the `extern`
-qualifier.
+The `extern` function qualifier allows providing function _definitions_ that can
+be called with a particular ABI:
```
eddyb: Hmm, this seems to imply it's specific to definitions but the exact same syntax with the same default also applies to `fn` pointers.

src/items/external-blocks.md
```diff
@@ -29,9 +29,12 @@
 > _NamedFunctionParametersWithVariadics_ :\
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
-External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external, non-Rust
-libraries.
+External blocks provide _declarations_ of items that are not _defined_ in the
```
eddyb: I think this could be more explicit that these are "imports" of a sort.

src/items/external-blocks.md
Centril: @eddyb `let _ = FFI_STATIC;` works without `unsafe`; https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5e582a3d7aa9dc49b3a250a6861a80b9

src/items/external-blocks.md
```diff
@@ -85,13 +90,16 @@ There are also some platform-specific ABI strings:
...
+=======
+Finally, there are some unstable `rustc`-specific ABI strings:
 
-* `extern "rust-intrinsic"` -- The ABI of rustc intrinsics.
-* `extern "rust-call"` -- The ABI of the Fn::call trait functions.
-* `extern "platform-intrinsic"` -- Specific platform intrinsics -- like, for
-  example, `sqrt` -- have this ABI. You should never have to deal with it.
+* `extern "rust-intrinsic"` -- The ABI of rustc intrinsics, equivalent to `"Rust"`.
+* `extern "rust-call"` -- The ABI of the Fn::call trait functions, equivalent to `"Rust"`.
+* `extern "platform-intrinsic"` -- The ABI of Specific platform intrinsics, equivalent to `"Rust"` -- like, for
```
eddyb: Not sure these should be documented, we've been meaning to remove all 3 for years...
cc @oli-obk at least for intrinsics we could maybe have an intermediary stage where we do `#[rustc_intrinsic] extern "Rust" {...}` instead of `extern "rust-intrinsic" {...}`?

I would maybe call them "pseudo-ABIs", in that they use the ABI string to indicate something special *not purporting to the ABI*, but are otherwise the "Rust" ABI.

Centril: They shouldn't; see https://github.com/rust-lang-nursery/reference/pull/652#discussion_r314060449.

For the question of moving towards `#[rustc_intrinsic]` please move this to a rust-lang/rust issue.

src/items/external-blocks.md
gnzlbg: That never happened :P 

I need to figure out a way to rebase the changes done to my branch via the suggestions on top of local modifications without "spurious" merge commits or these issues :/

src/items/external-blocks.md
gnzlbg: I've removed these :)

src/items/functions.md
gnzlbg: Not the exact same (e.g. function pointers have no identifier or body), but I see your point. Stretching this out a bit more, an also similar syntax applies to declarations in extern blocks. 

I think that the "extern block item" and the "function item" sections should just tangentially mention `extern "ABI"`, and that a different section describing the "extern qualifier" should be used to explain the whole thing. As it is done right now, we have links back and forth between both sections for no good reason.

src/items/functions.md
```diff
@@ -243,3 +281,7 @@ attributes macros.
 [`export_name`]: ../abi.md#the-export_name-attribute
 [`link_section`]: ../abi.md#the-link_section-attribute
 [`no_mangle`]: ../abi.md#the-no_mangle-attribute
+[external_block_abi]: external-blocks.md#abi
+[Unstable Book]: https://doc.rust-lang.org/unstable-book
+[rust_intrinsic]: https://doc.rust-lang.org/unstable-book/language-features/intrinsics.html?highlight=rust-intrin#intrinsics
```
Centril: Links need some garbage collection now.

src/items/external-blocks.md
gnzlbg: I've tried to clarify that.

src/items/external-blocks.md
gnzlbg: I've disambiguated that.

src/items/external-blocks.md
```diff
@@ -29,9 +29,13 @@
 > _NamedFunctionParametersWithVariadics_ :\
 > &nbsp;&nbsp; ( _NamedFunctionParam_ `,` )<sup>\*</sup> _NamedFunctionParam_ `,` `...`
 
-External blocks form the basis for Rust's foreign function interface.
-Declarations in an external block describe symbols in external, non-Rust
-libraries.
+External blocks provide _declarations_ of items that are not _defined_ in the
+current crate and are the basis of Rust's foreign function interface. These are
+sort of like unchecked imports.
```
Centril: Nit: "sort of like" reads rather like spoken English (not something you'd find in an IEEE paper...).

gnzlbg: I didn't intend to take @eddyb feedback literally here, but that's the best I could come up with. 

These blocks declare that something exists in a particular way. If you "use" (call, access, etc.) it and it doesn't exist, or it exists in a way that's different than how you declared it: boom. If you don't actually use it, whether it exists doesn't matter, and if it doesn't, that's ok.

OTOH, to me, an import is an operation that brings something into scope that exists. If it doesn't exist, then you can't import it. So I'm not sure what would be the best way to make an analogy with imports here.

Maybe something like

> Declarations are similar to imports, in that they bring an item into scope. The main difference is that an import requires an item to exist, while a declaration states that an item exists with a particular type signature, path, etc. As long as a declared function is not called or a declared static is not accessed, these declared items do not need to exist. This allows, for example, declaring a dynamically-linked function unconditionally, as long as it is only used when the function is actually dynamically linked. 

?

gnzlbg: Maybe with an example like this:

```rust
extern "C" { fn foo(); }

fn main() {
    let x = foo;
    if symbol_of_foo_not_linked() {
         link_symbol_of_foo();
    }
    unsafe { x() };
}
```

?

src/items/functions.md
```diff
@@ -148,30 +147,34 @@ extern "stdcall" fn new_i32_stdcall() -> i32 { 0 }
...
 
-Some ABIs that are identical to `"Rust"` are:
+**Non-normative note**: The LLVM backend of the current Rust implementation
+aborts the process by executing an illegal instruction.
 
-* `"rust-call"`
-* `"platform-intrinsic"`
-* `"rust-intrinsic"`
+**Non-normative note**: There are other ABIs available in unstable Rust that are
+equivalent to the `"Rust"` ABI, e.g.,
```
fstirlitz: Are they actually equivalent to `extern "Rust"`, though? I know for example that pointers to `extern "rust-intrinsic"` functions cannot be formed, since they have nowhere to point to.

gnzlbg: I see your point. The pointers can be formed (e.g. Rust programs that do this type check), but the compiler ICEs (this detects bugs in libcore, etc.). 

Maybe you are interested in [this comment](https://github.com/rust-lang-nursery/reference/pull/652#discussion_r314206468).

src/items/external-blocks.md
Centril: @eddyb @ehuss wdyt about the wording ^-- ?

eddyb: I use "FFI import" to mean something different from what the `use` item does in Rust, so I'm not comparing the two but rather using the more general meaning of "import".

Also, not sure that where the function comes from in terms of linkage is *that* important, compared to the fact that *it likely isn't even written in Rust*, hence "FFI".

src/items/functions.md
fstirlitz: I always assumed that because intrinsics are always inlined into the caller, they don't have their own distinct addresses and it's meaningless to speak of pointers to them. The compiler ICEs because it cannot generate code which forms such pointers, even when all types check out; in other words, it's not a deliberate feature (see also rust-lang/rust#15694). Am I wrong here?

gnzlbg: I'm not sure. I don't know what happens when you create a function pointer to, e.g., `llvm.assume` (does one get an LLVM error?), but for something like `core::intrinsic::sqrtf32` which emits `llvm.sqrt` creating a function pointer to it should work just fine. I'm not sure, but it might depend on the intrinsic.

src/items/external-blocks.md
```diff
@@ -48,6 +52,8 @@ extern "abi" for<'l1, ..., 'lm> fn(A1, ..., An) -> R`, where `'l1`, ... `'lm`
 are its lifetime parameters, `A1`, ..., `An` are the declared types of its
 parameters and `R` is the declared return type.
 
+Statics within external blocks are declared in the same way as other Rust statics,
```
Havvy:
```suggestion
Statics within external blocks are declared in the same way as statics outside of external blocks,
```

src/items/external-blocks.md
```diff
@@ -48,6 +52,8 @@ extern "abi" for<'l1, ..., 'lm> fn(A1, ..., An) -> R`, where `'l1`, ... `'lm`
...
+with the exception that they may not have an expression initializing their value.
```
Havvy:
```suggestion
except that they do not have an expression initializing their value.
```

src/items/functions.md
Havvy: IMO, punt this change to another PR. This one is getting a lot of back and forth, and I think we should setting with an improved base to work on.

That said, for that section, something like this could work:

```md
The `extern` qualifier specifies the [<abbr title="Application Binary Interface>ABI</abbr>][ABI] for calling the functions they apply to. It is applicable to ...

[ABI]: https://en.wikipedia.org/ABI
```

src/items/functions.md
```diff
@@ -107,35 +107,73 @@ component after the function name. This might be necessary if there is not
...
+```
+
+is equivalent to:
 
-// Declares an extern fn with "stdcall" ABI
+```rust
+extern "Rust" fn foo() {}
+```
+
+Functions in Rust can be called by foreign code, and using an ABI that
```
Havvy: This is a run-on sentence, or at least feels like one. Also, "Rust's functions". And what does "foreign code" here mean differently than code from other programming languages?

Centril: Rust code exposed with a C ABI.

gnzlbg:
cc @Gankra

src/items/functions.md
gnzlbg: So I'm going to leave this as is for now . Even if it still isn't 100% clear, it is still an improvement over what we have. I've opened #655 to track my suggestion for improving this further by just explaining the `extern` qualifier for all users somewhere else, and just referencing it from the function item and extern block item sections. 

If you have any better ideas about how to explain this more clearly, we can discuss those in that issue.

src/items/functions.md
gnzlbg: I'm not sure how to improve this.

Gankra:
Since most of the discussion is collapsed and there's 14 commits of history, it's hard for me to tell what happened here. 

I just want to get a clarification as to what has happened here with rust-call/rust-intrinsic/platform-intrinsic. They were never properly documented so it's unclear to be exactly what they implies (rust-intrinsic is only clearish one to me).

Are these things stricken from the record as pure perma-unstable implementation details that no one should ever have to think about? Is there another document somewhere that stdlib and rustc devs can refer to for the meaning of them? (these use cases after all *should* be the most active users of a reference)

In particular it *is* slightly interesting to me as to whether callers of rust-intrinsic and platform-intrinsic methods can rely on them never unwinding (I assume rust-call must in fact be allowed to unwind).

gnzlbg:
> I just want to get a clarification as to what has happened here with rust-call/rust-intrinsic/platform-intrinsic. [...] Are these things stricken from the record as pure perma-unstable implementation details that no one should ever have to think about?

IIUC, the reference is only intended (at least for the time being), to document stable Rust. All unstable Rust information should be documented in the Unstable Book. 

Now, the information that this PR removes is about perma-unstable features that are not intended to ever be stabilized. Most people, including most of those hacking in the standard library, shouldn't have to know about this. Knowledge about these is only required if you are touching standard library code that requires special compiler support. So the ideal place to document these would probably be the rustc book.

The information that this PR removes wasn't great either, and it is a bit out-of-date. For example,  most `"platform-intrinsic"` (thousands of them) were removed from rustc a while ago, and now AFAICT this ABI is only used for generic SIMD intrinsics... for no good reason. As @eddyb pointed out [here](https://github.com/rust-lang-nursery/reference/pull/652#discussion_r314206468), we probably want to just change these to be `extern "Rust"` and handle their "magicness" somehow else. 

> In particular it is slightly interesting to me as to whether callers of rust-intrinsic and platform-intrinsic methods can rely on them never unwinding 

They can't. These are essentially `extern "Rust"` which can unwind. Whether these unwind or not, would depend on the concrete intrinsic that we are dealing with (I don't have any concrete intrinsic in mind that actually can unwind, but this doesn't mean there isn't one).

Gankra:
I expect whatever primitive operation integer addition lowers to is a fine canonical example of an unwinding intrinsic.

gnzlbg:
@Havvy @Centril I'm uncertain of which changes are still required here.

src/items/external-blocks.md
Centril:
```suggestion
akin to unchecked imports. 
```
let's make it at least a bit less informal and then land this. :)

PR #654, opened 2019-08-15
include description of bindings in the reference

nikomatsakis:
This describes the behavior implemented in https://github.com/rust-lang/rust/pull/63376.

cc @Centril

### DIFF
--- a/src/lifetime-elision.md
+++ b/src/lifetime-elision.md
@@ -75,6 +75,14 @@ containing type is first used to try to infer a bound.
 * If there is more than one bound from the containing type then an explicit
   bound must be specified
 
+If the trait object is used as a binding for an associated type (e.g.,
+`Item = dyn Trait`) then the containing trait is first used to try to
+infer a bound, in an analogous way to type arguments:
+
+* If there is a unique bound from the containing type then that is the default
+* If there is more than one bound from the containing type then an explicit
+  bound must be specified
+
 If neither of those rules apply, then the bounds on the trait are used:
 
 * If the trait is defined with a single lifetime _bound_ then that bound is
@@ -128,6 +136,29 @@ impl<'a> dyn Foo<'a> + 'a {}
 struct TwoBounds<'a, 'b, T: ?Sized + 'a + 'b>
 TwoBounds<'a, 'b, dyn Foo<'c>>
 ```
+
+Here is an example featuring associated type bindings:
+
+```rust,ignore
+trait Bar<'a> {
+  type Item1: ?Sized;
+  type Item2: ?Sized + 'a;
+}
+trait Baz { }
+
+dyn Bar<
+  'x
+  Item1 = dyn Baz, // defaults to `dyn Baz + 'static`
+  Item2 = dyn Baz, // defaults to 'dyn Baz + `x`
+> 
+```
+
+**Note:** As of this writing, the rules for associated type bindings
+are incompletely implemented, and explicit bounds may sometimes be
+required when they ought not to be. See [rust-lang/rust#63618][] for
+more details.
+
+[rust-lang/rust#63618]: https://github.com/rust-lang/rust/issues/63618
 
 ## `'static` lifetime elision
 


src/lifetime-elision.md
```diff
@@ -71,6 +71,14 @@ rules.
 If the trait object is used as a type argument of a generic type then the
 containing type is first used to try to infer a bound.
 
+* If there is a unique bound from the containing type then that is the default
```
Havvy:
```suggestion
* If there is a unique bound from the containing type then that is the default.
```

src/lifetime-elision.md
```diff
@@ -71,6 +71,14 @@ rules.
...
+* If there is more than one bound from the containing type then an explicit
+  bound must be specified
```
Havvy:
```suggestion
  bound must be specified.
```

src/lifetime-elision.md
```diff
@@ -71,6 +71,14 @@ rules.
...
+
+If the trait object is used as a binding for an associated type (e.g.,
```
Havvy: I'm not sure the "e.g." parenthetical is useful here. But if we keep it, it should say "for example". It'd be nice if "trait object" and "associated type" were links. Or link to the full example?

src/lifetime-elision.md
```diff
@@ -129,6 +137,29 @@ struct TwoBounds<'a, 'b, T: ?Sized + 'a + 'b>
...
+trait Baz { }
+
+dyn Bar<
+  'x
+  Item1 = dyn Baz, // defaults to `dyn Baz + 'static`
+  Item2 = dyn Baz, // defaults to 'dyn Baz + `x`
+> 
+```
+
+**Note:** As of this writing, the rules for associated type bindings
```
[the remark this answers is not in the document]
nikomatsakis: Hmm, this phrase suggests a temporary condition that is expected to change. It doesn't feel redundant to me.

src/lifetime-elision.md
nikomatsakis: Well, perhaps the e.g. isn't the most elegant way to express it, but I think that showing the syntax will go along way towards helping people understand what the other words actually mean. 

I can definitely make links.

Separate but relevant: In general, I tend to think we should deprecate the term trait objects and move towards "dyn types" or something like that, but that'd be a shift obviously. I do think though that leveraging the keyword in our terminology makes a lot of sense.

src/lifetime-elision.md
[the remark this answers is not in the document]
Centril:
```suggestion
**Note:** As of 2019-08-16, the rules for associated type bindings
```

src/lifetime-elision.md
```diff
@@ -71,9 +71,17 @@ rules.
 If the trait object is used as a type argument of a generic type then the
 containing type is first used to try to infer a bound.
 
-* If there is a unique bound from the containing type then that is the default
+* If there is a unique bound from the containing type then that is the default.
```
Centril:
```suggestion
* If there is a unique lifetime bound from the containing type then that is the default.
```

src/lifetime-elision.md
```diff
@@ -71,9 +71,17 @@ rules.
...
 * If there is more than one bound from the containing type then an explicit
```
Centril:
```suggestion
* If there is more than one lifetime bound from the containing type then an explicit
```

src/lifetime-elision.md
```diff
@@ -71,9 +71,17 @@ rules.
...
-  bound must be specified
+  bound must be specified.
+
+If the trait object is used as a binding for an associated type (e.g.,
+`Item = dyn Trait`) then the containing trait is first used to try to
+infer a bound, in an analogous way to type arguments:
```
Centril:
```suggestion
infer a lifetime bound, in an analogous way to type arguments:
```

src/lifetime-elision.md
```diff
@@ -71,9 +71,17 @@ rules.
...
+
+* If there is a unique bound from the containing type then that is the default.
```
Centril:
```suggestion
* If there is a unique lifetime bound from the containing type then that is the default.
```

src/lifetime-elision.md
```diff
@@ -71,9 +71,17 @@ rules.
...
+* If there is more than one bound from the containing type then an explicit
```
Centril:
```suggestion
* If there is more than one lifetime bound from the containing type then an explicit
```

src/lifetime-elision.md
```diff
@@ -129,6 +137,27 @@ struct TwoBounds<'a, 'b, T: ?Sized + 'a + 'b>
...
+trait Bar<'a> {
+  type Item1: ?Sized;
+  type Item2: ?Sized + 'a;
+}
+trait Baz { }
+
+dyn Bar<
+  'x
+  Item1 = dyn Baz, // defaults to `dyn Baz + 'static`
+  Item2 = dyn Baz, // defaults to 'dyn Baz + `x`
```
Centril: This is a great *and more complicated* (which is why it is great) example to include as a test in the PR!

ehuss:
@nikomatsakis I was looking to merge this with the minor editorial changes, but I see that this example [doesn't compile](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2a1c0f0c8fa014f2afe27df011969707) (aside from the syntax issues).  I assume this is because of https://github.com/rust-lang/rust/issues/63618.  

IIUC, the rules for defaults for associated types are not implemented at all?  If that is correct, then I personally would prefer to document what is implemented, instead of what is intended in this case.  (Following the same kind of logic that we don't document unstable features, future intentions, unimplemented RFCs, etc.)

If there are cases that are implemented, then I think it would be better to use those as examples. If they are not implemented at all, then I would prefer to remove the associated type rules for now and change the example to make it clear that it fails to infer.

We can very easily add the rules back once they are implemented.

nikomatsakis:
@ehuss oof, I totally forgot about this PR.

Havvy:
What's the status on this?

traviscross:
Per the comment above here...

- https://github.com/rust-lang/reference/pull/654#issuecomment-619239004

@rustbot author
@rustbot labels +S-blocked

PR #659, opened 2019-08-25
update UB

RalfJung:
This syncs the Reference definition of UB with what the Nomicon has [recently been updated to](https://github.com/rust-lang-nursery/nomicon/pull/149), including the still-under-review changes in https://github.com/rust-lang-nursery/nomicon/pull/162.

Cc @Centril @Gankra

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -44,15 +44,15 @@ code.
   * A `!` (all values are invalid for this type).
   * [Uninitialized memory][undef] in the value of an integer (`i*`/`u*`),
     floating point value (`f*`), or raw pointer.
-  * A dangling or unaligned reference or `Box`, or one that points to an invalid value.
+  * A reference or `Box<T>` that is dangling, unaligned, or points to an invalid value.
   * Invalid metadata in a wide reference, `Box`, or raw pointer:
     * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the reference points to.
+      `Trait` that matches the actual dynamic trait the pointer or reference points to.
     * Slice metadata is invalid if if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
   * Non-UTF-8 byte sequences in a `str`.
   * Invalid values for a type with a custom definition of invalid values, such
-    as a `NonNull` that is null. (Requesting custom invalid values is an
+    as a `NonNull<T>` that is null. (Requesting custom invalid values is an
     unstable feature, but some stable libstd types, like `NonNull`, make use of
     it.)
 
@@ -67,7 +67,7 @@ points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
 pointer value and the size of the pointee type. As a consequence, if the span is
 empty, "dangling" is the same as "non-null". Note that slices point to their
-entire range, so it is very important that the length metadata is never too
+entire range, so it is important that the length metadata is never too
 large. In particular, allocations and therefore slices cannot be bigger than
 `isize::MAX` bytes.
 
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -57,12 +57,6 @@ code.
     > **Note**: For `rustc`, those types are [`NonNull<T>`] and [`NonZero*`].
     > Requesting custom invalid types requires the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
-> **Note**: Undefined behavior affects the entire program. For example, calling
-> a function in C that exhibits undefined behavior of C means your entire
-> program contains undefined behaviour that can also affect the Rust code. And
-> vice versa, undefined behavior in Rust can cause adverse affects on code
-> executed by any FFI calls to other languages.
-
 A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
@@ -71,6 +65,12 @@ empty, "dangling" is the same as "non-null". Note that slices point to their
 entire range, so it is important that the length metadata is never too
 large. In particular, allocations and therefore slices cannot be bigger than
 `isize::MAX` bytes.
+
+> **Note**: Undefined behavior affects the entire program. For example, calling
+> a function in C that exhibits undefined behavior of C means your entire
+> program contains undefined behaviour that can also affect the Rust code. And
+> vice versa, undefined behavior in Rust can cause adverse affects on code
+> executed by any FFI calls to other languages.
 
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
 [pointer aliasing rules]: http://llvm.org/docs/LangRef.html#pointer-aliasing-rules
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -26,7 +26,7 @@ code.
 * Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
 * Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
   [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
-* Mutating immutable data. All data inside a `const` is immutable. Moreover, all
+* Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
   data reached through a shared reference or data owned by an immutable binding
   is immutable, unless that data is contained within an [`UnsafeCell<U>`].
 * Invoking undefined behavior via compiler intrinsics.
@@ -73,6 +73,7 @@ large. In particular, allocations and therefore slices cannot be bigger than
 > vice versa, undefined behavior in Rust can cause adverse affects on code
 > executed by any FFI calls to other languages.
 
+[`const`]: items/constant-items.html
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
 [pointer aliasing rules]: http://llvm.org/docs/LangRef.html#pointer-aliasing-rules
 [undef]: http://llvm.org/docs/LangRef.html#undefined-values
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -52,10 +52,10 @@ code.
     * Slice metadata is invalid if if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
   * Non-UTF-8 byte sequences in a `str`.
-  * Invalid values for a type with a custom definition of invalid values, such
-    as a `NonNull<T>` that is null. (Requesting custom invalid values is an
-    unstable feature, but some stable libstd types, like `NonNull`, make use of
-    it.)
+  * Invalid values for a type with a custom definition of invalid values.
+
+    > **Note**: For `rustc`, those types are [`NonNull<T>`] and [`NonZero*`].
+    > Requesting custom invalid types requires the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
 > **Note**: Undefined behavior affects the entire program. For example, calling
 > a function in C that exhibits undefined behavior of C means your entire
@@ -78,3 +78,5 @@ large. In particular, allocations and therefore slices cannot be bigger than
 [`target_feature`]: attributes/codegen.md#the-target_feature-attribute
 [`UnsafeCell<U>`]: ../std/cell/struct.UnsafeCell.html
 [Rustonomicon]: ../nomicon/index.html
+[`NonNull<T>`]: ../core/ptr/struct.NonNull.html
+[`NonZero*`]: ../core/num/index.html
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -32,7 +32,7 @@ code.
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
   does not support (see [`target_feature`]).
-* Calling a function with the wrong call ABI (including unwind ABI).
+* Calling a function with the wrong call ABI (in particular, with the wrong unwind ABI).
 * Producing an invalid value, even in private fields and locals. "Producing" a
   value happens any time a value is assigned to or read from a place, passed to
   a function/primitive operation or returned from a function/primitive
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -27,7 +27,7 @@ code.
 * Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
   [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
 * Mutating non-mutable data (that is, data reached through a shared
-  reference or data owned by a `let` binding), unless that data is contained
+  reference or data owned by an immutable binding), unless that data is contained
   within an [`UnsafeCell<U>`].
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -53,9 +53,10 @@ code.
       (i.e., it must not be read from uninitialized memory).
   * Non-UTF-8 byte sequences in a `str`.
   * Invalid values for a type with a custom definition of invalid values.
+    In the standard library, this affects [`NonNull<T>`] and [`NonZero*`].
 
-    > **Note**: For `rustc`, those types are [`NonNull<T>`] and [`NonZero*`].
-    > Requesting custom invalid values requires the unstable `rustc_layout_scalar_valid_range_*` attributes.
+    > **Note**: For `rustc`, requesting custom invalid values requires the
+    > unstable `rustc_layout_scalar_valid_range_*` attributes.
 
 A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -42,15 +42,15 @@ code.
   * A null `fn` pointer.
   * A value in a `char` which is a surrogate or above `char::MAX`.
   * A `!` (all values are invalid for this type).
-  * A dangling or unaligned reference or `Box`, or one that points to an invalid value.
-  * Invalid metadata in a wide reference, `Box` or raw pointer:
-    * slice metadata is invalid if the slice has a total size larger than
-      `isize::MAX` bytes in memory.
-    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the reference points to.
-  * Non-UTF-8 byte sequences in a `str`.
   * [Uninitialized memory][undef] in the value of an integer (`i*`/`u*`),
     floating point value (`f*`), or raw pointer.
+  * A dangling or unaligned reference or `Box`, or one that points to an invalid value.
+  * Invalid metadata in a wide reference, `Box`, or raw pointer:
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+      `Trait` that matches the actual dynamic trait the reference points to.
+    * Slice metadata is invalid if if the length is not a valid `usize`
+      (i.e., it must not be read from uninitialized memory).
+  * Non-UTF-8 byte sequences in a `str`.
   * Invalid values for a type with a custom definition of invalid values, such
     as a `NonNull` that is null. (Requesting custom invalid values is an
     unstable feature, but some stable libstd types, like `NonNull`, make use of
@@ -68,7 +68,8 @@ part of *some* allocation). The span of bytes it points to is determined by the
 pointer value and the size of the pointee type. As a consequence, if the span is
 empty, "dangling" is the same as "non-null". Note that slices point to their
 entire range, so it is very important that the length metadata is never too
-large.
+large. In particular, allocations and therefore slices cannot be bigger than
+`isize::MAX` bytes.
 
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
 [pointer aliasing rules]: http://llvm.org/docs/LangRef.html#pointer-aliasing-rules
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -32,7 +32,7 @@ code.
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
   does not support (see [`target_feature`]).
-* Calling a function with the wrong call ABI (in particular, with the wrong unwind ABI).
+* Calling a function with the wrong call ABI or wrong unwind ABI.
 * Producing an invalid value, even in private fields and locals. "Producing" a
   value happens any time a value is assigned to or read from a place, passed to
   a function/primitive operation or returned from a function/primitive
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -55,7 +55,7 @@ code.
   * Invalid values for a type with a custom definition of invalid values.
 
     > **Note**: For `rustc`, those types are [`NonNull<T>`] and [`NonZero*`].
-    > Requesting custom invalid types requires the unstable `rustc_layout_scalar_valid_range_*` attributes.
+    > Requesting custom invalid values requires the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
 A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -26,9 +26,9 @@ code.
 * Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
 * Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
   [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
-* Mutating non-mutable data (that is, data reached through a shared
-  reference or data owned by an immutable binding), unless that data is contained
-  within an [`UnsafeCell<U>`].
+* Mutating immutable data. All data inside a `const` is immutable. Moreover, all
+  data reached through a shared reference or data owned by an immutable binding
+  is immutable, unless that data is contained within an [`UnsafeCell<U>`].
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
   does not support (see [`target_feature`]).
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
-  * Indexing outside of the bounds of an object with [`offset`] with
-    the exception of one byte past the end of the object.
-  * Using [`std::ptr::copy_nonoverlapping_memory`], a.k.a. the `memcpy32`and
-    `memcpy64` intrinsics, on overlapping buffers.
-* Invalid values in primitive types, even in private fields and locals:
-  * Dangling or null references and boxes.
+* Invoking undefined behavior via compiler intrinsics.
+* Executing code compiled with platform features that the current platform
+  does not support (see [`target_feature`]).
+* Unwinding into another language.
```
gnzlbg: Wouldn't this be UB in that other language, depending on that languages rules ?

RalfJung: Would it? I am not sure. Are you expecting other languages to have a clause like "it is UB to unwind *from* another language"? Rust does not have such a clause.

gnzlbg: Unwinding from another language would also be ok, as long as the unwind is identical to what Rust uses. E.g. in

```rust
extern "Rust" {
    fn foo();
}
unsafe { foo() }
```

`foo` can unwind there without issues, and whether `foo` is actually implemented in Rust, or say assembly, doesn't matter as long as it follows the Rust ABI.

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+* Producing an invalid value, even in private fields and locals. "Producing" a
+  value happens any time a value is assigned, passed to a function/primitive
+  operation or returned from a function/primitive operation.
+  The following values are invalid (at their respective type):
```
gnzlbg: Maybe mention here that this list isn't necessary complete, and that just because a value is not mentioned here as invalid for some type does not mean that value is valid?

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+      `Trait` that matches the actual dynamic trait the reference points to.
   * Non-UTF-8 byte sequences in a `str`.
-* Executing code compiled with platform features that the current platform
-  does not support (see [`target_feature`]).
+  * [Uninitialized memory][undef] in the value of an integer (`i*`/`u*`),
+    floating point value (`f*`), or raw pointer.
+  * Invalid values for a type with a custom definition of invalid values, such
+    as a `NonNull` that is null. (Requesting custom invalid values is an
+    unstable feature, but some stable libstd types, like `NonNull`, make use of
```
gnzlbg: maybe also add `NonZero*` to this list ?

src/behavior-considered-undefined.md
gnzlbg: > Would it? I am not sure. Are you expecting other languages to have a clause like "it is UB to unwind from another language"? Rust does not have such a clause.

Rust says that unwinding from Rust into Rust is ok. The behavior of unwinding from some other language into Rust is not defined. Most other languages do this.

gnzlbg: Maybe you are confusing this with unwinding from `nounwind` functions is UB ?

For example, we do explicitly say that if an `extern "C"` function declaration unwinds, the behavior is undefined. We assume that this does not happen, and e.g. we give these functions the `nounwind` attribute in LLVM-IR.

RalfJung: I am just copying this from the Nomicon, where this line existed [for quite some time already](https://doc.rust-lang.org/1.30.0/nomicon/what-unsafe-does.html).

I think it is crucial that the case of unwinding out of an `extern "C"` is covered here somehow. Any suggestions for better wording?

src/behavior-considered-undefined.md
RalfJung: The list is meant to be complete. But also, there's a big fat "WARNING: The following list is not exhaustive" further up this page.

src/behavior-considered-undefined.md
RalfJung: Isn't one example enough? I think it is.

src/behavior-considered-undefined.md
gnzlbg: Unwinding from extern function declarations whose ABI does not support unwinding, like `extern "C"`, is UB.

gnzlbg: Emphasis on "some ABIs" and "function declarations". Trying to unwind from an `extern "C"` function definition panics, and trying to unwind from an `extern "Rust"` declaration is ok.

src/behavior-considered-undefined.md
gnzlbg: I thought the only two types in libstd that we expose for which this holds were `NonZero*` and `NonNull`. If there are more then one example is enough, but otherwise we can be complete here.

RalfJung: AFAIK those are the only ones -- `NonNull*` is an entire family though, so it's more than two types.

However, I think it is a maintenance hazard to try to list those here.

src/behavior-considered-undefined.md
RalfJung: > Trying to unwind from an extern "C" function definition aborts

AFAIK it doesn't, not yet? See https://github.com/rust-lang/rust/pull/62505.
And even then there will be an attribute to opt ouf of this, with a chance for people to use it the wrong way.

I was asking for a wording for the "unwinding out of Rust to another language" case. I think that is the more interesting one.

RalfJung: Also this code looks a lot like unwinding is UB even when it comes from an `extern "Rust"` *declaration*:

https://github.com/rust-lang/rust/blob/d760df5aea483aae041c9a241e7acacf48f75035/src/librustc_codegen_llvm/attributes.rs#L278-L281

gnzlbg: > AFAIK it doesn't, not yet? See rust-lang/rust#62505.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=8665db1c91123fdaa3617ca8443d6989

That code aborts when the panic tries to escape `foo` :/

RalfJung: We abort-on-panic-from-extern only on nightly, but regularly backport a patch reverting that to beta. It is quite silly. :/

See https://github.com/rust-lang/rust/pull/62603.

gnzlbg: > And even then there will be an attribute to opt of of this, 

AFAIK what will happen here is still very unclear. Chances are there will just be different function ABIs w/o unwinding support, to make that part of the type system.

> I was asking for a wording for the "unwinding out of Rust to another language" case.

Whether you can or cannot unwinding out of an `extern "ABI" fn` depends on the function ABI - some allow this, some do not. The ones that do not, panic. 

When you call a function (from any language) you must uphold its ABI - if you don't, the behavior is undefined. 

For example, if you have a Rust function that unwinds, and call it from Rust code compiled with a slightly different toolchain that happens to be ABI incompatible, the behavior is undefined. Same language, but unwinding triggered UB. If you have a Rust function that unwinds, and after unwinding we land in a different programming language than Rust, saying that this is UB is possible, but kind of pointless, because the program execution isn't in Rust code anymore. It's in some other programming language that might say that unwinding into it from Rust is ok. Either way, we can't tell, and even if we could, I don't know which purpose does this serve.

gnzlbg: > We abort-on-panic-from-extern only on nightly, but regularly backport a patch reverting that to beta. It is quite silly. :/

I haven't used anything but nightly for so long that I forgot that this is still a thing. So yeah, this should mention this.

RalfJung: We agreed on chat that for now, the reference should say that cross-lang unwinding is just plain UB (following the Nomicon), and should be updated once https://github.com/rust-lang/rfcs/pull/2699 is accepted.

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+  * Invalid metadata in a wide reference, `Box` or raw pointer:
+    * slice metadata is invalid if the slice has a total size larger than
+      `isize::MAX` bytes in memory.
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+      `Trait` that matches the actual dynamic trait the reference points to.
   * Non-UTF-8 byte sequences in a `str`.
-* Executing code compiled with platform features that the current platform
-  does not support (see [`target_feature`]).
+  * [Uninitialized memory][undef] in the value of an integer (`i*`/`u*`),
+    floating point value (`f*`), or raw pointer.
```
Centril: and references? -- could we *add* a more blanket ban on uninit memory outside of `MaybeUninit<T>` and equivalents?

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
-* Breaking the [pointer aliasing rules] on accesses through raw pointers;
-  a subset of the rules used by C.
-* `&mut T` and `&T` follow LLVM’s scoped [noalias] model, except if the `&T`
-  contains an [`UnsafeCell<U>`].
-* Mutating non-mutable data &mdash; that is, data reached through a shared
+* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
+  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Mutating non-mutable data (that is, data reached through a shared
   reference or data owned by a `let` binding), unless that data is contained
```
Centril: Let bindings are too specific here... this is about bindings in general?

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
-  * Using [`std::ptr::copy_nonoverlapping_memory`], a.k.a. the `memcpy32`and
-    `memcpy64` intrinsics, on overlapping buffers.
-* Invalid values in primitive types, even in private fields and locals:
-  * Dangling or null references and boxes.
+* Invoking undefined behavior via compiler intrinsics.
+* Executing code compiled with platform features that the current platform
+  does not support (see [`target_feature`]).
+* Unwinding into another language.
+* Producing an invalid value, even in private fields and locals. "Producing" a
+  value happens any time a value is assigned, passed to a function/primitive
```
Centril: assigned to a place, right? this could use some elaboration...

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
 </div>
 
 * Data races.
-* Dereferencing a null or dangling raw pointer.
-* Unaligned pointer reading and writing outside of [`read_unaligned`]
```
Centril: Can you explain removing this?

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
-  and [`write_unaligned`].
-* Reads of [undef] \(uninitialized) memory.
-* Breaking the [pointer aliasing rules] on accesses through raw pointers;
```
Centril: and removing this?

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+* Producing an invalid value, even in private fields and locals. "Producing" a
+  value happens any time a value is assigned, passed to a function/primitive
+  operation or returned from a function/primitive operation.
+  The following values are invalid (at their respective type):
   * A value other than `false` (`0`) or `true` (`1`) in a `bool`.
   * A discriminant in an `enum` not included in the type definition.
+  * A null `fn` pointer.
   * A value in a `char` which is a surrogate or above `char::MAX`.
+  * A `!` (all values are invalid for this type).
+  * A dangling or unaligned reference or `Box`, or one that points to an invalid value.
```
Centril:
```suggestion
  * A reference or `Box<T>` that is dangling, unaligned, or points to an invalid value.
```
Grouping is easier to parse this way.

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+  * Invalid metadata in a wide reference, `Box` or raw pointer:
```
Centril:
```suggestion
  * Invalid metadata in a wide reference, `Box<T>` or raw pointer:
```

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
+    * slice metadata is invalid if the slice has a total size larger than
+      `isize::MAX` bytes in memory.
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+      `Trait` that matches the actual dynamic trait the reference points to.
```
Centril:
```suggestion
      `Trait` that matches the actual dynamic trait the pointer or reference points to.
```

RalfJung:
I adjusted the text to fix the [slice issue that came up](https://github.com/rust-lang/unsafe-code-guidelines/issues/166#issuecomment-524728054).

src/behavior-considered-undefined.md
RalfJung: It is redundant: unaligned references are UB when produced, unaligned pointers are still called out, the rest is `unsafe` lib functions and nothing language-inherent.

src/behavior-considered-undefined.md
RalfJung: It is still there, I just removed "on accesses through raw pointers" as that makes it sound like unsafe code doesn't have to worry when using references.

src/behavior-considered-undefined.md
RalfJung: Not sure. This is a pre-existing clause that the nomicon does not have. I think this is actually captured by "pointer aliasing rules" (at least, Stacked Borrows also enforces this).

src/behavior-considered-undefined.md
RalfJung: Assigned as in `=`. But sure, I can try to be more specific. I will say "assigned to or read from a place"?

Centril: sgtm

src/behavior-considered-undefined.md
RalfJung: An uninitialized reference is certainly not aligned and non-null, is it?

I much prefer to view "no UB outside certain types" as a consequence of our rules, than a primitive rule.

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,56 @@ code.
...
   * Non-UTF-8 byte sequences in a `str`.
-* Executing code compiled with platform features that the current platform
-  does not support (see [`target_feature`]).
+  * [Uninitialized memory][undef] in the value of an integer (`i*`/`u*`),
+    floating point value (`f*`), or raw pointer.
+  * Invalid values for a type with a custom definition of invalid values, such
+    as a `NonNull` that is null. (Requesting custom invalid values is an
```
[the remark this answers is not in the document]
RalfJung: I don't understand what you are suggesting I do here. Sure the rustc guide should explain that if it does not already, but this also must be called out here.

src/behavior-considered-undefined.md
Centril: Hmm; can we s/`let binding`/binding for now then?

src/behavior-considered-undefined.md
Centril: Indeed it is; 👍

src/behavior-considered-undefined.md
RalfJung: Well not all bindings but only immutable ones. Done.

src/behavior-considered-undefined.md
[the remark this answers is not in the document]
Centril: I'm suggesting that "(Requesting custom...)" should be moved to a note or removed as that ability is unstable. If you have a note then it should point to the relevant rustc guide section.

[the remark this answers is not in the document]
RalfJung: What do you mean by "note"? It already is a parenthetical remark.

I think this is needed here because otherwise people reading this will be confused. Most users of Rust do not and should not know about this feature.

Maybe @gnzlbg was right and we should just also exhaustively list the affected types here. @Gankra didn't like this but such stylistic differences between Nomicon and Reference seem fine.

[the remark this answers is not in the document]
Centril: > What do you mean by "note"? It already is a parenthetical remark.

https://doc.rust-lang.org/nightly/reference/#conventions

How rustc achieves the fact that certain types have various niches is not part of the language definition.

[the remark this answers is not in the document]
RalfJung: This is a document for human beings, not robots.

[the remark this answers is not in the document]
RalfJung: This parenthetical also does not fit "useful information about the state of the book or [...] useful, but mostly out of scope, information".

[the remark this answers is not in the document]
Centril: > This is a document for human beings, not robots.

I don't know what you mean by this.

Implementation details of rustc belong in notes; that's the convention in the reference that is used. cc @Havvy

src/behavior-considered-undefined.md
Centril: I'm confused; the parent bullet here talks about invalid values for certain types and specifically the current bullet talks about uninit memory.

src/behavior-considered-undefined.md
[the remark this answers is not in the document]
RalfJung: I mean that this should be readable, so we have to mention `NonNull` here and not send the user on a chase for looking up what "custom definition of invalid values" means.

src/behavior-considered-undefined.md
RalfJung: This bullet talks about specific types just like all the others. It talks about those types where uninitialized memory is the *only* invalid value.

src/behavior-considered-undefined.md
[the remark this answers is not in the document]
Centril: I don't understand how a note in the conventional style I linked above about how rustc achieves this would be unreadable (it's also not necessary to say it is unstable, only a note re. the names of the attributes + links with elaboration in the rustc guide).

In other words I think this should be changed into:

> **Note:** In `rustc`, such custom definitions are achieved with [`rustc_layout_scalar_valid_range_start`](link_to_rustc_guide), other thing, ..., and last thing.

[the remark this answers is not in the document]
RalfJung: I don't think the attribute is very interesting to most users; the types are. Also if this is what notes are for, I think the def.n you linked me to need adjustment; this is not "mostly out of scope". It is relevant for the vast majority of readers.

Is there an example for how a note rendered in a nested enumeration looks like?

src/behavior-considered-undefined.md
Centril: Oh... I see; should we add another bullet which talks about uninit memory more generally then?

RalfJung: That as my point above that there is nothing else to say about uninit memory.

src/behavior-considered-undefined.md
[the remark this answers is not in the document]
Centril: > I don't think the attribute is very interesting to most users; the types are.

But you've already mentioned one of those types. Saying that this is achieved with some unstable attribute without mentioning the name of that attribute is *vague* and doesn't say what the other types are. I think it's far easier to know the name of the attribute and then search for it in the source if you want to know what the full list of types are. Alternatively, mention all of the base types.

> Is there an example for how a note rendered in a nested enumeration looks like?

See https://github.com/rust-lang-nursery/reference/search?p=2&q=note&type=&utf8=%E2%9C%93 for a list of notes. I didn't see any example in a nested enumeration when skimming.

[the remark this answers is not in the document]
RalfJung: Something like this?

src/behavior-considered-undefined.md
```diff
@@ -23,30 +23,48 @@ code.
...
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+      `Trait` that matches the actual dynamic trait the pointer or reference points to.
+    * Slice metadata is invalid if if the length is not a valid `usize`
+      (i.e., it must not be read from uninitialized memory).
   * Non-UTF-8 byte sequences in a `str`.
-* Executing code compiled with platform features that the current platform
-  does not support (see [`target_feature`]).
+  * Invalid values for a type with a custom definition of invalid values.
+
+    > **Note**: For `rustc`, those types are [`NonNull<T>`] and [`NonZero*`].
```
RalfJung: I am not sure how much of the libstd surface we consider a rustc implementation detail vs. part of what every compiler should provide.

src/behavior-considered-undefined.md
```diff
@@ -23,44 +23,58 @@ code.
...
-  * Indexing outside of the bounds of an object with [`offset`] with
-    the exception of one byte past the end of the object.
-  * Using [`std::ptr::copy_nonoverlapping_memory`], a.k.a. the `memcpy32`and
-    `memcpy64` intrinsics, on overlapping buffers.
-* Invalid values in primitive types, even in private fields and locals:
-  * Dangling or null references and boxes.
+* Invoking undefined behavior via compiler intrinsics.
+* Executing code compiled with platform features that the current platform
+  does not support (see [`target_feature`]).
+* Unwinding into another language.
```
gnzlbg: I think this would be better as something like this:

```suggestion
* Unwinding into another language or into Rust compiled with a different toolchain version or into Rust compiled with the same toolchain version but different compilation options except for those options for which we note otherwise.
```

src/behavior-considered-undefined.md
Centril: Aside from operating system specific APIs, the public interface of the standard library should be provided by all implementations.

src/behavior-considered-undefined.md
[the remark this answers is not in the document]
Centril: Yeah that's better.

src/behavior-considered-undefined.md
Centril: All of this "toolchain" stuff is highly implementation (rustc) specific and falls under "the ABI is not stable".

src/behavior-considered-undefined.md
RalfJung: Okay, so I moved the type list out of the note.

src/behavior-considered-undefined.md
```diff
@@ -23,30 +23,49 @@ code.
...
+    * Slice metadata is invalid if if the length is not a valid `usize`
+      (i.e., it must not be read from uninitialized memory).
   * Non-UTF-8 byte sequences in a `str`.
-* Executing code compiled with platform features that the current platform
-  does not support (see [`target_feature`]).
+  * Invalid values for a type with a custom definition of invalid values.
+    In the standard library, this affects [`NonNull<T>`] and [`NonZero*`].
+
+    > **Note**: For `rustc`, requesting custom invalid values requires the
+    > unstable `rustc_layout_scalar_valid_range_*` attributes.
```
gnzlbg: Shouldn't this note be in the rustc book ?

src/behavior-considered-undefined.md
gnzlbg: @Centril do you have a better way to word this? `"Unwinding into another language"` is UB, but so is unwinding into Rust in many cases as well.

RalfJung: Can we replace that list of "or <different Rust>" by something like "or into a separately compiled Rust library"? Not sure what the best wording is, I am looking for something that basically days "if you link together two Rust dylib/cdylib, that's basically crossing a language barrier".

src/behavior-considered-undefined.md
RalfJung: I was asked to add the attribute so I did.

src/behavior-considered-undefined.md
gnzlbg: Maybe "Violating the call ABI and unwinding ABI of a function on function calls" ?

src/behavior-considered-undefined.md
Centril: We can shorten this to "`rustc` achieves this with the `rustc_layout_scalar_valid_range_*` attributes" but overall this should link to a rustc guide page with longer elaboration.

src/behavior-considered-undefined.md
gnzlbg: Another example is that, AFAICT, according to this list, `transmuting` an `extern "Rust" fn` to an `extern "C" fn` is not UB, and I can't find any item that says that calling the Rust function via the `extern "C" fn` is UB. 

A bullet point saying that "Calling a function with an incorrect call ABI (including unwinding ABI) is UB" would probably solve both issues.

src/behavior-considered-undefined.md
gnzlbg: Oh, I thought that the reference should contain no mentions of unstable features. If it is ok for it to contain those, then I'm fine with it.

src/behavior-considered-undefined.md
RalfJung: I made it "Calling a function with the wrong call ABI (including unwind ABI)."

RalfJung:
I think I resolved all comments.

src/behavior-considered-undefined.md
```diff
@@ -23,30 +23,49 @@ code.
...
-* `&mut T` and `&T` follow LLVM’s scoped [noalias] model, except if the `&T`
-  contains an [`UnsafeCell<U>`].
-* Mutating non-mutable data &mdash; that is, data reached through a shared
-  reference or data owned by a `let` binding), unless that data is contained
+* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
+  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Mutating non-mutable data (that is, data reached through a shared
+  reference or data owned by an immutable binding), unless that data is contained
   within an [`UnsafeCell<U>`].
```
RalfJung: Considering [this discussion](https://github.com/rust-lang/rust/pull/63955#discussion_r318447052), I think I'll have to add a clause like "this is always UB for memory in a `const`, no matter the `UnsafeCell`".

Centril: Yes please!

src/behavior-considered-undefined.md
```diff
@@ -23,30 +23,49 @@ code.
...
-  * Indexing outside of the bounds of an object with [`offset`] with
-    the exception of one byte past the end of the object.
-  * Using [`std::ptr::copy_nonoverlapping_memory`], a.k.a. the `memcpy32`and
-    `memcpy64` intrinsics, on overlapping buffers.
-* Invalid values in primitive types, even in private fields and locals:
-  * Dangling or null references and boxes.
+* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
+  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Mutating immutable data. All data inside a `const` is immutable. Moreover, all
```
Centril:
```suggestion
* Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
```

src/behavior-considered-undefined.md
```diff
@@ -57,10 +76,8 @@ code.
 [noalias]: http://llvm.org/docs/LangRef.html#noalias
```
Centril:
```suggestion
[`const`]: items/constant-items.html
[noalias]: http://llvm.org/docs/LangRef.html#noalias
```

Havvy:
@Centril Is there anything left here, or can it be merged?

PR #661, opened 2019-08-28
Break statements not allowed outside loops

hawkinsw:
According to the reference, `break` expressions are valid
expressions in the context of `if`, `if let` and `while let`
expressions but this is not true. Correct for that.


ehuss:
Oh, no worries! It is not a waste of time.

Issue #568, opened 2019-04-21
Document name resolution.


ehuss:
Name resolution should be documented. This is a large topic. A few pieces I can think of:

- How items are added to scope (#129).
    - Don't forget tricky things like `macro_use` and `macro_export`.
- The different namespaces (type, value, and macro?).
- Should ribs be explained? Seems like a useful way to talk about a scope region.
- What is a valid path?
- The different scope layers. I saw this list once, which is likely incorrect, but a starting point:
    - local variables
    - items in unnamed blocks
    - items in the current module
    - crate names in the "extern prelude"
    - standard library prelude
    - language prelude (built-in types)
- I think all prelude kinds should be documented and explicitly named. Currently the reference mentions "the prelude" which is ambiguous, and the extern prelude. There are other preludes like the language prelude, the `macro_use` prelude, the "tool prelude" (I think), etc. The order they are processed should be documented to clarify how shadowing works.
    - `no_implicit_prelude` should clarify which preludes are not included.
    - crates-and-source-files.md#preludes-and-no_std should also be updated.
    - modules.md#prelude-items doesn't mention extern crate prelude. It needs updating.
- I believe there are some complexities around name resolution during import, particularly with macros, since the modules are in flux.
- How does resolution work that a struct field and method can have the same name?
- There are a lot of details that will be difficult to cover. For example: https://github.com/rust-lang/rust/issues/56414.
- See also: https://github.com/rust-lang/rfcs/blob/master/text/1560-name-resolution.md
- Items declared inside functions cannot refer to the type parameters of the enclosing function.
- Interesting insight on how macro resolution works: https://internals.rust-lang.org/t/announcing-rust-2018-beta-release/8901/17
- Detailed comment from petrochenkov about item containers: https://github.com/rust-lang/rfcs/pull/1976#issuecomment-301903528
- This simplifying pr would be good to understand: https://github.com/rust-lang/rust/pull/58349
- Interesting comment: https://github.com/rust-lang/rust/pull/59047#discussion_r264027037
- This PR has some interesting insight into macro paths: https://github.com/rust-lang/rust/pull/52841
- How does `use` handle something that is available in multiple namespaces (such as a module and a type).

gnzlbg:
cc @petrochenkov

ehuss:
Just to kick off a little discussion of the outline: a long while back I started an outline here: https://github.com/ehuss/reference/commit/bb2e84bf3b84f4e7a9b2532feae82659c603d894

[...]

PR #665, opened 2019-08-30
Undocument `use` paths.

ehuss:
The current documentation is woefully out of date and incomplete.  I've been trying off and on to study the resolver to begin #568, which will be required to finish this, but I probably won't be able to finish for a long while.

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -82,13 +82,9 @@ In this example, the module `quux` re-exports two public names defined in
 
 ## `use` Paths
 
-Paths in `use` items must start with a crate name or one of the [path
-qualifiers] `crate`, `self`, `super`, or `::`. `crate` refers to the current
-crate. `self` refers to the current module. `super` refers to the parent
-module. `::` can be used to explicitly refer to a crate, requiring an extern
-crate name to follow.
+> **Note**: This section is incomplete.
 
-An example of what will and will not work for `use` items:
+Some examples of what will and will not work for `use` items:
 <!-- Note: This example works as-is in either 2015 or 2018. -->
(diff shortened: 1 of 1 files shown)


src/items/use-declarations.md
```diff
@@ -82,13 +82,9 @@ In this example, the module `quux` re-exports two public names defined in
 
 ## `use` Paths
 
-Paths in `use` items must start with a crate name or one of the [path
-qualifiers] `crate`, `self`, `super`, or `::`. `crate` refers to the current
-crate. `self` refers to the current module. `super` refers to the parent
-module. `::` can be used to explicitly refer to a crate, requiring an extern
-crate name to follow.
```
Centril: The snippet above you're removing seems roughly accurate? Rather than removing it, can we *just add* your note?

ehuss: I think it is misleading, because it seems to imply there is something special about `use` paths or that the first segment is restricted somehow. This hasn't been true since uniform paths (which feels like an eternity ago, but was only in January!).

The `use` path is really any "vaild" path for some definition of "valid".

The "paths" chapter describes what all these qualifiers mean, so I don't think they need to be repeated. 

From my incomplete notes (which are likely outdated), paths can also start with:
- Item? in the current scope.  (Item may be the wrong word.)
- A macro in the macro_use prelude.
- An item in the prelude.
- Built-in type.
- Built-in attribute (but can't be used for some reason).
- Built-in tool module (but can't be used for some reason).

Plus a bunch of restrictions on what *can't* be used (associated items, built-in macros, type parameters, local variables, `Self` types, `impl` items, etc).

PR #668, opened 2019-09-02
Fix #664: Review Oxford comma usage.

### LINKED ISSUE #664: Oxford comma usage is inconsistent
In PR #663, @hawkinsw stated that the comma usage is inconsistent.

I don't know which sites are missing commas, but they should be added where missing. @hawkinsw: Can you point out those locations?

hawkinsw:
This is an update to correct for the places where the Oxford
comma is not used.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -25,7 +25,7 @@
 *Conditionally compiled source code* is source code that may or may not be
 considered a part of the source code depending on certain conditions. <!-- This
 definition is sort of vacuous --> Source code can be conditionally compiled
-using [attributes], [`cfg`] and [`cfg_attr`], and the built-in [`cfg` macro].
+using the [attributes] [`cfg`] and [`cfg_attr`] and the built-in [`cfg` macro].
 These conditions are based on the target architecture of the compiled crate,
 arbitrary values passed to the compiler, and a few other miscellaneous things
 further described below in detail.
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -37,8 +37,8 @@ to be run.
 * Index expressions, [array indexing] or [slice] with a `usize`.
 * [Range expressions].
 * [Closure expressions] which don't capture variables from the environment.
-* Built in [negation], [arithmetic, logical], [comparison] or [lazy boolean]
-  operators used on integer and floating point types, `bool` and `char`.
+* Built-in [negation], [arithmetic], [logical], [comparison] or [lazy boolean]
+  operators used on integer and floating point types, `bool`, and `char`.
 * Shared [borrow]s, except if applied to a type with [interior mutability].
 * The [dereference operator].
 * [Grouped] expressions.
@@ -57,7 +57,7 @@ A _const context_ is one of the following:
   * [statics]
   * [enum discriminants]
 
-[arithmetic, logical]:  expressions/operator-expr.md#arithmetic-and-logical-binary-operators
+[arithmetic]:           expressions/operator-expr.md#arithmetic-and-logical-binary-operators
 [array expressions]:    expressions/array-expr.md
 [array indexing]:       expressions/array-expr.md#array-and-slice-indexing-expressions
 [array indexing]:       expressions/array-expr.md#array-and-slice-indexing-expressions
@@ -84,6 +84,7 @@ A _const context_ is one of the following:
 [lazy boolean]:         expressions/operator-expr.md#lazy-boolean-operators
 [let statements]:       statements.md#let-statements
 [literals]:             expressions/literal-expr.md
+[logical]:              expressions/operator-expr.md#arithmetic-and-logical-binary-operators
 [negation]:             expressions/operator-expr.md#negation-operators
 [overflow]:             expressions/operator-expr.md#overflow
 [paths]:                expressions/path-expr.md
--- a/src/dynamically-sized-types.md
+++ b/src/dynamically-sized-types.md
@@ -20,7 +20,7 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
   last field, this makes the struct itself a
   <abbr title="dynamically sized type">DST</abbr>.
 
-Notably: [variables], function parameters, [const] and [static] items must be
+> **Note**: [variables], function parameters, [const] items, and [static] items must be
 `Sized`.
 
 [sized]: special-types-and-traits.md#sized
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -265,7 +265,7 @@ a few specific cases:
 
 * Before an expression used as a [statement].
 * Elements of [array expressions], [tuple expressions], [call expressions],
-  tuple-style [struct] and [enum variant] expressions.
+  and tuple-style [struct] and [enum variant] expressions.
   <!--
     These were likely stabilized inadvertently.
     See https://github.com/rust-lang/rust/issues/32796 and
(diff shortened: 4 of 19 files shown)


src/conditional-compilation.md
```diff
@@ -25,7 +25,7 @@
 *Conditionally compiled source code* is source code that may or may not be
 considered a part of the source code depending on certain conditions. <!-- This
 definition is sort of vacuous --> Source code can be conditionally compiled
-using [attributes], [`cfg`] and [`cfg_attr`], and the built-in [`cfg` macro].
```
Havvy: This one is actually semantically a list of two items. Adding the comma there would be a comma splice as well. It's meant to be read as "using attributes (cfg and cf_attr) and the built-in cfg macro".

src/const_eval.md
```diff
@@ -37,8 +37,8 @@ to be run.
 * Index expressions, [array indexing] or [slice] with a `usize`.
 * [Range expressions].
 * [Closure expressions] which don't capture variables from the environment.
-* Built in [negation], [arithmetic, logical], [comparison] or [lazy boolean]
-  operators used on integer and floating point types, `bool` and `char`.
+* Built-in [negation], [arithmetic], [logical], [comparison] or [lazy boolean]
```
Havvy: Looks better this way, but you need to update the link refs at the bottom as well or the links will break.

src/dynamically-sized-types.md
```diff
@@ -20,7 +20,7 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
   last field, this makes the struct itself a
   <abbr title="dynamically sized type">DST</abbr>.
 
-Notably: [variables], function parameters, [const] and [static] items must be
+Notably: [variables], function parameters, [const], and [static] items must be
```
Havvy: add "items" after "[const]".

Side note: The "Notably:" part is probably from the original document and should either be changed to the Note format used elsewhere in the book or just remove the word and have this be part of the text.

src/expressions.md
```diff
@@ -265,7 +265,7 @@ a few specific cases:
 
 * Before an expression used as a [statement].
 * Elements of [array expressions], [tuple expressions], [call expressions],
-  tuple-style [struct] and [enum variant] expressions.
+  tuple-style [struct], and [enum variant] expressions.
```
Havvy: "tuple-style struct and enum variant expressions" is one thing.

src/expressions/block-expr.md
```diff
@@ -180,7 +180,7 @@ expression in the following situations:
 * Loop bodies ([`loop`], [`while`], [`while let`], and [`for`]).
 * Block expressions used as a [statement].
 * Block expressions as elements of [array expressions], [tuple expressions],
-  [call expressions], tuple-style [struct] and [enum variant] expressions.
+  [call expressions], tuple-style [struct], and [enum variant] expressions.
```
Havvy: "tuple-style struct and enum variant expressions" is one thing. Here too.

hawkinsw:
Thanks @Havvy !! I was getting delirious looking at these and I appreciate the extra set of eyes. I will address your comments first thing tomorrow and send a follow-up. Thanks again!

src/conditional-compilation.md
hawkinsw: Are cfg and cf_attr the only attributes that can do conditional compilation? If so, I think that it would read well like:

"using the attributes cfg and cf_attr and the built-in cfg macro."

I read it (without the "the") as a non-exhaustive list of example of attributes that enable conditional compilation. I am definitely not trying to nittle, I am just trying to make sure that I understand the sentence. 

Thank you!

src/dynamically-sized-types.md
hawkinsw: > add "items" after "[const]".
> 

Got it!

> Side note: The "Notably:" part is probably from the original document and should either be changed to the Note format used elsewhere in the book or just remove the word and have this be part of the text.

I agree but I can't discern a consistent format methodology for `Note`s. I see they are sometimes in code blocks, sometimes `***Note***` and sometimes `<strong>`. Just let me know what we prefer and I will make it comply!

Thanks :-)

Havvy: Notes shouldn't be in code blocks. Should be `***`, although `<strong>` is equivalent.

src/conditional-compilation.md
Havvy: The `the` there definitely helps. I also felt like something was missing, but couldn't think of what specifically.

hawkinsw: Perfect! Thanks!

src/expressions.md
```diff
@@ -264,8 +264,8 @@ exist in `core::ops` and `core::cmp` with the same names.
 a few specific cases:
 
 * Before an expression used as a [statement].
-* Elements of [array expressions], [tuple expressions], [call expressions],
-  tuple-style [struct] and [enum variant] expressions.
+* Elements of [array expressions], [tuple expressions], and [call expressions]
```
Havvy:
```suggestion
* Elements of [array expressions], [tuple expressions], [call expressions],
```

src/expressions/block-expr.md
```diff
@@ -180,7 +180,7 @@ expression in the following situations:
 * Loop bodies ([`loop`], [`while`], [`while let`], and [`for`]).
 * Block expressions used as a [statement].
 * Block expressions as elements of [array expressions], [tuple expressions],
-  [call expressions], tuple-style [struct] and [enum variant] expressions.
+  and [call expressions] and tuple-style [struct] and [enum variant] expressions.
```
Havvy:
```suggestion
  [call expressions], and tuple-style [struct] and [enum variant] expressions.
```

src/dynamically-sized-types.md
```diff
@@ -20,7 +20,7 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
   last field, this makes the struct itself a
   <abbr title="dynamically sized type">DST</abbr>.
 
-Notably: [variables], function parameters, [const] and [static] items must be
+***Note***: [variables], function parameters, [const] items, and [static] items must be
```
Havvy: Only need asterisks, and I forgot to mention the blockquote. Introduction.md is the source of truth here.

```suggestion
> **Note**: [variables], function parameters, [const] items, and [static] items must be
```

src/items/associated-items.md
```diff
@@ -133,7 +133,7 @@ Shorthand             | Equivalent
 `&'lifetime self`     | `self: &'lifetime Self`
 `&'lifetime mut self` | `self: &'lifetime mut Self`
 
-> Note: Lifetimes can be and usually are elided with this shorthand.
+> Note: Lifetimes can be, and usually are, elided with this shorthand.
```
Havvy: While we're correcting this line, here's another correction.

```suggestion
> **Note**: Lifetimes can be, and usually are, elided with this shorthand.
```

src/expressions.md
hawkinsw: For both of these, I waffled on writing it the way I suggested and the way you suggested. I think that either read perfectly okay and appreciate your feedback! Thanks!!

hawkinsw:
Thanks @Havvy for all the feedback on this. I am looking forward to digging in to other issues if/when we have this resolved! If you had time, I'd love some suggestions on where to look for opportunities.

Havvy:
💟 Thanks for going through the back and forth on this. The open issues list has quite a few things worth checking out, though many require doing research into how the language works. Feel free to work on anything you want.

hawkinsw:
> heart_decoration Thanks for going through the back and forth on this. The open issues list has quite a few things worth checking out, though many require doing research into how the language works. 

Good, because this is exactly what I want! I am a PL nerd so this stuff really excites me! Thanks again for working with me on this!

PR #676, opened 2019-09-12
document behavior of \r\n in string literals

matklad:


matklad:
@ehuss yeah, that makes sense. I'm not sure I'll manage to send a follow-up PR, so here's an unedited summary of things:

> does all source text unconditionally translate CRLF to LF?

Yes, with only two possible exceptions of `include_bytes` and `include_str`. 

This in general is not observable, except for literals and doc comments. 

>how exactly bare CRs are handled.

They are not translated, `\r\r\n` turns into `\r\n`, and the lexer treats *only* `\n` as a new line. However, we treat isolated \r in a string literal as an error. Argubly, the latter should be downgraded to a warning, but also warn about other control characters: https://github.com/rust-lang/rust/issues/63682

PR #685, opened 2019-09-26
Remove the warning that appears on every page.

### LINKED ISSUE #681: Hide warning about incompleteness
The red banner with "For now, this reference is a best-effort document..." scares me a lot. Would you please add a button "I got it" which hides the banner and sets a cookie that prevents its popup again?

ehuss:
I personally don't think this warning is necessary, and it seems a little scary.  I also think there is sufficient notice on the Introduction page that this is incomplete and non-normative.  Additionally, there is now a small link in the upper-right corner of every page to file github issues.  

Of course this is up to @Havvy who added this in #122.  Do you think the reference has come along enough in the past 2 years to remove this?

Closes #681


Havvy:
It's probably fine to remove the warning. I don't know if everything has been vetted so far, but most of it has.

r? @Centril

PR #686, opened 2019-09-29
Add examples to clarify the casting rules

emilk:
It think some illustrating examples will really help to convey what it means e.g. for an integer cast to truncate.


ehuss:
I like the examples in demonstrating some of the rules.  However, I personally think interleaving them makes it a bit harder to read the rules.  Maybe consolidate the examples into a single code block at the bottom of the section?  And place a comment before each group, like `// Same size, no-op conversions.`  Also, I would shy away from using `allow(overflowing_literals)`, can those examples use non-overflowing literals?

Overall I like the idea.

ehuss (approved):
Thanks!

Issue #487, opened 2018-12-27
Document `uniform_paths`


petrochenkov:
This is a documentation issue for https://github.com/rust-lang/rust/pull/56759

estebank:
Some information around 2018 edition has been added to https://github.com/rust-lang/reference/blob/master/src/items/use-declarations.md. Is the current state of the documentation enough to close this ticket @petrochenkov?

ehuss:
We just undocumented it (#665), because it was wrong in several ways.  There is an open PR #697 to update it, but I think the PR is still incomplete, and doesn't really address the underlying issue.  Personally, I don't think it can be properly documented without documenting all of name resolution, which is a huge project.

bstrie:
One subtask of this is to document the limitations of type aliases WRT associated items (including enum variants, which are secretly associated items):
```rust
mod my_mod {
    pub enum MyEnum {
        MyVariant
    }
    pub type TypeAlias = MyEnum;
}
use my_mod::MyEnum; // OK
use my_mod::MyEnum::MyVariant; // OK
use my_mod::TypeAlias; // OK
use my_mod::TypeAlias::MyVariant; // Doesn't work
let _ = my_mod::TypeAlias::MyVariant; // OK
```

Issue #291, opened 2018-04-03
Path and module changes (RFC 2126)


matthewjasper:
See rust-lang/rust#44660.

- [ ] Paths starting with `crate`
- [ ] Not needing `extern crate`
- [ ] `crate` visibility modifier
- [ ] Not using `mod.rs`
- [ ] Any edition lints
- [ ] Update existing examples

ehuss:
I have a branch for the `extern crate` changes, but it is blocked waiting on #415 (I either need to link to or write something that explains the extern prelude).  `crate` visibility is not stable (rust-lang/rust#53120) so I don't think that's relevant here.  I opened PR #450 for `mod.rs` changes.

I'm not sure what is meant by "edition lints".  My `extern crate` changes note that it is unidiomatic in 2018, but I'm not sure what other lints are relevant.  Also, I was under the impression lints aren't generally documented here and deferred to the rustc user guide.  Are there specific `rust_2018_idioms` you want to call out?

ehuss:
Regarding updating the examples, AFAIK the mdbook playpen links don't support changing the edition, so those links will always break.  I can't find this being tracked or mentioned anywhere.  I imagine they aren't used very often, but I think it should be tracked somewhere (should I file an mdbook issue?).

PR #697, opened 2019-10-18
Document `uniform_paths`

mattheww:
This updates use-declarations.md to describe the uniform_paths behaviour (requested in #487).

It also reorders things so that the main text and example describe the 2018 edition semantics, and the "Edition Differences" section says how the 2015 edition differs.

See also #291 "Path and module changes".

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -82,92 +82,111 @@ In this example, the module `quux` re-exports two public names defined in
 
 ## `use` Paths
 
-Paths in `use` items must start with an identifier or one of the [path
-qualifiers] `crate`, `self`, `super`, or `::`.
+A path in a `use` declaration must start with an identifier or one of the
+[path qualifiers] `crate`, `self`, `super`, or `::`.
 
-`crate` refers to the current crate. `self` refers to the current module.
-`super` refers to the parent module. `::` refers to the [extern prelude] (so
-it requires a crate name to follow).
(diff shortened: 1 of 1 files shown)


src/items/use-declarations.md
```diff
@@ -82,31 +82,44 @@ In this example, the module `quux` re-exports two public names defined in
...
 
-An example of what will and will not work for `use` items:
-<!-- Note: This example works as-is in either 2015 or 2018. -->
+`crate` refers to the current crate. `self` refers to the current module.
+`super` refers to the parent module. `::` refers to the [extern prelude] (so
+it requires a crate name to follow).
 
-```rust
+If the path starts with an identifier, the identifier is resolved in the same
+way as for a path in an expression (in particular, it can be a crate name or
```
petrochenkov: "Expression" here is a bit too specific, this is true for any non-import path.

petrochenkov: >If the path starts with an identifier

And if the path doesn't start with an identifier, the behavior is also the same.
The text leaves the non-identifier case up in the air though and never talks about it below.

mattheww:
I've pushed a new proposed version of the text.

I've reduced the text which describes resolution behaviour which isn't specific to `use` paths.

I've made it more explicit about which pieces of text describe which editions.

mattheww:
Closing: #1052 covers this properly

PR #698, opened 2019-10-19
Make Rust 2018 primary and use "Edition Differences" for Rust 2015

mattheww:
At present, some "Edition Differences" sections describe Rust 2018 additions to Rust 2015, some describe Rust 2015 limitations, and some are a mixture.

These commits attempt to make things consistent by putting the Rust 2018 description in the main text and information about Rust 2015 in the "Edition Differences" sections.

I've updated all the sections I could find, except for use-declarations.md
which needs more work (see #697).

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -171,21 +171,12 @@ The [trait implementation] must also begin with the `unsafe` keyword.
 
 ## Parameter patterns
 
-Function or method declarations without a body only allow [IDENTIFIER] or
-`_` [wild card][WildcardPattern] patterns. `mut` [IDENTIFIER] is currently
-allowed, but it is deprecated and will become a hard error in the future.
-<!-- https://github.com/rust-lang/rust/issues/35203 -->
+In function and method declarations, the parameter pattern is required.
 
-In the 2015 edition, the pattern for a trait function or method parameter is
-optional:
+If the function or method has a body, any irrefutable pattern is allowed.
 
-```rust
-trait T {
-    fn f(i32);  // Parameter identifiers are not required.
-}
-```
-
-The kinds of patterns for parameters is limited to one of the following:
+If the function or method has no body, the pattern must have one of the
+following forms:
 
 * [IDENTIFIER]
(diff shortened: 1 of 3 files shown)


mattheww:
As discussed on Discord, I've prepared some different forms of this patch series.

 - [2019-10_rust2018_primary]: has the 2018 edition in the main text (as in this PR), and the 2015 edition in "Edition differences"

 - [2019-10_rust2015_primary]: has the 2015 edition in the main text, and the 2018 edition in "Edition differences"

 - [2019-10_common_primary]: puts everything that varies between editions in "Edition differences", mostly leaving the main text describing what's true in both editions.

I've included changes to `paths.md` and `use-declarations.md` on those branches: those are some of the largest differences between the editions, so I think it makes sense to let people see what they might look like in the different arrangements.

But those commits are changing the text as well as rearranging it, so I don't think they're really part of this PR.

[2019-10_rust2018_primary]: https://github.com/mattheww/reference/commits/2019-10_rust2018_primary
[2019-10_rust2015_primary]: https://github.com/mattheww/reference/commits/2019-10_rust2015_primary
[2019-10_common_primary]: https://github.com/mattheww/reference/commits/2019-10_common_primary

ehuss:
The intent was to have something like the "common primary" option, though there are some places where it was a little inconsistent.  I think it will be difficult to have complete consistency, since some edition differences are large (like async/await, which just have brief callouts), so I suspect there will be a few special cases.

I'm not entirely happy with the "common" approach since it is a little disjoint. Some of the large callout blocks look a little strange to me.  But the other approaches have their own drawbacks (such as making it less clear where the edition differences are).

If anything, I would vote for some of the "common" changes to make it more consistent.  But I'd like to hear what other people think!

joshtriplett:
I like the idea of the reference always describing the current edition of Rust, and separately stating how past editions differ.

mattheww:
Closing: I don't think there's anything of value left here.

The Reference is now essentially in "Common primary" style (and this is documented in STYLE.md).

PR #703, opened 2019-10-27
Update coherence and orphan rules documentation to match RFC 2451

ohadravid:
As part of stabilizing [RFC 2451](https://rust-lang.github.io/rfcs/2451-re-rebalancing-coherence.html) (re-rebalance coherence) in https://github.com/rust-lang/rust/pull/65879, I wanted to update the reference to match it.

I copied the relevant parts of the RFC, with a slight rewording by @nikomatsakis ([comment](https://github.com/rust-lang/rust/issues/63599#issuecomment-522032863)).


src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
+## What is coherence and why do we care?
+
+Coherence means that for any given trait and type, there is one specific
```
Centril: Can you try to remove the motivational stuff from this text? The reference's only aim is to give a definition / specification of behavior, not explain why things matter and why we care.

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
...
+
+Coherence means that for any given trait and type, there is one specific 
+implementation that applies. This is important for Rust to be easy to reason 
+about. When you write `<Foo as Bar>::trait_method`, the compiler needs to know
+what actual implementation to use.
+
+In languages without coherence, the compiler has to have some way to choose
+which implementation to use when multiple implementations could apply. Scala
+does this by having complex scope resolution rules for "implicit" parameters.
+Haskell (when a discouraged flag is enabled) does this by picking an impl
```
Centril: This entire paragraph is irrelevant. Haskell (mainly) and other languages are only interesting for the purposes of "influences".

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
...
+
+- `Trait` is a local trait
+- All of
+  - At least one of the types `T0..=Tn` must be a local type. Let `Ti` be the
+    first such type.
+  - No uncovered type parameters `P1..=Pn` may appear in `T0..Ti` (excluding
+    `Ti`)
+
+We only restrict the appearance of *uncovered* type parameters. Once again, it is
+important to note that for the purposes of coherence, `#[fundamental]` types are
```
Centril: `#[funtamental]` is not part of the stable language definition. From the POV of the reference, some types and traits are just fundamental. We can specify with a note how rustc achieves this. What is part of the language definition is the actual set of fundamental types and traits. These should be itemized here.

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
...
+`T` is uncovered, but the `T` in `Vec<T>` is covered. This is only relevant for
+type parameters.
+
+Blanket Impl: Any implementation where a type appears uncovered. `impl<T> Foo
+for T`, `impl<T> Bar<T> for T`, `impl<T> Bar<Vec<T>> for T`, and `impl<T> Bar<T>
+for Vec<T>` are considered blanket impls. However, `impl<T> Bar<Vec<T>> for
+Vec<T>` is not a blanket impl, as all instances of `T` which appear in this impl
+are covered by `Vec`.
+
+Fundamental Type: A type for which you cannot add a blanket impl backwards
```
Centril: The social aspect of fundamental types is not important in the reference so the first sentence can be removed and the other ones restructured a bit.

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
...
+compatibly. This includes `&`, `&mut`, and `Box`. Any time a type `T` is
+considered local, `&T`, `&mut T`, and `Box<T>` are also considered local.
+Fundamental types cannot cover other types. Any time the term "covered type" is
+used, the `T` in `&T`, `&mut T`, and `Box<T>` is not considered covered.
```
Centril: Seems like some of these terms should be added to the glossary?

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
...
+about. When you write `<Foo as Bar>::trait_method`, the compiler needs to know
+what actual implementation to use.
+
+In languages without coherence, the compiler has to have some way to choose
+which implementation to use when multiple implementations could apply. Scala
+does this by having complex scope resolution rules for "implicit" parameters.
+Haskell (when a discouraged flag is enabled) does this by picking an impl
+arbitrarily.
+
+Rust's solution is to enforce that there is only one impl to choose from at all.
```
Centril: Most of this paragraph is not important either.

src/items/coherence.md
```diff
@@ -0,0 +1,79 @@
+## What is coherence and why do we care?
```
Centril: I think this text should mainly explain that overlap is prevented (and the rules for that) and that orphans are not allowed (and the rules for that).

Question to @nikomatsakis: is the orphan rule actually part of preventing overlap in the compiler?

ohadravid:
Hi @Centril!

I made the changes you requested, and since all the terms seemed relevant I moved all of them to the glossary, making the coherence explanation much shorter so I moved it back to where it was. If you think this is too much for the glossary I can also move them back/move only some of them/etc.

You question to Niko about the orphan rule/overlap is still relevant, because the old text is not too specific there either (I changed it back to the way it was).

One question about defining fundamental types: I looked for the attribute, and saw that `Pin` was added since the RFC (So I added it to the list).
`Generator` is also there, but I didn't add it because it unstable.
So should there be a sentence like "the lang may define additional types which are considered fundamental"?

nikomatsakis:
@ohadravid thanks for doing this!

src/glossary.md
```diff
@@ -30,6 +30,14 @@ items are defined in [implementations] and declared in [traits]. Only
 functions, constants, and type aliases can be associated. Contrast to a [free
 item].
 
+### Blanket Implementation
+
+Any implementation where a type appears [uncovered](#uncovered-type). `impl<T> Foo
+for T`, `impl<T> Bar<T> for T`, `impl<T> Bar<Vec<T>> for T`, and `impl<T> Bar<T>
+for Vec<T>` are considered blanket impls. However, `impl<T> Bar<Vec<T>> for
+Vec<T>` is not a blanket impl, as all instances of `T` which appear in this impl
```
Centril:
```suggestion
Vec<T>` is not a blanket impl, as all instances of `T` which appear in this `impl`
```

src/glossary.md
```diff
@@ -65,6 +73,13 @@ For example, `2 + (3 * 4)` is an expression that returns the value 14.
 An [item] that is not a member of an [implementation], such as a *free
 function* or a *free const*. Contrast to an [associated item].
 
+### Fundamental Type
+
+Includes `&`, `&mut`, `Box` and `Pin`. Any time a type `T` is
+considered [local](#local-type), `&T`, `&mut T`, `Box<T>` and `Pin<T>` are also considered local.
+Fundamental types cannot [cover](#uncovered-type) other types. Any time the term "covered type" is
+used, the `T` in `&T`, `&mut T`, `Box<T>` and `Pin<T>` is not considered covered.
```
Centril:
```suggestion
used, the `T` in `&T`, `&mut T`, `Box<T>`, and `Pin<T>` is not considered covered.
```

src/glossary.md
```diff
@@ -65,6 +73,13 @@ For example, `2 + (3 * 4)` is an expression that returns the value 14.
 An [item] that is not a member of an [implementation], such as a *free
 function* or a *free const*. Contrast to an [associated item].
 
+### Fundamental Type
+
+Includes `&`, `&mut`, `Box` and `Pin`. Any time a type `T` is
```
Centril: Some points & questions (which @nikomatsakis may help with):

- `&` and friends are not types; they are type constructors.

- Suppose we have `#[fundamental] struct W<P0, P1> { ... }`. How would `W<A0, A1>` behave with respect to locality?  In this case the addition of a second type parameter is intentional so that we can suss out the *essence* of how fundamental types behave. It seems lifetimes have no effect on locality.

- Before itemizing the list of fundamental types it would be good to give a definition of how such types behave.

- What about fundamental `trait`s?

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
 moved from. All other memory locations are assumed to be uninitialized. Only
 unsafe Rust can create such a memory without initializing it.
 
+### Local Trait
```
Centril:
```suggestion
### Local trait
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+
+A trait which was defined in the current crate. Whether a trait is
```
Centril:
```suggestion
A `trait` which was defined in the current crate. A trait definition is local
or not independent of applied type arguments. That is, given `trait Foo<T, U>`,
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+local or not has nothing to do with type parameters. Given `trait Foo<T, U>`,
```
Centril:
```suggestion
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+`Foo` is always local, regardless of the types used for `T` or `U`.
```
Centril:
```suggestion
`Foo` is always local, regardless of the types substituted for `T` and `U`.
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+
+### Local Type
+
+A struct, enum, or union which was defined in the current crate.
```
Centril:
```suggestion
A `struct`, `enum`, or `union` which was defined in the current crate.
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+This is not affected by type parameters. `struct Foo` is considered local, but
```
Centril:
```suggestion
This is not affected by applied type arguments. `struct Foo` is considered local, but
```

src/glossary.md
```diff
@@ -87,6 +102,19 @@ A variable is initialized if it has been assigned a value and hasn't since been
...
+`Vec<Foo>` is not. `LocalType<ForeignType>` is local. Type aliases and trait
+aliases do not affect locality.
```
Centril: Trait aliases do not exist on stable.

src/glossary.md
```diff
@@ -158,6 +186,12 @@ It allows a type to make certain promises about its behavior.
 
 Generic functions and generic structs can use traits to constrain, or bound, the types they accept.
 
+### Uncovered Type
+
+A type which does not appear as a parameter to another type. For example,
```
Centril:
```suggestion
A type which does not appear as an argument to another type. For example,
```

src/glossary.md
```diff
@@ -158,6 +186,12 @@ It allows a type to make certain promises about its behavior.
...
+`T` is uncovered, but the `T` in `Vec<T>` is covered. This is only relevant for
+type parameters.
```
Centril:
```suggestion
type arguments.
```

src/items/implementations.md
```diff
@@ -152,29 +152,31 @@ impl Shape for Circle {
...
 traits the implementation is for, the implementations can be instantiated with
 the same type. <!-- This is probably wrong? Source: No two implementations can
 be instantiable with the same set of types for the input type parameters. -->
 
-The `Orphan Check` states that every trait implementation must meet either of
-the following conditions:
+#### Orphan rules
 
-1.  The trait being implemented is defined in the same crate.
+Given `impl<P1..=Pn> Trait<T1..=Tn> for T0`, an impl is valid only if at
```
Centril:
```suggestion
Given `impl<P1..=Pn> Trait<T1..=Tn> for T0`, an `impl` is valid only if at
```

src/items/implementations.md
```diff
@@ -152,29 +152,31 @@ impl Shape for Circle {
...
-    meet the following grammar, where `C` is a nominal type defined
-    within the containing crate:
+- `Trait` is a [local trait]
+- All of
+  - At least one of the types `T0..=Tn` must be a [local type]. Let `Ti` be the
+    first such type.
+  - No [uncovered type] parameters `P1..=Pn` may appear in `T0..Ti` (excluding
+    `Ti`)
+
+We only restrict the appearance of *uncovered* type parameters. Once again, it is
```
Centril:
```suggestion
Only the appearance of *uncovered* type parameters is restricted.
```

src/items/implementations.md
```diff
@@ -152,29 +152,31 @@ impl Shape for Circle {
...
+important to note that for the purposes of coherence, [fundamental types] are
```
Centril:
```suggestion
Note that for the purposes of coherence, [fundamental types] are
```

Centril:
> If you think this is too much for the glossary I can also move them back/move only some of them/etc.

No it's good. :)

>  So should there be a sentence like "the lang may define additional types which are considered fundamental"?

I think we sorta want to give a description of behavior for any arbitrary fundamental type that is semi-syntactic in that it talks about type arguments and type constructors and how those relate wrt. locality and such. Then we can itemize a list of fundamental type constructors.

ohadravid:
I applied all your edits @Centril , also edited some more title case -> sentence case and such.

I changed the fundamental types part to use the type constructor wording, but I leave it to Niko to define what they are :)

I did use the order of: what are fundamentals, itemized list, locality.
Explaining it with locality first didn't sound right (you have to use at least `&T` and `Box<T>` when explaining locality, because of the different syntax, and than it seems like `&mut T` and `Pin` are somehow different..) but that's all more like a placeholder until we have the actual description.

ohadravid:
@Centril @nikomatsakis I used Niko's definitions from [RFC 1023](https://github.com/rust-lang/rfcs/blob/master/text/1023-rebalancing-coherence.md#errors-from-cargo-and-the-fundamental-attribute).
They seem to fit the reference, but I wasn't too sure if more explanation is needed. I also added the definition of fundamental traits for completeness.

Centril:
Filed https://github.com/rust-lang/reference/issues/714 for any follow-ups to the more formal spec.

Centril:
Thanks!

Issue #434, opened 2018-09-25
Keyword - Usage Table


Havvy:
Just like we have the punctuation table that has links to where it's used, it'd be nice to have a table for keywords and where the keywords are used.

ehuss:
Would it be sufficient to have the keywords in `keywords.md` link to where it's used?  

In relation to #211, I'm thinking whether or not there should be a single Appendix with all "weird things" just like the first edition book, or if it is sufficient to have it spread across multiple pages (`keywords.md`, `tokens.md`, etc.).  I kinda like the idea of a single page, but would want to avoid too much duplication.

ConveyorBusBuilder:
I think there should definitely be the keywords on their own at least to start just to get it out there (I'm worried that it may turn into another reference where it's not fully documented for a while.) but a sort of book of out-there Rust knowledge, a sort of 'liber arcanum' is absolutely something I would love to see. If I knew a bit about using mdBook, I'd start getting on documenting the large amount of obscure knowledge I know.

If there was a 'Doc Arcanum,' should there be a straight-up list of facts that can help you or would that be better suited for a sort of 'useful rust patterns' doc. Does the latter exist?

ConveyorBusBuilder:
OK funny enough with a comment from ehuss on a different issue, I learned a thing that I should have known long before: keywords are documented in the std docs. There's a section under Primitive Types, Modules, and Macros called Keywords and if you click on it it gives you a massive list of keywords and each entry there (when you click on one) gives some documentation on what it is and how it's used. Here's a link to the list: [https://doc.rust-lang.org/std/#keywords](https://doc.rust-lang.org/std/#keywords). Is this what you were thinking of,  @Havvy?

PR #708, opened 2019-11-02
Create links from keywords list to explanations

pflanze:
Note that I haven't built the docs (I'm out of diskspace building the mdbook tool right now, and short on time). The relative links should work both for the website (although the targets are in /stable/std not the apparently default toplevel /std), and for local docs as used by "rustup docs".

I'm adding these links especially since I was using the docs on the website without js enabled thus had a really hard time finding the keywords. "Just use the search bar" is easy to say if you don't see any :). In any case I think there should be a list linking to all the keyword docs somewhere and this page seems predestined for that.

### DIFF
--- a/src/keywords.md
+++ b/src/keywords.md
@@ -95,8 +95,8 @@ The following keywords are reserved beginning in the 2018 edition.
 These keywords have special meaning only in certain contexts. For example, it
 is possible to declare a variable or method with the name `union`.
 
-* `union` is used to declare a [union] and is only a keyword when used in a
-  union declaration.
+* [`union`](../std/keyword.union.html) is used to declare a [union]
+  and is only a keyword when used in a union declaration.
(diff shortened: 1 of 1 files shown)


ehuss:
cc #434

I would be reluctant to add links to these std docs because they are not normative, and not necessarily kept in sync.  I think I would prefer internal links, similar to how the [Punctuation](https://doc.rust-lang.org/nightly/reference/tokens.html#punctuation) section is.

Also, the grammar/lexer definitions don't link to explanations, but instead usually link to other grammar productions, leaving the leaf nodes unlinked.  This would make it inconsistent with the rest of the reference.  Also, adding these links may make it more difficult to finish the automatic grammar appendix generator script.

Centril:
Closing with @ehuss's rationale and since the opposite case has not been argued for some time.

PR #712, opened 2019-11-05
abi.md: clarify #[used] and linking

JustAPerson:
I originally interpreted this paragraph to mean `rustc` would also instruct the linker to keep the symbol. [This is not the case](https://github.com/rust-lang/rfcs/blob/master/text/2386-used.md#guide-level-explanation).


nagisa:
This is correct and should be merged. Let me try all the invocations:

r+
@bors r+
bors r+
r=nagisa

and hope any of them work ;)

PR #716, opened 2019-11-11
Helper attributes in attribute macros

Extrawurst:
This little gotcha took me a while to figure out, maybe others will benefit from this addition swell

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -273,6 +273,11 @@ fn invoke4() {}
 // out: item: "fn invoke4() {}"
 ```
 
+#### Attribute macro helper attributes
+
+Unlike for [Derive macros] helper attributes do not have to be marked explicitly.
+Since attribute macros can completely discard the input token stream filtering out custom `helper` attributes is entirely the job of the attribute macro.
+
 [Attribute macros]: #attribute-macros
 [Cargo's build scripts]: ../cargo/reference/build-scripts.html
 [Derive macros]: #derive-macros


src/procedural-macros.md
```diff
@@ -273,6 +273,11 @@ fn invoke4() {}
 // out: item: "fn invoke4() {}"
 ```
 
+#### Attribute macro helper attributes
+
+Unlike for [Derive macros] helper attributes do not have to be marked explicitly.
+Since attribute macros can completely discard the input token stream filtering out custom `helper` attributes is entirely the job of the attribute macro.
```
lzutao:
```suggestion
Since attribute macros can completely discard the input token stream filtering
out custom `helper` attributes is entirely the job of the attribute macro.
```

src/procedural-macros.md
```diff
@@ -273,6 +273,11 @@ fn invoke4() {}
 // out: item: "fn invoke4() {}"
 ```
 
+#### Attribute macro helper attributes
```
Centril: I do not understand this addition; until we have something like https://github.com/rust-lang/rust/issues/65823 there is no such thing in the language definition as "attribute macro helper attributes". (N.B. the reference is not a user guide -- it is an incomplete definition of the language).

Extrawurst: Thanks for pointing me to that issue, its basically the same reason for my edit: new comers might be confused how to get the same behaviour like the helper attributes in a full fledged proc macro attribute, not realising that it is the macro's job to actually strip those out (or what ever else for that matter) - I get that it is weird to put something in the docs that looks like a user-guide thing but clearly I am not the only person stumbling over this...

Centril: @Extrawurst I think adding this sort of documentation would be appropriate for e.g. a book on procedural macros or some other sort of tutorial (cc @dtolnay for possible locations). But people stumbling over things is not a good reason to add help notes to the reference, which is not teaching material.

Extrawurst: fair enough

PR #717, opened 2019-11-14
Document try expressions

Kampfkarren:
Tracking issue: https://github.com/rust-lang/rust/issues/31436

I am not sure if I am expected to fix the error relating to try blocks being unstable, part of the stabilizing process is to write documentation for the reference? Or is stabilization and documentation meant to happen at the same time? I'm unsure...


Kampfkarren:
> The master branch of the reference corresponds to what is stable on the master branch ("nightly") of rust-lang/rust. If you want to write about Rust in the future, you want the Unstable book.

Oops, sorry!

PR #724, opened 2019-11-30
Update for nested receivers.

ehuss:
Stabilized in https://github.com/rust-lang/rust/pull/64325.


ehuss:
cc @cramertj

I don't really know if this is worded correctly, but I figured I'd get the conversation started. I also wanted to double check if people are satisfied with how the [object safety](https://doc.rust-lang.org/nightly/reference/items/traits.html#object-safety) rules are worded.

src/items/associated-items.md
```diff
@@ -109,13 +109,15 @@ following types:
 - [`Arc<Self>`]
 - [`Pin<P>`] where `P` is one of the above types except `Self`.
 
-The `Self` term can be replaced with the type being implemented.
+The `Self` term can be replaced with the type being implemented, including
+type aliases for the type, or any nested combination of the above types.
```
Centril: Maybe add a note re. object safety (when it's not)?

cramertj:
Seems reasonable enough to me!

src/items/associated-items.md
ehuss: I added more info in the object safety section.  The intent is that it can all be rewritten when Receiver is stabilized.

I also added some examples. Do you think they are useful? I'm concerned they take up a lot of space.  The rules are fairly concise, but I find it hard to visualize how they translate to code without some examples.

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+
+```rust,compile_fail
+// These are object-safe, but cannot be dispatched on a trait object.
+trait NonDispatchable {
+    // Non-methods cannot be dispatched.
+    fn foo() where Self: Sized {}
+    // Self type isn't known until runtime.
+    fn returns(&self) -> Self where Self: Sized;
+    // `other` may be a different concrete type of the receiver.
+    fn param(&self, other: Self) where Self: Sized {}
```
ehuss: I can't find anything in the reference that mentions that this must be Sized for all traits. Am I missing something?

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+
+```rust,compile_fail
+// Not object safe if `Self` is a type parameter.
+trait Super<A> {}
+trait WithSelf: Super<Self> where Self: Sized {}
+
+struct S;
+impl<A> Super<A> for S {}
+impl WithSelf for S {}
+let obj: Box<dyn WithSelf> = Box::new(S); // ERROR: cannot use `Self` type parameter
```
ehuss: I don't see how this falls out of the rules listed above.  Is this an extra rule, or some combination I am not understanding?

src/items/associated-items.md
Centril: It's probably fine :)

Centril:
@cramertj can you take another look at the changes and approve the PR?

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
 
+When there isn't a `Self: Sized` bound on a method, the type of a method
+receiver must be one of the following types:
+
+* `&Self`
+* `&mut Self`
+* [`Box<Self>`]
+* [`Rc<Self>`]
+* [`Arc<Self>`]
+* [`Pin<P>`] where `P` is one of the object safe receiver types
```
cramertj: This is the first time the phrase  "object safe receiver types" has been mentioned. Consider possibly "where `P` is one of the types above" or similar.

ehuss: Good point, updated!

ehuss:
Ping @Centril I think this should be ready to go.

src/items/associated-items.md
Centril: Having another look at this now, I wonder if we could rephrase this section more formally as (with formatting to be fixed):

> If the type of the self parameter is specified, it is limited to *semantic* types generated by the following grammar (where `'lt` denotes some arbitrary lifetime):
>
> ```
> P = &'lt S | &'lt mut S | Box<S> | Rc<S> | Arc<S> | Pin<P>
> S = Self | P
> ```
>
> The `Self` terminal in this grammar is the semantic `Self` type and can be replaced with the type being implemented, including type aliases or associated type projections for the type.

src/items/associated-items.md
```diff
@@ -109,13 +109,15 @@ following types:
...
 
 ```rust
 # use std::rc::Rc;
 # use std::sync::Arc;
 # use std::pin::Pin;
 struct Example;
+type Alias = Example;
```
Centril:
```suggestion
type Alias = Example;
trait Trait { type Output; }
impl Trait for Example { type Output = Example; }
```

src/items/associated-items.md
```diff
@@ -126,6 +128,7 @@ impl Example {
     fn by_pin(self: Pin<&Self>) {}
     fn explicit_type(self: Arc<Example>) {}
     fn with_lifetime<'a>(self: &'a Self) {}
+    fn nested<'a>(self: &mut &'a Arc<Rc<Box<Alias>>>) {}
```
Centril: Additionally:
```suggestion
    fn nested<'a>(self: &mut &'a Arc<Rc<Box<Alias>>>) {}
    fn via_projection(self: <Example as Trait>::Output) {}
```

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+// These are object-safe, but cannot be dispatched on a trait object.
+trait NonDispatchable {
+    // Non-methods cannot be dispatched.
+    fn foo() where Self: Sized {}
+    // Self type isn't known until runtime.
+    fn returns(&self) -> Self where Self: Sized;
+    // `other` may be a different concrete type of the receiver.
+    fn param(&self, other: Self) where Self: Sized {}
+    // Generics are not compatible with vtables.
+    // Alternate solution is to use a trait object instead.
```
Centril: This line feels like user advice?

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+    fn with_lifetime<'a>(self: &'a Self) {}
+    fn nested_pin(self: Pin<Arc<Self>>) {}
+}
+# struct S;
+# impl TraitMethods for S {}
+# let t: Box<dyn TraitMethods> = Box::new(S);
+```
+
+```rust,compile_fail
+// These are object-safe, but cannot be dispatched on a trait object.
```
Centril: Does "these" refer to the trait or the methods? The trait is object safe, but the methods would not be if not for `Sized`, which makes the trait object safe since you cannot dispatch to the methods.

src/items/traits.md
Centril: Not sure I follow here.

src/items/traits.md
```diff
@@ -95,6 +95,99 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+// Self: Sized traits are not object-safe.
+trait TraitWithSize where Self: Sized {}
+
+struct S;
+impl TraitWithSize for S {}
+let obj: Box<dyn TraitWithSize> = Box::new(S); // ERROR
+```
+
+```rust,compile_fail
+// Not object safe if `Self` is a type parameter.
```
Centril: Not sure what this means -- I assume you are referring to the use of `Self` as a type *argument* in `Super<Self>`? `Self` is sort of an implicit extra type parameter in each `trait` definition.

src/items/traits.md
Centril: cc @nikomatsakis

src/items/traits.md
ehuss: Why is this an error? ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7af9cb55e569eb8d0e7f1c2a4f111ffd))

```rust
trait UnsizedSelf {
    fn param(&self, other: Self){}
}
```

Is there some rule listed somewhere that implies `other` must be sized?  I can't find it.

src/items/traits.md
ehuss: Reworded.

src/items/traits.md
ehuss: Yes, oops, an argument.

src/items/traits.md
Centril: Same result with:
```rust
trait UnsizedSelf { fn param(other: Self) {} }
```
or:
```rust
fn param<T: ?Sized>(other: T) {}
```
This is the general restriction that parameters, or local variables more generally, must have a known size.

Remember that traits are not `: Sized` by default.

By saying `where Self: Sized` you make `Self` have a known size for that function (meaning that it also is excluded from the VTable).

ehuss: More specifically, I was looking for a statement that says `Self` is `?Sized` for traits.  I guess the [Dynamically Sized Types](https://doc.rust-lang.org/nightly/reference/dynamically-sized-types.html#dynamically-sized-types) chapter covers that with the sentence "Unlike type parameters `Self: ?Sized` by default in trait definitions."  OK, nevermind, I just had a hard time finding that (I was expecting it in the traits chapter).

Centril: Ah, I see. Well, glad you found it. =P

src/items/traits.md
Centril: Seems like Niko hasn't had time to address this. Let's file an issue for now and then Niko can follow up on that one?

ehuss: Yea, I wasn't expecting to fix it in this PR, I was just a little confused.

Filed #756.

nikomatsakis:
Sorry y'all, I had this on my list to review .. but I'll check out the associated issue

src/items/associated-items.md
```diff
@@ -98,24 +98,28 @@ Associated functions whose first parameter is named `self` are called *methods*
 and may be invoked using the [method call operator], for example, `x.foo()`, as
 well as the usual function call notation.
 
-If the type of the `self` parameter is specified, it is limited to one of the
-following types:
+If the type of the `self` parameter is specified, it is limited to semantic
```
RalfJung: What's a "semantic type"? In the PL research community this has a very specific meaning (which you can read about e.g. in [this paper](https://plv.mpi-sws.org/rustbelt/popl18/paper.pdf)), but I doubt that is what you mean.

Centril: Indeed, it's not the "semantic type system" notion in PL research. Rather, this means to say that the `Self` here is not meant literally (as in those characters making up `Self`). So it could be a type alias to e.g. `Rc<TheImplementingType>`. Ideas for a better wording?

RalfJung: "unfolded type"? "fully expanded type"? "resolved type"?

Centril: "resolved type" seems good.

PR #730, opened 2019-12-14
introduce feature gate into never-type test

nikomatsakis:
Given that we are rolling back stabilization of the never type, this example needs a feature gate. The test was added only recently, by @ehuss -- I think that ordinarily we don't document unstable features, but given that this section of the book seems to be older, and that `!` can appear in some places, and that we hope to get `!` stabilized very soon, I opted not to remove it.

This needs to land semi-urgently so that https://github.com/rust-lang/rust/pull/67224 can land.

### DIFF
--- a/src/types/never.md
+++ b/src/types/never.md
@@ -8,7 +8,15 @@ computations that never complete. Expressions of type `!` can be coerced into
 any other type.
 
 ```rust,should_panic
+#![feature(never_type)]
 let x: ! = panic!();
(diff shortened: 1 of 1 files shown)


ehuss:
I thought this would fail the stable-check, but it looks like the stable-check doesn't recurse into subdirectories. 😢   I will try to fix that some point in the future.

Sorry this caused a hiccup!

PR #732, opened 2019-12-18
Remove `extern` from exception list

JohnTitor:
`r#extern` is now accepted
cf. https://github.com/rust-lang/rust/pull/57572


matthewjasper:
Thanks

PR #738, opened 2020-01-18
Add section on semver derived from rust-lang/rfcs#1105

guswynn:
As far as I can tell, rust-lang/rfcs#1105 are the *de-facto* guidelines that crate maintainers should use to guide their use of semver for this crates.

I have found it confusing in the past that these guidelines don't exist in any documentation outside that RFC, and discussion in rust-lang/rust#68004 led be to believe that it's worth it to add this to some official documentation.

In this PR, I largely **copied the RFC** into its own section in the reference, and edited it to make the language fit the reference (I could have gotten the tone wrong in some places)

The intention is that this documentation can also be extended (unlike the RFC) when new things with consequences for semver (like rust-lang/rust#68004) are landed in the language (whether or not this guide includes just nightly or just stable changes to the language is TBD, perhaps a version of this deserves to be in the unstable book).

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,4 +28,4 @@ And then generate the book:
 mdbook build
 ```
 
-The generated HTML will be in the `docs` folder.
+The generated HTML will be in the `book` folder.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -122,6 +122,8 @@
 
 - [The Rust runtime](runtime.md)
 
+- [Semver for Rust](semver.md)
+
 - [Appendices](appendices.md)
     - [Macro Follow-Set Ambiguity Formal Specification](macro-ambiguity.md)
     - [Influences](influences.md)
--- a/src/semver.md
+++ b/src/semver.md
@@ -1,0 +1,620 @@
+# Semver and Rust
+
+The merged [RFC 1105][rfc] provided a comprehensive set of guidelines for which changes to
+*stable* APIs within the standard library
+are considered breaking from a [semver][semver] perspective, and which are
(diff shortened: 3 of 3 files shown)


guswynn:
I'm somewhat new to github, am I able to add a label? I want to add T-docs

also I will fix the breakage

ehuss:
Thanks for the PR!  The reference is intended to be more of a language spec, and so it typically doesn't include guidelines of this nature.

I have been intending to include something like this in the Cargo book. It's not really clear where it should live, but I think since Cargo is the tool most concerned about semver compatibility, it seems like a decent place.  My intent is to have two chapters, one on version unification and versioning guidelines, and a second that includes this content about what a "breaking change" is.

If you'd be interested, I'd be happy if you could help out with getting it started in the Cargo repo.  I do desire to include some editing and enhancements beyond just copying the RFC (which is worded with the standard library in mind).

steveklabnik:
I'm of two minds about this; it's interesting because you're 100% right that this isn't really the language, but it is also the rules for how this specification can change. It's kind of like, meta information. That doesn't mean that I think closing is a bad idea, mind you...

guswynn:
@ehuss That makes a lot of sense! I would absolutely love to help out on getting it started on the cargo version of this pr (I assume I start here https://github.com/rust-lang/cargo/tree/master/src/doc/src/reference)

@steveklabnik yeah there are definitely diminishing returns on how many levels of meta there is a whole book for

ehuss:
> I assume I start here https://github.com/rust-lang/cargo/tree/master/src/doc/src/reference

Yep!

PR #740, opened 2020-01-25
Write documentation for the `Termination` trait

LeSeulArtichaut:
This PR adds documentation for the `Termination` trait proposed in rust-lang/rfcs#1937.
See tracking issue rust-lang/rust#43301.

### DIFF
--- a/src/special-types-and-traits.md
+++ b/src/special-types-and-traits.md
@@ -139,6 +139,29 @@ compile-time; that is, it's not a [dynamically sized type]. [Type parameters]
 are `Sized` by default. `Sized` is always implemented automatically by the
 compiler, not by [implementation items].
 
+## `Termination`
+
+Types implementing the [`Termination`] trait can be returned from `main()`.
+Those types implement logic to be converted to an integer value returned
+to the operating system when the program ends.
+
+The standard library has implementations for the following types:
+
+* [`ExitCode`] represents conventional return values, depending on the
+  operating system. This includes `SUCCESS` and `FAILURE` variants, but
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  Can you say more why you'd like to add this?  The reference doesn't generally include content on unstable features, so I'm not sure if this should be included (it already has a brief link [here](https://github.com/rust-lang/reference/blob/11e893fc1357bc688418ddf1087c2b7aa25d154d/src/crates-and-source-files.md#main-functions)).  The reference also generally doesn't include details about how the standard library is implemented, and instead focuses on just the language itself.

LeSeulArtichaut:
I thought this was part of the stabilization process. If you look at the tracking issue, the only missing step is « adjust documentation »

ehuss:
Ah, is there a proposal to stabilize it somewhere?

Generally, stabilization isn't usually gated on having the reference updated.  Since the reference only covers stable language features, it's usually best to update it during the stabilization process, or shortly afterwards.

As for this specific change, I'm not sure what the best way to document this is.  As I understand, `Termination` is not a language feature, since it is implemented solely in the standard library (using the unstable `start` lang item).  I would probably just update the note in the "Main Functions" section, and leave only the first paragraph from this PR in the special traits list.  The details of how Termination is implemented is up to the standard library, and seems like the kind of thing that is not relevant to the language, and very difficult to keep in sync.  It could then be updated or removed if/when `start` is stabilized.

LeSeulArtichaut:
Alright, closing this

PR #741, opened 2020-01-27
`enum`s can now be `#[repr(transparent)]`

Centril:
Stabilized in https://github.com/rust-lang/rust/pull/68122.

r? @ehuss 
cc @rkruppe

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -357,8 +357,8 @@ Like all ways to create undefined behavior in safe Rust, this is a bug.
 
 ### The `transparent` Representation
 
-The `transparent` representation can only be used on a [`struct`][structs] that has,
-or an [`enum`][enumerations] with a single variant which has:
+The `transparent` representation can only be used on a [`struct`][structs]
+or an [`enum`][enumerations] with a single variant that has:
 
 - a single field with non-zero size, and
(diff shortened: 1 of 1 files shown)


src/type-layout.md
```diff
@@ -332,13 +332,14 @@ Like all ways to create undefined behavior in safe Rust, this is a bug.
 
 ### The `transparent` Representation
 
-The `transparent` representation can only be used on `struct`s that have:
+The `transparent` representation can only be used on a [`struct`][structs] that has,
+or an [`enum`][enumerations] with a single variant which has:
```
ehuss: I was initially a little confused with how this is worded (it seemed a little awkward).  Instead of splitting it, maybe word it something like this?

```
The `transparent` representation can only be used on a [`struct`][structs] or
single variant [`enum`][enumerations] that has:
```

I'm guessing the intent was to avoid ambiguity of whether or not the two rules apply to both structs and enums?  I'm not terribly strong at grammar, but I would say without a comma it is not an independent clause, and should be considered together.

Centril: The idea was to avoid the implication that the attribute is placed on the variant as opposed to the `enum`. I'll update to your wording.

PR #742, opened 2020-02-03
Document #[track_caller].

anp:


ehuss:
Thanks for the PR.  What is the context for this change?  The reference only documents stable attributes, and AFAIK this is still unstable.  Will it be stabilized soon?

src/attributes/codegen.md
```diff
@@ -140,6 +140,25 @@ feature detection on the x86 platforms.
...
+An implementation of `core::intrinsics::caller_location` returns this implicit argument if called
+from a function with the attribute. Otherwise it returns the location at which the intrinsic itself
+was called.
+
+Creating a function pointer to a function with `#[track_caller]` creates a shim which passes the
+location at which the attributed function is defined. The pointer then refers to this shim,
+preventing unsoundness or the implicit ABI from leaking into the function pointer's type.
+
+In a const context, the attribute does not affect function ABI and instead informs the const
+evaluator to continue walking up the stack when probing for the topmost caller location.
```
Centril: This doesn't seem right... I did some investigation and filed https://github.com/rust-lang/rust/issues/68789.

anp:
> Thanks for the PR. What is the context for this change? The reference only documents stable attributes, and AFAIK this is still unstable. Will it be stabilized soon?

Ah -- right! I'm documenting the attribute here based on feedback in https://github.com/rust-lang/rust/issues/47809#issuecomment-572497766 while I'm working on overall feature docs for the rustc-guide. Perhaps I should hold off on this PR until closer to stabilization?

src/attributes/codegen.md
anp: Thanks for the followup, I think there's some confusion (aside from the legit bugs) here caused by me incorrectly including ABI in the definition of the attribute. From miri's perspective there isn't an observable ABI change, I'll clarify the details in a rustc-specific note as you suggested above.

Centril: Looks like I was using an outdated nightly instead of master, so there's no bug. I've filed https://github.com/rust-lang/rust/pull/68798 to ensure the correct behavior remains.

Centril:
> Perhaps I should hold off on this PR until closer to stabilization?

The PR won't be merged until this has been stabilized upstream in rustc, but feel free to polish the section meanwhile. It's certainly better to have the docs ready to go as opposed to having a bunch of work much after stabilization. ;) It's also useful for the eventual stabilization report.

src/attributes/codegen.md
```diff
@@ -140,6 +140,25 @@ feature detection on the x86 platforms.
...
+
+The `track_caller` attribute applies to function declarations and indicates to the compiler that the
+attributed function's generated code should have an implicit `&'static core::panic::Location`
+argument which is populated by its caller. If the caller of the attributed function is also
+annotated with `#[track_caller]`, then the attributed function should receive the location
+its caller received, and so on.
+
+An implementation of `core::intrinsics::caller_location` returns this implicit argument if called
+from a function with the attribute. Otherwise it returns the location at which the intrinsic itself
+was called.
```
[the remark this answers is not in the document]
anp: Coming back to this and I think it's actually most accurate to describe observing the caller with `core::panic::Location::caller`, which is a public function specified by the RFC. Sound good?

src/attributes/codegen.md
```diff
@@ -140,6 +140,28 @@ feature detection on the x86 platforms.
 > may be enabled or disabled for an entire crate with the
 > [`-C target-feature`] flag.
 
+## The `track_caller` attribute
+
+The `track_caller` attribute applies to function declarations, allowing code within the function to
+observe the callsite of the topmost `track_caller`-attributed function in the attributed function's
+callers. If the attributed function itself is called by an unattributed function, then the
+attributed function observes its own callsite. If the attributed function is called by another
+attributed function, then code within the callee observes the callsite of the caller, and so on.
```
Centril: General note re. this paragraph: There's a lot of "attributed function" which makes things feel somewhat dense to follow. I think we could attach some symbols to the cases to make the references in the text clearer (requiring less mental state). For example:
> If the attributed function `f` is called by another attributed function `g`, then code within the `f` observes the callsite of `g`, and so on.

src/attributes/codegen.md
```diff
@@ -140,6 +140,28 @@ feature detection on the x86 platforms.
...
+
+> Note: `rustc` implements the `core::intrinsics::caller_location` intrinsic for observing the
+> callsite of a function with `#[track_caller]`, wrapped by `core::panic::Location::caller` for
+> general use.
+
+Coercing a function with `#[track_caller]` to a function pointer creates a shim which appears to
+observers to have been called at the attributed function's definition site.
+
+> Note: Rules around shims for function pointers are necessary because `rustc` implements
```
Centril: Just to be clear, by "rules" you mean the coercion above? If so, can you reword to "The aforementioned shim for function pointers..."?

src/attributes/codegen.md
anp: Question: does the reference have a notion of the call stack?

anp: I ask because I think miri's implementation of walking up the call stack is actually the most intuitive way to understand the semantics of the attribute. Describing it is pretty natural in terms of the observation point "looking up" the stack until it finds an unattributed function's frame.

Centril: I "grepped" for "stack" and "call" and found the following:
- https://github.com/rust-lang/reference/blob/8e7d614303b0dec7492e048e63855fcd3b944ec8/src/expressions/return-expr.md#L9

- https://github.com/rust-lang/reference/blob/33dfdbda87931a8e0868a6c5066cc272463c8b7e/src/memory-ownership.md#L3

which seems like a "yes"?

I agree that what you suggest would be intuitive.

src/attributes/codegen.md
```diff
@@ -140,6 +140,28 @@ feature detection on the x86 platforms.
 > may be enabled or disabled for an entire crate with the
 > [`-C target-feature`] flag.
 
+## The `track_caller` attribute
+
+The `track_caller` attribute applies to function declarations, allowing code within the function to
```
anp: TODO document ABI requirements

anp: TODO document trait behavior

nikomatsakis:
(Looking back, I realize that most the things I listed were actually already marked as TODO by @anp, so I guess "seems good" =)

anp:
OK, I think I've addressed the outstanding feedback, ready for more!

src/attributes/codegen.md
```diff
@@ -143,6 +143,54 @@ feature detection on the x86 platforms.
...
+implementations, otherwise undefined behavior results. When applied to a function which is made
+available to an `extern` block, the declaration in the `extern` block must also have the attribute,
+otherwise undefined behavior results.
+
+### Behavior
+
+Applying the attribute to a function `f` allows code within `f` to observe the [`Location`] of the
+"topmost" tracked call that led to `f`'s invocation. At the point of observation, an implementation
+behaves as if it walks up the stack from `f`'s frame to find the nearest frame of an *unattributed*
+function `outer`, and it returns the [`Location`] of the tracked call in `outer`.
```
nikomatsakis, line 163: I feel like this paragraph should mention the "hint-like" character of this attribute. It reads to me as though the behavior is guaranteed. I think just using the word "hint" would suffice, perhaps like "Note however that this result is a "hint"; in some scenarios we may halt the walk early", or even just saying "However, be sure to read the Limitations section for some important caveats."

anp: Good call, think I've addressed this.

anp:
https://github.com/rust-lang/rust/commit/ae79c30d74a52e81687bb35af6481eeeeafaeb84 has landed so AIUI this is ready to as well. Or is it preferred to wait for a nightly release including the PR?

ehuss:
@anp would you be able to add a small example? It doesn't need to be complicated, just a basic usage of placing the attribute on a function and calling `std::panic::Location::caller()`.

Also, I'm wondering if "codegen" is the right chapter to put this in. Is there any particular reason you placed it there?  Generally my intent is that section is for things that influence how LLVM operates.  But my understanding of track_caller is that its support is injected into MIR, right?  I don't really know how this is implemented, but I'm guessing it changes the Rust ABI to pass the location along?  Would "runtime" make more sense since this is related to panic handling? Or ABI (if that's how it works)?  Hm, none of these seem like a really good fit...

anp:
Good call adding an example. LMK what you think of the ones I included, I just built off of the existing prose examples.

The attribute is mostly implemented in the codegen portions of the compiler, although I agree that runtime and/or ABI seem like good options too. Let me know where you'd like it?

ehuss:
Thanks!  I guess it is ok to keep it in the codegen section for now.  If a better organization presents itself in the future, we can always move things around later.

I'd like to get one other person to do a quick review before merging.  I'm going to randomly ping @joshtriplett, can you take a quick look at this to double-check it makes sense?

joshtriplett:
Looks good to me.

ehuss:
Ok, thanks @anp and @joshtriplett!

PR #747, opened 2020-02-09
fix TypeParamBound

king6cong:


ehuss:
Thanks for the PR!  Can you explain why you think this change needs to be made?  The `+` repetition is already captured in the _TypeParamBounds_ rule, so I don't see a reason that this is needed.

king6cong:
@ehuss yes, you are right 😄

PR #758, opened 2020-02-12
remove redundant `public`

king6cong:


Centril:
I'm not sure this is redundant as:
```rust
pub use foo::private;

mod foo {
    fn private() {}
}
```
results in an error.

king6cong:
@Centril 
As in https://doc.rust-lang.org/stable/reference/visibility-and-privacy.html#re-exporting-and-visibility

```rust
pub use self::implementation::api;

mod implementation {
    pub mod api {
        pub fn f() {}
    }
}
```

Although `fn f` has pub vis itself, `implementation::api::f` is private to extern crates, and reexport makes it accessible to extern crates. I don’t know if this is a public name as mentioned here.

Havvy:
That example is covered by "even a definition with a private canonical path", as far as I can tell.

Havvy:
Closing without a better argument that the word is redundant.

Nonetheless, thanks for attempting to help!

PR #760, opened 2020-02-12
isolate non-mut static part from static-item section

king6cong:


src/items/static-items.md
```diff
@@ -8,12 +8,16 @@
...
+## Non-mut statics
+
+Non-`mut` static items that contain a type
 that is not [interior mutable] may be placed in read-only memory. Static items
 do not call [`drop`] at the end of the program.
 
-All access to a static is safe, but there are a number of restrictions on
-statics:
+All access to a Non-`mut` static is safe, but there are a number of restrictions on
+Non-`mut` statics:
```
ehuss: Not all these rules seem to apply to non-`mut` static.  In fact, maybe only the first one?

Also, this sentence (the original) seems to be a very strong assertion (**all access**?).  I'm not up to speed on safety issues, so maybe @Centril could validate what should be safe, or if this wording makes sense.

Centril: Yeah this diff seems to remove rules about `static mut` like the thing about referring to constants.

cc @RalfJung and @eddyb on the access thing, but notably, a `static` in an  `extern` block is not safe to access and requires an `unsafe` context (https://doc.rust-lang.org/nightly/reference/items/external-blocks.html#statics). However, some `static` take be part of some internal contract that makes it unsound to access under some circumstances, even if the language rules do not enforce it. Under that reading it seems to me that the current wording makes sense.

src/items/static-items.md
```diff
@@ -8,12 +8,16 @@
...
+
+## Non-mut statics
+
+Non-`mut` static items that contain a type
 that is not [interior mutable] may be placed in read-only memory. Static items
 do not call [`drop`] at the end of the program.
 
-All access to a static is safe, but there are a number of restrictions on
-statics:
+All access to a Non-`mut` static is safe, but there are a number of restrictions on
```
RalfJung, line 19:
```suggestion
All access to a non-`mut` static is safe, but there are a number of restrictions on
```

src/items/static-items.md
RalfJung, line 20:
```suggestion
non-`mut` statics:
```

RalfJung: I mean you can access a non-`mut` static in safe code, so as a matter of fact this statement is correct, or at least some interpretation of it. ;) However, as a reflection of what the compiler is doing, `extern` statics should probably be excluded somewhere.

"all access"  is slightly misleading though; writing to a `static X: i32` is *not* safe and thus rejected by the compiler.

(This might also be a good place to call out that mutating a non-`mut` non-interior-mut `extern static X: T` *is* UB, even if that mutation happens by non-Rust code.)

PR #763, opened 2020-02-16
Add metavariable examples to macros by example

Lonami:
When learning macros, a few examples would have been very valuable. They help see "what do I need for my use case" more clearly without having to click through a dozen links. I also copied a brief description from the links and added it inline, to help better explain the concepts to newcomers.

[Rendered](https://github.com/Lonami/reference/blob/patch-1/src/macros-by-example.md#metavariables).

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -115,23 +115,60 @@ foo!(3);
 ## Metavariables
 
 In the matcher, `$` _name_ `:` _fragment-specifier_ matches a Rust syntax
-fragment of the kind specified and binds it to the metavariable `$`_name_. Valid
-fragment specifiers are:
+fragment of the kind specified and binds it to the metavariable `$`_name_.
 
-  * `item`: an [_Item_]
-  * `block`: a [_BlockExpression_]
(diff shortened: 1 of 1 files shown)


Centril:
I'm not sure we should be adding more redundancy to the reference as then we also need to maintain the same concepts in more places making things less canonical. Having to click through more links doesn't seem like an issue for the reference, which is really not a user guide and is especially not for beginners.

Lonami:
That's a fair point. I personally found [Macros By Example in the Reference](https://doc.rust-lang.org/reference/macros-by-example.html) easier to understand than the [Macros section in the Book](https://doc.rust-lang.org/book/ch19-06-macros.html), which is why I thought it would be fitting to improve it right here.

Maybe I should try to improve [the section in the book instead](https://github.com/rust-lang/book/blob/master/src/ch19-06-macros.md), then?

Centril:
Yeah, I think this would be more appropriate for the book, although you might want to talk to the book maintainers first re. how much "advanced" stuff you want to add and how to structure that.

Lonami:
Thanks for the directions, closing this then.

PR #766, opened 2020-02-22
Fix incorrect pseudocode for #[repr(C)] struct alignment

kinseytamsin:
Alternative to #559 incorporating some suggested changes to that PR.

@RalfJung I felt like your pseudocode approached actual code too much, and sorta requires understanding what's going on with `Layout` in order to follow, so I just took the route of making the padding calculation into an implicit function per other suggestions.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -197,6 +197,8 @@ Here is this algorithm described in pseudocode.
 
 <!-- ignore: pseudocode -->
 ```rust,ignore
+/// Returns the amount of padding needed after `offset` to ensure that the
+/// following address will be aligned to `alignment`.
 fn padding_needed_for(offset: usize, alignment: usize) -> usize {
     let misalignment = offset % alignment;
     if misalignment > 0 {
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -197,6 +197,17 @@ Here is this algorithm described in pseudocode.
 
 <!-- ignore: pseudocode -->
 ```rust,ignore
+fn padding_needed_for(offset: usize, alignment: usize) -> usize {
+    let misalignment = offset % alignment;
+    if misalignment > 0 {
+        // round up to next multiple of `alignment`
+        alignment - misalignment
+    } else {
+        // already a multiple of `alignment`
+        0
+    }
+}
+
 struct.alignment = struct.fields().map(|field| field.alignment).max();
 
 let current_offset = 0;
@@ -205,10 +216,6 @@ for field in struct.fields_in_declaration_order() {
     // Increase the current offset so that it's a multiple of the alignment
     // of this field. For the first field, this will always be zero.
     // The skipped bytes are called padding bytes.
-    //
-    // padding_needed_for() is equivalent to
-    // std::alloc::Layout::padding_needed_for(), but takes an integer rather
-    // than a Layout as the first argument.
     current_offset += padding_needed_for(current_offset, field.alignment);
 
     struct[field].offset = current_offset;
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -226,6 +226,14 @@ for field in struct.fields_in_declaration_order() {
 struct.size = current_offset + padding_needed_for(current_offset, struct.alignment);
 ```
 
+<div class="warning">
+
+Warning: This pseudocode uses a naive algorithm that ignores overflow issues for
+the sake of clarity. To perform memory layout computations in actual code, use
+[`Layout`].
+
+</div>
+
 > Note: This algorithm can produce zero-sized structs. This differs from
 > C where structs without data still have a size of one byte.
 
@@ -374,3 +382,4 @@ used with any other representation.
 [`C`]: #the-c-representation
 [primitive representations]: #primitive-representations
 [`transparent`]: #the-transparent-representation
+[`Layout`]: ../std/alloc/struct.Layout.html
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -205,14 +205,18 @@ for field in struct.fields_in_declaration_order() {
     // Increase the current offset so that it's a multiple of the alignment
     // of this field. For the first field, this will always be zero.
     // The skipped bytes are called padding bytes.
-    current_offset += field.alignment % current_offset;
+    //
+    // padding_needed_for() is equivalent to
+    // std::alloc::Layout::padding_needed_for(), but takes an integer rather
+    // than a Layout as the first argument.
+    current_offset += padding_needed_for(current_offset, field.alignment);
 
     struct[field].offset = current_offset;
 
     current_offset += field.size;
 }
 
-struct.size = current_offset + current_offset % struct.alignment;
+struct.size = current_offset + padding_needed_for(current_offset, struct.alignment);
 ```
 
 > Note: This algorithm can produce zero-sized structs. This differs from


rkanati:
Neat! I'll nix mine.

kinseytamsin:
@rkanati I assumed that the fact that the original PR stalled meant that you weren't likely to show up, I kinda wanted you to be able to fix yours and get the credit :sob: But I also wanted it done, so I guess I haven't really done anything wrong XD

rkanati:
No problem! It's just a doc fix, after all.

RalfJung:
I agree now that it's better not to use `Layout`, to keep this code self-contained.

However, could you please define `padding_needed_for`? As we have seen, it is somewhat subtle to get right.

src/type-layout.md
```diff
@@ -205,14 +205,18 @@ for field in struct.fields_in_declaration_order() {
...
+    // than a Layout as the first argument.
+    current_offset += padding_needed_for(current_offset, field.alignment);
 
     struct[field].offset = current_offset;
 
     current_offset += field.size;
 }
 
-struct.size = current_offset + current_offset % struct.alignment;
+struct.size = current_offset + padding_needed_for(current_offset, struct.alignment);
```
RalfJung: @rkanati you didn't touch this one in your PR... do you remember why? Was that an oversight, or was the original code actually correct for some reason?

kinseytamsin:
@RalfJung Yeah, I started including a definition of the function but wasn't sure whether that would be necessary. Luckily now that I know about `Layout` (and even which method to look at), I can just copy that code :)

kinseytamsin:
Okay so, for reference, here's the code for `Layout::padding_needed_for`.

```rust
pub fn padding_needed_for(&self, align: usize) -> usize {
    let len = self.size();

    // Rounded up value is:
    //   len_rounded_up = (len + align - 1) & !(align - 1);
    // and then we return the padding difference: `len_rounded_up - len`.
    //
    // We use modular arithmetic throughout:
    //
    // 1. align is guaranteed to be > 0, so align - 1 is always
    //    valid.
    //
    // 2. `len + align - 1` can overflow by at most `align - 1`,
    //    so the &-mask with `!(align - 1)` will ensure that in the
    //    case of overflow, `len_rounded_up` will itself be 0.
    //    Thus the returned padding, when added to `len`, yields 0,
    //    which trivially satisfies the alignment `align`.
    //
    // (Of course, attempts to allocate blocks of memory whose
    // size and padding overflow in the above manner should cause
    // the allocator to yield an error anyway.)

    let len_rounded_up = len.wrapping_add(align).wrapping_sub(1)
        & !align.wrapping_sub(1);
    len_rounded_up.wrapping_sub(len)
}
```

I'm feeling like this might be a bit overkill to directly adapt into what's supposed to be pseudocode. @RalfJung do you think I should use this directly for the definition of `padding_needed_for` in the pseudocode or should I write a definition according to the naive definition of the algorithm (i.e., without the measures to account for extreme values/overflowing)?

I'm actually having a bit of trouble fully understanding what's going on in that code, what with all the layered modular arithmetic and bitwise operations (I'm not really fully acquainted with the identities and such to be able to easily follow what bitwise operations are doing, so after puzzling it out a bit it's turning out to be extremely non-obvious to me how to translate that into a simple presentation of the algorithm suited to pseudocode). Maybe @pnkfelix can help since they wrote that code?

RalfJung:
I was actually thinking of using the implementation from the other PR:
```
fn padding_needed_for(offset, alignment) {
    let misalignment = current_offset % alignment;
    if misalignment > 0 {
        current_offset += alignment - misalignment;
    }
}
```
Maybe also add a note somewhere that we are ignoring overflow concerns here, and link to `Layout` for people that actually want to do these computations in real code.

kinseytamsin:
@RalfJung Yeah, I'll do that. I'm definitely overthinking this, heh.

kinseytamsin:
@RalfJung I added a function definition to the pseudocode as requested. I agree that the note directing the reader to `Layout::padding_needed_for` is a good idea, but should it be a note or a warning? I figured I'd ask for a second opinion before I actually add that, to be sure.

RalfJung:
What do you mean, "warning"? What would it warn about?

RalfJung:
Oh you mean, the fact that we ignore overflows. Yeah seems best to me to call it a warning, we don't want people to copy-paste this code.

src/type-layout.md
```diff
@@ -197,6 +197,17 @@ Here is this algorithm described in pseudocode.
 
 <!-- ignore: pseudocode -->
 ```rust,ignore
+fn padding_needed_for(offset: usize, alignment: usize) -> usize {
```
RalfJung: Maybe add a (brief) doc comment?

kinseytamsin:
@RalfJung Any further concerns with this PR? :slightly_smiling_face: (Besides the need to squash these commits before merging.)

src/type-layout.md
kinseytamsin: As far as I can tell, that was just an oversight.

src/type-layout.md
```diff
@@ -197,6 +197,8 @@ Here is this algorithm described in pseudocode.
 
 <!-- ignore: pseudocode -->
 ```rust,ignore
+// Returns the amount of padding needed after `offset` to ensure that the
+// following address will be aligned to `alignment`.
```
RalfJung: Please use `///`

src/type-layout.md
RalfJung: I think so, too. It would just have been good to get explicit confirmation.

src/type-layout.md
kinseytamsin: I figured it wouldn't matter in pseudocode, but I'll do it.

src/type-layout.md
```diff
@@ -226,6 +228,15 @@ for field in struct.fields_in_declaration_order() {
 struct.size = current_offset + padding_needed_for(current_offset, struct.alignment);
 ```
 
+<div class="warning">
+
+Warning: The function used to calculate the needed padding in this pseudocode is
+a naive implementation that doesn't account for overflow issues. For a rigorous
```
RalfJung: The pseudocode as a whole does not account for overflows -- it's not just the padding part.

IMO this should recommend to construct the layout using `extend` to add each field, and using `pad_to_align` for the final padding.

RalfJung:
So the code looks all right to me now, but I'd prefer to get this confirmed by some data layout guru... summoning @gnzlbg @hanna-kruppe @eddyb

src/type-layout.md
RalfJung: I figured if we are doing this we might as well do it right. ;)

kinseytamsin: Fair enough, haha. At this point I wonder if this can even be called pseudocode except by virtue of technically not being executable (keyword collision, implicit struct/field definitions), but the merits of using the pseudo-Rust syntax are beyond the scope of this PR. :P

src/type-layout.md
kinseytamsin: I wonder if it might just be better to just make the link part a general recommendation to use `std::alloc::Layout` to work with layouts in real code.

RalfJung: I mean, yes, that was basically the point of this warning in my eyes.

kinseytamsin:
@RalfJung All your review comments should be addressed now :slightly_smiling_face:

RalfJung:
Thanks, I think they are. :)

Now we are just waiting for some extra eyes on this.

hanna-kruppe:
Looks fine to me too, thanks @kinseytamsin!

Issue #771, opened 2020-03-03
Incomplete documentation of `while let` syntax regarding lazy boolean operator expressions.


steffahn:
Comparing the reference’s [section on `if let`](https://doc.rust-lang.org/nightly/reference/expressions/if-expr.html#if-let-expressions) vs. [the one on `while let`](https://doc.rust-lang.org/nightly/reference/expressions/loop-expr.html#predicate-pattern-loops), the latter contains no note about top-level `||` and `&&` being disallowed, even though this is the case by [the corresponding RFC](https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#rollout-plan-and-transitioning-to-rust-2018) as well as in practice.

Centril:
The corresponding RFC has not been stabilized, the syntax `if p && let p = q { ... }` is unstable. The reference only documents stable Rust.

steffahn:
@Centril No. I’m talking about the part of the RFC that makes Rust2018 disallow certain expressions. This rule, that `||` and `&&` cant appear inside `if let` or `while let` is part of stable Rust. It is even in the reference already, however only for `if let`, not for `while let`.

Centril:
Oh you mean the `Expression except struct or lazy boolean operator expression` bit in if-let vs. while-let; yeah that needs to be updated. Want to send in a PR?

steffahn:
Sure, I’ll do that. Playing around with this, there seems to be more to it though.. is seems as if the `=` changes precedence inside an `if let`. According to [the reference](https://doc.rust-lang.org/nightly/reference/expressions.html#expression-precedence) it is supposed to be binding weaker than `&&` and `||` and there’s also `..` and `..=` in between. Apparently the implementation of this part of the RFC changed things in a way that also disallows `..`: [see this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=78e90cc80d80640e6a860f03f629585c).

PR #772, opened 2020-03-04
Add information about || and && to grammar describing `while let`.

steffahn:
Also fix italics to be consistent with other pages.
See #771. I also copied and adapted the corresponding paragraph from the `if let` page.
_Edit: Instead of copying the paragraph from `if let` mentioning the unstable `if-let chains` feature, now the copy as well as the original paragraph are removed._

### DIFF
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -123,26 +123,7 @@ while let Some(v @ 1) | Some(v @ 2) = vals.pop() {
 }
 ```
 
-The expression cannot be a [lazy boolean operator expression][_LazyBooleanOperatorExpression_].
-Use of a lazy boolean operator is ambiguous with a planned feature change
-of the language (the implementation of if-let chains - see [eRFC 2947][_eRFCIfLetChain_]).
-When lazy boolean operator expression is desired, this can be achieved
-by using parenthesis as below:
-
-<!-- ignore: psuedo code -->
-```rust,ignore
-// Before...
-while let PAT = EXPR && EXPR { .. }
-
-// After...
-while let PAT = ( EXPR && EXPR ) { .. }
-
-// Before...
-while let PAT = EXPR || EXPR { .. }
-
-// After...
-while let PAT = ( EXPR || EXPR ) { .. }
-```
+As is the case in [`if let` expressions], the scrutinee cannot be a [lazy boolean operator expression][_LazyBooleanOperatorExpression_].
 
 ## Iterator loops
 
@@ -315,3 +296,5 @@ expression `()`.
 [`match` expression]: match-expr.md
 [scrutinee]: ../glossary.md#scrutinee
 [temporary values]: ../expressions.md#temporary-lifetimes
+[_LazyBooleanOperatorExpression_]: operator-expr.md#lazy-boolean-operators
+[`if let` expressions]: if-expr.md#if-let-expressions
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -45,7 +45,7 @@ have type compatible with the value of the `break` expression(s).
 
 > **<sup>Syntax</sup>**\
 > _PredicateLoopExpression_ :\
-> &nbsp;&nbsp; `while` [_Expression_]<sub>except struct expression</sub> [_BlockExpression_]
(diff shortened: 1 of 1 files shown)


src/expressions/loop-expr.md
```diff
@@ -123,11 +123,32 @@ while let Some(v @ 1) | Some(v @ 2) = vals.pop() {
...
+
+// After...
+while let PAT = ( EXPR && EXPR ) { .. }
+
+// Before...
+while let PAT = EXPR || EXPR { .. }
+
+// After...
+while let PAT = ( EXPR || EXPR ) { .. }
+```
```
Centril: This is user advice and discussion of unstable features, which is off-topic for the reference.

steffahn: Then perhaps it should also be removed from the page about `if` expressions?

Centril: Probably yes, cc @ehuss

ehuss: I think it's fine to keep this information, though place it in a "note" block. I don't see a reason to *never* talk about such things, since it is relevant here (explaining a strange peculiarity of the language).  The C++ reference does the same with side notes mentioning future changes or how to disambiguate things that aren't obvious.

The information only needs to live in one place.  `if let` would be fine, with `while let` linking to `if let` for more information.  I might even trim `while let` even more to remove redundant information, but doesn't necessarily need to be done in this PR.

ehuss:
Hm, I don't have the answer for that, @Centril should.   It looks like `let` has special precedence rules ([`parse_let_expr`](https://github.com/rust-lang/rust/blob/2b0cfa5b4c5099f45ca540ee1c7d8c1ecd5267d2/src/librustc_parse/parser/expr.rs#L1473-L1475)) from `if let` ([`parse_assoc_expr`](https://github.com/rust-lang/rust/blob/2b0cfa5b4c5099f45ca540ee1c7d8c1ecd5267d2/src/librustc_parse/parser/expr.rs#L135)).  I'm not too familiar with how that works.

I general, the reference is missing lots of information on precedence rules and disambiguations, so don't be too surprised about details like this not being covered (though they should!).

src/expressions/loop-expr.md
Centril: Putting it in a note and deduplicating seems like a reasonable resolution to me.

Centril:
> @ehuss I'm questioning if "Expression__except struct or lazy boolean operator expression_" is even accurate. It seems like the compiler is currently, internally, for parsing and some typechecking, treating

This only applies to parsing and AST level validation. Type checking (done on HIR) does not care about parenthesis and precedences. The old note is probably outdated by now.



> Hm, I don't have the answer for that, @Centril should. It looks like `let` has special precedence rules ([`parse_let_expr`](https://github.com/rust-lang/rust/blob/2b0cfa5b4c5099f45ca540ee1c7d8c1ecd5267d2/src/librustc_parse/parser/expr.rs#L1473-L1475)) from `if let` ([`parse_assoc_expr`](https://github.com/rust-lang/rust/blob/2b0cfa5b4c5099f45ca540ee1c7d8c1ecd5267d2/src/librustc_parse/parser/expr.rs#L135)). I'm not too familiar with how that works.

Yes, the parser treats `let pat = scrutinee` as an expression on its own, and specifically requires that `scrutinee` has a minimum precedence of one above the precedence of `&&`. The relevant code is:

```rust
/// In `let p = e`, operators with precedence `<=` this one requires parenthesis in `e`.
pub fn prec_let_scrutinee_needs_par() -> usize {
    AssocOp::LAnd.precedence()
}

    /// Parses a `let $pat = $expr` pseudo-expression.
    fn parse_let_expr(&mut self, attrs: AttrVec) -> PResult<'a, P<Expr>> {
        let pat = self.parse_top_pat(GateOr::No)?;
        self.expect(&token::Eq)?;
        let expr = self.with_res(Restrictions::NO_STRUCT_LITERAL, |this| {
            this.parse_assoc_expr_with(1 + prec_let_scrutinee_needs_par(), None.into())
        })?;
```

...so what we should do is document the minimum precedence requirement; everything else just falls out automatically. (We shouldn't document that `let` is an expression on its own, as that is still an unstable implementation detail.)

PR #776, opened 2020-03-04
Fix expression and statement grammar.

### LINKED ISSUE #773: Is a semicolon part of statement or not? Is a semicolon really never part of `Statement` nonterminal expansion containing `ExpressionWithBlock`?
According to [this page](https://doc.rust-lang.org/nightly/reference/expressions/block-expr.html), “The semicolon following a statement is not a part of the statement itself.” While this view may be true and relevant for using the `stmt` parser in procedural macros, the grammar, in particular a nonterminal literally named `Statement`, tells a different story. I believe this is confusing.

Apart from the above, there is a problem / inaccuracy in the nonterminal `ExpressionStatement`. [The respective page](https://doc.rust-lang.org/nightly/reference/statements.html#expression-statements) accurately conveys that an `ExpressionWithoutBlock` must be followed by `;` and an `ExpressionWithBlock` usually isn’t. However, an `ExpressionWithBlock` most definitely _can_ have a `;` following it. This might syntactically be well covered by the fact that `;` in by itself [is a valid `Statement`](https://doc.rust-lang.org/nightly/reference/statements.html#statements), and I wouldn’t complain if this kind of `;` was always redundant and like a no-op statement following the `ExpressionWithoutBlock`-statement, but this isn’t the case by means of what is stated below the grammar of `ExpressionStatement`, “The type of `ExpressionWithBlock` expressions when used as statements must be the unit type.”

It goes on further around there and speaks of a “trailing semicolon” one “can omit” and clarifies that more precisely, only “When the trailing semicolon is omitted, the result must be type `()`.” All this means is: A semicolon does change things: it allows the return type to be non-(). If the semicolon is relevant for type-checking an `ExpressionStatement` containing `ExpressionWithBlock`, in my view this should imply that said semicolon _must be part of the statement_.

I‘m not 100% sure what the best solution to improve this is. I’d suggest that the rule becomes:
`ExpressionStatement:  ExpressionWithoutBlock ; | ExpressionWithBlock ;?`
i.e.: allow an optional semicolon behind ExpressionWithBlock, being part of the ExpressionStatement. Maximum munch would mean that if any `;` exists it will then be considered part of the statement. But there’s also the issue #762 that seems related. Perhaps there is a different approach that solves both problems together.

### LINKED ISSUE #774: Wrong grammar for `match`.
[In the grammar for `match`](https://doc.rust-lang.org/nightly/reference/expressions/match-expr.html), i.e. for the nonterminal `MatchExpression`, in `MatchArms` the right-hand side of the non-last arms only allows skipping the `,` for `BlockExpression`. However [it seems to be the case](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1114d1a7917ae89458b351017ea1d49e) that any `ExpressionWithBlock` works without the `,`. _Is this correct?_

Furthermore, currently the case of a `BlockExpression` followed by `,` matches both alternatives of the `BlockExpression ,? | Expression ,`. Perhaps a nicer rule would use `ExpressionWithoutBlock , | ExpressionWithBlock ,?`, in effect being similar to my proposed rule for `ExpressionStatement` in #773, but with comma instead of semicolon.

Finally, in the last arm, the `BlockExpression | Expression` part seems a bit redundant as a `BlockExpression` is an `Expression`.

steffahn:
Semicolon or comma in expression statements or match arms, respectively, are optional, but part of the statement / match arm.
Also remove a note about proc-macros from the page about BlockExpression.
Closes #773 and closes #774.

### DIFF
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -28,9 +28,6 @@ a following semicolon except if its outer expression is a flow control
 expression. Furthermore, extra semicolons between statements are allowed, but
 these semicolons do not affect semantics.
 
-> Note: The semicolon following a statement is not a part of the statement
-> itself. They are invalid when using the `stmt` macro matcher.
-
 When evaluating a block expression, each statement, except for item declaration
 statements, is executed sequentially. Then the final expression is executed,
 if given.
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -9,10 +9,10 @@
 >
 > _MatchArms_ :\
 > &nbsp;&nbsp; ( _MatchArm_ `=>`
->                             ( [_BlockExpression_] `,`<sup>?</sup>
->                             | [_Expression_] `,` )
+>                             ( [_ExpressionWithoutBlock_][_Expression_] `,`
+>                             | [_ExpressionWithBlock_][_Expression_] `,`<sup>?</sup> )
 >                           )<sup>\*</sup>\
-> &nbsp;&nbsp; _MatchArm_ `=>` ( [_BlockExpression_] | [_Expression_] ) `,`<sup>?</sup>
+> &nbsp;&nbsp; _MatchArm_ `=>` [_Expression_] `,`<sup>?</sup>
 >
 > _MatchArm_ :\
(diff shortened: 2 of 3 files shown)


steffahn:
I just understood how the links in the markdown work. I'll have to fix those before merging this.

Centril:
Thanks!

Issue #626, opened 2019-06-26
Character and string token definitions need updating.


ehuss:
There are multiple issues here. Some of this has changed in 1.37 via https://github.com/rust-lang/rust/pull/60793.

- `RAW_BYTE_STRING_LITERAL` no longer allows bare CR (new 1.37).

- "Raw string" and "raw byte string" needs to be updated that CRLF is converted to LF (new 1.37).

- Several tokens need to sync the English text with the "Lexer" definition.
    - `STRING_LITERAL` indicates several rules (like isolated CR's are not allowed), but the text does not mention any of those restrictions.
    - `CHAR_LITERAL` says "single Unicode character…except U+0027" which is not complete.
    - `RAW_STRING_LITERAL` does not allow bare CR's.
    - `BYTE_LITERAL` escapes are not described.
    - `BYTE_STRING_LITERAL` restrictions are not described.
    - In general, just make sure they are all in sync!
    
- Typo in `RAW_BYTE_STRING_CONTENT`, points to `RAW_STRING_CONTENT` when it should be `RAW_BYTE_STRING_CONTENT`.

- I cannot find anywhere that mentions CRLF in a string is converted to LF. Am I blind?

- The description for string continuations says "`\` immediately before `U+000A`", but it can also be before CRLF.  How should this be handled?  I haven't looked at how it is implemented, but are all CRLF's translated everywhere?  Should there just be a blanket statement somewhere about this, to avoid having to discuss it in every string literal definition?

I may be missing some things here. Need to very thoroughly review everything to make sure it is correct and up-to-date with the changes from 60793.

ehuss:
See also https://github.com/rust-lang/rust/issues/62865

mattheww:
https://github.com/rust-lang/rust/pull/118699#issuecomment-1852867466
should be helpful.

mattheww:
A form like `b"\u{00a0}"` is rejected at lexing time ("unicode escape in byte string").

But as it doesn't match either `BYTE_STRING_LITERAL` or `RESERVED_TOKEN_DOUBLE_QUOTE`, the current description says there's a valid tokenisation as the identifier `b` followed by `"\u{00a0}"`.

So if we keep on with the current mechanism for documenting such rejected tokens, I think we'd need yet more reserved forms.

There are probably other similar cases. I think after rust-lang/rust#119172 a
C string literal containing a `NUL` is one.

Issue #567, opened 2019-04-21
Clean up and consolidate the lexical specification.


ehuss:
The lexical specification needs some cleanup and organization. Some things I can think of:

- There should be an overall introduction and overview of the lexical structure.
- Paths are in the Lexical chapter, but I don't think they should be.
- The UTF8BOM/SHEBANG definition is floating in a chapter outside of the Lexical chapter. I think it is relevant to lexing, so it should be somehow incorporated in the Lexical chapter. (Not sure how, probably need to rearrange things a little.)
- I think there should be an appendix consolidating all the Lexer rules blocks. This should be generated automatically.
- The "input format" subchapter is almost completely useless, and could be moved somewhere else.
- There should be a note about token ambiguity (this can be relatively brief, but should be mentioned). This depends on the lexer/parser implementation. rustc works by splitting tokens into smaller parts. The proc_macro parser works by only issuing the smaller tokens, and using the Spacing to determine if they should be combined later on. The tokens that I'm aware of that cause this issue are:

Token | Possibly Split Into
------|--------------------
`+=` | `+` `=`
`&&` | `&` `&`
<code>&#124;&#124;</code> | <code>&#124;</code> <code>&#124;</code>
`<<` | `<` `<`
`<-` | `<` `-`
`>>` | `>` `>`
`>>=` | `>` `>=`
`>=` | `>` `=`
`+=` | `+` `=`

See also:
https://github.com/rust-lang-nursery/wg-grammar/issues/3
https://internals.rust-lang.org/t/pre-pre-rfc-canonical-lexer-specification/4099

ehuss:
I'm going to close as I have split the remaining points to https://github.com/rust-lang/reference/issues/1982 and https://github.com/rust-lang/reference/issues/1983.

There's perhaps a large issue of whether or not there should be a separate lexical definition at all, and instead use a scannerless approach to defining the language. However, that is beyond the scope here.

PR #778, opened 2020-03-11
Fix some typos, misformattings and small mistakes in the lexical structure reference.

steffahn:
Multiple unrelated things in one commit since they’re each very small corrections.
Next to typos and bad formatting, this PR changes a word (“equivalent” ⟶ “equal”) and adds (more precisely, disallows) “\n” in a few obvious places in comments’ lexical structure.

### DIFF
--- a/src/comments.md
+++ b/src/comments.md
@@ -2,7 +2,7 @@
 
 > **<sup>Lexer</sup>**\
 > LINE_COMMENT :\
-> &nbsp;&nbsp; &nbsp;&nbsp; `//` (~[`/` `!`] | `//`) ~`\n`<sup>\*</sup>\
+> &nbsp;&nbsp; &nbsp;&nbsp; `//` (~[`/` `!` \\n] | `//`) ~\\n<sup>\*</sup>\
 > &nbsp;&nbsp; | `//`
 >
 > BLOCK_COMMENT :\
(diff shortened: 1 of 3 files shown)


steffahn:
Just noticing, perhaps the _IsolatedCR_ should be disallowed in the front part of OUTER_LINE_DOC, too...
Furthermore – but on this part I’m not sure about what exactly \n and _IsolatedCR_ mean here – should _IsolatedCR_ also be disallowed in LINE_COMMENT?

steffahn:
Hm.. looking at it further, so far _IsolatedCR_ is only forbidden in anything “DOC” comment, so perhaps it is allowed within a LINE_COMMENT. But I found another mistake in the beginning of OUTER_BLOCK_DOC where the ~`*` most likely needs an _IsolatedCR_ excluded, too. Will add this to the commit aswell.

ehuss:
Yea, that seems correct.

Also, just FYI, there are a number of problems with the lexer chapter.  CRLF handling was recently rewritten, which hasn't been updated in the reference.  See #626.  Also, I've been meaning to rewrite the lexer chapter, since it is a bit disorganized right now (#567 mentions some of the problems).  If you're interested in working on those (or any issues), that would be great!

(I'll leave this PR open for a bit to see if @Centril wants to look.)

steffahn:
Another thing I noticed: The whole thing with _BlockCommentOrDoc_ is totally wrong regarding _IsolatedCR_ s. As far as a few quick tests go, what actually matters is if the whole comment is a block comment or not, not if nested comments happen to be syntactically like a doc comment or a normal comment.
Finally, and this may just be me not quite getting how these grammars are supposed to be read regarding maximum munch etc... but I feel like I’m missing that `/*` be disallowed, too, inside the BLOCK_COMMENT rules where it currently says “~`*/`”.

Centril:
> (I'll leave this PR open for a bit to see if @Centril wants to look.)

cc @petrochenkov @matklad -- I'm not too familiar with the lexer myself.

PR #781, opened 2020-03-13
Fix grammar for variadic functions

matklad:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=17340777c82dc328a58a93f9db6d71c2


ehuss:
Hm, I'm wondering if this was stabilized accidentally.  This was added in https://github.com/rust-lang/rust/pull/57760 ([RFC 2137](https://github.com/rust-lang/rfcs/blob/master/text/2137-variadic.md)), see also https://github.com/rust-lang/rust/issues/44930 and https://github.com/rust-lang/wg-grammar/issues/45.  

Is there any meaning to the identifier of a variadic parameter?  Or is it just ignored?  That should probably be documented.

(Or, if this was truely a mistake, maybe it should be reverted/fixed/made unstable?)

matklad:
I don't know, I guess cc @Centril  / @petrochenkov ?

But apparently people use this in the real world already, this was reported against rust-analyzer.

Centril:
The function parameter grammar as implemented in `rustc` is as follows. I've invented new meta grammar notation here, but it should hopefully be understandable.

Let `RN` (*"require name*") be a function `Edition -> Bool`. Then the (higher order) grammar is:

```rust
FnParams<RN> = "(" head: FirstParam<RN> tail: { "," Param<RN> %% "," } ")" ;

FirstParam<RN> = SelfParam | Param<RN> ;
Param<RN> = CVariadicTy | AlwaysNamedParam | MaybeNamedParam<RN> ;

MaybeNamedParam<RN> if RN(edition(current_token)) = Pat ":" ParamTy ; // Named param.
MaybeNamedParam<RN> = ParamTy ; // Unnamed param.

// Ambiguous with ^-- but for precedence.
AlwaysNamedParam = NamedStart? Ident ":" ParamTy ;
NamedStart = NtPat | "&" | "&&" | "mut" ;

ParamTy = CVariadicTy | TyPlusAllowed ;
CVariadicTy = "..." ;
```

For:
- function pointers, let `RN = \_ -> False`,
- functions in traits, let `RN = \x -> x`,
- otherwise, let `RN = \_ -> True`.

This is translated starting with `parse_fn_params` in the rustc parser.

Beyond this, AST validation imposes other semantic restrictions.

Havvy:
cc @ehuss did #927 do this

ehuss:
Ah, yes, this should now be resolved.

PR #782, opened 2020-03-13
Size fo empty structs in C is zero byte.

ldm0:
```c
#include<stdio.h>

struct A {
};

int main() {
        struct A a;
        printf("%ld\n", sizeof(a));
}
```
```bash
$ gcc tmp.c
$ ./a.out
0
$ g++ tmp.c
$ ./a.out
1
```

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -236,9 +236,12 @@ the sake of clarity. To perform memory layout computations in actual code, use
 
 </div>
 
-> Note: This algorithm can produce zero-sized structs. In C, the size of
-> structs without data is zero. This is not the same as C++ where structs
-> without data still have a size of one byte.
+> Note: This algorithm can produce zero-sized structs. In C, an empty struct
+> declaration like `struct Foo { }` is illegal. However, both gcc and clang
+> support options to enable such structs, and assign them size zero. C++, in
+> contrast, gives empty structs a size of 1, unless they are inherited from or
+> they are fields that have the `[[no_unique_address]]` attribute, in which
+> case they do not increase the overall size of the struct.
 
 #### \#[repr(C)] Unions
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -236,8 +236,9 @@ the sake of clarity. To perform memory layout computations in actual code, use
 
 </div>
 
-> Note: This algorithm can produce zero-sized structs. This differs from
-> C where structs without data still have a size of one byte.
+> Note: This algorithm can produce zero-sized structs. In C, the size of
+> structs without data is zero. This is not the same as C++ where structs
+> without data still have a size of one byte.
 
 #### \#[repr(C)] Unions
 


eddyb:
AFAIK, empty structs in C are a GNU C extension (implemented in GCC and Clang).
That is, the standard disallows that situation but GCC went ahead and supported it.

I wonder if using `-std=c11` changes the behavior (the default is something like `gnu11`).

eddyb:
cc @nagisa @hanna-kruppe @gnzlbg

ldm0:
> I wonder if using `-std=c11` changes the behavior (the default is something like `gnu11`).

```bash
$ gcc tmp.c -std=c11
$ ./a.out
0
```

eddyb:
Maybe `-Wall -Wextra` might show something, I hate how many warnings are off by default.

ldm0:
> Maybe -Wall -Wextra might show something, I hate how many warnings are off by default.

The gcc doesn't complained about this though.

```bash
$ gcc tmp.c -Wall -Wextra
$ ./a.out
0
```

This is really a UB in C:
> If the struct-declaration-list contains no named members, the behavior is undefined.

But since this is what we got, we should at least note the reader right?

ldm0:
Note:
```bash
$ clang tmp.c -Wall -Wextra -std=c11
$ ./a.out
0
$ gcc tmp.c -Wall -Wextra -std=c11
$ ./a.out
0
```

hanna-kruppe:
For comparison, this is also [discussed in the UCG](https://rust-lang.github.io/unsafe-code-guidelines/layout/structs-and-tuples.html#c-compatible-layout-repr-c). I agree the reference should be more correct about this, but the current diff is not the best we can do. I suggest following the UCG in explicitly distinuishing C standard vs what common C compilers do vs C++.

@eddyb Indeed there is a warning about empty structs in C being an extension: [`-Wpedantic`](https://godbolt.org/z/-EDepB).

eddyb:
@hanna-kruppe Thanks, and it also looks like `-std=c11` doesn't affect that warning :disappointed:.

ehuss:
The changes here look good to me, and mostly repeats what is said in UCG.  @eddyb or @hanna-kruppe  should we just merge this?

src/type-layout.md
```diff
@@ -236,8 +236,12 @@ the sake of clarity. To perform memory layout computations in actual code, use
...
 </div>
 
-> Note: This algorithm can produce zero-sized structs. This differs from
-> C where structs without data still have a size of one byte.
+> Note: This algorithm can produce zero-sized structs. In C, an empty struct
+> declaration like `struct Foo { }` is illegal. However, both gcc and clang
+> support options to enable such structs, and assign them size zero. C++, in
+> contrast, gives empty structs a size of 1, unless they are inherited from or
+> they are fields that have the `[[no_unique_address]]` attribute, in which
+> case they do not increase the overall size of the struct.
```
hanna-kruppe: The last bit (empty base optimization / ``[[no_unique_address]]``) is not mentioned in UCG and seems dubiously relevant: `sizeof(Foo)` is not affected by those two tricks, they just modify the layout computation of *other* types. Those other types explicitly do things that don't have a Rust equivalent (inheritance / said attribute), so I don't think this will be as relevant as the rest of this note for people translating struct definitions from C++ to Rust.

hanna-kruppe:
I don't have permissions to merge this, but from my perspective it's a fine improvement (aside from the nit I just left).

Centril:
I'll leave merging this to @ehuss once @hanna-kruppe thinks its ready.

src/type-layout.md
ldm0: @hanna-kruppe The ``[[no_unique_address]]` is mentioned in UCG I guess? Actually the EBO is not limited to inheritance: 
<https://en.cppreference.com/w/cpp/language/ebo>
 ```cpp
#include <cassert>

struct Empty {}; // empty class
 
struct X {
    int i;
    [[no_unique_address]] Empty e;
};
 
int main()
{
    // the size of any object of empty class type is at least 1
    assert(sizeof(Empty) >= 1);
 
    // empty member optimized out
    assert(sizeof(X) == sizeof(int));
}
```

hanna-kruppe: Oh, you're right, both empty base optimization and `[[no_unique_address]]` are mentioned in UCG, sorry for the confusion. I also take issue with the wording in UCG (and I can't see how your link or example supports that "EBO is not limited to inheritance" -- the attribute is separate from empty *base* optimization), but that need not block this PR.

hanna-kruppe:
I think this is ready.

ldm0:
Ping @Centril

PR #784, opened 2020-03-23
Clarify rule for end-of-line backslashes in strings

bgeron:


ehuss:
I think I can see how it can be confusing due to the use of commas, but I think the original intent is: There is a preceding clause which establishes *when* a continuation happens ("when an unescaped `U+005C` character (`\`) occurs immediately before the line-break") and then lists 3 things that are ignored:

1. The backslash.
2. The linebreak.
3. The whitespace at the beginning of the next line.

Reading it that way, this change seems to remove number 2, which seems wrong.  Perhaps it can be rewritten to be clearer?  Maybe a simple change like this?

```diff
 when an unescaped `U+005C` character (`\`) occurs immediately before the
-line-break, the `U+005C` character, the line-break, and all whitespace at the
+line-break, then the `U+005C` character, the line-break, and all whitespace at the
 beginning of the next line are ignored. Thus `a` and `b` are equal:

```

bgeron:
Great find Eric! Integrated your suggestion and changed the PR title.

bgeron:
Can I just say how happy I am that the Rust community likes to clarify and make small improvements to things, as opposed to other communities with slow processes where things don't get fixed and everyone just accepts the mild snafus

♥

bgeron:
Ready to be merged, I think!

PR #793, opened 2020-04-11
UB definition: mention size_of_val for determining a reference's span

RalfJung:


RalfJung:
I wish we'd use one-sentence-per-line or one-thought-per-line instead of wrapping the file. ;)

Centril:
Also that, yes. :)

PR #802, opened 2020-04-16
Drive-by mention unsafe fn closure coercion

alercah:


ehuss:
I'm a bit uncertain about why it would call out `unsafe` specifically here. `unsafe fn` isn't a distinct pointer type from `fn` is it?  I wouldn't expect to call out all the function "front matter" like "extern" here, either.

alercah:
Yes, they are. Interestingly, though, `unsafe` closures can be assigned to regular function pointers: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=de1dee0081843286d774d1d21c27091c

(I'm not sure this is wrong, but it's interesting.)

I forgot about `extern`, but closures can't be converted to extern functions (other than usual Rust ones).

ldm0:
A drive-by developer here.

@alercah There is no "unsafe closure" actually. For the `let u = || unsafe {};`, `u` is just a closure with an unsafe block.

But I think this PR maybe useful. Since the rustc currently support coercing `non-capturing closure` to [safe](https://github.com/rust-lang/rust/issues/39817) and [unsafe](https://github.com/rust-lang/rust/pull/59580) function pointer. Coercing `non-capturing closure` to `extern` function pointer may be implemented in the future(<https://github.com/rust-lang/rust/issues/44291>) but now it won't pass compilation. I guess the reference need to point out either the compile-passing `unsafe fn` or the non-compile-passing `extern fn`.

nikomatsakis:
I do not think of `unsafe fn` as a distinct type from `fn` -- rather I think of "function pointers" as types that have a range of options, something like this (ignoring `for<...>`):

* `extern "ABI" unsafe? fn(A0..An) -> R`

alercah:
What about impls?

nikomatsakis:
@alercah 

> so perhaps we should just write "(e.g., fn())" instead.

what do you think about this? :)

alercah:
Ah, I see. That makes sense to me!

src/types/closure.md
```diff
@@ -131,8 +131,8 @@ more specific call traits:
 > not how it captures them.
 
 *Non-capturing closures* are closures that don't capture anything from their
-environment. They can be coerced to function pointers (`fn`) with the matching
-signature.
+environment. They can be coerced to function pointers (`fn` or `unsafe fn`)
```
nikomatsakis, line 134:
```suggestion
environment. They can be coerced to function pointers (e.g., `fn()`)
```

nikomatsakis:
@ehuss how do I merge, just click the green button? :)

ehuss:
Yes, after marking "approved" just click the merge button. 😃

PR #809, opened 2020-05-12
work on char/str descriptions

RalfJung:
Cc @ehuss (https://github.com/rust-lang/reference/pull/792)

### DIFF
--- a/src/types/textual.md
+++ b/src/types/textual.md
@@ -9,12 +9,12 @@ or 0xE000 to 0x10FFFF range. It is immediate [Undefined Behavior] to create a
 string of length 1.
 
 A value of type `str` is a Unicode string, represented as an array of 8-bit
-unsigned bytes holding a sequence of UTF-8 code points. Note that this is a
-library-level invariant: for the compiler and core language specification, `str`
-is the same as `[u8]`, but methods working on `str` may assume that the data in
-there is valid UTF-8 and may cause Undefined Behavior otherwise. Since `str` is
-a [dynamically sized type], it can only be instantiated through a pointer type,
-such as `&str`.
+unsigned bytes holding a sequence of UTF-8 encoded Unicode code points. Note
+that this is a library-level invariant: for the compiler and core language
+specification, `str` is the same as `[u8]`, but methods working on `str` may
+assume that the data in there is valid UTF-8 and may cause Undefined Behavior
+otherwise. Since `str` is a [dynamically sized type], it can only be
+instantiated through a pointer type, such as `&str`.
 
 [Unicode scalar value]: http://www.unicode.org/glossary/#unicode_scalar_value
 [Undefined Behavior]: ../behavior-considered-undefined.md
--- a/src/types/textual.md
+++ b/src/types/textual.md
@@ -8,13 +8,13 @@ or 0xE000 to 0x10FFFF range. It is immediate [Undefined Behavior] to create a
 `char` that falls outside this range. A `[char]` is effectively a UCS-4 / UTF-32
 string of length 1.
 
-A value of type `str` is a Unicode string, represented as an array of 8-bit
-unsigned bytes holding a sequence of UTF-8 encoded Unicode code points. Note
-that this is a library-level invariant: for the compiler and core language
-specification, `str` is the same as `[u8]`, but methods working on `str` may
-assume that the data in there is valid UTF-8 and may cause Undefined Behavior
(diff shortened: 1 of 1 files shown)


src/types/textual.md
```diff
@@ -2,15 +2,20 @@
...
-instantiated through a pointer type, such as `&str`.
+unsigned bytes holding a sequence of UTF-8 code points. Note that this is a
+library-level invariant: for the compiler and core language specification, `str`
+is the same as `[u8]`, but methods working on `str` may assume that the data in
+there is valid UTF-8 and may cause Undefined Behavior otherwise. Since `str` is
+a [dynamically sized type], it can only be instantiated through a pointer type,
+such as `&str`.
 
 [Unicode scalar value]: http://www.unicode.org/glossary/#unicode_scalar_value
+[Undefined Behavior]: ../behavior-considered-undefined.html
```
ehuss, line 20:
```suggestion
[Undefined Behavior]: ../behavior-considered-undefined.md
```

src/types/textual.md
```diff
@@ -2,15 +2,20 @@
...
+not a surrogate), represented as a 32-bit unsigned word in the 0x0000 to 0xD7FF
+or 0xE000 to 0x10FFFF range. It is immediate [Undefined Behavior] to create a
+`char` that falls outside this range. A `[char]` is effectively a UCS-4 / UTF-32
+string of length 1.
 
 A value of type `str` is a Unicode string, represented as an array of 8-bit
-unsigned bytes holding a sequence of UTF-8 code points. Since `str` is a
-[dynamically sized type], it is not a _first-class_ type, but can only be
-instantiated through a pointer type, such as `&str`.
+unsigned bytes holding a sequence of UTF-8 code points. Note that this is a
```
ehuss, line 12: I'm uncertain, but is the correct term maybe "UTF-8 code units"?

RalfJung: I don't know, I just kept the existing text here...

RalfJung: But Wikipedia says you are right

> UTF-8 (8-bit Unicode Transformation Format) is a variable width character encoding capable of encoding all 1,112,064[nb 1] valid character code points in Unicode using one to four one-byte (8-bit) **code units**.

RalfJung: Hm actually, "code points" might be correct here though -- that's the safety invariant. So I changed it ti "UTF-8 encoded Unicode code points" instead.

RalfJung:
@Havvy what about this?

src/types/textual.md
```diff
@@ -8,13 +8,13 @@ or 0xE000 to 0x10FFFF range. It is immediate [Undefined Behavior] to create a
...
-that this is a library-level invariant: for the compiler and core language
-specification, `str` is the same as `[u8]`, but methods working on `str` may
-assume that the data in there is valid UTF-8 and may cause Undefined Behavior
-otherwise. Since `str` is a [dynamically sized type], it can only be
-instantiated through a pointer type, such as `&str`.
+A value of type `str` is represented the same way as `[u8]`, it is a slice of
+8-bit unsigned bytes. However, the Rust standard library makes extra assumptions
+about `str`: methods working on `str` assume and ensure that the data in there
+is valid UTF-8. Calling a `str` method with a non-UTF-8 buffer can cause
+[Undefined Behavior] now or in the future. \
```
Havvy, line 15: Is the trailing backslash here on purpose? I can't remember what markdown does to it.

RalfJung: It's a newline without starting a new paragraph. I found like the next sentence is somewhat disconnected from the previous ones. You can see that in action by clicking the ["view file" link](https://github.com/rust-lang/reference/blob/ed7ad7290605b96b3c4415cbcd2afb16fce71cb6/src/types/textual.md) in the PR diff view.

I can remove it if you prefer.

Havvy: Ah, as long as it has a purpose, it looks fine to me. I recall there being discussion of changing to one sentence = one line anyways, although it never went anywhere.

RalfJung: > I recall there being discussion of changing to one sentence = one line anyways

That's different, that's for just the markdown file.

Havvy: Oh, I see on, uh, the third look at that file. It's subtle, and I don't see a difference in readability either way personally, but if you see it, I'm willing to trust a second set of eyes.

PR #810, opened 2020-05-12
Account for removal of UB in float-to-int casts

Mark-Simulacrum:
This updates the reference per PR rust-lang/rust#71269.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -366,9 +366,11 @@ same trait object.
         * zero-extend if the source is unsigned
         * sign-extend if the source is signed
     * Casting from a float to an integer will round the float towards zero
-        * **[NOTE: currently this will cause Undefined Behavior if the rounded
-          value cannot be represented by the target integer type][float-int]**.
-          This includes Inf and NaN. This is a bug and will be fixed.
+        * `NaN` will return `0`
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Sorry it took so long, @ehuss :)

src/expressions/operator-expr.md
```diff
@@ -366,9 +366,8 @@ same trait object.
         * zero-extend if the source is unsigned
         * sign-extend if the source is signed
     * Casting from a float to an integer will round the float towards zero
-        * **[NOTE: currently this will cause Undefined Behavior if the rounded
-          value cannot be represented by the target integer type][float-int]**.
-          This includes Inf and NaN. This is a bug and will be fixed.
+        * `NaN` will return `0` and values outside the integer range will
+          saturate to the integer minimum/maximum.
```
Havvy, line 370: We should state this more explicitly that values smaller than the integer range will saturate to the minimum value of the integer type and that values larger will saturate towards the maximum value of the integer type. Each one in its own bullet point.

I can do this as a follow-up PR if you don't want to.

Mark-Simulacrum:
Tried to correct both nits, also removed the leftover float-float link from 977499574e9e6241a4f719800b3be47ed40ef1a1.

Havvy:
💟 Thanks!

PR #812, opened 2020-05-14
cfg_attr needs a valid predicate

Tacklebox:
The example for cfg_attr needs a valid predicate. `linux` on it's own doesn't work. I think this likely slipped through because `windows` on it's own looks valid because it's a target_family which set their value like `unix` and `windows`. Not sure.


Havvy:
💟 Thanks!

PR #814, opened 2020-05-16
Update tuple index token.

### LINKED ISSUE #735: Tupple index vs integer literal
The [section about *integer literals*](https://doc.rust-lang.org/reference/tokens.html#integer-literals) got expanded in 8e79044b808a0e6b56b13285509d20b05892e472 to include *tupple inxex* but the surrounding text did not.

- It still says an *integer literal* has 4 forms instead of 5.

- It talks about suffixes but a *tupple index* can't have a suffix. It also discussed types but a *tupple index* probably doesn't really have a type since it only exists at compile time.

ehuss:
The tuple index is not a distinct token lexically, so don't treat it as such.

Also, add a note about the future-compatibility issue with suffixes.

Closes #735


matthewjasper:
Thanks!

PR #825, opened 2020-06-02
changed crate to super to fix compile error in the eg.

prejrdev:


src/visibility-and-privacy.md
```diff
@@ -134,8 +141,7 @@ pub mod submodule {
         }
     }
 }
-
-# fn main() {}
+fn main(){};
```
ehuss, line 144: These changes should be undone, it breaks the tests.

src/visibility-and-privacy.md
```diff
@@ -107,15 +107,22 @@ mod crate_helper_module {
...
+    use super::crate_helper_module;
+    use crate::rate_limit;
+    use super::public_api;
+    
     pub fn my_method() {
         // Any item in the local crate may invoke the helper module's public
         // interface through a combination of the two rules above.
         crate_helper_module::crate_helper();
+        public_api();
+        let doube_limit = rate_limit * 2;
```
ehuss, line 125: I'm not sure this really belongs here.  This example (especially this `my_method`) is here to illustrate the combination of the two rules above.  I'm not sure how this relates to those rules?

prejrdev: rate_limit is private only to the file itself and can't be accessed outside of it.  i added it as an example of that fact to contrast with the note about public_api()

src/visibility-and-privacy.md
prejrdev: i only added that because the previous version felt like the hashtag was made as a comment and the next example had it.

prejrdev: ok, changed and now it passes the tests.

src/visibility-and-privacy.md
```diff
@@ -89,9 +89,6 @@ Here's an example of a program which exemplifies the three cases outlined
 above:
 
 ```rust
-// This module is private, meaning that no external crate can access this
```
ehuss, line 92: Why was this removed?

src/visibility-and-privacy.md
```diff
@@ -134,7 +137,6 @@ pub mod submodule {
         }
     }
 }
-
```
ehuss, line 137: This whitespace change seems unnecessary.  It would probably be best to keep consistency with the rest of the file.

src/visibility-and-privacy.md
```diff
@@ -105,17 +102,23 @@ mod crate_helper_module {
 
 // This function is "public to the root" meaning that it's available to external
 // crates linking against this one.
-pub fn public_api() {}
+pub fn public_api(rate_limit:i32) {}
```
ehuss, line 105: It's probably best to use standard formatting.

```suggestion
pub fn public_api(rate_limit: i32) {}
```

src/visibility-and-privacy.md
```diff
@@ -105,17 +102,23 @@ mod crate_helper_module {
...
+
+//This is by default private, and it isn't available to external crates.
```
ehuss, line 107:
```suggestion
// This is by default private, and it isn't available to external crates.
```

src/visibility-and-privacy.md
```diff
@@ -105,17 +102,23 @@ mod crate_helper_module {
...
-
+    use super::crate_helper_module;
+    use super::MAX_RATE;
+    use super::public_api;
+    
     pub fn my_method() {
         // Any item in the local crate may invoke the helper module's public
         // interface through a combination of the two rules above.
         crate_helper_module::crate_helper();
+        public_api(MAX_RATE);
```
ehuss, line 121: It would probably be best to have a comment indicating what this is illustrating.

src/visibility-and-privacy.md
```diff
@@ -105,17 +102,23 @@ mod crate_helper_module {
 
 // This function is "public to the root" meaning that it's available to external
 // crates linking against this one.
-pub fn public_api() {}
+pub fn public_api(rate_limit:i32) {}
+
+//This is by default private, and it isn't available to external crates.
+const MAX_RATE:i32 = 32;
```
ehuss, line 108:
```suggestion
const MAX_RATE: i32 = 32;
```

ehuss:
I'm going to close since this is a couple years old.  Feel free to reopen if you want to address the above issues.

PR #830, opened 2020-06-09
Improve visibility and privacy page

camsteffen:
My main reason for making changes to this page was to add to this note in the part about "helper modules":

> Note: Consider marking items with pub(crate) to ensure that they are not re-exported (see below).

You can see my Stack Overflow [question](https://stackoverflow.com/questions/62205451/where-is-my-private-code-exposed-as-public) for background.

But then I found some other things to tweak just to improve readability. Some phrases I just removed that seemed irrelevant or redundant. If any of these changes seem ill-justified, I can reverse them.

### DIFF
--- a/src/visibility-and-privacy.md
+++ b/src/visibility-and-privacy.md
@@ -8,85 +8,84 @@
 > &nbsp;&nbsp; | `pub` `(` `super` `)`\
 > &nbsp;&nbsp; | `pub` `(` `in` [_SimplePath_] `)`
 
-These two terms are often used interchangeably, and what they are attempting to
-convey is the answer to the question "Can this item be used at this location?"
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  In general, the reference is trying to lean more towards a spec-like language, and less of a user guide. That means using precise language, and less "here's how to use this".  I understand the current visibility page is written more in the guide style, and there is a bit of a problem of not having a specific "guide" (aside from TRPL).  It's a problem that we don't really have an answer for now.

There definitely seems to be a few improvements here, but it also seems to remove content which I'm concerned about.  Perhaps @Havvy has some thoughts here?

camsteffen:
I don't see this PR gaining any traction.

PR #831, opened 2020-06-09
fix out of date info on type aliases

yaahc:
The current version of the reference for `type-aliases` currently claims that you cannot qualify constructors for enums through type aliases, but this is no longer true. AFAIK the only type of constructor that type aliases can't be used for are `tuple-structs`. tuple-enums, non-tuple-enums, and non-tuple-structs all construct just fine through type aliases.

### DIFF
--- a/src/items/type-aliases.md
+++ b/src/items/type-aliases.md
@@ -20,9 +20,9 @@ type Point = (u8, u8);
 let p: Point = (41, 68);
 ```
 
-A type alias to an tuple-struct type cannot be used to qualify the constructors:
+A type alias to a tuple-struct or unit-struct cannot be used to qualify that type's constructor:
 
-```rust
+```rust,edition2018,compile_fail
 pub struct MyStruct(u32);
 
 pub use self::MyStruct as PubUse;
@@ -30,7 +30,7 @@ pub type PubType = MyStruct;
 
 fn main() {
     let _ = PubUse(5); // OK
-    // let _ = PubType(5); // Doesn't work
+    let _ = PubType(5); // Doesn't work
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/items/type-aliases.md
```diff
@@ -20,13 +20,18 @@ type Point = (u8, u8);
 let p: Point = (41, 68);
 ```
 
-A type alias to an enum type cannot be used to qualify the constructors:
+A type alias to an tuple-struct type cannot be used to qualify the constructors:
```
ehuss, line 23: I think this applies to both tuple-structs and unit structs, correct?

src/items/type-aliases.md
```diff
@@ -20,13 +20,18 @@ type Point = (u8, u8);
...
-// let _: F = F::A;  // Doesn't work
+pub struct MyStruct(u32);
+
+pub use self::MyStruct as PubUse;
+pub type PubType = MyStruct;
+
+fn main() {
+    let _ = PubUse(5); // OK
+    // let _ = PubType(5); // Doesn't work
+}
```
ehuss, line 34: I think it is good when demonstrating something that doesn't work to actually include the broken code, and mark it as `compile_fail`.  I would also avoid the `main` function and the `pub` qualifications, since they aren't really relevant here.  Maybe something like this:

    ```rust,edition2018,compile_fail
    struct MyStruct(u32);
    use MyStruct as PubUse;
    type PubType = MyStruct;

    let _ = PubUse(5); // OK
    let _ = PubType(5); // ERROR: found type alias `PubType`
    ```

yaahc: I think that the `pub` usage in this case is actually slightly relevant because this is most relevant when dealing with renaming types without making a breaking change to your API, so the type being exported is the primary reason this information is important.

ehuss: I'm not sure I follow.  Type aliases are used in many circumstances unrelated to API compatibility.  I see the primary reason for this example is to demonstrate how tuple struct constructors are not aliased with the `type` item, not to illustrate various use cases for aliases.

PR #834, opened 2020-06-15
Document min pointer width.

### LINKED ISSUE #643: Document that isize/usize are at least 16-bit wide on all platforms we currently support
See this UCG PR: https://github.com/rust-lang/unsafe-code-guidelines/pull/154

@briansmith raised the issue that this guarantee would be more useful if it was documented elsewhere and they are right. We should document that here, in the `isize`/`usize` API docs, and maybe also on the trait impls justified by this observation, like the `From<u16>` impl for `usize`, for example.

ehuss:
Closes #643.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -169,11 +169,13 @@ on the endianness of the target's CPU.
 
 ### `target_pointer_width`
 
-Key-value option set once with the target's pointer width in bits. For example,
-for targets with 32-bit pointers, this is set to `"32"`. Likewise, it is set
-to `"64"` for targets with 64-bit pointers.
+Key-value option set once with the target's pointer width in bits.
 
-<!-- Are there targets that have a different bit number? -->
+Example values:
+
+* `"16"`
+* `"32"`
+* `"64"`
 
 ### `target_vendor`
 
--- a/src/types/numeric.md
+++ b/src/types/numeric.md
@@ -40,3 +40,8 @@ differences between pointers into an object or array and can address every byte
 within an object along with one byte past the end.
 
 `usize` and `isize` are at least 16-bits wide.
+
+> **Note**: Many pieces of Rust code may assume that pointers, `usize`, and
+> `isize` are at least 32-bit and at most 64-bit. As a consequence, 16-bit
+> pointer support is limited and may require explicit care and acknowledgment
+> from a library to support.
--- a/src/types/numeric.md
+++ b/src/types/numeric.md
@@ -38,3 +38,5 @@ platform's pointer type. The theoretical upper bound on object and array size
 is the maximum `isize` value. This ensures that `isize` can be used to calculate
 differences between pointers into an object or array and can address every byte
 within an object along with one byte past the end.
+
+`usize` and `isize` are at least 16-bits wide.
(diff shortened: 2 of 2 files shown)


ehuss:
If the lang team is not ready to make the min-size commitment, I can easily just remove the addition to `numeric.md`.  I think highlighting that `16` is a valid option to `target_pointer_width` is a good thing to add regardless.

joshtriplett:
I don't think we should add this without a very large caveat that it's *entirely reasonable* for 16-bit pointers to have limited support, and that many libraries may assume that usize is at least 32-bit. 16-bit pointer support is something I'd expect to require explicit care and acknowledgement of in a library crate that supports it, much like `no_std` support.

From a language point of view, yes, `usize` and `isize` are at least 16 bits. But I think there's value in at least a *descriptive* statement like "Note that many pieces of Rust code may assume that pointers and `usize`/`isize` are at least 32-bit and at most 64-bit."

ehuss:
Thanks for taking a look!  I've updated with a note.

src/conditional-compilation.md
```diff
@@ -169,11 +169,13 @@ on the endianness of the target's CPU.
...
-for targets with 32-bit pointers, this is set to `"32"`. Likewise, it is set
-to `"64"` for targets with 64-bit pointers.
+Key-value option set once with the target's pointer width in bits.
 
-<!-- Are there targets that have a different bit number? -->
+Example values:
+
+* `"16"`
+* `"32"`
+* `"64"`
```
chorman0773, line 178: Would this allow targets where pointers are really 24-bits, such as the 65816?
Technically speaking, I forcibly extend pointers to 32-bit, but they only have 24 significant bits.

joshtriplett: There are two separate questions to answer here, I think.

First, whether pointer size here refers to the "significant bits" or the actual size of a pointer value. I think it refers to the latter, so even if you only have 24 significant bits, if you extend pointers to 32-bit then the pointer size is 32-bit.

But if you *don't* extend pointers, then that raises an interesting question. I don't see a fundamental reason why (for instance) we couldn't support systems with 80-bit pointers and 80-bit `usize`, or why it would *have* to be rounded up to 128-bit, as long as LLVM (or whatever backend we used for the platform) supported those 80-bit values.

If we add mentions of such cases, though, then I think we'd need to change the clarifying note I suggested to say that "Many pieces of Rust code may assume that pointers, `usize`, and `isize` are either 32-bit or 64-bit". Because a system with 40-bit or 48-bit or 56-bit pointers would likely run into many of the same difficulties and limited library support as a system with 16-bit or 24-bit or 80-bit or 128-bit pointes.

src/types/numeric.md
```diff
@@ -40,3 +40,8 @@ differences between pointers into an object or array and can address every byte
 within an object along with one byte past the end.
 
 `usize` and `isize` are at least 16-bits wide.
+
+> **Note**: Many pieces of Rust code may assume that pointers, `usize`, and
+> `isize` are at least 32-bit and at most 64-bit. As a consequence, 16-bit
```
joshtriplett, line 45:
```suggestion
> `isize` are either 32-bit or 64-bit. As a consequence, 16-bit
```

Based on the discussion with @chorman0773. I think it's safe to say that these kinds of assumptions in the ecosystem would create similar difficulties for a target with pointer size *between* 32 and 64 bits.

src/conditional-compilation.md
chorman0773: I extend pointer size to a multiple of the alignment, 4-bytes. However I treat the 4th byte as padding, rather than treat it as part of the value.

Issue #799, opened 2020-04-16
If let and while let examples of irrefutable patterns should mention warning


alercah:
Both [`if let`][1] and [`while let`][2] have examples with irrefutable patterns. By default this is a compiler warning, so they make bad examples, or at least deserve mention of the compiler warning.

There should also probably be more links on these pages to the definition of irrefutable.

[1]: https://doc.rust-lang.org/nightly/reference/expressions/if-expr.html#if-let-expressions
[2]: https://doc.rust-lang.org/nightly/reference/expressions/loop-expr.html#predicate-pattern-loops

rimutaka:
As per discussion with @ehuss (https://github.com/rust-lang/reference/pull/835#discussion_r451035016)
> In general, the reference does not cover rustc lints, so I don't think it is appropriate to add this.

If that is true, this issue should be closed making PR #835 redundant.

traviscross:
For the reasons mentioned there, we'll go ahead and close this.

PR #835, opened 2020-06-21
Added a compiler warning and explanations for irrefutable patterns.

rimutaka:
In response to issue https://github.com/rust-lang/reference/issues/799

I am very new to Rust (10 days), so it needs to be reviewed for correctness.
I got the compiler warning from running the code and the irrefutability requirements from reading:

* https://doc.rust-lang.org/reference/expressions/closure-expr.html
* https://doc.rust-lang.org/reference/items/functions.html
* https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops
* https://doc.rust-lang.org/reference/statements.html#let-statements

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -111,17 +111,7 @@ if let (a, 3) = (1, 2) {           // "(a, 3)" is refutable, and will not match
 }
 ```
 
-Using an irrefutable pattern in `match`, `if let` or `while let` expressions produces a compiler warning.
-
-Example:
-```rust
-| /     if let (x, y) = (1, 2) {
-| |         print!("an irrefutable pattern, x={}, y={}", x, y)
-| |     }
-| |_____^
-|
-= note: `#[warn(irrefutable_let_patterns)]` on by default
-```
+Using an irrefutable pattern in `match`, `if let` or `while let` expressions produces a compiler warning. E.g. `if let (x, y) = (1, 2) { ... }` results in this compiler message `#[warn(irrefutable_let_patterns)]`.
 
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -111,6 +111,10 @@ if let (a, 3) = (1, 2) {           // "(a, 3)" is refutable, and will not match
 }
 ```
 
+Using an irrefutable pattern in `match`, `if let` or `while let` expressions produces a compiler warning. E.g. `if let (x, y) = (1, 2) { ... }` results in this compiler message `#[warn(irrefutable_let_patterns)]`.
```
ehuss, line 114: In general, the reference does not cover `rustc` lints, so I don't think it is appropriate to add this.

src/patterns.md
```diff
@@ -111,6 +111,10 @@ if let (a, 3) = (1, 2) {           // "(a, 3)" is refutable, and will not match
...
+
+Patterns used in `let` statements, `for` expressions, *Function* and *closure* parameters are always irrefutable.
```
ehuss, line 116: Typically we link each phrase to the appropriate section.  For example, "`let` statements" should link to statements.md#let-statements.  There are links to all of these above, so moving those to reference-style links would probably be a good idea.

Also, I think the `F` in function shouldn't be capitalized.

I also might use the phrase "must be irrefutable" instead of "are always".

src/patterns.md
rimutaka: Eric, in that case you should probably close issue #799 making this PR redundant.

traviscross:
Going through some backlog... for the reasons discussed above, we're going to just close this.  Thanks for filing the PR.

PR #837, opened 2020-06-24
Mention Box's partial field moves

alercah:


src/expressions/field-expr.md
```diff
@@ -30,7 +30,8 @@ Finally, the fields of a struct or a reference to a struct are treated as
 separate entities when borrowing. If the struct does not implement
 [`Drop`](../special-types-and-traits.md#drop) and is stored in a local variable,
 this also applies to moving out of each of its fields. This also does not apply
-if automatic dereferencing is done though user defined types.
+if automatic dereferencing is done though user defined types other than
+[`Box`](..//special-types-and-traits.html#boxt).
```
Havvy, line 34:
```suggestion
[`Box`](../special-types-and-traits.html#boxt).
```

Extra `/`?

Havvy:
`Box` is the exception here yes? If so, it should probably say "with the exception of Box" rather than "other than". The "other than" feels like a double negation of the negative at the start making it hard to read.

And yes, `Box` can be considered a user-defined type. It's very special, but it's still technically a struct.

Havvy (approved):

Havvy:
💟 Thanks!

PR #842, opened 2020-06-30
Update const eval chapter to latest changes

oli-obk:


src/const_eval.md
```diff
@@ -45,6 +43,8 @@ to be run.
 * [Cast] expressions, except pointer to address and
   function pointer to address casts.
 * Calls of [const functions] and const methods.
+* [loop], [while] and `while let` expressions.
```
ehuss, line 46:
```suggestion
* [loop], [while] and [`while let`] expressions.
```

oli-obk:
> Should the "const context" list be updated to include "The body of a `const fn`"?

hmm... good question. It isn't really, a const fn can be invoked in a const context, and its body has a bunch of rules it needs to uphold, but these differ from the current list of const context's rules in some small and some large ways.

src/const_eval.md
```diff
@@ -57,6 +57,22 @@ A _const context_ is one of the following:
...
 
+## Const Functions
+
+A _const fn_ is a function that one is permitted to call from a const context. Declaring a function
+`const` has no effect on any existing uses, it only restricts the types that arguments and the
+return type may use, as well as prevent various expressions from being used within it.
+
+Notable features that const contexts have, but const fn haven't are:
+
+* floating point types
```
ehuss, line 68: I'm a bit confused on this point. It seems that floating point types themselves are OK, it's floating-point operations that seem to be restricted?  That is, the following seems to be ok:

```rust
const fn foo(x: f32) -> f32 {
    x
}
```

Is it referring to something else?

oli-obk: ah yes, float ops are forbidden, not values

ehuss:
@oli-obk can you update the point on floating point types to clarify it is for floating point operations (and preferably which operations)?

oli-obk:
Thanks for the ping! I cleaned up a bunch of things, because there were more changes to the features in the language since I last edited this PR

src/const_eval.md
```diff
@@ -57,6 +59,22 @@ A _const context_ is one of the following:
...
+
+Notable features that const contexts have, but const fn haven't are:
+
+* floating point operations
+  * floating point values are treated just like generic parameters without trait bounds beyond
+  `Copy`. So you cannot do anything with them but copy/move them around.
+* `dyn Trait` types
+* generic bounds on generic parameters beyond `Sized`
+* comparing raw pointers
+* union field access or `transmute` invocations.
```
ehuss, line 76:
```suggestion
* union field access
* `transmute` invocations
```

These two points don't seem to be related to one another, are they?

Also, would you mind linking `transmute` to the library docs?  I think a link to `../core/intrinsics/fn.transmute.html` would be appropriate.

oli-obk: well... to me they are, because I was using them for the same thing, but you're right, that view is likely not what anyone else has on it.

PR #843, opened 2020-06-30
Improve Type-Coersion Documentation

### LINKED ISSUE #734: Examples of coercions are wrong
Several of the examples on [the coercions page](https://doc.rust-lang.org/stable/reference/type-coercions.html) claim that 42 gets coerced into i8. However, no conversion takes place, the 42 is i8 already thanks to type inference.

Potentially related PR: #342

AloeareV:
I've updated pull request #342 by @Havvy by fixing the merge conflicts, and additionally adding a fix to close #734 (Incorrect examples of type coercion).

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -4,11 +4,10 @@
 They happen automatically at specific locations and are highly restricted in 
 what types actually coerce.
 
-The [type cast operator], `as`, is not a coersion site. However, most type
-conversions allowed by coersion can also be explicitly performed by `as`.
+Any conversions allowed by coercion can also be explicitly performed by the
+[type cast operator], `as`.
 
 Coercions are originally defined in [RFC 401] and expanded upon in [RFC 1558].
-
 
 ## Coercion sites
 
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -4,7 +4,11 @@
 They happen automatically at specific locations and are highly restricted in 
 what types actually coerce.
 
+The [type cast operator], `as`, is not a coersion site. However, most type
+conversions allowed by coersion can also be explicitly performed by `as`.
+
 Coercions are originally defined in [RFC 401] and expanded upon in [RFC 1558].
+
 
 ## Coercion sites
 
@@ -64,10 +68,6 @@ sites are:
       x
   }
   ```
-
-* The [as] type cast operator&mdash;while not a coercion site&mdash;can be used to 
-  explicitly perform type coercion.
-
 
 If the expression in one of these coercion sites is a coercion-propagating
 expression, then the relevant sub-expressions in that expression are also
@@ -190,6 +190,6 @@ unsized coercion to `Foo<U>`.
 [RFC 1558]: https://github.com/rust-lang/rfcs/blob/master/text/1558-closure-to-fn-coercion.md
 [subtype]: subtyping.md
 [object safe]: items/traits.md#object-safety
-[as]: expressions/operator-expr.md#type-cast-expressions
+[type cast operator]: expressions/operator-expr.md#type-cast-expressions
 [`Unsize`]: ../std/marker/trait.Unsize.html
 [`CoerceUnsized`]: ../std/ops/trait.CoerceUnsized.html
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -52,7 +52,7 @@ sites are:
       Foo { x: &mut 42 };
   }
   ```
-  
+ 
 * Function results&mdash;either the final line of a block if it is not
   semicolon-terminated or any expression in a `return` statement
 
@@ -168,7 +168,7 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `[T; n]` to `[T]`.
 
-* `T` to `dyn U`, when `T` implements `U + Sized`
+* `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
 
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
     * `Foo` is a struct.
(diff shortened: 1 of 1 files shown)


src/type-coercions.md
```diff
@@ -15,55 +15,59 @@ sites are:
...
 
   fn main() {
-      Foo { x: 42 };
+      Foo { x: &mut 42 };
   }
   ```
-
-* Function results, either the final line of a block if it is not
+  
+* Function results&ndash;either the final line of a block if it is not
```
ehuss, line 56:
```suggestion
* Function results&mdash;either the final line of a block if it is not
```

src/type-coercions.md
```diff
@@ -15,55 +15,59 @@ sites are:
...
+  struct Foo<'a> { x: &'a i8 }
 
   fn main() {
-      Foo { x: 42 };
+      Foo { x: &mut 42 };
   }
   ```
-
-* Function results, either the final line of a block if it is not
+
```
ehuss, line 55: Try to avoid adding trailing whitespace

src/type-coercions.md
```diff
@@ -15,55 +15,59 @@ sites are:
...
   ```rust
-  fn foo() -> i8 {
-      42
+  use std::fmt::Display;
+  fn foo(x: &u32) -> &dyn Display {
+      x
   }
   ```
 
+* The [as] type cast operator can also explicitly perform type coersion.
```
ehuss, line 68:
```suggestion
* The [as] type cast operator can also explicitly perform type coercion.
```

I also feel like `as` was originally not included here because it isn't quite the same as a "coercion site".  What is the reason for adding it?

src/type-coercions.md
```diff
@@ -91,7 +95,7 @@ the block has a known type.
 
 Coercion is allowed between the following types:
 
-* `T` to `U` if `T` is a subtype of `U` (*reflexive case*)
+* `T` to `U` if `T` is a [subtype] of `U` (*reflexive case*)
```
ehuss, line 98: There seems to be an unanswered question here: https://github.com/rust-lang/reference/pull/342/files#r191050490

src/type-coercions.md
```diff
@@ -164,8 +168,7 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `[T; n]` to `[T]`.
 
-* `T` to `U`, when `U` is a trait object type and either `T` implements `U` or
-  `T` is a trait object for a subtrait of `U`.
+* `T` to `dyn U`, when `T` implements `U + Sized`
```
ehuss, line 171: There seems to be an unanswered question here: https://github.com/rust-lang/reference/pull/342/files#r191050567

src/type-coercions.md
AloeareV: I've had little experience with the term 'reflexive' in programming contexts, but in math, "is a subset of" is a reflexive relationship. I assume that "is a subtype of" would be the programming equivalent.

src/type-coercions.md
AloeareV: Ah. Trait Objects can only be created from Sized types. That said, that's an incomplete restriction, I'll add a line about object safety.

src/type-coercions.md
AloeareV: Didn't mean to resolve this. It's not a coercion site, but it's a connected concept with a lot of overlap, and I think it's worthy of a mention.

AloeareV: It might be unclear of me to list it with the coersion sites, I can add a clause about "While not a coersion site..." or maybe move it to the introduction.

AloeareV:
@ehuss What's the verdict on this?

src/type-coercions.md
ehuss: Since this list is explicitly a list of coercion sites, I don't think it should include something that is explicitly not a coercion site. Can you move it up to the introduction?  Just say something like, `For explicit type coercions, see [type cast expressions].`.

src/type-coercions.md
ehuss: Ah. I'm not super familiar with it, but since this is just adding a link, I won't hold up the PR on it.  AIUI, the reflexive subtype would be "T is a subtype of T", not U (in the math sense, T is a subset of T).

src/type-coercions.md
AloeareV: Done!

src/type-coercions.md
AloeareV: I guess.. Subtyping is an example of a reflexive relationship, because types are subtypes of themselves. So.. That means that a subtype coersion is reflexive, because it could be coercing to Self. That sounds really finicky now that I say it.

src/type-coercions.md
```diff
@@ -1,10 +1,14 @@
 # Type coercions
 
-Coercions are defined in [RFC 401]. [RFC 1558] then expanded on that.
-A coercion is implicit and has no syntax.
+**Type coercions** are implicit operations that change the type of a value.
+They happen automatically at specific locations and are highly restricted in 
+what types actually coerce.
+
+The [type cast operator], `as`, is not a coersion site. However, most type
+conversions allowed by coersion can also be explicitly performed by `as`.
```
ehuss, line 8: The phrasing here seems a little strange to me.  Nothing suggests or implies that `as` is a coercion site. The equivalence doesn't make sense because `as` is an expression, not a location. It also says "most type conversions...", but then doesn't say which are allowed. 

I think it would be best just to keep this simple and to the point.  I still suggest my original wording, something like `For explicit type coercions, see [type cast expressions].`.

Also, the word is spelled "coercion" with a `c`.

src/type-coercions.md
```diff
@@ -1,10 +1,14 @@
...
+
+Coercions are originally defined in [RFC 401] and expanded upon in [RFC 1558].
```
ehuss, line 11: Try to avoid double blank lines.

src/type-coercions.md
AloeareV: I guess I'm trying to make the distinction that, if it's explicit, it's not a coercion. You're right that that wording is funky, though.

PR #852, opened 2020-07-17
clarify when reading uninititalized memory is allowed

RalfJung:
This should hopefully help avoid confusion like [this one](https://internals.rust-lang.org/t/why-even-unused-data-needs-to-be-valid/12734/23?u=ralfjung).

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -57,6 +57,11 @@ code.
     > **Note**: `rustc` achieves this with the unstable
     > `rustc_layout_scalar_valid_range_*` attributes.
 
+Note that uninitialized memory is also implicitly invalid for any type that has
+a restricted set of valid values.  In other words, the only cases in which
+reading uninitialized memory is permitted is inside `union`s, and between the
+fields of a compound type (in the "padding").
+
 A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -59,7 +59,7 @@ code.
 
 **Note:** Uninitialized memory is also implicitly invalid for any type that has
 a restricted set of valid values. In other words, the only cases in which
-reading uninitialized memory is permitted is inside `union`s and in "padding"
+reading uninitialized memory is permitted are inside `union`s and in "padding"
 (the gaps between the fields/elements of a type).
 
 A reference/pointer is "dangling" if it is null or not all of the bytes it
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -57,6 +57,11 @@ code.
     > **Note**: `rustc` achieves this with the unstable
     > `rustc_layout_scalar_valid_range_*` attributes.
 
+Note that uninitialized memory is also implicitly invalid for any type that has
+a restricted set of valid values.  In other words, the only cases in which
```
Havvy, line 61: Extra space.

```suggestion
a restricted set of valid values. In other words, the only cases in which
```

src/behavior-considered-undefined.md
```diff
@@ -57,6 +57,11 @@ code.
     > **Note**: `rustc` achieves this with the unstable
     > `rustc_layout_scalar_valid_range_*` attributes.
 
+Note that uninitialized memory is also implicitly invalid for any type that has
```
Havvy, line 60:
```suggestion
**Note:** Uninitialized memory is also implicitly invalid for any type that has
```

src/behavior-considered-undefined.md
```diff
@@ -57,6 +57,11 @@ code.
...
+a restricted set of valid values.  In other words, the only cases in which
+reading uninitialized memory is permitted is inside `union`s, and between the
```
Havvy, line 62: You can probably just minimize the second part to just "padding". We also don't define "compound type" anywhere, so it's probably best not to use it. We do define "user defined type" for struct/enum/union though.

Havvy, line 62:
```suggestion
reading uninitialized memory is permitted is inside `union`s and between the
```

A list of two items doesn't take a comma.

RalfJung: It's all "types with fields" though that I mean by "compound type" -- so, including tuples and arrays.

RalfJung:
> Mostly nits. Only real concern is "compound types" since it's not a term used elsewhere in the reference.

I tried rewording that, what do you think?

src/behavior-considered-undefined.md
```diff
@@ -57,6 +57,11 @@ code.
     > **Note**: `rustc` achieves this with the unstable
     > `rustc_layout_scalar_valid_range_*` attributes.
 
+**Note:** Uninitialized memory is also implicitly invalid for any type that has
+a restricted set of valid values. In other words, the only cases in which
+reading uninitialized memory is permitted is inside `union`s and in "padding"
```
Havvy, line 62:
```suggestion
reading uninitialized memory is permitted are inside `union`s and in "padding"
```

RalfJung:
@Havvy I don't have merge permissions in this repo.

Havvy:
💟 Merged.

PR #854, opened 2020-07-21
Added Unpin to list of Auto Traits

### LINKED ISSUE #680: Unpin being an auto trait is not documented in list of auto traits
As documented in the [`Unpin`](https://doc.rust-lang.org/std/marker/trait.Unpin.html) stdlib docs, `Unpin` "is automatically implemented for almost every type." However, [the "Auto Traits" section](https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits) does not list it.

AloeareV:
Closes #680

### DIFF
--- a/src/special-types-and-traits.md
+++ b/src/special-types-and-traits.md
@@ -98,7 +98,7 @@ immutable [`static` items].
 
 ## Auto traits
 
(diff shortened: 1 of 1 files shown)


matthewjasper:
Thanks

PR #863, opened 2020-07-29
Metavariables can be keywords

jonas-schievink:


ehuss:
Thanks!  Do you happen to know the history here?  It looks like this changed in 1.17 (though there weren't any notes about it).  I'm guessing it was via https://github.com/rust-lang/rust/pull/39419?

It looks like this also allows raw identifiers.  Example:

```rust
macro_rules! foo {
    ($r#fn:expr) => {$r#fn};
}
```

which seems a little odd to me. 🤷   Maybe the "identifiers" page should have a production for whatever "ident or keyword or raw keyword" should be called?

Dunno if this is actually intended.

jonas-schievink:
I have no idea about the history here, sorry. I just noticed that rust-analyzer does some gymnastics to also allow keywords here, and realized that rustc also allows them.

jonas-schievink:
> Maybe the "identifiers" page should have a production for whatever "ident or keyword or raw keyword" should be called?

Is there any place in Rust where a normal ident is allowed but a raw ident isn't?

ehuss:
> Is there any place in Rust where a normal ident is allowed but a raw ident isn't?

[`LIFETIME_OR_LABEL`](https://doc.rust-lang.org/nightly/reference/tokens.html?highlight=lifetime_or_label#lifetimes-and-loop-labels)

mattheww:
This change was later made in #1130.

ehuss:
Indeed, I think this can be closed, thanks!

Issue #751, opened 2020-02-09
Update grammar for new fn grammar


ehuss:
Update for https://github.com/rust-lang/rust/pull/68764 and https://github.com/rust-lang/rust/pull/68788 and https://github.com/rust-lang/rust/pull/69023 and https://github.com/rust-lang/rust/pull/68728

Centril:
Also: https://github.com/rust-lang/rust/pull/69194

PR #864, opened 2020-07-30
Allow trait inner attributes

matklad:
r? @ehuss


ehuss:
Thanks!  This is a tiny part of #751 (from https://github.com/rust-lang/rust/pull/68728).  I believe there are much larger changes to the grammar needed here, but this seems good on its own.

The link is broken, though.  Would you mind adding `[_InnerAttribute_]: ../attributes.md` to the bottom of the file?

PR #867, opened 2020-08-03
Some constant/static updates.

### LINKED ISSUE #22: Clarify the meaning of reference-typed constants
Migrated from https://github.com/rust-lang/rust/issues/34817

---------------------------------------------------------------------

I used `pub const FOO: &'static Bar = &Bar { // crate-private stuff here ... };` in a crate. Within that crate, `FOO as *const Bar` was consistently the same address. However, when used in another crate, `FOO as *const Bar` became another address.

Apparently what's happening is that the RHS of `const` is inlined but within a given crate (effectively allowing a crate to inline stuff that's private to another crate), duplicates are later coalesced when making the values being referred to into the data section of the object file.

What I thought was supposed to be happening was this:

I expected the `Bar { // crate-private stuff here ... }` bit to be evaluated once in the crate where the constant was declared and the result getting uniquely baked into the data section of the object file. I then expected the value that gets inlined at contant use time to be the address of that unique location.

Especially considering that this appears to result in a constant address within a crate but blows up when the constant is used in another crate, it would be useful for the Book and the Reference to spell out more clearly that the constant isn't just the reference value of the right-hand-side expression but the entire right-hand-side expression itself and explicitly warn about the instantiations of the constant expression getting coalesced on a per-crate basis, which gives the appearance of the reference being the value of a uniquely evaluated expression rather than the expression itself as long as you are only testing within a single crate.

### LINKED ISSUE #582: Array type should say size expression is a *const* expression.
The array size expression needs to be a const expression (I believe). Currently types/array.md only says that it needs to evaluate to `usize`.

ehuss:
- Clarify that array length is a "constant expression". Closes #582.
- Remove restriction that statics must use other statics by reference. This was removed in https://github.com/rust-lang/rust/pull/51110 (1.29 I believe).
- Clarify that paths to statics are allowed only within static initializers.
- Add a clarification of usage of constants from external crates (closes #22 I believe).
- Mention that generic parameters cannot be used in a const context (https://github.com/rust-lang/rust/issues/43408)
- Some other minor tweaks.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -20,8 +20,9 @@ also constant expressions and do not cause any [`Drop::drop`][destructors] calls
 to be run.
 
 * [Literals].
-* [Paths] to [functions] and constants.
+* [Paths] to [functions] and [constants].
   Recursively defining constants is not allowed.
+* Paths to [statics]. These are only allowed within the initializer of a static.
 * [Tuple expressions].
 * [Array expressions].
 * [Struct] expressions.
@@ -53,7 +54,7 @@ to be run.
 A _const context_ is one of the following:
 
 * [Array type length expressions]
-* Repeat expression length expressions
+* [Array repeat expressions][array expressions]
 * The initializer of
   * [constants]
   * [statics]
@@ -75,6 +76,10 @@ Notable features that const contexts have, but const fn haven't are:
 * comparing raw pointers
 * union field access
 * [`transmute`] invocations.
+
+Conversely, the following are possible in a const function, but not in a const context:
+
+* Use of generic parameters.
 
 [arithmetic]:           expressions/operator-expr.md#arithmetic-and-logical-binary-operators
 [array expressions]:    expressions/array-expr.md
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -54,7 +54,7 @@ to be run.
 A _const context_ is one of the following:
 
 * [Array type length expressions]
-* [Array repeat expressions][array expressions]
+* [Array repeat length expressions][array expressions]
 * The initializer of
   * [constants]
   * [statics]
--- a/src/items/constant-items.md
+++ b/src/items/constant-items.md
@@ -7,11 +7,12 @@
 A *constant item* is an optionally named _[constant value]_ which is not associated
 with a specific memory location in the program. Constants are essentially inlined
 wherever they are used, meaning that they are copied directly into the relevant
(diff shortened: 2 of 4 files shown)


ehuss:
@oli-obk Do you think you'd be able to review this?

src/const_eval.md
```diff
@@ -53,7 +54,7 @@ to be run.
 A _const context_ is one of the following:
 
 * [Array type length expressions]
-* Repeat expression length expressions
+* [Array repeat expressions][array expressions]
```
oli-obk, line 57: Specifically only the length field of the repeat expression

src/const_eval.md
```diff
@@ -76,6 +77,10 @@ Notable features that const contexts have, but const fn haven't are:
 * union field access
 * [`transmute`] invocations.
 
+Conversely, the following are possible in a const function, but not in a const context:
+
+* Use of generic parameters.
```
oli-obk, line 82: that's not possible in `const fn` yet either.

src/items/constant-items.md
```diff
@@ -7,11 +7,12 @@
 A *constant item* is an optionally named _[constant value]_ which is not associated
 with a specific memory location in the program. Constants are essentially inlined
 wherever they are used, meaning that they are copied directly into the relevant
-context when used. References to the same constant are not necessarily
-guaranteed to refer to the same memory address.
+context when used. This includes usage of constants from external crates.
+References to the same constant are not necessarily guaranteed to refer to the
+same memory address.
```
oli-obk, line 12: Maybe also mention that this holds for `!Copy` types. So you can have a `const FOO: String = String::new()` and copy that around as many times as you want.

src/const_eval.md
ehuss: Oops, sorry!  I probably shouldn't open PRs late at night.

src/const_eval.md
ehuss: Hm, maybe I'm a bit confused. The following seems to work for me:

```rust
const fn cfunc<T>(x: &T) -> usize {
    std::mem::size_of::<T>()
}
```

oli-obk: ah, well, yes. I was thinking of trait bounds on generic parameters.

ehuss: I think the restriction on trait bounds is already mentioned above.  Let me know if you think this should be worded any differently.

ehuss:
@Havvy can you approve this?

PR #874, opened 2020-08-11
Add an extra fn() entry to the variance table in the subtyping chapter

Kimundi:
While discussing variance on a Rust help channel, I noticed that the subtyping chapter does not specifically mention the special case where a type parameter is used both in the input and output of a function, which, as far as I know, would make it be treated as invariant because its the intersection of both contravariance and covariance.

In case I'm not totally wrong with this, I propose adding that case to the table to make it a bit more clear.


nikomatsakis (approved):
correct

Havvy:
💟 Thanks!

PR #246, opened 2018-02-17
[RFC 2195] Document new type representations

Havvy:
This PR does two things:

1. Documents RFC 1240.
2. Splits the Enum sections for the C and Primitive representations into field-less enums and enums with fields.

Incidentally, fixes #244

PR #879, opened 2020-08-21
Document the enum changes in RFC 2195

### LINKED ISSUE #244: Replace remaining "C-like enumerations" with "fieldless enumerations"

### LINKED ISSUE #786: enum documentation not updated for RFC 2195
https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations talks about "Custom Discriminant Values for Field-Less Enumerations", but does not mention at all that since https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md Rust also supports custom discriminant values and defined layout for enums with fields.

poliorcetics:
Document the changes introduced in https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md.

Fixes #244 (I believe), fixes #786.

This is your RFC so I'll tag you @Gankra.

This is a restart from #246, applying the comments that were given at the time.

Disclaimer: I have never personally used this feature and only tested using the playground so I may be missing things, I'm sorry if that's the case.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -304,11 +304,11 @@ two fields, also called a "tagged union" in C:
 - a `repr(C)` union of `repr(C)` structs for the fields of each variant that had
   them ("the payload")
 
-> Note: due to the representation of `repr(C)` structs and unions, if a variant
+> Note: Due to the representation of `repr(C)` structs and unions, if a variant
 > has a single field there is no difference between putting that field directly
 > in the union or wrapping it in a struct; any system which wishes to manipulate
 > such an `enum`'s representation may therefore use whichever form is more
-> convenient/consistent for them
+> convenient or consistent for them
 
 ```rust
 // This Enum has the same layout as ...
@@ -334,10 +334,10 @@ enum MyEnumDiscriminant { A, B, C, D }
 // This is the variant union.
 #[repr(C)]
 union MyEnumFields {
-   A: MyAFields,
-   B: MyBFields,
-   C: MyCFields,
-   D: MyDFields,
+    A: MyAFields,
+    B: MyBFields,
+    C: MyCFields,
+    D: MyDFields,
 }
 
 #[repr(C)]
@@ -372,13 +372,6 @@ different behavior whether the enum has fields or no fields. It is an error
 for [zero-variant enumerations] to have a primitive representation. Combining
 two primitive representations together is an error.
 
-<span id="c-primitive-representation">Combining the `repr(C)`
-and a primitive representation is only defined for enums with fields. The
-primitive representation modifies the `repr(C)` by changing the representation
-of the discriminant enum to have the representation of the chosen primitive
-representation. So, if you chose the `u8` representation, then the discriminant
-enum would have a size and alignment of 1 byte.</span>
-
 > Note: Primitive representations are primarily intended for Rust code that
 > wants to interoperate with the idioms of preexisting C/C++ codebases. If you
 > have control over both the Rust and C code, such as using C as FFI glue
@@ -388,21 +381,21 @@ enum would have a size and alignment of 1 byte.</span>
 
 #### Primitive Representation of Field-less Enums
 
-For [field-less enums], they set the size and alignment to be the same as
-the primitive type of the same name. For example, a field-less enum with
-a `u8` representation can only have discriminants between 0 and 255 inclusive.
+For [field-less enums], primitive representations set the size and alignment to
+be the same as the primitive type of the same name. For example, a field-less
+enum with a `u8` representation can only have discriminants between 0 and 255
+inclusive.
 
 #### Primitive Representation of Enums With Fields
 
-The representation of a `repr(int)` enum is a `repr(C)` union of `repr(C)`
-structs for each variant with a field. The first field of each struct in the
-union is a `repr(int)` version of the enum with all fields removed ("the tag")
-and the remaining fields are the fields of that variant.
+The representation of a primitive representation enum is a `repr(C)` union of
+`repr(C)` structs for each variant with a field. The first field of each struct
+in the union is the primitive representation version of the enum with all fields
+removed ("the tag") and the remaining fields are the fields of that variant.
 
-> Note: this representation is unchanged if the tag is given its own member in
-> the union, should that make manipulation more clear for you (although in C++,
-> to follow The Exact Word Of The Standard the tag member should be wrapped in
-> a `struct`).
+> Note: This representation is unchanged if the tag is given its own member in
+> the union, should that make manipulation more clear for you (although to
+> follow the C++ standard the tag member should be wrapped in a `struct`).
 
 ```rust
 // This enum has the same layout as ...
@@ -417,10 +410,10 @@ enum MyEnum {
 // ... this union.
 #[repr(C)]
 union MyEnumRepr {
-   A: MyVariantA,
-   B: MyVariantB,
-   C: MyVariantC,
-   D: MyVariantD,
+    A: MyVariantA,
+    B: MyVariantB,
+    C: MyVariantC,
+    D: MyVariantD,
 }
 
 // This is the discriminant enum.
@@ -446,6 +439,12 @@ struct MyVariantD(MyEnumDiscriminant);
 ```
 
 > Note: `union`s with non-`Copy` fields are unstable, see [55149].
+
+Combining the `repr(C)` and a primitive representation is only defined for enums
+with fields. The primitive representation modifies the `repr(C)` by changing the
+representation of the discriminant enum to have the representation of the chosen
+primitive representation. So, if you chose the `u8` representation, then the
+discriminant enum would have a size and alignment of 1 byte.
 
 ### The alignment modifiers
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -308,7 +308,7 @@ two fields, also called a "tagged union" in C:
 > has a single field there is no difference between putting that field directly
 > in the union or wrapping it in a struct; any system which wishes to manipulate
 > such an `enum`'s representation may therefore use whichever form is more
-> convenient or consistent for them
+> convenient or consistent for them.
 
 ```rust
 // This Enum has the same layout as ...
@@ -493,7 +493,7 @@ enum Enum16 {
     Variant1,
 }
 
-// One byte for the discriminant and one byte for the value in Enum8::Variant0
+// The C representation, with the size of the discriminant unspecified
 assert_eq!(std::mem::size_of::<EnumC>(), 8);
 // One byte for the discriminant and one byte for the value in Enum8::Variant0
 assert_eq!(std::mem::size_of::<Enum8>(), 2);
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -373,11 +373,7 @@ for [zero-variant enumerations] to have a primitive representation. Combining
 two primitive representations together is an error.
 
 > Note: Primitive representations are primarily intended for Rust code that
-> wants to interoperate with the idioms of preexisting C/C++ codebases. If you
-> have control over both the Rust and C code, such as using C as FFI glue
-> between Rust and some third language, then you should use a
-> [primitive representation](#primitive-representation-of-enums-with-fields)
-> instead.
+> wants to interoperate with the idioms of preexisting C/C++ codebases.
 
 #### Primitive Representation of Field-less Enums
 
@@ -493,7 +489,7 @@ enum Enum16 {
     Variant1,
 }
 
-// The C representation, with the size of the discriminant unspecified
+// The size of the C representation is platform dependant
 assert_eq!(std::mem::size_of::<EnumC>(), 8);
 // One byte for the discriminant and one byte for the value in Enum8::Variant0
 assert_eq!(std::mem::size_of::<Enum8>(), 2);
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -361,20 +361,6 @@ struct MyDFields;
 
 > Note: `union`s with non-`Copy` fields are unstable, see [55149].
 
-<span id="c-primitive-representation">Combining the `repr(C)`
-and a primitive representation is only defined for enums with fields. The
-primitive representation modifies the `repr(C)` by changing the representation
-of the discriminant enum to have the representation of the chosen primitive
-representation. So, if you chose the `u8` representation, then the discriminant
-enum would have a size and alignment of 1 byte.</span>
-
-> Note: This representation is primarily intended for Rust code that wants to
-> interoperate with the idioms of preexisting C/C++ codebases. If you have
-> control over both the Rust and C code, such as using C as FFI glue between
-> Rust and some third language, then you should use a
-> [primitive representation](#primitive-representation-of-enums-with-fields)
-> instead.
-
 ### Primitive representations
 
 The *primitive representations* are the representations with the same names as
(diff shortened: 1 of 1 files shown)


src/type-layout.md
```diff
@@ -383,12 +369,22 @@ the primitive integer types. That is: `u8`, `u16`, `u32`, `u64`, `u128`,
 
 Primitive representations can only be applied to enumerations and have
 different behavior whether the enum has fields or no fields. It is an error
-for [zero-variant enumerations] to have a primitive representation.
+for [zero-variant enumerations] to have a primitive representation. Combining
+two primitive representations together is an error.
 
-Combining two primitive representations together is unspecified.
+<span id="c-primitive-representation">Combining the `repr(C)`
+and a primitive representation is only defined for enums with fields. The
```
RalfJung, line 376: This sentence makes it sound like the discussion should be in
```
#### Primitive Representation of Enums With Fields
```
That subsection really has two parts: with and without `repr(C)`.

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+The representation of a `repr(C)` enum with fields is a `repr(C)` struct with
+two fields, also called a "tagged union" in C:
 
-Likewise, combining the `C` representation with a primitive representation, the
-layout is unspecified.
+- a `repr(C)` version of the enum with all fields removed ("the tag")
+- a `repr(C)` union of `repr(C)` structs for the fields of each variant that had
+  them ("the payload")
+
+> Note: due to the representation of `repr(C)` structs and unions, if a variant
```
ehuss, line 307:
```suggestion
> Note: Due to the representation of `repr(C)` structs and unions, if a variant
```

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+> has a single field there is no difference between putting that field directly
+> in the union or wrapping it in a struct; any system which wishes to manipulate
+> such an `enum`'s representation may therefore use whichever form is more
+> convenient/consistent for them
```
ehuss, line 311:
```suggestion
> convenient or consistent for them.
```

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
 the primitive integer types. That is: `u8`, `u16`, `u32`, `u64`, `u128`,
 `usize`, `i8`, `i16`, `i32`, `i64`, `i128`, and `isize`.
 
-Primitive representations can only be applied to enumerations.
+Primitive representations can only be applied to enumerations and have
+different behavior whether the enum has fields or no fields. It is an error
+for [zero-variant enumerations] to have a primitive representation. Combining
+two primitive representations together is an error.
+
+<span id="c-primitive-representation">Combining the `repr(C)`
```
ehuss, line 375: I think this span can be removed if the link has been removed.

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+> Note: Primitive representations are primarily intended for Rust code that
+> wants to interoperate with the idioms of preexisting C/C++ codebases. If you
+> have control over both the Rust and C code, such as using C as FFI glue
+> between Rust and some third language, then you should use a
+> [primitive representation](#primitive-representation-of-enums-with-fields)
+> instead.
+
+#### Primitive Representation of Field-less Enums
+
+For [field-less enums], they set the size and alignment to be the same as
```
ehuss, line 391:
```suggestion
For [field-less enums], primitive representations set the size and alignment to be the same as
```

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+a `u8` representation can only have discriminants between 0 and 255 inclusive.
 
-For [C-like enumerations], they set the size and alignment to be the same as the
-primitive type of the same name. For example, a C-like enumeration with a `u8`
-representation can only have discriminants between 0 and 255 inclusive.
+#### Primitive Representation of Enums With Fields
 
-It is an error for [zero-variant enumerations] to have a primitive
-representation.
+The representation of a `repr(int)` enum is a `repr(C)` union of `repr(C)`
```
ehuss, line 397:
```suggestion
The representation of a primitive representation enum is a `repr(C)` union of `repr(C)`
```

`repr(int)` isn't ever defined or used elsewhere. I realize the RFC used it as a shorthand for "primitive representation", and "primitive representation" is wordy, but I think it can be confusing when seen in isolation.

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+structs for each variant with a field. The first field of each struct in the
+union is a `repr(int)` version of the enum with all fields removed ("the tag")
```
ehuss, line 399:
```suggestion
union is the primitive representation version of the enum with all fields removed ("the tag")
```

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+and the remaining fields are the fields of that variant.
+
+> Note: this representation is unchanged if the tag is given its own member in
```
ehuss, line 402:
```suggestion
> Note: This representation is unchanged if the tag is given its own member in
```

src/type-layout.md
```diff
@@ -285,40 +286,166 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+> the union, should that make manipulation more clear for you (although in C++,
+> to follow The Exact Word Of The Standard the tag member should be wrapped in
```
ehuss, line 404:
```suggestion
> the union, should that make manipulation more clear for you (although to
> follow the C++ standard the tag member should be wrapped in
```

src/type-layout.md
poliorcetics: Good catch, I had completely forgotten it

src/type-layout.md
poliorcetics: I moved the paragraph to the end of the `Primitive Representation of Enums With Fields`, is it better ?

src/type-layout.md
```diff
@@ -285,40 +286,165 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+#[derive(Clone, Copy)]
+struct MyVariantD(MyEnumDiscriminant);
+```
 
-For all other enumerations, the layout is unspecified.
+> Note: `union`s with non-`Copy` fields are unstable, see [55149].
 
-Likewise, combining two primitive representations together is unspecified.
+Combining the `repr(C)` and a primitive representation is only defined for enums
+with fields. The primitive representation modifies the `repr(C)` by changing the
```
RalfJung, line 444: Yes, much better. :) Some wording suggestions:
```suggestion
For enums with fields, it is also possible to combining `repr(C)` and a primitive representation.
This modifies the `repr(C)` by changing the
```
Also it would be nice to make "the `repr(C)`" in the 2nd line a link to the relevant section.

And finally, is it possible to summarize the difference between this and "normal" primitive reps in a sentence? Something like "union-of-structs vs struct-of-tag-and-unions"?

poliorcetics:
> And finally, is it possible to summarize the difference between this and "normal" primitive reps in a sentence? Something like "union-of-structs vs struct-of-tag-and-unions"?

I'm not sure I understand what it is you want

src/type-layout.md
```diff
@@ -440,12 +440,14 @@ struct MyVariantD(MyEnumDiscriminant);
 
 > Note: `union`s with non-`Copy` fields are unstable, see [55149].
 
-Combining the `repr(C)` and a primitive representation is only defined for enums
-with fields. The primitive representation modifies the `repr(C)` by changing the
+For enums with fields, it is also possible to combining `repr(C)` and a
```
RalfJung, line 443:
```suggestion
For enums with fields, it is also possible to combine `repr(C)` and a
```

RalfJung:
> > And finally, is it possible to summarize the difference between this and "normal" primitive reps in a sentence? Something like "union-of-structs vs struct-of-tag-and-unions"?
> 
> I'm not sure I understand what it is you want

I think it would be useful to directly contrast `repr(C)` and `repr(C, int)` and explain how they differ.

src/type-layout.md
```diff
@@ -440,12 +440,14 @@ struct MyVariantD(MyEnumDiscriminant);
...
+primitive representation. This modifies the [`repr(C)`] by changing the
```
RalfJung, line 444:
```suggestion
primitive representation (e.g., `repr(C, i32)`). This modifies the [`repr(C)`] by changing the
```

src/type-layout.md
```diff
@@ -440,12 +440,14 @@ struct MyVariantD(MyEnumDiscriminant);
...
 representation of the discriminant enum to have the representation of the chosen
 primitive representation. So, if you chose the `u8` representation, then the
```
```suggestion
representation of the discriminant enum to have the representation of the chosen
primitive. So, if you chose the `u8` representation, then the
```
At least I think that's what you mean? "the representation of the chose primitive representation" does make sense, I think.

poliorcetics:
> I think it would be useful to directly contrast `repr(int)` and `repr(C, int)` and explain how they differ.

I added a new subsection that explains it and shows some example. 😄

src/type-layout.md
RalfJung: This strange wording is still in the latest version I think.

src/type-layout.md
```diff
@@ -440,12 +440,56 @@ struct MyVariantD(MyEnumDiscriminant);
...
 
-For enums with fields, it is also possible to combining `repr(C)` and a
-primitive representation. This modifies the [`repr(C)`] by changing the
-representation of the discriminant enum to have the representation of the chosen
-primitive representation. So, if you chose the `u8` representation, then the
+#### Combining primitive representations of enums with fields and \#[repr(C)]
+
+For enums with fields, it is also possible to combine `repr(C)` and a
+primitive representation (e.g., `repr(C, u8)`). This modifies the [`repr(C)`] by
+changing the representation of the discriminant enum to have the representation
```
RalfJung, line 447: Here: "changing the representation [...] to have the representation of".  But representations do not have representations.

poliorcetics: I updated the wording and added a little context to the example below, is it better this way ?

RalfJung: Yes looks good. :)

src/type-layout.md
```diff
@@ -285,40 +286,223 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+different behavior whether the enum has fields or no fields. It is an error
+for [zero-variant enumerations] to have a primitive representation. Combining
+two primitive representations together is an error.
+
+> Note: Primitive representations are primarily intended for Rust code that
+> wants to interoperate with the idioms of preexisting C/C++ codebases. If you
+> have control over both the Rust and C code, such as using C as FFI glue
+> between Rust and some third language, then you should use a
+> [primitive representation](#primitive-representation-of-enums-with-fields)
+> instead.
```
ehuss, line 380: I'm a little unclear here.  "instead" of what?  `repr(C)`?

src/type-layout.md
```diff
@@ -285,40 +286,223 @@ the default `enum` size and alignment for the target platform's C ABI.
...
-layout is unspecified.
+- a `repr(C)` version of the enum with all fields removed ("the tag")
+- a `repr(C)` union of `repr(C)` structs for the fields of each variant that had
+  them ("the payload")
+
+> Note: Due to the representation of `repr(C)` structs and unions, if a variant
+> has a single field there is no difference between putting that field directly
+> in the union or wrapping it in a struct; any system which wishes to manipulate
+> such an `enum`'s representation may therefore use whichever form is more
+> convenient or consistent for them
```
ehuss, line 311:
```suggestion
> convenient or consistent for them.
```

src/type-layout.md
```diff
@@ -285,40 +286,223 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+    Variant1,
+}
+
+#[repr(C, u16)]
+enum Enum16 {
+    Variant0(u8),
+    Variant1,
+}
+
+// One byte for the discriminant and one byte for the value in Enum8::Variant0
```
ehuss, line 496: This doesn't explain why two bytes would be equal to 8.

Also, it says `Enum8::Variant0`, but I would think it should be `EnumC::Variant0`?

poliorcetics: That's a copy-paste mistake 😓

src/type-layout.md
poliorcetics: I'm not sure this whole note is useful. I could shorten it to:

> Note: Primitive representations are primarily intended for Rust code that wants to interoperate with the idioms of preexisting C/C++ codebases.

src/type-layout.md
```diff
@@ -285,40 +286,223 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+    Variant1,
+}
+
+#[repr(C, u16)]
+enum Enum16 {
+    Variant0(u8),
+    Variant1,
+}
+
+// The C representation, with the size of the discriminant unspecified
```
ehuss, line 496: I thought the discriminant was specified — as the size of a C `enum` of the target platform's C ABI?  Then there is 1 byte for the u8 value, plus padding.  It might also be worth indicating what platform this example is for, since the value could be different for other platforms?  Or maybe just scratch all that and say "the size is platform-dependent"?

poliorcetics: I'll go with "the size is platform dependant"

src/type-layout.md
poliorcetics: I used the above wording in the "Fix nits 2" commit

ehuss: Reading it again, I'm a little unclear, since there are multiple forms.  In this case, is it referring to `repr(C, Int)` or `repr(Int)`?  AIUI:

* `repr(C, Int)` is useful for C/C++ code that uses a struct with an explicitly-sized tag and payload (which is probably more common)
* `repr(Int)` is useful for C/C++ code that uses a union with an explicitly-sized tag in each struct, and can be more space-efficient than `repr(C, Int), but less common for pre-existing C code?
* `repr(C)` is useful for C/C++ code that uses a struct with a **C-enum tag** and payload? And that tag size is implementation-defined, and "best guess", so probably harder to get correct? That is, you really need to know which C compiler, platform, and C flags are being used?

Is that roughly correct?

poliorcetics: I would say you are correct. I don't have much experience with this but the hobby projects I did used `repr(C, int)` to ensure maximal compatibility, which coincides with what you said. Would you like me to properly write in the document ?

ehuss: Hm.  How about removing the note altogether for now?  This feels like it would be more useful in a guide on using FFI, and I don't want to dive too deep on it here. The reference has generally shied away from including information on how something is supposed to be used, as that tends more toward guide material.

ehuss:
@RalfJung Did you have any further comments on this, or does it look good to you?

poliorcetics:
Remind me to squash once all the changes are approved, I'm just waiting for all the reviews to be done. 😄

src/type-layout.md
poliorcetics: I removed it.

src/type-layout.md
```diff
@@ -285,40 +286,216 @@ the default `enum` size and alignment for the target platform's C ABI.
...
+  them ("the payload")
+
+> Note: Due to the representation of `repr(C)` structs and unions, if a variant
+> has a single field there is no difference between putting that field directly
+> in the union or wrapping it in a struct; any system which wishes to manipulate
+> such an `enum`'s representation may therefore use whichever form is more
+> convenient or consistent for them.
+
+```rust
+// This Enum has the same layout as ...
```
RalfJung, line 314: This seems to use the term "layout" for what is called "representation" above.

RalfJung (commented):
My only remaining comment is that this seems to use "layout" and "representation" synonymously. But this is likely pre-existing and it is something we also struggle with in the UCG. (We mostly switched to "layout" there but then realized "layout" can mean at least 3 different things... see https://github.com/rust-lang/unsafe-code-guidelines/issues/122.)

I did not fact-check all the statements about representations and C++ here, as I am not deeply familiar with those details.

src/type-layout.md
poliorcetics: No you're right, `representation` is used every where in those paragraphs, I'll change it.

ehuss (approved):
Thanks!  I appreciate sticking through with all the review comments.  I'll just squash via GitHub.

PR #880, opened 2020-08-27
Added variable back into example.

ciaranmooney:
This PR adds the assignment of x back into the match example. Without it, the example doesn't seem to make sense.


Havvy:
💟 Thanks! Yeah, it's weird that this one had the `x` omitted, but the next one shows the struct.

PR #883, opened 2020-09-04
Rust 1.46 now allows more features in const fn

jrincayc:
These have been added in https://github.com/rust-lang/rust/pull/72437/ so should be added to the reference.


ehuss:
Thanks, I've actually been meaning to delete this entire section. Would you mind doing that, and linking to the const-eval chapter (https://github.com/rust-lang/reference/blob/master/src/const_eval.md#const-functions)?  It might also be good to double check that the const-eval chapter is not missing anything here (it has been recently updated).

jrincayc:
> Thanks, I've actually been meaning to delete this entire section. Would you mind doing that, and linking to the const-eval chapter (https://github.com/rust-lang/reference/blob/master/src/const_eval.md#const-functions)? It might also be good to double check that the const-eval chapter is not missing anything here (it has been recently updated).

I deleted most of the section in the newest merge version and moved things that I thought useful to the const_eval.md.  I left a link to the const-functions section.

ehuss:
Thanks!  Just one minor thing, this leaves some dangling link references at the bottom of `src/items/functions.md`, can you remove them?  I think they are:

* `[trait bounds]`
* ``[`?Sized`]``
* `[external_block_abi]` (unrelated to this PR, but can also be removed)

jrincayc:
> Thanks! Just one minor thing, this leaves some dangling link references at the bottom of `src/items/functions.md`, can you remove them? I think they are:
> 
>     * `[trait bounds]`
> 
>     * `` [`?Sized`] ``
> 
>     * `[external_block_abi]` (unrelated to this PR, but can also be removed)

I removed those references.
P.S. is there some trick to finding them?

ehuss (approved):
Thanks!

ehuss:
>  P.S. is there some trick to finding them?

Not really, just check when removing a link if there are any other references. Unfortunately the markdown parser eats the references and doesn't provide a way to access them.

PR #885, opened 2020-09-15
Fix stable check.

ehuss:
The stable check was stopping at the first directory it found.

### DIFF
--- a/stable-check/src/main.rs
+++ b/stable-check/src/main.rs
@@ -1,5 +1,5 @@
-use std::error::Error;
 use std::env;
+use std::error::Error;
 use std::fs;
 use std::fs::File;
 use std::io::prelude::*;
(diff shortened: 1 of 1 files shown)


stable-check/src/main.rs
```diff
@@ -26,15 +26,15 @@ fn check_directory(dir: &Path) -> Result<(), Box<dyn Error>> {
...
-        }
-
-        let mut file = File::open(&path)?;
-        let mut contents = String::new();
-        file.read_to_string(&mut contents)?;
+            check_directory(&path)?;
+        } else {
+            let mut file = File::open(&path)?;
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)?;
```
lzutao, line 33: You could use `fs::read_to_string`.

ehuss: Yea, I noticed that, too. I rewrote it in #886.

Havvy:
Have you tested that your change works locally?

ehuss:
Yea, I've tested that this works as expected. However, #886 is a superset of this, and I would prefer to merge that instead. Do you think you can review that PR?

Havvy:
I presume this should be closed due to #886. If not, reopen.

PR #886, opened 2020-09-15
Add style checks.

ehuss:
This extends the stable-checker to do more style checks:

- `#![feature]` nightly features.
- CR linefeeds.
- Tab characters.
- Trailing whitespace.
- Missing newline at end of file.
- Non-md or html files.
- Broken link references.
- Indented code blocks (should use fenced code blocks).
- Code blocks without a language.

And includes the associated fixes.

There is some risk this can add frustration to contributing, but I think it is worthwhile. The `rust-lang/rust` repository already has most of these checks, and I don't think it has been too bad there.


Havvy (approved):
I haven't done a proper review for this, but the code only affects management of this repo, none of the code could possibly affect anything else, and I trust @ehuss to have thought about this enough, I'm going to approve.

PR #887, opened 2020-09-25
Update guarantees of unwinding past an extern "C" function

yuriks:
As prompted by discussion in rust-lang/rust#52652

The current docs say it will abort, however that behavior was actually reverted in rust-lang/rust#48445 and is not what currently happens in stable.

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -174,12 +174,15 @@ extern "C" fn new_i32() -> i32 { 0 }
 let fptr: extern "C" fn() -> i32 = new_i32;
 ```
 
-Functions with an ABI that differs from `"Rust"` do not support unwinding in the
-exact same way that Rust does. Therefore, unwinding past the end of functions
-with such ABIs causes the process to abort.
+Functions with an ABI that differs from `"Rust"` do not support unwinding in
+the exact same way that Rust does. Therefore, unwinding past the end of
+functions with such ABIs results in undefined behavior, and will differ between
+platforms and compiler editions. If you cannot guarantee this will not happen,
+you should use [`catch_unwind`] or similar to abort the program instead if
+unwinding would pass through the function.
 
-> **Note**: The LLVM backend of the `rustc` implementation
-aborts the process by executing an illegal instruction.
+> **Note**: Currently, in nightly `rustc` the LLVM backend aborts the process
+> by executing an illegal instruction. However, this is not the behavior in stable.
 
 ## Const functions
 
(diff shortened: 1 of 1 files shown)


src/items/functions.md
```diff
@@ -174,12 +174,15 @@ extern "C" fn new_i32() -> i32 { 0 }
...
+the exact same way that Rust does. Therefore, unwinding past the end of
+functions with such ABIs results in undefined behavior, and will differ between
+platforms and compiler editions. If you cannot guarantee this will not happen,
+you should use [`catch_unwind`] or similar to abort the program instead if
+unwinding would pass through the function.
 
-> **Note**: The LLVM backend of the `rustc` implementation
-aborts the process by executing an illegal instruction.
+> **Note**: Currently, in nightly `rustc` the LLVM backend aborts the process
+> by executing an illegal instruction. However, this is not the behavior in stable.
```
ehuss, line 185: We generally don't include notes about specific versions of the rustc compiler in the reference.  I'm also not sure what this is referring to, since it no longer aborts on unwind.

src/items/functions.md
```diff
@@ -174,12 +174,15 @@ extern "C" fn new_i32() -> i32 { 0 }
 let fptr: extern "C" fn() -> i32 = new_i32;
 ```
 
-Functions with an ABI that differs from `"Rust"` do not support unwinding in the
-exact same way that Rust does. Therefore, unwinding past the end of functions
-with such ABIs causes the process to abort.
+Functions with an ABI that differs from `"Rust"` do not support unwinding in
+the exact same way that Rust does. Therefore, unwinding past the end of
+functions with such ABIs results in undefined behavior, and will differ between
```
ehuss, line 179: Can you link `undefined behavior` to `behavior-considered-undefined.md`?

src/items/functions.md
```diff
@@ -174,12 +174,15 @@ extern "C" fn new_i32() -> i32 { 0 }
...
+platforms and compiler editions. If you cannot guarantee this will not happen,
```
ehuss, line 180: I'm not sure what "compiler editions" means here.  "Edition" has a specific meaning in Rust, and I don't think that is what is being referred to here.  I think it would be fine to end the sentence at "undefined behavior", with the link leading the reader to understanding what that means.

ehuss (commented):
Thanks for the PR!

ehuss:
I'm going to close as all the panic/unwinding stuff has been rewritten in https://github.com/rust-lang/reference/pull/1226. Thanks!

PR #888, opened 2020-10-04
Document execution order

### LINKED ISSUE #248: Document expression evaluation order
See also, https://internals.rust-lang.org/t/rust-expression-order-of-evaluation/2605/32

Havvy:
Thank eddyb and Steve Klabnik for complaining on Twitter for this PR.

Closes #248

The order is now documented. I considered using an example over expression evaluation using println!, but the twitter thread is explicitly using iterators and this is probably the most common reason to care about evaluation order in tuple expressions anyways.

I also cleaned up the page to bring it into style compliance. I did both at the same time, so there's only one noisy commit. Sorry to the reviewer.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -85,6 +85,30 @@ in the order given by their associativity.
 | `=` `+=` `-=` `*=` `/=` `%=` <br> `&=` <code>&#124;=</code> `^=` `<<=` `>>=` | right to left |
 | `return` `break` closures   |                     |
 
+## Evaluation order
+
+Most expressions include subexpressions. Unless otherwise stated on the
+expression's page, evaluation of these inner expressions is left to right as
+written in the source code.
+
+For example, the two `next` method calls will always be called in the same
+order:
+
+```rust
+# // Using vec instead of array to avoid references
+# // since there is no stable owned array iterator
+# // at the time this example was written.
+let mut one_two = vec![1, 2].into_iter();
+assert_eq!(
+    (1, 2),
+    (one_two.next().unwrap(), one_two.next().unwrap())
+);
+```
+
+> **Note**: Since this is applied recursively, expressions are also evaluated
+> from innermost to outermost, ignoring siblings until there are no inner
+> subexpressions.
+
 ## Place Expressions and Value Expressions
 
 Expressions are divided into two main categories: place expressions and
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -90,7 +90,12 @@ The following list of expressions all evaluate their operands the same way, as
 described after the list. Other expressions either don't take operands or
 evaluate them conditionally as described on their respective pages.
 
-* Operator expressions (except the Lazy boolean operators)
+* Dereference expression
+* Error propagation expression
+* Negation expression
+* Arithmetic and logical binary operators
+* Comparison operators
+* Type cast expression
 * Grouped expression
 * Array expression
 * Await expression
@@ -127,9 +132,9 @@ assert_eq!(
 );
 ```
 
-> **Note**: Since this is applied recursively, expressions are also evaluated
-> from innermost to outermost, ignoring siblings until there are no inner
-> subexpressions.
+> **Note**: Since this is applied recursively, these expressions are also
+> evaluated from innermost to outermost, ignoring siblings until there are no
+> inner subexpressions.
 
 ## Place Expressions and Value Expressions
 
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -48,10 +48,9 @@ value, and has effects during *evaluation*. Many expressions contain
 sub-expressions, called the *operands* of the expression. The meaning of each
 kind of expression dictates several things:
 
-* Whether or not to evaluate the sub-expressions when evaluating the expression
-* The order in which to evaluate the sub-expressions
-* How to combine the sub-expressions' values to obtain the value of the
-  expression
+* Whether or not to evaluate the operands when evaluating the expression
+* The order in which to evaluate the operands
+* How to combine the operands' values to obtain the value of the expression
 
 In this way, the structure of expressions dictates the structure of execution.
 Blocks are just another kind of expression, so blocks, statements, expressions,
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -86,9 +86,29 @@ in the order given by their associativity.
 
 ## Evaluation order of operands
 
-Unless otherwise stated on the expression's page, evaluation of the operands of
-an expression is done before applying the effect of the expression. The operands
-are executed left to right as written in the source code.
+The following list of expressions all evaluate their operands the same way, as
+described after the list. Other expressions either don't take operands or
+evaluate them conditionally as described on their respective pages.
+
+* Operator expressions (except the Lazy boolean operators)
+* Grouped expression
+* Array expression
+* Await expression
+* Index expression
+* Tuple expression
+* Tuple index expression
+* Struct expression
+* Enumeration variant expression
+* Call expression
+* Method call expression
+* Field expression
+* Break expression
+* Range expression
+* Return expression
+
+The operands of these expressions are evaluated prior to applying the effects of
+the expression. Expressions taking multiple operands are evaluated left to right
+as written in the source code.
 
 > **Note**: Which subexpressions are the operands of an expression is
 > determined by expression precedence as per the previous section.
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -45,8 +45,8 @@
 An expression may have two roles: it always produces a *value*, and it may have
 *effects* (otherwise known as "side effects"). An expression *evaluates to* a
 value, and has effects during *evaluation*. Many expressions contain
-sub-expressions (operands). The meaning of each kind of expression dictates
-several things:
+sub-expressions, called the *operands* of the expression. The meaning of each
+kind of expression dictates several things:
 
 * Whether or not to evaluate the sub-expressions when evaluating the expression
 * The order in which to evaluate the sub-expressions
@@ -85,11 +85,14 @@ in the order given by their associativity.
 | `=` `+=` `-=` `*=` `/=` `%=` <br> `&=` <code>&#124;=</code> `^=` `<<=` `>>=` | right to left |
 | `return` `break` closures   |                     |
 
-## Evaluation order
+## Evaluation order of operands
 
-Most expressions include subexpressions. Unless otherwise stated on the
-expression's page, evaluation of these inner expressions is left to right as
-written in the source code.
+Unless otherwise stated on the expression's page, evaluation of the operands of
+an expression is done before applying the effect of the expression. The operands
+are executed left to right as written in the source code.
+
+> **Note**: Which subexpressions are the operands of an expression is
+> determined by expression precedence as per the previous section.
 
 For example, the two `next` method calls will always be called in the same
 order:
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -427,9 +427,9 @@ operand.
 > is evaluated before the left one.
 
 It then has the effect of first [dropping] the value at the assigned place,
-unless the place is an uninitialized local variable or field of a local
-variable. Next it either [copies or moves] the assigned value to the assigned
-place.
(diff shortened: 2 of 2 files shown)


eddyb:
Why document any one thing in particular? The order is universally "postorder" (on the AST), i.e. "more nested" expressions first (effectively mandated by "expressions depend on the result of their subexpressions"), and between siblings it's specifically left-to-right (outside of perhaps assignments' LHS? I forget how we resolved that one).

Can't we specify it in one place? Doing it on a case-by-case basis seems inefficient and would suggest the order isn't uniformly *defined*, whereas *it is* (otherwise e.g. the borrow-checker couldn't possibly work).

cc @rust-lang/lang

Havvy:
I thought about that after I went to bed as well. I could add an evaluation order section to `expressions.md` instead.

Havvy:
I've completely rewritten this PR to now document the default at the top level. I'll be doing the tuple changes as their own PR.

ehuss:
I don't think I know enough to review this. Is this an actual commitment the language can make? Does `rustc` actually enforce this? Doesn't LLVM do some amount of reordering? Can an implementation reorder expressions if it can prove it results in the same value and observable side-effects?

I would expect this to have some more precise language. Perhaps make it clear that this left-to-right ordering is referring to the *operands* of an expression, and this is not related operator precedence and associativity? (Or, at least, spell out the relationship with operator precedence and associativity?) I think "operands" is a little more specific than "subexpressions" or "inner expressions".

I'm also a little unclear on how this relates to *execution order*, which is sometimes defined relative to "sequence" or "critical execution" points.

Havvy:
For reordering expressions: If there are no observable side effects done by doing so, then expressions can be reordered in their evaluation order. Heck, they can even be evaluated at compile time under certain conditions. But this is also true about the ordering of statements in a block expression. And yet, those aren't semantics of the language. They're optimizations in the implementation. Anywhere there are actual side effects, this becomes true.

With respect to "operands", I think "subexpressions" makes a more clear term. Most expressions contain inner expressions that are subsumed by that expression. Operand would have to be defined and outside of arithemetic operator expressions, you don't see the term used.

With respect to "operator precedence", I should add a note about that.

Evaluation order and execution order are basically synonymous to me. I don't think I've seen a distinction between the two terms in any literature I've read.

ehuss:
The reason I suggest "operand" is because for me, "sub-expression" can be ambiguous if it is recursive. Like, the "borrow expression" clearly (to me) has one operand, but has a limitless number of sub-expressions if you consider it recursively. Also, "operand" is already defined and used throughout `expressions.md` and `operator-expr.md`, and it seems a little more direct to me.  Just my opinion, though, I don't think it matters too much.

I guess I just feel a little uneasy to have an hard rule of "left to right", when that's not universally true if optimizers are allowed to reorder based on some unwritten rules.  Like [C++17](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0145r3.pdf) very clearly defines ordering with respect to value computation and side effects. And [Java](https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.7) very clearly expresses the order and caveats on when it can be rewritten.  Many other specs just vaguely mention "left-to-right" (like [Go](https://golang.org/ref/spec#Order_of_evaluation) and Javascript and [D](https://dlang.org/spec/expression.html#order-of-evaluation) and [C#](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#run-time-evaluation-of-argument-lists)), so it is not unusual, but it seems strange to me to be so imprecise in a spec.

I think it is interesting that C# [used to have](https://web.archive.org/web/20120312085757/http://msdn.microsoft.com/en-us/library/Aa691105) an ordering based on well-defined side effects, but that seems to have been removed and it only mentions left-to-right ordering now.

Havvy:
See the commit message.

Since operand is used already, I'll continue to use it. Didn't actually see it before.

What optimizer do is something that we don't concern ourselves with in the reference. It can do much worse things to the semantics of the language already by eliding copies and movies and function calls or replacing function calls with a constant instead. Yes, it can reorder if it doesn't change the ultimate effect. Nothing in the reference in universal in the face of an optimizer.

Havvy:
@ehuss Do you still feel unqualified to review this?

ehuss:
I would be more comfortable if someone from the language team reviewed this.  It looks good to me, and it does follow the convention used in several other languages as noted above.  I'm just uncertain if that's the most appropriate way to define the order.  Otherwise I'm fine with it.

RalfJung:
> I guess I just feel a little uneasy to have an hard rule of "left to right", when that's not universally true if optimizers are allowed to reorder based on some unwritten rules.

It *is* universally true though. Optimizations are only allowed to reorder things in a way that causes no change of observable behavior (the "as-if" rule). The rules for what optimizations are allowed to do are not unwritten, they are "you must only do things where the resulting program only behaves in ways that the original program could already behave as" -- this is the universal correctness condition for all compilers of all programming languages.

That's just like how the spec says that `x*2` results in the value of `x` being multiplied by 2, but optimizations can still change this to something else, like `x << 1`, or remove the computation entirely if the result is unused (and no overflow occurred). And yet the spec will just talk about multiplication and not about how the compiler could choose to implement that multiplication.

RalfJung:
> Nothing in the reference is universal in the face of an optimizer.

That would be horrible! Everything in the reference is universal for all *correct* optimizers, when considering the *observable effects* of running a Rust program. Don't think in terms of assembly, think of the resulting program as a black box that you can only probe by running it, feeding it input and observing its output. Everything that happens that way must conform to what the reference says.

The reference says nothing about the inside of that black box, but that does not at all mean that the reference is any less universal. The inside of that black box is simply out of scope, just like, e.g., the exact formatting of error messages.

src/expressions.md
```diff
@@ -45,13 +45,12 @@
 An expression may have two roles: it always produces a *value*, and it may have
 *effects* (otherwise known as "side effects"). An expression *evaluates to* a
 value, and has effects during *evaluation*. Many expressions contain
-sub-expressions (operands). The meaning of each kind of expression dictates
```
joshtriplett, line 48: This is a nit, and not a semantic issue with the description, but: I have a moderate preference that we continue to use "sub-expressions" in general, and only use "operands" for "sub-expressions of an operator".

src/expressions.md
```diff
@@ -85,6 +84,33 @@ in the order given by their associativity.
 | `=` `+=` `-=` `*=` `/=` `%=` <br> `&=` <code>&#124;=</code> `^=` `<<=` `>>=` | right to left |
 | `return` `break` closures   |                     |
 
+## Evaluation order of operands
+
+Unless otherwise stated on the expression's page, evaluation of the operands of
```
joshtriplett, line 89: Can we cross-reference any cases where this *isn't* true from within this section, to not leave people wondering "what expressions *don't* evaluate their sub-expressions left-to-right"?

joshtriplett:
A couple of minor nits, but otherwise, this looks great. With my lang team hat on (though not speaking for the rest of the lang team), :+1: to documenting "left to right" as the semantics.

src/expressions.md
joshtriplett: Withdrawing this comment; the only case that really bugs me is the idea that in `f(a, b)`, `f`, `a`, and `b` would all be considered "operands" of the function call expression, and I can live with that terminology in the interests of precision.

Havvy: FWIW, I didn't like the word beforehand, but after this and the tuple PR, I've found it works well. The only alternative I can think of would be to state that when we talk about sub-expressions, we only mean the top-most level, not the subexpressions of the subexpressions. And requiring global context seems like a bad idea.

src/expressions.md
Havvy: I listed every case where it is true. I was being lazy with that sentence prior. I'm still being lazy by not actually linking to those expression pages, but that can always be done later and they're in the sidebar.

nikomatsakis: One prominent case that is not mentioned here is `a = b`. Our execution odrer there, I *believe*, is to evaluate `b` before `a`, though I want to double check now =)

nikomatsakis: [Yes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dedd696fe07142ecf4bb4a4c53d5496f)

src/expressions.md
```diff
@@ -85,6 +84,53 @@ in the order given by their associativity.
...
+let mut one_two = vec![1, 2].into_iter();
+assert_eq!(
+    (1, 2),
+    (one_two.next().unwrap(), one_two.next().unwrap())
+);
+```
+
+> **Note**: Since this is applied recursively, expressions are also evaluated
+> from innermost to outermost, ignoring siblings until there are no inner
+> subexpressions.
```
nikomatsakis, line 132: I would prefer to extend this PR with coverage of 

* assignment expressions
* the order of "augmented assignment" like `+=`, particularly in the face of operator overloading -- this gets a bit subtle, see e.g. https://github.com/rust-lang/rust/issues/28160 -- I think at this point it's clear though that we just have to document the current behavior and live with it

How much did you test to verify the execution order for various corner cases? (e.g., overloaded index?) I think that what you wrote above is correct, but I'd love to see a comprehensive set of test cases added to the repository, e.g., in a place like `src/test/ui/evaluation_order.rs` =) this doesn't all have to fall on you, of course, we could just write to write up a list of those tests and see if we can advertise folks to write them up

nikomatsakis (changes requested):

src/expressions.md
RalfJung: Hm, I seem to recall @eddyb saying that it actually is even more complicated than that, but I forgot the details...

src/expressions.md
Havvy: I haven't done any testing of the corner cases. Although I did just make one for [Index](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=beecf916cf1c426d58e9f2261112b538). But given that for all of these operations, the actual method call happens after the operands are evaluated, I can't see execution order changing by using language known impls or trait impls.

I also am not sure a single `evaluation_order.rs` test file is the best place. I would think you'd want `src/test/ui/exprs/%expr_kind%/evaluation_order.rs` so that it's grouped with other tests of the expression kind. But of course, that'd require quite a lot of reorganization of the tests and ultimately all of this falls outside of the domain of the reference docs.

I also just did a quick test for [assignment operators](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7c3793070869bd4bff3687d8a2144bac), and they apparently evaluate the right expression before the left expression. That I did not know, so that does require updating this PR.

Havvy:
I need to still do compound assignment expressions, but I'm getting tired. Since I was modifying assignment expressions, I figured I'd give it the same treatment I gave tuples (and would like to give to all expressions).

<details>
<summary>Image of rendered change</summary>

image
</details>

src/expressions.md
nikomatsakis: @Havvy yeah I totally agree that a "directory" is the right place to sort those kinds of tests.

nikomatsakis:
@Havvy this is looking good, are you still expecting to make more additions here, or would you rather land for now?

Havvy:
I keep putting off finishing up the section on compound assignment; lack of energy after work or last week a patch dropped in an MMO. If you want to merge as is, I can send that section as its own PR.

Havvy:
Given that there will probably be needed discussion on compound assignment operators, I wish to land that part separately from this.

src/expressions/operator-expr.md
```diff
@@ -409,20 +409,35 @@ halfway between two floating point numbers.
...
 
+Evaluating assignment expressions begins by evaluating its operands. The
+assigned value operand is evaluated first, followed by the assigned place
+operand.
+
+> **Note**: This is different than other expressions in that the right operand
+> is evaluated before the left one.
+
+It then has the effect of first [dropping] the value at the assigned place,
+unless the place is an uninitialized local variable or field of a local
```
joshtriplett, line 430:
```suggestion
unless the place is an uninitialized local variable or an uninitialized field of a local
```

joshtriplett:
One minor clarification, and otherwise, r=me.

Issue #891, opened 2020-10-11
Documentation says that SSE4.1 includes SSE3, but LLVM says SSSE3


eduardosm:
The [list of available target features](https://doc.rust-lang.org/reference/attributes/codegen.html#the-target_feature-attribute) says that `sse4.1` implicitly enables `sse3` (but not `ssse3`). However, looking at LLVM source code, it looks like that SSE4.1 also includes SSSE3.

https://github.com/llvm/llvm-project/blob/544a6aa2674e3875e4014eafb101a982f9296439/llvm/lib/Target/X86/X86.td#L66-L82
```
def FeatureSSE1    : SubtargetFeature<"sse", "X86SSELevel", "SSE1",
                                      "Enable SSE instructions">;
def FeatureSSE2    : SubtargetFeature<"sse2", "X86SSELevel", "SSE2",
                                      "Enable SSE2 instructions",
                                      [FeatureSSE1]>;
def FeatureSSE3    : SubtargetFeature<"sse3", "X86SSELevel", "SSE3",
                                      "Enable SSE3 instructions",
                                      [FeatureSSE2]>;
def FeatureSSSE3   : SubtargetFeature<"ssse3", "X86SSELevel", "SSSE3",
                                      "Enable SSSE3 instructions",
                                      [FeatureSSE3]>;
def FeatureSSE41   : SubtargetFeature<"sse4.1", "X86SSELevel", "SSE41",
                                      "Enable SSE 4.1 instructions",
                                      [FeatureSSSE3]>;
def FeatureSSE42   : SubtargetFeature<"sse4.2", "X86SSELevel", "SSE42",
                                      "Enable SSE 4.2 instructions",
                                      [FeatureSSE41]>;
```

ehuss:
Yea, looks like a mistake. Ugh, I remember spending a lot of time pouring through LLVM code to try to make sure those were correct. 😦  Would you be willing to submit a PR to fix it?

eduardosm:
>  Would you be willing to submit a PR to fix it?

No problem, I'll do it

lzutao:
ssse3 implicits sse3.

PR #892, opened 2020-10-11
Specify that SSE4.1 includes SSSE3 instead of SSE3

eduardosm:
According to LLVM source code: https://github.com/llvm/llvm-project/blob/544a6aa2674e3875e4014eafb101a982f9296439/llvm/lib/Target/X86/X86.td#L66-L82

Fixes https://github.com/rust-lang/reference/issues/891


ehuss (approved):
Thanks!

PR #896, opened 2020-10-17
Update deprecated GitHub Actions commands.

ehuss:
The old method of echoing commands is being deprecated, see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/.


Havvy:
I don't know anything about how github actions works to review this.

Havvy (approved):
At worst, this can make CI fail. So approving since I don't think we can get a reviewer who knows github actions in a timely manner.

PR #897, opened 2020-10-18
mention how unions interact with dropping

RalfJung:
Cc https://github.com/rust-lang/rust/pull/77547
r? @ehuss

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -79,6 +79,16 @@ u.f1 = 2;
 Commonly, code using unions will provide safe wrappers around unsafe union
 field accesses.
 
+## Unions and `Drop`
+
+When a union is dropped, it cannot know which of its fields needs to be dropped.
+For this reason, all union fields must either be of a `Copy` type or of the
+shape `ManuallyDrop<_>`.  This ensures that a union does not need to drop
+anything when it goes out of scope.
+
+Like for structs and enums, it is possible to `impl Drop` for a union to
+manually define what happens when it gets dropped.
+
 ## Pattern matching on unions
 
 Another way to access union fields is to use pattern matching. Pattern matching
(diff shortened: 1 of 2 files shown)


src/items/unions.md
```diff
@@ -79,6 +79,16 @@ u.f1 = 2;
 Commonly, code using unions will provide safe wrappers around unsafe union
 field accesses.
 
+## Unions and `Drop`
+
+When a union is dropped, it cannot know which of its fields needs to be dropped.
+For this reason, all union fields must either be of a `Copy` type or of the
+shape `ManuallyDrop<_>`.  This ensures that a union does not need to drop
```
ehuss, line 86: Can you add a link for this? 
```suggestion
shape [`ManuallyDrop<_>`].  This ensures that a union does not need to drop
```

```
[`ManuallyDrop<_>`]: ../../std/mem/struct.ManuallyDrop.html

ehuss (commented):
Thanks! I think my earlier confusion was that I was looking at it from the perspective before the recent stabilization. Am I correct to say that before https://github.com/rust-lang/rust/pull/77547 that all fields were required to be `Copy`? If so, then my confusion was that it said `write to a field that doesn't implement Copy`, but why mention "Copy" if it wasn't possible to define a non-Copy field in the first place?

I see that it is now possible to have a non-Copy field with `ManuallyDrop`, which makes that statement make more sense, but previously it seemed confusing to me (unless I'm being dense and misunderstanding something).

RalfJung:
> Am I correct to say that before rust-lang/rust#77547 that all fields were required to be Copy? If so, then my confusion was that it said write to a field that doesn't implement Copy, buy why mention "Copy" if it wasn't possible to define a non-Copy field in the first place?

You are correct for *stable* Rust. The docs were likely written by someone who thought of what is possible when enabling nightly feature gates.

ehuss (approved):
Makes sense, TYVM!

PR #899, opened 2020-11-05
Referencify tuples

Havvy:
I'm not sure "initialized" is the correct word, but I feel like our
verbiage around place and value expressions need to be cleaned up
anyways. I originally went with "moved", but that's not really correct.
I think I want to call it a "value operand", but I need to discuss with
others on that first. In any case, it's a lot more correct than what
was there before.

In one example, I renamed `unit_x` to `point` because while it's technically a mathematical unit, it clashes with the unit tuple.

I also kept `element` instead of renaming to `field`.

### DIFF
--- a/src/expressions/tuple-expr.md
+++ b/src/expressions/tuple-expr.md
@@ -29,7 +29,7 @@ Examples of tuple expressions:
 | -------------------- | ------------ |
 | `()`                 | `()` (unit)  |
 | `(0.0, 4.5)`         | `(f64, f64)` |
-| `("x".to_string(), ) | `(String, )  |
+| `("x".to_string(), )` | `(String, )`  |
 | `("a", 4usize, true)`| `(&'static str, usize, bool)` |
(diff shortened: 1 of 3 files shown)


src/expressions/tuple-expr.md
```diff
@@ -9,23 +9,28 @@
...
-(0,); // single-element tuple
-(0); // zero in parentheses
-```
+Examples of tuple expressions:
+
+| Expression           | Type         |
+| -------------------- | ------------ |
+| `()`                 | `()` (unit)  |
+| `(0.0, 4.5)`         | `(f64, f64)` |
+| `("x".to_string(), ) | `(String, )  |
```
ehuss, line 32:
```suggestion
| `("x".to_string(), )` | `(String, )`  |
```

ehuss (commented):
:+1:

ehuss (approved):

PR #900, opened 2020-11-09
Document RangeFrom patterns

workingjubilee:
This PR documents RangeFrom-style patterns as a prelude to stabilizing https://github.com/rust-lang/rust/issues/67264, per the stabilization policy.

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -401,8 +401,8 @@ match tuple {
 
 > **<sup>Syntax</sup>**\
 > _RangePattern_ :\
-> &nbsp;&nbsp; _RangePatternBound_ `..=` _RangePatternBound_
-> &nbsp;&nbsp; _RangePatternBound_ `..`
+> &nbsp;&nbsp; &nbsp;&nbsp; _RangePatternBound_ `..=` _RangePatternBound_\
+> &nbsp;&nbsp; | _RangePatternBound_ `..`
 >
 > _ObsoleteRangePattern_ :\
 > &nbsp;&nbsp; _RangePatternBound_ `...` _RangePatternBound_
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -10,7 +10,6 @@
 > &nbsp;&nbsp; | [_IdentifierPattern_]\
 > &nbsp;&nbsp; | [_WildcardPattern_]\
 > &nbsp;&nbsp; | [_RestPattern_]\
-> &nbsp;&nbsp; | [_ObsoleteRangePattern_]\
 > &nbsp;&nbsp; | [_ReferencePattern_]\
 > &nbsp;&nbsp; | [_StructPattern_]\
 > &nbsp;&nbsp; | [_TupleStructPattern_]\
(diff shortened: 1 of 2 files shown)


src/patterns.md
```diff
@@ -402,6 +402,8 @@ match tuple {
 > **<sup>Syntax</sup>**\
 > _RangePattern_ :\
 > &nbsp;&nbsp; _RangePatternBound_ `..=` _RangePatternBound_
+> _RangeFromPattern_ :\
+> &nbsp;&nbsp; _RangePatternBound_ `..`
```
ehuss, line 406:
```suggestion
>
> _RangeFromPattern_ :\
> &nbsp;&nbsp; _RangePatternBound_ `..`
```

ehuss (commented):
Can this more formally define the meaning of the half-open range? Generally the definition shouldn't be given as an example.  

Additionally, can you add a small example to the example block below?

Can you also update the [Punctuation section](https://github.com/rust-lang/reference/blob/master/src/tokens.md#punctuation) to include a link for `..` to the "Range patterns"?

workingjubilee:
Whoops, sorry for falling off the face of the planet.
Need to think of a good example still, but was this more what you had in mind re: writing style?

src/patterns.md
```diff
@@ -402,6 +402,7 @@ match tuple {
 > **<sup>Syntax</sup>**\
 > _RangePattern_ :\
 > &nbsp;&nbsp; _RangePatternBound_ `..=` _RangePatternBound_
+> &nbsp;&nbsp; _RangePatternBound_ `..`
```
ehuss, line 405:
```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; _RangePatternBound_ `..=` _RangePatternBound_\
> &nbsp;&nbsp; | _RangePatternBound_ `..`
```

ehuss (changes requested):
It looks like there were some changes in the stabilization PR.  Do you think you can update with those changes?  Was it just the restriction of slice patterns, or were there other changes?

workingjubilee:
Just the restriction on slice patterns, but will update. Blargh!

ehuss:
Can you update with the restrictions on slice patterns?

workingjubilee:
Only if I can remember what I was doing, unfortunately.

Does the grammar look right?

src/patterns.md
```diff
@@ -694,7 +716,11 @@ match int_reference {
 > &nbsp;&nbsp; `[` _SlicePatternItems_<sup>?</sup> `]`
 >
 > _SlicePatternItems_ :\
-> &nbsp;&nbsp; [_Pattern_] \(`,` [_Pattern_])<sup>\*</sup> `,`<sup>?</sup>
+> &nbsp;&nbsp; _InSlicePattern_ \(`,` _InSlicePattern_)<sup>\*</sup> `,`<sup>?</sup>
+>
+> _InSlicePattern_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_PatternWithoutRange_]\
+> &nbsp;&nbsp; | [_ClosedRangePattern_]
```
ehuss, line 723: I believe the restriction is done after parsing, so this shouldn't be expressed as part of the grammar.  That is, `[0..]` is syntactically valid as a pattern, but is then later rejected.  I think it would be fine just to list that as a restriction below in prose.

(and, fwiw, _ClosedRangePattern_ isn't defined)

workingjubilee: Fair enough then.
Whoops.
Alright, let's see...

workingjubilee: Also a good point because: I forgot that it was only the version _without parentheses_ that is currently "up in the air".

ehuss (approved):
Thanks!

workingjubilee:
Thank _you_.

PR #903, opened 2020-11-12
Update where our chat is.

Havvy:
We're no longer using Discord as we're not a part of the docs team
anymore since there is no docs team anymore.


ehuss (approved):
:+1:

PR #906, opened 2020-11-21
Use `doc`, not `test` for fn item inner attr example

Havvy:
Using `test` causes a deny-by-default lint to trigger. See
https://github.com/rust-lang/rust/pull/79003 for more information.

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -290,12 +290,12 @@ responsibility to ensure that.
 [Outer attributes][attributes] are allowed on functions. [Inner
 attributes][attributes] are allowed directly after the `{` inside its [block].
 
-This example shows an inner attribute on a function. The function will only be
-available while running tests.
+This example shows an inner attribute on a function. The function is documented
+with just the word "Example".
 
 ```rust
-fn test_only() {
-    #![doc("Example")]
+fn documented() {
+    #![doc = "Example"]
 }
 ```
 
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -295,7 +295,7 @@ available while running tests.
 
 ```rust
 fn test_only() {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!  The sentence above the example should probably be updated (`The function will only be available while running tests.`) and the function name (`test_only`) should also probably be changed.

I think it would also be a good idea to document this behavior somehow.  AFAIK ([rustc_builtin_macros](https://github.com/rust-lang/rust/blob/40624dde6cdfad2576b4d795c97583188a814884/compiler/rustc_builtin_macros/src/lib.rs#L89-L95)), the only attributes this affects is `test` and `bench`.  I don't know how soon https://github.com/rust-lang/rust/issues/54726 might get stabilized, but if it might be a long time, maybe a brief mention at https://github.com/rust-lang/reference/blob/master/src/attributes/testing.md#the-test-attribute that says something along the lines of:  "similar to proc-macros, the `#[test]` attribute only works as an [outer attribute], and cannot be used as an [inner attribute]".

src/items/functions.md
```diff
@@ -295,7 +295,7 @@ available while running tests.
 
 ```rust
 fn test_only() {
-    #![test]
+    #![doc("Example")]
```
petrochenkov, line 298:
```suggestion
    #![doc = "Example"]
```
`doc("Example")` is not rejected by validation, but it doesn't do anything.

Havvy:
I want to focus this PR on just getting toolstate working, so the update to the test file shall be in its own PR.

ehuss (approved):
Thanks! Seems reasonable, I don't think there is any rush.  It's an extreme edge case, that hopefully some day will be removed, in which case there will be no need to document it.

PR #911, opened 2020-12-15
Removing ambiguity in type-layout.md

marcospb19:


marcospb19:
This line sounded confusing to me at first:
```
Types where all values have the same size and alignment known at compile time implement...
```

Like, the values have the same size, **and** they have the alignment known at compile time?

The context that comes ahead of it explains better, but I found this part to be confusing to people whose primary language isn't english.

Changed it to:
```
Types where all values have the same size and alignment (both known at compile time) implement...
```

Other possible change would be:

```
Types where all values have the same size and alignment, and both are known at compile time, implement...
```

Havvy:
Parenthesis are used for secondary information, and this is definitely not secondary. Your alternative looks good to me.

Havvy:
Oh, and this is my writing, so a personal thanks for the editing.

marcospb19:
hahaha no problem, thank you for your amazing writing!!!

I force pushed with the second suggestion, here's the diff now:

**-**
```
where all values have the same size and alignment known at compile time...
```
**+**
```
where all values have the same size and alignment, and both are known at compile time, ...
```

Havvy (approved):

Havvy:
💟 Thanks!

PR #912, opened 2020-12-16
Update unions for safe ManuallyDrop assignment.

ehuss:
Update for the change in https://github.com/rust-lang/rust/pull/78068.

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -66,14 +66,18 @@ unsafe {
 }
 ```
 
-Writes to `Copy` union fields do not require reads for running destructors, so
-these writes don't have to be placed in `unsafe` blocks
+Writes to [`Copy`] or [`ManuallyDrop`][ManuallyDrop] union fields do not
+require reads for running destructors, so these writes don't have to be placed
+in `unsafe` blocks
 
 ```rust
-# union MyUnion { f1: u32, f2: f32 }
-# let mut u = MyUnion { f1: 1 };
-#
+# use std::mem::ManuallyDrop;
+union MyUnion { f1: u32, f2: ManuallyDrop<String> }
+let mut u = MyUnion { f1: 1 };
+
+// These do not require `unsafe`.
 u.f1 = 2;
+u.f2 = ManuallyDrop::new(String::from("example"));
 ```
 
 Commonly, code using unions will provide safe wrappers around unsafe union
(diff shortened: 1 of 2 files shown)


ehuss:
@RalfJung Can you review this?

Havvy (approved):

src/types/union.md
```diff
@@ -5,12 +5,14 @@ a [`union` item][item].
 
 Unions have no notion of an "active field". Instead, every union access
 transmutes parts of the content of the union to the type of the accessed
-field. Since transmutes can cause unexpected or undefined behaviour, `unsafe` is
-required to read from a union field, or to write to a field that doesn't
-implement [`Copy`]. See the [item] documentation for further details.
+field. Since transmutes can cause unexpected or undefined behaviour, `unsafe`
+is required to read from a union field, or to write to a field that doesn't
+implement [`Copy`] or is a [`ManuallyDrop`] type. See the [item] documentation
```
RalfJung, line 10: a field that "is a `ManuallyDrop` type" makes little sense grammatically, IMO.
```suggestion
implement [`Copy`] or has a [`ManuallyDrop`] type. See the [item] documentation
```

src/items/unions.md
```diff
@@ -66,14 +66,18 @@ unsafe {
 }
 ```
 
-Writes to `Copy` union fields do not require reads for running destructors, so
-these writes don't have to be placed in `unsafe` blocks
+Writes to [`Copy`] or [`ManuallyDrop`][ManuallyDrop] union fields do not
+require reads for running destructors, so these writes don't have to be placed
+in `unsafe` blocks
```
RalfJung, line 71: The actual implementation goes a bit further, so `u.f1.g2 = ...` is safe if *that assignment* is at `ManuallyDrop` or `impl Copy` type... but on stable code that is impossible to observe since all fields must be `ManuallyDrop` or `impl Copy`. Not if the reference should even mention it, then.

RalfJung:
Looks good, aside from the nit and the comment that I left. :)

src/items/unions.md
ehuss: I think it's fine to leave out for now.  Is it even possible on nightly to have a non-Copy/ManuallyDrop type?

RalfJung: Yeah, with a feature flag you can use any type without drop glue.

In principle even those could be safe to assign, so we could remove this safety check entirely... for now I went with the smaller change that should just make the safety check unreachable on stable.

Issue #599, opened 2019-04-21
Style, formatting, and policy guide


ehuss:
I think it might be useful to have a guide for style, formatting, and what should and should not be included. If there is, care should be taken to communicate that they are guidelines on how documentation *should* be written, not necessarily how it *must* be written, since overly pedantic rules can discourage contribution.

## Policy
- When should the reference include *why* something is done in a particular way? Often the reference avoids the "why", but I think it can sometimes provide interesting context.
    - For example, `type_length_limit` explains what it does, but not why it is there. I think it could be useful to say something like: `type_length_limit` is used to prevent the compiler from hanging and to better deal with [polymorphic recursion](https://en.wikipedia.org/wiki/Polymorphic_recursion).
    - There are many other examples where the motivation for something is hidden away in some RFC.

- How can reference writers know when something should be intentionally undefined? (Like certain evaluation orders.) Seems like it would be easy to accidentally document some behavior without knowing it shouldn't be specified. Maybe someone in-the-know could call out exactly what should be left undefined, I suspect the list is not very long.

- How to deal with rustc version numbers? Is the rustc version number the same as the "The Rust Language" version number? Should the reference actively avoid mentioning version numbers? In general, they could be useful in some situations, but generally are difficult to mark what is changed/added/removed with enough granularity, and to deal with changes over time. (See #60)

- Should the reference document actual rust behavior (including bugs)?  Or should it document the *intended* behavior? Or both?
    - Related: If there's a known issue, should it be encouraged or discouraged to link to the Github issue?

- Should the reference mention things that are deprecated?

- Should the reference avoid talking about past behavior?

- How should overlap with other documentation be handled? Should the reference duplicate everything and be completely standalone? Or can it link to other books?
    - It could be useful to link to other sources like TRPL. The reference is tends to be terse, and pointing to guide-style documentation could helpful for someone learning how to use something. However, lots of links would be very noisy.

- How should the reference handle `rustc`-specific things? Where does the line get drawn? What is OK to mention? For example, compiler flags? Should it just be opportunistically helpful? For example, `extern crate` briefly mentions Cargo's dash-to-underscore rename. Another example: is the filesystem layout part of the language or part of the implementation? What is implementation-specific?

- What is the policy for linking to historical information, such as RFCs, PRs, GitHub issues, blog posts, etc?  These sources are mostly static and not updated, and thus may provide outdated or incorrect information. However, as historical context they are useful.

[...]

PR #914, opened 2020-12-22
Document array expression with a const.

ehuss:
Documentation for https://github.com/rust-lang/rust/pull/79270.

I did not include any details about drop behavior with `[a; 0]`, as it seems like https://github.com/rust-lang/rust/issues/74836 is unresolved, though I can add that if desired.  (I'm not sure if that would go in the array-expr chapter, or the destructor chapter.)

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -10,19 +10,16 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_] ( `,` [_Expression_] )<sup>\*</sup> `,`<sup>?</sup>\
 > &nbsp;&nbsp; | [_Expression_] `;` [_Expression_]
 
-An _[array](../types/array.md) expression_ can be written by
-enclosing zero or more comma-separated expressions of uniform type in square
-brackets. This produces an array containing each of these values in the
-order they are written.
+An _[array] expression_ can be written by enclosing zero or more
+comma-separated expressions of uniform type in square brackets. This produces
+an array containing each of these values in the order they are written.
 
 Alternatively there can be exactly two expressions inside the brackets,
-separated by a semi-colon. The expression after the `;` must have type
-`usize` and be a [constant expression],
-such as a [literal](../tokens.md#literals) or a [constant
-item](../items/constant-items.md). `[a; b]` creates an array containing `b`
-copies of the value of `a`. If the expression after the semi-colon has a value
-greater than 1 then this requires that the type of `a` is
-[`Copy`](../special-types-and-traits.md#copy).
+separated by a semicolon. The expression after the `;` must have type `usize`
+and be a [constant expression], such as a [literal] or a [constant item]. `[a;
+b]` creates an array containing `b` copies of the value of `a`. If the
+expression after the semicolon has a value greater than 1 then this requires
+that the type of `a` is [`Copy`].
 
 ```rust
 [1, 2, 3, 4];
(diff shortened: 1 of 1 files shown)


ehuss:
@RalfJung Sorry, bugging you to review another one.

RalfJung:
This sounds correct. However, do you want to mention the subtleties around `drop`? See https://github.com/rust-lang/rust/pull/79270#issuecomment-735253236 and the surrounding discussion.

ehuss:
Hm, yea, as mentioned in the description, I intentionally left that out because I was uncertain what to say.  Since https://github.com/rust-lang/rust/issues/74836 is unresolved, should it document that `[non_const_expr; 0]` evaluates once and then drops, and then include a note that this is not working as intended?  Or just leave that part out completely and only say `[CONST; 0]` does not evaluate `CONST` (it seems a little weird to say that without contrasting it with the non-const example).   Also, would you recommend putting that in the Destructors chapter, the Array Expression chapter, the Constant Evaluation chapter, or maybe in multiple locations?

Havvy:
I would document it how it is now (`[CONST; 0]`) does not evaluate `CONST`), but include a warning directly after contrasting this behavior with the non-const example.

RalfJung:
> I would document it how it is now (`[CONST; 0]`) does not evaluate `CONST`), but include a warning directly after contrasting this behavior with the non-const example.

Sounds good.

As to where to put this... I have little idea about the structure of the reference, so I'll leave that up to you. I would not document the same thing in multiple places, but adding cross-references might be a good idea.

ehuss:
OK, I included a note on the `[a; 0]` behavior.  I waffled a little on whether or not to note the bug.  We talked about the policy a little bit in #599, and it seemed like the consensus is that it doesn't hurt to include the occasional note about a bug.  I can remove it whenever https://github.com/rust-lang/rust/issues/74836 is closed.

src/expressions/array-expr.md
```diff
@@ -16,14 +16,21 @@ brackets. This produces an array containing each of these values in the
...
+separated by a semicolon. The expression after the `;` must have type
 `usize` and be a [constant expression],
 such as a [literal](../tokens.md#literals) or a [constant
 item](../items/constant-items.md). `[a; b]` creates an array containing `b`
-copies of the value of `a`. If the expression after the semi-colon has a value
+copies of the value of `a`. If the expression `b` after the semicolon has a value
 greater than 1 then this requires that the type of `a` is
 [`Copy`](../special-types-and-traits.md#copy), or `a` must be a path to a
-constant item.
+constant item. When `b` evaluates to 0, the expression `a` is evaluated once
```
RalfJung, line 26: "the expression `a` is evaluated once" is *always* true for non-constant-items. So I think it is a bit strange to make `0` sound like a special case here.

I'd describe this (probably in a separate paragraph) as:
* for constant items, the expression is instantiated `b` times. (So if `b == 0`, it is not instantiated at all.)
* for other `a`, the expression is evaluated exactly once (and then the result is copied as needed).

ehuss: Oh, that makes sense and is much clearer to me now.  Thanks for your patience!

RalfJung:
LGTM. :)

src/expressions/array-expr.md
```diff
@@ -10,26 +10,35 @@
...
+b]` creates an array containing `b` copies of the value of `a`. If the
+expression after the semicolon has a value greater than 1 then this requires
+that the type of `a` is [`Copy`], or `a` must be a path to a constant item.
+
+When the repeat expression `a` is a constant item, it is evaluated `b` times.
+If `b` is 0, the constant item is not evaluated at all. For expressions that
+are not a constant item, it is evaluated exactly once, and then the result is
+copied `b` times.
+
+> **Note:** In the case where `b` is 0, and `a` is a non-constant item, there
```
Havvy, line 29: Can you use the warning for this?

```
<div class="warning">

Warning: This is an example warning.

</div>
```

Havvy (changes requested):
r+ after switching the note to a warning.

src/expressions/array-expr.md
ehuss: Done!

Havvy (approved):

Havvy:
And merged. 💟 Thanks!

PR #915, opened 2020-12-27
Compound operator expressions

Havvy:
This is the final part of #248, split off from #888 since I have questions on this part and don't want to block #888 on it.

Specifically, is this desugaring correct? I asked on Zulip a few weeks ago, but got no definitive answer.

Also, is it possible to have a relative link to a test file?

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -498,8 +498,10 @@ fn example() {
 <div class="warning">
 
 Warning: The evaluation order of operands swaps depending on the types of the
-operands. Try not to write code that depends on the evaluation order of operands
-in compound assignment expressions. See [this test] for an example of using this
+operands: with primitive types the right-hand side will get evaluated first,
+while with non-primitive types the left-hand side will get evaluated first.
+Try not to write code that depends on the evaluation order of operands in
+compound assignment expressions. See [this test] for an example of using this
 dependency.
 
 </div>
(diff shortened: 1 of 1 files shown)


ehuss:
I can't answer if this is correct, but reading through it looks good to me (and #888 also still looks good to me).

> Also, is it possible to have a relative link to a test file?

It is not.  If you want to leave the link in, I think that's fine, but we unfortunately don't currently have the ability to link-check external links, so there is some risk that it will become broken over time.

Havvy:
@nikomatsakis Do you have bandwidth to review this?

src/expressions/operator-expr.md
```diff
@@ -441,24 +441,74 @@ x = y;
...
+
+# let (mut a1, a2) = (Addable, Addable);
+  AddAssign::add_assign(&mut a1, a2);
+}
 ```
 
+<div class="warning">
+
+Warning: The evaluation order of operands swaps depending on the types of the
+operands. Try not to write code that depends on the evaluation order of operands
```
joshtriplett, line 501: This seems sufficiently unusual to warrant further elaboration. The text above ("Otherwise, this expression is syntactic sugar ...") doesn't make it clear that *because* of that sugar, the evaluation order changes.

```suggestion
operands: with primitive types the right-hand side will get evaluated first, while with non-primitive types the left-hand side will get evaluated first.  Try not to write code that depends on the evaluation order of operands
```

joshtriplett:
I posted one comment seeking further clarification; other than that, this LGTM.

src/expressions/operator-expr.md
Havvy: Yeah, I think the distinction got lost somewhere in the ten different ways I tried to formulate it before settling on this one. Adding it explicitly to the warning sounds good to me, so I did so.

PR #917, opened 2021-01-05
linkage.md: update link to FFI section of the Book.

wsh:
Hi! I've tried to verify that this works but using `mdbook build` creates a broken link locally (in the output, `book` is the directory containing the reference, not the Book). This format, however, matches the link to the book from [`runtime.md`](https://doc.rust-lang.org/stable/reference/runtime.html#standard-behavior), which does work.


Havvy (approved):
💟 Thanks!

PR #918, opened 2021-01-05
Update mdbook

weihanglo:
- Changelog to 0.4.5: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-045)
- mdbook security advisory: https://blog.rust-lang.org/2021/01/04/mdbook-security-advisory.html

Some breaking changes were introduced in 0.4.0 but I skimmed The References built by 0.4.5 and no regression detected.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Install mdbook
       run: |
         mkdir bin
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!  The version here doesn't really matter since it is only used for tests, but it doesn't hurt to update!

PR #919, opened 2021-01-06
Add "Logic errors" as behavior not considered unsafe

ojeda:
In https://github.com/rust-lang/rust/issues/80657 and https://github.com/rust-lang/rust/pull/80681 it is discussed how to clarify/define what a "logic error" is and what are their consequences. The reference should mention them as well.

We could use other term instead of "unspecified", e.g. "unpredictable"; in case we want to emphasize that the results may not the same between runs, builds, kinds of build, etc. For instance, from the description of integer overflow, it looks like the implementation needs to choose some reasonable behavior (e.g. a panic or wrap around), even if not documented (therefore, "unspecified", in terms of the C and C++ standards). However, logic errors seem to require more freedom: the implementation may not even be able (or want) to predict what may happen, so it cannot "reasonably choose" a particular behavior, even if such behavior is guaranteed to be safe (similar to deadlocks, although in that case the set of possible behavior is known, even if nondeterministic between runs).

As I mentioned in one of the linked discussions, it might be a good idea to define a separate, clear term for this kind of behavior that is considered erroneous and unpredictable, yet safe. So not exactly "unspecified", at least as used in the C and C++ standards -- it is closer to "safe UB" than "unspecified". In other words, in the same spirit as C's "bounded UB" or Ada's "bounded error". Ideally, the term would not overload another common term in other language specs (like UB).

### DIFF
--- a/src/behavior-not-considered-unsafe.md
+++ b/src/behavior-not-considered-unsafe.md
@@ -36,4 +36,21 @@ semantics.
 See [RFC 560] for error conditions, rationale, and more details about
 integer overflow.
 
+##### Logic errors
+
+Safe code may impose extra logical constraints that can be checked
+at neither compile-time nor runtime. If a program breaks such
+a constraint, the behavior may be unspecified but will not result in
+undefined behavior. This could include panics, incorrect results,
+aborts, and non-termination. The behavior may also differ between
+runs, builds, or kinds of build.
+
+For example, implementing both `Hash` and `Eq` requires that values
+considered equal have equal hashes. Another example are data structures
+like `BinaryHeap`, `BTreeMap`, `BTreeSet`, `HashMap` and `HashSet`
+which describe constraints on the modification of their keys while
+they are in the data structure. Violating such constraints is not
+considered unsafe, yet the program is considered erroneous and
+its behavior unpredictable.
+
 [RFC 560]: https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md


src/behavior-not-considered-unsafe.md
```diff
@@ -37,3 +37,18 @@ See [RFC 560] for error conditions, rationale, and more details about
 integer overflow.
 
 [RFC 560]: https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md
```
Havvy, line 39: As per reference style rules, this link needs to be at the bottom.

src/behavior-not-considered-unsafe.md
```diff
@@ -37,3 +37,18 @@ See [RFC 560] for error conditions, rationale, and more details about
...
+
+##### Logic errors
+
+Some safe functionality may impose extra logic constraints that cannot be
```
Havvy, line 43:
```suggestion
Some safe functionality may impose extra logical constraints that can be
```

The way you have it written now, it's a double negative.

src/behavior-not-considered-unsafe.md
```diff
@@ -37,3 +37,18 @@ See [RFC 560] for error conditions, rationale, and more details about
...
+checked at neither compile-time nor runtime, for different reasons. If a user
```
Havvy, line 44:
```suggestion
checked at neither compile-time nor runtime, for different reasons. If a program
```

Programs, not users, break constraints.

```suggestion
checked at neither compile-time nor runtime. If a user
```

"for different reasons" doesn't actually mean anything.

src/behavior-not-considered-unsafe.md
Havvy, line 43:
```suggestion
Safe code may impose extra logic constraints that cannot be
```

src/behavior-not-considered-unsafe.md
```diff
@@ -37,3 +37,18 @@ See [RFC 560] for error conditions, rationale, and more details about
...
+breaks such a constraint, the behavior is unspecified but will not result in
+undefined behavior. This could include panics, incorrect results,
+aborts, and non-termination. The behavior may also differ between runs,
+builds or kinds of build.
+
+For instance, data structures like `BinaryHeap`, `BTreeMap`, `BTreeSet`,
```
Havvy, line 50:
```suggestion
For example, data structures like `BinaryHeap`, `BTreeMap`, `BTreeSet`,
```

Trying to keep all examples greppable with `example` somewhere in them.

src/behavior-not-considered-unsafe.md
```diff
@@ -37,3 +37,18 @@ See [RFC 560] for error conditions, rationale, and more details about
 integer overflow.
 
 [RFC 560]: https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md
+
+##### Logic errors
+
+Some safe functionality may impose extra logic constraints that cannot be
+checked at neither compile-time nor runtime, for different reasons. If a user
+breaks such a constraint, the behavior is unspecified but will not result in
```
Havvy, line 45:
```suggestion
breaks such a constraint, the behavior may be unspecified but will not result in
```

It's not unreasonable to possibly specify behavior for code that breaks contracts.

Havvy (changes requested):
Mostly little wording changes.

Can you also point to the Hash+Eq contract defined by Hash? https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq ; That way it's not just data structures that shown, but also that trait invariants can be broken in safe code. This is important because it means that unsafe code can't rely on invariants defined by safe traits.

src/behavior-not-considered-unsafe.md
ojeda: > "for different reasons" doesn't actually mean anything.

Yeah, it is a bit informal for the reference. Initially I was thinking that a reader might want to know what kind of things wouldn't be possible to enforce, so I thought about giving some examples, and then reduced it... but yeah, it looks out of place.

src/behavior-not-considered-unsafe.md
ojeda: Ah, yeah, that is my Spanish speaking :-)

src/behavior-not-considered-unsafe.md
ojeda: I see your point. Hmm... when I was writing this, I was thinking about this as a definition, i.e. that if somewhere else a "logic error" is referenced, then it means the behavior is unspecified. Thus if you actually specify what happens when some constraints are broken for some function, then you shouldn't use the term "logic error" (or whatever we chose for that term). So, for instance, an `unwrap()` panicking wouldn't be regarded as a "logic error".

But I agree with you, because this is not actually written as a definition (it uses "logic error" rather than defining it) and the file isn't meant to be a list of definitions either. Plus giving a very specific meaning to a term as common as "logic error" would create confusion. Nevertheless, I wonder if such a list of semi-formal definitions would be a good idea (similar to what I comment in the OP).

ojeda:
> This is important because it means that unsafe code can't rely on invariants defined by safe traits.

Indeed, which is one reason it would be great to actually list all the "logic errors" instances in core etc.; similar to the UB and unspecified lists in the C standard.

(I haven't spelled this out since it was just an example, but I can split the two examples into two paragraphs and give more detail to this one if you prefer).

src/behavior-not-considered-unsafe.md
Havvy: FWIW, this entire page is basically non-normative. It exists mainly as a reminder to the user that `safe` doesn't actually mean safe in general.

ojeda: Ah, thanks for the pointer, I didn't know that. I thought everything was more or less normative (well, "equally normative", since the intro says the document isn't normative anyway).

src/behavior-not-considered-unsafe.md
```diff
@@ -36,4 +36,21 @@ semantics.
...
+aborts, and non-termination. The behavior may also differ between
+runs, builds or kinds of build.
+
+For example, implementing both `Hash` and `Eq` requires that values
+considered equal have equal hashes. Another example are data structures
+like `BinaryHeap`, `BTreeMap`, `BTreeSet`, `HashMap` and `HashSet`
+which describe constraints on the modification of their keys while
+they are in the data structure. Violating such constraints is not
+considered unsafe, yet the program is considered erroneous and
+its behavior, unpredictable.
```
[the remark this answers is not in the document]
ojeda: I was mainly putting it there in place of the verb (since I omitted it to avoid repeating it). I have now tried to look up if that is done or not in English (like in Spanish), but it looks like it is only done when the clauses are split by a semicolon and no conjunction? I am curious now... :-)

Havvy:
I forgot to mention, please don't force push after a review. It makes it impossible to see what has changed from push to push.

Havvy (approved):

Havvy:
cc @RalfJung ; does this look alright?

ojeda:
> I forgot to mention, please don't force push after a review. It makes it impossible to see what has changed from push to push.

Yeah... GitHub should really fix that.

Thanks for the review and your patience with my English!

Havvy:
(This entire comment is off topic)

Github cannot really preserve the history after a force push because you are explicitly telling Github to rewrite the history for the branch that you force push to. It's how git works. So, Github can't "fix" it as it's not actually broken. As such, the general rule of thumb is that once you know somebody has interacted with a commit, you consider the commit immutable. If the project wants to later squash all your commits into one, they have that ability as an option on the Merge button (or can do so manually).

ojeda:
(This entire comment is off topic too)

> Github cannot really preserve the history after a force push because you are explicitly telling Github to rewrite the history for the branch that you force push to. It's how git works. So, Github can't "fix" it as it's not actually broken.

No, that's not how Git works, nor how GitHub works. Git does preserve the history, as does GitHub. You can even do the diff between force pushes if you build the URL yourself, e.g. for this case you would have copy-pasted the two hashes as indicated by GitHub above:

    https://github.com/rust-lang/reference/compare/950d856..a747328

It is just they never bothered to fix their UI while in the "Conversation" tab.

(In fact, all the commits from all the forks are shared in their backend so they added a warning because it is easy to make it look like a commit is in an upstream repo)

>  As such, the general rule of thumb is that once you know somebody has interacted with a commit, you consider the commit immutable.

No, the whole "push new commits to fix a PR" is a broken workflow. Force-pushing is wrong when you are overwriting what is already merged, not when re-submitting a PR because that is outside the repo's history. The commits and their reviews are preserved in all cases.

In fact, in Git "native" projects such as the Linux kernel and Git itself have always used the equivalent of "force-pushing" for updating patch series. That is, we resend the patch series; and similarly, the previous versions of the series and their reviews are always preserved in the ML archive.

GitHub simply never added a proper UI for that, so people started adding commits upon commits to workaround it. So the story is actually backwards ;-)

I am OK with projects that prefer commits since in GitHub's UI that is easier, but it is definitely not how it is supposed to work in a proper Git workflow.

> If the project wants to later squash all your commits into one, they have that ability as an option on the Merge button (or can do so manually).

The commit message is mainly the responsibility of the author of the commit, not of the reviewers or the committer. That is why you want to send clean PRs and force push them. In fact, some projects require extra annotations if the committer has done any changes at all, however trivial they are (including in the commit message).

RalfJung:
Hm... yes the text makes sense, but I am not sure about the best place for putting it. Currently we probably don't have a better place than this though.

This touches on the larger story of "language UB" vs "library UB" -- there's assumptions *the compiler* makes about what the code does (and violating them is language UB), and there are assumptions that *libraries* make about what their callers do (and violating them is library UB). The "list of UB" in the reference is certainly about language UB, these logic errors are a concept related to library APIs and thus more closely related to library UB. Related to this is also the distinction of validity invariants vs safety invariants, which one could call "language invariants" and "library invariants", respectively.

My concern with this change is that it might lead to even more confusion between language UB and library UB. But maybe this is still okay for now. Longer-term, the reference should more clearly tease apart these two concepts.

Mark-Simulacrum:
It might make sense to try and put this in the API guidelines, which is closer to library and generally feels better than documenting non-language concerns in the reference.

Havvy:
API guidelines sounds like a good place to me. This chapter is entirely about things that don't actually concern the language, so it doesn't technically fit the reference, but it's a good chapter to have in juxtaposition of the chapter that says what is, so I'm fine with having it here nonetheless.

RalfJung:
> This chapter is entirely about things that don't actually concern the language, so it doesn't technically fit the reference, but it's a good chapter to have in juxtaposition of the chapter that says what is, so I'm fine with having it here nonetheless.

I think it'd be better to have a proper introduction of language UB vs library UB. But this is probably a longer-term project which also involves adjusting the stdlib documentation to better distinguish these two kinds of UB.

Mark-Simulacrum:
Yes, I agree that having the distinction between the kinds of UB in the reference makes sense. I think the impacts of that on libraries (e.g. this notion of logic error / safe but unspecified behavior) might be better suited for API guidelines or similar places.

ojeda:
What is considered library? Is it always a clear cut? For instance, I guess one could argue the `Hash` and `Eq` example to be part of the language itself (and not just the library) since it involves a built-in operator trait. In such case, one could argue it is "language UB" (but safe, in this case).

Similarly, in C one could argue `stddef.h` is part of the "core language" since it is one of the headers that are always present, even in freestanding implementations (plus a "language operator" like `sizeof` or pointer subtraction return values of type `size_t` or `ptrdiff_t` respectively, which are introduced by that header, etc. etc.).

So I definitely agree it is a good idea to define these differences between kinds of UB in the reference somewhere.

Havvy:
@ojeda The language aspect is that `x == y` desugars to an `Eq::eq` method call, but the library semantics are what `Eq::eq()` is supposed to mean. The language does not care at that is means equality. This is very clear cut.

@RalfJung Language UB vs Library UB should definitely be documented in the Behavior Considered Unsafe chapter. That said, I think it's orthogonal to logic errors not being UB and that might be surprising, so specifically stating it is important.

ojeda:
@Havvy I am not talking about what equality means, but about the constraint itself. The constraint may be spelled out in a library file now; but it could very well be that it had been spelled out in the language itself. C++ also has its fair share of magic bits that blend the line between the library and the language; even if the standard delineates what is "library" and what is "language". 

Anyway, my point is that I believe it is not that important whether some UB comes from the library or from the language -- defining clear terms and what kind of consequences each entails (bounded/safe UB? unbounded/unsafe UB? unspecified? ...) is more important; because those are then used to document things in code/libraries/etc. and in day-to-day conversation.

RalfJung:
> What is considered library? Is it always a clear cut?

Yes, I think it is fairly clear-cut: the *language* is everything that an interpreter like Miri needs to implement "by hand"; everything else is a library.

> I am not talking about what equality means, but about the constraint itself. The constraint may be spelled out in a library file now; but it could very well be that it had been spelled out in the language itself.

Sure, but it wasn't. That's like saying "we are using `fn` as the keyword for functions but we could have used `fun`". It's true, but besides the point. When designing a language, there is some wiggle room for what to make a language primitive and what to leave to the library; but Rust made certain choices here and once that choice is made, *for Rust*, the answer is clear-cut.

Now, one could argue that we should permit other Rust implementations to make different choices here than `rustc`. Maybe this is what you mean.  But I still think there is such a thing as "core Rust" that we can describe precisely and than contains all language UB, and we can define the API surface of libstd on top of that and add library UB where needed.

Anyway, for `Eq` and `Hash`, *there is no UB*, this this is certainly a library concept and not a language concept.

> Language UB vs Library UB should definitely be documented in the Behavior Considered Unsafe chapter. That said, I think it's orthogonal to logic errors not being UB and that might be surprising, so specifically stating it is important.

I think it might be confusing because "logic error" is a library concept, which I feel like shouldn't occur on a page dedicated to language UB. But, as I already said, I have no better place to suggest either.

ojeda:
> Yes, I think it is fairly clear-cut: the _language_ is everything that an interpreter like Miri needs to implement "by hand"; everything else is a library.

And a language *can* include such a constraint. In fact, one can argue we don't know if it does or doesn't, since the reference is non-normative and not complete (see below).

> Sure, but it wasn't. That's like saying "we are using `fn` as the keyword for functions but we could have used `fun`". It's true, but besides the point. When designing a language, there is some wiggle room for what to make a language primitive and what to leave to the library; but Rust made certain choices here and once that choice is made, _for Rust_, the answer is clear-cut.

First of all, I was merely posing a question to try to understand the mind-frame of the writers of the reference, not arguing for or against it particular choices, *precisely* because I know there is this wiggle room.

Second, the reference is non-normative and, most importantly, a *work in progress*. So no, things are never set in stone -- not even in formally standardized documents that are supposed to be normative. Furthermore, it doesn't really matter that "Rust made certain choices" when discussing the changes to the reference itself: it is a cyclic argument.

Third, the key point here is where are these "certain choices" documented and how. In the introduction, the reference claims the "standard library" is not documented by it. Not only that is not clear-cut at all, it is not even defined *what* the "library" is supposed to be or include or not. One can claim it is "everything not in the reference", but that would require assuming the reference is complete, which it isn't (as it claims itself). One can also claim it is "everything documented in the library" too, but as said above, that requires clarification when written from the point of view of the reference.

Finally, if one wants to take away something from this, it should be "this probably needs to be clarified", not "you are making a pointless/vacuous argument"; which is non-constructive.

> Now, one could argue that we should permit other Rust implementations to make different choices here than `rustc`. Maybe this is what you mean.

Well, one of the points of having a reference is to make it clear what and what cannot be done by implementations, definitely; but more importantly, and this is what I was referring about, is to define terms and vocabulary that everyone can refer back to. While instances of "library UB" may be defined in the library itself, the *concept* of "library UB" should be in the reference. That is to say: to define e.g. "language UB" and "library UB", one needs to know where the line between language and library is -- it does not particularly matter where it is in a given language, just that it is clearly defined what is what.

> But I still think there is such a thing as "core Rust" that we can describe precisely and than contains all language UB, and we can define the API surface of libstd on top of that and add library UB where needed.

Agreed.

PR #901, opened 2020-11-11
document const generics

lcnr:
blocked on the stabilization of `min_const_generics`.

PR #258, opened 2018-03-03
Generic parameters

matthewjasper:
Putting this up to get some feedback. Second PR contains all the major changes.

* Add grammar for trait bounds and type parameters (and uses it for trait objects) based on the grammar from @brauliobz's [branch](https://brauliobz.github.io/rust-reference/grammar.html#items), as well as nikomatsakis's [parser](https://github.com/nikomatsakis/rustypop/blob/master/src/rusty.lalrpop) and jorendorff's [grammar](https://github.com/jorendorff/rust-grammar).
* Gives lifetime parameters some recognition
* Documents HRTB (breifly)

cc #256

PR #921, opened 2021-01-06
Add more details about const generics.

ehuss:
This is an extension of #901 with some more details about const generics.

This is not intended to be exhaustive of everything regarding const generics, just to add a little bit more information that I'm aware of.

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -33,8 +33,9 @@ struct Ref<'a, T> where T: 'a { r: &'a T }
 struct InnerArray<T, const N: usize>([T; N]);
 ```
 
-The generic parameters are in scope within the item definition where they are
-declared.
+Generic parameters are in scope within the item definition where they are
+declared. They are not in scope for items declared within the body of a
+function as described in [item declarations].
 
 [References], [raw pointers], [arrays], [slices][arrays], [tuples], and
 [function pointers] have lifetime or type parameters as well, but are not
@@ -42,7 +43,7 @@ referred to with path syntax.
 
 ### Const generics
 
-Const generic parameters allow items to be generic over constant values. The
+*Const generic parameters* allow items to be generic over constant values. The
 const identifier introduces a name for the constant parameter, and all
 instances of the item must be instantiated with a value of the given type.
 
@@ -52,11 +53,10 @@ instances of the item must be instantiated with a value of the given type.
 The only allowed types of const parameters are `u8`, `u16`, `u32`, `u64`, `u128`, `usize`
 `i8`, `i16`, `i32`, `i64`, `i128`, `isize`, `char` and `bool`.
 
-Const parameters can generally be used anywhere a [const item] can be used,
-with the exception of the definition of any [item] within the body of a
-function, and can only be used as standalone expressions in [types] and
-[array repeat expressions] (described below). That is, they are allowed in the
-following places:
+Const parameters can be used anywhere a [const item] can be used, with the
+exception that when used in a [type] or [array repeat expression], it must be
+standalone (as described below). That is, they are allowed in the following
+places:
 
 1. As an applied const to any type which forms a part of the signature of the
    item in question.
@@ -111,7 +111,7 @@ fn foo<const N: usize>() {
 ```
 
 As a further restriction, const parameters may only appear as a standalone
-argument inside of [types] and [array repeat expressions]. In those contexts,
+argument inside of a [type] or [array repeat expression]. In those contexts,
 they may only be used as a single segment [path expression], possibly inside a
 [block] (such as `N` or `{N}`). That is, they cannot be combined with other
 expressions.
@@ -131,9 +131,10 @@ A const argument in a [path] specifies the const value to use for that item.
 The argument must be a [const expression] of the type ascribed to the const
 parameter. The const expression must be a [block expression][block]
 (surrounded with braces) unless it is a single path segment (an [IDENTIFIER])
-or a [literal] (with a possibly leading `-` token). This syntactic restriction
-is necessary to avoid requiring infinite lookahead when parsing an expression
-inside of a type.
+or a [literal] (with a possibly leading `-` token).
+
+> **Note**: This syntactic restriction is necessary to avoid requiring
+> infinite lookahead when parsing an expression inside of a type.
 
 ```rust
 fn double<const N: i32>() {
@@ -155,6 +156,10 @@ fn example() {
 When there is ambiguity if a generic argument could be resolved as either a
 type or const argument, it is always resolved as a type. Placing the argument
 in a block expression can force it to be interpreted as a const argument.
+
+<!-- TODO: Rewrite the paragraph above to be in terms of namespaces, once
+    namespaces are introduced, and it is clear which namespace each parameter
+    lives in. -->
 
 ```rust,compile_fail
 type N = u32;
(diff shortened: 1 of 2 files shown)


ehuss:
Some notes about things I'm uncertain about:

- I don't understand lazy_normalization_consts, so I don't know what needs to be said about "anonymous constants".
- I did not mention that impl parameters must be constrained.  I believe that goes for all parameters (type, lifetime, and const), and should probably be mentioned [here](https://github.com/rust-lang/reference/blob/master/src/items/implementations.md#generic-implementations), but I do not think there is anything special about const (though I may be wrong). I figure this could be added separately.
- I don't think `expr as TYPE < 0` needs to be called out as wrong (since `<` is interpreted as part of the type), but I am uncertain. I also don't think this is specific to const generics, but was something I noticed while reviewing it.
- Does not discuss inference, since type inference is mostly undocumented.
- Should it include the note about exhaustiveness?  See the "exhaustiveness checking" section in https://github.com/rust-lang/rust/pull/79135.  I think that applies to all generic parameters (including types and lifetimes), so I don't think it is specific to const generics. I'm also uncertain how to write it (is it just the impl params must be used in the impl type?).

src/items/generics.md
```diff
@@ -43,18 +43,137 @@ struct Ref<'a, T> where T: 'a { r: &'a T }
...
+<!-- TODO: update above to say "introduces a name in the [value namespace]"
+    once namespaces are added. -->
+
 The only allowed types of const parameters are `u8`, `u16`, `u32`, `u64`, `u128`, `usize`
 `i8`, `i16`, `i32`, `i64`, `i128`, `isize`, `char` and `bool`.
 
-Const parameters may only be be used as standalone arguments inside
-of [types] and [repeat expressions] but may be freely used elsewhere:
+Const parameters can generally be used anywhere a [const item] can be used,
+with the exception of the definition of any [item] within the body of a
```
Havvy, line 66: This exception applies to all generics. Can it be placed outside of the `const generics` section? Or we can point to [item declaration statements](https://doc.rust-lang.org/stable/reference/statements.html#item-declarations) as the source of truth. Moving it out of this section also lets us trim down this sentence:

> Const parameters can be used anywhere a const item can be used, with the exception that when used in a type or array repeat expression, it must be standalone (as described below). That is, they are allowed in the following places:

src/items/generics.md
```diff
@@ -43,18 +43,137 @@ struct Ref<'a, T> where T: 'a { r: &'a T }
...
+    // Similarly not allowed for array repeat expressions.
+    [1; {N + 1}]
+}
+```
+
+A const argument in a [path] specifies the const value to use for that item.
+The argument must be a [const expression] of the type ascribed to the const
+parameter. The const expression must be a [block expression][block]
+(surrounded with braces) unless it is a single path segment (an [IDENTIFIER])
+or a [literal] (with a possibly leading `-` token). This syntactic restriction
```
Havvy, line 144: This final statement should be in a note since it's non-normative.

src/items/generics.md
```diff
@@ -43,18 +43,137 @@ struct Ref<'a, T> where T: 'a { r: &'a T }
...
+    // Example usage of a const argument.
+    double::<9>();
+    double::<-123>();
+    double::<{7 + 8}>();
+    double::<SOME_CONST>();
+    double::<{ SOME_CONST + 5 }>();
+}
+```
+
+When there is ambiguity if a generic argument could be resolved as either a
```
Havvy, line 165: This case should talk about different namespaces for clarity, even if we haven't properly documented namespaces (and we can leave a note saying that we haven't yet done so as well).

The way to describe this without mentioning ambiguity would be to say that generic arguments check the type namespace before the value namespace, yes?

Havvy (changes requested):
For exhaustiveness checking, it's basically saying that even if there are individual impls for all values of a const generic, that does not mean there is a blanket impl over all values of that const generic. Unlike types and lifetimes, there's not a constrained listing of all possible values that Rust knows about, so it's probably worth mentioning.

src/items/generics.md
```diff
@@ -43,18 +43,137 @@ struct Ref<'a, T> where T: 'a { r: &'a T }
 struct InnerArray<T, const N: usize>([T; N]);
 ```
 
+The generic parameters are in scope within the item definition where they are
+declared.
+
+[References], [raw pointers], [arrays], [slices][arrays], [tuples], and
```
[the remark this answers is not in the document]
ehuss: Introduced in #258. If I had to guess, it is maybe trying to say is that the syntax for specifying arguments to those types does not use the same syntax that the other items (structs, fns, etc.) use. That is, instead of IDENT<…> they each have their own unique syntax for specifying arguments.

I agree, it seems a bit confusing. Any suggestions? Maybe @matthewjasper could clarify?

src/items/generics.md
ehuss: Hm, that might be tricky, since it would need to also specify which namespace each parameter goes in.  I'd kinda like to tackle that separately if it would be OK?

About a year ago I started on writing an outline of chapters for name resolution, but kinda got overwhelmed and stopped.  I would still like to move it forward.  I was thinking with a rough outline, a good start would just be introducing namespaces. It's hard to document individual pieces in isolation since it is intermingled with scopes and other topics, but I think it can be done incrementally with a plan.  I mainly want to avoid incrementally adding pieces without having a picture of how it should look in the long term, since I think it would otherwise require reorganizing things at each step which can be disruptive and more work.

ehuss:
> For exhaustiveness checking,

Sorry, I think my brain crossed its wires when I wrote that.  The sentences towards the end were intended for the question about `impl` constrained parameters (the [`report_unsued_parameters`](https://github.com/rust-lang/rust/blob/46c35c76fedb90b547201d111e6423bb1de78cd6/compiler/rustc_typeck/src/impl_wf_check.rs#L150-L189) check).

I took a stab at adding a small blurb about exhaustiveness checking, but I'm not comfortable with the wording. I suspect there are others who can word it more accurately.

src/items/generics.md
ehuss: I added a sentence at the top when talking about parameter scopes that links to _Item declaration statements_.

src/items/generics.md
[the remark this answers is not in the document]
Havvy: Yeah, I let my brain think about it overnight. I'm betting this statement was originally in `paths.html` where it makes sense and was moved here as part of a split out. We should probably bring most of `GenericArgument` into `generics.html` and then rename `GenericArguments` to `GenericPathArguments` or something. I'll open an issue for that one.

src/items/generics.md
Havvy: Sure. Opened #924 to discuss this.

Havvy (approved):
The blurb on exhaustive checking is a little clunky, but it gets the job done well enough. If somebody wants to word it better, they can always open a new PR.

ehuss:
@Havvy I'm not sure if you had a chance to edit this, but I went ahead and pushed a small clarification on when const generics must be used.

Havvy (approved):
I have a more comprehensive explanation of a constraining generic parameter, but it has nothing to do with const parameters specifically, so I'm making that its own PR. The changes to `generics.md` you have are better than what I have, though.

Can you fix the file conflicts and we can merge this?

ehuss:
> Can you fix the file conflicts and we can merge this?

Done, I somehow missed the notification.

PR #922, opened 2021-01-08
The rust compiler is permissive of f32 and f64 suffix

philberty:
During development of gccrs it was found in PR https://github.com/Rust-GCC/gccrs/pull/126 that the rust compiler is permissive of the float suffix on literals but the spec does not show this.

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -371,8 +371,8 @@ type of the literal. The integer suffix must be the name of one of the
 integral types: `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `i64`,
 `u128`, `i128`, `f32`,`f64`, `usize`, or `isize`.
 
-Note the suffix of `f32` and `f64` enforce that an integer literal can coerced
-into a float of repestive suffix type, which the lexer can be permissive of.
+Note the suffix of `f32` and `f64` ensures that the literal has a floating point type.
+Even when it would otherwise be lexed as an integer.
 
 The type of an _unsuffixed_ integer literal is determined by type inference:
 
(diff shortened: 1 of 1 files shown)


src/tokens.md
```diff
@@ -368,7 +369,10 @@ Like any literal, an integer literal may be followed (immediately,
 without any spaces) by an _integer suffix_, which forcibly sets the
 type of the literal. The integer suffix must be the name of one of the
 integral types: `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `i64`,
-`u128`, `i128`, `usize`, or `isize`.
+`u128`, `i128`, `f32`,`f64`, `usize`, or `isize`.
+
+Note the suffix of `f32` and `f64` enforce that an integer literal can coerced
+into a float of repestive suffix type, which the lexer can be permissive of.
```
bjorn3, line 375:
```suggestion
into a float of respective suffix type.
```

bjorn3, line 375:
```suggestion
Note the suffix of `f32` and `f64` ensures that the literal has a floating point type.
Even when it would otherwise be lexed as an integer.
```

ehuss:
Thanks for the PR!  I think this is already covered in the floating point grammar:

```
DEC_LITERAL (. DEC_LITERAL)? FLOAT_EXPONENT? FLOAT_SUFFIX
```

Is there any particular reason to mention this as part of the integer grammar?

Havvy:
The floating point grammar, as quoted by ehuss, is defined at https://doc.rust-lang.org/stable/reference/tokens.html#floating-point-literals

The changes in this PR are factually incorrect, but there is at least one change that can be made: Adding `5f32` (or any other whole number...) as an example in the examples list. If you want to do that, go for it.

philberty:
Apologies if this is not the correct place to add this. When testing gccrs vs rust it was noticed that f32 and f64 suffix was allowed on integers abut the integer token spec didn't mention thus our lexer enforced that f32 and f64 was an invalid suffix for this integer token.

Maybe just adding f32 and f64 is enough to the list of accepted suffix types is enough for this change.

ehuss:
`f32` and `f64` suffixes are already listed in the `FLOAT_LITERAL` production. `f32` and `f64` suffixed `DEC_LITERAL` values aren't integers.  The grammar needs to be considered as a whole to understand what kind of lookahead is necessary to parse the text to differentiate between different rules.

Havvy:
Since this PR is factually incorrect due to a misunderstanding of how to read the grammar, I am closing it.

Nonetheless, thanks for the attempt to help.

PR #923, opened 2021-01-08
Fix formatting in the tuple section

jyn514:
Before:

image

After:

image

Caught by cargo-deadlinks in https://github.com/deadlinks/cargo-deadlinks/issues/133#issuecomment-748399781.


ehuss (approved):
Thanks!

PR #930, opened 2021-01-12
Document extra behavior of #[no_mangle]

jonas-schievink:


ehuss (approved):
Thanks, I pushed a small change to linkify the `used` attribute.

PR #932, opened 2021-01-14
Fix plural and delete spurious words in comparison ops

ojeda:
Best seen with a word or rich diff.


Havvy (approved):
💟 Thanks!

Although this section in general could be rewritten to be more reference-like. The assumption for what comparison means is not a language level concern. And we should show the syntactic sugar explicitly like in the compound assignment expressions chapter. But none of that concerns this PR, and a lot of expressions on this page are like that.

Havvy (approved):

Issue #934, opened 2021-01-17
Type Layout -> Layout; Make values primary


Havvy:
Values have a defined layout, but types don't always have defined layout.

E.g. dynamically sized types have dynamic types and where padding is located depends on which enum discriminant is active.

As such, we should rename the type layout chapter to just layout and describe it as properties of values and then state that types may have defined properties if all values of the type have the same value for that property.

PR #935, opened 2021-01-17
Padding

### LINKED ISSUE #882: Define the properties of Padding
We don't have a good explanation of the properties of padding. Currently we talk about it in two places:

1. Type layout; It's pointed out that a type can have alignment padding that adds to the size.
2. Added in #852, it's pointed out that it can be set as uninitialized.

I think padding should probably be described in the type layout chapter.

My question though, other than being an unnameable psuedo-field that can theoretically have any value, does it have any other properties? E.g., can they be repurposed by enums to do niche optimizations (making it unsafe to set the value)? Other restrictions?

Havvy:
Fixes #882

There are two initial commits. The first one does some general cleanup while the second actually defines padding. When reviewing, review each commit in order for better understanding.

You can see that while I'm still trying to describe padding as a property of a type, I have to actually say it's a property of a value for padding from fields. Thus I filed #934. Since Size and Alignment also are similarly contorted right now, I don't think figuring what to do about this should block this PR.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -5,11 +5,26 @@ The *layout* of a type is the combination of its:
 * size
 * alignment
 * relative offsets of its fields or elements
+* relative offsets of padding
 * for enums: how the discriminant is laid out and interpreted
 
 > **Note**: Type layout is allowed to change with each compilation. Instead of
 > trying to document exactly how each type is laid out, we only document the
 > properties that are guaranteed.
+
+## Padding
+
+Types may contain *padding*: bytes that have no semantic meaning that may take
+on any bit pattern. When a type is [moved or copied], the value of padding bytes
+may be changed.
+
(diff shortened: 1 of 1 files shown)


src/type-layout.md
```diff
@@ -1,11 +1,30 @@
...
+> properties that are guaranteed.
+
+## Padding
+
+Types may contain *padding*: bytes that have no semantic meaning that may take
+on any bit pattern. When a type is [moved or copied], the value of padding bytes
+may be changed.
+
+If a field can be accessed, its padding is also the padding of the containing
+type.
```
ehuss, line 22: I'm having a little difficulty understanding this sentence.  Can you elaborate on what it is saying? I've always had the impression that padding is the space *between* fields, not a property of the field itself.  And I'm uncertain what "the padding of the containing type" is.

PR #936, opened 2021-01-17
Refactored build steps for better readability

saki-osive:
In this PR, I have refactored the build steps of the project in the _README.md_ file so they are better readable, easier to follow, and more detailed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 book
 stable-check/
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 book
 stable-check/
-.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 book
 stable-check/
+.idea
```
bjorn3, line 3: You can add this to `~/.config/git/ignore` to apply it for your whole system or `.git/info/exclude` to apply it to this repo without having to commit it.

saki-osive:
Thanks @bjorn3, please verify now.

Havvy (approved):
💟 Thanks!

I'm not sure telling people to download the zip file is the best idea, unless they don't want to contribute and just build it locally, but the rest is definitely an improvement.

PR #937, opened 2021-01-17
Start documenting name resolution.

### LINKED ISSUE #924: Document Namespaces
N.B.: This is a subpart of #568.

@ehuss: you state in #921 that they started an outline for this. Do you still have it, and if so, are you willing to show it?

If this is as daunting as you think it is, should we create a new branch to work on it?

ehuss:
This is a very small start to making progress on documenting name resolution (#568). This PR focuses on:

- Create the "names" chapter, with an introduction, and list of all things that introduce names.
- Document namespaces.
- Consolidate and extend preludes.
- Move the "Paths" chapter underneath the names chapter.
- Move the "Visibility" chapter underneath the names chapter.

Closes #924.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -11,7 +11,6 @@
     - [Comments](comments.md)
     - [Whitespace](whitespace.md)
     - [Tokens](tokens.md)
-    - [Paths](paths.md)
 
 - [Macros](macros.md)
     - [Macros By Example](macros-by-example.md)
@@ -37,7 +36,6 @@
     - [External blocks](items/external-blocks.md)
     - [Generic parameters](items/generics.md)
     - [Associated Items](items/associated-items.md)
-    - [Visibility and Privacy](visibility-and-privacy.md)
 
 - [Attributes](attributes.md)
     - [Testing](attributes/testing.md)
@@ -102,6 +100,14 @@
     - [Lifetime elision](lifetime-elision.md)
 
 - [Special types and traits](special-types-and-traits.md)
+
+- [Names](names.md)
+    - [Namespaces](names/namespaces.md)
+    - [Scopes](names/scopes.md)
+    - [Preludes](names/preludes.md)
+    - [Paths](paths.md)
+    - [Name resolution](names/name-resolution.md)
+    - [Visibility and privacy](visibility-and-privacy.md)
 
 - [Memory model](memory-model.md)
     - [Memory allocation and lifetime](memory-allocation-and-lifetime.md)
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -144,10 +144,13 @@ identifiers, and labels are used to refer to an entity.
 
 ### Namespace
 
-A [*namespace*] is a set of [names](#name) where each name unambiguously
-refers to a specific [entity](#entity). Namespaces are organized in a
-hierarchy, where each level of the hierarchy has its own collection of named
-entities.
+A *namespace* is a logical grouping of declared [names](#name) based on the
+kind of [entity](#entity) the name refers to. Namespaces allow the occurrence
+of a name in one namespace to not conflict with the same name in another
+namespace.
+
+Within a namespace, names are organized in a hierarchy, where each level of
+the hierarchy has its own collection of named entities.
 
 ### Nominal types
 
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -142,9 +142,10 @@ identifiers, and labels are used to refer to an entity.
 
 ### Namespace
 
-A [*namespace*] is a collection of uniquely named [entities](#entity).
-Namespaces can be organized in a hierarchy, where each level of the hierarchy
-has its own collection of named entities.
+A [*namespace*] is a set of [names](#name) where each name unambiguously
+refers to a specific [entity](#entity). Namespaces are organized in a
+hierarchy, where each level of the hierarchy has its own collection of named
+entities.
 
 ### Nominal types
 
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -63,8 +63,10 @@ A dynamically sized type (DST) is a type without a statically known size or alig
 
 ### Entity
 
-An [*entity*] is a [type], [item], [generic parameter], [variable binding],
-[loop label], [lifetime], [field], or [attribute].
+An [*entity*] is a language construct that can be referred to in some way
+within the source program, usually via a [path][paths]. Entities include
+[types], [items], [generic parameters], [variable bindings], [loop labels],
+[lifetimes], [fields], [attributes], and [lints].
 
 ### Expression
 
@@ -251,23 +253,24 @@ example of an uninhabited type is the [never type] `!`, or an enum with no varia
 
 [alignment]: type-layout.md#size-and-alignment
 [associated item]: #associated-item
-[attribute]: attributes.md
+[attributes]: attributes.md
 [*entity*]: names.md
 [enums]: items/enumerations.md
-[field]: expressions/field-expr.md
+[fields]: expressions/field-expr.md
 [free item]: #free-item
-[generic parameter]: items/generics.md
+[generic parameters]: items/generics.md
 [identifier]: identifiers.md
 [identifiers]: identifiers.md
 [implementation]: items/implementations.md
 [implementations]: items/implementations.md
 [inherent implementation]: items/implementations.md#inherent-implementations
 [item]: items.md
-[item]: items.md
+[items]: items.md
 [labels]: tokens.md#lifetimes-and-loop-labels
 [lifetime or loop label]: tokens.md#lifetimes-and-loop-labels
-[lifetime]: tokens.md#lifetimes-and-loop-labels
-[loop label]: tokens.md#lifetimes-and-loop-labels
+[lifetimes]: tokens.md#lifetimes-and-loop-labels
+[lints]: attributes/diagnostics.md#lint-check-attributes
+[loop labels]: tokens.md#lifetimes-and-loop-labels
 [method]: items/associated-items.md#methods
 [*Name resolution*]: names/name-resolution.md
 [*name*]: names.md
@@ -280,7 +283,7 @@ example of an uninhabited type is the [never type] `!`, or an enum with no varia
 [structs]: items/structs.md
 [trait objects]: types/trait-object.md
 [traits]: items/traits.md
-[type]: types.md
+[types]: types.md
 [undefined-behavior]: behavior-considered-undefined.md
 [unions]: items/unions.md
-[variable binding]: patterns.md
+[variable bindings]: patterns.md
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -61,6 +61,11 @@ through a mechanism called ‘trait objects’.
 
 A dynamically sized type (DST) is a type without a statically known size or alignment.
 
+### Entity
+
+An [*entity*] is a [type], [item], [generic parameter], [variable binding],
+[loop label], [lifetime], [field], or [attribute].
+
 ### Expression
 
 An expression is a combination of values, constants, variables, operators
@@ -123,6 +128,24 @@ This is not affected by applied type arguments. `struct Foo` is considered local
 `Vec<Foo>` is not. `LocalType<ForeignType>` is local. Type aliases do not
 affect locality.
 
+### Name
+
+A [*name*] is an [identifier] or [lifetime or loop label] that refers to an
+[entity](#entity). A *name binding* is when an entity declaration introduces
+an identifier or label associated with that entity. [Paths],
+identifiers, and labels are used to refer to an entity.
+
+### Name resolution
+
+[*Name resolution*] is the compile-time process of tying [paths],
+[identifiers], and [labels] to [entity](#entity) declarations.
+
+### Namespace
+
+A [*namespace*] is a collection of uniquely named [entities](#entity).
+Namespaces can be organized in a hierarchy, where each level of the hierarchy
+has its own collection of named entities.
+
 ### Nominal types
 
 Types that can be referred to by a path directly. Specifically [enums],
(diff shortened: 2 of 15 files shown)


ehuss:
I tried hard to make this accurate, but there are almost certainly some mistakes (or things people would present differently). Please be gentle. :smile:

I explicitly did not update the rest of the reference to specify how things are introduced into different namespaces. For example, the `struct` item page should indicate that it introduces the struct name into the type namespace, and the constructor in the value namespace (and constructors generally aren't documented at all). I plan to follow up with that separately, as I think it will touch a large number of places, and I'm uncertain exactly how I'm going to integrate it. Another big part of this is to document `use` and how it introduces things into various namespaces, which I think is a big topic on its own when considering globs and such.

Other uncertainties:
* I documented lifetimes and labels as separate namespaces, but I'm not sure if that's the best way to present it. I think it could consider the lifetimes as part of the type namespace, and labels as part of the value namespace (since they are syntactically incapable of clashing with anything else). Or, they could be considered part of a combined lifetime/label namespace (since [#21633](https://github.com/rust-lang/rust/issues/21633) intended to reject shadowing, but in practice they are still treated independently.

## Terminology

This introduces a new term "entity" to refer to types/items/values/macros/lifetimes/etc. I'd like to defend that choice, as I think it is useful to have a single term to refer to that soup of things. Also, I find that term is used in several other language specs and programming literature.

I waffled a bit on using "name" as an explicit term. There is a bit of a vague overlap between "name", "identifier", "label", and "path". I think "name" is very commonly used in other language specs and literature (and follows more naturally with "namespaces" and "name resolution", and is how rustc internally refers to them as "[name bindings](https://en.wikipedia.org/wiki/Name_binding)"). I'm not sure if how I presented it is entirely clear, though.

petrochenkov:
I'll look at this tomorrow.

Havvy (changes requested):
So, overall, I like what I'm seeing. This is a huge chunk of language that makes describing other parts of the language a lot more precise. Like, I could gleefully work on `expressions/path-expr.md` using this terminology.

---

Except, I don't like saying that the things that introduce names are entities themselves. I rather think that items introduce entities and names separately. The entity stands alone in some abstract entity space where names don't exist while the name is attached to a scope in a specified namespace that points to the element in entity space. I'm probably missing some subtlety in my definition even; this stuff is hard to think about. But by separating it out like this, it lets us:

* separate the lexical element (e.g. an item) from what the lexical element introduces
* say that a single lexical element introduces multiple entities and names
* say the crate root is an entity, and if you think it is not, one [can create an alias to it](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cee55c52f0bd5be9facfa25dcb418323)

---

I have some half-finished thoughts in my head that a name can also be attached to an entity. For example, `mod foo { struct Bar; }` would have `Bar` attached to `Foo`, and that this attachment is what lets `foo::Bar` work. Then the preludes can introduce the names into module scope but without attaching them. This also lets us attach associated items to the item they associate to.

---

I'm also not sure how non-static method names fit in all of this.

src/names/namespaces.md
```diff
@@ -0,0 +1,148 @@
+# Namespaces
+
+Entity names are segregated into separate *namespaces* based on the kind of
```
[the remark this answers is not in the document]
bjorn3: If two glob imports import the same name inside the value namespace, that it is still a namespace despite ambiguity.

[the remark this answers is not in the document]
ehuss: I'm not sure I follow the point about glob imports. A conflict with two glob imports like that is an error (that is, an invalid program). Other valid circumstances have rules that disambiguate what name is introduced. I wouldn't consider that to be ambiguous when there are rules on which name is used.

src/SUMMARY.md
```diff
@@ -103,6 +101,14 @@
 
 - [Special types and traits](special-types-and-traits.md)
 
+- [Names](names.md)
+    - [Namespaces](names/namespaces.md)
+    - [Scopes](names/scopes.md)
+    - [Preludes](names/preludes.md)
+    - [Paths](paths.md)
+    - [Name resolution](names/name-resolution.md)
```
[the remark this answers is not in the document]
ehuss: I think "name resolution" is a common term when discussing compilers. I've only ever seen "name lookup" used when discussing C++, since that's what the spec writers decided to use. I don't think it is super important, since both phrases will be understood, but I would lean towards the one that is used more often for clarity. Examples:

Python: https://docs.python.org/3/reference/executionmodel.html#resolution-of-names
Ada: http://www.ada-auth.org/standards/12rm/html/RM-4-1.html
PHP: https://www.php.net/manual/en/language.namespaces.rules.php
EcmaScript: https://262.ecma-international.org/9.0/index.html#sec-lexical-environment-and-environment-record-specification-types
PL/SQL: https://docs.oracle.com/cd/A87860_01/doc/appdev.817/a77069/d_names.htm
C++ spec uses "name resolution" in the context of template processing.
https://en.wikipedia.org/wiki/Name_resolution_(programming_languages)
A Theory of Name Resolution: https://web.cecs.pdx.edu/~apt/esop15.pdf
Declarative Name Binding and Scope Rules: https://gkonat.github.io/assets/publication/declarative_name_binding_and_scope_rules-sle12.pdf
Cooper & Torczon "Engineering a Compiler" 2nd ed p. 258

src/names.md
```diff
@@ -0,0 +1,133 @@
...
+[*Name resolution*] is the compile-time process of tying paths, identifiers,
+and labels to entity declarations.
+
+Access to certain names may be restricted based on their [*visibility*].
+
+## Explicit entities
+
+Entities that explicitly introduce a name in the source code are:
+
+* [Items]
```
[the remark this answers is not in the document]
ehuss: Yea, I think I had that originally, but removed it due to the risk that markdown would "eat" it.  For example, if you have a single word after the colon, like `* [Boolean type]: `bool``, it just disappears since it is interpreted as a link reference. It should be safe otherwise, though.

src/glossary.md
```diff
@@ -61,6 +61,11 @@ through a mechanism called ‘trait objects’.
 
 A dynamically sized type (DST) is a type without a statically known size or alignment.
 
+### Entity
+
+An [*entity*] is a [type], [item], [generic parameter], [variable binding],
```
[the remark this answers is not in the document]
ehuss: Hm, yea, it is difficult to define "entity" since it is a bit of a nebulous concept that is only defined by the list of things it encompasses. Some examples of how others handle it:

C++: "An entity is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field,template, template specialization, namespace, or parameter pack."
[Ada](http://www.ada-auth.org/standards/12rm/html/RM-3-1.html#I1290): Doesn't directly say what an "entity" is, other than "A declaration is a language construct that associates a name with (a view of) an entity."
[Java](https://docs.oracle.com/javase/specs/jls/se11/html/jls-6.html): Defines entity as the list (similar to how I did): "A declared entity (§6.1) is a package, class type (normal or enum), …". It tends to be use the term "declared entity" more formally.
ECMA 334 (C#): Doesn't seem to ever define it, but uses it all over such as "The scope of a name is the region of program text within which it is possible to refer to the entity declared by the name without qualification of the name."
[Go](https://golang.org/ref/spec): Doesn't define it, but uses it offhand like "While the identifier of the inner declaration is in scope, it denotes the entity declared by the inner declaration."
[Haskell 2010](https://www.haskell.org/definition/haskell2010.pdf): "We use the term entity to refer to a value, type, or class defined in, imported into, or perhaps exported from a module"
[Typescript](https://github.com/microsoft/TypeScript/blob/master/doc/spec-ARCHIVED.md): Uses the term "entity" all over without explicitly defining it.

I took another stab and making it more explicit, though I think defining at as a list of things is also valid.

ehuss:
> I'm also not sure how non-static method names fit in all of this.

I believe this is #23, correct? I think [method-call-expr](https://github.com/rust-lang/reference/blob/master/src/expressions/method-call-expr.md) doesn't do a horrible job of covering it.

petrochenkov:
>I documented lifetimes and labels as separate namespaces, but I'm not sure if that's the best way to present it.

+1
This is mostly a philosophical question due to the lexical separation of identifier and lifetime tokens, but it would be better to treat them separately.
Yes, lifetimes behave kind of like type parameters from type namespace (but also like const parameters from value namespace), and labels use same hygiene rules as local variables, but they still share zero common resolution code with regular types and values in the compiler, for example, can't form paths, can't be imported. I think that makes them sufficiently different.

src/SUMMARY.md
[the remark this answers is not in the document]
petrochenkov: "Resolution" also sounds like a more "involved" process than just looking the name up, that may include things like fixed-point iteration (which is needed to resolve imports).

src/glossary.md
```diff
@@ -123,6 +130,25 @@ This is not affected by applied type arguments. `struct Foo` is considered local
...
+
+[*Name resolution*] is the compile-time process of tying [paths],
+[identifiers], and [labels] to [entity](#entity) declarations.
+
+### Namespace
+
+A [*namespace*] is a set of [names](#name) where each name unambiguously
+refers to a specific [entity](#entity). Namespaces are organized in a
+hierarchy, where each level of the hierarchy has its own collection of named
+entities.
```
petrochenkov, line 150: Looks like this is what is called "scope" in the compiler (https://github.com/rust-lang/rust/blob/master/compiler/rustc_resolve/src/lib.rs#L102-L115)?
"Namespace" in the compiler is one of "type", "value" and "macro" (+ perhaps "lifetime" and "label", but they are separate).

petrochenkov: Ah, ok, it's a "scope", but the wording is confusing.

>set of [names](#name) where each name unambiguously
refers to a specific [entity](#entity)

This is plain indecipherable.

>Namespaces are organized in a hierarchy

And this should somehow highlight that the hierarchies exist *inside* the namespaces, and not between them.

src/names/preludes.md
```diff
@@ -0,0 +1,144 @@
+# Preludes
+
+A *prelude* is a collection of names that are automatically brought into scope
+of every module in a crate.
```
petrochenkov, line 4: AFAIK, originally "prelude" meant simply a piece of code inserted by the compiler at the start of the crate, like
```rust
#[prelude_import]
use std::prelude::v1::*;
#[macro_use]
extern crate std;
```
, not a set of names.

Then I used it as a temporary name / analogy for built-in types (["language prelude"](https://github.com/rust-lang/rust/pull/32131)) and extern crates (["extern prelude"](https://github.com/rust-lang/rust/pull/49789)), because they behaved similarly to the actual prelude import, and then it stuck.

I still have no idea whether the name "prelude" sounds reasonable in the meaning "set of names" to native speakers, or to users of [languages](https://hackage.haskell.org/package/base-4.14.1.0/docs/Prelude.html) from which this term originally came.

src/names/preludes.md
```diff
@@ -0,0 +1,144 @@
...
+namespace]. See the [tool attributes] section for more details.
+
+## The `no_implicit_prelude` attribute
+
+The *`no_implicit_prelude` [attribute]* may be applied at the crate level or
+on a module to indicate that it should not automatically bring the [standard
+library prelude], [extern prelude], or [tool prelude] into scope for that
+module or any of its descendants.
+
+This attribute does not affect the [language prelude] or [`macro_use` prelude].
```
petrochenkov, line 115: It affects `macro_use` prelude on >=2018 edition.
The plan was to exclude `macro_use` prelude on both editions (https://github.com/rust-lang/rust/pull/55630), but I never got to it.

ehuss: Oops, sorry I missed that. For various reasons, I do all my testing in 2015, and I didn't think to try 2018. I even saw that code, and somehow forgot to double-check it. I thought it was weird that it didn't affect `macro_use`.

src/names/preludes.md
ehuss: I'm personally comfortable with it. I think the behavior is similar enough ("importing these things into every module"). I think it will help later when talking about behaviors in how names are resolved.

src/glossary.md
ehuss: Yea, I'm struggling to define it clearly. Lexicography is not my strong suit. I took another stab at it.

I suspect this will receive updates and clarifications once I start working on the "scopes" chapter.

src/SUMMARY.md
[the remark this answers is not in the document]
Havvy: Alright. If everybody else also uses it, that's fine. I'm putting it on record that I still don't like the term.

src/names.md
```diff
@@ -0,0 +1,143 @@
+# Names
+
+An *entity* is a language construct that can be referred to in some way within
+the source program, usually via a [*path*]. Entities include [types], [items],
```
Havvy, line 4:
```suggestion
the source program, usually via a [path][*path*]. Entities include [types], [items],
```

src/names.md
```diff
@@ -0,0 +1,143 @@
...
+
+Names are segregated into different [*namespaces*], allowing entities in
+different namespaces to share the same name without conflict.
+
+[*Name resolution*] is the compile-time process of tying paths, identifiers,
+and labels to entity declarations.
+
+Access to certain names may be restricted based on their [*visibility*].
+
+## Explicit entities
```
Havvy, line 28: Can we change these section titles to "Explicitly declared entities" and "Implicitly declared entities"?

Havvy (changes requested):

Havvy (commented):
I see nothing further I'd like to see changed. This comment is to remove my "needs change" mark on the PR.

@petrochenkov can you approve and merge once you're satisfied yourself?

petrochenkov (approved):

petrochenkov:
@Havvy 
I have no power in this repo, I can neither assign myself nor approve the PR.

Havvy (approved):

src/names/preludes.md
jyn514: > I still have no idea whether the name "prelude" sounds reasonable in the meaning "set of names" to native speakers, or to users of languages from which this term originally came.

As a native speaker, I think of "prelude" as a jargon term the same way "macro" or "scope" is. Those words all have meanings outside programming, but the connection to the jargon term is tenuous at best. I think "prelude" is useful as an existing jargon term that's used in other languages, though.

PR #940, opened 2021-01-21
Referencify bool type

Havvy:
I never knew there was so much to say about booleans.

### DIFF
--- a/src/types/boolean.md
+++ b/src/types/boolean.md
@@ -32,6 +32,67 @@ Like all primitives, the boolean type [implements][p-impl] the
 
 > **Note**: See the [standard library docs][std] for library operations.
 
+## Operations on boolean values
+
+<!-- This is washy wording --> When using certain operator expressions with a
+boolean type for its operands, they evaluate using the rules of [boolean logic].
+
+### Logical not
+
+| `b` | [`!b`][op-not] |
+|- | - |
+| `true` | `false` |
+| `false` | `true` |
+
+### Logical or
+
+| `a` | `b` | [<code>a &#124; b</code>][op-or] |
+|- | - | - |
+| `true` | `true` | `true` |
+| `true` | `false` | `true` |
+| `false` | `true` | `true` |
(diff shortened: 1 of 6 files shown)


src/types/boolean.md
```diff
@@ -1,17 +1,52 @@
...
+The boolean type is the type of many operands in various [expressions]:
+
+* The condition operand in [if expressions] and [while expressions]
+* The operands in [lazy boolean operator expressions][lazy]
+
+> **Note**: The boolean type acts similarly to but is not an [enumerated type].
+In practice, this mostly means that constructors are not associated to the type
+(e.g. `bool::true`).
+
+Like all primitives, the boolean type [implements][p-impl] the
```
ehuss, line 29: Is there a reason this doesn't list other traits like the [operator traits](https://doc.rust-lang.org/nightly/reference/special-types-and-traits.html#operator-traits)?  It does seem like a little bit of a gray area of which trait impls are part of the "language", but it is not clear to me how that line is drawn.

Havvy: I just referenced the traits on the special traits page, and didn't think too much about the operator traits. It is an oversight, sorta. There's inherent behavior with the operators themselves, but the trait implementations are [library-side](https://doc.rust-lang.org/src/core/ops/bit.rs.html#167):

```rust
macro_rules! bitand_impl {
    ($($t:ty)*) => ($(
        #[stable(feature = "rust1", since = "1.0.0")]
        impl BitAnd for $t {
            type Output = $t;

            #[inline]
            fn bitand(self, rhs: $t) -> $t { self & rhs }
        }

        forward_ref_binop! { impl BitAnd, bitand for $t, $t }
    )*)
}

bitand_impl! { bool usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 }
```

Comparison operators are also built-in that way.

---

How about I add truth tables for all the things you can do with a boolean, and then link to those from the operator expressions?

ehuss (approved):
:+1:

PR #952, opened 2021-01-26
Remove "Memory Ownership" chapter

Havvy:
It's a holdover from pre-Rust 1.0 when the reference was the main
document to understand anything about the language. Its contents are
completely meaningless and to avoid having hopeful people step into this
useless page and have their hopes smashed, we are removing it.


ehuss (approved):
👍

PR #956, opened 2021-02-03
remove `>= | Ge` punctuation reference to Generics

kw-fn:


ehuss:
`Generics` is intentionally included here because the token can show up in generic parameters.  For example:

```
struct S<T: Trait<T>=Foo> {
    field: T
}
```

Here the `>=` token shows up as part of parsing the generic parameters (or, at least, must be handled in some way).

The reference doesn't (yet — #567) really go into details on token splitting. It's a bit tricky, since different implementations will take different approaches.  When making the table, I took the approach of including these strange scenarios where tokens may need to be split into smaller pieces.  (AndAnd and ShrEq are other examples.)

Perhaps these strange entries should be marked with a footnote that mentions this split token behavior?

PR #958, opened 2021-02-04
Standardize reference of "Attribute Macro"

ETCaton:
The Attributes section uses "macro attributes" but links to a section calling them "attribute macros"

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -25,14 +25,14 @@ the bang after the hash, apply to the thing that follows the attribute.
 
 The attribute consists of a path to the attribute, followed by an optional
 delimited token tree whose interpretation is defined by the attribute.
-Attributes other than macro attributes also allow the input to be an equals
+Attributes other than attribute macros also allow the input to be an equals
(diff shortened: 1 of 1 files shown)


JohnTitor:
I'm thinking of this but I'd prefer "macro attributes" if the wording hasn't been standardized yet. Because we have a similar wording "procedural macro attributes" and if we call the earlier "attribute macros", the word order isn't consistent. @ehuss thoughts?

petrochenkov:
I use both depending on context - "macro attributes and other attributes" or "attribute macros and other macros".

JohnTitor:
Hmm, that makes sense, it seems to make it clearer what we talk about now. So the question now is that whether we should standardize it or not. We should do if it makes some confusion, but it doesn't seem to be a serious problem right now.

traviscross:
We're going to close this in favor of:

- https://github.com/rust-lang/reference/pull/1660

...and due to the reason that @petrochenkov mentioned, which is that there are kind of two things here depending on context.

PR #959, opened 2021-02-04
Fix typo in type cast expression table

osa1:
Typo originally reported in the compiler issue tracker:

https://github.com/rust-lang/rust/issues/81686#issuecomment-772820785

It doesn't make sense to cast a pointer to a float, and it's currently
rejected by the compiler:

    error[E0606]: casting `*const T` as `f32` is invalid
     --> test.rs:2:5
      |
    2 |     a as f32
      |     ^^^^^^^^

I think what's meant here is "integer type" rather than "numeric type".


ehuss (approved):
Thanks!

PR #960, opened 2021-02-06
Document file size limitation

Smittyvb:
Source files must be less than 4 GiB. (the file size is stored as a `u32` in `rustc`)
```shell
$ truncate -s 4294967295 large.rs # create a sparse 4 GiB file
$ rustc large.rs # compile
fatal error: rustc does not support files larger than 4GB
```


tesuji:
Isn't that compiler implementation detail rather than a language spec need to be documented ?

ehuss:
Yea, this is an implementation limitation, and probably not something we want as part of the language. Thanks for opening the PR, though!

PR #961, opened 2021-02-10
Update type casts section

osa1:
- Add reference-to-pointer casts
- Update array-to-pointer casts for [rust#81479][1]

[1]: https://github.com/rust-lang/rust/pull/81479


Havvy:
Looks good to me. The only thing of note is that this would be first place using subscript notation. I'm fine with that, but cc @ehuss for their thoughts on that specifically.

Havvy:
@osa1 Can you rebase?

osa1:
@Havvy done.

Havvy (approved):

Havvy:
💟 Thanks!

Gonna merge this using the subscript notation. If it's decided we don't want to use that, we can change it in the future.

PR #963, opened 2021-02-14
Remove enum variant expr

Havvy:
This PR builds on top of #962, so that one needs to be approved and merged before this one.

This PR also gets rid the enum variant expression page. I think it was originally considered different when the expressions page was split into multiple chapters, and then quietly ignored. The struct expression page gives the same amount of information, plus more. It could give more (e.g. mentioning that it sets the discriminant), but I'll do that when I clean up the struct expression page itself.


ehuss:
Can you rebase so the diff is easier to read?

Havvy:
Rebased.

src/expressions/struct-expr.md
```diff
@@ -101,7 +101,8 @@ Point3d { x, y: y_value, z };
 ## Tuple struct expression
 
 A struct expression with fields enclosed in parentheses constructs a tuple struct.
-Though it is listed here as a specific expression for completeness, it is equivalent to a [call expression] to the tuple struct's constructor. For example:
+Though it is listed here as a specific expression for completeness, it is equivalent to a [call expression] to the tuple struct's constructor. For example: <!--
+This is false. Example: Position(0, ..other_position) -->
```
ehuss, line 105: I don't understand this comment. Can you show what it means? 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a295289c9b9e281e95c80e220b3fd3ad

ehuss (commented):
Can you add this to `book.toml`:

```toml
[output.html.redirect]
"/expressions/enum-variant-expr.html" = "struct-expr.html"
```

src/expressions/struct-expr.md
Havvy: Oh, I figured that would work. Guess it does not, and its falseness is false.

ehuss:
Also, should there be a mention somewhere that the enum variant doesn't support functional record update syntax?  Or did I miss that?

Havvy:
Added the redirect, removed the comment, and changed to enum variant.

Also added a thing about the syntax only working for struct type values. It's a bit wordy, and I'll do a cleanup pass on it later.

ehuss (approved):
Thanks!

PR #965, opened 2021-02-18
Clarify object safety rules for methods striked from the vtable

guswynn:
When trying to figure out how `async_trait` object safety worked and thinking about `Stream::next`'s object safety, I found it very hard to understand 

This clarifies that object-safety can still be possible even if a method is striked from the vtable, and that `self` and `where `Self: Sized` are the ONLY ways to do that.
This was only really explained clearly in the examples below, so this includes this clarification in the object safety rules in the beginning of the section

---
Separately, NonDispatchable seems like you are out of luck, but this pattern: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5529e6424895cccda19d4dd81f389b69, where we impl the trait for &T where T: Trait + ?Sized is actually quite common in the std (albeit with `self` parameter's so you don't need the `(&obj_ref)` dance to get method resolution to be happy, see https://rust-lang.zulipchat.com/#narrow/stream/249502-wg-async-foundations.2Fstream-trait-rfc/topic/Object.20safety/near/226749056 for more info). 

This seems like a pattern we should explain somewhere, but is the reference the right place for it? I have not mentioned it in this PR, but would be curious what people think

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -67,23 +67,23 @@ trait Seq<T> {
 Object safe traits can be the base trait of a [trait object]. A trait is
 *object safe* if it has the following qualities (defined in [RFC 255]):
 
-* It must not require `Self: Sized`
-* All associated functions must either have a `where Self: Sized` bound, or
-    * Not have any type parameters (although lifetime parameters are allowed),
-      and
-    * Be a [method] that does not use `Self` except in the type of the receiver.
+* All [supertraits] must also be object safe.
+* It must not require `Self: Sized` (i.e. `Sized` must not be a [supertrait][supertraits])
 * It must not have any associated constants.
-* All supertraits must also be object safe.
-
-When there isn't a `Self: Sized` bound on a method, the type of a method
-receiver must be one of the following types:
-
-* `&Self`
-* `&mut Self`
-* [`Box<Self>`]
-* [`Rc<Self>`]
-* [`Arc<Self>`]
-* [`Pin<P>`] where `P` is one of the types above
+* All associated functions must "dispatchable from a trait object" or "explicitly non-dispatchable from a trait object":
+    * Dispatchable functions require:
+        * Not have any type parameters (although lifetime parameters are allowed),
+        * Be a [method] that does not use `Self` except in the type of the receiver.
+        * Have a receiver with one of the following types:
+            * `&Self` (i.e. `&self`)
+            * `&mut Self` (i.e `&mut self`)
+            * [`Box<Self>`]
+            * [`Rc<Self>`]
+            * [`Arc<Self>`]
+            * [`Pin<P>`] where `P` is one of the types above
+        * Does not have a `where Self: Sized` bound (reciever type of `Self` (i.e. `self`) implies this).
+    * Explicitly non-dispatchable functions require:
+        * Have a `where Self: Sized` bound (reciever type of `Self` (i.e. `self`) implies this).
 
 ```rust
 # use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


src/items/traits.md
```diff
@@ -68,12 +68,13 @@ Object safe traits can be the base trait of a [trait object]. A trait is
 *object safe* if it has the following qualities (defined in [RFC 255]):
 
 * It must not require `Self: Sized`
-* All associated functions must either have a `where Self: Sized` bound, or
+* All associated functions must either have a `where Self: Sized` bound (reciever type of `self: Self`implies this), or
```
bjorn3, line 71:
```suggestion
* All associated functions must either have a `where Self: Sized` bound (reciever type of `self: Self` implies this), or
```

Havvy (commented):
I don't think these changes make it clear at all that the methods are not part of the vtable, mainly because it causes convoluted logic. Specifically, it has the structure  of:

```md
* All assoc fns must have X (Y implies X) OR
** Not have A AND
** Be M but don't S
** OTHERWISE Zeta
```

I had to spend five minutes just to figure out what was trying to be said. The whole list should be reorganized to say that the associated function is dispatchable (and the criteria for that) or non-dispatchable (and again, which criteria are for that). Do you want to do that?

---

Unrelated to this change in particular, but it should also probably say `Sized` is not a supertrait instead of saying `Self: Sized`.

guswynn:
@Havvy  I agree it's somewhat convoluted. Unfortunately, the reality is that object safety is somewhat convoluted. I do like your suggestion, though it would be a sign it departure from the original rfc structure, which I am not sure if it's a constraint people want to keep or not

Havvy:
Keeping the original RFC structure for documentation is definitely not a constraint. If it can be improved, it should.

guswynn:
@Havvy let me know what you think of this change!

src/items/traits.md
```diff
@@ -67,23 +67,23 @@ trait Seq<T> {
...
-When there isn't a `Self: Sized` bound on a method, the type of a method
-receiver must be one of the following types:
-
-* `&Self`
-* `&mut Self`
-* [`Box<Self>`]
-* [`Rc<Self>`]
-* [`Arc<Self>`]
-* [`Pin<P>`] where `P` is one of the types above
+* All associated functions must "dispatchable from a trait object" or "explicitly non-dispatchable from a trait object":
```
Havvy, line 73:
```suggestion
* All associated functions must either be dispatchable from a trait object or be explicitly non-dispatchable:
```

src/items/traits.md
```diff
@@ -67,23 +67,23 @@ trait Seq<T> {
 Object safe traits can be the base trait of a [trait object]. A trait is
 *object safe* if it has the following qualities (defined in [RFC 255]):
 
-* It must not require `Self: Sized`
-* All associated functions must either have a `where Self: Sized` bound, or
-    * Not have any type parameters (although lifetime parameters are allowed),
-      and
-    * Be a [method] that does not use `Self` except in the type of the receiver.
+* All [supertraits] must also be object safe.
+* It must not require `Self: Sized` (i.e. `Sized` must not be a [supertrait][supertraits])
```
Havvy, line 71:
```suggestion
* `Sized` must not be a [supertrait][supertraits]. In other words, it must not require `Self: Sized`.
```

Let's avoid the latinisms like "i.e."

Havvy (changes requested):
I'll approve it with my suggested edits.

The things in parenthesis should be in the `Note:` format, but I can do that later; especially since mixing it with a list is going to be interesting.

Havvy:
@guswynn Any status update on this?

guswynn:
@Havvy sorry about the delay!

Havvy:
💟 Thanks!

And no worries about the delay. I know I've spent more than enough time delaying on things I've been needing to do myself.

PR #967, opened 2021-02-20
Editorial cleanup on expressions (part 1)

Havvy:
* Use "the syntax of" uniformly.
* Use asterisks for all defined terms.
* Define more terms, espcially in the syntax section.
* * Reword things so that definitions are generally first.
* * I did not necessarily go with the best wording; the idea is to
   improve, not perfect. I still need to dedicate time to each
   expressions individually.
* Remove usage of "the compiler" and "Rust".
* * This also involved rewording.
* * How to deal with closure types and closure expressions is gonna be
   an interesting question to solve. I avoided solving it here.
* Remove non-normative information or put them in a `Note`.
* A few added section headers
* Move links to the bottom

Note that I've left quite a few nonsensical statements alone, such as
any that use "denote". They'll be treated separately.

About halfway through the list of expressions and this PR is getting
large. So I'm gonna cut this one here, stopping at grouped expressions
in the alphabetical list.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -55,6 +55,9 @@ In this way, the structure of expressions dictates the structure of execution.
 Blocks are just another kind of expression, so blocks, statements, expressions,
 and blocks again can recursively nest inside each other to an arbitrary depth.
 
+> **Note**: We give names to the operands of expressions so that we may discuss
+> them, but these names are not stable and may be changed.
+
 ## Expression precedence
 
 The precedence of Rust operators and expressions is ordered as follows, going
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -137,7 +137,7 @@ assert_eq!(
 ## Place Expressions and Value Expressions
 
 Expressions are divided into two main categories: place expressions and
-value expressions. Likewise within each expression, sub-expressions may occur
+value expressions. Likewise within each expression, operands may occur
 in either place context or value context. The evaluation of an expression
 depends both on its own category and the context it occurs within.
 
--- a/src/expressions/await-expr.md
+++ b/src/expressions/await-expr.md
@@ -5,12 +5,12 @@
 > &nbsp;&nbsp; [_Expression_] `.` `await`
 
 *Await expressions* suspend the current computation until the given future is ready to produce a value.
-The syntax for an await expression is an operand with a type that implements the Future trait, `.`, and then the `await` keyword.
+The syntax for an await expression is an expression with a type that implements the Future trait, called the *future operand*, then the token `.`, and then the `await` keyword.
 Await expressions are legal only within an [async context], like an [`async fn`] or an [`async` block].
 
-More specifically, an `<expr>.await` expression has the following effect.
+More specifically, an await expression has the following effect.
 
-1. Evaluate `<expr>` to a [future] `tmp`;
+1. Evaluate the future operand to a [future] `tmp`;
 2. Pin `tmp` using [`Pin::new_unchecked`];
 3. This pinned future is then polled by calling the [`Future::poll`] method and passing it the current [task context](#task-context);
 3. If the call to `poll` returns [`Poll::Pending`], then the future returns `Poll::Pending`, suspending its state so that, when the surrounding async context is re-polled,execution returns to step 2;
@@ -25,11 +25,11 @@ Because `await` expressions are only legal in an async context, there must be so
 
 ## Approximate desugaring
 
-Effectively, an `<expr>.await` expression is roughly equivalent to the following non-normative desugaring:
+Effectively, an await expression is roughly equivalent to the following non-normative desugaring:
 
 <!-- ignore: example expansion -->
 ```rust,ignore
-match /* <expr> */ {
+match future_operand {
     mut pinned => loop {
         let mut pin = unsafe { Pin::new_unchecked(&mut pinned) };
         match Pin::future::poll(Pin::borrow(&mut pin), &mut current_context) {
--- a/src/expressions/await-expr.md
+++ b/src/expressions/await-expr.md
@@ -4,9 +4,9 @@
 > _AwaitExpression_ :\
 > &nbsp;&nbsp; [_Expression_] `.` `await`
 
+*Await expressions* suspend the current computation until the given future is ready to produce a value.
+The syntax for an await expression is an operand with a type that implements the Future trait, `.`, and then the `await` keyword.
 Await expressions are legal only within an [async context], like an [`async fn`] or an [`async` block].
-They operate on a [future].
-Their effect is to suspend the current computation until the given future is ready to produce a value.
 
 More specifically, an `<expr>.await` expression has the following effect.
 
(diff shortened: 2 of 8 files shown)


src/expressions/field-expr.md
```diff
@@ -4,28 +4,48 @@
...
+> // Invalid: Parsed as calling the method "callable"
+> // holds_callable.callable();
+>
+> // Valid
+> (holds_callable.callable)();
+> ```
+
 A field expression denotes a field of a [struct] or [union].
+
 To call a function stored in a struct, parentheses are needed around the field expression.
```
ehuss, line 29: This seems to be redundant now?

src/expressions/field-expr.md
```diff
@@ -4,28 +4,48 @@
...
 (Struct {a: 10, b: 20}).a;
-mystruct.method();          // Method expression
 (mystruct.function_field)() // Call expression containing a field expression
 ```
 
-A field access is a [place expression] referring to the location of that field.
-When the subexpression is [mutable], the field expression is also mutable.
+## Automatic dereferencing
+
+Also, if the type of the operand implements [`Deref`] or [`DerefMut`][`Deref`] depending on whether the operand is [mutable], it is *automatically dereferenced* as many times as necessary to make the field access possible.
```
ehuss, line 41: Not sure what the "Also" is for.

```suggestion
If the type of the operand implements [`Deref`] or [`DerefMut`][`Deref`] depending on whether the operand is [mutable], it is *automatically dereferenced* as many times as necessary to make the field access possible.
```

Also, this removes the sentence about `In cases of ambiguity, we prefer fewer autoderefs to more.`.  Do you know what that was referring to?  What ambiguity is there?

src/expressions/field-expr.md
```diff
@@ -4,28 +4,48 @@
...
+This processes is also called *autoderef* for short.
 
-Also, if the type of the expression to the left of the dot is a pointer, it is automatically dereferenced as many times as necessary to make the field access possible.
-In cases of ambiguity, we prefer fewer autoderefs to more.
+## Borrowing
 
 Finally, the fields of a struct or a reference to a struct are treated as separate entities when borrowing.
```
ehuss, line 46:
```suggestion
The fields of a struct or a reference to a struct are treated as separate entities when borrowing.
```

src/expressions/grouped-expr.md
```diff
@@ -4,8 +4,12 @@
 > _GroupedExpression_ :\
 > &nbsp;&nbsp; `(` [_InnerAttribute_]<sup>\*</sup> [_Expression_] `)`
 
-An expression enclosed in parentheses evaluates to the result of the enclosed expression.
-Parentheses can be used to explicitly specify evaluation order within an expression.
+A *parenthesized expression* wrap a single expression, evaluating to that expression.
```
ehuss, line 7:
```suggestion
A *parenthesized expression* wraps a single expression, evaluating to that expression.
```

src/expressions/grouped-expr.md
```diff
@@ -4,8 +4,12 @@
...
+The syntax for a parenthesized expression is a `(`, then an operand, the *enclosed operand*, and then a `)`.
+
+An expression enclosed in parentheses evaluates to the value of the enclosed operand.
```
ehuss, line 10: This seems to repeat the first sentence, they both say that it evaluates to the enclosed expression.

src/expressions/field-expr.md
```diff
@@ -42,9 +62,14 @@ let d: String = x.f3;           // Move out of x.f3
 ```
 
 [_Expression_]: ../expressions.md
+[`Box`]: ../special-types-and-traits.html#boxt
+[`Deref`]: ../special-types-and-traits.html#deref-and-derefmut
```
ehuss, line 66:
```suggestion
[`Box`]: ../special-types-and-traits.md#boxt
[`Deref`]: ../special-types-and-traits.md#deref-and-derefmut
```

ehuss (commented):
I really like the initial sentences that explain what the expression does.

src/expressions/block-expr.md
```diff
@@ -16,10 +16,12 @@ A *block expression*, or *block*, is a control flow expression and anonymous nam
 As a control flow expression, a block sequentially executes its component non-item declaration statements and then its final optional expression.
 As an anonymous namespace scope, item declarations are only in scope inside the block itself and variables declared by `let` statements are in scope from the next statement until the end of the block.
 
-Blocks are written as `{`, then any [inner attributes], then [statements], then an optional expression, and finally a `}`.
-Statements are usually required to be followed by a semicolon, with two exceptions.
-Item declaration statements do not need to be followed by a semicolon.
-Expression statements usually require a following semicolon except if its outer expression is a flow control expression.
+The syntax for a block is `{`, then any [inner attributes], then [statements], then an optional operand, and finally a `}`.
```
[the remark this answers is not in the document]
Havvy: How about syntactically calling it an expression, but keeping the name operand? So, for example,

"then statements, then an optional expression, (called) the xyz operand, and finally a `}`."

Also, what do you think of adding the word "called" there, where we name the syntactic element?

Alternatively, we could keep operand there, and always link to its definition where we use it. Or add a tooltip for it, but that's probably not a good idea.

src/expressions/call-expr.md
```diff
@@ -23,13 +24,13 @@ let name: &'static str = (|| "Rust")();
 
 ## Disambiguating Function Calls
 
-Rust treats all function calls as sugar for a more explicit, [fully-qualified syntax].
+All function calls are sugar for a more explicit [fully-qualified syntax].
 Upon compilation, Rust will desugar all function calls into the explicit form.
 Rust may sometimes require you to qualify function calls with trait, depending on the ambiguity of a call in light of in-scope items.
```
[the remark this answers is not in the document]
Havvy: I think it's meant to be "with traits", but also that comma should not be there.

src/expressions/field-expr.md
Havvy: The ambiguity is if you can autoderef a type that has the field to another type that also has the field. The entire paragraph needs to be changed, but I'm trying to avoid making too many content changes in this PR, expecting other PRs for each expression to handle them.

src/expressions/call-expr.md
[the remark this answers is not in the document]
Havvy: The section is definitely underspecified. I am not trying to fix that in this PR.

src/expressions/field-expr.md
Havvy: Yes it is.

Havvy:
This is ready for re-review.

src/expressions/await-expr.md
```diff
@@ -4,45 +4,32 @@
 > _AwaitExpression_ :\
 > &nbsp;&nbsp; [_Expression_] `.` `await`
 
+*Await expressions* suspend the current computation until the given future is ready to produce a value.
+The syntax for an await expression is an expression with a type that implements the Future trait, called the *future operand*, then the token `.`, and then the `await` keyword.
```
ehuss, line 8:
```suggestion
The syntax for an await expression is an expression with a type that implements the [Future] trait, called the *future operand*, then the token `.`, and then the `await` keyword.
```

src/expressions.md
```diff
@@ -55,6 +55,9 @@ In this way, the structure of expressions dictates the structure of execution.
 Blocks are just another kind of expression, so blocks, statements, expressions,
 and blocks again can recursively nest inside each other to an arbitrary depth.
 
+> **Note**: We give names to the operands of expressions so that we may discuss
+> them, but these names are not stable and may be changed.
```
ehuss, line 59: I'm curious why this needs to be stated.  Are you concerned that other people will start using these names, and you don't want that?

ehuss (commented):
Sorry about slow reviews.  I recently had surgery, and have been a bit distracted.

src/expressions.md
Havvy: Mainly because I'm coming up with them off the spot without any rigorous discussion or approval from the lang team. I might try an RFC for making these terms part of the language, and dealing with the bikeshed that that entails. I also want to be able to freely change them if we do find better names.

Havvy:
Finally stopped feeling super lazy and added the newline. That addresses review comments and should be ready to merge.

ehuss (approved):
Nice!

I still think the note about operand naming isn't necessary, since the reference introduces all sorts of names that are just a way of putting labels on what is being discussed, but don't change the language in any way.  But I think it's fine to leave in, it can be removed later if desired.

Issue #950, opened 2021-01-25
Slowly convert reference to one sentence is one line


Havvy:
Instead of wrapping paragraphs at 80 characters, each sentence should be on one line.

There's not enough bandwidth to convert the entire reference at once, so PRs that touch a page substantially should switch the style. Do not allow mixed style on the same page; either do an entire page or none at all. Also, do so as a separate commit so it doesn't drown out other changes. Maybe even send in a PR that just does that and send in the substantial change based on the commit of that PR.

- [ ] Update the style guide.

PR #980, opened 2021-03-12
apply #950 to STYLE.md

ben0x539:
🌞

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -47,7 +47,7 @@ Rust examples are tested via rustdoc, and should include the appropriate annotat
 * `compile_fail` — The example is expected to fail to compile.
 * `ignore` — The example shouldn't be built or tested.
   This should be avoided if possible.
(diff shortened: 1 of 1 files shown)


ben0x539:
Sorry, I haven't really used github actions before and I don't know how to find out what's wrong.

ehuss:
The linkcheck script is a little out-of-sync with the latest nightly docs, it should get resolved after the next nightly is released (~7 hours from now).

Havvy:
Is there a way to re-run style checks?

ehuss:
@Havvy Not sure if you have "write" permissions, but if you do you can click on the "Details" link for the Actions run, and then in the upper-right corner is a drop-down to re-run an action.

I re-ran this, but it ran into another error.  It will need to be rebased on latest master to pass.

ehuss (approved):
I'm sorry for the delay, this got lost in my queue.   Thanks for the help!

PR #981, opened 2021-03-15
Document cfg(version)

est31:


Havvy (changes requested):
Can you pick a version number that has the feature implemented? Reading this, I'd think that I could use the feature in Rust 1.50.0.

Also, can you add a `Note` specifying when the minimum version that understands this notation? You can leave it as X.YY.0 for now until it's stabilized, but I assume with the greatest confidence that people will be looking for that specifically or will not know it doesn't work for e.g. Rust 1.0.0.

Also, are other forms allowed? E.g. is `version("false")` valid?

--

There's also some tone mismatches compared to the rest of the Reference, but it'd be easier to send a PR to your repo for those than to try to explain them.

--

Otherwise, looks good and thanks for writing this ahead of time!

est31:
> Also, are other forms allowed? E.g. is version("false") valid?

It does not follow the format so the cfg will evaluate to false (and a warning will be emitted, does the reference cover warnings?). It's "allowed" in the sense that it doesn't yield an *error*, to be forward compatible with future changes to rust's versioning without sacrificing compatibility with older compilers. `version(false)` on the other hand is not allowed, it does yield errors.

est31:
I've written the section to specifically cover the question of `version("false")`'s validity, but, while *containing* the info, apparently it didn't *convey* it, so improvements are needed. Suggestions welcome!

Havvy:
Perhaps explicitly stating "Strings that don't match this version format are allowed, and are always false."?

Issue #979, opened 2021-03-12
I wish it were clearer the "Type Cast Expressions" table is exhaustive


mcclure:
I am reading the Rust reference to learn Rust.
I reached [8.2.4, "Type cast expressions"](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#type-cast-expressions).

I read over the table. Immediately I found myself asking: What happens if you cast to `bool`? It's not listed.
I tested it, and immediately found that casting to bool is illegal.
Looking at the table after testing it, in retrospect this is "obvious". The reference says: "`as` can be used to explicitly perform coercions *[link to coercion section in reference]*, as well as the following additional casts". It does not say `as` can be used for anything else. Therefore, any cast not listed in either the coercion rules or the table must be illegal.
But, I was not able to figure that out just from reading the text! It is too easy to misinterpret "it *can* be used to do this..." as an example, rather than (as it is apparently intended) a normative definition.

**"Expected behavior"**: What would have helped me is somehow explicitly noting that "coercions, as well as the following additional casts" is exhaustive. So for example putting "(and no others)" after "following additional casts", or putting a sentence after the table saying "Any cast which is not either a defined coercion, or in the above table, is an error".

mcclure:
Addressed in #982

PR #982, opened 2021-03-16
Explicitly note any cast not defined in the documentation is illegal.

mcclure:
My attempt to address #979. Does this help?

This may cause a merge conflict with #961 .

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -305,7 +305,7 @@ fn average(values: &[f64]) -> f64 {
 }
 ```
 
-`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts.
+`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts. Any cast that does not fit either a coercion rule or an entry in the table is a compiler error.
 Here `*T` means either `*const T` or `*mut T`.
 
 | Type of `e`           | `U`                   | Cast performed by `e as U`       |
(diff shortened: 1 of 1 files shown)


Havvy:
#961 already has a merge conflict that needs to be resolved.

I think the sentence is good to add. I think it should be after line 308 instead of after the table. I'll defer to your judgement though if you disagree with that though.

Here's what it would look like:

```md
`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts.
Any cast that does not fit either a coercion rule or an entry in the table is a compiler error.
Here `*T` means either `*const T` or `*mut T`.
```

mcclure:
I think your way is clearer, I switched it.

src/expressions/operator-expr.md
```diff
@@ -305,7 +305,7 @@ fn average(values: &[f64]) -> f64 {
 }
 ```
 
-`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts.
+`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts. Any cast that does not fit either a coercion rule or an entry in the table is a compiler error.
```
Havvy, line 308:
```suggestion
`as` can be used to explicitly perform [coercions](../type-coercions.md), as well as the following additional casts.
Any cast that does not fit either a coercion rule or an entry in the table is a compiler error.
```

As per #950

Havvy (approved):

Havvy:
💟 Thanks

PR #984, opened 2021-03-18
Document current limitation of type aliases on enums

bstrie:
Question: is the intent of the reference to document how the language *currently* is, or how it *should* be? Because the behavior documented in this PR is something that I would like to be changed in the future (https://github.com/rust-lang/rust/issues/83248), but in the meantime it might be useful to readers to better understand the limitations of type aliases as they are currently implemented.

### DIFF
--- a/src/items/type-aliases.md
+++ b/src/items/type-aliases.md
@@ -36,7 +36,7 @@ A type alias without the [_Type_] specification may only appear as an
 
 A type alias to an enum cannot refer to the enum's variants within a [use declaration]:
 
-```rust,edition2018,compile_fail
+```rust,edition2018,compile_fail,E0432
 mod my_mod {
     pub enum MyEnum {
         MyVariant
--- a/src/items/type-aliases.md
+++ b/src/items/type-aliases.md
@@ -1,3 +1,4 @@
+
 # Type aliases
 
 > **<sup>Syntax</sup>**\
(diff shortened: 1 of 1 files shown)


src/items/type-aliases.md
```diff
@@ -33,10 +34,30 @@ let _ = TypeAlias(5); // Doesn't work
 A type alias without the [_Type_] specification may only appear as an
 [associated type] in a [trait].
 
+A type alias to an enum cannot refer to the enum's variants within a [use declaration]:
+
+```rust,edition2018,compile_fail
```
jyn514, line 39:
```suggestion
```rust,edition2018,compile_fail,E0432
```
https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#error-numbers-for-compile-fail-doctests

petrochenkov:
This is not a separate rule though, just an example of type-relative resolution not being supported in imports due to compiler staging.

bstrie:
@petrochenkov It would still be useful to mention that rule somewhere on this page. Is there a better way to word it, or some canonical part of the reference to link to?

Havvy:
I looked around in different places for the general rule, but I don't see it documented anywhere. I'm also not sure how to formulate the general rule, so I don't know where it would go either.

ehuss:
I think this boils down to documenting uniform paths (#487).  I thought there was something already in the reference about multi-segment paths and the requirements of the non-terminal segments, but I can't find anything.  

I think this would end up being part of (possibly significant) changes to the "Paths" chapter, the overhaul of the "Use declarations" chapter, and writing the "Resolution" chapter which could help clarify how paths are handled.

I have been slowly making progress in the background on name resolution, which ties into all of this.

bstrie:
Shall I close this in favor of your future additions, then?

Havvy:
FWIW, having it now would be better than waiting for the indeterminate future since it can always be removed later. But I don't know @ehuss's progress, so I'll `.await` their answer to the question.

bstrie:
I'll close this for now; I've left a comment on #487 to help remember to document this specific case in the future.

Issue #983, opened 2021-03-17
Please disambiguate phrase "these traits" in 8.2.13


mcclure:
I am reading the Rust reference to help me learn Rust.

I am looking at [8.2.13 "closure expressions"](https://doc.rust-lang.org/stable/reference/expressions/closure-expr.html) and I do not understand this sentence (bolded):

> The compiler will determine which of the closure traits the closure's type will implement by how it acts on its captured variables. The closure will also implement Send and/or Sync if all of its captured types do. **These traits allow functions to accept closures using generics, even though the exact types can't be named.**

The first appearance of "closure traits" is a link to 10.1.12. I am confused whether the phase "these traits" refers to:

- The closure traits defined in 10.1.12
- The traits Send and Sync, defined in the previous sentence.
- The informal English construction meaning "these characteristics…"

**"Expected behavior"**: You should replace "These traits" with some more specific wording. My proposed alternate text for the paragraph would be (based on my current best guess at the meaning):

> The compiler will determine which of the [closure traits](https://doc.rust-lang.org/stable/reference/types/closure.html#call-traits-and-coercions) the closure's type will implement by how it acts on its captured variables. The closure will also implement the closure traits [`Send`](https://doc.rust-lang.org/stable/reference/special-types-and-traits.html#send) and/or [`Sync`](https://doc.rust-lang.org/stable/reference/special-types-and-traits.html#sync) if all of its captured types do. Closure traits allow functions to accept closures using generics, even though the exact types can't be named.

I can create a PR for this change if someone can confirm that this is actually an accurate capture of the original paragraph's intent.

PR #991, opened 2021-03-25
Fix ambiguous phrasing in 8.2.13 in paragraph on closure traits

mcclure:
Addresses #983.

8.2.13 contains a paragraph of three sentences, in which the first sentence talks about the set of all closure traits, the second sentence talks about the specific closure traits `send` and `sync`, and the third sentence uses the phrase "these traits". In the third sentence it is unclear if the referent of "these traits" is closure traits (first sentence), the specific closure traits described in the second sentence, or nothing at all (IE it could mean "these characteristics" in the informal English sense).

This patch clarifies the text so that it is totally clear the two traits in sentence two are a subset of closure traits and totally clear that "these traits" in sentence three refers to "closure traits". I didn't get any answers in #983 but that is my best reasonable guess what the text means. Obviously only merge this if this is actually the text's intent…!

### DIFF
--- a/src/expressions/closure-expr.md
+++ b/src/expressions/closure-expr.md
@@ -31,8 +31,8 @@ A closure can be forced to capture its environment by copying or moving values b
 This is often used to ensure that the closure's type is `'static`.
 
 The compiler will determine which of the [closure traits](../types/closure.md#call-traits-and-coercions) the closure's type will implement by how it acts on its captured variables.
-The closure will also implement [`Send`](../special-types-and-traits.md#send) and/or [`Sync`](../special-types-and-traits.md#sync) if all of its captured types do.
-These traits allow functions to accept closures using generics, even though the exact types can't be named.
+The closure will also implement the closure traits [`Send`](../special-types-and-traits.md#send) and/or [`Sync`](../special-types-and-traits.md#sync) if all of its captured types do.
+Closure traits allow functions to accept closures using generics, even though the exact types can't be named.
 
 In this example, we define a function `ten_times` that takes a higher-order function argument, and we then call it with a closure expression as an argument, followed by a closure expression that moves values from its environment.
 


src/expressions/closure-expr.md
```diff
@@ -31,8 +31,8 @@ A closure can be forced to capture its environment by copying or moving values b
 This is often used to ensure that the closure's type is `'static`.
 
 The compiler will determine which of the [closure traits](../types/closure.md#call-traits-and-coercions) the closure's type will implement by how it acts on its captured variables.
-The closure will also implement [`Send`](../special-types-and-traits.md#send) and/or [`Sync`](../special-types-and-traits.md#sync) if all of its captured types do.
-These traits allow functions to accept closures using generics, even though the exact types can't be named.
+The closure will also implement the closure traits [`Send`](../special-types-and-traits.md#send) and/or [`Sync`](../special-types-and-traits.md#sync) if all of its captured types do.
+Closure traits allow functions to accept closures using generics, even though the exact types can't be named.
```
ehuss, line 35: I wouldn't add "the closure traits" for Send/Sync.  "Closure traits" specifically refers to `Fn`, `FnMut`, and `FnOnce`.  The other traits that closures may implement (send/sync/copy/clone) are just extras, and are not directly related to allowing use as function parameters.

I'm not sure why they are even called closure traits, since they are just the general call traits which are used for things other than closures. And I also question why it explicitly says "using generics", since closures can also be passed as trait objects.

Since the closure type chapter uses a clearer term ("call traits"), I would change it to use that phrasing, and change the last sentence to something like "The call traits allow functions to accept closures as parameters, even though…".

ehuss:
Oh, I forgot.  This will be rewritten as part of #967.  Perhaps you could take a look at that PR and see if it is clearer there?

ehuss:
I'm going to close this since this section has been rewritten, which I think has a bit clearer wording.

mcclure:
Sorry I failed to respond before, and thanks for clarifying the text

Issue #987, opened 2021-03-21
Typo in 10.2 "Dynamically Sized Types"


mcclure:
I am reading the Rust reference. Looking at [section 10.2 "Dynamically sized types"](https://doc.rust-lang.org/stable/reference/dynamically-sized-types.html) I find:

> DSTs can be provided as type arguments when a bound of `?Sized`.

This sentence does not verb. I don't understand what it is trying to say, so I cannot make a PR. Should the words "is provided" be at the end?

PR #992, opened 2021-03-25
Fix various grammatical errors around ?Sized

mcclure:
* Section 10.2 contains two sentences which are not sentences (I.E., they do not contain verbs). Patch makes the sentences grammatical and less ambugious, also links on the phrase ?Sized to the doc for ?Sized.
* Section 10.6 contains the text '? is only used to declare that the Sized trait may not be implemented for a type parameter or associated type'. 'May not' makes it sound as if it is prohibited to implement the Sized trait, whereas in fact the text means that the type may at its option fail to implement the sized trait. Patch clarifies.

I am not completely certain about this text. I think my text is *better* (because the existing text is not valid English) but my text is definitely more ungainly. (I would love to find an excuse to remove the words "a non-trait item" in the first bullet point.) I also do not know for a fact that my replacement text is *correct*— it is my best guess from looking at the other documentation (in particular [Sized](https://doc.rust-lang.org/std/marker/trait.Sized.html) in the stdlib). I am new at Rust and have yet to learn traits.

[EDIT: Removed note about ?sized link]

### DIFF
--- a/src/dynamically-sized-types.md
+++ b/src/dynamically-sized-types.md
@@ -11,7 +11,7 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
     * Pointers to slices also store the number of elements of the slice.
     * Pointers to trait objects also store a pointer to a vtable.
 * <abbr title="dynamically sized types">DSTs</abbr> can be provided as
-  type arguments when the type parameter has the special lifetime bound [`?Sized`][sized] specified. (In other words by default any type parameter on a non-trait item
+  type arguments when the type parameter has the special lifetime bound [`?Sized`](trait-bounds.html#sized) specified. (In other words by default any type parameter on a non-trait item
   has a `Sized` bound, and specifying the `?Sized` bound removes it.)
 * Traits may be implemented for <abbr title="dynamically sized
   types">DSTs</abbr>. Unlike type parameters on other kinds of items, `Self: ?Sized` is the default in trait
--- a/src/dynamically-sized-types.md
+++ b/src/dynamically-sized-types.md
@@ -11,10 +11,10 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
     * Pointers to slices also store the number of elements of the slice.
     * Pointers to trait objects also store a pointer to a vtable.
 * <abbr title="dynamically sized types">DSTs</abbr> can be provided as
-  type arguments when a bound of `?Sized`. By default any type parameter
-  has a `Sized` bound.
+  type arguments when the type parameter has the special lifetime bound [`?Sized`][sized] specified. (In other words by default any type parameter on a non-trait item
+  has a `Sized` bound, and specifying the `?Sized` bound removes it.)
 * Traits may be implemented for <abbr title="dynamically sized
-  types">DSTs</abbr>. Unlike type parameters `Self: ?Sized` by default in trait
+  types">DSTs</abbr>. Unlike type parameters on other kinds of items, `Self: ?Sized` is the default in trait
   definitions.
 * Structs may contain a <abbr title="dynamically sized type">DST</abbr> as the
   last field, this makes the struct itself a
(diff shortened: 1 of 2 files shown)


mcclure:
Addresses #987

mcclure:
A note: The best explanation of ?Sized I've found is attached to [Sized, in the stdlib](https://doc.rust-lang.org/std/marker/trait.Sized.html).

In the current version of this patch, when I mention ?Sized I link to "?Sized" in the reference.

This means the current status of Sized/?Sized documentation is:
- Stdlib— best documentation.
- Second-least-clear documentation, now that I've added additional verbosity in this patch: In the dynamically sized type section, 10.2.
- Third- and fourth- least clear documentation: "Sized" is defined in 11 "Special Types and Traits" and "?Sized" is defined in 10.6 "Trait and lifetime bounds". Neither 10.6 or 11 make it clear what the defaults are (IE: default ?sized for trait items, default sized for everything else) and the description of defaults in 10.6 is, to my read, actively misleading.

This all feels real backward. Getting the defaults wrong/misleading in the reference is especially concerning because this (defaults) is the exact material which is relevant for the reference rather than the stdlib. This PR currently constitutes a minimal change to make the text readable, but in my opinion a better way to structure this overall would be:
- Put Sized and ?Sized together in the reference— maybe in 10.6, maybe in 11, but in only one of those places and put them next to each other, because they're clearest when read in conjunction with each other. In this single place, explain ?Sized is the default for trait type parameters and Sized is the default for all other type parameters.
- Remove the verbosity from 10.2 and instead say something like:
    - DSTs can be provided as type arguments if the type parameter specifies `?Sized` (see [sized](link) and [?sized])
    - Traits may be implemented for DSTs unless the trait specifies `Sized` (see [sized](link) and [?sized]).
- Leave stdlib "Sized" doc unchanged, it's already ok

src/dynamically-sized-types.md
```diff
@@ -11,10 +11,10 @@ types">DSTs</abbr>. Such types can only be used in certain cases:
     * Pointers to slices also store the number of elements of the slice.
     * Pointers to trait objects also store a pointer to a vtable.
 * <abbr title="dynamically sized types">DSTs</abbr> can be provided as
-  type arguments when a bound of `?Sized`. By default any type parameter
-  has a `Sized` bound.
+  type arguments when the type parameter has the special lifetime bound [`?Sized`](trait-bounds.html#sized) specified. (In other words by default any type parameter on a non-trait item
```
Havvy, line 14:
```suggestion
  type arguments when the type parameter has the special trait bound [`?Sized`](trait-bounds.html#sized) specified. (In other words by default any type parameter on a non-trait item
```

Havvy, line 14: The parenthesis are not needed. The "In other words" already guards the sentence enough.

There's also needed commas after the "In other words" and "by default".

Also, as per the style guide, links in prose should just be the text and then have link refs at the bottom in alphabetical (well, UTF-8 codepoint) order.

Havvy (changes requested):
Review addresses the PR specifically.

I don't see any problem with the prose. I agree it is an improvement over what is currently there. I do have a few changes, but they are just one (small but crucial) factual inconsistency and three grammar/style things.

src/trait-bounds.md
```diff
@@ -77,8 +77,8 @@ Trait and lifetime bounds are also used to name [trait objects].
 
 ## `?Sized`
 
-`?` is only used to declare that the [`Sized`] trait may not be
-implemented for a type parameter or associated type. `?Sized` may
+`?` is only used to declare that the [`Sized`] trait may be
+unimplemented for a type parameter or associated type. `?Sized` may
```
tlyu, line 81: I think it might be better to say "is not required" instead of "may be unimplemented". Maybe add a parenthetical "(Type parameters and associated types have an implicit `Sized` bound.)" after this sentence.

PR #993, opened 2021-03-26
Link to ptr::addr_of on raw pointer docs

Havvy:
The second commit has the actual change. Just linking to the macros so that people can find them easier.

There are also a few switches from commas to periods in the first commit where they were actually different sentences.


ehuss (approved):
👍

PR #994, opened 2021-03-27
Correct casing type in notation.md

bjbagher:
Correct casing type from camelCase to PascalCase

### DIFF
--- a/src/notation.md
+++ b/src/notation.md
@@ -4,22 +4,22 @@
 
 The following notations are used by the *Lexer* and *Syntax* grammar snippets:
 
-| Notation          | Examples                      | Meaning                                   |
-|-------------------|-------------------------------|-------------------------------------------|
-| CAPITAL           | KW_IF, INTEGER_LITERAL        | A token produced by the lexer             |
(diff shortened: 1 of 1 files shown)


ehuss:
Hm, I'm uncertain that making this change is the right thing.  CamelCase can mean either upper or lowercase for the first character. I realize that is confusing, and usually it is better to be precise, but I think in this instance by example it is clear what is meant.  The Rust project as a whole uses "camel case" to mean uppercase, in various places ([compiler diagnostics](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#non-camel-case-types), [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html), and other documentation).  I'm reluctant to be inconsistent here, as that can sow further confusion (though I doubt too many people actually read this).

bjbagher:
Hey, no problem. I've never heard of CamelCase being either uppercase or lowercased. I come from the JS world, and camelCase is always lowercase for the first letter, otherwise, the ambiguity would bring confusion, hence PascalCase. I realize that this might not be the case here.

PR #995, opened 2021-03-30
Add const generic args to const contexts.

ehuss:
Const generic arguments are a "const context", so add them to the list.

Also, a const context has limited support for using const parameters.


Havvy (approved):

Havvy:
💟 Thanks!

PR #996, opened 2021-04-01
Fix typo in macros-by-example.md

Marcythm:
complete the missing parenthesis and move the parenthesis out of the code block into it.


Marcythm:
I'm not sure about the first modification, because it can't compile(But I think it's closer to what the author want to express).
Besides, I'm not a native English speaker, please excuse my typing errors.

Marcythm:
Well, it can compile now because I forgot that the outermost parentheses are parsed as delimiters...

ehuss (approved):
Thanks!

PR #997, opened 2021-04-02
Add rustdoc to tool lints.

ehuss:
`rustdoc` was added in https://github.com/rust-lang/rust/pull/80527.  Also, moved this to a callout since I consider it implementation-specific.


Havvy (approved):

Havvy:
This is stable, right?

ehuss:
Yes, I believe it is on the beta channel right now.

Havvy:
💟 Thanks!

PR #998, opened 2021-04-02
Update lint level documentation.

### LINKED ISSUE #574: Document lint control scoping.
The diagnostics chapter should explain how allow/deny/etc works when the same rule is applied at different levels.  For example:

```rust
#[allow(unused)]
{
    #[deny(unused)]
    {
        // Everything here is deny.
    }
}
```

This is pretty self-evident, but the docs should be explicit how this scoping works (it's fairly unusual attribute behavior).

Also, it should probably mention lint groups. For example:

```rust
#[allow(unused)]
{
    #[deny(unused_variables)]
    {
        // all unused lints are allow, except unused_variables.
    }
}
```

ehuss:
This adds some clarification about some details of lint level behavior. 

This should cover https://github.com/rust-lang/rust/pull/77534 in a roundabout fashion (by specifying the order that the attributes are processed).

This does not include a special callout for https://github.com/rust-lang/rust/issues/81670. I figure that is just a small bug in rustc, and this falls out of my personal threshold for being mentioned. This documents what was intended (that groups conceptually expand to the list of lints within them, and that the order of lints usually matters).

Some of this documentation could potentially be moved to the rustc book if it is decided that it is too implementation-specific. In particular, this describes the "warnings" group due to its unusual behavior. To me, this is a gray area between the language definition and what different implementations could do. I did feel like the command-line settings are implementation-specific, so just left that as a callout to the rustc book.

Closes #574

### DIFF
--- a/src/attributes/diagnostics.md
+++ b/src/attributes/diagnostics.md
@@ -39,6 +39,11 @@ pub mod m1 {
 }
 ```
 
+Lint attributes can override the level specified from a previous attribute, as
+long as the level does not attempt to change a forbidden lint. Previous
+attributes are those from a higher level in the syntax tree, or from a
+previous attribute on the same entity as listed in left-to-right source order.
+
 This example shows how one can use `allow` and `warn` to toggle a particular
 check on and off:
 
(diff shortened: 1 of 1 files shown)


src/attributes/diagnostics.md
```diff
@@ -39,6 +39,11 @@ pub mod m1 {
 }
 ```
 
+Lint attributes can override the level specified from a previous attribute, as
+long as the level does not attempt to change a forbidden lint. Previous
```
Havvy, line 43:
```suggestion
long as the level does not attempt to change a lint away from the forbidden level. Previous
```

"Forbidden lint" feels too ominous.

Havvy (approved):

Havvy:
I cannot merge since there's a merge conflict.

PR #1000, opened 2021-04-03
clarify UB for raw ptr deref

RalfJung:
Following [discussions on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/Deferencing.20Bad.20Value.20UB.20Reference.20Description)

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -23,8 +23,9 @@ code.
 </div>
 
 * Data races.
-* Evaluating a dereference [place expression] (`*expr`) on a raw pointer that is
-  [dangling] or unaligned.
+* Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
+  [dangling] or unaligned, even in [place expression context]
+  (e.g. `addr_of!(&*expr)`).
 * Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
   [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
 * Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
@@ -91,4 +92,5 @@ cannot be bigger than `isize::MAX` bytes.
 [Rustonomicon]: ../nomicon/index.html
 [`NonNull<T>`]: ../core/ptr/struct.NonNull.html
 [`NonZero*`]: ../core/num/index.html
-[place expression]: expressions.md#place-expressions-and-value-expressions
+[dereference expression]: expressions/operator-expr.md#the-dereference-operator
+[place expression context]: expressions.md#place-expressions-and-value-expressions
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -23,7 +23,8 @@ code.
 </div>
 
 * Data races.
-* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Evaluating a dereference [place expression] (`*expr`) on a raw pointer that is
+  [dangling] or unaligned.
 * Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
   [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
 * Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -23,7 +23,8 @@ code.
 </div>
 
 * Data races.
-* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Evaluating a dereference [place expression] (`*expr`) on a raw pointer that is
+  [dangling] or unaligned.
```
danielhenrymantilla, line 27: I think this was raised in the Zulip conversation, but I'll nevertheless raise it here: I wonder if a word other than e-_valu_-ating could be used when talking about a _place_ expression (although I couldn't come up with one satisfactory enough, so feel free to disregard this nit).

danielhenrymantilla (approved):
Mentioning / linking to an actual distinction between _value_ and _place_ expression seems totally reasonable when talking about pointer dereference (especially if the rules ended up loosened for to-place-dereferences), so 💯 from me! ✅

src/behavior-considered-undefined.md
RalfJung: Clearly the right term is "placating".

On a more serious note, this is one reason why I argued fiercly against calling these things "values", but I was unable to convince enough people to make a difference...

src/behavior-considered-undefined.md
```diff
@@ -62,6 +63,17 @@ a restricted set of valid values. In other words, the only cases in which
 reading uninitialized memory is permitted are inside `union`s and in "padding"
 (the gaps between the fields/elements of a type).
 
+> **Note**: Undefined behavior affects the entire program. For example, calling
+> a function in C that exhibits undefined behavior of C means your entire
+> program contains undefined behaviour that can also affect the Rust code. And
+> vice versa, undefined behavior in Rust can cause adverse affects on code
+> executed by any FFI calls to other languages.
+
+[place expression]: expressions.md#place-expressions-and-value-expressions
```
Havvy, line 72: Link refs should be located at the bottom of the page.

src/behavior-considered-undefined.md
```diff
@@ -23,7 +23,8 @@ code.
 </div>
 
 * Data races.
-* Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.
+* Evaluating a dereference [place expression] (`*expr`) on a raw pointer that is
```
Havvy, line 26: This feels weird to me, because a dereference is always a place expression.

RalfJung: Well, it can be used as a value expression as in
```rust
let x = *expr;
```
Arguably, here it is correct to say that `*expr` is (used as) a value expression.

Havvy: We call that a "value expression context" (and "place expression context" for where a place is wanted). Is it only UB to evaluate an unaligned/dangling pointer in place expression context, or always?

RalfJung: "value expression contexts" are a superset of "place expression contexts". IOW, the code above is sugar for
```rust
let x = place2value(*expr);
```
Evaluating a place expression in value expression context consists of first evaluating the place expression as normal, as then performing place-to-value conversion.

So, it is impossible to write `*expr` anywhere without it being *also* a place expression, but sometimes, it is both a place expression and a value expression (or really, `place2value(*expr)` is the value expression, but since we have no syntax for this, people tend to say that `*expr` is a value expression).

So, the answer is that it is always UB to evaluate an unaligned/dangling `*expr` since doing so always evaluates the place expression -- and then sometimes goes on performing place-to-value conversion. The important point that I hope to clarify in the docs is that it is the place expression evaluation, and *not* the place-to-value conversion, that is causing the UB.

That's the way I am thinking about this, anyway.

Havvy: > "value expression contexts" are a superset of "place expression contexts"

We have a difference of perspectives here; these two categories are distinct and split the set of expression contexts. A place expression in value expression context is allowed and has the `place2value` effect, but it's still always a place expression.

So, under my perspective, the `[place expression]` should be `, even in [place expression contexts],`.

---

> Evaluating a place expression in value expression context consists of first evaluating the place expression as normal, as then performing place-to-value conversion.

I like this sentence. I think I'll try and get it into the PR I submitted last night (#1003 )

RalfJung: So, more like this?

@digama0 I'd be interested in your opinion here, is this current wording sufficiently clear? (I know you'd like the rules to be more relaxed; so do I. That is a longer process. For now the goal is to make sure that the rules as they currently are are described unambiguously.)

Havvy (approved):

src/behavior-considered-undefined.md
digama0: LGTM

Havvy:
💟 Thanks!

PR #1001, opened 2021-04-03
document raw-addr-of operators

RalfJung:
as requested by @Havvy 

This is a bit unusual for being exposed as a library feature, fut it really is a fundamental language feature without stable syntax, which is stably accessible through the standard library.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -74,7 +74,7 @@ let a = & & & & mut 10;
 ### Raw address-of operators
 
 Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
-Like with `&`/`&mut`, the expression is evaluated in place expression context.
+Like with `&`/`&mut`, the expression `expr` is evaluated in place expression context.
 The difference is that `&`/`&mut` create *references* of type `&T`/`&mut T`, while `ptr::addr_of!(expr)` creates a (const) raw pointer of type `*const T` and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
 
 The raw address-of operators must be used whenever the place expression could evaluate to a place that is not properly aligned or does not store a valid value as determined by its type.
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -73,9 +73,9 @@ let a = & & & & mut 10;
 
 ### Raw address-of operators
 
-Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
+Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros [`ptr::addr_of!(expr)`][addr_of] and [`ptr::addr_of_mut!(expr)`][addr_of_mut].
 The expression `expr` is evaluated in place expression context.
-`ptr::addr_of!(expr)` then creates a (const) raw pointer of type `*const T` to the given place, and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
+`ptr::addr_of!(expr)` then creates a const raw pointer of type `*const T` to the given place, and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
 
 The raw address-of operators must be used instead of a borrow operator whenever the place expression could evaluate to a place that is not properly aligned or does not store a valid value as determined by its type, or whenever creating a reference would introduce incorrect aliasing assumptions.
 In those situations, using a borrow operator would cause [undefined behavior] by creating an invalid reference, but a raw pointer may still be constructed using an address-of operator.
@@ -640,6 +640,8 @@ See [this test] for an example of using this dependency.
 [Function pointer]: ../types/function-pointer.md
 [Function item]: ../types/function-item.md
 [undefined behavior]: ../behavior-considered-undefined.md
+[addr_of]: ../../std/ptr/macro.addr_of.html
+[addr_of_mut]: ../../std/ptr/macro.addr_of_mut.html
 
 [_BorrowExpression_]: #borrow-operators
 [_DereferenceExpression_]: #the-dereference-operator
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -73,14 +73,14 @@ let a = & & & & mut 10;
 
 ### Raw address-of operators
 
-Related to the borrow operators are the raw address-of operators, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
+Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
 Like with `&`/`&mut`, the expression is evaluated in place expression context.
 The difference is that `&`/`&mut` create *references* of type `&T`/`&mut T`, while `ptr::addr_of!(expr)` creates a (const) raw pointer of type `*const T` and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
 
-The raw address-of operators must be used whenever the place expression denotes a place that is not properly aligned or does not store a valid value as determined by its type.
+The raw address-of operators must be used whenever the place expression could evaluate to a place that is not properly aligned or does not store a valid value as determined by its type.
 In those situations, using a borrow operator would cause [undefined behavior] by creating an invalid reference, but a raw pointer may still be constructed using an address-of operator.
 
-**Example of creating a raw pointer to an unaligned place (through a `packed` struct):**
+The following is an example of creating a raw pointer to an unaligned place through a `packed` struct:
 
 ```rust
 use std::ptr;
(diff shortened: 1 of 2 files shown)


Havvy (changes requested):
Just woke up and saw this, and it was a great way to start my day. 😍 

I have to go to work, so no time for a full review, but I will at least ask that you switch the style to one line/one sentence to match the rest of the page.

RalfJung:
> I will at least ask that you switch the style to one line/one sentence to match the rest of the page.

Done.

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
+### Raw address-of operators
+
+Related to the borrow operators are the raw address-of operators, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
+Like with `&`/`&mut`, the expression is evaluated in place expression context.
+The difference is that `&`/`&mut` create *references* of type `&T`/`&mut T`, while `ptr::addr_of!(expr)` creates a (const) raw pointer of type `*const T` and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
+
+The raw address-of operators must be used whenever the place expression denotes a place that is not properly aligned or does not store a valid value as determined by its type.
```
Havvy, line 79:
```suggestion
The raw address-of operators must be used whenever the place expression denotes a place that is not properly aligned or does not store a valid value as determined by its type.
```
```suggestion
The raw address-of operators must be used whenever the place expression could evaluate to a place that is not properly aligned or does not store a valid value as determined by its type.
```

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
+### Raw address-of operators
+
+Related to the borrow operators are the raw address-of operators, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
+Like with `&`/`&mut`, the expression is evaluated in place expression context.
```
Havvy, line 76: "the expression" here refers to the expression passed to the macro, right? If so, can you name that expression something; perhaps "the referenced operand". Also if so, it needs to be added to the list of place expression contexts on `expressions.md`.

Also, the compare/contrast with borrowing expressions feels weird in the context of the reference. It would make sense in material teaching for the first time, but this isn't a first time reader's document.

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
...
+The difference is that `&`/`&mut` create *references* of type `&T`/`&mut T`, while `ptr::addr_of!(expr)` creates a (const) raw pointer of type `*const T` and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
+
+The raw address-of operators must be used whenever the place expression denotes a place that is not properly aligned or does not store a valid value as determined by its type.
+In those situations, using a borrow operator would cause [undefined behavior] by creating an invalid reference, but a raw pointer may still be constructed using an address-of operator.
+
+**Example of creating a raw pointer to an unaligned place (through a `packed` struct):**
```
Havvy, line 82:
```suggestion
The following is an example of creating a raw pointer to an unaligned place through a `packed` struct:
```

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
...
+    f2: u16,
+}
+
+let packed = Packed { f1: 1, f2: 2 };
+// `&packed.f2` would create an unaligned reference, and thus be Undefined Behavior!
+let raw_f2 = ptr::addr_of!(packed.f2);
+assert_eq!(unsafe { raw_f2.read_unaligned() }, 2);
+```
+
+**Example of creating a raw pointer to a place that does not contain a valid value:**
```
Havvy, line 99:
```suggestion
The following is an example of creating a raw pointer to a place that does not contain a valid value:
```

Havvy (changes requested):

src/expressions/operator-expr.md
RalfJung: > "the expression" here refers to the expression passed to the macro, right? If so, can you name that expression something; perhaps "the referenced operand". Also if so, it needs to be added to the list of place expression contexts on expressions.md.

I did some editing here, please check.

> Also, the compare/contrast with borrowing expressions feels weird in the context of the reference. It would make sense in material teaching for the first time, but this isn't a first time reader's document.

I feel like this also makes sense just to increase clarity of the document, no matter how much experience we assume the reader to have. Also note that currently, `### Raw address-of operators` is a *third*-level header, so this is basically "part of" the borrow operator section. The operators are very similar to I felt it would make sense to treat them together, at least to some extend.

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
+### Raw address-of operators
+
+Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
+Like with `&`/`&mut`, the expression `expr` is evaluated in place expression context.
```
Havvy, line 76: The first and third sentences together seem to make up the main definition and then this second sentence is just awkwardly in the middle. Perhaps the second and third sentences should be swapped.

Havvy, line 76: It is more clear calling it `expr`, but if you don't name the operand, I will have to when I go through this chapter unifying the style. Your choice if you want to leave it like this or change it.

Havvy (changes requested):
Overall I am happy with the changes. I just see one thing that really stands out to me.

src/expressions/operator-expr.md
RalfJung: I don't know what you mean -- the operand, `expr`, is named, isn't it?

RalfJung: I don't know what you mean -- the operand, `expr`, is named, isn't it?

RalfJung: I made the initial definition self-contained and moved the references to the borrow operator to the second paragraph.

RalfJung:
@Havvy I think I did all the changes you were asking for.

RalfJung:
This PR has been waiting for the next round of review for a while (and the feature it documents is now stable since many releases). How can we make progress?

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
...
+
+The raw address-of operators must be used instead of a borrow operator whenever the place expression could evaluate to a place that is not properly aligned or does not store a valid value as determined by its type, or whenever creating a reference would introduce incorrect aliasing assumptions.
+In those situations, using a borrow operator would cause [undefined behavior] by creating an invalid reference, but a raw pointer may still be constructed using an address-of operator.
+
+The following is an example of creating a raw pointer to an unaligned place through a `packed` struct:
+
+```rust
+use std::ptr;
+
+#[repr(packed)]
```
ehuss, line 87: Does this maybe need to be a well-defined layout like C?  I would assume that the Rust layout could theoretically rearrange the order?

```suggestion
#[repr(C, packed)]
```

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
+### Raw address-of operators
+
+Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros `ptr::addr_of!(expr)` and `ptr::addr_of_mut!(expr)`.
```
ehuss, line 75: It can be helpful to point to the standard library documentation to give more context.

```suggestion
Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros [`ptr::addr_of!(expr)`][addr_of] and [`ptr::addr_of_mut!(expr)`][addr_of_mut].
```

And at the bottom:

```
[addr_of]: ../../std/ptr/macro.addr_of.html
[addr_of_mut]: ../../std/ptr/macro.addr_of_mut.html
```

src/expressions/operator-expr.md
```diff
@@ -70,6 +70,50 @@ let a = && && mut 10;
...
+The expression `expr` is evaluated in place expression context.
+`ptr::addr_of!(expr)` then creates a (const) raw pointer of type `*const T` to the given place, and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
```
ehuss, line 77: The parenthetical here seems to stick out to me here, and I'm not sure what it is intending to convey.  

```suggestion
`ptr::addr_of!(expr)` then creates a const raw pointer of type `*const T` to the given place, and `ptr::addr_of_mut!(expr)` creates a mutable raw pointer of type `*mut T`.
```

ehuss:
Can you resolve the conflict?

> This PR has been waiting for the next round of review for a while (and the feature it documents is now stable since many releases). How can we make progress?

I don't feel super comfortable hitting the merge button as I am not very familiar with this.  I'd feel better if someone on the lang team approved it.  Perhaps @joshtriplett?

src/expressions/operator-expr.md
RalfJung: > I would assume that the Rust layout could theoretically rearrange the order?

It could, but nothing in the example relies on field order. So this is correct without `C`, I think.
We could still add it though, if you want?

ehuss: Ah, makes sense. Leaving as-is seems fine to me.

RalfJung:
@rust-lang/lang it's now been a year since the macros have been stabilized, might be good to also get the reference docs landed at some point. :D

joshtriplett:
I believe all feedback has been addressed at this point. Merging.

PR #1003, opened 2021-04-06
Expr

### LINKED ISSUE #972: Rename "Value expressions and place expressions" section to "Evaluation categories"
C++ calls them [value categories](https://docs.microsoft.com/en-us/cpp/cpp/lvalues-and-rvalues-visual-cpp?view=msvc-160). Since we call one "value expression", I'd like to avoid that term and instead call them a category of some kind. Since the distinction is about what kind of thing the expression evaluates to, I figured "evaluation category" would be the best term.

Having this term would also let me say parenthesized expressions are of the same evaluation category as their enclosed operand.

Havvy:
Fixes #972 

There's more that I want to do, but I think it requires creating a brand new chapter for places. Main information on temporaries don't really belong in the expressions chapter, nor does the definition of moving and copying.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -57,8 +57,14 @@ Blocks are just another kind of expression, so blocks, statements, expressions, 
 
 ## Expression precedence
 
-The precedence of Rust operators and expressions is ordered as follows, going from strong to weak.
+The grammar for expressions as presented in this book is wrong.
+It purposefully ignores that certain operands cannot be certain expressions due to precedence.
+For example, `a * b + c` is actually parsed as an addition operator expression with the multiplication operator expression as its left operand.
+The grammar, taking precedence into account, is too unweildy to read and understand.
+
+The precedence of operators and expressions is ordered as follows, going from strongest to weakest.
 Binary Operators at the same precedence level are grouped in the order given by their associativity.
+Expressions in this table cannot have operands that are expressions lower in the table.
 
 | Operator/Expression         | Associativity       |
 |-----------------------------|---------------------|
@@ -81,6 +87,8 @@ Binary Operators at the same precedence level are grouped in the order given by 
 | `..` `..=`                  | Require parentheses |
 | `=` `+=` `-=` `*=` `/=` `%=` <br> `&=` <code>&#124;=</code> `^=` `<<=` `>>=` | right to left |
 | `return` `break` closures   |                     |
+
+> **Note**: Wrapping an expression with a [parenthetical expression] increases its precedence.
 
 ## Evaluation order of operands
 
@@ -250,6 +258,7 @@ They are never allowed before:
 [`if let`]:             expressions/if-expr.md#if-let-expressions
 [match]:                expressions/match-expr.md
 [method-call]:          expressions/method-call-expr.md
+[parenthetical expression]: expressions/grouped-expr.md
 [paths]:                expressions/path-expr.md
 [struct]:               expressions/struct-expr.md
 [tuple expressions]:    expressions/tuple-expr.md
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -135,28 +135,41 @@ assert_eq!(
 
 > **Note**: Since this is applied recursively, these expressions are also evaluated from innermost to outermost, ignoring siblings until there are no inner subexpressions.
 
-## Place Expressions and Value Expressions
+## Expression categories
 
-Expressions are divided into two main categories: place expressions and value expressions.
-Likewise within each expression, operands may occur in either place context or value context.
-The evaluation of an expression depends both on its own category and the context it occurs within.
+<!-- This section was renamed, but heavily linked to. So adding this span to keep the links working. -->
+<span id="place-expressions-and-value-expressions"></span>
 
-A *place expression* is an expression that represents a memory location.
-These expressions are [paths] which refer to local variables, [static variables], [dereferences][deref] (`*expr`), [array indexing] expressions (`expr[expr]`), [field] references (`expr.f`) and parenthesized place expressions.
-All other expressions are value expressions.
+Expressions and expression contexts are divided into two *expression categories*: *place* and *value*.
+For expressions, what they evaluate into determines their categories.
+Expressions that evaluate to a value are *value expressions* while those that evaluate to a place are *place expressions*.
+For expression contexts, whether they operate on a place or value determines their context.
 
-A *value expression* is an expression that represents an actual value.
+Evaluating an expression in the opposite expression context has effects after evaluation of the expression to produce the expected place or value of the context.
+When evaluating a value expression in place expression context, a temporary place is initialized to that value.
+When evaluating a place expression in value expression context, the value at that place is read, moving or copying it.
 
-The following contexts are *place expression* contexts:
+The following lists all place expressions.
+Expressions not listed here are value expressions.
 
-* The left operand of an [assignment][assign] or [compound assignment] expression.
+* [Paths] which refer to local variables or [static variables].
+* [The dereference operator][deref]
+* [Field access expressions][field]
+* [Tuple indexing expressions]
+* [Array indexing expressions]
+* [Parenthetical expressions] when its enclosed operand is a place expression
+
+The following list all place expression contexts.
+All other contexts are value expression contexts.
+
+* The initializer of a [let statement].
+* The assigned place operand of an [assignment][assign] or [compound assignment] expression.
 * The operand of a unary [borrow] or [dereference][deref] operator.
-* The operand of a field expression.
+* The container operand of a field expression.
 * The indexed operand of an array indexing expression.
 * The operand of any [implicit borrow].
-* The initializer of a [let statement].
-* The [scrutinee] of an [`if let`], [`match`][match], or [`while let`] expression.
-* The base of a [functional update] struct expression.
+* The [scrutinee] operand of an [`if let`], [`match`][match], or [`while let`] expression.
+* The base operand of a [functional update] struct expression.
 
 > Note: Historically, place expressions were called *lvalues* and value expressions were called *rvalues*.
 
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -41,14 +41,17 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_MatchExpression_]\
 > &nbsp;&nbsp; )
 
-An expression may have two roles: it always produces a *value*, and it may have *effects* (otherwise known as "side effects").
-An expression *evaluates to* a value, and has effects during *evaluation*.
-Many expressions contain sub-expressions, called the *operands* of the expression.
+Expressions are the fundamental unit of computation.
+When evaluated, they may have *effects* and evaluates to either a value or place.
+
+Where expressions are allowed are *expression contexts*.
+Expressions commonly contain sub-expressions, called the *operands* of the expression.
+
 The meaning of each kind of expression dictates several things:
 
 * Whether or not to evaluate the operands when evaluating the expression
 * The order in which to evaluate the operands
-* How to combine the operands' values to obtain the value of the expression
+* How to combine the operands' values to obtain the value or place of the expression
 
 In this way, the structure of expressions dictates the structure of execution.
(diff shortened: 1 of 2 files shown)


src/expressions.md
```diff
@@ -41,28 +41,33 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_MatchExpression_]\
 > &nbsp;&nbsp; )
 
-An expression may have two roles: it always produces a *value*, and it may have
-*effects* (otherwise known as "side effects"). An expression *evaluates to* a
-value, and has effects during *evaluation*. Many expressions contain
-sub-expressions, called the *operands* of the expression. The meaning of each
-kind of expression dictates several things:
+Expressions are the fundamental unit of computation.
+When evaluated, they may have *effects* and evaluates to either a value or place.
```
programmerjake, line 45: I think we should probably keep the `(otherwise known as "side effects")` part

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
 
 ## Expression precedence
 
-The precedence of Rust operators and expressions is ordered as follows, going
-from strong to weak. Binary Operators at the same precedence level are grouped
-in the order given by their associativity.
+The grammar for expressions as presented in this book is wrong.
+It purposefully ignores that certain operands cannot be certain expressions due to precedence.
+For example, `a * b + c` is actually parsed as an addition operator expression with the multiplication operator expression as its left operand.
+The grammar, taking precedence into account, is too unweildy to read and understand.
```
programmerjake, line 66:
```suggestion
The grammar, taking precedence into account, is too unwieldy to easily read and understand.
```

Spell fix and insert `easily` since the grammar *can* be understood, it just takes effort.

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
+
+The precedence of operators and expressions is ordered as follows, going from strongest to weakest.
+Binary Operators at the same precedence level are grouped in the order given by their associativity.
+Expressions in this table cannot have operands that are expressions lower in the table.
```
programmerjake, line 70: maybe say something here about how you can have operands that are lower precedence if wrapped in parenthesis or similar -- we don't want to imply it's impossible to write `(a + b) * c` after all

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
 );
 ```
 
-> **Note**: Since this is applied recursively, these expressions are also
-> evaluated from innermost to outermost, ignoring siblings until there are no
-> inner subexpressions.
+> **Note**: Since this is applied recursively, these expressions are also evaluated from innermost to outermost, ignoring siblings until there are no inner subexpressions.
 
-## Place Expressions and Value Expressions
+## Expression categories
```
RalfJung, line 141: Is there any way to keep old links to this section working? I know I linked to this section several times; I forgot if this was in other parts of the documentation (e.g. doc comments) or is GH issues / forum posts.

bjorn3: You can add `<a name="place-expressions-and-value-expressions"></a>`  I think.

src/expressions.md
Havvy: I have that as a note after the table. I could put it before the table.

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
 
-The precedence of Rust operators and expressions is ordered as follows, going
-from strong to weak. Binary Operators at the same precedence level are grouped
-in the order given by their associativity.
+The grammar for expressions as presented in this book is wrong.
+It purposefully ignores that certain operands cannot be certain expressions due to precedence.
+For example, `a * b + c` is actually parsed as an addition operator expression with the multiplication operator expression as its left operand.
+The grammar, taking precedence into account, is too unweildy to read and understand.
+
+The precedence of operators and expressions is ordered as follows, going from strongest to weakest.
```
Havvy, line 68:
```suggestion
The precedence of operators and expressions is ordered, from strongest to weakest, in the following table.
```

src/expressions.md
Havvy: The `name` attribute is deprecated. The span below with an `id` keeps the links working. I just tested this locally, and it scrolls to right below the chapter name, so I need to move the span to before the chapter name, but it otherwise works. We have links throughout the reference as well, and I didn't change them in this PR, yet.

src/expressions.md
Havvy: I could also put `a * (b + c)` in the example as well.

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
 
-An expression may have two roles: it always produces a *value*, and it may have
-*effects* (otherwise known as "side effects"). An expression *evaluates to* a
-value, and has effects during *evaluation*. Many expressions contain
-sub-expressions, called the *operands* of the expression. The meaning of each
-kind of expression dictates several things:
+Expressions are the fundamental unit of computation.
+When evaluated, they may have *effects* and evaluates to either a value or place.
+
+Where expressions are allowed are *expression contexts*.
```
RalfJung, line 47: This sentence sounds rather awkward. What is it you are trying to say? E.g.
"The syntactic positions that permit expressions to be written are called 'expression contexts'."

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
+Expressions commonly contain sub-expressions, called the *operands* of the expression.
+
+The meaning of each kind of expression dictates several things:
 
 * Whether or not to evaluate the operands when evaluating the expression
 * The order in which to evaluate the operands
-* How to combine the operands' values to obtain the value of the expression
+* How to combine the operands' values to obtain the value or place of the expression
```
RalfJung, line 54:
```suggestion
* How to combine the operands' values or places to obtain the value or place of the expression
```
Expressions like `place.field` have operands that are places, so I think if we want to be consistent we need to repeat this 'or place'.

Or we could find some term for "the result of evalauting an expression", i.e., a term for "value or place". *Usually* that term is "value" but we already burned that. In the Stacked Borrows paper formalization, we hence used the term "result".

src/expressions.md
```diff
@@ -41,28 +41,33 @@
...
+> **Note**: We give names to the operands of expressions so that we may discuss them, but these names are not stable and may be changed.
 
 ## Expression precedence
 
-The precedence of Rust operators and expressions is ordered as follows, going
-from strong to weak. Binary Operators at the same precedence level are grouped
-in the order given by their associativity.
+The grammar for expressions as presented in this book is wrong.
+It purposefully ignores that certain operands cannot be certain expressions due to precedence.
+For example, `a * b + c` is actually parsed as an addition operator expression with the multiplication operator expression as its left operand.
```
RalfJung, line 65:
```suggestion
For example, `a * b + c` is actually parsed as an addition operator expression with the multiplication operator expression as its left operand (like `(a * b) + c`).
```

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
+<!-- This section was renamed, but heavily linked to. So adding this span to keep the links working. -->
+<span id="place-expressions-and-value-expressions"></span>
 
-A *place expression* is an expression that represents a memory location. These
-expressions are [paths] which refer to local variables, [static variables],
-[dereferences][deref] (`*expr`), [array indexing] expressions (`expr[expr]`),
-[field] references (`expr.f`) and parenthesized place expressions. All other
-expressions are value expressions.
+Expressions and expression contexts are divided into two *expression categories*: *place* and *value*.
+For expressions, what they evaluate into determines their categories.
```
RalfJung, line 147: This feels awkward. We already established the "place or value" dichotomy earlier, so IMO it'd be better to say upfront (L45) that there are "place expressions" and "value expressions" -- or we find some way to not make the place/value distinction at all until here.

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
+Expressions that evaluate to a value are *value expressions* while those that evaluate to a place are *place expressions*.
+For expression contexts, whether they operate on a place or value determines their context.
```
RalfJung, line 149: I have no idea what this last sentence is supposed to tell me.

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
 
-A *value expression* is an expression that represents an actual value.
+Evaluating an expression in the opposite expression context has effects after evaluation of the expression to produce the expected place or value of the context.
```
RalfJung, line 151: I only understood this sentence after reading the rest of this paragraph. What about something more like "When a place expression is evaluated in value context or vice versa, extra effects apply to perform the required conversion"?

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
+When evaluating a value expression in place expression context, a temporary place is initialized to that value.
+When evaluating a place expression in value expression context, the value at that place is read, moving or copying it.
 
-The following contexts are *place expression* contexts:
+The following lists all place expressions.
+Expressions not listed here are value expressions.
 
-* The left operand of an [assignment][assign] or [compound assignment]
-  expression.
+* [Paths] which refer to local variables or [static variables].
```
RalfJung, line 158: Aren't all paths place expressions?
This sounds like only some paths are place expressions.

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
+* [The dereference operator][deref]
+* [Field access expressions][field]
+* [Tuple indexing expressions]
+* [Array indexing expressions]
+* [Parenthetical expressions][paren] when its enclosed operand is a place expression
+
+The following list all place expression contexts.
```
RalfJung, line 165:
```suggestion
The following lists all place expression contexts.
```

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
+All other contexts are value expression contexts.
+
+* The initializer of a [let statement].
+* The assigned place operand of an [assignment][assign] or [compound assignment] expression.
 * The operand of a unary [borrow] or [dereference][deref] operator.
```
RalfJung, line 170: Should "address-of operator" be listed here?

src/expressions.md
```diff
@@ -133,96 +136,88 @@ assert_eq!(
...
-* The operand of a field expression.
+* The container operand of a field expression.
```
RalfJung, line 171: What's a "container operand"?

src/expressions.md
Havvy: Hmm, I'll think of a way to phrase it better.

src/expressions.md
Havvy: My thought was that L45 gives a summary, and then this section expands upon the idea. It would feel weird to not include the dichotomy in the main definition but including the terms in the main definition would be more detail then necessary.

Also, the fact that the category exists for both the expression and the expression context makes this harder to describe.

src/expressions.md
Havvy: I actually don't know. This was here beforehand.

src/expressions.md
Havvy: That's in your PR which I need to get back to. But my work schedule this week is night/morning/night/morning and I had a migraine over my weekend. I'll rebase against master once that one is merged.

src/expressions.md
Havvy: It's the operand of the field expression, given a name. See the field expression page, which I should add a link to on this list.

src/expressions.md
RalfJung: > Also, the fact that the category exists for both the expression and the expression context makes this harder to describe.

These are not two independent facts, so maybe there is a way to present them as one fact? There are two kinds of expressions, basically two kinds of "non-terminals" in the grammar (`ValExpr`, `PlaceExpr`), so naturally there are two questions: which class does eache xpression fall in, and which `Expr` in the grammar is refined to each of these two non-terminals?

src/expressions.md
RalfJung: Maybe this was meant as "Paths, which refer to ...". IOW, this is a definition of "Paths".

src/expressions.md
RalfJung: No worries, get a good cup of sleep first. :)

src/expressions.md
RalfJung: That name strikes me as rather counterintuitive. Maybe "base" would be better?

Since there is only one operand, just saying "the operand" would also work IMO.

PR #1005, opened 2021-04-09
Improve clarity and style consistency of crate type list

r00ster91:
It adds a little bit more information to the `bin` crate type and standardizes the style below where the operating systems are listed.

### DIFF
--- a/src/linkage.md
+++ b/src/linkage.md
@@ -18,7 +18,8 @@ be ignored in favor of only building the artifacts specified by command line.
 * `--crate-type=bin`, `#[crate_type = "bin"]` - A runnable executable will be
   produced. This requires that there is a `main` function in the crate which
   will be run when the program begins executing. This will link in all Rust and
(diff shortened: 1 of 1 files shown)


r00ster91:
Actually there is still some unclarity here:
We have
> This output type will create `*.so` files on Linux, `*.dylib` files on macOS, and `*.dll` files on Windows.

for `dylib` and `cdylib` and for `staticlib` there's:

> This output type will create an `*.a` archive on Linux, macOS and Windows (MinGW), and a `*.lib` file on Windows (MSVC).

Is this true, `staticlib` will always create a single file and `dylib` and `cdylib` will always create multiple files?

ehuss:
They are all single files, it is just s difference in phrasing.

(FWIW, most of this section is intended to be moved to the rustc book, but I'm happy to accept small updates and fixes to it.)

ehuss (approved):
Thanks!

PR #1006, opened 2021-04-10
Adjust the definition of `target_family`

nagisa:
This generalizes the `target_family` to allow it being specified more
than once and apply to other properties than just the operating system
family.

Of particular interest is an ability to match architectures that are a
part of a family of architecutes (e.g. various versions of ARM, WASM or
other targets)


ehuss (approved):
Thanks!

PR #1008, opened 2021-04-21
added macro_rules to weak keywords

### LINKED ISSUE #1007: Missing weak keywords?
Currently comparing: [this line line symbol.rs](https://github.com/rust-lang/rust/blob/9d9c2c92b834c430f102ea96f65119e37320776e/compiler/rustc_span/src/symbol.rs#L96) and [this file](https://github.com/rust-lang/reference/blob/8db4edd7eab7931859713ac8016ba123ce42b061/src/keywords.md) and it looks like `auto`, `catch`, `default`, `macro_rules`, `raw`, and `union` are missing from the reference?

Is this intentional? If not, I should be able to add a quick PR to fix.

jakevossen5:
Closes #1007

### DIFF
--- a/src/keywords.md
+++ b/src/keywords.md
@@ -95,6 +95,7 @@ The following keywords are reserved beginning in the 2018 edition.
 These keywords have special meaning only in certain contexts. For example, it
 is possible to declare a variable or method with the name `union`.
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1009, opened 2021-04-22
fixed a typo in traits.md

adsick:
line 46: Trait functions are **are** not allowed to be [`async`] or [`const`].
now: Trait functions are not allowed to be [`async`] or [`const`].


ehuss (approved):
Thanks!

PR #1011, opened 2021-05-01
Rearrange HRTB grammar.

ehuss:
_ForLifetimes_ are used in multiple places, so I don't think it makes sense to define them as part of _WhereClause_. Also, I often got confused hitting the "Higher-ranked trait bounds" section and not seeing the grammar there.


tlyu:
This looks good to me, for what it's worth.

matthewjasper (approved):

matthewjasper:
Thanks

PR #1013, opened 2021-05-04
Explicitly state result of compound assignment

tlyu:
The existing text implies but doesn't explicitly say that
compound assignment produces the unit value. A reader could also
infer it from the signatures of the compound assignment trait
functions, but it seems better to be explicit.


ehuss (approved):
Thanks!

PR #1016, opened 2021-05-05
Add crate and module to glossary.

### LINKED ISSUE #1015: Glossary: add "crate" and "module"
Please add "crate" and "module" to the glossary. As someone coming from Python, I'm a bit unsure of what Rust terminology to use instead of Python's "package".

ehuss:
Closes #1015

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -50,6 +50,16 @@ Combinators are higher-order functions that apply only functions and
 earlier defined combinators to provide a result from its arguments.
 They can be used to manage control flow in a modular fashion.
 
+### Crate
+
+A crate is the unit of compilation and linking. There are different [types of
+crates], such as libraries or executables. Crates may link and refer to other
+library crates, called external crates. A crate has a self-contained tree of
+[modules], starting from an unnamed root module called the crate root. [Items]
+may be made visible to other crates by marking them as public in the crate
+root, including through [paths] of public modules.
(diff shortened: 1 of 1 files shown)


tlyu:
These look good to me, for what it's worth. I wonder if it would help to mention that modules might or might not be in separate files, and that crates can consist of multiple source files. (Arguably, mentioning these in the glossary might be redundant with the full description in the body of the reference. Also, Rust's terminology here might be sufficiently idiosyncratic that a small amount of clarification in the glossary would help.)

src/glossary.md
```diff
@@ -130,6 +140,14 @@ This is not affected by applied type arguments. `struct Foo` is considered local
 `Vec<Foo>` is not. `LocalType<ForeignType>` is local. Type aliases do not
 affect locality.
 
+### Module
+
+A module is a container for zero or more [items]. Modules are organized in a
+tree, starting from an unnamed module at the root called the crate root or the
+root module. [Paths] may be used to refer to items outside of a module, which
```
matthewjasper, line 147: "outside of the module that defines them"? "from other modules"?

ehuss: Yea, the sentence was quite awkward. I went with your second suggestion, thanks!

matthewjasper (approved):

matthewjasper:
Thanks

PR #1022, opened 2021-05-10
Expand on Unicode identifiers.

ehuss:
This expands on #999 with more details about the definition of an identifier.


JohnTitor (approved):
The changes here should match the RFC and implementation, LGTM.

PR #1026, opened 2021-05-17
Fix type_length_limit example.

ehuss:
The existing example for the `type_length_limit` attribute wasn't failing as expected.  Somewhere between 2020-09-18 and 2020-09-19 the old example stopped failing (maybe https://github.com/rust-lang/rust/pull/72412 or https://github.com/rust-lang/rust/pull/76575, I can't tell).

I adjusted the example to require some recursion which now triggers the error. This is more similar to the [existing test](https://github.com/rust-lang/rust/blob/fe72845f7bb6a77b9e671e6a4f32fe714962cec4/src/test/ui/type_length_limit.rs).

Additionally, since https://github.com/rust-lang/rust/pull/68664, `ignore` can be removed and it should check the example now.

### DIFF
--- a/src/attributes/limits.md
+++ b/src/attributes/limits.md
@@ -42,19 +42,14 @@ to set the limit based on the number of type substitutions.
 
 > Note: The default in `rustc` is 1048576.
 
-<!-- This code should fail to compile. Unfortunately rustdoc's `compile_fail`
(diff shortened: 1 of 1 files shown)


matthewjasper (approved):

matthewjasper:
thanks

PR #1027, opened 2021-05-17
example for bindings after at

MarcusDunn:
This adds a example of binding a varible after an `@`. `bindings_after_at` is currently unstable, but if [85305](https://github.com/rust-lang/rust/pull/85305) is merged, it may merit an example (so dont merge this untill then :sweat_smile:). I think one could argue the feature is obscure enough that the example may be more noise than useful. I'll leave that call up to those with more expirence than myself.


ehuss (approved):
Thanks!

PR #1030, opened 2021-05-24
missing parameter name in Trait Implementations

tlyu:
The example in Trait Implementations was missing a parameter name in a
trait method declaration, which is no longer allowed in the
2018 edition.


ehuss (approved):
Thanks!

Most of the examples are written for 2015.  We should probably change `book.toml` to have:

```toml
[rust]
edition = "2018"
```

and update all the failing tests.  If that's something you'd be interested in helping with, it would be greatly appreciated!

PR #1034, opened 2021-06-03
Be more precise about array offset in type layouts

owengage:
Currently it states that the nth element has an offset of n*(size of type). I think this is assuming that people are starting at 0, but that's not really what `nth` implies. The first element has an offset of 0, but currently it implies that `n=1`, giving it a non-zero offset.

This might want a larger rewording, but I added the zero-case explicitly to try and be clearer.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -70,9 +70,9 @@ at least equal to the size and alignment of a pointer.
 
 ## Array Layout
 
-Arrays are laid out so that the `nth` element of the array is offset from the
-start of the array by `n * the size of the type` bytes. An array of `[T; n]`
-has a size of `size_of::<T>() * n` and the same alignment of `T`.
+An array of `[T; N]` has a size of `size_of::<T>() * N` and the same alignment
+of `T`. Arrays are laid out so that the zero-based `nth` element of the array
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!

I'm not sure I agree that `nth` implies 1-based indexing.  The standard library [`nth`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth) is 0-based.  

Perhaps it could be qualified something like "so that the zero-based *n*th element of the"?

owengage:
I guess the fact that we disagree means there's potential for confusion at the least. I've reworded it slightly:

1. moving the sentence about the size first, since that makes it more obvious that the first element must be at offset 0 in order to fit them in. 
2. switched to using `N` for the size and `n` for indexing to try and disambiguate between the two.
3. added 'zero-indexed' before the nth as you suggested.

I hope that makes it more clear regardless of what someone expects `nth` to mean.

owengage:
@ehuss any thoughts on this? If it's too pedantic that's fine with me, I'd just rather this doesn't sit open. Thanks for looking. :)

ehuss (approved):
Looks good, thanks!

PR #1036, opened 2021-06-03
Missing TypeParamBounds in TypeAlias

### LINKED ISSUE #1032: `TypeAlias` syntax missing a `TypeParamBounds`
this is valid syntactically:
```rs
type TypeAlias: TraitBound = SomeType;
```
however it is not part of the reference.

expectation:
> **<sup>Syntax</sup>**\
> _TypeAlias_ :\
> &nbsp;&nbsp; `type` [IDENTIFIER]()&nbsp;[_GenericParams_]()<sup>?</sup> ( `:` [TypeParamBounds]() )<sup>?</sup> [_WhereClause_]()<sup>?</sup> ( `=` [_Type_]() )<sup>?</sup> `;`

*should be stated that this syntax is only useful in the context of an associated type and will yield ``bounds on `type`s in this context have no effect`` anywhere else.

edit: I accidently pressed `enter` before writing the description

GilRtr:
Fixes #1032.
I did not know how to add the information that this documented syntax is not always correct. (the `*` in the original issue)

### DIFF
--- a/src/items/type-aliases.md
+++ b/src/items/type-aliases.md
@@ -3,6 +3,7 @@
 > **<sup>Syntax</sup>**\
 > _TypeAlias_ :\
 > &nbsp;&nbsp; `type` [IDENTIFIER]&nbsp;[_GenericParams_]<sup>?</sup>
+>              ( `:` [_TypeParamBounds_] )<sup>?</sup>
 >              [_WhereClause_]<sup>?</sup> ( `=` [_Type_] )<sup>?</sup> `;`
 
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! To address the limitation, I would add a sentence at the bottom that is similar to the statement about _Type_ being limited. Maybe something along these lines:

The [_TypeParamBounds_] may only be specified when used as an [associated type] in a [trait].

ehuss (approved):
Thanks!

PR #1037, opened 2021-06-03
Fix typos + grammar

kklibo:


JohnTitor (approved):

src/destructors.md
```diff
@@ -1,7 +1,7 @@
 # Destructors
 
 When an [initialized]&#32;[variable] or [temporary] goes out of
-[scope](#drop-scopes) its *destructor* is run, or it is *dropped*. [Assignment]
+[scope](#drop-scopes), its *destructor* is run (also phrased as: "it is *dropped*"). [Assignment]
```
ehuss, line 4: I'm not entirely certain this is what was meant by the the original.  @matthewjasper can you clarify here?

My understanding is that "dropping" and "running destructors" have different definitions, and that not all values have destructors.  I don't think the two terms can always be used interchangeably. That is:

- When any value goes out of scope, it is "dropped".
- If the value has a destructor, it is run.

ehuss (commented):
Thanks, these a great fixes, very much appreciated!

src/destructors.md
kklibo: I didn't intend to change the meaning of this sentence. If "dropping" and "running destructors" have different meanings here, then my change was wrong. It would be better to revert it and add an explanation here about the difference between the terms.

ehuss: I went ahead and pushed a revert for this, and filed #1078 to follow up on this particular question.

ehuss (approved):
Thank you very much!

PR #1040, opened 2021-06-04
Add "scopes" chapter.

ehuss:
This adds the chapter on scopes.

I tried to make this accurate and complete, but it is probably neither. There are some things that are intentionally not covered, such as some ambiguities (will handle in other places or at a later date), peculiarities with macros (perhaps deferred for more explicit "expansion" docs?), details regarding `use` (will add to `use` chapter), and known bugs (hopefully they will get fixed some day).

cc #568

### DIFF
--- a/src/items.md
+++ b/src/items.md
@@ -53,15 +53,14 @@ There are several kinds of items:
 * [implementations]
 * [`extern` blocks]
 
-Some items form an implicit scope for the declaration of sub-items. In other
-words, within a function or module, declarations of items can (in many cases)
-be mixed with the statements, control blocks, and similar artifacts that
-otherwise compose the item body. The meaning of these scoped items is the same
-as if the item was declared outside the scope &mdash; it is still a static item
-&mdash; except that the item's *path name* within the module namespace is
-qualified by the name of the enclosing item, or is private to the enclosing
-item (in the case of functions). The grammar specifies the exact locations in
-which sub-item declarations may appear.
+Items may be declared in the [root of the crate], a [module][modules], or a [block expression].
+A subset of items, called [associated items], may be declared in [traits] and [implementations].
+A subset of items, called external items, may be declared in [`extern` blocks].
+
+Items may be defined in any order, with the exception of [`macro_rules`] which has its own scoping behavior.
+[Name resolution] of item names allows items to be defined before or after where the item is referred to in the module or block.
+
+See [item scopes] for information on the scoping rules of items.
 
 [_ConstantItem_]: items/constant-items.md
 [_Enumeration_]: items/enumerations.md
@@ -82,15 +81,23 @@ which sub-item declarations may appear.
 [_Visibility_]: visibility-and-privacy.md
 [`extern crate` declarations]: items/extern-crates.md
 [`extern` blocks]: items/external-blocks.md
+[`macro_rules`]: macros-by-example.md
 [`use` declarations]: items/use-declarations.md
+[associated items]: items/associated-items.md
+[block expression]: expressions/block-expr.md
 [constant items]: items/constant-items.md
 [enumeration definitions]: items/enumerations.md
 [function definitions]: items/functions.md
 [implementations]: items/implementations.md
+[item scopes]: names/scopes.md#item-scopes
 [modules]: items/modules.md
+[name resolution]: names/name-resolution.md
 [paths]: paths.md
+[root of the crate]: crates-and-source-files.md
+[statement]: statements.md
 [static items]: items/static-items.md
 [struct definitions]: items/structs.md
 [trait definitions]: items/traits.md
+[traits]: items/traits.md
 [type definitions]: items/type-aliases.md
 [union definitions]: items/unions.md
--- a/src/names/scopes.md
+++ b/src/names/scopes.md
@@ -1,3 +1,350 @@
 # Scopes
 
-> **Note**: This is a placeholder for future expansion.
+A *scope* is the region of source text where a named [entity] may be referenced with that name.
+The following sections provide details on the scoping rules and behavior, which depend on the kind of entity and where it is declared.
+The process of how names are resolved to entities is described in the [name resolution] chapter.
+More information on "drop scopes" used for the purpose of running destructors may be found in the [destructors] chapter.
+
+## Item scopes
+
+The name of an [item][items] declared directly in a [module] has a scope that extends from the start of the module to the end of the module. These items are also members of the module and can be referred to with a [path] leading from their module.
+
+The name of an item declared as a [statement] has a scope that extends from the start of the block the item statement is in until the end of the block.
+
+It is an error to introduce an item with a duplicate name of another item in the same [namespace] within the same module or block.
+[Asterisk glob imports] have special behavior for dealing with duplicate names and shadowing, see the linked chapter for more details.
+Items in a module may shadow items in a [prelude](#prelude-scopes).
+
+Item names from outer modules are not in scope within a nested module.
+A [path] may be used to refer to an item in another module.
+
+### Associated item scopes
+
+[Associated items] are not scoped and can only be referred to by using a [path] leading from the type or trait they are associated with.
+[Methods] can also be referred to via [call expressions].
+
+Similar to items within a module or block,  it is an error to introduce an item within a trait or implementation that is a duplicate of another item in the trait or impl in the same namespace.
+
+## Pattern binding scopes
+
+The scope of a local variable [pattern] binding depends on where it is used:
+
+* [`let` statement] bindings range from just after the `let` statement until the end of the block where it is declared.
+* [Function parameter] bindings are within the body of the function.
+* [Closure parameter] bindings are within the closure body.
+* [`for`] and [`while let`] bindings are within the loop body.
+* [`if let`] bindings are within the consequent block.
+* [`match` arms] bindings are within the [match guard] and the match arm expression.
+
+Local variable scopes do not extend into item declarations.
+<!-- Not entirely, see https://github.com/rust-lang/rust/issues/33118 -->
+
+### Pattern binding shadowing
+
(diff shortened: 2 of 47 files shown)


src/names/scopes.md
```diff
@@ -1,3 +1,324 @@
...
+[Derive macro helper attributes] are in scope in the item where their corresponding [`derive` attribute] is specified.
+The scope extends from just after the `derive` attribute to the end of the item. <!-- Note: Not strictly true, see https://github.com/rust-lang/rust/issues/79202, but this is the intention. -->
+Helper attributes shadow other attributes of the same name in scope.
+
+## `Self` scope
+
+Although [`Self`] is a keyword with special meaning, it interacts with name resolution in a way similar to normal names.
+
+The implicit `Self` type in the definition of a [struct], [enum], [union], [trait], or [implementation] is treated similarly to a [generic parameter](#generic-parameter-scopes), and is in scope in the same way as a generic type parameter.
+For struct, enum, and unions, it is in scope starting after the generic parameters.
```
JohnTitor, line 260:
```suggestion
For structs, enums, and unions, it is in scope starting after the generic parameters.
```

JohnTitor (approved):
Overall LGTM 👍

ehuss:
@petrochenkov Is this something you could potentially review/look over?

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
 # Scopes
 
-> **Note**: This is a placeholder for future expansion.
+A *scope* is the region of source text where a named [entity] may be referenced with that name.
+The following sections provide details on the scoping rules and behavior, which depend on the kind of entity and where it is declared.
+The process of how names are resolved to entities is described in the [name resolution] chapter.
+More information on "drop scopes" used for the purpose of running destructors maybe be found in the [destructors] chapter.
```
jyn514, line 6:
```suggestion
More information on "drop scopes" used for the purpose of running destructors may be found in the [destructors] chapter.
```

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+[Associated items] are not scoped and can only be referred to by using a [path] leading from the type they are associated with.
+[Methods] can also be referred to via [call expressions].
+
+Similar to items within a module or block, it is an error to introduce an item within a trait or implementation that is a duplicate of another item in the trait or impl in the same namespace.
+
+## Pattern binding scopes
+
+The scope of a local variable [pattern] binding depends on where it is used:
+
+* [`let` statement] bindings range from just after the `let` statement until the end of the block where it is declared.
```
jyn514, line 32: should this mention shadowing?

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+* [Function parameter] bindings are within the body of the function.
+* [Closure parameter] bindings are within the closure body.
+* [`for`] and [`while let`] bindings are within the loop body.
+* [`if let`] bindings are within the consequent block.
```
jyn514, line 36: is "consequent" a technical term here? Trying to figure out whether this implies that it goes out of scope in the `else` block.

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+# trait SomeTrait<'a, T> {}
+// The <'a, U> for `SomeTrait` refer to the 'a and U parameters of `bounds_scope`.
+fn bounds_scope<'a, T: SomeTrait<'a, U>, U>() {}
+
+fn where_scope<'a, T, U>()
+    where T: SomeTrait<'a, U>
+{}
+```
+
+Generic scopes do not extend into [items] declared inside a function.
```
petrochenkov, line 93: This is kind of a philosophical question, but they do extend (*), it's just an error to use a parameter from an outer item.

(*) until a `mod` item is encountered, which is a true barrier for names.

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+
+### Generic parameter shadowing
+
+It is an error to shadow a generic parameter.
+Items declared within functions are allowed to reuse generic parameter names from the function because generic scopes do not extend to inner items.
```
petrochenkov, line 98: The motivation after "because" is artificial.
This is allowed because shadowing of names is allowed in general, items in inner scopes can shadow items from outer scopes, et cetera.

The restrictions on generic parameters that appear in some cases are exceptions.
They exist for historical reasons, but technically nothing prevents removing them at any moment.

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+    // Items within functions are allowed to reuse generic parameter in scope
+    // because all generics are not in scope within inner items.
+    fn inner<'a>() {} // OK
+}
+```
+
+```rust,compile_fail
+trait SomeTrait<'a> {
+    fn example<'a>() {} // ERROR: 'a is already in scope
+}
```
petrochenkov, line 111: It would be good to add type and const parameters to these examples.
I'm not actually sure that the rules here are enforced consistently for all of them.

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+
+// The label is in scope for the expression of `while` loops.
+'a: while break 'a {}         // Loop does not run.
+'a: while let _ = break 'a {} // Loop does not run.
+
+// The label is not in scope in the defining `for` loop:
+'a: for outer in 0..5 {
+    // This will break the outer loop, skipping the inner loop and stopping
+    // the outer loop.
+    'a: for inner in { break 'a; 0..1 } {
```
petrochenkov, line 209: Fascinating.

src/names/scopes.md
```diff
@@ -1,3 +1,329 @@
...
+The scope extends from just after the `derive` attribute to the end of the item. <!-- Note: Not strictly true, see https://github.com/rust-lang/rust/issues/79202, but this is the intention. -->
+Helper attributes shadow other attributes of the same name in scope.
+
+## `Self` scope
+
+Although [`Self`] is a keyword with special meaning, it interacts with name resolution in a way similar to normal names.
+
+The implicit `Self` type in the definition of a [struct], [enum], [union], [trait], or [implementation] is treated similarly to a [generic parameter](#generic-parameter-scopes), and is in scope in the same way as a generic type parameter.
+For structs, enums, and unions, it is in scope starting after the generic parameters.
+For traits and implementations, it is in scope starting just before the generic parameters.
```
petrochenkov, line 266: I'm not sure that this difference is not a bug.
The original implementation of `Self` is for traits and impls, and it is certainly correct.

src/names/scopes.md
ehuss: That seems to be an implementation choice, though?  Like, the choice is:

1) Generic names do not extend into an inner item. Using a parent name will be a "name not found" error.
2) Generic names do extend into an inner item, but can be shadowed. Using a parent name without shadowing it will be a "can't use generic parameter from outer function" error.

I felt like the first approach was a little easier to explain, even if that's not how `rustc` works (which I see as just a way to provide better errors, or maybe just easier to implement).

That being said, I reworded it to say it is an error.

src/items.md
```diff
@@ -53,15 +53,13 @@ There are several kinds of items:
...
-Some items form an implicit scope for the declaration of sub-items. In other
-words, within a function or module, declarations of items can (in many cases)
-be mixed with the statements, control blocks, and similar artifacts that
-otherwise compose the item body. The meaning of these scoped items is the same
-as if the item was declared outside the scope &mdash; it is still a static item
-&mdash; except that the item's *path name* within the module namespace is
-qualified by the name of the enclosing item, or is private to the enclosing
-item (in the case of functions). The grammar specifies the exact locations in
-which sub-item declarations may appear.
+Items may be declared in the [root of the crate], a [module][modules], or a [statement].
```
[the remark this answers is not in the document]
ehuss: Yea, that was a bit awkward. I switched to a block expression.

src/names/scopes.md
ehuss: Yea, the consequent block is the block immediately after the conditional expression. It is defined in https://doc.rust-lang.org/nightly/reference/expressions/if-expr.html.  Else blocks are called "else blocks".

src/names/scopes.md
ehuss: I would prefer not to add too many up-front qualifications so it doesn't obscure the statements being made. Shadowing is mentioned immediately afterwards, which applies to all of the pattern binding sites.  

One problem with defining something (like Rust) is that one needs to take into consideration the whole of the document, which includes many refinements and clarifications on previously stated things (and to make it worse, those interactions can be a tangled web).  It can be a difficult balance to warn a reader about upcoming qualifications, since to some degree it would be repeating what is about to come next.

src/names/scopes.md
ehuss: I reworded this a bit.

src/names/scopes.md
ehuss: I'm not aware of any inconsistent rules, but if any come to mind, let me know.

src/names/scopes.md
ehuss: I think I made a mistake here, I'm not sure what I was thinking at the time.  Perhaps I got confused by [this code](https://github.com/rust-lang/rust/blob/f9b2e3c87b76c03cc398be8b8f65d096c0583eb2/compiler/rustc_resolve/src/late.rs#L1082-L1094), but based on [this](https://github.com/rust-lang/rust/blob/f9b2e3c87b76c03cc398be8b8f65d096c0583eb2/compiler/rustc_resolve/src/late.rs#L1532-L1554) I'm almost certain that struct/enum/union have the same behavior as traits and impls.


```rust
struct S<T: Into<Self>> { // OK
    f1: T,
}
```

I have pushed an update, does that seem correct?

ehuss:
Thanks for the reviews!

traviscross (approved):

PR #1044, opened 2021-06-08
Make explicit reference to scrutinee expression in grammar snippet

AloeareV:
Adding an explicit mention of the scrutinee in the grammar snippet makes the usage of the term in the description clearer, at @zancas's suggestion.

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -41,12 +41,15 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` Scrutinee
 >              [_BlockExpression_]\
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!  Can you also update `if let` and `while let` to use the new named production?

AloeareV:
I hadn't realized that the term applied there, even though it's right there in the text!

src/expressions/if-expr.md
```diff
@@ -41,12 +41,15 @@ assert_eq!(y, "Bigger");
 
 > **<sup>Syntax</sup>**\
 > _IfLetExpression_ :\
-> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Expression_]<sub>_except struct or lazy boolean operator expression_</sub>
+> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` _Scrutinee_
```
ehuss, line 44: Instead of re-defining it, can you link to a single definition?  Something like:

```suggestion
> &nbsp;&nbsp; `if` `let` [_Pattern_] `=` [_Scrutinee_]
```

With a link to the `match` section, like:

```md
[_Scrutinee_]: match-expr.md
```

ehuss (approved):
Thanks!

PR #1046, opened 2021-06-11
Add a note why the same size int casting is a no-op

JohnTitor:
This is a counterpart to https://github.com/rust-lang/nomicon/pull/277.


ehuss (approved):
Thanks!

PR #1047, opened 2021-06-11
Add notes on `#[target_feature]` for wasm

alexcrichton:
Also document the `simd128` feature stabilized in rust-lang/rust#86204


ehuss (approved):
Thanks! ❤️

PR #1050, opened 2021-06-17
update UB list for safe target_feature

RalfJung:
Cc @alexcrichton

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -75,7 +75,7 @@ The following is a list of the available feature names.
 
 #### `x86` or `x86_64`
 
-Executing code with unsupported features is UB on this platform.
+Executing code with unsupported features is undefined behavior on this platform.
 Hence this platform requires that `#[target_feature]` is only applied to [`unsafe`
 functions][unsafe function].
 
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -63,7 +63,8 @@ error to specify a feature for a target architecture that the crate is not
 being compiled for.
 
 It is [undefined behavior] to call a function that is compiled with a feature
-that is not supported on the current platform the code is running on.
+that is not supported on the current platform the code is running on, *except*
+if the platform explicitly documents this to be safe.
 
 Functions marked with `target_feature` are not inlined into a context that
 does not support the given features. The `#[inline(always)]` attribute may not
@@ -134,9 +135,12 @@ Feature     | Implicitly Enables | Description
 
 #### `wasm32` or `wasm64`
 
-Executing code with unsupported features is allowed (i.e., is not UB) on this platform.
-Hence this platform allows `#[target_feature]` to be applied to both safe and
-[`unsafe` functions][unsafe function].
+`#[target_feature]` may be used with both safe and
+[`unsafe` functions][unsafe function] on Wasm platforms. It is impossible to
+cause undefined behavior via the `#[target_feature]` attribute because
+attempting to use instructions unsupported by the Wasm engine will fail at load
+time without the risk of being interpreted in a way different from what the
+compiler expected.
 
 Feature     | Description
 ------------|-------------------
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -75,7 +75,8 @@ The following is a list of the available feature names.
 
 #### `x86` or `x86_64`
 
-This platform requires that `#[target_feature]` is only applied to [`unsafe`
+Executing code with unsupported features is UB on this platform.
+Hence this platform requires that `#[target_feature]` is only applied to [`unsafe`
 functions][unsafe function].
 
 Feature     | Implicitly Enables | Description
@@ -133,7 +134,8 @@ Feature     | Implicitly Enables | Description
 
 #### `wasm32` or `wasm64`
 
-This platform allows `#[target_feature]` to be applied to both safe and
+Executing code with unsupported features is allowed (i.e., is not UB) on this platform.
+Hence this platform allows `#[target_feature]` to be applied to both safe and
 [`unsafe` functions][unsafe function].
 
 Feature     | Description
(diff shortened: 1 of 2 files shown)


src/attributes/codegen.md
```diff
@@ -133,7 +134,8 @@ Feature     | Implicitly Enables | Description
 
 #### `wasm32` or `wasm64`
 
-This platform allows `#[target_feature]` to be applied to both safe and
+Executing code with unsupported features is allowed (i.e., is not UB) on this platform.
+Hence this platform allows `#[target_feature]` to be applied to both safe and
```
alexcrichton, line 138: This may want to be reworded slightly, since I think we still want to reserve the right to say that if you actually execute undefined instructions that's UB. That being said the target inherently provides no mechanism or means by which to do this.

Put another way "executing code with unsupported feautres" is not necesarily explicitly safe on wasm platforms, it's just not inherently possible in the platforms itself.

Basically I want to leave us room to handle [feature detection](https://github.com/WebAssembly/feature-detection/blob/main/proposals/feature-detection/Overview.md) at some point in the future. Even then though that will be implemented in a way that you won't be able to get it wrong so it won't be `unsafe` (just sort of inherently with the wasm platform).

I dunno this is definitely sort of lawyer-y, but I think ideally this would be reworded a bit to say that `#[target_feature]` can be applied to safe functions because it's not possible to execute unknown instructions on wasm.

RalfJung: > This may want to be reworded slightly, since I think we still want to reserve the right to say that if you actually execute undefined instructions that's UB. That being said the target inherently provides no mechanism or means by which to do this.

UB is a Rust concept here, not a target concept. So this is about explaining what you are or are not allowed to do *in Rust* when compiling to this target.

In Rust it is definitely possible to write code that ends up executing unknown instructions, even when compiling to wasm. My understanding is that this is guaranteed to trap, but that is not the same as "it's impossible to do so".

bjorn3: > My understanding is that this is guaranteed to trap, but that is not the same as "it's impossible to do so".

AFAIK it causes the wasm module to fail to load at all if the simd instructions are not optimized away. The wasm runtime needs to know about all used instructions in a wasm module to be able to parse and typecheck them.

alexcrichton: Ok, I'll leave  this to you then

RalfJung: To be clear, I am open to suggestions. :)

src/attributes/codegen.md
```diff
@@ -133,7 +134,8 @@ Feature     | Implicitly Enables | Description
 
 #### `wasm32` or `wasm64`
 
-This platform allows `#[target_feature]` to be applied to both safe and
+Executing code with unsupported features is allowed (i.e., is not UB) on this platform.
```
bjorn3, line 137:
```suggestion
Attempting to use instructions unsupported by the wasm engine will fail at load time
without the risk of being interpreted in a way different from what the compiler expected.
```

RalfJung: I think it should explicitly say something about there being no Rust surface UB -- as the UB list says, this is defined per-target, so the targets should explicitly define it.

basically, there's a `bool` per target whether this is UB or not; we need to say explicitly which value taht `bool` ahs for wasm.

RalfJung:
So how do we make progress here? The current state of the reference is in contradiction with target_feature being safe on some platforms.

src/attributes/codegen.md
```diff
@@ -75,7 +75,8 @@ The following is a list of the available feature names.
 
 #### `x86` or `x86_64`
 
-This platform requires that `#[target_feature]` is only applied to [`unsafe`
+Executing code with unsupported features is UB on this platform.
```
ehuss, line 78:
```suggestion
Executing code with unsupported features is undefined behavior on this platform.
```

src/attributes/codegen.md
ehuss: I like the way bjorn3 has phrased it here. If you specifically want to make sure that the term "undefined behavior" shows up here, then I think a small adjustment could be made:

> `#[target_feature]` may be used with both safe and unsafe functions on Wasm platforms.
> Although it is undefined behavior to execute unsupported instructions, attempting to use instructions unsupported by the Wasm engine will fail at load time without the risk of being interpreted in a way different from what the compiler expected.

RalfJung: > Although it is undefined behavior to execute unsupported instructions

The thing is though that this is *not* UB. Rust makes it safe to invoke a function with the wrong target_feature on wasm, so if that was UB that would be unsound.

(The UB we care about here is Rust UB. It is the Rust backend responsibility to worry about target UB, and certainly of no concern for the Rust Reference to worry about what is or is not UB in wasm programs.)

ehuss: I understand there are two perspectives of UB, but that is a very fine distinction that doesn't come across in the current wording. The statement that "Executing code with unsupported features is allowed on this platform" by itself doesn't come across as being correct to me because the platform does not allow executing code with unsupported features (sort of the exact opposite of what is written). I understand that you're trying to say Rust doesn't care, but it isn't coming across that way to me.

Perhaps another crack at it:

> `#[target_feature]` may be used with both safe and unsafe functions on Wasm platforms.
> It is not undefined behavior to use the `#[target_feature]` attribute because attempting to use instructions unsupported by the Wasm engine will fail at load time without the risk of being interpreted in a way different from what the compiler expected.

RalfJung: Yes I like that, thanks! I just slightly changed the second sentence.

src/attributes/codegen.md
RalfJung: @alexcrichton this paragraph got reworded, what do you think about the latest version?

> `#[target_feature]` may be used with both safe and
[`unsafe` functions][unsafe function] on Wasm platforms. It is impossible to
cause undefined behavior via the `#[target_feature]` attribute because
attempting to use instructions unsupported by the Wasm engine will fail at load
time without the risk of being interpreted in a way different from what the
compiler expected.

alexcrichton: Sure!

ehuss (approved):
Thanks, I appreciate it!

PR #1051, opened 2021-06-24
Remove inner attributes from non-block expressions.

### LINKED ISSUE #1021: Mistake on where attributes are allowed in StructExpr and more
According to the [reference](https://doc.rust-lang.org/nightly/reference/expressions/struct-expr.html), only inner attributes are allowed in an StructExpr:

```
StructExprStruct :
   PathInExpression { InnerAttribute* (StructExprFields | StructBase)? }
```

However, this seems not to be the behaviour of rustc. rustc instead accepts only outer attributes for each field, but not inner attributes for the entire expression.

For example, this compiles in rustc:

``` rust
struct S { x: i64, y: i64 }
fn main() {
    let s =  S { x: 1, #[deprecated] y: 2};
}
```

But not this:

``` rust
struct S { x: i64, y: i64 }
fn main() {
    let s =  S { #![deprecated] x: 1, y: 2};
}
```

ehuss:
Inner attributes were removed in https://github.com/rust-lang/rust/pull/83312, with https://github.com/rust-lang/rust/pull/85193 re-adding support in match expressions.

Closes #1021

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -4,7 +4,7 @@
 
 > **<sup>Syntax</sup>**\
 > _ArrayExpression_ :\
-> &nbsp;&nbsp; `[` [_InnerAttribute_]<sup>\*</sup> _ArrayElements_<sup>?</sup> `]`
+> &nbsp;&nbsp; `[` _ArrayElements_<sup>?</sup> `]`
 >
 > _ArrayElements_ :\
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):
LGTM 👍

JohnTitor:
@ehuss Do you want an additional reviewer or could I merge this?

ehuss:
I was thinking it would be nice if someone on the lang team could give it a stamp of approval, but I think that is unlikely so I would say just go ahead and merge. I'm pretty confident it covers the linked PRs.

PR #1052, opened 2021-07-01
Add details on how names are introduced.

### LINKED ISSUE #129: Items must show what paths they make available.

ehuss:
This expands on several sections to specify exactly which names are introduced by items. This also contains a few tangential updates, such as explaining what `Self` is and how it is defined, and a major update to the `use` chapter.

This does not cover everything; that's a multi-year effort. This is intended as a small step along that journey. One giant gaping hole is covering expansion and peculiarities related to that. The ambiguities section is also incomplete.

Closes #129
cc #487
cc #568

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -83,7 +83,7 @@ enum Enum {
 
 Variant constructors are similar to [struct] definitions, and can be referenced by a path from the enumeration name, including in [use declarations].
 Each variant defines its type in the [type namespace], though that type cannot be used as a type specifier.
-Each variant also defines a constructor in the [value namespace].
+Tuple-like and unit-like variants also define a constructor in the [value namespace].
 
 A struct-like variant can be instantiated with a [struct expression].
 A tuple-like variant can be instantiated with a [call expression] or a [struct expression].
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -30,6 +30,7 @@ nominal [enumerated type] as well as a set of *constructors*, that can be used
 to create or pattern-match values of the corresponding enumerated type.
 
 Enumerations are declared with the keyword `enum`.
+The `enum` declaration defines the enumeration type in the [type namespace] of the module or block where it is located.
 
 An example of an `enum` item and its use:
 
@@ -78,6 +79,27 @@ enum Enum {
     Bar = 2,
     Baz = 1,
 }
+```
+
+Variant constructors are similar to [struct] definitions, and can be referenced by a path from the enumeration name, including in [use declarations].
+The constructors are defined in both the [type namespace] and [value namespace] within the enumeration.
+
+A struct-like variant can be instantiated with a [struct expression].
+A tuple-like variant can be instantiated with a [call expression].
+A unit-like variant can be instantiated with a [path expression].
+For example:
+
+```rust
+enum Examples {
+    UnitLike,
+    TupleLike(i32),
+    StructLike { value: i32 },
+}
+
+use Examples::*; // Creates aliases to all variants.
+let x = UnitLike; // Path expression of the const item.
+let y = TupleLike(123); // Call expression.
+let z = StructLike { value: 123 }; // Struct expression.
 ```
 
 <span id="custom-discriminant-values-for-fieldless-enumerations"></span>
@@ -299,20 +321,27 @@ enum E {
 }
 ```
 
-[IDENTIFIER]: ../identifiers.md
-[_GenericParams_]: generics.md
-[_WhereClause_]: generics.md#where-clauses
 [_Expression_]: ../expressions.md
-[_TupleFields_]: structs.md
+[_GenericParams_]: generics.md
 [_StructFields_]: structs.md
+[_TupleFields_]: structs.md
 [_Visibility_]: ../visibility-and-privacy.md
-[enumerated type]: ../types/enum.md
+[_WhereClause_]: generics.md#where-clauses
+[`C` representation]: ../type-layout.md#the-c-representation
 [`mem::discriminant`]: ../../std/mem/fn.discriminant.html
-[never type]: ../types/never.md
-[unit-only]: #unit-only-enum
-[numeric cast]: ../expressions/operator-expr.md#semantics
+[call expression]: ../expressions/call-expr.md
 [constant expression]: ../const_eval.md#constant-expressions
 [default representation]: ../type-layout.md#the-default-representation
-[primitive representation]: ../type-layout.md#primitive-representations
-[`C` representation]: ../type-layout.md#the-c-representation
+[enumerated type]: ../types/enum.md
 [Field-less enums]: #field-less-enum
+[IDENTIFIER]: ../identifiers.md
+[never type]: ../types/never.md
+[numeric cast]: ../expressions/operator-expr.md#semantics
+[path expression]: ../expressions/path-expr.md
+[primitive representation]: ../type-layout.md#primitive-representations
+[struct expression]: ../expressions/struct-expr.md
+[struct]: structs.md
+[type namespace]: ../names/namespaces.md
+[unit-only]: #unit-only-enum
+[use declarations]: use-declarations.md
+[value namespace]: ../names/namespaces.md
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -82,11 +82,12 @@ enum Enum {
 ```
 
 Variant constructors are similar to [struct] definitions, and can be referenced by a path from the enumeration name, including in [use declarations].
-The constructors are defined in both the [type namespace] and [value namespace] within the enumeration.
+Each variant defines its type in the [type namespace], though that type cannot be used as a type specifier.
+Each variant also defines a constructor in the [value namespace].
 
 A struct-like variant can be instantiated with a [struct expression].
-A tuple-like variant can be instantiated with a [call expression].
-A unit-like variant can be instantiated with a [path expression].
+A tuple-like variant can be instantiated with a [call expression] or a [struct expression].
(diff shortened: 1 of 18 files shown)


src/items/enumerations.md
```diff
@@ -61,6 +62,27 @@ integer associated to it that is used to determine which variant it holds. An
 opaque reference to this discriminant can be obtained with the
 [`mem::discriminant`] function.
 
+Variant constructors are similar to [struct] definitions, and can be referenced by a path from the enumeration name, including in [use declarations].
+The constructors are defined in both the [type namespace] and [value namespace] within the enumeration.
+
+A struct-like variant can be instantiated with a [struct expression].
+A tuple-like variant can be instantiated with a [call expression].
```
petrochenkov, line 69: Or a struct expression (`TupleVariant(value)` / `TupleVariant { 0: value }`).

src/items/enumerations.md
```diff
@@ -61,6 +62,27 @@ integer associated to it that is used to determine which variant it holds. An
...
+A unit-like variant can be instantiated with a [path expression].
```
petrochenkov, line 70: Or a struct expression (`UnitVariant` / `UnitVariant {}`).

src/paths.md
```diff
@@ -206,10 +206,19 @@ fn bar() {
 
 ### `Self`
 
-`Self`, with a capital "S", is used to refer to the implementing type within
-[traits] and [implementations].
+`Self`, with a capital "S", is used to refer to the current type being implemented or defined. It may be used in the following situations:
+
+* In a [trait] definition, it refers to the type implementing the trait.
+* In an [implementation], it refers to the implementing type.
+  When implementing a tuple or unit [struct], [enumeration], or [union], it also refers to the constructor in the [value namespace].
```
petrochenkov, line 213:
```suggestion
  When implementing a tuple or unit [struct], it also refers to the constructor in the [value namespace].
```
Enumerations and unions never have constructors.

src/items/use-declarations.md
```diff
@@ -13,6 +13,7 @@ A _use declaration_ creates one or more local name bindings synonymous with
 some other [path]. Usually a `use` declaration is used to shorten the path
 required to refer to a module item. These declarations may appear in [modules]
 and [blocks], usually at the top.
+A `use` declaration is also sometimes called an "import".
```
petrochenkov, line 16: (Or "reexport" if it's not private.)

src/items/use-declarations.md
```diff
@@ -85,75 +86,186 @@ In this example, the module `quux` re-exports two public names defined in
...
+
+```rust
+// Creates bindings to:
+// std::collections::BTreeSet
+// std::collections::hash_map
+// std::collections::hash_map::HashMap
+use std::collections::{BTreeSet, hash_map::{self, HashMap}};
+```
+
+An empty brace does not import anything. `a::b::{}` is treated as `a::b::{self as _}` and `use {};` has no effect.
```
petrochenkov, line 166: The `self as _` desugaring performed by rustc results in a bug - https://github.com/rust-lang/rust/issues/61826.
It causes "import nothing" declarations like `use Trait::{};` act as "import trait" declarations `use Trait as _;`, bringing the trait into scope and making its methods available.

src/items/use-declarations.md
```diff
@@ -85,75 +86,186 @@ In this example, the module `quux` re-exports two public names defined in
...
+// This only imports the module `foo`. The function `foo` lives in
+// the value namespace and is not imported.
+use bar::foo::{self};
+
+fn main() {
+    foo(); // Error, `foo` is a module
+}
+```
+
+> **Note**: `self` as the first segment of a `use` path has a different meaning from `self` inside braces.
```
petrochenkov, line 208: It's actually exactly the same meaning - current module (of the parent segment), or simply the current module if there's no parent segment.

We just don't support this syntax consistently in all positions.
`use foo::self::bar;` would work with exactly the same meaning too (equivalent to `use foo::bar;`), if it was supported.

src/items/use-declarations.md
```diff
@@ -85,75 +86,186 @@ In this example, the module `quux` re-exports two public names defined in
...
+    pub fn bar() {
+        // Creates local aliases to V1 and V2 of the Example enum.
+        use Example::*;
+        let x = V1;
+    }
+}
+```
+
+Items and named imports are allowed to shadow names from glob imports in the same [namespace].
+That is, if there is a name already defined by another item in the same namespace, the glob import will skip it.
```
petrochenkov, line 235: Technically the glob import won't skip it, it will still import the name, it will just be shadowed by the explicit import/item, in the same way as explicit imports/items shadow names from preludes.
(At least that's the intention, the actual implementation is pretty messy right now.)

src/items/enumerations.md
```diff
@@ -61,6 +62,27 @@ integer associated to it that is used to determine which variant it holds. An
 opaque reference to this discriminant can be obtained with the
 [`mem::discriminant`] function.
 
+Variant constructors are similar to [struct] definitions, and can be referenced by a path from the enumeration name, including in [use declarations].
+The constructors are defined in both the [type namespace] and [value namespace] within the enumeration.
```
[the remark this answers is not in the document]
ehuss: Can you clarify something here?  The tuple-struct constructor in the value namespace doesn't currently have much of a purpose, does it?  At least, I can't think of a way it could be used. The only thing I noticed is that you are not allowed to shadow it, as in:

```rust
enum E {
    V1 {f1: i32}
}

fn foo() {
    use E::*;
    let V1 = 123; // ERROR: let bindings cannot shadow struct variants
}
```

ehuss:
Thanks so much for the review, I do appreciate it!

src/items/use-declarations.md
```diff
@@ -199,8 +313,100 @@ m!(use std as _;);
...
+
+// These both import the same `Qux`.
+// The visibility of `Qux` is `pub` because that is the maximum
+// visibility between these two `use` declarations.
+use foo::*;
+pub use bar::*;
+# fn main() {}
+```
+
+If an in-scope item has the same name as a crate name in the [extern prelude], then `use` of that crate name requires a leading `::` to unambiguously select the crate name or `crate::` to use the item from the crate root.
```
mattheww, line 381: For the record: this section is no longer true as of [Rust 1.72][1].

The ambiguity is now resolved in favour of the in-scope item.

[1]:https://github.com/rust-lang/rust/pull/112086

ehuss: Thanks, resolved in https://github.com/rust-lang/reference/pull/1052/commits/325b14ee07067f42bebee8d0816ee205516c4045 by just removing this.

ehuss:
@traviscross I have rebased, and pushed a small fix for this. I'm not certain what else has changed or is missing since it was written, though. I though I had some additional notes somewhere, and if I find them I'll follow up later.

src/items/use-declarations.md
```diff
@@ -201,8 +315,88 @@ m!(use std as _;);
...
+  use TypeAlias::MyVariant; // ERROR
+  ```
+
+## Ambiguities
+
+> **Note**: This section is incomplete.
+
+Some situations are an error when there is an ambiguity as to which name a `use` declaration refers to, when there are two name candidates that do not resolve to the same entity.
+
+Glob imports are allowed to import conflicting names in the same namespaces as long as the name is not used or shadowed.
```
traviscross, line 343: What's meant by *or shadowed* here?  E.g., this works:

```rust
mod foo {
    pub struct Qux;
}

mod bar {
    pub struct Qux;
}

use bar::*;
use foo::*; // Ok, no name conflict.

struct Qux; // Shadows imports.

fn main() {}
```

ehuss: Oof, I don't remember. It's possible I meant the case exactly what you specify is OK. To reword it as "...as long as the name is shadowed or not used"? However, that doesn't really make much sense (it isn't "used" when it is shadowed, either).

Another possibility is that I got `AmbiguityKind::GlobVsOuter` confused here. But I believe that restriction is a bit more complicated. See these examples:

https://github.com/rust-lang/rust/blob/HEAD/tests/ui/imports/glob-shadowing.rs
https://github.com/rust-lang/rust/blob/HEAD/tests/ui/imports/issue-57539.rs
https://github.com/rust-lang/rust/blob/HEAD/tests/ui/imports/shadow_builtin_macros.rs
https://github.com/rust-lang/rust/blob/HEAD/tests/ui/macros/macro-path-prelude-shadowing.rs

Perhaps it is best to just delete "or shadowed" for now, and follow up on completing this section later?

traviscross: Done.

src/items/enumerations.md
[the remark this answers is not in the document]
petrochenkov: The value namespace component for braced variants was removed in https://github.com/rust-lang/rust/pull/103578.

petrochenkov (approved):

petrochenkov:
LGTM except for the braced variant constructor detail - https://github.com/rust-lang/reference/pull/1052#discussion_r1691628631.

src/items/enumerations.md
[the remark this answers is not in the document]
ehuss: Thanks!

traviscross (approved):

PR #1054, opened 2021-07-04
Mention (negative) infinity values on float-to-int casting

JohnTitor:
A follow-up to https://github.com/rust-lang/nomicon/pull/256.


ehuss (approved):
Thanks!

PR #1055, opened 2021-07-04
Update documentation for arbitrary_enum_discriminant feature

fee1-dead:
Continuation of #639. This should only be merged after the feature is stabilized.

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -59,7 +59,7 @@ In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
 called an enum variant.
 
 An enum where no constructors contain fields are called a
-*<a id="field-less-enum">field-less enum</a>*. For example, this is a fieldless enum:
+*<span id="field-less-enum">field-less enum</span>*. For example, this is a fieldless enum:
 
 ```rust
 enum Fieldless {
@@ -70,7 +70,7 @@ enum Fieldless {
 ```
 
 If a field-less enum only contains unit variants, the enum is called an
-*<a id="unit-only-enum">unit-only enum</a>*. For example:
+*<span id="unit-only-enum">unit-only enum</span>*. For example:
 
 ```rust
 enum Enum {
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -59,7 +59,7 @@ In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
 called an enum variant.
 
 An enum where no constructors contain fields are called a
-*<a name="field-less-enum">field-less enum</a>*.
+*<a id="field-less-enum">field-less enum</a>*.
 
 ## Discriminants
 
@@ -77,36 +77,30 @@ another means of distinguishing variants) in its actual memory layout.
 In two circumstances, the discriminant of a variant may be explicitly set by
 following the variant name with `=` and a [constant expression]:
 
-<ol>
-<li>
 
-if the enumeration is "C-like" (i.e., it has no tuple or struct variants); e.g.:
+1. if the enumeration is fieldless; e.g.:
 
-```rust
-enum Enum {
-    Foo = 3,
-    Bar = 2,
-    Baz = 1,
-}
-```
-</li>
-<li>
+   ```rust
+   enum Enum {
+       Foo = 3,
+       Bar() = 2,
+       Baz {} = 1,
+   }
+   ```
 
-if a [primitive representation] is used. For example:
+2. if a [primitive representation] is used. For example:
 
-```rust
-#[repr(u8)]
-enum Enum {
-    Unit = 3,
-    Tuple(u16),
-    Struct {
-        a: u8,
-        b: u16,
-    } = 1,
-}
-```
-</li>
-</ol>
+   ```rust
+   #[repr(u8)]
+   enum Enum {
+       Unit = 3,
+       Tuple(u16),
+       Struct {
+           a: u8,
+           b: u16,
+       } = 1,
+   }
+   ```
 
 #### Implicit Discriminants
 
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -13,8 +13,8 @@
 >
 > _EnumItem_ :\
 > &nbsp;&nbsp; _OuterAttribute_<sup>\*</sup> [_Visibility_]<sup>?</sup>\
-> &nbsp;&nbsp; [IDENTIFIER]&nbsp;( _EnumItemTuple_ | _EnumItemStruct_
->                                | _EnumItemDiscriminant_ )<sup>?</sup>
+> &nbsp;&nbsp; [IDENTIFIER]&nbsp;( _EnumItemTuple_ | _EnumItemStruct_ )<sup>?</sup>
+>                                _EnumItemDiscriminant_<sup>?</sup>
 >
 > _EnumItemTuple_ :\
 > &nbsp;&nbsp; `(` [_TupleFields_]<sup>?</sup> `)`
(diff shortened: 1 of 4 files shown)


src/items/enumerations.md
```diff
@@ -56,22 +56,64 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
 In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
-called an enum variant. Each enum instance has a _discriminant_ which is an
-integer associated to it that is used to determine which variant it holds. An
-opaque reference to this discriminant can be obtained with the
-[`mem::discriminant`] function.
+called an enum variant.
 
-## Custom Discriminant Values for Fieldless Enumerations
+An enum where no constructors contain fields are called a
+*<a name="field-less-enum">field-less enum</a>*.
```
ehuss, line 62:
```suggestion
*<a id="field-less-enum">field-less enum</a>*.
```

The other form breaks our linkchecker.

src/items/enumerations.md
```diff
@@ -56,22 +56,64 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+
+#### Explicit Discriminants
+
+In two circumstances, the discriminant of a variant may be explicitly set by
+following the variant name with `=` and a [constant expression]:
+
+<ol>
+<li>
+
+if the enumeration is "C-like" (i.e., it has no tuple or struct variants); e.g.:
```
ehuss, line 83:
```suggestion
if the enumeration is fieldless (i.e., it has no tuple or struct variants); e.g.:
```

src/items/enumerations.md
```diff
@@ -56,22 +56,64 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+
+### Assigning Discriminant Values
+
+#### Explicit Discriminants
+
+In two circumstances, the discriminant of a variant may be explicitly set by
+following the variant name with `=` and a [constant expression]:
+
+<ol>
+<li>
```
ehuss, line 81: I think the HTML can be removed by just putting numbers in front of the items, and indenting the code blocks at least 3 characters.

ehuss (commented):
Thanks, this looks great! 🎉 

There's another link in `const_eval.md` that references `#custom-discriminant-values-for-fieldless-enumerations` that needs to be updated.

src/items/enumerations.md
fee1-dead: This is wrong because the compiler would tell me to add a repr for the enum.

ehuss: I don't understand what this has to do with the representation.  We have been moving the reference to use the term "fieldless enum" instead of "c-like"; that is the terminology we would prefer.

fee1-dead: True, but only c-like enums are allowed to have a discriminant without a `#[repr()]` attribute.

ehuss: Sorry, I'm still confused.  The term "c-like enum" has historically been used for Rust enums that only has unit variants (that is, variants without fields).  We have changed the terminology in the reference to use the phrase "fieldless enum" instead.  We want to stop using the phrase "c-like" altogether.  Indeed, this particular section is showing a fieldless enum.  Why do you still want to use the term "c-like" here?

fee1-dead: The following "c-like enum" does not need a repr attribute: (this is also "fieldless")
```rust
enum NoRepr {
    Foo = 5,
    Bar = 3,
    Baz = 1,
}
```

While the following "fieldless enum" (but not "c-like") **needs** a `repr` attribute:

```rust
#[repr(u8)] // MUST do this
enum B {
    Foo = 5,
    Bar() = 3, 
    Baz { } = 1,
}
```

Is this intended by the RFC? I think we might allow these kinds of "fieldless" but not "c-like" enums like this without a `repr` attribute.

cc @jswrenn @nox

ehuss: Hm, I can see how the literal interpretation of "fieldless" could be confused with those.  The two examples with brackets define tuple and struct variants, they just have 0 fields.  The unit variant has a distinctly different meaning which has different semantics.  As defined here in the reference, a "fieldless enum" is one with only unit variants (your first example).

fee1-dead: I don't think that is correct. The reference says 

> An enum where no constructors contain fields are called a
*<a id="field-less-enum">field-less enum</a>*.

ehuss: Well, this PR is the one adding that definition, no?  I guess my review could be more thorough here, but if you want to make that statement more precise, I would reword it to be that a fieldless enum is one that only has unit variants.  I would also drop the hyphen.  AFAIK, the language has always generally treated unit variants different from struct and tuple variants, even if the later two don't have fields.

fee1-dead: all "fieldless" enums can be cast to their discriminant values. however only c-like enums may specify discriminants without a repr. @jswrenn has already said this on the original PR: https://github.com/rust-lang/reference/pull/639#issuecomment-511515290

fee1-dead: Opened https://github.com/rust-lang/rust/pull/88203. This should be made consistent and I see no problems for fieldless enums to have an unspecified repr.

src/items/enumerations.md
```diff
@@ -56,22 +56,58 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
+
+### Assigning Discriminant Values
+
+#### Explicit Discriminants
+
+In two circumstances, the discriminant of a variant may be explicitly set by
+following the variant name with `=` and a [constant expression]:
+
+
+1. if the enumeration is "C-like" (i.e., it has no tuple or struct variants); e.g.:
```
ehuss, line 81: I feel like we tried quite hard to scrub the presence of the "C-like" terminology, and this PR seems to bring it back. Would it be possible to come up with some other terminology here? Or, perhaps it doesn't need an explicit term, but could just say "if the enumeration only has Unit variants…"?

src/items/enumerations.md
```diff
@@ -122,6 +155,59 @@ enum OverflowingDiscriminantError2 {
...
+#### Via `mem::discriminant`
+
+[`mem::discriminant`] returns an opaque reference to the discriminant of
+an enum value which can be compared. This cannot be used to get the value
+of the discriminant.
+
+#### Casting
+
+If an enumeration is C-like (with no tuple and struct variants), then its
+discriminant can be directly accessed with a [numeric cast]; e.g.:
```
ehuss, line 169: This seems to be missing the grandfathered support for casting enums with empty constructors, like:

```rust
enum Fieldless {
    Tuple(),
    Struct{},
    Unit,
}

assert_eq!(Fieldless::Tuple() as u8, 0);
assert_eq!(Fieldless::Struct{} as u8, 1);
assert_eq!(Fieldless::Unit as u8, 2);
```

Was that excluded for a reason?

src/items/enumerations.md
```diff
@@ -56,22 +56,58 @@ a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
...
 In this example, `Cat` is a _struct-like enum variant_, whereas `Dog` is simply
-called an enum variant. Each enum instance has a _discriminant_ which is an
-integer associated to it that is used to determine which variant it holds. An
-opaque reference to this discriminant can be obtained with the
-[`mem::discriminant`] function.
+called an enum variant.
 
-## Custom Discriminant Values for Fieldless Enumerations
+An enum where no constructors contain fields are called a
+*<a id="field-less-enum">field-less enum</a>*.
```
ehuss, line 62: I think the field-less term got a little confusing, and it's still not clear to me, does it also cover empty constructors, like:

```rust
enum Fieldless {
    Tuple(),
    Struct{},
    Unit,
}
```

If so, perhaps that needs to be made explicit?

ehuss (commented):
Returning to this now that it has landed. I think the structure and content look great. I just want to nail down the details. Other than the comments I added, is this otherwise up to date with what eventually finalized?

src/items/enumerations.md
fee1-dead: The problem is that they cannot be casted to their discriminant if their discriminants are specified. This is summarized in https://github.com/rust-lang/rust/issues/60553#issuecomment-985007785. This will likely need more discussion around the design, but none of the design concerns block the stabilization of arbitrary_enum_discriminant.

If you would like this to be included, then I can put that in.

src/items/enumerations.md
fee1-dead: Yeah, maybe we can use `Unit-only` and link to a section when referring to it somewhere else. Will do so in a moment.

src/items/enumerations.md
fee1-dead: Yeah, I will add that as an example.

src/items/enumerations.md
```diff
@@ -122,6 +167,59 @@ enum OverflowingDiscriminantError2 {
...
+#[repr(u8)]
+enum Enum {
+    Unit,
+    Tuple(bool),
+    Struct{a: bool},
+}
+
+impl Enum {
+    fn discriminant(&self) -> u8 {
+        unsafe { *(self as *const Self as *const u8) }
```
scottmcm, line 210: This doesn't look right to me?  Just because it's `repr(u8)` I don't think that means that the discriminant is necessarily *first*.

If the `enum` were `repr(C)` or `repr(C, u8)` *then* it'd be ok because the field order would be specified.  But for `repr(Rust)` I'm pretty sure we're allowed to put the discriminant *after* the payload, if we wanted to.

scottmcm: Oh, never mind, it looks like <https://rust-lang.github.io/rfcs/2195-really-tagged-unions.html#guide-level-explanation> means that `repr(u8)` disables all field reordering too, not just sets the discriminant size :(

tmandry: I found that odd, too. The RFC justifies this by saying the discriminant position needs to be deterministic, which seems possible to do without disabling other field reordering?

tmandry: (At this point it seems too late to change, but an edition change could require `repr(C)` for deterministic field ordering.)

scottmcm: That would need another flag to distinguish between the union-of-structs and struct-containing-union-of-structs layouts, though.  Which might be nice, though!

I suppose the argument is that the size of the discriminant can't matter to safe code, so the only point of ever specifying its width is to get deterministic layout?

ehuss (approved):
I went ahead and pushed some edits to fix some issues, but also to add the grandfathered behavior of allowing `as` casting for fieldless enums. If anyone feels that it should be different, feel free to open a PR.

ehuss:
Thanks again fee1-dead, and sorry for the super long delay.

PR #1056, opened 2021-07-05
fix comment in function parameter drop scope example

tlyu:
In "Scopes of function parameters" under "Destructors", the first
comment in the example that demonstrates function parameter
drop order contradicted both the text and the program behavior.


ehuss (approved):
Thanks!

PR #1057, opened 2021-07-05
fix grammar in Expressions

tlyu:
Fix minor grammar errors in Expressions, mostly adjusting
punctuation, but also fixing missing or incorrect words.


ehuss (approved):
Thanks!

PR #1058, opened 2021-07-05
fix typo in macro-ambiguity.md

yume-chan:
Checked https://raw.githubusercontent.com/rust-lang/rfcs/master/text/0550-macro-future-proofing.md, there is no `\`s in that file.


ehuss (approved):
Thanks!

PR #1059, opened 2021-07-06
Update rust reference info about closures

### LINKED ISSUE #1066: 2021: Update for disjoint captures (RFC 2229)
The closure type page needs to be updated for the 2021 changes. See [RFC 2229](https://github.com/rust-lang/rfcs/blob/master/text/2229-capture-disjoint-fields.md) and [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/disjoint-capture-in-closures.md).

(Note: PR #1059 is already open to resolve this.)

roxelo:
This PR updates the closure types page so it includes up to date information on the new behavior introduced by RFC2229/Edition 2021

cc: @arora-aman
r? @nikomatsakis 

Closes #1066

### DIFF
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -317,7 +317,7 @@ If a closure captures a field of a composite types such as structs, tuples, and 
 * Input:
     * Analyzing the closure C yields a mapping of `Place -> Mode` that are accessed
     * Access mode is `ref`, `ref uniq`, `ref mut`, or `by-value` (ordered least to max)
-        * For a `Place` that is used in two different acess modes within the same closure, the mode reported from closure analysis is the maximum access mode.
+        * For a `Place` that is used in two different access modes within the same closure, the mode reported from closure analysis is the maximum access mode.
         * Note: `ByValue` use of a `Copy` type is seen as a `ref` access mode.
     * Closure mode is `ref` or `move`
 * Output:
@@ -383,6 +383,8 @@ If a closure captures a field of a composite types such as structs, tuples, and 
 
 ### box-mut
 
+This test shows how a `move` closure can sometimes capture values by mutable reference, if they are reached via a `&mut` reference.
+
 ```rust
 struct Foo { x: i32 }
 
@@ -405,10 +407,10 @@ fn box_mut() {
 <!-- ignore: Omit error about unterminated string literal when representing c_prime -->
 ```ignore
 Closure mode = move
-C = {
+C_in = {
     (ref mut, (*(*bx)).x)
 }
-C' = C
+C_out = C_in
 ```
 
 Output is the same: `C' = C`
@@ -439,13 +441,16 @@ fn main() {
 <!-- ignore: Omit error about unterminated string literal when representing c_prime -->
 ```ignore
 Closure mode = ref
-C = {
+C_in = {
     (ref mut, packed)
 }
-C' = C
+C_out = C_in
 ```
 
 ### Optimization-Edge-Case
+
+This test shows an interesting edge case. Normally, when we see a borrow of something behind a shared reference (`&T`), we truncate to capture the entire reference, because that is more efficient (and we can always use that reference to reach all the data it refers to). However, in the case where we are dereferencing two shared references, we have to be sure to preserve the full path, since otherwise the resulting closure could have a shorter lifetime than is necessary.
+
 ```edition2021
 struct Int(i32);
 struct B<'a>(&'a i32);
@@ -465,10 +470,10 @@ fn foo<'a, 'b>(m: &'a MyStruct<'b>) -> impl FnMut() + 'static {
 <!-- ignore: Omit error about unterminated string literal when reprenting c_prime -->
 ```ignore
 Closure mode = ref
-C = {
+C_in = {
     (ref mut, *m.a)
 }
-C' = C
+C_out = C_in
 ```
 
 # Edition 2018 and before
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -312,6 +312,139 @@ If a closure captures a field of a composite types such as structs, tuples, and 
 } // tuple.1 dropped here -----------------------------+
 ```
 
+
+## Overall Capture analysis algorithm
+
+* Input:
+    * Analyzing the closure C yields a set of `(Mode, Place)` pairs that are accessed
+    * Access mode is `ref`, `ref uniq`, `ref mut`, or `by-value` (ordered least to max)
+    * Closure mode is `ref` or `move`
+* Output:
+    * Minimal `(Mode, Place)` pairs that are actually captured
+* Cleanup and truncation
+    * Generate C' by mapping each (Mode, Place) in C:
+        * `(Mode1, Place1) = ref_opt(unsafe_check(copy_type(Mode, Place)))`
+        * if this is a ref closure:
+            * Add `ref_xform(Mode1, Place1)` to C'
+        * else:
+            * Add `move_xform(Mode1, Place1)` to C'
+* Minimization
+    * Until no rules apply:
+        * For each two places (M1, P1), (M2, P2) where P1 is a prefix of P2:
+            * Remove both places from the set
+            * Add (max(M1, M2), P1) into the set
+* Helper functions:
+    * `copy_type(Mode, Place) -> (Mode, Place)`
+        * "By-value use of a copy type is a ref"
+        * If Mode = "by-value" and type(Place) is `Copy`:
+            * Return (ref, Place)
+        * Else
+            * Return (Mode, Place)
+    * `unsafe_check(Mode, Place) -> (Mode, Place)`
+        * "Ensure unsafe accesses occur within the closure"
+        * If Place contains a deref of a raw pointer:
+            * Let Place1 = Place truncated just before the deref
+            * Return (Mode, Place1)
+        * If Mode is `ref *` and the place contains a field of a packed struct:
+            * Let Place1 = Place truncated just before the field
+            * Return (Mode, Place1)
+        * Else
+            * Return (Mode, Place1)
+    * `move_xform(Mode, Place) -> (Mode, Place)` (For move closures)
+        * "Take ownership if data being accessed is owned by the variable used to access it (or if closure attempts to move data that it doesn't own)."
+        * "When taking ownership, only capture data found on the stack."
+        * "Otherwise, reborrow the reference."
+        * If Mode is `ref mut` and the place contains a deref of an `&mut`:
+            * Return (Mode, Place)
+        * Else if Mode is `ref *` and the place contains a deref of an `&`:
+            * Return (Mode, Place)
+        * Else if place contains a deref:
+            * Let Place1 = Place truncated just before the deref
+            * Return (ByValue, Place1)
+        * Else:
+            * Return (ByValue, Place)
+    * `ref_xform(Mode, Place) -> (Mode, Place)` (for ref closures)
+        * "If taking ownership of data, only move data from enclosing stack frame."
+        * Generate C' by mapping each (Mode, Place) in C
+            * If Mode is ByValue and place contains a deref:
+                * Let Place1 = Place truncated just before the deref
+                * Return (ByValue, Place1)
+            * Else:
+                * Return (Mode, Place)
+    * `ref_opt(Mode, Place) -> (Mode, Place)` (for ref closures)
+        * "Optimization: borrow the ref, not data owned by ref."
+        * If Place contains a deref of an `&`...
+            * ...or something
+
+## Key examples
+
+### box-mut
+
+```rust
+fn box_mut() {
+    let mut s = Foo { x: 0 } ;
+    
+    let px = &mut s;
+    let bx = Box::new(px);
+    
+    
+    let c = #[rustc_capture_analysis] move || bx.x += 10;
+    // Mutable reference to this place:
+    //   (*(*bx)).x
+    //    ^ ^
+    //    | a Box
+    //    a &mut
+}
+```
+
+```
+Closure mode = move
+C = {
+    (ref mut, (*(*bx)).x)
+}
+C' = C
+```
+
+Output is the same: `C' = C`
+
+### Packed-field-ref-and-move
+
+When you have a closure that both references a packed field (which is unsafe) and moves from it (which is safe) we capture the entire struct, rather than just moving the field. This is to aid in predictability, so that removing the move doesn't make the closure become unsafe:
+
+```rust
+print(&packed.x);
+move_value(packed.x);
+```
+
+
+```rust
+struct Point { x: i32, y: i32 }
+fn f(p: &Point) -> impl Fn() {
+    let c = move || {
+      let x = p.x; 
+    }; 
+    
+    // x.x -> ByValue
+    // after rules x -> ByValue
+
+    c
+} 
+
+struct Point { x: i32, y: i32 }
+fn g(p: &mut Point) -> impl Fn() {
+    let c = move || {
+      let x = p.x; // ought to: (ref, (*p).x)
+    };
+    
+    move || {
+       p.y += 1;
+    }
+    
+    
+    // x.x -> ByValue
+   
+```
+
 # Edition 2018 and before
 
 ## Closure types difference
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -76,14 +76,15 @@ In the case where a path and one of the ancestor’s of that path are both captu
 
 Note that this might need to be applied recursively.
 
-```rust=
-let s = String::new("S");
-let t = (s, String::new("T"));
-let mut u = (t, String::new("U"));
+```rust
+# fn move_value<T>(_: T){}
+let s = String::from("S");
+let t = (s, String::from("T"));
+let mut u = (t, String::from("U"));
 
 let c = || {
     println!("{:?}", u); // u captured by ImmBorrow
-    u.0.truncate(0); // u.0 captured by MutBorrow
+    u.1.truncate(0); // u.0 captured by MutBorrow
     move_value(u.0.0); // u.0.0 captured by ByValue
 };
 ```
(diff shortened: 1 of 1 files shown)


src/types/closure.md
```diff
@@ -178,3 +289,98 @@ Because captures are often by reference, the following general rules arise:
...
+```
+and the call to `f` would work as follows:
+<!-- ignore: continuation of above -->
+```rust,ignore
+f(Closure { rect: rect });
+```
+
+## Capture precision difference
+
+Composite types such as structs, tuples, and enums are always captured in its intirety,
```
ehuss, line 339:
```suggestion
Composite types such as structs, tuples, and enums are always captured in its entirety,
```

ehuss (commented):
Thanks!  Just some minor editorial comments. I'll let Niko review the actual content. 😉

src/types/closure.md
```diff
@@ -42,48 +50,153 @@ so that the call to `f` works as if it were:
...
+struct S(i32);
+
+let b = Box::new(S(10));
+let c_box = || {
+    println!("{}", (*b).0); // closure captures `(*b).0`
+};
+
+let r = Rc::new(S(10));
+let c_rc = || {
+    println!("{}", (*r).0); // closure caprures `r`
```
nikomatsakis, line 173:
```suggestion
    println!("{}", (*r).0); // closure captures `r`
```

nikomatsakis (changes requested):
The content is good. It'd be nice to also add in the "formal algorithm" that we came up with in our last meeting.

src/types/closure.md
```diff
@@ -316,65 +316,69 @@ If a closure captures a field of a composite types such as structs, tuples, and
...
 * Input:
-    * Analyzing the closure C yields a set of `(Mode, Place)` pairs that are accessed
+    * Analyzing the closure C yields a mapping of `Place -> Mode` that are accessed
     * Access mode is `ref`, `ref uniq`, `ref mut`, or `by-value` (ordered least to max)
+        * For a `Place` that is used in two different acess modes within the same closure, the mode reported from closure analysis is the maximum access mode.
+        * Note: `ByValue` use of a `Copy` type is seen as a `ref` access mode.
     * Closure mode is `ref` or `move`
 * Output:
-    * Minimal `(Mode, Place)` pairs that are actually captured
+    * Minimal `(Place, Mode)` pairs that are actually captured
```
arora-aman, line 325: Made the change to (Place, Mode) to reflect the way information is stored and passed around in code

src/types/closure.md
```diff
@@ -316,65 +316,69 @@ If a closure captures a field of a composite types such as structs, tuples, and
 ## Overall Capture analysis algorithm
 
 * Input:
-    * Analyzing the closure C yields a set of `(Mode, Place)` pairs that are accessed
+    * Analyzing the closure C yields a mapping of `Place -> Mode` that are accessed
```
arora-aman, line 319: Add details for how things are currently stored in the initial analysis phase

src/types/closure.md
```diff
@@ -316,65 +316,69 @@ If a closure captures a field of a composite types such as structs, tuples, and
...
+            * Add `move_xform(Place1, Mode1)` to C'
 * Minimization
     * Until no rules apply:
-        * For each two places (M1, P1), (M2, P2) where P1 is a prefix of P2:
+        * For each two places (P1, M1), (P2, M2) where P1 is a prefix of P2:
             * Remove both places from the set
-            * Add (max(M1, M2), P1) into the set
+            * Add (P1, max(M1, M2)) into the set
 * Helper functions:
-    * `copy_type(Mode, Place) -> (Mode, Place)`
```
arora-aman: This is not a concern of the capture analysis anymore

src/types/closure.md
```diff
@@ -316,65 +316,69 @@ If a closure captures a field of a composite types such as structs, tuples, and
...
         * "Optimization: borrow the ref, not data owned by ref."
-        * If Place contains a deref of an `&`...
-            * ...or something
+        * Disjoint capture over immutable reference doesn't add too much value because the fields can either be borrowed immutably or copied.
+        * Edge case: Field that is accessed via the referece lives longer than the reference.
+            * Resolution: Only consider the last Deref
+        * If Place is (Base, Projections), where Projections is a list of size N.
+            * For all `i, 0 <= i < N`, Projections[i] != Deref
+                * Return (Place, Mode)
+            * If `l, 0 <= l < N` is the last/rightmost Deref Projection i.e. for any `i, l < i < N` Projection[i] != Deref,
```
arora-aman, line 378: Not sure if this became too mathematical  😅

src/types/closure.md
```diff
@@ -412,37 +416,55 @@ Output is the same: `C' = C`
...
+   b: B<'a>,
+}
+
+fn foo<'a, 'b>(m: &'a MyStruct<'b>) -> impl FnMut() + 'static {
+    let c = || drop(&m.a.0);
     c
-} 
+}
 
-struct Point { x: i32, y: i32 }
```
arora-aman: These examples were us discussiong proper result of what analysis should do in moving copy types and working through some exampels

src/types/closure.md
```diff
@@ -178,3 +294,258 @@ Because captures are often by reference, the following general rules arise:
...
+Closure mode = ref
+C = {
+    (ref mut, packed)
+}
+C' = C
+```
+
+### Optimization-Edge-Case
+```edition2021
+struct Int(i32);
```
arora-aman, line 451: @ehuss  @nikomatsakis should I just make this ignore since tests seem to not allow edition 2021 since it's unstable?

src/types/closure.md
```diff
@@ -316,79 +316,85 @@ If a closure captures a field of a composite types such as structs, tuples, and
...
-            * Return (Mode, Place1)
+            * Return (Place1, Mode)
         * If Mode is `ref *` and the place contains a field of a packed struct:
             * Let Place1 = Place truncated just before the field
-            * Return (Mode, Place1)
+            * Return (Place1, Mode)
         * Else
-            * Return (Mode, Place1)
-    * `move_xform(Mode, Place) -> (Mode, Place)` (For move closures)
+            * Return (Place, Mode)
```
arora-aman, line 348: This I changed ~~Place1~~ Place, because i think it was a typo in the algorithm we wrote on Hackmd

src/types/closure.md
```diff
@@ -42,48 +50,150 @@ so that the call to `f` works as if it were:
...
 followed by unique immutable borrow (see below), by mutable borrow, and finally
-by move. It will pick the first choice of these that is compatible with how the
-captured variable is used inside the closure body. The compiler does not take
-surrounding code into account, such as the lifetimes of involved variables, or
-of the closure itself.
+by move. It will pick the first choice of these that allows the closure to
+compile. The choice is made only with regards to the contents of the closure
+expression; the compiler does not take into account surrounding code, such as
+the lifetimes of involved variables or fields.
+>>>>>>> 881f305... Update closure types documentation so it includes information about RFC2229
```
lqd, line 64: a stray merge conflict marker ?

roxelo: Fixed

src/types/closure.md
nikomatsakis: seems good to me :)

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+} // tuple.1 dropped here -----------------------------+
+```
+
+
+## Overall Capture analysis algorithm
+
+* Input:
+    * Analyzing the closure C yields a mapping of `Place -> Mode` that are accessed
+    * Access mode is `ref`, `ref uniq`, `ref mut`, or `by-value` (ordered least to max)
+        * For a `Place` that is used in two different acess modes within the same closure, the mode reported from closure analysis is the maximum access mode.
```
nikomatsakis, line 320:
```suggestion
        * For a `Place` that is used in two different access modes within the same closure, the mode reported from closure analysis is the maximum access mode.
```

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+    } // 'c' and 'tuple.0' dropped here ------------+  |
+} // tuple.1 dropped here -----------------------------+
+```
+
+
+## Overall Capture analysis algorithm
+
+* Input:
+    * Analyzing the closure C yields a mapping of `Place -> Mode` that are accessed
+    * Access mode is `ref`, `ref uniq`, `ref mut`, or `by-value` (ordered least to max)
```
nikomatsakis, line 319: I'm still a bit unclear on the best way to incorporate ref uniq here -- my sense is that the results that come in initially should never have ref uniq, but it gets determined later, whenever we do a truncation through a `*` of an `&mut` (as we talked about).

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+            * Resolution: Only consider the last Deref
+        * If Place is (Base, Projections), where Projections is a list of size N.
+            * For all `i, 0 <= i < N`, Projections[i] != Deref
+                * Return (Place, Mode)
+            * If `l, 0 <= l < N` is the last/rightmost Deref Projection i.e. for any `i, l < i < N` Projection[i] != Deref,
+              and `Place.type_before_projection(l) = ty::Ref(.., Mutability::Not)`
+                * Let Place1 = (Base, Projections[0..=l])
+                * Return (Place1, Ref)
+
+## Key examples
```
nikomatsakis, line 382: These are pretty inscrutable if you didn't attend our meeting, I think

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+    
+    
+    let c = move || bx.x += 10;
+    // Mutable reference to this place:
+    //   (*(*bx)).x
+    //    ^ ^
+    //    | a Box
+    //    a &mut
+}
+```
```
nikomatsakis, line 403: Probably we should give the initial and final results. Something like

```
C_in  = { (ref mut, (*(*bx)).x) }
C_out = C_in
```

src/types/closure.md
nikomatsakis: By the time this lands, they will, I believe

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+            * For all `i, 0 <= i < N`, Projections[i] != Deref
+                * Return (Place, Mode)
+            * If `l, 0 <= l < N` is the last/rightmost Deref Projection i.e. for any `i, l < i < N` Projection[i] != Deref,
+              and `Place.type_before_projection(l) = ty::Ref(.., Mutability::Not)`
+                * Let Place1 = (Base, Projections[0..=l])
+                * Return (Place1, Ref)
+
+## Key examples
+
+### box-mut
```
nikomatsakis, line 384: Let's add a comment here:

This test shows how a `move` closure can sometimes capture values by mutable reference, if they are reached via a `&mut` reference.

src/types/closure.md
```diff
@@ -178,3 +293,258 @@ Because captures are often by reference, the following general rules arise:
...
+<!-- ignore: Omit error about unterminated string literal when representing c_prime -->
+```ignore
+Closure mode = ref
+C = {
+    (ref mut, packed)
+}
+C' = C
+```
+
+### Optimization-Edge-Case
```
nikomatsakis, line 448: Let's add a comment here:

This test shows an interesting edge case. Normally, when we see a borrow of something behind a shared reference (`&T`), we truncate to capture the entire reference, because that is more efficient (and we can always use that reference to reach all the data it refers to). However, in the case where we are dereferencing *two* shared references, we have to be sure to preserve the full path, since otherwise the resulting closure could have a shorter lifetime than is necessary.

nikomatsakis (changes requested):
This is very nice. I left some comments.

src/types/closure.md
```diff
@@ -2,38 +2,46 @@
 
 A [closure expression] produces a closure value with a unique, anonymous type
 that cannot be written out. A closure type is approximately equivalent to a
-struct which contains the captured variables. For instance, the following
+struct which contains the captured values. For instance, the following
 closure:
 
 ```rust
+#[derive(Debug)]
+struct Point { x:i32, y:i32 }
```
joshtriplett, line 10:
```suggestion
struct Point { x: i32, y: i32 }
```

ehuss:
@roxelo / @arora-aman / @nikomatsakis Is there an update on how this is going?  It looks like niko left some comments that may not have been addressed?

arora-aman:
> @roxelo / @arora-aman / @nikomatsakis Is there an update on how this is going? It looks like niko left some comments that may not have been addressed?

@ehuss I think they have been addressed, github didn't auto resolve them because lines were added instead of them being changed.

There was one style change that josh left that I have now committed. Hopefully tests pass given Edition 2021 has been stabilized

arora-aman:
I'll fix the style errors hopefully tomorrow.

roxelo:
Just saw this now, can help if needed @arora-aman

src/types/closure.md
```diff
@@ -42,48 +50,153 @@ so that the call to `f` works as if it were:
...
+    _ => println!("Hello World!")
+};
 ```
 
-If, instead, the closure were to use `self.vec` directly, then it would attempt
-to capture `self` by mutable reference. But since `self.set` is already
-borrowed to iterate over, the code would not compile.
+### Capturing references in move contexts
+
+Moving fields out of references is not allowed. As a result, in the case of move closures, when values accessed through a shared references are moved into the closure body, the compiler will truncate right before a dereference.
```
shepmaster, line 109: I would love to understand the logic behind 

> Moving fields out of references is not allowed

Where could I learn more?

src/types/closure.md
```diff
@@ -42,48 +50,153 @@ so that the call to `f` works as if it were:
...
+
+```rust
+let x = 10;
+let c = || {
+    let _ = x;
+};
+
+let c = || match x {
+    _ => println!("Hello World!")
+};
```
Veykril, line 104: This seems to be wrong. Both of these examples are captured by value today -> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4044e24fa1cffb795c2dd03239eb5fb2

src/types/closure.md
```diff
@@ -42,48 +50,153 @@ so that the call to `f` works as if it were:
...
+```rust
+struct S(i32);
+
+let b = Box::new(S(10));
+let c_box = || {
+    println!("{}", (*b).0); // closure captures `b`
+};
+```
 
 ## Unique immutable borrows in captures
```
HKalbasi, line 201: The example in this section needs to be updated, as it now captures `*x` by mutable borrow, not `x` by unique immutable borrow.

ehuss:
@Veykril or @HKalbasi Would either of you be interested in helping to get this PR over the line? Are either of you familiar with how they are implemented? I am not familiar with how closures are implemented in rustc. I could start reviewing all of the RFC 2229 PR's and reviewing all of the current implementation, but I have been dragging because that sounds like weeks of work. It would help to have someone else to ask questions to and to help with reviewing.

HKalbasi:
I guess I can help since I reimplemented this in rust-analyzer and got more or less familiar with the rustc implementation. What should I do?

ehuss:
Thanks! I'm not sure exactly how to make progress, but perhaps some questions might help to get started:

1. Have you carefully reviewed this PR? If not, would you be willing to give it a detailed examination? Are there any details missing?
2. Can you point me to a rough outline of the areas in rustc that I would need to look at to understand the text of this PR?
3. How difficult do you think it would be for me to figure it out? I'm moderately familiar with rustc, but not this area (nor type inference or type checking, which I'm not sure is relevant here). I'm not sure if I'm vastly over- or under-estimating how hard this is to review.

I can try to review the comment you left and push changes where I feel confident in them, but if I end up having questions it would be helpful to have someone to ask.

Hopefully there aren't too many gray areas where it isn't clear if some behavior was intended or not.

HKalbasi:
> Have you carefully reviewed this PR? If not, would you be willing to give it a detailed examination? Are there any details missing?

Not yet, but will do it in the following days when I found time.

> Can you point me to a rough outline of the areas in rustc that I would need to look at to understand the text of this PR?

The whole thing happens in [`expr_use_visitor.rs`](https://github.com/rust-lang/rust/blob/master/compiler/rustc_hir_typeck/src/expr_use_visitor.rs) and [`upvars.rs`](https://github.com/rust-lang/rust/blob/master/compiler/rustc_hir_typeck/src/upvar.rs). The entry point is the `analyze_closure` in `upvars.rs`, which calls `ExprUseVisitor::consume_body` to collect all of the places, and then aggregates them, applies some changes and optimizations, trims the places in old editions and computes the closure kind.

> How difficult do you think it would be for me to figure it out? I'm moderately familiar with rustc, but not this area (nor type inference or type checking, which I'm not sure is relevant here). I'm not sure if I'm vastly over- or under-estimating how hard this is to review.

It's pretty straightforward, specially if you are familiar with the `Place` concept. I think you won't have difficulties in finding the corresponding code for any part of text, but checking that if every small detail is covered might becomes a little difficult.

src/types/closure.md
```diff
@@ -42,48 +50,153 @@ so that the call to `f` works as if it were:
...
+
+```rust
+#[repr(packed)]
+struct T(String, String);
+
+let t = T(String::from("foo"), String::from("bar"));
+let c = || unsafe {
+    println!("{}", t.0); // closure captures t
+};
+```
```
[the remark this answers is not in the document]
HKalbasi: Yes the example is wrong and doesn't compile, but I'm not sure if your example is relevant here. I guess the `addr_of!` macro will always capture the whole item, since it is expanded into something like `packed as usize + some_offset`.

I think this example would work:
```Rust
#[repr(packed)]
struct T(i32, i32);

fn main() {
    let t = T(2, 5);
    let c = || {
        t.0; // closure captures t.0 if not packed, but t if packed
        t.1;
    };
    println!("{}", core::mem::size_of_val(&c))
}
```
Normally, `t.0` should be captured by immutable borrow (since `i32` is copy) but now that `t` is packed and taking `&t.0` is UB, the whole `t` will be captured and this code will print `8` instead of `16` when `T` is not packed.

ehuss:
I have posted https://github.com/rust-lang/reference/pull/1521 with some updates to this PR, since I am unable to push directly to this one.

ehuss:
Closing as this is now resolved by #1521.

PR #1060, opened 2021-07-07
move text about checking of bounds on generics

### LINKED ISSUE #1024: "Where clauses" subsection has paragraphs that belong in 10.6
In 6.14 (Generic parameters), the "Where clauses" subsection has these two paragraphs that probably belong in 10.6 (Trait and lifetime bounds).

> Bounds that don't use the item's parameters or higher-ranked lifetimes are checked when the item is defined. It is an error for such a bound to be false.
> 
> Copy, Clone, and Sized bounds are also checked for certain generic types when defining the item. It is an error to have Copy or Clone as a bound on a mutable reference, trait object or slice or Sized as a bound on a trait object or slice.

I think the quoted paragraphs apply more generally than to where clauses.

Alternatively, the "Where clauses" subsection could move to 10.6, because where clauses are arguably specifically about bounds, rather than about generic parameters in general. I think the two quoted paragraphs still belong outside any subsection that's specific to where clauses.

I'm willing to work on a pull request for this.

tlyu:
Move text about the timing of checking of bounds on generics
from the "Where clauses" subsection to "Trait and lifetime
bounds", where it makes more sense. Split parts of the example
accordingly. Correct an error about when `Clone`, `Copy`, and
`Sized` trait bounds are checked.

Fixes #1024.

I'm not sure if the previous text about checking trait bounds on generics at the definition instead of at the use site was ever correct, but it doesn't seem correct now on stable.

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -77,7 +77,7 @@ Bounds that don't use the item's parameters or [higher-ranked lifetimes] are che
 It is an error for such a bound to be false.
 
 [`Copy`], [`Clone`], and [`Sized`] bounds are also checked for certain generic types when using the item, even if the use does not provide a concrete type.
-It is an error to have `Copy` or `Clone` as a bound on a mutable reference, [trait object], or [slice][arrays].
+It is an error to have `Copy` or `Clone` as a bound on a mutable reference, [trait object], or [slice].
 It is an error to have `Sized` as a bound on a trait object or slice.
 
 ```rust,compile_fail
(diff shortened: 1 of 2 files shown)


src/trait-bounds.md
```diff
@@ -73,6 +73,26 @@ fn name_figure<U: Shape>(
 }
 ```
 
+Bounds that don't use the item's parameters or [higher-ranked lifetimes] are
+checked when the item is defined. It is an error for such a bound to be false.
```
JohnTitor, line 77: Seems you're turning the second paragraph into "one sentence per line" but the first doesn't.

```suggestion
Bounds that don't use the item's parameters or [higher-ranked lifetimes] are checked when the item is defined.
It is an error for such a bound to be false.
```

JohnTitor (approved):
Looks reasonable to me with one nit.

src/trait-bounds.md
tlyu: Thanks for the review! I pushed an update.

JohnTitor:
Thanks!

PR #1061, opened 2021-07-07
Mention std::is_aarch64_feature_detected

adamgemmell:
This macro has recently been updated and we hope for it to be stabilised soon.


ehuss (approved):
Thanks!

Looks like this was stabilized in rust-lang/rust#90271.  Links to stabilization PRs and stabilization reports can be helpful to understand what's going on.

PR #1062, opened 2021-07-08
fix wording/punctuation in "Lifetime bounds"

tlyu:
Clarify that "outlives" really means "lives at least as long as".
Also fix some more minor punctuation and wording issues.

Based on #1060.

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -77,7 +77,7 @@ Bounds that don't use the item's parameters or [higher-ranked lifetimes] are che
 It is an error for such a bound to be false.
 
 [`Copy`], [`Clone`], and [`Sized`] bounds are also checked for certain generic types when using the item, even if the use does not provide a concrete type.
-It is an error to have `Copy` or `Clone` as a bound on a mutable reference, [trait object], or [slice][arrays].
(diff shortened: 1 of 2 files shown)


src/trait-bounds.md
```diff
@@ -73,6 +73,26 @@ fn name_figure<U: Shape>(
 }
 ```
 
+Bounds that don't use the item's parameters or [higher-ranked lifetimes] are
+checked when the item is defined. It is an error for such a bound to be false.
+
+[`Copy`], [`Clone`], and [`Sized`] bounds are also checked for certain generic types when using the item, even if the use does not provide a concrete type.
+It is an error to have `Copy` or `Clone` as a bound on a mutable reference, [trait object], or [slice][arrays].
```
ehuss, line 80: I think we've added a slice chapter since this was written, can you update the link?

```suggestion
It is an error to have `Copy` or `Clone` as a bound on a mutable reference, [trait object], or [slice].
```

I think it would point to:

```
[slice]: types/slice.md

ehuss (commented):
Looks good, just one small change if you don't mind.

src/trait-bounds.md
tlyu: Thanks! Updated in the upstream branch (for #1060) and rebased.

ehuss (approved):
Thanks!

PR #1063, opened 2021-07-08
document implicit lifetime bounds

### LINKED ISSUE #289: Infered lifetime bounds (RFC 2093)
See rust-lang/rust#44493

- [ ] Document existing inferred lifetime bounds in functions and impls
- [ ] Document feature. Should this be documented with the existing inferred bounds?
- [ ] Make use of in examples

tlyu:
Add documentation about implicit lifetime bounds that the compiler
can infer for functions, trait implementations, and structs.
This draws from RFC 2093 (for structs) and RFC 1214 for functions
and trait implementations.

This is built on #1062 and #1060.

It does substantially increase the length of the page it's on, so I would be open to some reorganization.

Fixes #289.

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -119,93 +119,6 @@ fn f<'a, 'b>(x: &'a i32, mut y: &'b i32) where 'a: 'b {
 r[bound.lifetime.outlive-type]
 `T: 'a` means that all lifetime parameters of `T` outlive `'a`. For example, if `'a` is an unconstrained lifetime parameter, then `i32: 'static` and `&'static str: 'a` are satisfied, but `Vec<&'a ()>: 'static` is not.
 
-In the following text, the words "inferred" and "implied" refer to similar lifetime bounds that may be omitted.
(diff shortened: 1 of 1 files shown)


traviscross:
We talked about this on the lang docs call.  This is good work.

There are two things that we want to change before merging this.  One is that we'd like the verbiage to state the rules more directly rather than leaning so heavily on the examples.  The other is that we'd like to roll this into the section on implied bounds.

tlyu:
I've done a rebase. I'm still refreshing my memory about the subject matter after a few years away, so I'll address the review comments once I've done that.

tlyu:
It looks like #1261 added a lot of overlapping content. I'm going to see how much overlap there is, and check whether the content that I wrote is still needed.

tlyu:
Updated. I think pointing to the RFCs that specify the detailed rules might be enough for now? They're sufficiently complex that they might deserve their own chapter.

@rustbot label -S-waiting-on-author +S-waiting-on-review

PR #1064, opened 2021-07-08
Clarify closure capture modes slightly

jonas-schievink:
This improves the wording a bit, which came up in https://users.rust-lang.org/t/why-compiler-prefer-failed-than-choosing-move/62071


ehuss (approved):
Thanks!

PR #1073, opened 2021-07-13
Add note about the sign of the remainder

### LINKED ISSUE #978: I wish it were clearer what % does on negative numbers
In [Rust reference 8.2.4](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#arithmetic-and-logical-binary-operators), the table explains only that the % operator performs "remainder".

I write video games and I spend a lot of time performing the modulo operation on negative numbers. I get very easily confused about what, and I don't feel I clearly understand from reading 8.2.4 what Rust % would do if one or both operands were negative.

Maybe there are people out there for whom the use of "remainder" rather than "modulo", combined with "Integer division rounds towards zero", would be unambiguous, but I am not one :(

**"Expected behavior"**: It would be nice to have at least a line of text indicating how we should interpret "remainder", in particular in the case of negative operands. It would be even better if the "examples" block with the assert_eqs included some different negative-number examples for %.

I asked about this on Twitter and it was pointed out to me [the stdlib entry for numbers](https://doc.rust-lang.org/beta/std/ops/trait.Rem.html#impl-Rem%3Cu64%3E) is much more explicit. I don't know exactly what your rules are for which things need to be included in the Reference and which should be deferred to the stdlib.

notriddle:
Fixes #978

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -192,16 +192,18 @@ The operands of all of these operators are evaluated in [value expression contex
 | `-`    | Subtraction             |               | Subtraction    | `std::ops::Sub`    | `std::ops::SubAssign`                 |
 | `*`    | Multiplication          |               | Multiplication | `std::ops::Mul`    | `std::ops::MulAssign`                 |
 | `/`    | Division*               |               | Division       | `std::ops::Div`    | `std::ops::DivAssign`                 |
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
Thanks!

PR #1075, opened 2021-07-20
items/generics.md: fix generic params order

LittleEntity:
Example:
```rust
struct Test<const N: usize, A,'s> {
    a: A,
    s: &'s str,
    arr: [u8; N],
}
```
This does not compile. 

1. error: type parameters must be declared prior to const parameters
2. error: lifetime parameters must be declared prior to const parameters

> help: reorder the parameters: lifetimes, then types, then consts: `<'s, A, const N: usize>`

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -3,19 +3,29 @@
 > **<sup>Syntax</sup>**\
 > _GenericParams_ :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `<` `>`\
-> &nbsp;&nbsp;  | `<` (_GenericParam_ `,`)<sup>\*</sup> _GenericParam_ `,`<sup>?</sup> `>`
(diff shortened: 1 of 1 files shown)


LittleEntity:
If the description for this PR is too small: Let me know.

Thankyou for making Rust my favorite language 💖

The tooling, the language and the documentation is outstanding!

ehuss:
Thanks for the PR!  The grammar describes the parsing stage of the language (essentially what gets converted to an AST). The restriction on ordering comes at a later stage, and the order restriction is described in that section in English.

For example, the sample you show is valid input to a macro. You can see whether or not something parses correctly by using a `#[cfg]` attribute with a false value, like this:

```rust
// This compiles successfully.
#[cfg(IGNORE_THIS)]
struct Test<const N: usize, A,'s> {
    a: A,
    s: &'s str,
    arr: [u8; N],
}
```

ehuss:
Closing as the original is the intended format.

PR #1076, opened 2021-07-21
Remove incorrect apostrophe

carols10cents:
This should be the other form of its.

I couldn't unsee it after I noticed it. ❤️


ehuss (approved):
Thanks 😄

PR #1079, opened 2021-07-27
Mention "function item type to `fn pointer`" coercion

JohnTitor:
Fixes https://github.com/rust-lang/nomicon/issues/237
This is mentioned in [RFC 401](https://github.com/rust-lang/rfcs/blob/master/text/0401-coercions.md#function-type-polymorphism) but it seems the reference doesn't mention it currently.


ehuss (approved):
Thanks!

PR #1082, opened 2021-08-12
Document if-let guards

LeSeulArtichaut:
cc @pnkfelix https://github.com/rust-lang/rfcs/pull/2294 https://github.com/rust-lang/rust/issues/51114

### DIFF
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -21,7 +21,14 @@
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> [_Pattern_] _MatchArmGuard_<sup>?</sup>
 >
 > _MatchArmGuard_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _MatchArmIfGuard_\
+> &nbsp;&nbsp; | _MatchArmIfLetGuard_
+>
+> _MatchArmIfGuard_:\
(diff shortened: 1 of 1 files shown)


src/expressions/match-expr.md
```diff
@@ -102,6 +111,21 @@ let message = match maybe_digit {
 };
 ```
 
+`if let` guards are introduced by the `if let` keywords and are composed of a pattern, an `=` and a scrutinee expression.
```
kw-fn, line 114:
```suggestion
`if let` guards are composed following [The Rust Reference `if let` keywords syntax](https://doc.rust-lang.org/stable/reference/expressions/if-expr.html#if-let-expressions).
```

ehuss: @kw-fn, I think the original is correct here, as it is just describing the syntax, and is a style used throughout the reference as a way to anchor the description of what the text is talking about.  Also, we do not use external links, or refer to the reference by name within itself.

ehuss:
Thanks @LeSeulArtichaut, this looks good to me!  It will need to wait for it to be stabilized on nightly, though happy to discuss any feedback from the lang team in the meantime.

Fishrock123:
This text reads well to me.

ehuss:
Closing as superseded by https://github.com/rust-lang/reference/pull/1957.

PR #1084, opened 2021-08-28
expressions.md: Attempt fixing broken grammar in Mutability paragraph

daerich:
Turning to the reference for a quick brush-up of my knowledge, I stumbled upon this peculiar grammar in the mutability paragraph. Attempt to correct it, keeping its original meaning.


JohnTitor (approved):
Good catch, thanks!

PR #1088, opened 2021-09-07
Document using liveness to infer auto traits for async blocks

eholk:
In order to fix some of the issues around overly conservative generator captures (rust-lang/rust#69663), I am working on changing the analysis to be based on liveness of values rather than their scope. This is an update to the reference that describes how I propose things to work in the future, and I'd appreciate some feedback on it. Obviously we shouldn't merge this until the rustc changes are ready too, but I at least wanted to notify the lang team and get some feedback.


eholk:
@rustbot label +I-nominate +T-lang

I'd appreciate some feedback. I'm also open to doing a video or chat meeting if that's helpful.

Mark-Simulacrum:
Niko mentioned this in today's T-lang meeting and I raised that it'd be good to document the *current* state for this inference before adding changes, so that there's a good diff to look at.

eholk:
@Mark-Simulacrum - That's a good idea, I'll put up a PR documenting the current rules and then rebase this one on top of that so that the differences are clearer.

eholk:
I just pushed #1093 which documents the current behavior. This PR now includes that one as well, so once we merge that one it will be easier to see what is changing.

PR #1089, opened 2021-09-08
Add I-nominated and T-lang

eholk:
This will allow regular users to nominate PRs to this repo for lang team consideration.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -1,6 +1,7 @@
 [relabel]
 allow-unauthenticated = [
-    "S-*", "A-*", "New Content", "Language Cleanup", "Easy", "Formatting", "Enhancement", "Bug",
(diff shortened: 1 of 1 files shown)


ehuss:
I don't think rust-lang org typically allows unauthenticated access to I-nominated (see for example [here](https://github.com/rust-lang/rust/blob/c9db3e0fbc84d8409285698486375f080d361ef3/triagebot.toml#L9)).  If the lang team wants this, then that's fine with me, but otherwise I don't think this is normally done.  Also, I think we've had a grand total of two PRs ever nominated, so it doesn't happen too often, and I'm fine adding the labels manually if someone asks for it.

eholk:
Sounds good! I guess this is kind of a big hammer to fix a small problem. I'll just ping you next time I need this.

PR #1093, opened 2021-09-29
Document auto trait inference for async blocks

eholk:
We are in the process of changing this (rust-lang/rust#69663), but it would be good to document the existing rules before changing them. This should also help explain the compilation errors people are getting in the meantime.

### DIFF
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -117,6 +117,24 @@ loop {
 }
 ```
 
+### Auto traits and `async` blocks
+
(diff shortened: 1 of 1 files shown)


ehuss:
@pnkfelix Did you have any more comments on this?

eholk:
This came up in the Lang Docs call today. It'd be good to get some up to date documentation in this area. The behavior described in this PR has changed now that drop-tracking-mir is on by default, so it needs to be brought up to date.

#1088 might describe the current behavior, but it also should be checked for correctness.

PR #1095, opened 2021-10-02
Use subtrait/supertrait, not sub-trait, super-trait.

waywardmonkeys:
There are definitions of subtrait and supertrait, so use the
official terminology.


ehuss (approved):
Thanks!

Issue #24, opened 2017-03-23
 Default type parameters are not in the language reference


steveklabnik:
Migrated from https://github.com/rust-lang/rust/issues/38011

-------------------------------------

Default type parameters (e g `struct Foo<A = u32>(A)`) is a language feature, but they are not mentioned in the [language reference](https://doc.rust-lang.org/reference.html#type-parameters).

It would be nice if such a reference also mentioned the tricky semantics around how it interacts with inference, as [mentioned here](https://github.com/rust-lang/rust/issues/36980#issuecomment-251726254).

mankinskin:
It would be even better if the semantics could be less tricky :) Where is the problem with always trying to infer unspecified types and falling back on defaults when they can't be inferred? This is the behaviour I would expect, at least.

SilvanCodes:
I would like the first case to work:
```rust
struct Foo<T = u8>(T);
    
impl<T: Default> Default for Foo<T> {
    fn default() -> Self {
        Self(Default::default())
    }
}
    
// This breaks with: error[E0282]: type annotations needed for `main::Foo<T>`
let a = Foo::default();

// This works.
let b: Foo = Foo::default();
```

See it live here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0577801bc59bb19cc6c61039863d130a

Havvy:
You should file an issue against rust-lang/rust about that. The reference just documents what's already there.

ehuss:
See also [RFC 213](https://github.com/rust-lang/rfcs/blob/master/text/0213-defaulted-type-params.md).

tlyu:
I might try to work on this. Is [RFC 213](https://rust-lang.github.io/rfcs/0213-defaulted-type-params.html) the correct reference? Or is there a newer one?

tlyu:
The issue mentioned in [this comment](https://github.com/rust-lang/rust/issues/36980#issuecomment-251726254) seems to be another undocumented semantic difference. A QualifiedTypePath (in a call expression) apparently must resolve without inference (but possibly by using default types for type parameters). A PathInExpression (in a call expression) that leads to a type is apparently still subject to inference.

I'm still trying to work through whether the behavior observed by that commenter conflicts with what the RFC says.

Note that some of the features in the RFC are now [deprecated](https://github.com/rust-lang/rust/issues/36887). (No defaulted type parameters on functions, at least.)

PR #1098, opened 2021-10-21
document `#![feature(const_generics_defaults)]`

BoxyUwU:
[stabilisation report](https://github.com/rust-lang/rust/pull/90207#issuecomment-958843546) for this feautre

cc @rust-lang/project-const-generics

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -22,7 +22,7 @@
 parameters are listed in angle <span class="parenthetical">brackets (`<...>`)</span>,
 usually immediately after the name of the item and before its definition. For
 implementations, which don't have a name, they come directly after `impl`.
-The order of generic parameters is restricted to lifetime parameters and then type/const parameters intermixed.
+The order of generic parameters is restricted to lifetime parameters and then type and const parameters intermixed.
 
 Some examples of items with type, const, and lifetime parameters:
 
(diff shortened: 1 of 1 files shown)


src/items/generics.md
```diff
@@ -31,6 +31,7 @@ fn foo<'a, T>() {}
 trait A<U> {}
 struct Ref<'a, T> where T: 'a { r: &'a T }
 struct InnerArray<T, const N: usize>([T; N]);
+struct Uwu<const N: bool, U>(U);
```
joshtriplett, line 34:
```suggestion
struct EitherOrderWorks<const N: bool, U>(U);
```

No issue with the pwayfuwness, just twying to document what this example is twying to show. :smiley_cat:

ehuss (commented):
Thanks for the PR!  There seems to be some information missing, and I was wondering if you could incorporate some more?  For example, some of the following are things that I've noticed:

* What are the restrictions for the expression in the default?
* Restrictions on the type of the default expression.  I believe it must be the same type as the declaration, but can there be any coercion? Is it essentially a [constant expression](https://doc.rust-lang.org/nightly/reference/const_eval.html#constant-expressions)? 

And anything else you can think of?

I understand that default generic parameters are not documented at all (#24). Although it would be great to have that, I think it's fine to not include it now.

Also, it can usually be helpful to provide some links to more information in the PR.  I found the [stabilization report](https://github.com/rust-lang/rust/pull/90207#issuecomment-958843546) which is very useful.

src/items/generics.md
yaahc: Might I suggest `EithewOwdewWowks`

ehuss (approved):
I'm going to go ahead and merge as-is. However, I would appreciate if anyone can help fill out the details more.  I opened #1159 to track the rest of this.

PR #1101, opened 2021-10-31
Byte literal and string syntax: add missing quote escapes

osa1:
Current definitions of byte characters and byte literals do not allow escaped
single and double quotes. Updated lexical syntax for byte characters and
strings to allow these characters. This is in sync with the implementation.

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -235,7 +235,7 @@ r##"foo #"# bar"##;                // foo #"# bar
 >
 > BYTE_ESCAPE :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `\x` HEX_DIGIT HEX_DIGIT\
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!

I believe both byte literals and byte strings support `\'` and `\"`, so perhaps just add those to `BYTE_ESCAPE`?

osa1:
Thanks for the review @ehuss and sorry for the late response.

> I believe both byte literals and byte strings support \' and \", so perhaps just add those to BYTE_ESCAPE?

You're right. I added `\'` and `\"` escapes to `BYTE_ESCAPE` so now they're valid in both byte and byte string syntax. This is again in sync with the implementation.

ehuss (approved):
Thanks!

PR #1102, opened 2021-11-05
Document aarch64 `target_feature` options

adamgemmell:
Draft documentation update https://github.com/rust-lang/rust/issues/90620

Of note is the fact that I haven't documented the "v8.Xa" target features in order to discourage their use. It's much better to use the more granular target features available.

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -133,6 +133,71 @@ Feature     | Implicitly Enables | Description
 [`xsaveopt`]: https://www.felixcloutier.com/x86/xsaveopt
 [`xsaves`]: https://www.felixcloutier.com/x86/xsaves
 
+#### `aarch64`
+
+This platform requires that `#[target_feature]` is only applied to [`unsafe`
+functions][unsafe function].
+
+Further documentation on these features can be found in the [ARM Architecture
+Reference Manual], or elsewhere on [developer.arm.com].
+
+[ARM Architecture Reference Manual]: https://developer.arm.com/documentation/ddi0487/latest
+[developer.arm.com]: https://developer.arm.com
+
+> ***Note***: The following pairs of features should both be marked as enabled
+> or disabled together if used:
+> - `fp` and `neon`, in order facilitate inlining in more places, among other reasons.
+> - `paca` and `pacg`, which LLVM currently implements as one feature.
+
+
+Feature        | Implicitly Enables | Feature Name
+---------------|--------------------|-------------------
+`aes`          | `neon`         | FEAT_AES - Advanced <abbr title="Single Instruction Multiple Data">SIMD</abbr> AES instructions
+`bf16`         |                | FEAT_BF16 - BFloat16 instructions
+`bti`          |                | FEAT_BTI - Branch Target Identification
+`crc`          |                | FEAT_CRC - CRC32 checksum instructions
+`dit`          |                | FEAT_DIT - Data Independent Timing instructions
+`dotprod`      |                | FEAT_DotProd - Advanced SIMD Int8 dot product instructions
+`dpb`          |                | FEAT_DPB - Data cache clean to point of persistence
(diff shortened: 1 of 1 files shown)


adamgemmell:
@ehuss This and https://github.com/rust-lang/reference/pull/1061 are no longer blocked on stabilisation - we're currently merging PRs for this (see https://github.com/rust-lang/rust/issues/86941#issuecomment-1032838605)

src/attributes/codegen.md
```diff
@@ -131,6 +131,69 @@ Feature     | Implicitly Enables | Description
...
+`rcpc`         |                | FEAT_LRCPC - Release consistent Processor Consistent
+`rcpc2`        | `rcpc`         | FEAT_LRCPC2 - RcPc with immediate offsets
+`rdm`          |                | FEAT_RDM - Rounding Double Multiply accumulate
+`sb`           |                | FEAT_SB - Speculation Barrier
+`sha2`         | `neon`         | FEAT_SHA1 & FEAT_SHA256 - Advanced SIMD SHA instructions
+`sha3`         | `neon`, `sha2` | FEAT_SHA512 & FEAT_SHA3 - Advanced SIMD SHA instructions
+`sm4`          | `neon`         | FEAT_SM3 & FEAT_SM4 - Advanced SIMD SM3/4 instructions
+`spe`          |                | FEAT_SPE - Statistical Profiling Extension
+`ssbs`         |                | FEAT_SSBS - Speculative Store Bypass Safe
+`sve`          | `fp`           | FEAT_SVE - Scalable Vector Extension
```
adamgemmell, line 188: SVE relies on fp16

src/attributes/codegen.md
```diff
@@ -131,6 +131,69 @@ Feature     | Implicitly Enables | Description
...
+`bti`          |                | FEAT_BTI - Branch Target Identification
+`crc`          |                | FEAT_CRC - CRC32 checksum instructions
+`dit`          |                | FEAT_DIT - Data Independent Timing instructions
+`dotprod`      |                | FEAT_DotProd - Advanced SIMD Int8 dot product instructions
+`dpb`          |                | FEAT_DPB - Data cache clean to point of persistence
+`dpb2`         |                | FEAT_DPB2 - Data cache clean to point of deep persistence
+`f32mm`        | `sve`          | FEAT_F32MM - SVE single-precision FP matrix multiply instruction
+`f64mm`        | `sve`          | FEAT_F64MM - SVE double-precision FP matrix multiply instruction
+`fcma`         | `neon`         | FEAT_FCMA - Floating point complex number support
+`fhm`          | `fp`           | FEAT_FHM - Half-precision FP FMLAL instructions
```
adamgemmell, line 162: FHM relies on `fp16`

ehuss:
Thanks for putting this together!

I can't find how to verify the implied feature mappings.  How can I validate those?  I see the [feature mapping](https://github.com/rust-lang/rust/blob/984204814e00f60c5e1ec99e2e184f326782a586/compiler/rustc_codegen_llvm/src/llvm_util.rs#L190-L211) on the rust side, but those don't seem to completely encompass the mappings here. The x86 side has the implicit mappings defined ([here](https://github.com/llvm/llvm-project/blob/cb22d71806b784a9105803f38b2740ca6888392a/llvm/lib/Support/X86TargetParser.cpp#L464-L591)), but I can't find the same thing for the aarch64 side.

adamgemmell:
@ehuss I've been using https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/AArch64.td combined with the "to_llvm_features" mapping you linked. The array of features at the end of each def provides the implied mapping.

I can see X86.td also exists with this information, not really sure why it's duplicated there. The two sources don't seem to conflict at least.

ehuss (approved):
Thanks!  I'm going to approve even though I am not particularly an expert here. It seems to match what has been stabilized. Followup PRs to fix anything amiss are welcome.

PR #1104, opened 2021-11-08
Update list of types with magic Copy impls

bstrie:
Numeric types, char, bool, `!`, shared references, and raw pointers all have ordinary Copy impls in std, not
provided by the compiler. Also, mention closures here.

### DIFF
--- a/src/special-types-and-traits.md
+++ b/src/special-types-and-traits.md
@@ -71,7 +71,7 @@ For unions, this means all variants have to be `Copy`.
 * [Arrays] of `Copy` types
 * [Function pointers]
 * [Function items]
-* [Closures] that capture no values or that only capture values of `Copy` types.
+* [Closures] that capture no values or that only capture values of `Copy` types
 
 ## `Clone`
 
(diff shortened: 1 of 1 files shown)


src/special-types-and-traits.md
```diff
@@ -67,13 +67,11 @@ For unions, this means all variants have to be `Copy`.
...
-* [Numeric types]
-* `char`, `bool`, and [`!`]
 * [Tuples] of `Copy` types
 * [Arrays] of `Copy` types
-* [Shared references]
-* [Raw pointers]
-* [Function pointers] and [function item types]
+* [Function pointers]
+* [Function items]
+* [Closures] that capture no values or that only capture values of `Copy` types.
```
JohnTitor, line 74:
```suggestion
* [Closures] that capture no values or that only capture values of `Copy` types
```

JohnTitor (approved):
Makes sense to me.

bstrie:
I've pushed an additional commit to remove the lines about providing Copy/Clone impls on arrays, since libstd does that as of recently: https://github.com/rust-lang/rust/pull/86041

ehuss (approved):
Thanks!

PR #1105, opened 2021-11-09
Add inline assembly to the reference

Amanieu:
This PR adds inline assembly to the reference, pending stabilization in https://github.com/rust-lang/rust/issues/72016.

I have included all features currently supported by inline assembly, including the ones that are not being stabilized immediately (`sym` & `const` operands, support for additional architectures). This is intentional since it keeps all of the documentation about inline assembly in a single place.


src/inline-assembly.md
```diff
@@ -0,0 +1,462 @@
...
+- ARM
+- AArch64
+- RISC-V
+- NVPTX
+- PowerPC
+- Hexagon
+- MIPS
+- WebAssembly
+- BPF
+- SPIR-V
```
bjorn3, line 32: Maybe clarify that for many it is currently unstable?

src/inline-assembly.md
```diff
@@ -0,0 +1,462 @@
...
+  - Behavior is undefined if the direction flag is set on exiting an asm block.
+- The requirement of restoring the stack pointer and non-output registers to their original value only applies when exiting an `asm!` block.
+  - This means that `asm!` blocks that never return (even if not marked `noreturn`) don't need to preserve these registers.
+  - When returning to a different `asm!` block than you entered (e.g. for context switching), these registers must contain the value they had upon entering the `asm!` block that you are *exiting*.
+    - You cannot exit an `asm!` block that has not been entered. Neither can you exit an `asm!` block that has already been exited.
+    - You are responsible for switching any target-specific state (e.g. thread-local storage, stack bounds).
+    - The set of memory locations that you may access is the intersection of those allowed by the `asm!` blocks you entered and exited.
+- You cannot assume that an `asm!` block will appear exactly once in the output binary. The compiler is allowed to instantiate multiple copies of the `asm!` block, for example when the function containing it is inlined in multiple places.
```
bjorn3, line 460: Maybe explain that the implication of this is that only numbered labels are allowed and not named labels?

ehuss (commented):
Thank you very much for preparing this!

I am very reluctant to add unstable documentation here. The reference very explicitly tries to avoid adding any unstable documentation, and I don't know why we would add an exception here. Would it be possible to relegate any unstable documentation to the unstable book? Or leave unstable parts commented out? I understand the desire to have some level of consolidation of documentation, but going by that logic, everything from the unstable book would end up here.

Can you add a newline after each sentence? That is the style used in the reference.

Would it be possible to sprinkle a few examples in? I think it is very helpful to have a few illustrative examples, and we try to do that throughout the reference.

I notice that the unstable book has more in-depth documentation. Is there a plan for how that documentation will live on?

Amanieu:
I can move the unstable parts (unstable arch registers, `sym` & `const`) to the unstable book.

The guide-level explanation from the unstable book (which is from the RFC) will most likely go into Rust by example.

Amanieu:
`asm!` and `global_asm!` are now stable on the latest nightly, so the tests should pass.

Amanieu:
Ping! I'd like to get this merged before the beta branch in 2 weeks which include stable inline assembly.

src/inline-assembly.md
```diff
@@ -0,0 +1,474 @@
...
+dir_spec := "in" / "out" / "lateout" / "inout" / "inlateout"
+reg_spec := <register class> / "<explicit register>"
+operand_expr := expr / "_" / expr "=>" expr / expr "=>" "_"
+reg_operand := dir_spec "(" reg_spec ")" operand_expr
+operand := reg_operand
+clobber_abi := "clobber_abi(" <abi> *("," <abi>) [","] ")"
+option := "pure" / "nomem" / "readonly" / "preserves_flags" / "noreturn" / "nostack" / "att_syntax" / "raw"
+options := "options(" option *("," option) [","] ")"
+asm := "asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," clobber_abi) *("," options) [","] ")"
+global_asm := "global_asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," options) [","] ")"
```
ehuss, line 51: `format_string` isn't defined in this mini-grammar. Perhaps add something like this?

```
format_string := STRING_LITERAL / RAW_STRING_LITERAL
```

ehuss:
Thanks for the updates!

If I'm understanding this correctly, this seems to introduce a new set of undefined behaviors to the language? Would it make sense to add an entry to the bullet list at https://github.com/rust-lang/reference/blob/master/src/behavior-considered-undefined.md that links to this chapter?  

Otherwise, this looks good to me, though I can't vouch for the accuracy of all the information as I have no experience with it. I did at least scanned over the parser and it seems to match.

I don't know if anyone else would want to comment.  Maybe @joshtriplett would be willing to chime in?

src/inline-assembly.md
```diff
@@ -0,0 +1,474 @@
...
+  - RISC-V
+    - Floating-point exception flags in `fcsr` (`fflags`).
+    - Vector extension state (`vtype`, `vl`, `vcsr`).
+- On x86, the direction flag (DF in `EFLAGS`) is clear on entry to an asm block and must be clear on exit.
+  - Behavior is undefined if the direction flag is set on exiting an asm block.
+- The requirement of restoring the stack pointer and non-output registers to their original value only applies when exiting an `asm!` block.
+  - This means that `asm!` blocks that never return (even if not marked `noreturn`) don't need to preserve these registers.
+  - When returning to a different `asm!` block than you entered (e.g. for context switching), these registers must contain the value they had upon entering the `asm!` block that you are *exiting*.
+    - You cannot exit an `asm!` block that has not been entered.
+      Neither can you exit an `asm!` block that has already been exited.
```
joshtriplett, line 466:
```suggestion
      Neither can you exit an `asm!` block that has already been exited (without first entering it again).
```

src/inline-assembly.md
```diff
@@ -0,0 +1,474 @@
...
+    - You are responsible for switching any target-specific state (e.g. thread-local storage, stack bounds).
```
joshtriplett, line 467:
```suggestion
    - You are responsible for switching any target-specific state (e.g. thread-local storage, stack bounds).
    - You cannot jump from an address in one `asm!` block to an address in another, even within the same function or block, without treating their contexts as potentially different and requiring context switching. You cannot assume that any particular value in those contexts (e.g. current stack pointer, temporary values below the stack pointer, flags) will remain unchanged between the two `asm!` blocks.
```

src/inline-assembly.md
```diff
@@ -0,0 +1,474 @@
...
+    - The set of memory locations that you may access is the intersection of those allowed by the `asm!` blocks you entered and exited.
+- You cannot assume that an `asm!` block will appear exactly once in the output binary.
```
joshtriplett, line 469:
```suggestion
- You cannot assume that two `asm!` blocks adjacent in source code, even without any other code between them, will end up in successive addresses in the binary without any other instructions between them.
- You cannot assume that an `asm!` block will appear exactly once in the output binary.
```

Amanieu:
Thanks for the feedback! I've applied all the suggestions, so it should be good to go.

joshtriplett:
I remain saddened by the restrictions on assuming a single instance of an `asm!`, and I think people are highly likely to end up ignoring that and getting the results they expect with some careful use of `#[inline(never)]` and further care. I am not suggesting at this time that we know all the steps necessary to make that work reliably, but people will figure it out in practice when they need it, and I think we should be prepared to consider documenting those findings.

But I don't consider this a blocking concern, insofar as we can always *loosen* a restriction.

joshtriplett:
I think this is ready to merge.

joshtriplett (approved):

ehuss (approved):
Thank you!!

I'll update the books early next week before the branch.

src/inline-assembly.md
```diff
@@ -0,0 +1,477 @@
...
+    Other registers follow target-specific rules.
+  - Note that a `lateout` may be allocated to the same register as an `in`, in which case this rule does not apply.
+    Code should not rely on this however since it depends on the results of register allocation.
+- Behavior is undefined if execution unwinds out of an asm block.
+  - This also applies if the assembly code calls a function which then unwinds.
+- The set of memory locations that assembly code is allowed to read and write are the same as those allowed for an FFI function.
+  - Refer to the unsafe code guidelines for the exact rules.
+  - If the `readonly` option is set, then only memory reads are allowed.
+  - If the `nomem` option is set then no reads or writes to memory are allowed.
+  - These rules do not apply to memory which is private to the asm code, such as stack space allocated within the asm block.
```
Freax13, line 429: Do the unsafe code guidelines actually have any information about this? I can't find anything in the [published reference](https://rust-lang.github.io/unsafe-code-guidelines/) and I also can't find anything in the [unsafe code guidelines' github issues](https://github.com/rust-lang/unsafe-code-guidelines/issues?q=label%3AT-ffi). While not mentioned here, I also couldn't find anything in the [Rustonomicon](https://doc.rust-lang.org/nomicon/). Did I miss something or is there some other place that has information about this?

Besides that, isn't the whole point of the unsafe code guidelines to be a place for discussion before being moved into the proper reference? It seems a bit weird to have the official reference point to an unofficial reference.

PR #1107, opened 2021-11-18
Try to clarify destructor not being run scenario.

nbdd0121:
This is being discussed in https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Guarantees.20around.20Drop.20.28.2390752.29.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -349,9 +349,8 @@ let x = (&temp()).use_temp();  // ERROR
 and [`std::mem::ManuallyDrop`] provides a wrapper to prevent a
 variable or field from being dropped automatically.
 
-> Note: Preventing a destructor from being run via `forget` or other means is safe in Rust
-> even if it has a type that isn't `'static`. This means that publicly exposed APIs cannot
-> rely on destructor being run for soundness.
+> Note: Preventing a destructor from being run via [`std::mem::forget`] or other means is safe even if it has a type that isn't `'static`.
+> Besides the places where destructors are guaranteed to run as defined by this document, types may *not* safely rely on a destructor being run for soundness.
 
 [Assignment]: expressions/operator-expr.md#assignment-expressions
 [binding modes]: patterns.md#binding-modes
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -345,9 +345,13 @@ let x = (&temp()).use_temp();  // ERROR
 
 ## Not running destructors
 
-Not running destructors in Rust is safe even if it has a type that isn't
-`'static`. [`std::mem::ManuallyDrop`] provides a wrapper to prevent a
+[`std::mem::forget`] can be used to prevent the destructor of a variable from being run,
+and [`std::mem::ManuallyDrop`] provides a wrapper to prevent a
 variable or field from being dropped automatically.
+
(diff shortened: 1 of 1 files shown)


ecstatic-morse:
rust-lang/nomicon#135 has some background.

ehuss (approved):
Thanks!  This looks much better to me.

However, there seems to be quite a bit of discussion that you linked to.  Is that still ongoing? Was there more feedback that you want before merging this?

ecstatic-morse:
@ehuss Niko mentioned a lang-team FCP on the PR is the usual procedure. Should we kick one off?

src/destructors.md
```diff
@@ -345,10 +345,14 @@ let x = (&temp()).use_temp();  // ERROR
...
 
-Not running destructors in Rust is safe even if it has a type that isn't
-`'static`. [`std::mem::ManuallyDrop`] provides a wrapper to prevent a
+[`std::mem::forget`] can be used to prevent the destructor of a variable from being run,
+and [`std::mem::ManuallyDrop`] provides a wrapper to prevent a
 variable or field from being dropped automatically.
 
+> Note: Preventing a destructor from being run via `forget` or other means is safe in Rust
+> even if it has a type that isn't `'static`. This means that publicly exposed APIs cannot
+> rely on destructor being run for soundness.
```
Mark-Simulacrum, line 354: It seems like it might be good to expand a little on this: for one thing, publicly exposed APIs seems like a somewhat "interesting" term. Certainly even in a crate-local context I would expect to see unsafe fn on the constructor or similar if the destructor must run for soundness reasons.

Maybe we can adjust to something like?

> Note that while code can safely prevent a destructor from running, they are still guaranteed to automatically run in places defined by this document (such as the end of scopes). Authors may rely on destructors running at these locations for soundness, but must not otherwise assume destructors are run.

ehuss: Ping @nbdd0121, was wondering if you'd be willing to change the note to something like the suggestion given here?

nbdd0121: Sorry for not being responsive, I am quite busy recently.

I don't think it's good to explicit saying that "you can expect destructor at the end of scope to run for soundness", as I think it could create additional confusion, as these are really just part of control flow that users expect.

The purpose of this change is to keep its original intention while removing the possibility to interpret it as a permit for compilers to omit destructor calls.

ehuss: No worries, I was just checking in.

How about something like this?

> Note: Preventing a destructor from being run via `std::mem::forget` or other
> means is safe even if it has a type that isn't `'static`.
> Besides the places where destructors are guaranteed to run as defined by
> this document, types may **not** safely rely on a destructor being run for
> soundness.

I'm trying to blend the different ideas here:

* It should always be safe to call `forget`.
* There are places where destructors are guaranteed to run, and you may rely on that.
* Types may not rely on their destructor being run for soundness.

I'm agree with Mark about the "publicly exposed APIs", so I'm trying to avoid that phrasing.

nbdd0121: Sounds very good. I have applied your suggestion.

ehuss (approved):
Thanks!

PR #1109, opened 2021-11-22
fix crate_name attribute description

neetsdkasu:
I think the underscore is correct, not the hyphen

image

```rust
#![crate_name = "foo-bar"]

fn main() {
    println!("Hello, world!");
}
```
```bash
$ rustc src\main.rs
error: invalid character `-` in crate name: `foo-bar`
 --> src\main.rs:1:1
  |
1 | #![crate_name = "foo-bar"]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^

```

```rust
#![crate_name = "foo\u{002D}bar"]

fn main() {
    println!("Hello, world!");
}
```
```bash
$ rustc src\main.rs
error: invalid character `-` in crate name: `foo-bar`
 --> src\main.rs:1:1
  |
1 | #![crate_name = "foo\u{002D}bar"]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

`#![crate_name = "foo_bar"]` is no error
`#![crate_name = "foo\u{005F}bar"]` is no error

### DIFF
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -132,7 +132,7 @@ name of the crate with the [_MetaNameValueStr_] syntax.
 ```
 
 The crate name must not be empty, and must only contain [Unicode alphanumeric]
-or `-` (U+002D) characters.
+or `_` (U+005F) characters.
 
 [^phase-distinction]: This distinction would also exist in an interpreter.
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thank you!  Not sure how I flubbed that.

PR #1110, opened 2021-11-27
Update 'Subtyping and Variance' example to use `dyn Trait` syntax

mattheww:
This will allow the example to be built under edition 2021.


ehuss (approved):
Thanks!

PR #1111, opened 2021-11-27
Say that bare trait objects are rejected in the 2021 edition

### LINKED ISSUE #1070: 2021: Update for bare trait objects
The [trait object chapter](https://github.com/rust-lang/reference/blob/master/src/types/trait-object.md) needs to be updated for changing the `bare_trait_objects` to a hard error in 2021. The grammar and examples need to be updated, and a callout for the older behavior needs to be added. See the [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/warnings-promoted-to-error.md) for more.

mattheww:
Closes #1070.

This patch leaves the grammar block describing `dyn` as an optional keyword, treating the 2021-edition rejection of missing `dyn` as a post-parsing error (which is after all how it's implemented).

I think that's consistent with how the grammar works elsewhere: it describes what's allowed in any edition, while the text outside 'Edition Differences' blocks describes what's allowed in all editions.

### DIFF
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -14,15 +14,14 @@ number of [auto traits].
 Trait objects implement the base trait, its auto traits, and any [supertraits]
 of the base trait.
 
-Trait objects are written as the optional keyword `dyn` followed by a set of
-trait bounds, but with the following restrictions on the trait bounds. All
-traits except the first trait must be auto traits, there may not be more than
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Ah, indeed, for some reason I was thinking that the `dyn` checking was done during parsing. Thanks for the PR!

Issue #1071, opened 2021-07-09
2021: Update for ellipsis_inclusive_range_patterns


ehuss:
The [range patterns](https://github.com/rust-lang/reference/blob/master/src/patterns.md#range-patterns) needs to be updated for changing `ellipsis_inclusive_range_patterns` to a hard error in 2021. The grammar and examples need to be updated, and a callout for the older behavior needs to be added. See the [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/warnings-promoted-to-error.md) for more.

ehuss:
I'm going to close this as fixed by #1114.

PR #1114, opened 2021-12-03
Say that `...` range patterns are rejected in the 2021 edition

mattheww:
Corresponding issue is #1071

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -533,7 +533,7 @@ minimum to maximum value. The range of values for a `char` type are precisely th
 ranges containing all Unicode Scalar Values: `'\u{0000}'..='\u{D7FF}'` and
 `'\u{E000}'..='\u{10FFFF}'`.
 
-> **Edition Differences**: Before the 2021 edition, closed range patterns may be written using `...` in place of `..=`, with the same meaning.
+> **Edition Differences**: Before the 2021 edition, closed range patterns may also be written using `...` as an alternative to `..=`, with the same meaning.
 
 ## Reference patterns
 
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -443,8 +443,6 @@ A half-open range pattern in the style `a..` cannot be used to match within the 
 A pattern `a..=b` must always have a &le; b. It is an error to have a range pattern
 `10..=0`, for example.
 
-The `...` syntax is kept for backwards compatibility.
-
 Range patterns only work on scalar types. The accepted types are:
 
 * Integer types (u8, i8, u16, i16, usize, isize, etc.).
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -535,6 +533,8 @@ minimum to maximum value. The range of values for a `char` type are precisely th
 ranges containing all Unicode Scalar Values: `'\u{0000}'..='\u{D7FF}'` and
 `'\u{E000}'..='\u{10FFFF}'`.
 
+> **Edition Differences**: Before the 2021 edition, closed range patterns may be written using `...` in place of `..=`, with the same meaning.
```
bjorn3, line 536: Can you clarify that before the 2021 edition `..=` is also allowed?

mattheww: I've reworded in b83fcc1

ehuss:
Thanks for the PR!

The `_ObsoleteRangePattern_` grammar needs to have some kind of indication that it is for < 2021 editions.  The grammar blocks are intended to be standalone for the purpose of parsing (text in the English prose is intended for things done after parsing). There isn't much precedent for how that should be formatted. The only things currently are the footnotes in the [functions](https://doc.rust-lang.org/nightly/reference/items/functions.html) grammar.  I think it would be fine to do something similar here.  It's a bit awkward to define multiple languages simultaneously.

mattheww:
I've added a footnote.

I'm not sure anything currently defines whether this rule is considered to be part of the grammar or something that happens after parsing: I don't know a way to make rustc distinguish these cases.

As far as the implementation is concerned, `...` isn't rejected when parsing, but by a lint that runs over the generated AST.

I see rust-lang/wg-grammar#6 and rust-lang/wg-grammar#41 have been left unsettled.

Out of interest, I checked what `syn` does, and it unconditionally accepts both `...` and `..=` without distinguishing them in the returned AST. It doesn't seem to contemplate parsing varying by edition.

ehuss:
I'm sorry, I misremembered how that was implemented. I went ahead and removed the last commit, since I was mistaken about how this works.  I was thinking this was done in the parser.

From a language perspective, you can test how parsing works via macros, or using false `#[cfg]` attributes. For example, the following are some examples that work in 2021:

```rust
macro_rules! pat {
    ($e:pat) => {};
}

#[cfg(FALSE)]
fn f(x: i32) {
    match x {
        1...2 => {}
    }
}

fn main() {
    pat!(1...2);
}
```

The idea for the grammar in the reference is to capture what ends up in the AST, which is relevant to macros and expansion.

As for how this is actually implemented, the key part is [here](https://github.com/rust-lang/rust/blob/b8dc6aa673317cf72a8abb9c421f573f2b34b47a/compiler/rustc_lint/src/builtin.rs#L1733) where the lint is defined with `FutureIncompatibilityReason::EditionError`.  The majority of lints are handled after expansion.

ehuss (approved):

PR #1117, opened 2021-12-13
README.md: link to mdbook docs

SamB:
Makes things easier for would-be contributors,
though sometimes it might just tell them that
mdbook doesn't support what they wanted to do.


ehuss (approved):
Thanks!

Issue #1112, opened 2021-11-28
Primitive types list inconsistent


steffahn:
Small thing to fix, just haven’t got the time to do it myself right now.

On [this page](https://github.com/rust-lang/reference/blob/8db4edd7eab7931859713ac8016ba123ce42b061/src/types.md), 

https://github.com/rust-lang/reference/blob/8db4edd7eab7931859713ac8016ba123ce42b061/src/types.md?plain=1#L14-L18

> * Primitive types:
>    * Boolean — `true` or `false`
>    * Numeric — integer and float
>    * Textual — `char` and `str`
>    * Never — `!` — a type with no values

Every entry lists _types_ except for `Boolean` which lists _values_. Probably better should be

> * Primitive types:
>    * Boolean — `bool`
>    * Numeric — integer and float
>    * Textual — `char` and `str`
>    * Never — `!` — a type with no values

PR #1118, opened 2021-12-14
keep consistent for primitive types

Junnplus:
fixed: https://github.com/rust-lang/reference/issues/1112

### DIFF
--- a/src/types.md
+++ b/src/types.md
@@ -12,7 +12,7 @@ limited capabilities.
 The list of types is:
 
 * Primitive types:
-    * [Boolean] — `true` or `false`
+    * [Boolean] — `bool`
     * [Numeric] — integer and float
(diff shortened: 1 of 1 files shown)


src/types.md
```diff
@@ -12,8 +12,8 @@ limited capabilities.
 The list of types is:
 
 * Primitive types:
-    * [Boolean] — `true` or `false`
-    * [Numeric] — integer and float
+    * [Boolean] — `bool`
+    * [Numeric] — `integer` and `float`
```
ehuss, line 16: I would prefer to not use code spans for integer and float. The others are code spans because they are literal identifiers that can be used in code. "integer" and "float" are just words to encompass the types that are numeric.

ehuss (approved):
Thanks!

PR #1120, opened 2021-12-24
Note difference in CTFE timing between associated and free constants

XrXr:
Since const panics are now [stabilized], I figure it would be good to
document some of its gotchas. I couldn't really find documents for
the specifics I ran into. I am no const eval expert, and what I
wrote is basically a paraphrase of Ralf's comments[^1][^2],
but I hope to get the ball rolling for adding some docs.

I deliberately chose to not mention the guarantee about syntactically
referenced constants in functions that require code generation as it
seems hard to explain completely without some ambiguity. It also seems
to me that the rules are not completely stable[^3][^4] yet.

[stabilized]: https://github.com/rust-lang/rust/issues/89006
[^1]: https://github.com/rust-lang/rust/issues/91877#issuecomment-995026270
[^2]: https://github.com/rust-lang/rust/issues/91877#issuecomment-995977016
[^3]: https://github.com/rust-lang/rust/issues/71800
[^4]: https://github.com/rust-lang/rust/issues/51999#issuecomment-832086617

### DIFF
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -296,6 +296,22 @@ type. It is written the same as a [constant item].
 Unlike [free] constants, associated constant definitions undergo
 [constant evaluation] only when referenced.
 
+```rust
+struct Struct;
+
+impl Struct {
+    const ID: i32 = 1;
+    // Definition not immediately evaluated
+    const PANIC: () = panic!("compile-time panic");
+}
+
+fn main() {
+    assert_eq!(1, Struct::ID);
+    // Referencing Struct::PANIC causes compilation error
+    // let _ = Struct::PANIC;
+}
+```
+
 ### Associated Constants Examples
 
 A basic example:
@@ -335,24 +351,6 @@ impl ConstantIdDefault for OtherStruct {
 fn main() {
     assert_eq!(1, Struct::ID);
     assert_eq!(5, OtherStruct::ID);
-}
-```
-
-[Constant evaluation] timing:
-
-```rust
-struct Struct;
-
-impl Struct {
-    const ID: i32 = 1;
-    // Definition not immediately evaluated
-    const PANIC: () = panic!("compile-time panic");
-}
-
-fn main() {
-    assert_eq!(1, Struct::ID);
-    // Referencing Struct::PANIC causes compilation error
-    // let _ = Struct::PANIC;
 }
 ```
 
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -293,6 +293,9 @@ type that the definition has to implement.
 An *associated constant definition* defines a constant associated with a
 type. It is written the same as a [constant item].
 
+Unlike [free] constants, associated constant definitions undergo
+[constant evaluation] only when referenced.
+
 ### Associated Constants Examples
 
 A basic example:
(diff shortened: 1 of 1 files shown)


ehuss:
@oli-obk Would you be willing to review this?

oli-obk:
I'm not sure these docs belong here. This file is specifically about constant evaluation itself. You could add some documentation to associated constants clarifying that they are only evaluated at their use sites.

About the panic section, as it is written right now, it seems like no new information to me, can you elaborate on why you think it would be useful here to mention that panicking during CTFE will cause an error?

XrXr:
> You could add some documentation to associated constants clarifying that they are only evaluated at their use sites.

Sounds good, I will move it.

> can you elaborate on why you think it would be useful here to mention that panicking during CTFE will cause an error?

I added the panic section because the page had no explicit mention of panicking during CTFE, so I thought I would talk about it to show that it is a possible outcome of CTFE. Maybe a whole section for it is too much. I also put the section for proximity to the eval timing section since timing (whether CTFE happens at all) is more relevant if it could abort compilation.

XrXr:
@oli-obk I've reduced the scope of the change and relocated the additions as suggested. 77387cb

oli-obk (approved):

oli-obk:
I don't have merge powers. @ehuss this lgtm

src/items/associated-items.md
```diff
@@ -335,6 +338,24 @@ fn main() {
 }
 ```
 
+[Constant evaluation] timing:
```
ehuss, line 341: When I immediately read this, I didn't know what it meant by "timing".  

Instead of placing this down in the "examples" section, can this code block be moved up to the paragraph that you added above?  Generally in the reference, when there is an example demonstrating some concept, we try to write that example immediately after the paragraph that introduces that concept.

XrXr: That makes sense. Moved in be494280f42e3f998309087250d71601150a036e

ehuss (approved):
Thanks!

src/items/associated-items.md
```diff
@@ -293,6 +293,25 @@ type that the definition has to implement.
 An *associated constant definition* defines a constant associated with a
 type. It is written the same as a [constant item].
 
+Unlike [free] constants, associated constant definitions undergo
+[constant evaluation] only when referenced.
```
RalfJung, line 297: This seems odd, to make a guarantee by stating its *absence* on associated consts. Do we say explicitly anywhere that free constants are always evaluated, even when they are not used?

XrXr: > Do we say explicitly anywhere that free constants are always evaluated, even when they are not used?

I went through [Constant Evaluation](https://doc.rust-lang.org/stable/reference/const_eval.html) and [Constant items] and could not find it explicitly stated. Some possible ways forward:
 1. reword this line to say "Associated constant definitions undergo constant evaluation only when referenced.", removing the reference to free constants
 2. mention that unused free constants always undergo CTFE in [Constant items] and reword this sentence to link to it.

Do you have any preferences or recommendations?

[Constant items]:https://doc.rust-lang.org/stable/reference/items/constant-items.html

RalfJung: 2 sounds like something we will eventually want to do, but will probably require T-lang FCP to ensure that we are willing to commit to this as a guarantee.

PR #1125, opened 2022-01-01
Switch the default edition for examples to 2021

### LINKED ISSUE #1065: 2021: Switch default to 2021
The default for tests needs to be update to 2021. This involves changing [this line](https://github.com/rust-lang/reference/blob/ab60513a3a5a0591e237fddff5d027a982648392/book.toml#L14) to 2021, and fixing any broken tests.

Note that this is blocked on a new mdbook being published (I will try to do that soon). There will also need to be a new mdbook published once 2021 is stabilized on nightly.

mattheww:
This changes book.toml as requested in #1065, and updates the Github Actions to use a new enough version of mdBook.

I've also suggested an update to the part of the style guide that talks about edition annotations on examples.

Closes #1065

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Install mdbook
       run: |
         mkdir bin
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks!

PR #1126, opened 2022-01-01
Clarify behavior of x87 FP registers in inline assembly

Amanieu:
This came up in a discussion on Zulip.

cc @joshtriplett

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -460,6 +460,8 @@ To avoid undefined behavior, these rules must be followed when using function-sc
     - Vector extension state (`vtype`, `vl`, `vcsr`).
 - On x86, the direction flag (DF in `EFLAGS`) is clear on entry to an asm block and must be clear on exit.
   - Behavior is undefined if the direction flag is set on exiting an asm block.
+- On x86, the x87 floating-point register stack must remain unchanged unless all of the `st(*)` registers have been marked as clobbered.
+  - If all x87 registers are clobbered then the x87 register stack is guaranteed to be empty upon entering an `asm` block. Assembly code must ensure that the x87 register stack is also empty when exiting the asm block.
 - The requirement of restoring the stack pointer and non-output registers to their original value only applies when exiting an `asm!` block.
   - This means that `asm!` blocks that never return (even if not marked `noreturn`) don't need to preserve these registers.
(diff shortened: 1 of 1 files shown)


joshtriplett:
Can you clarify and spell out how the writer of an `asm!` block can clobber all x87 registers? Do they need to write out a clobber for each register individually? You wrote `st(*)` here; someone might expect that syntax to work. Does it, or could it?

Amanieu:
All 8 registers need to be explicitly clobbered, there is no wildcard syntax. The reason for this is that in the future we may decide to support x87 registers as proper operands. Also there is no precedent for a wildcard register name that covers multiple real registers.

joshtriplett:
> Also there is no precedent for a wildcard register name that covers multiple real registers.

In theory we could do that with something like `clobber_abi("x87")`, but perhaps that's not worth catering for.

joshtriplett (approved):

loveJesus:
Happy Easter and hello thanks for your work on rust. I am trying to implement the following in the new `asm!` syntax but i'm not really sure particularly about how to go about having the `fstp` command store into a rust variable. Thank you!
 
```rust
#[inline(always)]
pub fn cos_chirho(a: f32) -> f32 { 
let mut res: f32 = unsafe { mem::uninitialized() };

unsafe { asm!(
    r##"
        flds $1;
        fcos;
        fstps $0;
    "##
    : "=*m"(&mut res as *mut f32)
    : "*m"(&a as *const f32)
) };

res
}
```

Amanieu:
Just pass the memory address in a `reg` and then use it as normal:
```rust
asm!(
	"flds [{}]",
	"fcos",
	"fstps [{}]",
	in(reg) &a
	in(reg) &mut res,
);
```

loveJesus:
God is good! thank you.
Changed the `flds` to `fld dword ptr` and likewise for `fstp`. It works! 

I did not know however to use an `in(reg)` instead of `out(reg)` for the `&mut` parameter. It was giving me `(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)` when it would compile with `out(reg) *&mut`. It makes sense that an `in` can work since we modify the contents pointed to by the register and not the register itself. Sorry to have bothered you with this. Do you think me attempting to make a pull request with this example would be helpful?

Amanieu:
Rust-by-example might be a better place for some example code.

PR #1129, opened 2022-01-06
State in the UAX31 profile description that a lone `_` is not an identifier

mattheww:
This is following on from discussion in PR #1128.

UAX31 says an implementation that uses a variant of the default rules should

>  declare that it uses a **profile** and define that profile with a precise specification of the characters that are added to or removed from Start, Continue, and Medial and/or provide a list of additional constraints on identifiers. 

so I suppose this counts as such an additional constraint.


ehuss (approved):
Thanks!

PR #1133, opened 2022-01-08
Say that macros-by-example `ident` metavariables can match raw identifiers

### LINKED ISSUE #610: `ident` macro fragment should include raw keywords
The `ident` fragment is currently documented as matching IDENTIFIER_OR_KEYWORD.  However, I think it should also include RAW_IDENTIFIER.

mattheww:
Closes #610


ehuss (approved):
Thanks!

Issue #569, opened 2019-04-21
No-brace restrictions should mention that they are recursive.


ehuss:
There are various places in the grammar that restrict expressions with various types of braces.  These restrictions also apply recursively into the expressions. This should be captured somehow.  Examples:

if-expr.md mentions that the expression cannot be a struct expression:

```rust
if S{} {/*...*/}  // This is syntactically invalid.
```

However this applies recursively:

```rust
if x == S{} {/*...*/}  // This is a conditional expression with a struct inside it, also invalid.
```

Another example is that a final expression in _Statements_ must be an _ExpressionWithoutBlock_.  However, _ExpressionWithoutBlock_ does not capture that the block restriction also applies recursively.  For example `{{1}+{1}}` does not parse.

I don't know the exact rules here, presumably anything with brackets as the last expression?  Or parsing a block ends an expression?  This needs to be spelled out explicitly, and somehow captured in the grammar.

See https://github.com/rust-lang/rust/pull/59981.

Also, if, if-let, and match should explain *why* struct expressions aren't allowed, and it should indicate that parentheses can be used as a work-around.

NOTE: This applies to all the places where structs aren't allowed, like loop expressions.

See also https://github.com/rust-lang/rust/pull/59981 for some example tests that illustrate similar issues.

Also, it's not just tail expressions.  Method call expressions also can't be structs, as in `if S{}.foo() {}`.

Another example: `match () { () => 1 } + 2` the match requires parentheses.
Another example: `{{1}==0}` doesn't work, but `{1=={0}}` does.

Another example (see https://github.com/rust-lang/rust/pull/88690): `{"abc"}.len();` works but `{1}-1` does not.

ehuss:
I'm going to split this into two issues: https://github.com/rust-lang/reference/issues/1808 and https://github.com/rust-lang/reference/issues/1809, because these really are two separate concerns.

PR #1134, opened 2022-01-09
Restrict call expression grammar

absurdhero:
I made the minimal change to the doc for correctness but I welcome ideas for improvement if anyone feels
the below examples should be called out. I suspect that no additional descriptive text is warranted since
no one has noticed this issue before and commented on it.

This change restricts the CallExpression grammar to match rustc and to resolve a grammatical ambiguity.

CallExpression was described as accepting expressions (including ExpressionWithBlock) as a function operand.
However, rustc parses a block followed by a parenthetical expression as a sequence of Statements, not as a function call.

Furthermore, the `Expression ( CallParams? )` grammar leads to an ambiguity as shown:

`{expression}(param)`

This can be parsed as two statements in sequence or as a function call according to the documented grammar.

Or in this example using `if`:

`if true {String::new} else {String::new}(some_str)`

This can be rewritten as an unambiguous CallExpression by wrapping the if block in parenthesis:

`(if true {String::new} else {String::new})(some_str)`

### DIFF
--- a/src/expressions/call-expr.md
+++ b/src/expressions/call-expr.md
@@ -2,7 +2,7 @@
 
 > **<sup>Syntax</sup>**\
 > _CallExpression_ :\
-> &nbsp;&nbsp; [_Expression_] `(` _CallParams_<sup>?</sup> `)`
+> &nbsp;&nbsp; [_ExpressionWithoutBlock_] `(` _CallParams_<sup>?</sup> `)`
 >
 > _CallParams_ :\
 > &nbsp;&nbsp; [_Expression_]&nbsp;( `,` [_Expression_] )<sup>\*</sup> `,`<sup>?</sup>
@@ -87,6 +87,7 @@ Refer to [RFC 132] for further details and motivations.
 
 [RFC 132]: https://github.com/rust-lang/rfcs/blob/master/text/0132-ufcs.md
 [_Expression_]: ../expressions.md
+[_ExpressionWithoutBlock_]: ../expressions.md
 [`default()`]: ../../std/default/trait.Default.html#tymethod.default
 [`size_of()`]: ../../std/mem/fn.size_of.html
 [`std::ops::FnMut`]: ../../std/ops/trait.FnMut.html


pushkine:
The documentation is correct although that behavior is a confusing part of the syntax, you can read about it at the third paragraph of the [block expression](https://doc.rust-lang.org/reference/expressions/block-expr.html) reference

absurdhero:
Thanks for the reply, pushkine. I used blocks (the `BlockExpression` you linked to) as one example of an `ExpressionWithBlock` but there are others such as the `if` statement which I also demonstrated.

The third paragraph in the doc you linked is concerned with the semantics of blocks and their return value. But I'm actually talking about an issue with the syntactic description, not the semantics.

 I don't think any of the productions for ExpressionWithBlock are valid syntax as part of a function call. `BlockExpression`.

 Can you provide a counter-example where a function call can be expressed as a block followed by parenthesis?

In my testing, the compiler consistently interprets a block (or another kind of ExpressionWithBlock) followed `( CallParams? )` as two statements in sequence.

pushkine:
> Can you provide a counter-example where a function call can be expressed as a block followed by parenthesis?

```rs
fn main() {
	println!("{}", if true { || 1 } else { || 2 }())
}
```
https://replit.com/@pushkine/block-as-callee#src/main.rs

ehuss:
Sorry for not responding sooner.

So, the rules here are kind of a mess, and it is not entirely clear if things are working as intended.

Here is an example where a call expression can be a block expressions:

```rust
fn foo() {
    println!("hi");
}

fn main() {
    // `x` is a closure that calls `foo`.
    let x = ||{foo}();
    x();
}
```

I think some of these oddities are related to [`expr_is_complete`](https://github.com/rust-lang/rust/blob/08df8b81d6e723c66fee579d57d61dc60bc21fc1/compiler/rustc_parse/src/parser/expr.rs#L449-L453).

As a similar thing, #569 discusses an adjacent issue.  Part of the problem is the parser is stateful, and the state changes in different contexts.

So, in short, I'm not sure how to resolve this issue, but I don't think a change like this is entirely correct.

absurdhero:
Fascinating! Thanks for the examples, both of you.
This does look hairy to document and the reference doesn't seem like the right place for this level of detail.

I was dimly aware of #569 due to its impact on the rustypop grammar. I appreciate the pointer.

I'll close the PR since there isn't anything else to do.

PR #957, opened 2021-02-03
Document or-patterns

mark-i-m:
cc https://github.com/rust-lang/rust/pull/79278 https://github.com/rust-lang/rust/issues/54883 @joshtriplett 

@petrochenkov This also allows the leading vert change according to the T-lang FCP in https://github.com/rust-lang/rust/issues/81415

PR #1135, opened 2022-01-09
Document the 2021 edition changes to macros-by-example `pat` metavariables

### LINKED ISSUE #1072: 2021: Update for or-patterns changes
The [metavariables section](https://github.com/rust-lang/reference/blob/master/src/macros-by-example.md#metavariables) needs to be updated for the 2021 or-patterns changes. `pat` needs to be changed to `[_Pattern_]`, and a separate callout section for older behavior needs to be added. See the [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/or-patterns-macro-rules.md) for more details.

(I *think* that's all that needs to be done, but I am unsure. I'm also unsure on how to integrate this change. Typically we use "Edition Differences" callout blocks, but that seems like overkill here, but it seems strange to be inconsistent.)

mattheww:
Closes #1072

I've tried doing it with explicit "Edition Differences" blocks.

I've included a change under Follow-set Ambiguity Restrictions too.

I took some of the wording from parts of PR #957 which were left out at the time.

Do you think it's worth adding a note mentioning that the relevant edition
is the one in effect where the macro is defined, as opposed to invoked?

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -143,6 +143,8 @@ transcribed more than once or not at all.
 > **Edition Differences**: Starting with the 2021 edition, `pat` fragment-specifiers match top-level or-patterns (that is, they accept [_Pattern_]).
 >
 > Before the 2021 edition, they match exactly the same fragments as `pat_param` (that is, they accept [_PatternNoTopAlt_]).
+>
+> The relevant edition is the one in effect for the `macro_rules!` definition.
 
 ## Repetitions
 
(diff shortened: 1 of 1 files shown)


ehuss:
> Do you think it's worth adding a note mentioning that the relevant edition
> is the one in effect where the macro is defined, as opposed to invoked?

Yea, I think that would be good to add.   Thanks!

ehuss (approved):
Thanks!

PR #1136, opened 2022-01-15
Update the Preludes chapter for the 2021 edition changes to the standard library prelude

### LINKED ISSUE #1067: 2021: Update for prelude changes (RFC 3114)
I *think* the prelude documentation needs to be updated for the new edition behavior. See [RFC 3114](https://github.com/rust-lang/rfcs/blob/master/text/3114-prelude-2021.md) and the [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/prelude.md) for more. IIUC, all the references to `v1` in https://github.com/rust-lang/reference/blob/master/src/names/preludes.md need to be updated to the new model where the prelude depends on the edition. See also [`standard_library_imports.rs`](https://github.com/rust-lang/rust/blob/e916b7cb7708e470be8d0134bdf39479051a5c44/compiler/rustc_builtin_macros/src/standard_library_imports.rs#L67-L87) where this is implemented.

mattheww:
Closes #1067


src/names/preludes.md
```diff
@@ -18,9 +18,22 @@ There are several different preludes:
...
-from the [`core::prelude::v1`] module.
+Each crate has a standard library prelude, which consists of the names from a single standard library module. The module used depends on the crate's edition, and on whether the [`no_std` attribute] is applied to the crate:
+
+Edition | `no_std` not applied        | `no_std` applied
+--------| --------------------------- | ----------------------------
+2015    | [`std::prelude::rust_2015`] | [`core::prelude::rust_2015`]
+2018    | [`std::prelude::rust_2018`] | [`core::prelude::rust_2018`]
+2021    | [`std::prelude::rust_2021`] | [`core::prelude::rust_2021`]
+
+This choice of prelude is in effect throughout the crate, including code generated by macros defined in crates with a different edition or `no_std` attribute.
```
ehuss, line 29: I'm having a bit of trouble fully understanding what this sentence is trying to say.  Might I suggest to just remove it? The start of this chapter describes how every prelude is part of name resolution for every module.  Also, there isn't anything special about the standard library prelude with respect to macros.  

In the future, once [Name Resolution](https://doc.rust-lang.org/nightly/reference/names/name-resolution.html) is filled out, it might become more clear how names from preludes are resolved. Also, I think the [Hygiene section](https://doc.rust-lang.org/nightly/reference/macros-by-example.html#hygiene) explains how names are resolved in macros.

src/names/preludes.md
```diff
@@ -18,9 +18,22 @@ There are several different preludes:
 
 ## Standard library prelude
 
-The standard library prelude includes names from the [`std::prelude::v1`]
-module. If the [`no_std` attribute] is used, then it instead uses the names
-from the [`core::prelude::v1`] module.
+Each crate has a standard library prelude, which consists of the names from a single standard library module. The module used depends on the crate's edition, and on whether the [`no_std` attribute] is applied to the crate:
```
ehuss, line 21: Can you split sentences with newlines so that each sentence is on a line of its own?

src/names/preludes.md
```diff
@@ -130,13 +141,19 @@ This attribute does not affect the [language prelude].
 [`alloc`]: ../../alloc/index.html
 [`Box`]: ../../std/boxed/struct.Box.html
 [`core::prelude::v1`]: ../../core/prelude/index.html
+[`core::prelude::rust_2015`]: ../../core/prelude/index.html
+[`core::prelude::rust_2018`]: ../../core/prelude/index.html
+[`core::prelude::rust_2021`]: ../../core/prelude/index.html
```
ehuss, line 146: Is there a particular reason not to link the to the specific modules?  For example, [`rust_2021`](https://doc.rust-lang.org/core/prelude/rust_2021/index.html) precisely shows what is exported from the module.

I realize the original `v1` links went to the top-level module, but I honestly don't remember why I did that (I think linking to [`v1`](https://doc.rust-lang.org/1.25.0/std/prelude/v1/index.html) would also be more clear).

ehuss (commented):
Thanks!

src/names/preludes.md
mattheww: I've removed it

src/names/preludes.md
mattheww: Done

src/names/preludes.md
mattheww: I've changed them, including v1

ehuss (approved):
Thanks!

PR #1138, opened 2022-01-18
(minor) Remove Expression Path sub-types splits in Pattern specs

pushkine:
Removes complexity in `PathPattern` spec introduced by splitting `PathExpression` into its subcomponents `PathInExpression | QualifiedPathInExpression`, replaces the same segment by `PathPattern` in `RangePatternBound`.


ehuss (approved):
Thanks!

PR #1141, opened 2022-01-20
Link to associated constants section rather than glossary

XrXr:
I think the section about associated constants is more informative
than the glossary definition for the word "associated" in this case.


ehuss (approved):
Thanks!

PR #1143, opened 2022-01-24
Improve rendering

kirillbobyrev:
Span the link along the whole "const contexts", the text looks unnatural when the last `s` is left out of the hyperlink.


ehuss (approved):
Thanks!

PR #1144, opened 2022-01-25
Clarify where `impl Trait` can be used

WaffleLapkin:
This clarifies where exactly `impl Trait` can be used. This also includes changes from https://github.com/rust-lang/rust/pull/93082.


WaffleLapkin:
(CI fails because examples depend on rust-lang/rust#93082)

traviscross:
Seeing that:

- https://github.com/rust-lang/rust/pull/93082

...was closed, it seems we should just close this one also for now.  So we'll do that.

PR #1145, opened 2022-01-25
Fix micro typo in async/unsafe function docs

mlodato517:
**This Commit**
Changes "callers responsibility" to "caller's responsibility".

**Why?**
Just saw it while reading and figured it was a typo.


ehuss (approved):
Thanks!

PR #1147, opened 2022-01-27
Improve associated constant item CTFE timing section

XrXr:
@RalfJung pointed out in a [comment] that the previous phrasing of the
sentence can read like it is giving guarantees about free constant
definitions always undergoing CTFE, even when unused. That seems to be
how the compiler behaves right now, but it's unclear whether it's
intentional.

Be more precise and don't talk about free constants at all.

---

Related to this, I'm not sure if https://github.com/rust-lang/rust/issues/89006
stabilized the fact that *unamed constants* (in contrast to named but unused free constants) always undergo CTFE. It's part
of the issue, and it's used in the release notes to demo the feature, so I had assumed that it does. Going through the RFC and the issue though, it does not state it explicitly, so it's possible that nothing with regards to when CTFE is guaranteed to happen was actually stabilized.

I'm fairly new to the language overall, but maybe it's time that I try to write an RFC about the _whens_ of CTFE? 

[comment]: https://github.com/rust-lang/reference/pull/1120#discussion_r791928575

### DIFF
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -293,8 +293,8 @@ type that the definition has to implement.
 An *associated constant definition* defines a constant associated with a
 type. It is written the same as a [constant item].
 
-Unlike [free] constants, associated constant definitions undergo
-[constant evaluation] only when referenced.
+Associated constant definitions undergo [constant evaluation] only when
+referenced.
 
 ```rust
 struct Struct;
@@ -381,5 +381,4 @@ fn main() {
 [regular function parameters]: functions.md#attributes-on-function-parameters
 [generic parameters]: generics.md
 [where clauses]: generics.md#where-clauses
-[free]: ../glossary.md#free-item
 [constant evaluation]: ../const_eval.md
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -294,7 +294,8 @@ An *associated constant definition* defines a constant associated with a
 type. It is written the same as a [constant item].
 
 Associated constant definitions undergo [constant evaluation] only when
-referenced.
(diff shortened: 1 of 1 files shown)


RalfJung:
> I'm fairly new to the language overall, but maybe it's time that I try to write an RFC about the whens of CTFE?

If you want to help with this, I think the first step is to contact the lang team (e.g. in their Zulip stream) -- I am not sure if an RFC is even required; possibly an FCP (final comment period) is enough to turn this current implementation fact into a stable guarantee.

ehuss:
@oli-obk / @RalfJung Are you OK with this change?

src/items/associated-items.md
```diff
@@ -293,8 +293,8 @@ type that the definition has to implement.
 An *associated constant definition* defines a constant associated with a
 type. It is written the same as a [constant item].
 
-Unlike [free] constants, associated constant definitions undergo
-[constant evaluation] only when referenced.
+Associated constant definitions undergo [constant evaluation] only when
+referenced.
```
oli-obk, line 297: And only after monomorphization. Nothing is actually done on generic constants, even if we could attempt it nowadays

XrXr: I pushed [d0b33a2](https://github.com/rust-lang/reference/pull/1147/commits/d0b33a2990c305e8793a2e53dbc77b6edec1b018) and [37e3c15](https://github.com/rust-lang/reference/pull/1147/commits/37e3c15c52b31f3cda9fd359d3bb66a3e69bb5ff) to try to explain this.

oli-obk (approved):

ehuss (approved):
Thanks!

PR #1148, opened 2022-01-28
(minor) Add backticks to token

pushkine:
Viewing the repository locally on VSCode's markdown preview shows an error at that position

Code_2022-01-28_11-24-11

Adding backticks fixes the issue
Code_2022-01-28_11-24-19


JohnTitor (approved):
Good catch, thanks!

PR #1149, opened 2022-01-29
Add undocumented Punctuation token Tilde `~`

pushkine:
The token `~` is a valid punctuation token inside macro invocations and macro rule declarations: https://replit.com/@pushkine/Tilde-token#src/main.rs

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -596,7 +596,7 @@ usages and meanings are defined in the linked pages.
 | `#`    | Pound       | [Attributes]
 | `$`    | Dollar      | [Macros]
 | `?`    | Question    | [Question mark operator][question], [Questionably sized][sized], [Macro Kleene Matcher][macros]
-| `~`    | Tilde       | The tilde operator is deprecated, but its token may still be used
+| `~`    | Tilde       | The tilde operator has been unused since before Rust 1.0, but its token may still be used
 
 ## Delimiters
 
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -596,6 +596,7 @@ usages and meanings are defined in the linked pages.
 | `#`    | Pound       | [Attributes]
 | `$`    | Dollar      | [Macros]
 | `?`    | Question    | [Question mark operator][question], [Questionably sized][sized], [Macro Kleene Matcher][macros]
+| `~`    | Tilde       | The tilde operator is deprecated, but its token may still be used
 
 ## Delimiters
 


ehuss:
Thanks, this has been on my todo list.

Since this is new content that was never explicitly decided upon, I'm going to check with the libs team.

@rust-lang/lang The `~` token has been part of rust since the very early days as part of the old box syntax.  It was never removed or feature gated.  Since it has been on stable ever since, I think we should include it as a documented token since it can be used in macros.  I also believe there is some interest in using it with `~const`.

@rfcbot fcp merge

joshtriplett:
People can already use it in stable code, so this isn't so much a decision we can make as an acknowledgement of the current state of things; changing that would be a breaking change. Documenting this seems reasonable.

Separately, we might ask whether we can usefully use `~` for some new feature, but that won't remove the ability to use the token in macros.

(As an aside, speaking only for myself and not others on the lang team, I really hope that `~const` is not the final syntax for that mechanism; I'd much rather use a different token for that.)

ehuss:
Can someone at @rust-lang/lang issue the `@rfcbot fcp merge` command? I think rfcbot only allows that command for members of that specific team.

ehuss:
Yea, I think it should be documented. The reason I wanted to bring it up is:

A. Get awareness of this.
B. An alternative to documenting is to future-incompat sweep it under the rug (since this wasn't ever documented or necessarily intended).  I doubt that is worth the hassle.
C. Check on the wording here. I'm not sure if "deprecated" is really the correct term.  More like "unused" maybe?

src/tokens.md
```diff
@@ -596,6 +596,7 @@ usages and meanings are defined in the linked pages.
 | `#`    | Pound       | [Attributes]
 | `$`    | Dollar      | [Macros]
 | `?`    | Question    | [Question mark operator][question], [Questionably sized][sized], [Macro Kleene Matcher][macros]
+| `~`    | Tilde       | The tilde operator is deprecated, but its token may still be used
```
joshtriplett, line 599:
```suggestion
| `~`    | Tilde       | The tilde operator has been unused since before Rust 1.0, but its token may still be used
```

joshtriplett:
@rfcbot merge

scottmcm:
I agree that trying to change the tokenizing rules here wouldn't be valuable enough to be worth the effort.

ehuss (approved):
Thanks!

PR #1150, opened 2022-01-30
Add undocumented outer attributes above StructExpr fields

pushkine:
`StructExprField` may use outer attributes: https://replit.com/@pushkine/Attributes-in-Struct-Expressions#src/main.rs

Here's an occurrence in rust's compiler:

https://github.com/rust-lang/rust/blob/a00e130dae74a213338e2b095ec855156d8f3d8a/compiler/rustc_data_structures/src/sync.rs#L571-L577

### DIFF
--- a/src/expressions/struct-expr.md
+++ b/src/expressions/struct-expr.md
@@ -18,7 +18,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; &nbsp;&nbsp; [IDENTIFIER]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | ([IDENTIFIER] | [TUPLE_INDEX]) `:` [_Expression_]\
 > &nbsp;&nbsp; )
-> 
+>
 > _StructBase_ :\
 > &nbsp;&nbsp; `..` [_Expression_]
 >
--- a/src/expressions/struct-expr.md
+++ b/src/expressions/struct-expr.md
@@ -124,7 +124,7 @@ let a = Gamma;  // Gamma unit value.
 let b = Gamma{};  // Exact same value as `a`.
 ```
 
-[_OuterAttribute_]: attributes.md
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!  Looks like this was implemented in #38814 and stabilized in #42656.

pushkine:
@ehuss it's actually indirectly documented at the bottom of the expression index page 
https://doc.rust-lang.org/reference/expressions.html#expression-attributes 

Note that `CallExpression`'s `CallParams` and `StructExprTuple`'s members are also missing outer attributes in their reference syntax, I was not comfortable editing them myself since it requires introducing new pseudo AST nodes

ehuss:
Sorry, I'm not sure I understand your comment.  The `Expression` rule begins with `OuterAttribute`, with a marker to that section that specifies exactly where attributes are allowed. Thus something like `CallParams` doesn't need to specify `OuterAttribute` because it is part of `Expression`.

pushkine:
From what I understand Expressions cannot have attributes on their own, it's only their position as the child of another specific node that grants them that ability

I think it is very confusing to put a blanket `OuterAttribute* (...)` around the syntax definition of all expressions with a foot note listing the AST context that allows that to happen, I'd expect that information to be in the definition of each of those parents instead

ehuss:
Ah, yea, I can understand that it can be confusing.  The grammar in the reference defines the AST, which is essentially the parsed structure before certain validations are done.  The reason it is done this way is primarily for macros, which work with AST fragments.  For example, the following is valid:

```rust
macro_rules! expr {
    ($e:expr) => {};
}

fn main() {
    // Attributes are valid before any expression for `expr`.
    expr!(#[some_attr]1 + #[some_attr]2);
    expr!([
        #[some_attr]1,
        #[some_attr]2,
        #[some_attr]3,
    ]);
}
```

In this situation, attribute validation is done in a later phase.

PR #1154, opened 2022-02-01
Document pre-Rust-2021 special case for IntoIterator method lookup

### LINKED ISSUE #1068: 2021: Update for into_iter changes
I *think* the [iterator loops](https://github.com/rust-lang/reference/blob/master/src/expressions/loop-expr.md#iterator-loops) section needs to be updated for the new edition behavior. See the [edition guide chapter](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2021/IntoIterator-for-arrays.md) for more. IIUC, there is edition-specific behavior that changes how the language interfaces with the standard library for arrays specifically.

mattheww:
Closes #1068

This behaviour is also documented in the standard library docs for the [array type][1].

[1]: https://github.com/rust-lang/rust/blob/ad88831cd50ffe9cb9006bbdcb7fc9d97142e410/library/std/src/primitive_docs.rs#L585

The edition guide doesn't say that the special-case behaviour applies when the array type is reached via deref coercions, but this is clear from the implementation. It's also mentioned in https://github.com/rust-lang/rust/pull/84147#issuecomment-819000436.

### DIFF
--- a/src/expressions/method-call-expr.md
+++ b/src/expressions/method-call-expr.md
@@ -66,6 +66,11 @@ Once a method is looked up, if it can't be called for one (or more) of those rea
 If a step is reached where there is more than one possible method, such as where generic methods or traits are considered the same, then it is a compiler error.
 These cases require a [disambiguating function call syntax] for method and function invocation.
 
+> **Edition Differences**: Before the 2021 edition, during the search for visible methods, if the candidate receiver type is an [array type], methods provided by the standard library [`IntoIterator`] trait are ignored.
+>
+> The edition used for this purpose is determined by the token representing the method name.
+
+
 <div class="warning">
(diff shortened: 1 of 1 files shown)


est31:
IMO this should also mention that this behaviour might be changed in the future, see https://github.com/rust-lang/rust/pull/84590 .

mattheww:
I've added a sentence saying that.

I observe that `CONTRIBUTING.md` says "Documentation is written as if the current stable release of Rust is the last release", but I see there are a number of existing exceptions of this sort.

ehuss (approved):
Thanks, looks good!

This is dancing precariously with the divide between the language and the standard library, and it is a bit unfortunate to include.

PR #1156, opened 2022-02-03
Document let else statements

est31:
For the stabilization of `let else` statements.

Tracking issue: https://github.com/rust-lang/rust/issues/87335

### DIFF
--- a/src/statements.md
+++ b/src/statements.md
@@ -54,16 +54,26 @@ fn outer() {
 > **<sup>Syntax</sup>**\
 > _LetStatement_ :\
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
->     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] )<sup>?</sup> `;`
+>     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] [†](#let-else-restriction)
+>     ( `else` [_BlockExpression_]) <sup>?</sup> ) <sup>?</sup> `;`
+>
+> <span id="let-else-restriction">† When an `else` block is specified, the
+> _Expression_ must not be a [_LazyBooleanExpression_], or end with a `}`.</span>
 
-A *`let` statement* introduces a new set of [variables], given by an
-irrefutable [pattern]. The pattern is followed optionally by a type
-annotation and then optionally by an initializer expression. When no
-type annotation is given, the compiler will infer the type, or signal
+A *`let` statement* introduces a new set of [variables], given by a [pattern].
+The pattern is followed optionally by a type annotation and then either ends,
+or is followed by an initializer expression plus an optional `else` block.
+When no type annotation is given, the compiler will infer the type, or signal
 an error if insufficient type information is available for definite
 inference. Any variables introduced by a variable declaration are visible
 from the point of declaration until the end of the enclosing block scope,
 except when they are shadowed by another variable declaration.
+
+If an `else` block is not present, the pattern must be irrefutable.
+If an `else` block is present, the pattern may be refutable.
+If the pattern does not match (this requires it to be refutable), the `else`
+block is executed.
+The `else` block must always diverge (evaluate to the [never type]).
 
 ## Expression statements
 
@@ -121,6 +131,7 @@ statement are [`cfg`], and [the lint check attributes].
 [function]: items/functions.md
 [item]: items.md
 [module]: items/modules.md
+[never type]: types/never.md
 [canonical path]: paths.md#canonical-paths
 [implementations]: items/implementations.md
 [variables]: variables.md
@@ -128,9 +139,11 @@ statement are [`cfg`], and [the lint check attributes].
 [`cfg`]: conditional-compilation.md
 [the lint check attributes]: attributes/diagnostics.md#lint-check-attributes
 [pattern]: patterns.md
+[_BlockExpression_]: expressions/block-expr.md
 [_ExpressionStatement_]: #expression-statements
 [_Expression_]: expressions.md
 [_Item_]: items.md
+[_LazyBooleanExpression_]: expressions/operator-expr.md#lazy-boolean-operators
 [_LetStatement_]: #let-statements
 [_MacroInvocationSemi_]: macros.md#macro-invocation
 [_OuterAttribute_]: attributes.md
--- a/src/statements.md
+++ b/src/statements.md
@@ -75,6 +75,17 @@ If the pattern does not match (this requires it to be refutable), the `else`
 block is executed.
 The `else` block must always diverge (evaluate to the [never type]).
 
+```rust
+let (mut v, w) = (vec![1, 2, 3], 42); // The bindings may be mut or const
+let Some(t) = v.pop() else { // Refutable patterns require an else block
+    panic!(); // The else block must diverge
+};
+let [u, v] = [v[0], v[1]] else { // This pattern is irrefutable, so the compiler
+                                 // will lint as the else block is redundant.
+    panic!();
+};
+```
+
 ## Expression statements
 
 > **<sup>Syntax</sup>**\
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -59,8 +59,8 @@ fn answer_to_life_the_universe_and_everything() -> i32 {
 
 ## Function parameters
 
-As with `let` bindings, function parameters are irrefutable [patterns], so any
-pattern that is valid in a let binding is also valid as a parameter:
+Function parameters are irrefutable [patterns], so any pattern that is valid in
+an else-less `let` binding is also valid as a parameter:
 
 ```rust
 fn first((value, _): (i32, i32)) -> i32 { value }


src/statements.md
```diff
@@ -65,6 +66,25 @@ inference. Any variables introduced by a variable declaration are visible
...
+
+### `let else` statements
+
+> **<sup>Syntax</sup>**\
+> _LetElseStatement_ :\
+> &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
+>     ( `:` [_Type_] )<sup>?</sup> `=` [_Expression_] `else` [_BlockExpression_] `;`
+
+A *`let else` statement* introduces a new set of [variables], given by a
+refutable [pattern]. The pattern is followed optionally by a type
```
Fishrock123, line 78: Is this true for nested type values? I am a little unsure how the type annotation feature works

```rust
// If it were this, this documentation statement would be true:
let Ok(v): Result<u32, _> = expr else { };

// _I thought_ it as supposed to be able to be like this though, but maybe I am wrong:
let Ok(v: u32) = expr else { };
```

Fishrock123 (commented):
I would probably write this documentation a bit differently. I think to the developer there is not much different between `let` and `let else`. Consider that you can already match against irrefutable pattern in `let`, and this is essentially just an extension to `let` in practice.

But perhaps it's direct relation to `let` as a natural extension belongs more in a guide level explanation.

est31:
> Consider that you can already match against irrefutable pattern in let, and this is essentially just an extension to let in practice.

Originally I tried to extend the `let` section as you suggest. There are commonalities, but there are also a lot of differences. For `let`, the pattern *must* be irrefutable, for `let else` it *must* be refutable. For `let`, the `= expr` is optional, for `let else` it isn't. Explaining those differences would have made it harder to explain/comprehend how the thing works in the first place.

src/statements.md
cormacrelf: No, let-else has the same type annotation syntax as let, not sure where you got that idea from.

cormacrelf:
> for `let else` it *must* be refutable.

No, by default it is only a warning to have an irrefutable binding in there.

```
warning: irrefutable `let...else` pattern
 --> src/main.rs:4:5
  |
4 | /     let x = 5 else {
5 | |         return;
6 | |     };
  | |______^
  |
  = note: `#[warn(irrefutable_let_patterns)]` on by default
  = note: this pattern will always match, so the `else` clause is useless
  = help: consider removing the `else` clause

```

est31:
Good point, it's only a warning. Should I change the text of the reference to extend it to all patterns?

cormacrelf:
Yes. But refutable patterns deserve a mention, something along the lines of

> If the pattern is a refutable pattern, then the else block will be executed when the pattern is not matched. The else block must escape [/return/whatever we call it] ...

I haven't read much of the reference before so not sure whether it's appropriate to mention that refutable patterns are the whole point of it.

src/statements.md
Fishrock123: Oh sorry I thought that was part of the upcoming "type ascription" for some reason.

src/statements.md
```diff
@@ -75,11 +75,13 @@ except when they are shadowed by another variable declaration.
...
-or signal an error if insufficient type information is available for definite
-inference. It is is then followed by an initializer expression as well as a
-diverging `else` block.
+[pattern]. The pattern is followed optionally by a type annotation. When no type
+annotation is given, the compiler will infer the type, or signal an error if
+insufficient type information is available for definite inference. It is is
+then followed by an initializer expression as well as a diverging `else` block.
+Compared to `let` statement patterns which have to be irrefutable, `let else`
+patterns can be both irrefutable and refutable. However, as it is besides the
+point of, irrefutable patterns in `let else` are linted against.
```
Fishrock123, line 84: `as it is besides the point of` reads really strange...

Something like this would be better:
```
However, irrefutable patterns in `let else` are linted against due to their lack of usefulness (the `else` would be dead code).
```

est31: What about "Irrefutable patterns are not forbidden, however there is a lint for them as the `else` clause is redundant"?

Fishrock123: That sounds better.

est31: I've applied a slightly different version, but one that avoids the "besides the point" formulation.

Fishrock123: Looks good to me

src/statements.md
```diff
@@ -65,6 +66,27 @@ inference. Any variables introduced by a variable declaration are visible
...
+> &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
+>     ( `:` [_Type_] )<sup>?</sup> `=` [_Expression_] `else` [_BlockExpression_] `;`
+
+A *`let else` statement* introduces a new set of [variables], given by a
+[pattern]. The pattern is followed optionally by a type annotation. When no type
+annotation is given, the compiler will infer the type, or signal an error if
+insufficient type information is available for definite inference. It is is
+then followed by an initializer expression as well as a diverging `else` block.
+Compared to `let` statement patterns which have to be irrefutable, `let else`
+patterns can be both irrefutable and refutable. In the case of irrefutable
```
digama0, line 83:
```suggestion
patterns can be either irrefutable or refutable. In the case of irrefutable
```

est31:
With let else stable, this is ready for review!

ehuss:
I think I too would prefer to see this defined as a single statement. I understand it can be a little awkward since it changes the semantics quite a bit, but I think it shouldn't be too difficult to describe. As it is, it looks like there is a lot of duplication.

To make it fit in the original `let` statements section, I think it should be sufficient to:
* Remove the word "irrefutable".
* Add a paragraph describing the `else` block behavior. Here's a rough draft, but feel free to adjust:

When an `else` block follows the initializer, then the pattern may be a refutable pattern. 
When the pattern does not match, the `else` block is executed.
The type of the `else` block must be the [never type](types/never.md).
If there is no `else` block, then the pattern must be irrefutable.

* The grammar for _Expression_ in the initializer needs to reject forbidden expressions. I would probably just add a dagger (†) that links to a sentence in the grammar section that explains it.  Like:

> **<sup>Syntax</sup>**\
> … [_Expression_] [†](#let-else-restriction) …
> 
> <span id="let-else-restriction">† When an `else` block is specified, the _Expression_ must not be a [_LazyBooleanExpression_](expressions/operator-expression.md#lazy-boolean-operators) or end with a `}`.</span>

I'm also a bit surprised this is how it works, as this adds unprecedented complexity to the grammar and I think would be quite difficult for many parsers.

* Add a note about irrefutability:

> **Note**: `rustc` will emit a warning when using an irrefutable pattern with an `else` block since that makes it impossible for the `else` block to execute.

* Add an example. Illustrate both a basic `let` binding, and one with an `else` block.

est31:
@ehuss okay I gave up, I refactored the section so that it is unified with `let`. My second attempt of unifying feels less complicated than what I remember of my first attempt, so this was a good idea already, and I'm quite open towards this approach now.

src/statements.md
```diff
@@ -54,17 +54,36 @@ fn outer() {
...
+or is followed by an initializer expression plus an optional `else` block.
+When no type annotation is given, the compiler will infer the type, or signal
 an error if insufficient type information is available for definite
 inference. Any variables introduced by a variable declaration are visible
 from the point of declaration until the end of the enclosing block scope,
 except when they are shadowed by another variable declaration.
 
+If an `else` block is not present, the pattern has to be irrefutable.
+If an `else` block is present, the pattern may be refutable, and if it is, it
+will be checked on run time. On a mismatch, the `else` block is executed.
```
ehuss, line 71: I feel like specifying "run time" here may be over-specifying the behavior. For example, a `let` in a `const` will be evaluated at compile-time. It also isn't specifically stated what it means "to check", which isn't a verb typically used in the reference with pattern matching. These seem like two clauses to me, so I think it would be fine to state them separately, such as:

```suggestion
If an `else` block is present, the pattern may be refutable.
If the pattern does not match the initializer expression, the `else` block is executed.
```

src/statements.md
```diff
@@ -54,17 +54,36 @@ fn outer() {
...
+The `else` block has to always diverge (evaluate to type `!`).
+If an `else` block is present, restrictions apply on the expression:
+It might not be a [_LazyBooleanExpression_], or end in a `}` token.
```
ehuss, line 74: I think this should be up in the grammar section since it is part of the grammar.

Also, similar to the must/should/may terminology:

```suggestion
If an `else` block is present, restrictions apply on the expression:
It must not be a [_LazyBooleanExpression_], or end in a `}` token.
```

Although I do lean towards the earlier wording I suggested. I think it can sometimes be a little difficult to follow when using "it" when it is not very clear what "it" is. Here it is referring to an "expression", but there are different expressions at play (such as the block expression, and all the expressions within the block). Up in the grammar, the word _Expression_ refers to a specific expression (there is only one of that form; that is a rule with a capital E and italicized).

src/statements.md
```diff
@@ -54,17 +54,36 @@ fn outer() {
 > **<sup>Syntax</sup>**\
 > _LetStatement_ :\
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
->     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] )<sup>?</sup> `;`
+>     ( `:` [_Type_] )<sup>?</sup> ( | `=` [_Expression_] | `=` [_Expression_]
+>     `else` [_BlockExpression_]) `;`
```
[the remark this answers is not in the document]
est31: Do you know how to do the note in source code?

src/statements.md
est31: const-evaluation still has a "run-time" and compile time separation. But shrug.

src/statements.md
est31: I think it's clear because there is only one expression in the grammar. It is also mentioned in the first paragraph.

src/statements.md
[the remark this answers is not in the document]
est31: (also yes the leading | was to indicate the epsilon case, but I have changed it to your suggestion)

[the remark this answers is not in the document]
ehuss: Oh, sorry, I should have included the source from the comment above:

```
> **<sup>Syntax</sup>**\
> … [_Expression_] [†](#let-else-restriction) …
> 
> <span id="let-else-restriction">† When an `else` block is specified, the _Expression_ must not be a [_LazyBooleanExpression_](expressions/operator-expression.md#lazy-boolean-operators) or end with a `}`.</span>
```

Just make it a link, and use a `<span>` to set the target.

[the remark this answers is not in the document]
est31: Thanks! I've tried out your code, it causes an error in CI:

> error in ../src/statements.md (line 57): broken Reference link (reference `†`)
some style checks failed

Apparently pulldown-cmark does not parse the inline html.

I have tried using a footnote, but that shows a <sup>1</sup>, which in this context can be very easily misunderstood as a repetition:
Screenshot_20220924_031548

However, it is possible to hide the number and put a † on its stead. I've pushed a commit to do that.

Havvy (changes requested):
On https://doc.rust-lang.org/reference/items/functions.html#function-parameters we mention that let patterns are always irrefutable. This is no longer true, so that section needs to be updated as well.

est31:
@Havvy good catch! I've updated that sentence in the last commit.

src/statements.md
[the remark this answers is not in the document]
ehuss: I would prefer to not change the behavior of footnotes. Those are used in other places of the reference.

I would also prefer to avoid using them in this situation. This isn't really a footnote, but more of an attempt to stuff a bunch of information in a small amount of space.

I tried the suggestion I posted above, and the `style-check` and the linkchecker both passed.  It should look exactly like:

```md
> **<sup>Syntax</sup>**\
> _LetStatement_ :\
> &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
>     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] [†](#let-else-restriction)
>     ( `else` [_BlockExpression_]) <sup>?</sup> ) <sup>?</sup> `;`
>
> <span id="let-else-restriction">† When an `else` block is specified, the _Expression_ must not be a [_LazyBooleanExpression_](expressions/operator-expr.md#lazy-boolean-operators) or end with a `}`.</span>
```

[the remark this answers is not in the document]
est31: @ehuss regarding the CI failure, please see [this](https://github.com/rust-lang/reference/actions/runs/3116465400/jobs/5054319422) CI run, ran from [this](https://github.com/rust-lang/reference/pull/1156/commits/3eacf0029c5298c481fca9db1e1adf5d313ed293) commit. If you click on the prev button you will get to the commit that has precisely the state that you suggest.

```
    Compiling style-check v0.1.0 (/home/runner/work/reference/reference/style-check)
    Finished dev [unoptimized + debuginfo] target(s) in 37.63s
     Running `target/debug/style-check ../src`
error in ../src/statements.md (line 57): broken Reference link (reference `†`)
some style checks failed
```

```markdown
> &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
>     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] )<sup>?</sup> `;`
>     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_][†](#let-else-restriction)
>     ( `else` [_BlockExpression_]) <sup>?</sup> ) <sup>?</sup> `;`
A *`let` statement* introduces a new set of [variables], given by an
<span id="let-else-restriction">† When an `else` block is specified, the
irrefutable [pattern]. The pattern is followed optionally by a type
_Expression_ must not be a [_LazyBooleanExpression_], or end with a `}`.</span>
```

I can try it again, but I doubt it will be any different.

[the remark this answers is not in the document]
ehuss: That's because in that commit you have:

```
[_Expression_][†](#let-else-restriction)
```

The text I have above has a space:

```
[_Expression_] [†](#let-else-restriction)
```

Without the space, markdown thinks the `[†]` is a reference link for the preceding `[_Expression_]`.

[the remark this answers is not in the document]
est31: Ohhhh... I see... Will adjust it right away.

src/statements.md
```diff
@@ -54,17 +54,38 @@ fn outer() {
...
 > &nbsp;&nbsp; [_OuterAttribute_]<sup>\*</sup> `let` [_PatternNoTopAlt_]
->     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] )<sup>?</sup> `;`
+>     ( `:` [_Type_] )<sup>?</sup> (`=` [_Expression_] [†](#let-else-restriction)
+>     ( `else` [_BlockExpression_]) <sup>?</sup> ) <sup>?</sup> `;`
 
-A *`let` statement* introduces a new set of [variables], given by an
-irrefutable [pattern]. The pattern is followed optionally by a type
-annotation and then optionally by an initializer expression. When no
-type annotation is given, the compiler will infer the type, or signal
+<span id="let-else-restriction">† When an `else` block is specified, the
```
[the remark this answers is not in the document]
est31: Done

ehuss (approved):
Thanks for the patience with the reviews.

est31:
No problem! It is great that consistency is important to you.

PR #1157, opened 2022-02-04
Documentation PR for cfg_panic

cchiw:
[issue](https://github.com/rust-lang/rust/issues/77443)

### DIFF
--- a/src/attributes/testing.md
+++ b/src/attributes/testing.md
@@ -81,6 +81,50 @@ fn mytest() {
     assert_eq!(1, 2, "values don't match");
 }
 ```
+### Different panic strategies
+
+The `cfg_panic` feature makes it possible to exercise different lines of code depending on the panic strategy. 
+The possible value is either `unwind` or `abort`.
+The following is a playful example on choosing the right beverage. 
+
+```toml
+#![feature(cfg_panic)]
+
+#[cfg(panic = "unwind")]
+fn ah(){ println!("Spit it out!!!!");}
+
+#[cfg(not(panic="unwind"))]
+fn ah(){ println!("This is not your party. Run!!!!");}
+
+fn drink(beverage: &str){
+   if beverage == "lemonade"{ ah();}
+   else{println!("Some refreshing {} is all I need.", beverage);}
+}
+
+fn main(){
+   drink("water");
+   drink("lemonade");
+}
+```
+
+Here is the same example rewritten.
+```toml
+#![feature(cfg_panic)]
+
+fn drink(beverage: &str) {
+   // You shouldn't drink too much sugary beverages.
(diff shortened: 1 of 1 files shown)


src/attributes/testing.md
```diff
@@ -81,6 +81,48 @@ fn mytest() {
     assert_eq!(1, 2, "values don't match");
 }
 ```
+## Different panic strategies
+
+The `cfg_panic` feature makes it possible to exercise different lines of code depending on the panic strategy. 
+The possible value is either `unwind` or `abort`.
```
bjorn3, line 87: Should this say make clear that code compiled with `#[cfg(panic = "unwind")]` being true may still aborting on panic if any crate is compiled with `-Cpanic=abort`?

cchiw: This is a good thing to point out.  Note, I'm not as familiar with crates so i don't know too much about this but I'll take your word for it.

I updated the file, let me know what you think :)

src/attributes/testing.md
```diff
@@ -124,6 +125,9 @@ fn main() {
 }
 ```
 
+Note that when the code is compiled with a certain panic strategy it still might do something different if a crate was compiled with a different strategy.
```
bjorn3, line 128:
```suggestion
Note that when the code is compiled with a certain panic strategy it still might do something different if another crate was compiled with a different strategy.
```

cchiw:
Hi, just following up. How would you like to proceed on this PR? Should I adjust it or would you like to not include it? The  [issue ](https://github.com/rust-lang/rust/issues/77443) is still open pending this thread.  

Just FYI, the compiler changes will be accepted to version 1.60, the Rust by Example documentation was merged in, and the Rust Book chose not to include it.

ehuss:
Can you update the PR to instead add a section in the Conditional Compilation chapter?  There are several sections there with examples of how `cfg` values are documented, and you can follow that format for adding `cfg_panic`.  Thanks!

ehuss:
Sorry if I wasn't clear.  The documentation should look like the other sections in the conditional compilation chapter.  For example, there should be a section (probably just under the `proc_macro` one) that looks roughly like this:

----

### `panic`

Key-value option set for …

Example values:

* `"abort"`
* `"unwind"`

----

Also, the example using `#[cfg(not(panic="unwind"))]`  is using a form that is assuming there are only two panic strategies. There is a note [here](https://github.com/rust-lang/rust/pull/93658#issuecomment-1032940963) where it was stated that we should explicitly avoid that.  I think the text here should very briefly note that other values may be added in the future.

cchiw:
Ah, OK. I added the suggested section.

ehuss (approved):
Thanks!

PR #1158, opened 2022-02-07
Update patterns allowed in @ patterns

pushkine:
Changes the allowed right hand side pattern from `Pattern` to `PatternNoTopAlt`

`ident @ A | B` is always parsed as `[ident @ [A]] | [B]`
Even using the optional heading `|` throws an error
```rs
match () {
    a @ | _ | _ => {}
//      ^ expected a binding pattern
}
```


ehuss (approved):
Thanks!

PR #1160, opened 2022-02-08
`LifetimeWhereClauseItem` can have missing `LifetimeBounds`

Veykril:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7509c10ee125dbeec7f77ce5bb1d9295:
```rs
fn __<'a>() where 'a: {}
```
is valid rust


Veykril:
Ah sorry, I read the definition of `LifetimeTimeBounds` incorrectly, it already covers this.

PR #1161, opened 2022-02-08
Fix link to the Bastion of the Turbofish

SabrinaJewson:
Since <https://github.com/rust-lang/rust/pull/89488> it was moved to a different directory.

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -322,7 +322,7 @@ example of an uninhabited type is the [never type] `!`, or an enum with no varia
 [structs]: items/structs.md
 [trait objects]: types/trait-object.md
 [traits]: items/traits.md
-[turbofish test]: https://github.com/rust-lang/rust/blob/master/src/test/ui/bastion-of-the-turbofish.rs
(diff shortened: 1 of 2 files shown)


src/glossary.md
```diff
@@ -322,7 +322,7 @@ example of an uninhabited type is the [never type] `!`, or an enum with no varia
 [structs]: items/structs.md
 [trait objects]: types/trait-object.md
 [traits]: items/traits.md
-[turbofish test]: https://github.com/rust-lang/rust/blob/master/src/test/ui/bastion-of-the-turbofish.rs
+[turbofish test]: https://github.com/rust-lang/rust/blob/master/src/test/ui/parser/bastion-of-the-turbofish.rs
```
ehuss, line 325: Thanks!  Perhaps we should make this a permanent link that won't change?  Such as:

```suggestion
[turbofish test]: https://github.com/rust-lang/rust/blob/1.58.0/src/test/ui/parser/bastion-of-the-turbofish.rs
```

SabrinaJewson: Good idea! I also went ahead and the other `rust-lang/rust` link to a permalink

PR #1162, opened 2022-02-08
Update statements syntax

pushkine:
Statements syntax:

* Removed `MacroInvocationSemi` since it is already a part of `Item`
* Restricted `ExpressionWithoutBlock` to `ExpressionWithoutBlock`<sub> `except MacroInvocation with curly brackets` </sub>

If a statement begins with a MacroInvocation with curly brackets, it is always parsed as a `MacroInvocationSemi` `Item`

Example:
```rs
macro_rules! a {
	()=>{""}
}
fn main() {
	  a!().to_string();
	  a![].to_string();
	  let a = a!{}.to_string();
	  a!{}.to_string();
	  //  ^ [rustc] expected expression, found `.`
}
```
https://replit.com/@pushkine/QuietPriceySlope#src/main.rs

### DIFF
--- a/src/statements.md
+++ b/src/statements.md
@@ -5,7 +5,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; `;`\
 > &nbsp;&nbsp; | [_Item_]\
 > &nbsp;&nbsp; | [_LetStatement_]\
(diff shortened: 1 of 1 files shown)


pushkine:
It appears the issue only occurs on `repl.it`, it indeed compiles fine. I apologize for the confusion

PR #1163, opened 2022-02-09
Clarify guarantees provided by repr(packed)

joshlf:
Clarify that `repr(packed)` minimizes inter-field padding and, in the special case of `repr(packed(1))`/`repr(packed)`, guarantees no inter-field padding.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -515,6 +515,11 @@ For `packed`, if the specified alignment is greater than the type's alignment
 without the `packed` modifier, then the alignment and layout is unaffected.
 The alignments of each field, for the purpose of positioning fields, is the
 smaller of the specified alignment and the alignment of the field's type.
+Inter-field padding is guaranteed to be the minimum required in order to
+satisfy each field's (possibly altered) alignment (although note that, on its
+own, `packed` does not provide any guarantee about field ordering). An 
+important special case is `#[repr(packed(1))]` (or `#[repr(packed)]`), which
+guarantees that no inter-field padding will be present.
 
 The `align` and `packed` modifiers cannot be applied on the same type and a
 `packed` type cannot transitively contain another `align`ed type. `align` and
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -501,7 +501,7 @@ assert_eq!(std::mem::size_of::<Enum16>(), 4);
 
 The `align` and `packed` modifiers can be used to respectively raise or lower
 the alignment of `struct`s and `union`s. `packed` may also alter the padding
-between fields.
+between fields (although it will not alter the padding inside of any field).
 
 The alignment is specified as an integer parameter in the form of
 `#[repr(align(x))]` or `#[repr(packed(x))]`. The alignment value must be a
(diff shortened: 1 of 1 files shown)


Gankra:
See also https://github.com/rust-lang/reference/pull/1152 which is similarly trying to clarify repr details which are seemingly-obvious but worth stating explicitly.

joshlf:
> See also #1152 which is similarly trying to clarify repr details which are seemingly-obvious but worth stating explicitly.

Thanks, that's awesome to see!

joshlf:
What are the next steps to get this approved/merged?

ehuss:
I'm not entirely certain what this is trying to clarify.  If the fields are aligned in a certain way, the size of the padding falls out as a consequence of the field layout. In what way would there be more padding than what is required?  I also don't see `packed(1)` as being much of a special case, or at least not different in the layout method.

djkoloski:
`repr(Rust)` structs explicitly do not make any guarantees about data layout. Under the previous wording, a `repr(Rust, packed)` struct could have inter-field padding as `packed` only "may also alter the padding between fields". This updated wording guarantees that `repr(Rust, packed)` structs have no inter-field padding and other `packed(N)` reprs have the minimum amount of padding to satisfy the repr.

ehuss:
Perhaps the "no guarantees" is worded too specifically, and #1152 is trying to soften that.  Using `packed` with the default representation still follows the rules set out in the `packed` section.  

The note about "may also alter padding between fields" is more of a heads-up warning that `packed(N)` may change the offsets of the fields and *as a consequence* changes to padding may occur. The real rules are specified in the paragraph below.  It is not a statement that padding may be arbitrarily altered.  Perhaps that sentence could be reworded to make that clearer?  Like say ``` `packed` may also alter the alignment of each field as specified below.``` and just not mention "padding" if it is misleading?

joshlf:
> Even with that clarification, I think the problem still remains. Even though a struct/union and all of its fields are lowered to 1-alignment, that does not guarantee that it will not have padding. The layout could add arbitrary padding (for example, to reach some more desirable total size) unless it's specifically guaranteed that `repr(packed(1))` structs/unions do not have any padding. At least, that's my understanding of how alignment interacts with type layout.

This is my understanding as well - the reference, as currently written, does not guarantee that a `packed` struct will have the minimum possible padding. It allows that that _could_ happen, but it doesn't guarantee that it will. This PR is intended to change that so that it can be relied upon.

> I also don't see packed(1) as being much of a special case, or at least not different in the layout method.

It's a natural consequence of the guarantee I'd like to add, and doesn't technically need to be there, agreed. However, it's a very common use case, and so I figured it was useful to call it out specifically. I'd be happy to rephrase it as "an important consequence of these rules" or something like that.

ehuss:
> I'd be happy to rephrase it as "an important consequence of these rules" or something like that.

That might be better.
There also appears to be a CI error.

@rust-lang/lang Nominating this for discussion/approval.  This seems reasonable to me, though I don't know if there is something I haven't considered.

nikomatsakis:
These changes look good to me -- I like the new wording.

nikomatsakis:
We [discussed this in today's @rust-lang/lang meeting](https://github.com/rust-lang/lang-team/blob/master/minutes/2022-03-22.md#extern-macroabi-fn-lang-team151) and the general consensus was that this change made sense, but that it also made sense to do an FCP since this was a change to the lang spec, if only a small one. Therefore...

@rfcbot fcp merge

ehuss (approved):
Thanks!

PR #1165, opened 2022-02-10
📝 Fix grammar

vegerot:
At least in the US, `foos have, but bars haven't are` doesn't sound quite right.

I normally hear it as `foos have, but bars don't are` or `foos have, but bars don't have are`.


icecream17:
Maybe

```Notable features foos have that bars don't```

ehuss:
As part of #1166, this sentence got rewritten.  Hopefully the new one is a little clearer.  Thanks for the PR, though!

PR #1166, opened 2022-02-10
Update const_eval.md for feature stabilization

eholk:
We are in the process of stabilizing `const_fn_fn_ptr_basics` and `const_fn_trait_bound` (see https://github.com/rust-lang/rust/pull/93827). This edit updates const_eval.md to reflect the new capabilities after stabilizing these features.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -90,7 +90,6 @@ Notable features that are allowed in const contexts but not in const functions i
   * floating point values are treated just like generic parameters without trait bounds beyond
   `Copy`. So you cannot do anything with them but copy/move them around.
 * comparing raw pointers
-* union field access
 
 Conversely, the following are possible in a const function, but not in a const context:
 
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -89,7 +89,6 @@ Notable features that are allowed in const contexts but not in const functions i
 * floating point operations
   * floating point values are treated just like generic parameters without trait bounds beyond
   `Copy`. So you cannot do anything with them but copy/move them around.
-* comparing raw pointers
 
 Conversely, the following are possible in a const function, but not in a const context:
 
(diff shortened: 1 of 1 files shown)


eholk:
I removed the line about trait bounds. I didn't see much in the text that said one way or another about function pointers, so I didn't directly mention them yet.

ehuss:
Thanks!  Yea, it doesn't look like it mentioned the previous restrictions on function pointers.  

@oli-obk or @RalfJung, is there anything else you would like to see here?

src/const_eval.md
```diff
@@ -84,13 +84,11 @@ slightly) different results. It is advisable to not make array lengths and enum
...
-Notable features that const contexts have, but const fn haven't are:
+Notable features that are allowed in const contexts but not in const functions include:
 
 * floating point operations
   * floating point values are treated just like generic parameters without trait bounds beyond
   `Copy`. So you cannot do anything with them but copy/move them around.
-* `dyn Trait` types
-* generic bounds on generic parameters beyond `Sized`
 * comparing raw pointers
 * union field access
```
oli-obk, line 93: Drive-by fix: unions work just fine nowadays

oli-obk (approved):

src/const_eval.md
eholk: Sounds good, I went ahead and removed that too.

RalfJung:
LGTM.

But, since when do we allow comparing raw pointers in const contexts? [This](https://github.com/rust-lang/rust/blob/b7e95dee65c35db8f8e07046d445b12d92cbae12/compiler/rustc_const_eval/src/transform/check_consts/ops.rs#L670) sure looks like it is always rejected.

RalfJung:
Looks like that note goes [all the way back](https://github.com/rust-lang/reference/commit/167d75106949fcd83eb6925ee91a0b6e15fcc7a4), but I don't think we ever allowed ptr-to-int casts or raw ptr comparisons in const contexts? At least we never allowed actually executing them.

eholk:
I went ahead and remove "comparing raw pointers" too since those aren't allowed in const contexts either.

ehuss (approved):
Looks like this has been merged, so I'm going to go ahead and merge this.  If anyone has anything they'd like to follow up here, feel free to open a PR.

eholk:
Thanks!

PR #1168, opened 2022-02-14
Add supported Directives list

chorman0773:
This seeks to limit the directives guaranteed to be supported by inline assembly to a subset of those supported by LLVM and GNU AS. The list is chosen as a compromise between utility and implementability, and may be increased in the future.

See discussions here:
<https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/asm!.20and.20backends> and <https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/Defect.20Report.3A.20GNU.20AS.20support>

CC: @joshtriplett

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -491,6 +491,7 @@ The following directives are guaranteed to be supported by the assembler:
 .4byte
 .8byte
 .byte
+.short
 .word
 .long
 .quad
@@ -523,6 +524,9 @@ The following directives are guaranteed to be supported by the assembler:
 .comm
 .lcomm
 .option
+.equ
+.set
+.align
 ```
 
 The following directives are guaranteed to be supported for `global_asm` only:
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -527,6 +527,7 @@ The following directives are guaranteed to be supported by the assembler:
 .equ
 .set
 .align
+.inst
 ```
 
 The following directives are guaranteed to be supported for `global_asm` only:
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -593,7 +593,7 @@ On x86 targets, both 32-bit and 64-bit, the following additional directives are 
 - `.intel_syntax`
 - `.nops`
 
-Use of the `.att_syntax` and `.intel_syntax` with no parameters is supported, but the syntax must be restored to the option at entry (`.intel_syntax` without the `att_syntax` asm option, or `.att_syntax` with that option) or the behaviour is undefined (the output of the compiler may be corrupted as a result). Use of `.att_syntax` and `.intel_syntax` with an option (such as `.intel_syntax prefix` or `.att_syntax noprefix`) is unsupported and results in assembler-dependant behaviour. If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.
+Use of the `.att_syntax` and `.intel_syntax` directives with no parameters (or with parameters equivalent to the defaults) is supported, but the syntax must be restored to the option at entry (`.intel_syntax` without the `att_syntax` asm option, or `.att_syntax` with that option) or the behavior is undefined (the compiled output may be corrupted as a result). Use of `.att_syntax` and `.intel_syntax` with a non-default option (such as `.intel_syntax prefix` or `.att_syntax noprefix`) is unsupported. If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.
 
 On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
 
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -486,57 +486,57 @@ The result of using other directives is assembler-specific (and may cause an err
 
 The following directives are guaranteed to be supported by the assembler:
 
-```as
-.2byte
-.4byte
-.8byte
-.byte
-.short
-.word
-.long
-.quad
-.float
-.double
-.octa
-.sleb128
-.uleb128
-.ascii
-.asciz
-.string
-.skip
-.space
-.balign
-.balignl
-.balignw
-.balign
-.balignl
-.balignw
-.fill
-.section
-.pushsection
-.popsection
-.text
-.bss
-.data
-.def
-.endef
-.scl
-.comm
-.lcomm
-.option
-.equ
-.set
-.align
-.inst
-```
+
+- `.2byte`
+- `.4byte`
+- `.8byte`
+- `.byte`
+- `.short`
+- `.word`
+- `.long`
+- `.quad`
+- `.float`
+- `.double`
+- `.octa`
+- `.sleb128`
+- `.uleb128`
+- `.ascii`
+- `.asciz`
+- `.string`
+- `.skip`
+- `.space`
+- `.balign`
+- `.balignl`
+- `.balignw`
+- `.balign`
+- `.balignl`
+- `.balignw`
+- `.fill`
+- `.section`
+- `.pushsection`
+- `.popsection`
+- `.text`
+- `.bss`
+- `.data`
+- `.def`
+- `.endef`
+- `.scl`
+- `.comm`
+- `.lcomm`
+- `.option`
+- `.equ`
+- `.set`
+- `.align`
+- `.inst`
+
 
 The following directives are guaranteed to be supported for `global_asm` only:
-```as
-.alt_entry
-.private_extern
-.globl
-.global
-```
+
+- `.alt_entry`
+- `.private_extern`
+- `.globl`
+- `.global`
+
 
 #### Target Specific Directive Support
 
@@ -544,69 +544,66 @@ The following directives are guaranteed to be supported for `global_asm` only:
 
 The following directives are supported on ELF targets that support DWARF unwind info:
 
-```as
-.cfi_adjust_cfa_offset
-.cfi_def_cfa
-.cfi_def_cfa_offset
-.cfi_def_cfa_register
-.cfi_endproc
-.cfi_escape
-.cfi_lsda
-.cfi_offset
-.cfi_personality
-.cfi_register
-.cfi_rel_offset
-.cfi_remember_state
-.cfi_restore
-.cfi_restore_state
-.cfi_return_column
-.cfi_same_value
-.cfi_sections
-.cfi_signal_frame
-.cfi_startproc
-.cfi_undefined
-.cfi_window_save
-```
+
+- `.cfi_adjust_cfa_offset`
+- `.cfi_def_cfa`
+- `.cfi_def_cfa_offset`
+- `.cfi_def_cfa_register`
+- `.cfi_endproc`
+- `.cfi_escape`
+- `.cfi_lsda`
+- `.cfi_offset`
+- `.cfi_personality`
+- `.cfi_register`
+- `.cfi_rel_offset`
+- `.cfi_remember_state`
+- `.cfi_restore`
+- `.cfi_restore_state`
+- `.cfi_return_column`
+- `.cfi_same_value`
+- `.cfi_sections`
+- `.cfi_signal_frame`
+- `.cfi_startproc`
+- `.cfi_undefined`
+- `.cfi_window_save`
+
 
 ##### Structured Exception Handling
 
 On targets with structured exception Handling, the following additional directives are guaranteed to be supported:
-```as
-.seh_endproc
-.seh_endprologue
-.seh_proc
-.seh_pushreg
-.seh_savereg
-.seh_setframe
-.seh_stackalloc
-```
+
+- `.seh_endproc`
+- `.seh_endprologue`
+- `.seh_proc`
+- `.seh_pushreg`
+- `.seh_savereg`
+- `.seh_setframe`
+- `.seh_stackalloc`
+
 
 ##### x86
 
 On x86, the following additional directives are guaranteed to be supported:
-```as
-.nops
-```
+- `.nops`
 
 On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
-```as
-.code16
-.code32
-```
+
+- `.code16`
+- `.code32`
+
 
 ##### ARM
 
 On ARM for `global_asm!` only, the following additional directives are guaranteed to be supported:
-```as
-.code 
-.thumb
-.thumb_func
-```
+
+- `.code`
+- `.thumb`
+- `.thumb_func`
+
 
 On ARM, the following additional directives are guaranteed to be supported:
-```as
-.fnstart
-.fnend
-.save
-.movsp
-```
+
+- `.fnstart`
+- `.fnend`
+- `.save`
+- `.movsp`
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -87,7 +87,7 @@ On x86, the `.intel_syntax noprefix` mode of GAS is used by default.
 On ARM, the `.syntax unified` mode is used.
 These targets impose an additional restriction on the assembly code: any assembler state (e.g. the current section which can be changed with `.section`) must be restored to its original value at the end of the asm string.
 Assembly code that does not conform to the GAS syntax will result in assembler-specific behavior. 
-Further constraints on the directives used by the assembly are indicated by [Directives Support](#directives-support). 
+Further constraints on the directives used by inline assembly are indicated by [Directives Support](#directives-support). 
 
 [format-syntax]: ../std/fmt/index.html#syntax
 [rfc-2795]: https://github.com/rust-lang/rfcs/pull/2795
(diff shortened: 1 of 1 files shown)


Havvy:
Is there a reason to list the directives in a single code block and not a bulleted list of directives (each in a code block).

Also, is the behavior of these directives documented anywhere we could link to?

chorman0773:
Some of them was documented in various parts of the GNU AS documentation.

chorman0773:
https://sourceware.org/binutils/docs/as/

chorman0773:
I've turned it into a list of bullets

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
   - The compiler is currently unable to detect this due to the way inline assembly is compiled, but may catch and reject this in the future.
 
 > **Note**: As a general rule, the flags covered by `preserves_flags` are those which are *not* preserved when performing a function call.
+
+### Directives Support
+
+Inline ASM supports a subset of the directives supported by both GNU AS and LLVM's internal assembler, given as follows.
```
joshtriplett, line 484:
```suggestion
Inline assembly supports a subset of the directives supported by both GNU AS and LLVM's internal assembler, given as follows.
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+##### x86
+
+On x86, the following additional directives are guaranteed to be supported:
+- `.nops`
+
+On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
+
+- `.code16`
+- `.code32`
+
```
joshtriplett, line 596:
```suggestion

The directives `.att_syntax` and `.intel_syntax` (including options such as `noprefix`) are not fully supported. Using these directives to affect literal assembly syntax with no operands will work, as long as the state is restored before the end of the assembly block. Using these directives to affect assembly that includes operands is not defined.

```

src/inline-assembly.md
```diff
@@ -86,7 +86,8 @@ Currently, all supported targets follow the assembly code syntax used by LLVM's
 On x86, the `.intel_syntax noprefix` mode of GAS is used by default.
 On ARM, the `.syntax unified` mode is used.
 These targets impose an additional restriction on the assembly code: any assembler state (e.g. the current section which can be changed with `.section`) must be restored to its original value at the end of the asm string.
-Assembly code that does not conform to the GAS syntax will result in assembler-specific behavior.
+Assembly code that does not conform to the GAS syntax will result in assembler-specific behavior. 
+Further constraints on the directives used by the assembly are indicated by [Directives Support](#directives-support).
```
joshtriplett, line 90:
```suggestion
Further constraints on the directives used by inline assembly are indicated by [Directives Support](#directives-support). 
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
   - The compiler is currently unable to detect this due to the way inline assembly is compiled, but may catch and reject this in the future.
 
 > **Note**: As a general rule, the flags covered by `preserves_flags` are those which are *not* preserved when performing a function call.
+
+### Directives Support
+
+Inline ASM supports a subset of the directives supported by both GNU AS and LLVM's internal assembler, given as follows.
+The result of using other directives is assembler-specific (and may cause an error, or may be accepted as-is).
+
```
joshtriplett, line 486:
```suggestion

If inline assembly includes any "stateful" directive that modifies how subsequent assembly is processed, the block must undo the effects of any such directives before the inline assembly ends.

```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.lcomm`
+- `.option`
+- `.equ`
+- `.set`
+- `.align`
+- `.inst`
+- `.p2align`
+
+
+The following directives are guaranteed to be supported for `global_asm` only:
```
joshtriplett, line 534: I would prefer to unify these lists, and not suggest that some directives will work within `global_asm` but not within `asm`.

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.seh_setframe`
+- `.seh_stackalloc`
+
+
+##### x86
+
+On x86, the following additional directives are guaranteed to be supported:
+- `.nops`
+
+On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
```
joshtriplett, line 592: This should not be unspecified; they should be valid within `asm` as well.

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.seh_endproc`
+- `.seh_endprologue`
+- `.seh_proc`
+- `.seh_pushreg`
+- `.seh_savereg`
+- `.seh_setframe`
+- `.seh_stackalloc`
+
+
+##### x86
```
joshtriplett, line 587:
```suggestion
##### x86 (32-bit and 64-bit)
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+On x86, the following additional directives are guaranteed to be supported:
```
joshtriplett, line 589:
```suggestion
On x86 targets, both 32-bit and 64-bit, the following additional directives are guaranteed to be supported:
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+### Directives Support
+
+Inline ASM supports a subset of the directives supported by both GNU AS and LLVM's internal assembler, given as follows.
+The result of using other directives is assembler-specific (and may cause an error, or may be accepted as-is).
+
+The following directives are guaranteed to be supported by the assembler:
+
+
+- `.2byte`
```
joshtriplett, line 490: Could you please either sort the list, or group the list into logical groups?

It may make some sense to group things like `byte`, `short`, `word`, `quad`, and similar together, but that seems like substantially more organizational work, and in practice I would expect that a fully sorted list may be easier to use as a reference.

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
+
+- `.code16`
+- `.code32`
+
+
+##### ARM
+
+On ARM for `global_asm!` only, the following additional directives are guaranteed to be supported:
```
joshtriplett, line 600: As above, I'd prefer to see these lists unified.

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+On x86, the following additional directives are guaranteed to be supported:
+- `.nops`
+
+On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
+
+- `.code16`
+- `.code32`
+
+
+##### ARM
```
joshtriplett, line 598: This should specify whether it applies to ARM, or to both ARM and ARM64. (I'm guessing some of these directives aren't valid on ARM64.)

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.scl`
+- `.comm`
+- `.lcomm`
+- `.option`
+- `.equ`
+- `.set`
+- `.align`
+- `.inst`
+- `.p2align`
+
```
joshtriplett, line 532:
```suggestion
- `.bundle_align_mode`
- `.bundle_lock`
- `.bundle_unlock`
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.cfi_restore`
+- `.cfi_restore_state`
+- `.cfi_return_column`
+- `.cfi_same_value`
+- `.cfi_sections`
+- `.cfi_signal_frame`
+- `.cfi_startproc`
+- `.cfi_undefined`
+- `.cfi_window_save`
+
```
joshtriplett, line 572: There are several CFI directives not listed here, such as:
- `.cfi_personality_id`
- `.cfi_fde_data`
- `.cfi_inline_ldsa`
(and more, those are just the first three I noticed when reading https://sourceware.org/binutils/docs/as/CFI-directives.html )

Are those verifiably *not* supported by LLVM, or are they undocumented?

src/inline-assembly.md
joshtriplett: Also, directives like `.dc`, `.dcb`, and `.ds`. Those simply emit data; if they're supported by LLVM we should include them.

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.text`
+- `.bss`
+- `.data`
+- `.def`
+- `.endef`
+- `.scl`
+- `.comm`
+- `.lcomm`
+- `.option`
+- `.equ`
```
joshtriplett, line 527:
```suggestion
- `.equ`
- `.equiv`
- `.eqv`
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.balign`
+- `.balignl`
+- `.balignw`
+- `.balign`
+- `.balignl`
+- `.balignw`
+- `.fill`
+- `.section`
+- `.pushsection`
+- `.popsection`
```
joshtriplett, line 517:
```suggestion
- `.popsection`
- `.subsection`
```

src/inline-assembly.md
```diff
@@ -477,3 +478,135 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+- `.thumb`
+- `.thumb_func`
+
+
+On ARM, the following additional directives are guaranteed to be supported:
+
+- `.fnstart`
+- `.fnend`
+- `.save`
+- `.movsp`
```
joshtriplett, line 612:
```suggestion
- `.movsp`
- `.even`
```

joshtriplett:
OK, I've posted comments for every addition or modification I think this needs.

joshtriplett:
BTW, I reviewed all the inline assembly used in the Linux kernel, and my comments here also include all the directives found there (with the exception of macro directives, `.org` which seems too stateful for in inline asm, `.arch` which would have various issues and is hopefully unnecessary, and a few target-specific ones for targets we don't have).

chorman0773 (commented):
Commented on some suggestions

src/inline-assembly.md
joshtriplett: You already have to support both syntaxes, given the att_syntax option.

I agree that the state restoration is covered above, but I felt that it couldn't hurt to remind people that this in particular is something stateful.

src/inline-assembly.md
joshtriplett: cc @Amanieu Any idea here?

src/inline-assembly.md
joshtriplett: Not critical, we can always add these later.

src/inline-assembly.md
Amanieu: Yea these are specific to ARM32. Also there are [a lot more directives](https://sourceware.org/binutils/docs/as/ARM-Directives.html), the ones listed here are just ones I happen to be using.

src/inline-assembly.md
```diff
@@ -477,3 +478,143 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+
+##### x86
+
+On x86, the following additional directives are guaranteed to be supported:
+- `.att_syntax`
+- `.intel_syntax`
+- `.nops`
+
+Use of the `.att_syntax` and `.intel_syntax` with no parameters is supported, but the syntax must be restored to the option at entry (`.intel_syntax` without the `att_syntax` asm option, or `.att_syntax` with that option) or the behaviour is undefined (the output of the compiler may be corrupted as a result). Use of `.att_syntax` and `.intel_syntax` with an option (such as `.intel_syntax prefix` or `.att_syntax noprefix`) is unsupported and results in assembler-dependant behaviour. If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.
```
chorman0773, line 597: > If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.

This could probably easily be unspecified, but depending on how interpolations are implemented, the results could vary wildly.

src/inline-assembly.md
```diff
@@ -477,3 +478,143 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+On x86 for `global_asm!` only, the following additional directives are guaranteed to be supported (it is unspecified whether `.code16` or `.code32` are supported for `asm!()`):
+
+- `.code16`
+- `.code32`
+
+
+##### ARM
```
joshtriplett, line 604:
```suggestion
##### ARM (32-bit)
```

src/inline-assembly.md
```diff
@@ -477,3 +478,143 @@ To avoid undefined behavior, these rules must be followed when using function-sc
...
+
+
+##### x86 (32-bit and 64-bit)
+
+On x86 targets, both 32-bit and 64-bit, the following additional directives are guaranteed to be supported:
+- `.att_syntax`
+- `.intel_syntax`
+- `.nops`
+
+Use of the `.att_syntax` and `.intel_syntax` with no parameters is supported, but the syntax must be restored to the option at entry (`.intel_syntax` without the `att_syntax` asm option, or `.att_syntax` with that option) or the behaviour is undefined (the output of the compiler may be corrupted as a result). Use of `.att_syntax` and `.intel_syntax` with an option (such as `.intel_syntax prefix` or `.att_syntax noprefix`) is unsupported and results in assembler-dependant behaviour. If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.
```
joshtriplett, line 596: Minor tweaks, notably allowing parameters equivalent to the default and only disallowing non-default parameters.

```suggestion
Use of the `.att_syntax` and `.intel_syntax` directives with no parameters (or with parameters equivalent to the defaults) is supported, but the syntax must be restored to the option at entry (`.intel_syntax` without the `att_syntax` asm option, or `.att_syntax` with that option) or the behavior is undefined (the compiled output may be corrupted as a result). Use of `.att_syntax` and `.intel_syntax` with a non-default option (such as `.intel_syntax prefix` or `.att_syntax noprefix`) is unsupported. If operand interpolations are used between setting the syntax mode with one of these directives, and restoring it to the block's default, the behaviour is undefined.
```

src/inline-assembly.md
joshtriplett: Submitted a suggestion (below) to that effect.

src/inline-assembly.md
chorman0773: It may be nice to say an implementation just needs to support `.intel_syntax`/`.att_syntax` and nothing after that. This isn't a huge deal, though, it's just a note.

joshtriplett: In practice, handling a single fixed optional parameter that can only take a single value or be omitted entirely, and which has no semantic effect, doesn't seem like it'd be difficult.

I can easily imagine a compiler detecting them and complaining, but if it has the capability to do so, it should detect the whole directive instead and lint towards `options(...)`. I don't think the argument in particular makes that any harder.

joshtriplett:
Context for @rust-lang/lang:

- Inline assembly (`asm!` and `global_asm!`) is stabilizing soon. :tada:
- In principle, and as defined in the inline assembly RFC, we want to limit the supported directives to the subset supported by both LLVM and GNU AS, not least of which so that people can use either as an external assembler.
- There are also a few complex directives that we don't want to force the integrated assemblers of future backends to have to support, most notably the macro directives.
- As stated in the original RFC, we aren't proposing to actually filter directives.
- However, we should at least *document* the set of supported directives, to discourage people from using some of the more obscure ones.
- We may, in the future, consider some best-effort linting, such that people *could* work around those lints but they'd be more likely to know they're doing something unsupported.

This reference PR adds documentation of the set of assembler directives we intend to support. We can always expand this list in the future.

@rfcbot merge

joshtriplett:
@rfcbot concern unify-asm-and-global-asm-supported-directives

joshtriplett:
@rfcbot resolved unify-asm-and-global-asm-supported-directives

nikomatsakis:
@rfcbot reviewed

joshtriplett:
Since there are a few seconds and no objections, and this is easily reversed or edited, I'm going to go ahead and merge this:

@bors r+

joshtriplett (approved):

PR #1169, opened 2022-02-15
Document compatibility between declarative and procedural macro tokens

petrochenkov:
This PR assumes https://github.com/rust-lang/rust/pull/92472 is merged.

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -271,12 +271,73 @@ fn invoke4() {}
 // out: item: "fn invoke4() {}"
 ```
 
+### Declarative macro tokens and procedural macro tokens
+
+Declarative `macro_rules` macros and procedural macros use similar, but
+different definitions for tokens (or rather [`TokenTree`s].)
+
+Token trees in `macro_rules` (corresponding to `tt` matchers) are defined as
+- Delimited groups (`(...)`, `{...}`, etc)
+- All operators supported by the language, both single-character and
+  multi-character ones (`+`, `+=`).
+    - Note that this set doesn't include the single quote `'`.
+- Literals (`"string"`, `1`, etc)
+    - Note that negation (e.g. `-1`) is never a part of such literal tokens,
+      but a separate operator token.
+- Identifiers, including keywords (`ident`, `r#ident`, `fn`)
+- Lifetimes (`'ident`)
+- Metavariable substitutions in `macro_rules` (e.g. `$my_expr` in
+  `macro_rules! mac { ($my_expr: expr) => { $my_expr } }` after the `mac`'s
+  expansion, which will be considered a single token tree regardless of the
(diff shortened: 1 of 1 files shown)


src/procedural-macros.md
```diff
@@ -271,12 +271,72 @@ fn invoke4() {}
...
+
+When passed to a proc-macro
+- All multi-character operators are broken into single characters.
+- Lifetimes are broken into a `'` character and an identifier.
+- All metavariable substitutions are represented as their underlying token
+  streams.
+    - Such token streams may be wrapped into delimited groups ([`Group`]) with
+      implicit delimiters ([`Delimiter::None`]) when it's necessary for
+      preserving parsing priorities.
+    - `tt` and `ident` substitutions are never wrapped into such groups and
```
petrochenkov, line 316: The `ident` part here requires https://github.com/rust-lang/rust/pull/92472.

src/procedural-macros.md
```diff
@@ -271,12 +271,72 @@ fn invoke4() {}
...
+      always represented as their underlying token trees.
+
+When passed from a proc macro
```
dtolnay, line 319: I suggest "When emitted from a proc macro" or "When returned by a proc macro". It's passing in a sense, but this isn't the clearest way to say that.

dtolnay (changes requested):

petrochenkov:
I've addressed review comments.
https://github.com/rust-lang/rust/pull/92472 has also landed.

petrochenkov:
@rustbot ready
(Is the bot working on this repo?)

ehuss (commented):
Thanks!

src/procedural-macros.md
```diff
@@ -271,12 +271,72 @@ fn invoke4() {}
...
+- Lifetimes (`'ident`)
+- Metavariable substitutions in `macro_rules` (e.g. `$my_expr` in
+  `macro_rules! mac { ($my_expr: expr) => { $my_expr } }` after the `mac`'s
+  expansion, which will be considered a single token tree regardless of the
+  passed expression)
+
+Token trees in procedural macros are defined as
+- Delimited groups (`(...)`, `{...}`, etc)
+- All punctuation characters used in operators supported by the language (`+`,
+  but not `+=`), and also the single quote `'` character
```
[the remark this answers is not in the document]
petrochenkov: I languages defined by proc macros `'` is an independent token, it doesn't necessarily have to be joined with identifiers to form lifetimes.

petrochenkov:
I've added the changes from https://github.com/rust-lang/reference/pull/1169#discussion_r839962540 and https://github.com/rust-lang/reference/pull/1169#discussion_r839968049 in a slightly reworded form.

ehuss (approved):
Thanks!

PR #1170, opened 2022-02-15
Document native library modifier syntax and the `whole-archive` modifier specifically

petrochenkov:
This is an accompanying PR for linking modifier stabilization in https://github.com/rust-lang/rust/pull/93901.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -125,6 +125,15 @@ specifies the kind of library with the following possible values:
 
 The `name` key must be included if `kind` is specified.
 
+The optional `modifiers` argument is a way to specify linking modifiers for the
+library to link.
+Modifiers are specified as a comma-delimited string with each modifier prefixed
+with either a `+` or `-` to indicate that the modifier is enabled or disabled,
+respectively.
+Specifying multiple `modifiers` arguments in a single `link` attribute,
+or multiple identical modifiers in the same `modifiers` argument is not currently supported. \
+Example: `#[link(name = "mylib", kind = "static", modifiers = "+whole-archive")`.
+
 The `wasm_import_module` key may be used to specify the [WebAssembly module]
 name for the items within an `extern` block when importing symbols from the
 host environment. The default module name is `env` if `wasm_import_module` is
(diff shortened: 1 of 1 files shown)


src/items/external-blocks.md
```diff
@@ -153,6 +160,22 @@ this to satisfy the linking requirements of extern blocks elsewhere in your
...
+`+whole-archive` means that the static library is linked as a whole archive
+without throwing any object files away.
+
+This modifier translates to `--whole-archive` for `ld`-like linkers,
+to `/WHOLEARCHIVE` for `link.exe`, and to `-force_load` for `ld64`.
+The modifier does nothing for linkers that don't support it.
+
+The default for this modifier is `-whole-archive`. \
+NOTE: The default may currently be different when building dylibs for some targets,
+but it is not guaranteed.
```
petrochenkov, line 177: This paragraph is a copypaste from the compiler command line doc in https://github.com/rust-lang/rust/pull/93901.

Not sure where this text should live, seems more like a compiler stuff.
In that case we need to link from the reference to the compiler command line doc page.

petrochenkov:
I have updated the PR and https://github.com/rust-lang/rust/pull/93901 has also landed.

@rustbot label -S-waiting-on-stabilization +S-waiting-on-review

src/items/external-blocks.md
```diff
@@ -186,3 +205,4 @@ restrictions as [regular function parameters].
 [_Visibility_]: ../visibility-and-privacy.md
 [attributes]: ../attributes.md
 [regular function parameters]: functions.md#attributes-on-function-parameters
+[documentation for rustc]: https://doc.rust-lang.org/rustc/command-line-arguments.html#-l-link-the-generated-crate-to-a-native-library
```
ehuss, line 208: We generally prefer relative links to other books that are published by rust-lang. This helps with offline viewing, and keeping the user within the same release when viewing release-specific URLs like https://doc.rust-lang.org/1.59.0/reference/. Our linkchecker also only supports relative links, and this helps prevent links from going stale. This would also avoid confusion, as the existing link doesn't have any information on whole-archive which can be a bit confusing for someone looking at the nightly docs.

Also, I feel like it would be a little clearer if it linked directly to the whole-archive section, unless you had some specific reason to link to the `-l` section?

```suggestion
[documentation for rustc]: ../../rustc/command-line-arguments.html#linking-modifiers-whole-archive
```

petrochenkov: Looks like linkchecker doesn't like it though.

petrochenkov: Ah, ok, external links should indeed use html, not md.

ehuss (approved):
Thanks!

PR #1172, opened 2022-02-16
Document ADX `target_feature`

tarcieri:
Related stabilization PR: https://github.com/rust-lang/rust/pull/93745


ehuss (approved):
Looks like this is accepted, so I'm going to go ahead and merge.  I made a small change to keep the list alphabetized. Thanks!

PR #1174, opened 2022-03-02
inline-asm.md. Fix `.intel_syntax` directive

chorman0773:
See: https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm

The current supported directives list has `.intel_syntax` and `.att_syntax` without options supported, and claims them to be the default option (for no `att_syntax` option vs. `att_syntax` option). However, the default behaviour of the `.intel_syntax` directive differ between llvm-mc and gnu as, and GNU AS differs from the guaranteed behaviour: See https://godbolt.org/z/3js4Pfds7 (GNU AS) vs. https://godbolt.org/z/rKhqPxKvr (llvm-mc).

This PR removes the guaranteed support for the no-parameter form of`.intel_syntax`, and from `.att_syntax` as well to mirror.

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -587,19 +587,15 @@ On targets with structured exception Handling, the following additional directiv
 ##### x86 (32-bit and 64-bit)
 
 On x86 targets, both 32-bit and 64-bit, the following additional directives are guaranteed to be supported:
-- `.att_syntax`
-- `.intel_syntax`
 - `.nops`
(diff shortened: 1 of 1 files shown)


joshtriplett:
Per discussion on Zulip: perhaps we should just drop these entirely from the list of supported items. If I'd known they had varying behavior I wouldn't have suggested adding them.

joshtriplett (approved):

Issue #939, opened 2021-01-21
Split out literal expression and literal tokens


Havvy:
Right now the literal tokens chapter describes the behavior of literal expressions and not just the tokens. The behavior of matching the token to a specific primitive value should be described in the literal expressions chapter while the token literals chapter just describes the tokens.

ehuss:
Can you clarify exactly which parts you are referring to?  One oddity I see is the discussion about type inference with unsuffixed numbers, but I'm curious what else you're thinking of.

Havvy:
Ah, that's what I get for not actually checking the tokens page (it's a long page...) while quickly looking at how sparse https://doc.rust-lang.org/reference/expressions/literal-expr.html is. This is actually something just not documented, though yeah, that section on type inference should be moved.

Issue #298, opened 2018-04-03
Document RFC 0879 (small-base-lexing) and RFC 0463 (future-proof-literal-suffixes)


alercah:
[RFC 0879](https://github.com/rust-lang/rfcs/tree/master/text/0879-small-base-lexing.md)
[RFC 0463](https://github.com/rust-lang/rfcs/blob/master/text/0463-future-proof-literal-suffixes.md

kubycsolutions:
Just checking my understanding: You're saying that the lexer sees `0b110127` as a malformed token rather than a BIN_LITERAL immediately followed by a DEC_LITERAL as the documentation suggests it should?

If so, are you sure you want to change the doc rather than the lexer?

mattheww:
The current documentation is correct for numeric literals (and what it says is that 0b110127 is a [reserved form][1]).

This issue is still open because literal suffixes on string literals aren't completely documented.

[1]: https://doc.rust-lang.org/reference/tokens.html#reserved-forms-similar-to-number-literals

Issue #1167, opened 2022-02-13
Tokens `true` and `false` are defined twice.


mattheww:
At present, `true` and `false` are each defined twice in the `Lexer` blocks: once as keywords and once as part of BOOLEAN_LITERAL.

I think we should do one of:

 - change the BOOLEAN_LITERAL definition to use KW_TRUE | KW_FALSE rather than `true` | `false`

 - remove BOOLEAN_LITERAL and update literal-expr.md and patterns.md accordingly

The second is a larger change but seems more in the spirit of #939. It would also make the text

> Any kind of literal (string, integer, etc) with any suffix is valid as a token, and can be passed to a macro without producing an error

in tokens.md no longer appear to make a claim about suffixes following `true` and `false`.

Havvy:
I was asked if I had an opinion on this. I do not; the nitty gritty grammar decisions aren't something I've spent much brainpower thinking about.

PR #1177, opened 2022-03-06
Numeric literal expressions and literal suffixes

mattheww:
This documents integer and floating-point literal expressions, and improves the docs for integer and floating-point literal tokens.

Covers some of #939, for integer and floating-point literals.

Covers the RFC 0879 part of #298, and some of the RFC 0463 part (ie, for numeric literals but not string literals).

The details of floating-point conversion deserve a more throrough description, but maybe that needs looking at together with the standard library docs.

I had a false-positive error from linkchecker; the last commit is a workaround for that, but I think the right thing would be an addition to `LINKCHECK_EXCEPTIONS` in `linkcheck/main.rs`.

### DIFF
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -59,6 +59,18 @@ let a: u64 = 123;                  // type u64
 0usize;                            // type usize
 ```
 
+The value of the expression is determined from the string representation of the token as follows:
+
+* An integer radix is chosen by inspecting the first two characters of the string: `0b` indicates radix 2, `0o` indicates radix 8, `0x` indicates radix 16; otherwise the radix is 10.
+
+* If the radix is not 10, the first two characters are removed from the string.
+
+* Any underscores are removed from the string.
+
+* The string is converted to a `u128` value as if by [`u128::from_str_radix`] with the chosen radix.
+
+* The `u128` value is converted to the expression's type via a [numeric cast].
+
 ## Floating-point literal expressions
 
 A floating-point literal expression consists of a single [FLOAT_LITERAL] token.
@@ -87,8 +99,10 @@ let x: f64 = 2.; // type f64
 [constant expression]: ../const_eval.md#constant-expressions
 [floating-point types]: ../types/numeric.md#floating-point-types
 [literal tokens]: ../tokens.md#literals
+[numeric cast]: operator-expr.md#numeric-cast
 [numeric types]: ../types/numeric.md
 [suffix]: ../tokens.md#suffixes
+[`u128::from_str_radix`]: ../../core/primitive.u128.md#method.from_str_radix
 [CHAR_LITERAL]: ../tokens.md#character-literals
 [STRING_LITERAL]: ../tokens.md#string-literals
 [RAW_STRING_LITERAL]: ../tokens.md#raw-string-literals
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -108,6 +108,10 @@ The value of the expression is determined from the string representation of the 
 
 * The string is converted to the expression's type as if by [`f32::from_str`] or [`f64::from_str`].
 
+> **Note**: `inf` and `NaN` are not literal tokens.
+> The [`f32::INFINITY`], [`f64::INFINITY`], [`f32::NAN`], and [`f64::NAN`] constants can be used instead of literal expressions.
+> A literal large enough to be evaluated as infinite will trigger the `overflowing_literals` lint check.
+
 [constant expression]: ../const_eval.md#constant-expressions
 [floating-point types]: ../types/numeric.md#floating-point-types
 [lint check]: ../attributes/diagnostics.md#lint-check-attributes
@@ -116,7 +120,11 @@ The value of the expression is determined from the string representation of the 
 [numeric types]: ../types/numeric.md
 [suffix]: ../tokens.md#suffixes
 [`f32::from_str`]: ../../core/primitive.f32.md#method.from_str
+[`f32::INFINITY`]: ../../core/primitive.f32.md#associatedconstant.INFINITY
+[`f32::NAN`]: ../../core/primitive.f32.md#associatedconstant.NAN
 [`f64::from_str`]: ../../core/primitive.f64.md#method.from_str
+[`f64::INFINITY`]: ../../core/primitive.f64.md#associatedconstant.INFINITY
+[`f64::NAN`]: ../../core/primitive.f64.md#associatedconstant.NAN
 [`u128::from_str_radix`]: ../../core/primitive.u128.md#method.from_str_radix
 [CHAR_LITERAL]: ../tokens.md#character-literals
 [STRING_LITERAL]: ../tokens.md#string-literals
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -74,6 +74,9 @@ If the value does not fit in `u128`, the expression is rejected by the parser.
 
 * The `u128` value is converted to the expression's type via a [numeric cast].
 
+> **Note**: The final cast will truncate the value of the literal if it does not fit in the expression's type.
+> There is a [lint check] named `overflowing_literals`, defaulting to `deny`, which rejects expressions where this occurs.
+
 ## Floating-point literal expressions
 
 A floating-point literal expression consists of a single [FLOAT_LITERAL] token.
@@ -101,6 +104,7 @@ let x: f64 = 2.; // type f64
 
 [constant expression]: ../const_eval.md#constant-expressions
 [floating-point types]: ../types/numeric.md#floating-point-types
+[lint check]: ../attributes/diagnostics.md#lint-check-attributes
 [literal tokens]: ../tokens.md#literals
 [numeric cast]: operator-expr.md#numeric-cast
 [numeric types]: ../types/numeric.md
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -75,7 +75,7 @@ If the value does not fit in `u128`, the expression is rejected by the parser.
 * The `u128` value is converted to the expression's type via a [numeric cast].
 
 > **Note**: The final cast will truncate the value of the literal if it does not fit in the expression's type.
-> There is a [lint check] named `overflowing_literals`, defaulting to `deny`, which rejects expressions where this occurs.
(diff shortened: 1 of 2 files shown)


src/tokens.md
```diff
@@ -464,7 +441,7 @@ let horse = example.0b10;  // ERROR no field named `0b10`
 > **<sup>Lexer</sup>**\
 > FLOAT_LITERAL :\
 > &nbsp;&nbsp; &nbsp;&nbsp; DEC_LITERAL `.`
->   _(not immediately followed by `.`, `_` or an [identifier]_)\
+>   _(not immediately followed by `.`, `_` or an [identifier] or [keyword][keywords]_)\
```
ehuss, line 444: To be pedantic here, I think it is an identifier, raw identifier, or keyword.  We don't have have anything to succinctly refer to that. What do you think about adding a new rule to the identifiers chapter?  Maybe something like:

> IDENTIFIER_OR_KEYWORD_OR_RAW: IDENTIFIER_OR_KEYWORD | RAW_IDENTIFIER

I...can't think of a good name for that, maybe you can improve that.  That rule can then be referenced here, and the macros-by-example chapter (where it refers to `ident` in the Metavariables section).

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
 
-The representation semantics of floating-point numbers are described in
-["Machine Types"][machine types].
+Note that `-1.0`, for example, is analyzed as two tokens: `-` followed by `1.0`.
+
+#### Number pseudoliterals
+
+> **<sup>Lexer</sup>**\
+> NUMBER_PSEUDOLITERAL :\
+> &nbsp;&nbsp; &nbsp;&nbsp; DEC_LITERAL ( . DEC_LITERAL)? FLOAT_EXPONENT NUMBER_PSEUDOLITERAL_SUFFIX\
```
ehuss, line 490:
```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; DEC_LITERAL ( . DEC_LITERAL)<sup>?</sup> FLOAT_EXPONENT NUMBER_PSEUDOLITERAL_SUFFIX\
```

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
+> &nbsp;&nbsp; | DEC_LITERAL ( . DEC_LITERAL)? NUMBER_PSEUDOLITERAL_SUFFIX_NO_E\
```
ehuss, line 491:
```suggestion
> &nbsp;&nbsp; | DEC_LITERAL ( . DEC_LITERAL)<sup>?</sup> NUMBER_PSEUDOLITERAL_SUFFIX_NO_E\
```

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
+> &nbsp;&nbsp; | ( BIN_LITERAL | OCT_LITERAL | HEX_LITERAL ) NUMBER_PSEUDOLITERAL_SUFFIX_NO_E
+>
+> NUMBER_PSEUDOLITERAL_SUFFIX :\
+> &nbsp;&nbsp; IDENTIFIER_OR_KEYWORD <sub>_not matching INTEGER_SUFFIX or FLOAT_SUFFIX_</sub>
+>
+> NUMBER_PSEUDOLITERAL_SUFFIX_NO_E :\
+> &nbsp;&nbsp; NUMBER_PSEUDOLITERAL_SUFFIX <sub>_not beginning with `e` or `E`_</sub>
```
ehuss, line 498: This "beginning with e" is a bit surprising to me.  So strange!

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
+#### Number pseudoliterals
+
+> **<sup>Lexer</sup>**\
+> NUMBER_PSEUDOLITERAL :\
+> &nbsp;&nbsp; &nbsp;&nbsp; DEC_LITERAL ( . DEC_LITERAL)? FLOAT_EXPONENT NUMBER_PSEUDOLITERAL_SUFFIX\
+> &nbsp;&nbsp; | DEC_LITERAL ( . DEC_LITERAL)? NUMBER_PSEUDOLITERAL_SUFFIX_NO_E\
+> &nbsp;&nbsp; | ( BIN_LITERAL | OCT_LITERAL | HEX_LITERAL ) NUMBER_PSEUDOLITERAL_SUFFIX_NO_E
+>
+> NUMBER_PSEUDOLITERAL_SUFFIX :\
+> &nbsp;&nbsp; IDENTIFIER_OR_KEYWORD <sub>_not matching INTEGER_SUFFIX or FLOAT_SUFFIX_</sub>
```
ehuss, line 495: Unless I'm misunderstanding this, it seems to not cover something like `1.3E10u64`, which seems to be a single token.

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
+>
+> NUMBER_PSEUDOLITERAL_SUFFIX_NO_E :\
+> &nbsp;&nbsp; NUMBER_PSEUDOLITERAL_SUFFIX <sub>_not beginning with `e` or `E`_</sub>
+
+As described [above](#suffixes), tokens with the same form as numeric literals other than in the content of their suffix are accepted by the lexer (with the exception of some reserved forms described below).
```
ehuss, line 500: Perhaps it would be good to make it clear that these aren't used for anything other than tokenizing?  I might perhaps write something like this:

```suggestion
Tokenization of numeric literals allows arbitrary suffixes as described in the grammar above.
These values generate valid tokens, but are not a valid [literal expression], so are usually an error except as macro arguments.
```

src/tokens.md
```diff
@@ -477,47 +454,105 @@ let horse = example.0b10;  // ERROR no field named `0b10`
...
+> &nbsp;&nbsp; | OCT_LITERAL \[`8`-`9`&ZeroWidthSpace;]\
+> &nbsp;&nbsp; | ( BIN_LITERAL | OCT_LITERAL | HEX_LITERAL ) `.` \
+> &nbsp;&nbsp; &nbsp;&nbsp; &nbsp;&nbsp; _(not immediately followed by `.`, `_` or an [identifier] or [keyword][keywords]_)\
+> &nbsp;&nbsp; | ( BIN_LITERAL | OCT_LITERAL ) `e`\
+> &nbsp;&nbsp; | `0b` `_`<sup>\*</sup> _end of input or not BIN_DIGIT_\
+> &nbsp;&nbsp; | `0o` `_`<sup>\*</sup> _end of input or not OCT_DIGIT_\
+> &nbsp;&nbsp; | `0x` `_`<sup>\*</sup> _end of input or not HEX_DIGIT_\
+> &nbsp;&nbsp; | DEC_LITERAL ( . DEC_LITERAL)? (`e`|`E`) (`+`|`-`)? _end of input or not DEC_DIGIT_
+
+The following lexical forms similar to number literals are _reserved forms_:
```
ehuss, line 528: This introduction a little unclear to me.  Perhaps it could explain what these are up-front?  Maybe something like:

```suggestion
The following lexical forms similar to number literals are _reserved forms_.
Due to the possible ambiguity these raise, they are rejected by the tokenizer instead of being interpreted as separate tokens.
```

ehuss (commented):
Thanks for putting this together, it looks like a good improvement and step in the right direction.  I had a few questions and a few nit-picks.

src/tokens.md
mattheww: You're right. It also wasn't covering non-decimal with a float-type suffix, eg `0b1111_f32`.

I think the best way to write it will be to add back the extra suffixes rather than inventing more kinds of `NUMBER_PSEUDOLITERAL_SUFFIX_NO_XXX`, something like this:

````
NUMBER_PSEUDOLITERAL :
     DEC_LITERAL ( . DEC_LITERAL)? FLOAT_EXPONENT ( NUMBER_PSEUDOLITERAL_SUFFIX | INTEGER_SUFFIX )
   | DEC_LITERAL . DEC_LITERAL ( NUMBER_PSEUDOLITERAL_SUFFIX_NO_E | INTEGER SUFFIX )
   | DEC_LITERAL NUMBER_PSEUDOLITERAL_SUFFIX_NO_E
   | ( BIN_LITERAL | OCT_LITERAL | HEX_LITERAL ) ( NUMBER_PSEUDOLITERAL_SUFFIX_NO_E | FLOAT_SUFFIX )
````

**Comment by ehuss** [comment]
I think adding a single commit with the updates would be fine.

File: src/tokens.md (hunk 1 above)
Comment (mattheww) [reply src/tokens.md:444]: Ah, it could be a byte literal or a raw string literal too.

I think it will be best to switch to just talking about the following character directly, rather than how the following input would be tokenised (so matching the implementation more closely).

File: src/expressions/literal-expr.md [hunk 8]
```diff
@@ -8,19 +8,129 @@
...
+
+0b1111_1111_1001_0000;             // type i32
+0b1111_1111_1001_0000i64;          // type i64
+
+0usize;                            // type usize
+```
+
+The value of the expression is determined from the string representation of the token as follows:
+
+* An integer radix is chosen by inspecting the first two characters of the string: `0b` indicates radix 2, `0o` indicates radix 8, `0x` indicates radix 16; otherwise the radix is 10.
```
Havvy, line 66: I think the information after the colon should either be a sublist or a table. ~or a match expression~

Havvy (changes requested):
First off, I want to personally thank you for working on this page; this is a good chunk of content and it reads nicely. I'm not going to be nitpicky (I don't have the energy to do so and @ehuss is already doing so), but I do notice one glaring thing:

We no longer have prose about boolean, character, and string literals. We should have something, even if it's just section for them and the statements `An x literal statement consists of a single [X_LITERAL] token.` & `This section is incomplete.`.

src/expressions/literal-expr.md
mattheww: I've made it a sublist.

mattheww:
@Havvy I've added placeholder sections for the other types of literal expression. Some of them will be very easy to fill in, but I suppose that doesn't belong in this PR.

(Do you have an opinion on the choice in #1167? I'm leaning towards the second option.)

ehuss (approved):
Looks good, thanks!

ehuss:
@Havvy Are you OK with the changes?

Havvy (approved):

Havvy:
Yes. Quite pleased actually.

PR #1179, opened 2022-03-16
Add `let_chains` references

c410-f3r:
cc @nikomatsakis

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -2,29 +2,30 @@
 
 ## Syntax
 
-The same syntax is used by `if`, `if let` and chains of expressions.
-
 > **<sup>Syntax</sup>**\
 > _IfExpression_ :\
-> &nbsp;&nbsp; `if` _IfLetList_ [_BlockExpression_]\
-> &nbsp;&nbsp; ( `else` _IfLetList_ [_BlockExpression_] )<sup>\?</sup>
+> &nbsp;&nbsp; `if` _IfConditions_ [_BlockExpression_]\
+> &nbsp;&nbsp; (`else` ( [_BlockExpression_] | _IfExpression_ ) )<sup>\?</sup>
 >
-> _IfLet_ :\
->  &nbsp;&nbsp; [_Expression_]<sub>_except struct expression_</sub>
-> | `let` [_Pattern_] `=` [_Expression_]<sub>_except struct expression_</sub>
+> _IfConditions_ :\
+> &nbsp;&nbsp; _IfCondition_ ( && _IfCondition_ )*
 >
-> _IfLetList_ :\
-> &nbsp;&nbsp; _IfLet_ ( && _IfLet_ )*
-
-## `if`
+> _IfCondition_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_]<sub>_except struct expression_</sub>\
+> &nbsp;&nbsp; | `let` [_Pattern_] `=` [_Scrutinee_]
 
 An `if` expression is a conditional branch in program control.
-The syntax of an `if` expression is a condition operand, followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
-The condition operands must have the [boolean type].
-If a condition operand evaluates to `true`, the consequent block is executed and any subsequent `else if` or `else` block is skipped.
-If a condition operand evaluates to `false`, the consequent block is skipped and any subsequent `else if` condition is evaluated.
+The syntax of an `if` expression is a sequence of one or more condition operands separated by `&&`,
+followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
+
+Condition operands must be either an [_Expression_] with a [boolean type] or a conditional `let` match.
+If all of the condition operands evaluate to `true` and all of the `let` patterns successfully match their [scrutinee]s,
+the consequent block is executed and any subsequent `else if` or `else` block is skipped.
+If any condition operand evaluates to `false` or any `let` pattern does not match its scrutinee,
+the consequent block is skipped and any subsequent `else if` condition is evaluated.
 If all `if` and `else if` conditions evaluate to `false` then any `else` block is executed.
-An if expression evaluates to the same value as the executed block, or `()` if no block is evaluated.
+
+An `if` expression evaluates to the same value as the executed block, or `()` if no block is evaluated.
 An `if` expression must have the same type in all situations.
 
 ```rust
@@ -37,6 +38,7 @@ if x == 4 {
     println!("x is something else");
 }
 
+// `if` can be used as an expression.
 let y = if 12 * 15 > 150 {
     "Bigger"
 } else {
@@ -45,17 +47,15 @@ let y = if 12 * 15 > 150 {
 assert_eq!(y, "Bigger");
 ```
 
-## `if let`
+## `if let` patterns
 
-An `if let` expression is semantically similar to an `if` expression but in place of a condition operand it expects the keyword `let` followed by a pattern, an `=` and a [scrutinee] operand.
-If the value of the scrutinee matches the pattern, the corresponding block will execute.
-Otherwise, flow proceeds to the following `else` block if it exists.
-Like `if` expressions, `if let` expressions have a value determined by the block that is evaluated.
+`let` patterns in an `if` condition allow binding new variables into scope when the pattern matches successfully.
+The following examples illustrate bindings using `let` patterns:
 
 ```rust
 let dish = ("Ham", "Eggs");
 
-// this body will be skipped because the pattern is refuted
+// This body will be skipped because the pattern is refuted.
 if let ("Bacon", b) = dish {
     println!("Bacon is served with {}", b);
 } else {
@@ -63,7 +63,7 @@ if let ("Bacon", b) = dish {
     println!("No bacon will be served");
 }
 
-// this body will execute
+// This body will execute.
 if let ("Ham", b) = dish {
     println!("Ham is served with {}", b);
 }
@@ -73,23 +73,8 @@ if let _ = 5 {
 }
 ```
 
-`if` and `if let` expressions can be intermixed:
-
-```rust
-let x = Some(3);
-let a = if let Some(1) = x {
-    1
-} else if x == Some(2) {
-    2
-} else if let Some(y) = x {
-    y
-} else {
-    -1
-};
-assert_eq!(a, 3);
-```
-
-Multiple patterns may be specified with the `|` operator. This has the same semantics as with `|` in `match` expressions:
+Multiple patterns may be specified with the `|` operator.
+This has the same semantics as with `|` in [`match` expressions]:
 
 ```rust
 enum E {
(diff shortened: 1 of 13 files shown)


src/expressions/if-expr.md
```diff
@@ -4,11 +4,21 @@
...
+>
+> _inIf_ :\
+> &nbsp;&nbsp; ( `let` [_Pattern_] `=` [_Expression_]<sub>_except struct expression_</sub>
+> | [_Expression_]<sub>_except struct expression_</sub>
+> | _inIf_ )
+>
+> _inIfList_ :\
+> &nbsp;&nbsp; _inIf_ ( && _inIf_ )*
+
+[_Expression_]<sub>_except struct expression_</sub>
```
c410-f3r, line 21: @ehuss Before jumping into `while` and `if let`, do you recommend some way to avoid duplication? `if`, `if let` and `while` will mostly have the same grammar

nikomatsakis:
Thanks @c410-f3r

src/expressions/if-expr.md
ehuss: Sorry, I'm not sure I understand the question.  

If I understand the RFC correctly, there no longer is a grammatical difference between `if` and `if let`.  What I would do is:

* Remove distinctions between `if` and `if let`:
     * Remove _IfLetExpression_ as a distinct rule. 
     * Change the name of the chapter to just be "`if` expressions"
     * Rewrite `if-expr.md` so that it doesn't discuss them as two separate kinds of expressions.
* Define _InIfList_ once in `if-expr.md`
* Perform the same treatment to `while let`, removing _PredicatePatternLoopExpression_, and joining those two sections.  In the rules, you can link to other pages, so `while` would become something like:
  > `while` _InIfList_ _BlockExpression_
  
  Where _InIfList_ links to `if-expr.md`

src/expressions/if-expr.md
```diff
@@ -90,27 +89,6 @@ let a = if let Some(1) = x {
...
-is equivalent to
-
-<!-- ignore: expansion example -->
-```rust,ignore
-match EXPR {
-    PATS => { /* body */ },
-    _ => { /* else */ },    // () if there is no else
-}
-```
-
```
c410-f3r, line 113: It hasn't been the case for quite some time-> https://github.com/rust-lang/rust/pull/80357

src/expressions/if-expr.md
c410-f3r: Thanks @ehuss 

The PR was updated according to your indications

src/expressions/if-expr.md
```diff
@@ -126,30 +104,43 @@ if let E::X(n) | E::Y(n) = v {
...
-Use of a lazy boolean operator is ambiguous with a planned feature change of the language (the implementation of if-let chains - see [eRFC 2947][_eRFCIfLetChain_]).
-When lazy boolean operator expression is desired, this can be achieved by using parenthesis as below:
 
-<!-- ignore: psuedo code -->
-```rust,ignore
-// Before...
-if let PAT = EXPR && EXPR { .. }
+## Chains of expressions
+
+It is possible to chain multiple expressions into a single statement to avoid nested declarations.
```
ehuss, line 110: I think "statement" here is a little confusing.  This also doesn't really present any context, or explain how to chain or how chains work. We generally don't use code blocks as a replacement for explaining how something works, but rather as a helpful illustration and example.

Can you try to add some more detail here? I would expect at least a couple paragraphs explaining what chaining is, what the syntax is, what the semantics are, that you can combine boolean values and `let` pattern bindings, how combining multiple pattern bindings works, etc.

src/expressions/if-expr.md
```diff
@@ -126,30 +104,43 @@ if let E::X(n) | E::Y(n) = v {
 ```
 
 The expression cannot be a [lazy boolean operator expression][_LazyBooleanOperatorExpression_].
-Use of a lazy boolean operator is ambiguous with a planned feature change of the language (the implementation of if-let chains - see [eRFC 2947][_eRFCIfLetChain_]).
```
ehuss, line 129: Could this also retain the old note about how to deal with use of lazy boolean expressions when combined with `if let`?  Maybe something like:

> The scrutinee expression cannot be a [lazy boolean operator expression] due to ambiguity and precedence with [chains of expressions].
> If a lazy boolean operator expression is needed, then parentheses can be used.  For example:
>
> ```rust
> if let PAT = ( EXPR && EXPR ) { .. }
> if let PAT = ( EXPR || EXPR ) { .. }
> ```

src/expressions/if-expr.md
```diff
@@ -126,30 +104,43 @@ if let E::X(n) | E::Y(n) = v {
...
+fn single() {
+    let outer_opt = Some(Some(1i32));
+
+    if let Some(inner_opt) = outer_opt
+        && let Some(number) = inner_opt
+        && number == 1
+    {
+        println!("Peek a boo");
+    }
+}
```
ehuss, line 122: This example isn't super clear to me. Can it perhaps be broken up in two and explain what it is illustrating?  Maybe something like this:

~~~suggestion

The following is an example of chaining multiple expressions, mixing `let` bindings and boolean expressions, and with expressions able to reference pattern bindings from previous expressions:

```rust
let outer_opt = Some(Some(1i32));

if let Some(inner_opt) = outer_opt
    && let Some(number) = inner_opt
    && number == 1
{
    println!("Peek a boo");
}
```

The above is equivalent to the following without using expression chains:

```rust
~~~

File: src/expressions/if-expr.md [hunk 6]
```diff
@@ -126,30 +104,43 @@ if let E::X(n) | E::Y(n) = v {
...
+        if let Some(number) = inner_opt {
+            if number == 1 {
+                println!("Peek a boo");
+            }
+        }
+    }
+}
 ```
 
+The only remark is the fact that parenthesis (`if (let Some(a) = opt && (let Some(b) = a)) && b == 1`) and `||` operators (`if let A(x) = e1 || let B(x) = e2`) are not currently supported.
```
ehuss, line 137: The note about `||` not being allowed is already stated above, so I don't think it needs to be repeated here.

And the grammar doesn't actually define `if (let …)` as being valid, so this is a little confusing.  I left a comment on the stabilization PR, but it seems like there is some disconnect between what this documentation is stating and how it actually works.  

Generally the reference doesn't specify what may or may not be supported in the future, so I'm not sure it is worth stating anything.

src/expressions/if-expr.md
```diff
@@ -1,14 +1,22 @@
-# `if` and `if let` expressions
+# `if` expressions
 
-## `if` expressions
+## Syntax
+
+The same syntax is used by `if`, `if let` and chains of expressions.
```
ehuss, line 5: I was perhaps intending maybe a little bit of a greater rewrite here, as I'm not sure I entirely see `if` and `if let` being fundamentally different.  Is there a reason to keep them as separate sections?  

I think what I was thinking is it would say something like:

> The condition operand can be a list of one or more _IfLet_ expressions separated by `&&`.
> If all of the boolean values evaluate to true and all patterns match their scrutinee value, the consequent block is executed and…
> …and include a description of pattern bindings, scopes, `&&` priority, etc.

I guess I'm not seeing why they would be kept separate.  `if EXPR {}` just seems to be a special case where there is an `if` expression with a single boolean value.

src/expressions/if-expr.md
c410-f3r: My mistake, code desugaring only changed for `while` expressions. `for` expressions stay the same

src/expressions/if-expr.md
```diff
@@ -1,14 +1,22 @@
...
-> &nbsp;&nbsp; (`else` (
->   [_BlockExpression_]
-> | _IfExpression_
-> | _IfLetExpression_ ) )<sup>\?</sup>
+> &nbsp;&nbsp; `if` _IfLetList_ [_BlockExpression_]\
+> &nbsp;&nbsp; ( `else` _IfLetList_ [_BlockExpression_] )<sup>\?</sup>
+>
+> _IfLet_ :\
+>  &nbsp;&nbsp; ( [_Expression_]<sub>_except struct expression_</sub>
+> | `let` [_Pattern_] `=` [_Expression_]<sub>_except struct expression_</sub> )
```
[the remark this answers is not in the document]
c410-f3r: Removed

src/expressions/if-expr.md
c410-f3r: Wrote what was suggested in https://github.com/rust-lang/reference/pull/1179#discussion_r850776586

src/expressions/if-expr.md
c410-f3r: Done without invalid snippets

src/expressions/if-expr.md
c410-f3r: Done

src/expressions/if-expr.md
c410-f3r: Removed

src/expressions/if-expr.md
c410-f3r: I don't have enough free time to perform a greater rewrite

src/expressions/if-expr.md
```diff
@@ -1,14 +1,22 @@
...
 
 > **<sup>Syntax</sup>**\
 > _IfExpression_ :\
-> &nbsp;&nbsp; `if` [_Expression_]<sub>_except struct expression_</sub> [_BlockExpression_]\
-> &nbsp;&nbsp; (`else` (
->   [_BlockExpression_]
-> | _IfExpression_
-> | _IfLetExpression_ ) )<sup>\?</sup>
+> &nbsp;&nbsp; `if` _IfLetList_ [_BlockExpression_]\
+> &nbsp;&nbsp; ( `else` _IfLetList_ [_BlockExpression_] )<sup>\?</sup>
```
ehuss, line 10: This grammar doesn't look correct to me. The `else` part seems to have multiple issues:

* Doesn't allow `if foo {} else {}` where it is an unconditional else block.
* Doesn't allow `if foo {} else if bar {}` where there is an `else if`.
* Seems to allow `if foo {} else bar {}` which isn't valid.

src/expressions/if-expr.md
```diff
@@ -125,31 +103,71 @@ if let E::X(n) | E::Y(n) = v {
 }
 ```
 
-The expression cannot be a [lazy boolean operator expression][_LazyBooleanOperatorExpression_].
-Use of a lazy boolean operator is ambiguous with a planned feature change of the language (the implementation of if-let chains - see [eRFC 2947][_eRFCIfLetChain_]).
-When lazy boolean operator expression is desired, this can be achieved by using parenthesis as below:
+The expression cannot be a [lazy boolean operator expression][_LazyBooleanOperatorExpression_]. Scrutinee expressions also cannot be a [lazy boolean operator expression] due to ambiguity and precedence with [chains of expressions].
```
ehuss, line 106: These links seem to be broken.

Also, can you keep it to one sentence per line?

src/expressions/if-expr.md
ehuss: Do you mind if I push some changes to your branch?

src/expressions/if-expr.md
c410-f3r: Done

src/expressions/if-expr.md
c410-f3r: Not at all

ehuss (approved):
Thanks!

I pushed the changes that I was thinking of in terms of unifying if/if let and while/while let.

PR #1180, opened 2022-03-16
Upper bound on hashes in raw string literals

GrishaVar:


Havvy:
Has this been decided upon by the lang team, or is this just what it currently is in rustc?

ehuss:
I started a conversation at https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/Max.20raw.20string.20delimiters/near/275921936, which contains some links to the history here.

I don't think this change can be made easily, as we'll probably need to go through some process to decide on an appropriate limit (and to decide how to apply limits to the language in general).

And, BTW, this change just modifies the set of examples. I don't think those are proper definitions. If we do decide to go forward with something like this, the lexer definition and text describing the format will need to be updated.

Havvy:
The reference follows the same train as the compiler. So anything that is merged as stable on the compiler can be documented in the reference repository and it'll show up in the stable docs at the same time. I want to merge this, but @ehuss 's last paragraph still needs to be resolved.

> And, BTW, this change just modifies the set of examples. I don't think those are proper definitions. If we do decide to go forward with something like this, the lexer definition and text describing the format will need to be updated.

GrishaVar:
For the formal definitions, both instances of `0 or more` were replaced with `fewer than 256`. I reckon inclusion of zero and exclusion of 256 should be self-evident. Is this enough?

It might be a good idea to also mention the consequence of the upper limit, namely that`(~IsolatedCR)*` and `ASCII*` musn't contain more than 254 consecutive hash signs. I think that might need overhauling more definitions though, since the kleene as given can't restrict values in this way.

pnkfelix:
I'm nominating this for lang design team (T-lang) discussion so that we can figure out if we want include this as part of the reference spec (or if we want to encode the limit in some other way. E.g., maybe "implementation defined limit on the number of octothorpes in a raw-string literal")

PR #1181, opened 2022-03-17
Document the `target_feature_11` feature

LeSeulArtichaut:
cc tracking issue https://github.com/rust-lang/rust/issues/69098

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -53,6 +53,7 @@ features. It uses the [_MetaListNameValueStr_] syntax with a single key of
 `enable` whose value is a string of comma-separated feature names to enable.
 
 ```rust
+# #[cfg(target_feature = "avx2")]
 #[target_feature(enable = "avx2")]
 fn foo_avx2() {}
 ```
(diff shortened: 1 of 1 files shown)


joshtriplett:
We talked about this in today's lang meeting. This looks good, and we should merge it as soon as we stabilize the feature.

src/attributes/codegen.md
```diff
@@ -66,20 +65,42 @@ It is [undefined behavior] to call a function that is compiled with a feature
...
 if the platform explicitly documents this to be safe.
 
-Functions marked with `target_feature` are not inlined into a context that
-does not support the given features. The `#[inline(always)]` attribute may not
-be used with a `target_feature` attribute.
+For this reason, a function marked with `target_feature` is unsafe, except in
+a context that supports the given features. For example:
+
+```rust
+# #[target_feature(enable = "avx2")]
```
ehuss, line 72: This will similarly need a cfg to restrict it to platforms that support avx2.  You can do something similar to #1334.

src/attributes/codegen.md
```diff
@@ -66,20 +65,42 @@ It is [undefined behavior] to call a function that is compiled with a feature
...
+#[target_feature(enable = "avx2")]
+fn bar_avx2() {
+    // Calling `foo_avx2` here is safe.
+    foo_avx2();
+    || foo_avx2();
+}
+```
+
+Like unsafe functions, functions marked with `target_feature` cannot be
+assigned to a safe function pointer and do not implement `FnOnce`.
```
ehuss, line 92: Can this clarify that it *can* be used with an unsafe function pointer?  Like another sentence along the lines of "They may be assigned to an `unsafe` function pointer, for example:"

And also include an example, perhaps something like:

```rust
#[target_feature(enable = "avx2")]
fn fn_ptr_example() {}

// `unsafe` is required here
static MY_FN_PTR: unsafe fn () -> () = fn_ptr_example;
```

src/attributes/codegen.md
```diff
@@ -66,20 +65,42 @@ It is [undefined behavior] to call a function that is compiled with a feature
 that is not supported on the current platform the code is running on, *except*
 if the platform explicitly documents this to be safe.
 
-Functions marked with `target_feature` are not inlined into a context that
-does not support the given features. The `#[inline(always)]` attribute may not
-be used with a `target_feature` attribute.
+For this reason, a function marked with `target_feature` is unsafe, except in
+a context that supports the given features. For example:
```
ehuss, line 69: This doesn't really define what "a context that supports the given features" means. Can you add a little more detail here? For example, the RFC seems to contain more well-defined text as to exactly what this means. I'd like this to be able to address:

* The context means functions that have at least the exact same `target_features` enabled.
* This does not include implied features (that is, a fn with sse2 can't call a function with sse, even though sse2 implicitly enabled sse).
* This does not include features enabled by default on the platform, or manually enabled as compiler flags.
* Also include a short discussion of what is or is not allowed with trait definitions and impls.

ehuss:
It looks like there are some changes in the implementation recently. Should those be reflected in the documentation here? For example:
* https://github.com/rust-lang/rust/pull/108651 — It might not hurt to explicitly mention that it is not allowed on main.
* https://github.com/rust-lang/rust/pull/108983 — I'm not sure, but I don't think there is anything to add here? My only thought is that the current wording of "safe trait method implementations" is maybe ambiguous, since default trait methods aren't normally called an "implementation" (usually I would reserve that for `impl`).

Is there anything I missed?

traviscross:
In discussing this today, we note the recent related work here:

- https://github.com/rust-lang/rust/pull/127731

ehuss:
Closing as superseded by https://github.com/rust-lang/reference/pull/1720. Thanks!

PR #1182, opened 2022-03-19
Fixed inconsistency in the usage of semicolon at end of scopes

IsakNyberg:
Other scopes have a semicolon at end of their scopes, but these two do not have one.


ehuss (approved):
Thanks!

PR #1184, opened 2022-03-24
Update aarch64 to use neon as fp

workingjubilee:
As discussed in https://github.com/rust-lang/rust/pull/91608
At the very least, which feature to use for assembly should be clear.

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -148,7 +148,6 @@ Reference Manual], or elsewhere on [developer.arm.com].
 
 > ***Note***: The following pairs of features should both be marked as enabled
 > or disabled together if used:
-> - `fp` and `neon`, in order facilitate inlining in more places, among other reasons.
 > - `paca` and `pacg`, which LLVM currently implements as one feature.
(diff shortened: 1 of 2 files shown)


workingjubilee:
@ehuss I know that the reference is considered non-normative at the time, but we might want to strike the entire "implications" column for both the x86-64 and AArch64 suites of features, as they depend on a property of LLVM and not of Rust. What say you?

ehuss:
I don't object to removing the column. I think my concern was the situation where someone has `#[target_feature(enable = "foo")]` and they rely on it also enabling the `bar` feature.  If in some hypothetical future that implicit feature is removed (like in a different backend), that could break the author's expectations.  

I agree it would be better for `rustc` to be a little more explicit about how these are enabled.  There also hasn't been much discussion about this behavior, and it isn't entirely clear to me if it is even all that important.

ehuss (approved):
Thanks!  I'm going to merge assuming I understand https://github.com/rust-lang/rust/pull/91608 correctly.

PR #1186, opened 2022-03-26
clarify that references size_of_val can never exceed isize::MAX

RalfJung:
This is the underlying UB that https://github.com/rust-lang/rust/pull/95252 and https://github.com/rust-lang/rust/pull/95295 fix.

I also removed the statement 
> if the span is empty, "dangling" is the same as "non-null"

since it contradicts what [the library docs](https://doc.rust-lang.org/nightly/std/ptr/index.html) say about zero-sized accesses/types.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -78,11 +78,12 @@ reading uninitialized memory is permitted are inside `union`s and in "padding"
 A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
-pointer value and the size of the pointee type (using `size_of_val`). As a
-consequence, if the span is empty, "dangling" is the same as "non-null". Note
-that slices and strings point to their entire range, so it is important that the length
-metadata is never too large. In particular, allocations and therefore slices and strings
-cannot be bigger than `isize::MAX` bytes.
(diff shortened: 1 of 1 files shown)


ehuss:
I'm a bit unclear about the interaction of ZST and dangling (I'm likely misunderstanding this). Per the std docs:

> However, casting any non-zero integer literal to a pointer is valid for zero-sized accesses, even if some memory happens to exist at that address and gets deallocated.

This seems to provide an exception to the dangling rule.  That is, something like this?

```rust
// This points to some arbitrary location that was not allocated.
let x: *mut () = 1024 as *mut ();
unsafe {
    // This should not be undefined behavior?
    let y = *x;
}
```

If so, would it make sense to add a similar sentence here?  Otherwise, this seems to be in contradiction to the std docs, since this description of "dangling" doesn't carve out this exception.

RalfJung:
> This seems to provide an exception to the dangling rule. 

Well, it clarifies what the definition of "dangling" is when the size is 0.

The reference has so far been more vague on this subject, but yeah I can make that more clear here as well. We probably want an entire subsection on the question of "dangling pointers" then, since it is all rather subtle. However, this interacts closely with a bunch of questions around pointer provenance, which we are not (yet) willing to give final answers to...

ehuss (approved):
Thanks!  Given the discussion in https://github.com/rust-lang/rust/pull/95295#issuecomment-1097011792 and the std docs and other places, I'm going to go ahead and merge.

PR #1189, opened 2022-04-07
Boolean literal expressions

### LINKED ISSUE #1167: Tokens `true` and `false` are defined twice.

mattheww:
Closes #1167

This removes the `BOOLEAN_LITERAL` token, as `true` and `false` are already listed as keywords.

Then fills in the "Boolean literal expressions" section (one more small piece of #939).

### DIFF
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -150,8 +150,12 @@ The value of the expression is determined from the string representation of the 
 
 A boolean literal expression consists of one of the keywords `true` or `false`.
 
-> **Note**: This section is incomplete.
+The expression's type is the primitive [boolean type], and its value is:
+ * true if the keyword is `true`
+ * false if the keyword is `false`
 
+
+[boolean type]: ../types/boolean.md
 [constant expression]: ../const_eval.md#constant-expressions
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks!  

I don't feel strongly one way or the other in this case.  I can see how BOOLEAN_LITERAL is somewhat different from all the others. However, I also suspect that the original intent was to just give a label to the concept of "boolean", and to not be a perfectly minimal non-overlapping definition.  And it seems a little unfortunate to lose that label.  However, I don't think it matters too much and this looks fine to me.

PR #1191, opened 2022-04-12
Inline assembly: Clarify references to quoted regs

fw-immunant:
For `lateout` and similar operand direction specifiers, a quoted argument must be an explicit register name, per the EBNF grammar we give. As such, we should not use quotes when describing syntax used with register classes.

Further, correct the EBNF: in EBNF, quoted strings are always terminals, but `"<explicit register>"` is intended to be a nonterminal surrounded by literal (terminal) quotation marks. Moving the quotes to their own terminal symbols does require escaping, but makes the terminal/nonterminal distinction clearer. The only other place double-quotes are present in the grammar is inside `STRING_LITERAL`/`RAW_STRING_LITERAL`, so calling attention to this case is worthwhile for clarity.

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -41,7 +41,7 @@ The following ABNF specifies the general syntax:
 ```text
 format_string := STRING_LITERAL / RAW_STRING_LITERAL
 dir_spec := "in" / "out" / "lateout" / "inout" / "inlateout"
-reg_spec := <register class> / "<explicit register>"
+reg_spec := <register class> / "\"" <explicit register> "\""
 operand_expr := expr / "_" / expr "=>" expr / expr "=>" "_"
 reg_operand := dir_spec "(" reg_spec ")" operand_expr
 operand := reg_operand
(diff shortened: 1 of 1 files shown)


ehuss:
This looks good to me. @Amanieu do you perhaps have any comments?

src/inline-assembly.md
```diff
@@ -184,7 +184,7 @@ Here is the list of currently supported register classes:
 >
 > - On x86-64 the high byte registers (e.g. `ah`) are not available in the `reg_byte` register class.
 >
-> - Some register classes are marked as "Only clobbers" which means that they cannot be used for inputs or outputs, only clobbers of the form `out("reg") _` or `lateout("reg") _`.
```
Amanieu, line 187: This change is incorrect: clobber-only registers can only be used with explicit registers, not register classes.

src/inline-assembly.md
```diff
@@ -184,7 +184,7 @@ Here is the list of currently supported register classes:
...
+> - Some register classes are marked as "Only clobbers" which means that they cannot be used for inputs or outputs, only clobbers of the form `out(<register class>) _` or `lateout(<register class>) _`.
```
fw-immunant, line 187:
```suggestion
> - Some register classes are marked as "Only clobbers" which means that registers in these classes cannot be used for inputs or outputs, only clobbers of the form `out(<explicit register>) _` or `lateout(<explicit register>) _`.
```

Amanieu (approved):

ehuss (approved):
Thanks!

PR #1192, opened 2022-04-14
Add stable references of `macro_metavar_expr`

c410-f3r:
Requested in https://github.com/rust-lang/rust/pull/95860

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -193,6 +193,53 @@ compiler knows how to expand them properly:
     not have the same number. This requirement applies to every layer of nested
     repetitions.
 
+## Dollar-dollar ($$)
+
+`$$` expands to a single `$`.
+
+Since metavariable expressions always apply during the expansion of a macro, they cannot be used in recursive macro definitions and this is where `$$` expressions comes into play, i.e., `$$` can be used to resolve ambiguities in nested macros.
+
+The following example illustrates a macro that fails to compile due to the ambiguity of the repetition in a nested macro:
+
+```rust,compile_fail
+macro_rules! foo_error {
+    () => {
+        macro_rules! bar_error {
+            ( $( $any:tt )* ) => { $( $any )* };
+            // ^^^^^^^^^^^ error: attempted to repeat an expression containing no syntax variables matched as repeating at this depth
+        }
+    };
+}
+
+foo_error!();
+```
+
+The following resolves the problem by escaping the `$` in the repetition with `$$`:
+
+```rust
+macro_rules! foo_ok {
+    () => {
+        macro_rules! bar_ok {
+            ( $$( $any:tt )* ) => { $$( $any )* };
+        }
+    };
+}
+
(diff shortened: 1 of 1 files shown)


src/macros-by-example.md
```diff
@@ -193,6 +193,76 @@ compiler knows how to expand them properly:
...
+}
+```
+
+The above snippet will expand to a sequence of numbers respecting the number of times that `ignored_identifier` repeats, resulting in a final `0 + 1 + 1 + 1` output.
+
+### Dollar-dollar ($$)
+
+`$$` expands to a single `$`.
+
+Since metavariable expressions always apply during the expansion of a macro, they cannot be used in recursive macro definitions and this is where `$$` expressions comes into play, i.e., `$$` can se used to resolve ambiguities in nested macros.
```
Skgland, line 227:
```suggestion
Since metavariable expressions always apply during the expansion of a macro, they cannot be used in recursive macro definitions and this is where `$$` expressions comes into play, i.e., `$$` can be used to resolve ambiguities in nested macros.
```

src/macros-by-example.md
```diff
@@ -193,6 +193,76 @@ compiler knows how to expand them properly:
...
+
+```rust
+#![feature(macro_metavar_expr)]
+
+macro_rules! foo_error {
```
ehuss, line 232: Can you split this code block into two separate blocks?  That is, something like the following:

~~~

The following example illustrates a macro that fails to compile due to the ambiguity of the repetition in the nested macro:


```rust,compile_fail
macro_rules! foo_error {
…
}

foo_error!();
```

The following resolves the problem by escaping the `$` in the repetition with `$$`:

```rust
macro_rules! foo_ok {
…
}

foo_ok!();
```
~~~

ehuss (commented):
Thanks!

src/macros-by-example.md
c410-f3r: Done as requested

src/macros-by-example.md
c410-f3r: Thanks

ehuss:
@c410-f3r It looks like this was pared down, can you update this to reflect what was finalized?

c410-f3r:
Done

ehuss (approved):
Thanks!

I opened #1229 with a curious behavior of `$` matching, if you want to take a look. I don't feel it is too directly tied to this change, but seemed relevant.

c410-f3r:
> Thanks!
> 
> I opened #1229 with a curious behavior of `$` matching, if you want to take a look. I don't feel it is too directly tied to this change, but seemed relevant.

I am not an authority in this subject so I guess @petrochenkov can give a better opinion.

petrochenkov:
No opinion on https://github.com/rust-lang/reference/issues/1229.
If it produces an error then it's fine, more options for the future.
If it doesn't produce an error, and making it an error breaks code, then it's also fine.

PR #1194, opened 2022-04-17
Document the Termination trait for main() and test functions

### LINKED ISSUE #1193: Update for Termination trait
The [`Termination`](https://doc.rust-lang.org/nightly/std/process/trait.Termination.html) trait is now stable. There are several places in the reference that should be updated:

* The return type of `main` discussed here: https://github.com/rust-lang/reference/blob/master/src/crates-and-source-files.md#main-functions
* The return type of `#[test]` functions discussed here: https://github.com/rust-lang/reference/blob/master/src/attributes/testing.md#the-test-attribute
* I would add it to the [Special types and traits](https://github.com/rust-lang/reference/blob/master/src/special-types-and-traits.md) chapter. Just mention it as the return type of the above two points, and link to them.

mattheww:
Closes #1193

There doesn't seem to be a satisfactory way to document when tests are considered to pass: the implementation checks whether `ExitCode::to_i32()` returns 0, but `to_i32()` is intentionally hidden and marked as perma-unstable, and nothing in `ExitCode`'s docs provides a notion of a "successful" code.

So I've vaguely said "whether the resulting `ExitCode` represents successful termination", and listed some specific cases.

### DIFF
--- a/src/attributes/testing.md
+++ b/src/attributes/testing.md
@@ -9,16 +9,12 @@ enables the [`test` conditional compilation option].
 
 The *`test` attribute* marks a function to be executed as a test. These
 functions are only compiled when in test mode. Test functions must be free,
-monomorphic functions that take no arguments, and the return type must be one
-of the following:
+monomorphic functions that take no arguments, and the return type must implement the [`Termination`] trait, for example:
 
 * `()`
-* `Result<(), E> where E: Error`
-<!-- * `!` -->
-<!-- * Result<!, E> where E: Error` -->
-
-> Note: The implementation of which return types are allowed is determined by
-> the unstable [`Termination`] trait.
+* `Result<(), E> where E: Debug`
+* `!`
+<!-- * Result<!, E> where E: Debug` -->
 
 <!-- If the previous section needs updating (from "must take no arguments"
   onwards, also update it in the crates-and-source-files.md file -->
@@ -26,9 +22,12 @@ of the following:
 > Note: The test mode is enabled by passing the `--test` argument to `rustc`
 > or using `cargo test`.
 
-Tests that return `()` pass as long as they terminate and do not panic. Tests
-that return a `Result<(), E>` pass as long as they return `Ok(())`. Tests that
-do not terminate neither pass nor fail.
+The test harness calls the returned value's [`report`] method, and classifies the test as passed or failed depending on whether the resulting [`ExitCode`] represents successful termination.
+In particular:
+* Tests that return `()` pass as long as they terminate and do not panic.
+* Tests that return a `Result<(), E>` pass as long as they return `Ok(())`.
+* Tests that return `ExitCode::SUCCESS` pass, and tests that return `ExitCode::FAILURE` fail.
+* Tests that do not terminate neither pass nor fail.
 
 ```rust
 # use std::io;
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!

Although they specifically did not stabilize `to_i32`, an `ExitCode` can be created from any arbitrary `u8`:

```rust
use std::process::{ExitCode, Termination};

struct Foo {
    code: u8,
}

impl Termination for Foo {
    fn report(self) -> ExitCode {
        ExitCode::from(self.code)
    }
}

fn main() -> Foo {
    Foo { code: 123 }
}

#[test]
fn failing() -> Foo {
    Foo { code: 99 }
}
```

So, at least for tests, I think it should be defined as any non-zero exit code (that's how I wrote it in the [tests chapter](https://doc.rust-lang.org/nightly/rustc/tests/index.html)).

Also, something I didn't really think about is whether or not to document the return code of the process for the exit code of `main`.  I think it probably should be documented, but it seems like there could be some platform-specific issues here.

src/attributes/testing.md
```diff
@@ -9,26 +9,25 @@ enables the [`test` conditional compilation option].
...
 
-Tests that return `()` pass as long as they terminate and do not panic. Tests
-that return a `Result<(), E>` pass as long as they return `Ok(())`. Tests that
-do not terminate neither pass nor fail.
+The test harness calls the returned value's [`report`] method, and classifies the test as passed or failed depending on whether the resulting [`ExitCode`] represents successful termination.
+In particular:
+* Tests that return `()` pass as long as they terminate and do not panic.
+* Tests that return a `Result<(), E>` pass as long as they return `Ok(())`.
+* Tests that return `ExitCode::SUCCESS` pass, and tests that return `ExitCode::FAILURE` fail.
+* Tests that do not terminate neither pass nor fail.
```
ehuss, line 30: I'm concerned this may be over-specifying the behavior. There's a bit of an awkward separation between "Rust the language" and "the standard library".  The reference typically tends to focus on just the language behavior, and not delve into the standard library's implementation.

Also, I'm not really comfortable with specifying the non-termination behavior. I would consider that a failure, and it is possible in the future that either libtest or cargo could have some sort of timeout controls which would consider the test to fail.

I think leaving this up to the `Termination` docs in the standard library should be sufficient.

mattheww:
I've filed #1196 with my suggestion for how what to say about the process exit status.

mattheww:
I don't think writing "non-zero exit code" works, because the standard library docs don't say that all exit codes have a numeric value, and in particular they don't say that `SUCCESS` is zero and `FAILURE` isn't (presumably because they want it be possible for them not to be on some platforms).

I would be much happier if the Reference text could simply defer to an `ExitCode::is_success()` method, but that's evidently not an option for Rust 1.61.

If we want to avoid the Reference having a list like the one in this PR, perhaps an alternative is to make the standard library docs classify exit codes as "successful" or "unsuccessful", even in the absence of a public way to retrieve that classification.

src/attributes/testing.md
mattheww: The "Tests that do not terminate neither pass nor fail." sentence was present already; this PR only moves it.

ehuss:
Good points!  And thanks for opening that issue, it does seem like something that needs some input.

Then I think I'm good with this PR as-is.  Did you want to make any edits, or is it ready to go?

mattheww:
I think this is ready to go.

ehuss (approved):
Thanks!

PR #1198, opened 2022-04-22
Attempt to explain that &mut cannot alias 

XrXr:
This is a clumsy attempt at making the "`&mut` is an exclusive
reference" part of the aliasing rule more discoverable. I feel like this
is an inherent property of the type so I was surprised to not find that
information here. This is echoing parts of the docs for [UnsafeCell] and
[ptr::as_mut], so another place where a newcomer could potentially get
this info.

I think it's easy for learners to assume that `&mut` can alias since
many popular languages have comparatively lax aliasing rules.

---

Do you also think adding this here make sense? Maybe I only think so because I've been thinking about this recently. 

[UnsafeCell]: https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html
[ptr::as_mut]: https://doc.rust-lang.org/std/primitive.pointer.html#method.as_mut-1

### DIFF
--- a/src/types/pointer.md
+++ b/src/types/pointer.md
@@ -26,6 +26,11 @@ These also point to memory owned by some other value.
 A mutable reference type is written `&mut type` or `&'a mut type`.
 A mutable reference (that hasn't been borrowed) is the only way to access the value it points to, so is not `Copy`.
 
+While a mutable reference is alive, it is the only reference pointing to the referent; no other reference,
+mutable or shared, point to the same value. For the safe subset of the language, this constraint is
+verified by the compiler. For code that use `unsafe`, it's up to the programmer to follow this constraint to avoid
+[undefined behavior] for correctness.
+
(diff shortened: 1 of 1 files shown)


XrXr:
So I just saw https://github.com/rust-lang/unsafe-code-guidelines/issues/133 and realized that the way I'm explaining this is maybe too restrictive. What I wanted to do is to add another place where people can discover the current rules, even though not everything is worked out at the moment.

Maybe the reference is not a good place to raise awareness about Stacked Borrows so I'm closing this. I'll point people at other pieces of documentation scattered around that hint that the current aliasing rules.

Havvy:
Stacked borrows is implemented in miri, but rustc is more permissive than it. The actual rules aren't really well known (and I think they're inconsistent in the same way that clang's is), but stacked borrows is a set of rules that works on top of it. And then there's Ganka's "quit transmuting integers to pointers" thing on top of that that miri now has a flag to run under as well.

PR #1199, opened 2022-05-01
Say that that the default function return type is the unit type

mattheww:
I've tried to make minimal changes to the text, because I think there are other open PRs affecting this paragraph.


ehuss (approved):
Thanks!

PR #1200, opened 2022-05-02
Remove OuterAttributes for `StructPatternEtCetera`

pushkine:
Rust in stable and nightly treats outer attributes above `..` as whitespace.
Running `cargo fmt` even removes it.

Snippet:
```rust
struct Foo {
	a: u8
}

fn main() {
	let Foo { #[rest_attribute] .. } = Foo { a: 1 };
	let Foo { #[prop_attribute] a  } = Foo { a: 1 };
}
```

`cargo run`:
rest_attribute is ignored
```md
error: cannot find attribute `prop_attribute` in this scope
 --> src\main.rs:7:14
  |
7 |     let Foo { #[prop_attribute] a  } = Foo { a: 1 };
  |                 ^^^^^^^^^^^^^^
```

`cargo fmt`:
rest_attribute is removed
```rs
struct Foo {
    a: u8,
}

fn main() {
    let Foo { .. } = Foo { a: 1 };
    let Foo {
        #[prop_attribute]
        a,
    } = Foo { a: 1 };
}

```

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -585,7 +585,6 @@ Reference patterns are always irrefutable.
 > &nbsp;&nbsp; )
 >
 > _StructPatternEtCetera_ :\
-> &nbsp;&nbsp; [_OuterAttribute_] <sup>\*</sup>\
 > &nbsp;&nbsp; `..`
 
 [_OuterAttribute_]: attributes.md


ehuss:
Thanks for the PR!  I don't think this is the correct change, as rustc does parse these attributes. This particular thing is a known issue tracked in https://github.com/rust-lang/rust/issues/81282, where there is more discussion if this should be allowed or not.

pushkine:
@ehuss 
The fact that they are parsed and silently ignored is a bug, not a feature

Unless this is a "macro" situation, like visibility specifiers that are allowed but ignored in certain contexts?
In that case this should be mentioned, note on the other hand that `syn` does not parse that case.

https://astexplorer.net/#/gist/4a0c99b76353c971afe9576405234183/b8c362bf14d52215dd8410294069b10a2a305b18

ehuss:
I don't think this is necessarily a macro situation.  It's just a bug, and the lang team needs to make a call as to whether or not attributes should be allowed in that position.  I personally would vote that it should be allowed to be consistent with other field patterns (and https://github.com/rust-lang/rust/issues/81282#issuecomment-822376089 gives a motivating example).

pushkine:
Is the argument that this is an agreed upon incoming feature that is already parsed by rustc?

I don't think that this is a feature that exists anywhere outside docs

- The idea of such a feature was introduced by the docs 4+ years ago, there is no RFC supporting its implementation

- The compiler does not even define `..` as a node with attributes, there is a bug in `rustc_parse` skipping attributes silently
```rs
    /// A struct or struct variant pattern (e.g., `Variant {x, y, ..}`).
    /// The `bool` is `true` in the presence of a `..`.
    Struct(Option<QSelf>, Path, Vec<PatField>, bool),
```
https://github.com/rust-lang/rust/blob/21d613b1116eef86d613d5f7f10d8fa5f22e27b3/compiler/rustc_ast/src/ast.rs#L725-L727

ehuss:
I'm not sure I understand your question.  Attributes were added in https://github.com/rust-lang/rust/pull/38814. There wasn't any discussion of attributes on the `..` rest part. It parses it, but ignores it. I think that was an oversight, so the current behavior is a bug.  Either rustc should honor the attributes, or it should reject them.  That's a question for the lang team to answer in https://github.com/rust-lang/rust/issues/81282.

ehuss:
I'm going to close. Further followup should be done in https://github.com/rust-lang/rust/issues/81282, and either `rustc` or the reference can be updated based on the outcome of that issue.

PR #1201, opened 2022-05-04
Fix crate_type attribute examples

bstrie:
As a crate-level attribute, this must be used as an inner attribute.


ehuss (approved):
Thanks!

PR #1202, opened 2022-05-04
Set type when purposely discarding return type

kraktus:
Raised by https://github.com/rust-lang/rust-clippy/issues/6841, as it is safer in case of refactoring.

It does not catch all, and for complex type return is probably not worth it in terms of readability but still think it's a good addition.

Wording can probably be improved.


ehuss:
Thanks for the PR!  I would prefer to not include too many style suggestions in the reference.  There are too many guidelines to add here, and there is not a specific process for deciding guidelines on how Rust code should be written. Something like this would probably be better served in something like clippy.

PR #1203, opened 2022-05-04
#[must_use] on traits also affects trait objects

### LINKED ISSUE #796: #[must_use] on traits also affects trait objects
The reference says:

> When used on a trait declaration, a call expression of an expression statement to a function that returns an impl trait of that trait violates the unused_must_use lint.

But it also applies to `dyn Trait`, not just `impl Trait`.

botahamec:
This is meant to resolve #796 

Closes #796


ehuss (approved):
Thanks!

PR #1204, opened 2022-05-05
Don't use PathPattern in RangePattern bounds

flodiebold:
The use of `PathPattern` instead of `PathExpression` in `RangePattern` is IMO confusing, since these values are not patterns and are not represented as patterns in the AST.


ehuss (approved):
Seems reasonable, thanks!

PR #1205, opened 2022-05-05
Inline assembly: Add kreg0 register class

Amanieu:
This was added by https://github.com/rust-lang/rust/pull/95740.


ehuss (approved):
Thanks!

PR #1207, opened 2022-05-07
Stop saying that const functions cannot use 'extern'

### LINKED ISSUE #1197: Update for const_extern_fn
https://github.com/rust-lang/rust/pull/95346 stabilized the `const_extern_fn` feature which allows `const extern fn` and `const unsafe extern fn`.   https://github.com/rust-lang/rust/issues/64926 also contains some more information.

The paragraph [here](https://github.com/rust-lang/reference/blob/b5f6c2362baf932db9440fbfcb509b309237ee85/src/items/functions.md?plain=1#L222-L223) needs to be updated for this change.

mattheww:
Closes #1197

I don't think there's anything to do other than remove the claim in the text that these are forbidden. The grammar already allows it.

rust-lang/rust#64926 says this is stabilised only for the "Rust" and "C" ABIs, but I think that would be better covered in documentation for particular ABIs, if/when that exists.

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -216,8 +216,7 @@ Functions qualified with the `const` keyword are [const functions], as are
 [tuple struct] and [tuple variant] constructors. _Const functions_  can be
 called from within [const contexts].
 
-Const functions are not allowed to be [async](#async-functions), and cannot
-use the [`extern` function qualifier](#extern-function-qualifier).
+Const functions are not allowed to be [async](#async-functions).
 
 ## Async functions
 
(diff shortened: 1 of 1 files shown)


mattheww:
I've added that. I'm a bit worried that it might get stale.

Ideally I'd subscribe to the F-const_extern_fn label in the Rust repository but I don't think there's a way, unless one of the bots makes that possible?

ehuss:
Thanks!  Yea, there is a bit of risk for things getting stale (it often happens). Hopefully most language changes go through an FCP which I usually monitor.  I also suspect there will be little demand for other ABIs.

As for subscribing to issues, I *think* you can do it (I can't remember if you have to be an org member or not). In the "Watch" drop-down at the top, choose "Custom", then next to the "Issues" checkbox, there should be a "Filter by labels":

image

ehuss (approved):

PR #1208, opened 2022-05-07
Moved the option variant imports

### LINKED ISSUE #1115: Please add an example of "use" being used inside a function
`use` can be used inside a function, but that's not always obvious to a newcomer. (I, for example, didn't realize this was possible from the documentation.) It would be nice if the `use-declarations.html` page had a brief example of it so it's more obvious.

botahamec:
#1115 asks for an example of a `use` declaration inside of a function. I think the easiest way to do that is to move the already existing declaration inside of the `main` function.

Closes #1115

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -35,13 +35,15 @@ Use declarations support a number of convenient shortcuts:
 An example of `use` declarations:
 
 ```rust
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Seems reasonable, thanks!

PR #1212, opened 2022-05-09
Add a note to the turbofish section about impl Trait

nrc:
This is part of the stabilisation process for [explicit_generic_args_with_impl_trait](https://github.com/rust-lang/rust/issues/83701); it will be legal to use turbofish in the presence of `impl Trait` argument types, but there is still no way to specify those types


nrc:
@ehuss https://github.com/rust-lang/rust/pull/96868 has now landed, so this can be reviewed/merged now.

ehuss (approved):
Thanks!

Issue #1213, opened 2022-05-09
Remove reference to `#[cold]` in section on match expressions


e-rhodes:
The section [Attributes on match arms](https://doc.rust-lang.org/stable/reference/expressions/match-expr.html?highlight=cold#attributes-on-match-arms) lists `#[cold]` as one of the macros that can be applied to match guards:

> The only attributes that have meaning on match arms are [cfg](https://doc.rust-lang.org/stable/reference/conditional-compilation.html), [cold](https://doc.rust-lang.org/stable/reference/attributes/codegen.html#the-cold-attribute), and the [lint check attributes](https://doc.rust-lang.org/stable/reference/attributes/diagnostics.html#lint-check-attributes).

However, attempting to do so generates this warning:

```
     |
1272 |                 #[cold]
     |                 ^^^^^^^
     |
     = note: `#[warn(unused_attributes)]` on by default
     = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
     = note: see issue #80564 <https://github.com/rust-lang/rust/issues/80564> for more information
```

Therefore, it seems like the reference to `cold` should be removed here.

ehuss:
Yea, I believe `cold` only applies to functions.

ehuss:
Closed by #1216

PR #1216, opened 2022-05-13
Update "Match Expressions" to remove 'cold'

merelymyself:
As per issue #1213 , `#[cold]` only applies to functions.


ehuss (approved):
Thanks!

PR #1217, opened 2022-05-13
Fix backtick-typo in procedural-macros.md

chdalski:


src/procedural-macros.md
```diff
@@ -51,7 +51,7 @@ token, `.` is a `Punct` token, and `1.2` is a `Literal` token. The `TokenStream`
 type, unlike `Vec<TokenTree>`, is cheap to clone.
 
 All tokens have an associated `Span`. A `Span` is an opaque value that cannot
-be modified but can be manufactured. `Span`s represent an extent of source
+be modified but can be manufactured. `Spans` represent an extent of source
```
eddyb, line 54: There is no type named `Spans` AFAIK, this is just pluralizing \`Span\`, which is common in Rust documentation, when wanting to talk about multiple/any/all values of a type.

ehuss:
I'm going to close as this isn't a typo, but an intentional way to pluralize a defined name.

PR #1220, opened 2022-05-19
Add tile registers to inline-assembly documentation

chorman0773:
This documents the availablity of the tmm registers for clobbers in inline assembly for x86-64. It also documents that they are clobbered by the x86-64 abi_clobbers.


ehuss:
@Amanieu Can you review this?

Amanieu (approved):

ehuss (approved):
Thanks!

PR #1221, opened 2022-05-20
Reflect changes about unsizing casts in const context

AnthonyMikh:
Since Rust 1.61.0 it is possible to do unsizing casts in const context, so reflect that in reference.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -42,9 +42,8 @@ to be run.
 * The [dereference operator] except for raw pointers.
 * [Grouped] expressions.
 * [Cast] expressions, except
-  * pointer to address casts,
-  * function pointer to address casts, and
-  * unsizing casts to trait objects.
+  * pointer to address casts and
+  * function pointer to address casts.
 * Calls of [const functions] and const methods.
 * [loop], [while] and [`while let`] expressions.
 * [if], [`if let`] and [match] expressions.


ehuss:
Thanks!  It's not entirely clear to me that this is related to 1.61.  From what I can tell, unsizing casts in a const fn has been stable since https://github.com/rust-lang/rust/pull/85078 (1.54).  Unsizing casts for dyn trait in a const context seem to go back further, but I can't pinpoint it (at least 1.28?)

@oli-obk Do you perhaps remember what this line was referring to?  The only thing I can find is https://github.com/rust-lang/rust/issues/64992, where @RalfJung noticed that `dyn` unsizing is disallowed, but only in const fn.  I tried making an example where `dyn` unsizing was rejected in a const context, but kept running into other const limitations preventing me from creating an example.

RalfJung:
Mentioning dyn traits in `const fn` (in any expression, including casts) was disallowed until https://github.com/rust-lang/rust/pull/93827. So 1.61 sounds right I think.

ehuss:
This section defines what is allowed in any const context (not just const fn).  From what I can tell, dyn traits in something like a `const` works fine since at least 1.28 (needed `const_let` till 1.33, 1.27 and earlier complains about not allowing blocks).  But perhaps I am misunderstanding about what it means about an "unsizing cast"?

(Sorry for the questions, I'm trying to fully understand what this means.)

RalfJung:
> This section defines what is allowed in any const context (not just const fn).

Well, the rules were different between const and const fn for quite a while. (They still are different, but now floating point operations are the only difference.)

In const/static initializers, this has worked since 1.0:
```rust
trait Trait: Send + Sync {}
impl Trait for i32 {}

static X: i32 = 5;
static Y: &'static Trait = &X; // implicit dyn Trait unsizing
```
and this has worked since 1.2:
```rust
trait Trait: Send + Sync {}
impl Trait for i32 {}

static X: i32 = 5;
static Y: &'static Trait = &X as &Trait; // explicit dyn Trait unsizing cast
```

ehuss:
Hm, OK, I'm still unsure then why it was included since there is a separate section that explicitly lists the differences for const fn.

There are several other errors in the list. It would be nice if someone could audit it, though I'm not sure if anyone really reads this page.

Thanks @AnthonyMikh for the fix!

ehuss (approved):

PR #1224, opened 2022-05-27
modify confusing variance example

lcnr:
Given `fn(T) -> T` the type `T` is only invariant while wrapped inside of an item. If you have `fn(T) -> T` directly you can freely "shrink" the `T` in return position or "grow" the `T` in argument position. E.g.

```rust
fn split_lt<'a, 'short, 'long>(
    value: fn(&'a ()) -> &'a (),
) -> fn(&'long ()) -> &'short ()
where
    'a: 'short,
    'long: 'a,
{
    value
}

```

### DIFF
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -61,25 +61,51 @@ Variance of types is automatically determined as follows
 | `[T]` and `[T; n]`            |                   | covariant         |
 | `fn() -> T`                   |                   | covariant         |
 | `fn(T) -> ()`                 |                   | contravariant     |
-| `fn(T) -> T`                  |                   | invariant         |
 | `std::cell::UnsafeCell<T>`    |                   | invariant         |
 | `std::marker::PhantomData<T>` |                   | covariant         |
 | `dyn Trait<T> + 'a`           | covariant         | invariant         |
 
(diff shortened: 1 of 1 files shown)


ehuss:
This was added in #874 where Niko reviewed it as being correct.  @nikomatsakis Do you perhaps have any comments on this case?

lcnr:
Extended that section to talk about `fn(T) -> T` a bit more in-depth

nikomatsakis (approved):

nikomatsakis:
Seems right. To be 100% pedantic, you would talk about the "variance of a (struct/enum/union) type `Foo<T, U>` with respect to its type parameters `T` (or `U`)". i.e., if `Foo` is covariant with respect to `T`, then if you have `Foo<A>` and `Foo<B>` for some types `A` and `B`, `Foo<A> <: Foo<B>` iff `A <: B`.

Given that `fn` is not a struct/enum/union type that is parameterized by `T`, it's not really *correct* to say that `fn(T)` is contravariant with respect to `T`. You say that `fn` is contravariant with respect to its *arguments* (and covariant with respect to its return type).

ehuss:
Merging per Niko's approval.

PR #1225, opened 2022-05-28
Clarify the implication of using `no_mangle`.

ramnivas:
The `no_mangle` attribute on a function seems to imply `pub extern "C"`,
but it is not clear from the documentation. Thus in the wild, you see
that specify both `#[no_mangle]` and `pub extern "C"`.

This PR clarifies this implication.


bjorn3:
`#[no_mangle]` doesn't imply any calling convention. By default you still have the Rust calling convention which is unstable.

ramnivas:
Thanks for clarifying.

PR #1232, opened 2022-06-24
Describe what `windows_subsystem` does

ChrisDenton:
The reference documents the syntax of the `windows_subsystem` attribute but neglects to describe what it's intended to do. This is my attempt to rectify that.


ehuss (approved):
Thanks!

PR #1237, opened 2022-07-09
Document new `#[expect]` attribute and `reasons` parameter (RFC 2383)

xFrednet:
This PR is created in preparation to stabilize the `lint_reasons` feature. I hope this covers everything :upside_down_face:

Tracking issue: https://github.com/rust-lang/rust/issues/54503

### DIFF
--- a/src/attributes/diagnostics.md
+++ b/src/attributes/diagnostics.md
@@ -16,8 +16,8 @@ For any lint check `C`:
 * `#[allow(C)]` overrides the check for `C` so that violations will go
    unreported.
 * `#[expect(C)]` indicates that lint `C` is expected to be emitted. The
-  attribute will suppres the emission of `C` or issue a warning, if the
-  expectation is unfillfilled.
+  attribute will suppress the emission of `C` or issue a warning, if the
+  expectation is unfulfilled.
 * `#[warn(C)]` warns about violations of `C` but continues compilation.
 * `#[deny(C)]` signals an error after encountering a violation of `C`,
 * `#[forbid(C)]` is the same as `deny(C)`, but also forbids changing the lint
@@ -69,7 +69,7 @@ pub mod m2 {
 }
 ```
 
-This example shows how one can use `forbid` to disallow uses of `allow` or 
+This example shows how one can use `forbid` to disallow uses of `allow` or
 `expect` for that lint check:
 
 ```rust,compile_fail
@@ -92,7 +92,7 @@ All lint attributes support an additional `reason` parameter, to give context wh
 a certain attribute was added. This reason will be displayed as part of the lint
 message if the lint is emitted at the defined level.
 
-```rust,edition2015
+```edition2015,fail
 // `keyword_idents` is allowed by default. Here we deny it to
 // avoid migration of identifies when we update the edition.
 #![deny(
@@ -143,7 +143,7 @@ fn main() {
 
     // This `#[expect]` attribute creates a lint expectation that will be fulfilled, since
     // the `answer` variable is never used. The `unused_variables` lint, that would usually
-    // be emitted, is supressed. No warning will be issued for the statement or attribute.
+    // be emitted, is suppressed. No warning will be issued for the statement or attribute.
     #[expect(unused_variables)]
     let answer = "SpongeBob SquarePants!";
 }
@@ -178,7 +178,7 @@ fn select_song() {
 }
 ```
 
-If the `expect` attribute contains several lints, each one is expected separatly. For a
+If the `expect` attribute contains several lints, each one is expected separately. For a
 lint group it's enough if one lint inside the group has been emitted:
 
 ```rust
@@ -193,7 +193,7 @@ pub fn another_example() {
     // This attribute creates two lint expectations. The `unused_mut` lint will be
     // suppressed and with that fulfill the first expectation. The `unused_variables`
     // wouldn't be emitted, since the variable is used. That expectation will therefore
-    // be unsatified, and a warning will be emitted.
+    // be unsatisfied, and a warning will be emitted.
     #[expect(unused_mut, unused_variables)]
     let mut link = "https://www.rust-lang.org/";
 
@@ -202,8 +202,7 @@ pub fn another_example() {
 ```
 
 > Note: The behavior of `#[expect(unfulfilled_lint_expectations)]` is currently
-> defined to always generate the `unfulfilled_lint_expectations` lint. This may
-> change in the future.
+> defined to always generate the `unfulfilled_lint_expectations` lint.
 
 ### Lint groups
 
@@ -514,7 +513,6 @@ error[E0277]: My Message for `ImportantTrait<i32>` implemented for `String`
 [let statement]: ../statements.md#let-statements
 [macro definition]: ../macros-by-example.md
 [module]: ../items/modules.md
-[RFC 2383]: https://rust-lang.github.io/rfcs/2383-lint-reasons.html
 [rustc book]: ../../rustc/lints/index.html
 [rustc-lint-caps]: ../../rustc/lints/levels.html#capping-lints
 [rustc-lint-cli]: ../../rustc/lints/levels.html#via-compiler-flag
--- a/src/attributes/diagnostics.md
+++ b/src/attributes/diagnostics.md
@@ -92,9 +92,9 @@ All lint attributes support an additional `reason` parameter, to give context wh
 a certain attribute was added. This reason will be displayed as part of the lint
 message if the lint is emitted at the defined level.
 
-```edition2015,fail
+```rust,edition2015,compile_fail
 // `keyword_idents` is allowed by default. Here we deny it to
-// avoid migration of identifies when we update the edition.
+// avoid migration of identifiers when we update the edition.
 #![deny(
     keyword_idents,
     reason = "we want to avoid these idents to be future compatible"
--- a/src/attributes/diagnostics.md
+++ b/src/attributes/diagnostics.md
@@ -14,9 +14,10 @@ to which the attribute applies.
 For any lint check `C`:
 
 * `#[allow(C)]` overrides the check for `C` so that violations will go
-   unreported,
-* `#[expect(C)]` suppresses all lint emissions of `C`, but will issue
-   a warning if the lint wasn't emitted in the expected scope.
+   unreported.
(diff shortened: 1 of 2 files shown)


xFrednet:
The CI will fail, until the `#[feature(lint_reasons)]` has been stabilized in rustc. Everything else should be ready :upside_down_face:

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+
+With the `#[expect]` attributes lints can be expected in a certain scope. If
+this expectation is not fulfilled a new warning is emitted to the user. The
+lint levels can be overridden with other lint attributes as usual.
+
+```rust
+#[warn(missing_docs)]
+pub mod m2{
+    #[expect(missing_docs)]
+    pub mod nested {
```
ehuss, line 116: This example may be a little confusing, because these `mod` statements also trigger the missing_docs lint, meaning that the `undocumented_one` below isn't necessary. I'm not sure how to avoid that without too much complication, but maybe just add some doc comments to the modules?

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+    /// This comment explains something cool about the function. The
+    /// expectation will not be fulfilled and in turn issue a warning.
+    pub fn undocumented_too() -> i32 { 3 }
+}
+```
+
+> Note: Lint expectations have been proposed in [RFC 2383]. It was not defined
+> how expectations of the expectation lint should be handled. The rustc
+> implementation currently doesn't allow the expextation of the
+> `unfulfilled_lint_expectation` lint. This can change in the future.
```
ehuss, line 136: I'm not sure the link to the RFC is particularly important here.  There are also a few typos here (`unfulfilled_lint_expectation` (no s) and `expextation`). It is also not clear to me what it means by "currently doesn't allow", as it seems to allow it just fine:

```rust
#![feature(lint_reasons)]
#![expect(unfulfilled_lint_expectations)]
```

Produces:

```
warning: this lint expectation is unfulfilled
 --> src/lib.rs:2:11
  |
2 | #![expect(unfulfilled_lint_expectations)]
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unfulfilled_lint_expectations)]` on by default
  = note: the `unfulfilled_lint_expectations` lint can't be expected and will always produce this message
```

It does seem like it is *always* a warning, even if `unfulfilled_lint_expectations` is triggered (which seems strange).
Is there some more context here? 

Perhaps it could say something like:

> **Note**: The behavior of `#[expect(unfulfilled_lint_expectations)]` is currently defined to always generate the `unfulfilled_lint_expectations` lint. This may change in the future.

Perhaps this should be clarified before it is stabilized?

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+
+    file_name
+}
+```
+
+### Lint expectations
+
+With the `#[expect]` attributes lints can be expected in a certain scope. If
+this expectation is not fulfilled a new warning is emitted to the user. The
+lint levels can be overridden with other lint attributes as usual.
```
ehuss, line 110: This sentence wasn't really clear to me what it meant. I might suggest breaking the example up into two parts. The first example can show where `unfulfilled_lint_expectations` is unfulfilled. Then, after that example, write a small paragraph and a second example that explains how *changing* the level of a lint within an expectation scope will override the `expect` attribute. Perhaps something like:

----

If the level of a lint is modified within an `expect` scope with a `warn` or `deny` attribute, then triggering that lint does *not* satisfy the `expect` attribute, and the `unfulfilled_lint_expectations` will also trigger. Similarly, squelching a lint with `allow` will also prevent the lint from being fulfilled.

----

This behavior seems kinda confusing to me, and it is not really clear to me how it is working, or if that is really intended behavior? 

I might also include some text explaining the interaction with `forbid`, since that also isn't entirely clear to me. I believe it is something like "if a lint has previously been marked with `forbid`, then it is not allowed to use the `expect` attribute with that lint".

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+pub mod m2{
+    #[expect(missing_docs)]
+    pub mod nested {
+        // This missing documentation fulfills the expectation above
+        pub fn undocumented_one() -> i32 { 1 }
+
+        // Missing documentation signals a warning here, despite the expectation
+        // above. This emission would not fulfill the expectation
+        #[warn(missing_docs)]
+        pub fn undocumented_two() -> i32 { 2 }
```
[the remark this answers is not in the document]
ehuss: Oh, I think I was a bit confused by the absence of docs on the modules.  I see that this does prevent it from being fulfilled.  This seems a little confusing, per my comment below.

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+    #[allow(unused_mut, reason = "this is only modified on some platforms")]
+    let mut file_name = PathBuf::from("git");
+
+    #[cfg(target_os = "windows")]
+    file_name.set_extension("exe");
+
+    file_name
+}
+```
+
```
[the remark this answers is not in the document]
xFrednet: I'll add another example in the next commit :)

src/attributes/diagnostics.md
```diff
@@ -83,6 +85,56 @@ pub mod m3 {
...
+### Lint expectations
+
+With the `#[expect]` attributes lints can be expected in a certain scope. If
+this expectation is not fulfilled a new warning is emitted to the user. The
```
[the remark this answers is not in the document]
xFrednet: I'm unsure about this one, I intentionally left it a bit vague, since I feel like the way the warning is emitted and the lint name should not directly be part of the language specification.

I might be overthinking this, but this was my initial impression and the reason I wrote it like this :upside_down_face:

src/attributes/diagnostics.md
xFrednet: Ahh, nice catch

src/attributes/diagnostics.md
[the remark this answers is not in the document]
xFrednet: Good call, I fixed the example. I thought about using another lint, but decided against it, since this example was already used and might therefore be simpler to understand :)

src/attributes/diagnostics.md
xFrednet: > Note: The behavior of `#[expect(unfulfilled_lint_expectations)]` is currently defined to always generate the `unfulfilled_lint_expectations` lint. This may change in the future.

Yes this is what I intended to say. The context sadly cannot be condensed to one link. This behavior was implemented in https://github.com/rust-lang/rust/pull/94670. This has been discussed in the RFC PR (https://github.com/rust-lang/rfcs/pull/2383#issuecomment-396794262) with the suggestion to allow the `unfulfilled_lint_expectations` to be caught by outer expectations.

This can also be implemented in rustc. However, I decided against this for two reasons:
1. I didn't see a benefit. In the RFC Discussion, it was also said that there seems to be no use case for `#[expect(unfulfilled_lint_expectations)]`
2. Supporting `#[expect(unfulfilled_lint_expectations)]` would add a bunch of complexity. Currently, the lint is emitted at the item the attribute is attached to (For reasons regarding, incremental compilation and hash stability). To support such an expectation, we would need to either store the lint level and then order the expectations by which *could* fulfill which one or depend on the processing order which has caused some issues in the past. So, either solution would add complexity to the implementation without having an use case. We can also not simply emit the lint at the parent item due to the way lint levels are processed.

---

btw how did you add the `(i) Note` marker? I can sadly not see the MD text

src/attributes/diagnostics.md
xFrednet: > This behavior seems kinda confusing to me, and it is not really clear to me how it is working, or if that is really intended behavior?

This is intended behavior, see https://github.com/rust-lang/rust/issues/54503#issuecomment-1179701134

I've added more examples and documentation. :upside_down_face:

xFrednet:
With https://github.com/rust-lang/rust/issues/115980#issuecomment-1926001471 the semantics of the `#[expect]` attribute have been decided. I've reworked some parts of the documentation, to make the semantics clearer according to the decision.

src/attributes/diagnostics.md
```diff
@@ -83,6 +86,125 @@ pub mod m3 {
...
+    #[expect(unused_mut, unused_variables)]
+    let mut link = "https://www.rust-lang.org/";
+
+    println!("Welcome to our community: {link}");
+}
+```
+
+> Note: The behavior of `#[expect(unfulfilled_lint_expectations)]` is currently
+> defined to always generate the `unfulfilled_lint_expectations` lint. This may
+> change in the future.
```
ehuss, line 206: In what way might this change?

src/attributes/diagnostics.md
```diff
@@ -322,6 +444,7 @@ When used on a function in a trait implementation, the attribute does nothing.
 [let statement]: ../statements.md#let-statements
 [macro definition]: ../macros-by-example.md
 [module]: ../items/modules.md
+[RFC 2383]: https://rust-lang.github.io/rfcs/2383-lint-reasons.html
```
ehuss, line 447: This does not appear to be used.

```suggestion
```

src/attributes/diagnostics.md
xFrednet: The original RFC had a discussion about expecting the `unfulfilled_lint_expectations` lint, but never came up with a valid use case. The current behavior is the result of the way that rustc emits lints. If someone comes up with a valid use case, I think it would be valid to change the behavior of expecting this specific lint, but I honestly can't think of one.

Maybe it would be better, to remove the last sentence? (I'll do that in the typo fix commit)

xFrednet:
https://github.com/rust-lang/rust/pull/120924 has been merged and this feature will be part of Rust 1.81 :tada:

This PR should be ready for review. We'll have to wait one nightly to allow the example code blocks to compile in CI. I'll rebase it tomorrow. I'll be gone over the weekend, but can gat back to it afterwards :D

xFrednet:
I've rebased and the CI is green :tada: It should be ready for review an maybe merge :D

ehuss (approved):
Thanks! I pushed up a small fix.

PR #1238, opened 2022-07-17
update union field type rules

RalfJung:
Cc https://github.com/rust-lang/rust/pull/97995#issuecomment-1186263679
r? @ehuss


ehuss (approved):
Thanks!

PR #1239, opened 2022-07-19
Remove outdated note about non-Copy union fields

GoldsteinE:
Unions with non-Copy fields are now stable.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -359,8 +359,6 @@ struct MyCFields { x: u32, y: u8 }
 struct MyDFields;
 ```
 
-> Note: `union`s with non-`Copy` fields are unstable, see [55149].
-
 ### Primitive representations
 
 The *primitive representations* are the representations with the same names as


Mark-Simulacrum:
I think this isn't quite right - only a small sunset of non-Copy types is permitted, and probably should be documented here.

GoldsteinE:
I thought any non-Copy types without custom destructors are allowed? I’m going to re-read that issue, thanks for correction.

ehuss:
We recently changed the documentation for union drop requirements (#1238). I think it would be fine to keep some kind of note here, just as a reminder as to why `derive(Copy)` is included in the example. Perhaps it could be something like:

> **Note**: Union fields are restricted to a certain subset of types, such as deriving `Copy` in this example. See [Unions](https://github.com/rust-lang/reference/blob/master/src/items/unions.md) for more information.

Also, there are two identical notes in this chapter that need to be updated.

PR #1244, opened 2022-07-29
Add documentation for raw-dylib and link_ordinal

dpaoliello:
* Documents the `raw-dylib` link kind (https://github.com/rust-lang/rust/issues/58713), including how it differs from `dylib`.
* Documents the `link_ordinal` attribute (https://github.com/rust-lang/rust/pull/89025).

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -122,6 +122,9 @@ specifies the kind of library with the following possible values:
 - `static` — Indicates a static library.
 - `framework` — Indicates a macOS framework. This is only valid for macOS
   targets.
+- `raw-dylib` - Indicates a dynamic library where the compiler will generate
+  an import library to link against (see [`dylib` versus `raw-dylib`] below
+  for details). This is only valid for Windows targets.
 
 The `name` key must be included if `kind` is specified.
 
@@ -198,9 +201,26 @@ The default for this modifier is `-whole-archive`.
 More implementation details about this modifier can be found in
 [`whole-archive` documentation for rustc].
 
+#### `dylib` versus `raw-dylib`
+
+On Windows, linking against a dynamic library requires that an import library
+is provided to the linker: this is a special static library that declares all
+of the symbols exported by the dynamic library in such a way that the linker
+knows that they have to be dynamically loaded at runtime.
+
+Specifying `kind = "dylib"` instructs the Rust compiler to link an import
+library based on the `name` key, the linker will then use its normal library
+resolution logic to find that import library. Alternatively, specifying
+`kind = "raw-dylib"` instructs the compiler to generate an import library
+during compilation and provide that to the linker instead.
+
+`raw-dylib` is only supported on Windows and not supported on x86
+(`target_arch="x86"`), see [58713]. Using it when targeting other platforms or
+x86 on Windows will result in a compiler error.
+
 ### The `link_name` attribute
 
-The `link_name` attribute may be specified on declarations inside an `extern`
+The *`link_name` attribute* may be specified on declarations inside an `extern`
 block to indicate the symbol to import for the given function or static. It
 uses the [_MetaNameValueStr_] syntax to specify the name of the symbol.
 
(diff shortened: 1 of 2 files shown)


src/items/external-blocks.md
```diff
@@ -198,6 +201,19 @@ The default for this modifier is `-whole-archive`.
 More implementation details about this modifier can be found in
 [`whole-archive` documentation for rustc].
 
+#### `dylib` versus `raw-dylib`
+
+On Windows, linking against a dynamic library requires that an import library
+is provided to the linker: this is a special static library that declares all
+of the symbols exported by the dynamic library in such a way that the linker
+knows that they have to be dynamically loaded at runtime.
```
bjorn3, line 209: raw-dylib can be extended to other platforms in the future. For example on macOS there are .tbd files which are roughly equivalent to import libraries. And on ELF based systems it should be possible to synthesize an .so file with SONAME set to the name of the target dylib and all symbols declared with a fake value. The RFC for raw-dylib mentions other platforms as future possibility. Maybe you could rewrite this section a bit to note that other platforms have similar requirements that raw-dylib could fix, but that this is only implemented on Windows currently?

dpaoliello: Added a note about other platforms and only Windows being supported at the moment.

src/items/external-blocks.md
```diff
@@ -211,6 +230,40 @@ extern {
 }
 ```
 
+Using this attribute with the `link_ordinal` attribute will result in a
+compiler error.
+
+### The `link_ordinal` attribute
+
```
ehuss, line 237: There is generally a template for defining an attribute that looks like:

```
The *`link_ordinal` attribute* …
```

The first sentence should briefly explain what the attribute does. I'm not familiar with this, but it might look something like this:

The *`link_ordinal` attribute* specifies the numeric ordinal of an item in an [`extern` block](external-blocks.md) to link against.

ehuss (commented):
Thanks!

src/items/external-blocks.md
```diff
@@ -211,6 +230,40 @@ extern {
...
+This attribute is only used with the `raw-dylib` linking kind.
+It is ignored if used with any other kind.
```
dpaoliello, line 239: @ehuss thoughts on this? (Should we raise an error if `link_ordinal` is used for other linking kinds?)

src/items/external-blocks.md
```diff
@@ -198,6 +201,22 @@ The default for this modifier is `-whole-archive`.
...
+knows that they have to be dynamically loaded at runtime.
+
+Specifying `kind = "dylib"` instructs the Rust compiler to link an import
+library based on the `name` key, the linker will then use its normal library
+resolution logic to find that import library. Alternatively, specifying
+`kind = "raw-dylib:` instructs the compiler to generate an import library
+during compilation and provide that to the linker instead.
+
+Some non-Windows platforms have concepts similar to import libraries, but Rust
+currently only supports `raw-dylib` on Windows.
```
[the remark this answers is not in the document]
dpaoliello: I modified this to be a simple note that `raw-dylib` is not supported on Windows: I know that it is repeating the note above, but I want to make sure that anyone jumping to this section is clear about why it only talks about Windows.

src/items/external-blocks.md
dpaoliello: Ok, rewrote this so that the first sentence explains what the attribute does, and then I follow up with an explanation of what an ordinal is.

src/items/external-blocks.md
ehuss: It seems reasonable to make that an error.  The RFC seemed to imply that it can only be specified with raw-dylib.

dpaoliello: Ok, updated the docs and created a PR to change the behavior: https://github.com/rust-lang/rust/pull/100069

dpaoliello:
@ehuss or @michaelwoerister can I please get a review?

ehuss:
Seems fine to me. It looks like there is an exception for x86 that isn't mentioned here. Perhaps a small note like "is only supported on Windows and is not supported on `target_arch="x86"`" or something like that.

Merging this has to wait for the rust stabilization PR to merge.

dpaoliello:
> Seems fine to me. It looks like there is an exception for x86 that isn't mentioned here. Perhaps a small note like "is only supported on Windows and is not supported on `target_arch="x86"`" or something like that.
> 
> Merging this has to wait for the rust stabilization PR to merge.

@ehuss The stabilization PR has been merged, and I added a not about being unsupported on x86.

ehuss (approved):
Thanks!

I pushed a few minor editing changes. I removed the link to 58713 since we generally try to shy away from linking to issues. I recognize that they can provide some helpful context, but there are countless that are relevant to large parts of the reference, and it just doesn't scale to link them.

I was also on the fence about whether or not to make that compatibility note a side-note, or to make it prescriptive about what is supported. That is, does Rust the Language make any stance on which targets this attribute supports? This hits a gray area where I'm not sure. I'm not a fan of letting things be implementation-defined. However, this feels like a very implementation-defined thing.  It can always be changed later if someone feels that it should be decided by the implementation.

PR #1245, opened 2022-08-08
a fix about .await

liigo:


ehuss (approved):
Thanks!

cc @yoshuawuyts in case you want to double check, but this seems to be correct based on my understanding of the desugaring.

yoshuawuyts (commented):
Yep, this looks great - thanks for the ping @ehuss!

PR #1246, opened 2022-08-08
Be less specific about the representation of `+bundle`

petrochenkov:
Same as https://github.com/rust-lang/rust/pull/100074, but for the reference.


ehuss (approved):
Thanks!

PR #1249, opened 2022-08-14
#[non_exhaustive] on variant blocks cross-crate as casts

lambinoo:
Fixes: #1250 

Adding missing documentation after changing in rust-lang/rust#91161 (rust-lang/rust#92744)

### DIFF
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -127,7 +127,7 @@ match message {
 }
 ```
 
-It's also impossible to cast non-exhaustive types from foreign crates.
+It's also not allowed to cast non-exhaustive types from foreign crates.
 ```rust, ignore
 use othercrate::NonExhaustiveEnum;
 
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -127,6 +127,14 @@ match message {
 }
 ```
 
+It's also impossible to cast non-exhaustive types from foreign crates.
+```rust, ignore
+use othercrate::NonExhaustiveEnum;
+
+// Cannot cast a non-exhaustive enum outside of its defining crate.
+let _ = NonExhaustiveEnum::default() as u8;
+```
+
 Non-exhaustive types are always considered inhabited in downstream crates.
 
 [_MetaWord_]: ../attributes.md#meta-item-attribute-syntax


src/attributes/type_system.md
```diff
@@ -127,6 +127,14 @@ match message {
 }
 ```
 
+It's also impossible to cast non-exhaustive types from foreign crates.
```
ehuss, line 130: Just a minor wording suggestion. I'm not sure why, but "impossible" doesn't seem to quite fit here for me. It is possible, it just results in an error. I kinda view things that are "impossible" as fundamentally not even possible to try.

```suggestion
It's also not allowed to cast non-exhaustive types from foreign crates.
```

ehuss (commented):
Thanks!

src/attributes/type_system.md
lambinoo: Instant answer!

I agree with your logic here. The MR in the rust repository forbade this cast, so it was possible beforehand.

ehuss (approved):
Thanks!

mixy1:
Is this still the case as of today? 

```rust
pub use keyboard_types::{Code, Modifiers};

fn main() {
    let key: Code = Code::KeyR;
    println!("{}", key as u32);
}
```

Works just fine, https://docs.rs/keyboard-types/latest/keyboard_types/enum.Code.html and it's indeed marked as non-exhaustive.

obi1kenobi:
A quick check in the playground confirms the rule is a bit more involved, and the reference is inaccurate: [playground (make sure to run "test" not "build")](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c75b7c52eba06bb967b744a8fdc80097)
```rust
/// ` ``rust
/// use playground::Example;
///
/// let value = Example::default();
/// println!("{}", value as isize)
/// ` ``
#[non_exhaustive]  // it doesn't matter whether this is commented-out or not
#[derive(Default)]
pub enum Example {
    #[default]
    First,
    
    // uncommenting the next line breaks the doctest:
    // #[non_exhaustive]
    Second,
}
```

The [reference states](https://github.com/rust-lang/reference/blob/master/src/attributes/type_system.md#the-non_exhaustive-attribute) that "it's also not allowed to cast non-exhaustive types from foreign crates" and gives an example of (failing while) casting the value of a non-exhaustive enum. This is inaccurate.

Per my example above, the rule as implemented today appears to be "it's not allowed to cast enums from foreign crates if the enum contains a non-exhaustive variant."

In the case of `keyboard_types`, the `Code` enum is non-exhaustive but all its variants are exhaustive, so under my inferred rule it's allowed. I agree that the reference seems to say that wouldn't work, and I assume a PR would be welcome. Just a note that RustConf is this week, so feedback on the PR would likely be delayed.

obi1kenobi:
It may also be a good idea to update the reference page on enum variants and discriminants, since the section that describes numeric `as` casting doesn't mention `#[non_exhaustive]` at all: https://doc.rust-lang.org/reference/items/enumerations.html#casting

In `cargo-semver-checks`, we were just in the middle of implementing some new lints for numeric `as` casting of enums based on that reference page. After reading that page, I didn't know about the `#[non_exhaustive]` rule for variants until trying it in the playground just now, which is less than ideal 😅

mixy1:
Not sure if this is the right place to ask, but what's the point of marking a variant as non_exhaustive? It intuitively makes sense to mark the enum as non_exhaustive but I can't understand why one would use it for a variant.

obi1kenobi:
It can't be constructed by any downstream crate, and pattern-matching it requires allowing for extra fields with `..` since exhaustive matching is disallowed. This means it can have additional fields added to it without causing an exhaustiveness-related major breaking change.

mixy1:
I've opened PR #1607

PR #1251, opened 2022-08-14
Revert let chains reference docs

compiler-errors:
Reverts #1179 because let-chains are getting unstabilized in rust-lang/rust/#100538

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -62,7 +62,7 @@
         - [Closure expressions](expressions/closure-expr.md)
         - [Loop expressions](expressions/loop-expr.md)
         - [Range expressions](expressions/range-expr.md)
-        - [If expressions](expressions/if-expr.md)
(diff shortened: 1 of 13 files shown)


ehuss:
Is this definitely going to be reverted on the master branch?

compiler-errors:
Well yes, I think so, but regardless, it seems we can't revert on beta without reverting here apparently? https://github.com/rust-lang/rust/pull/100538#issuecomment-1214442616

ehuss:
I can make a beta branch to revert here if it isn't going to be reverted on master.

I'll assume it will be reverted and go ahead and merge here on master. There have been a few small changes to the reference since the beta branch, but I think they are fine to pull along (none of them are version-specific).

ehuss (approved):

ehuss:
You should be able to update the submodule to the latest master on your beta branch. Let me know if you need any help.

compiler-errors:
Makes sense, thanks for clarifying.

PR #1254, opened 2022-08-22
allow to quickly edit a page directly on github

tshepang:


ehuss (approved):
Thanks! I'm a little concerned that this might encourage low-quality PRs, but hopefully it will lean more towards useful fixes.

Also, using the `edit` link can be a little problematic since you are forced to log in to GitHub to view the page. We'll see if anyone reports issues with that.

tshepang:
> ... using the `edit` link can be a little problematic since you are forced to log in to GitHub to view the page. We'll see if anyone reports issues with that.

Why is that a concern, since one has already viewed the page by the time they are clicking the link?

ehuss:
If a random internet user is browsing a page like https://doc.rust-lang.org/reference/items/modules.html, and they click on the edit link in the upper-right corner, they are greeted with a page asking for a password. They don't need a GitHub account to browse https://doc.rust-lang.org/.  I think it could be a confusing experience if you are not familiar with what the button does, or are not familiar with GitHub.

An alternative is to use a URL like `https://github.com/rust-lang/reference/tree/master/{path}`. A user clicking on that link doesn't need a GitHub account, and will be presented with the page source view. If they have a GitHub account, they can click the little pencil icon to edit it.

It's not a big deal. The `/edit/` path is the intended use of that feature. I just don't like the user experience if someone isn't already logged into GitHub.

PR #1256, opened 2022-08-27
Update examples of what implements `Termination`

mattheww:
rust-lang/rust#97803 in Rust 1.63 added some implementations of `Termination`; I think it's worth updating the notes in the Reference to match.


ehuss (approved):
Thanks!

PR #1259, opened 2022-09-02
Add trait_upcasting related languages changes

crlf0710:
cc https://github.com/rust-lang/rust/issues/65991

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -34,6 +34,7 @@ code.
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
   does not support (see [`target_feature`]), *except* if the platform explicitly documents this to be safe.
+* Performing non-nop coercion on a dangling or unaligned raw pointer.
 * Calling a function with the wrong call ABI or unwinding from a function with the wrong unwind ABI.
 * Producing an invalid value, even in private fields and locals. "Producing" a
   value happens any time a value is assigned to or read from a place, passed to
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -172,7 +172,7 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
 
-* `dyn T` to `dyn U`, when `T` has `U` as one of its ancestor trait.
+* `dyn T` to `dyn U`, when `U` is one of `T`'s supertraits.
 
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
     * `Foo` is a struct.
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -159,7 +159,7 @@ Coercion is allowed between the following types:
 ### Unsized Coercions
 
 The following coercions are called `unsized coercions`, since they
-relate to converting sized types to unsized types, and are permitted in a few
+relate to converting types to unsized types, and are permitted in a few
 cases where other coercions are not, as described above. They can still happen
 anywhere else a coercion can occur.
 
@@ -171,6 +171,8 @@ an implementation of `Unsize<U>` for `T` will be provided:
 * `[T; n]` to `[T]`.
 
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
+
+* `dyn T` to `dyn U`, when `T` has `U` as one of its ancestor trait.
 
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
     * `Foo` is a struct.


src/type-coercions.md
```diff
@@ -172,6 +172,8 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
 
+* `dyn T` to `dyn U`, when `T` has `U` as one of its ancestor trait.
```
compiler-errors, line 175:
```suggestion
* `dyn T` to `dyn U`, when `U` is one of `T`'s supertraits.
```

compiler-errors: I don't think we use "ancestor traits" conventionally anywhere in the compiler, so I think we should prefer a term that *is* used in the compiler.

src/behavior-considered-undefined.md
```diff
@@ -34,6 +34,7 @@ code.
 * Invoking undefined behavior via compiler intrinsics.
 * Executing code compiled with platform features that the current platform
   does not support (see [`target_feature`]), *except* if the platform explicitly documents this to be safe.
+* Performing non-nop coercion on a dangling or unaligned raw pointer.
```
ehuss, line 37: I'm not exactly clear why this is being added in this PR, since I can't find any reference to it related to RFC 3324. Can you say more about why this is being added? Is there some context here? 

cc @RalfJung since you often know about UB stuff.

compiler-errors: I believe it was decided in a T-lang FCP in https://github.com/rust-lang/rust/issues/101336, the stabilization of trait_upcasting just becomes the first time that this behavior can be exercised, as far as I recall.

ehuss:
> What is the status here? The feature is being stabilized so maybe we should merge the reference changes :)

The reference changes have to wait for the stabilization PR to merge.

Sorry I didn't get to reviewing this earlier, I did not know it was being stabilized, and I tend to not prioritize pre-stabilization PRs when it seems like they are a long way off from being stabilized.

src/behavior-considered-undefined.md
compiler-errors, line 37: Though now that I read this, I think this should be clarified though to integrate language which more precisely reflects the decision in rust-lang/rust#101336. 

Specifically, should we mention that the *vtable pointer* needs to point to a valid vtable?

src/type-coercions.md
```diff
@@ -172,6 +172,8 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
 
+* `dyn T` to `dyn U`, when `U` is one of `T`'s supertraits.
```
compiler-errors, line 175: This should also mention that upcasting coercions allow dropping auto traits. i.e. `dyn Trait + Send -> dyn Trait`.

src/behavior-considered-undefined.md
RalfJung: My expectation was that it is completely okay for the pointer itself to be dangling or unaligned; only the vtable must be valid. And we already require the vtable to be valid; we say it is UB when one creates the following:
```
Invalid metadata in a wide reference, `Box<T>`, or raw pointer
```

So I don't think any change is needed to the UB list. At least I highly doubt we want to require the data part of the wide raw pointer to be valid in any way for such a coercion.

Cc @rust-lang/opsem

src/type-coercions.md
```diff
@@ -159,7 +159,7 @@ Coercion is allowed between the following types:
 ### Unsized Coercions
 
 The following coercions are called `unsized coercions`, since they
-relate to converting sized types to unsized types, and are permitted in a few
+relate to converting types to unsized types, and are permitted in a few
 cases where other coercions are not, as described above. They can still happen
 anywhere else a coercion can occur.
```
RalfJung, line 164: IMO it is worth stating explicitly that "unsizing" is a misnomer now and this also covers coercions from unsized types to unsized types.

ehuss:
Ping @crlf0710 Do you think you'd be able to address the comments here?

RalfJung:
This feature is being [reverted](https://github.com/rust-lang/rust/pull/120233#issuecomment-1903912456) so we should not land this PR.

That said, it'd still be good to have the PR ready for when this gets stabilized for real. :)

ehuss:
Yea, I wanted to get this ready for when it lands again. The stabilization should not have been done without this being ready, but I let it fall through the cracks.

crlf0710:
> Do you think you'd be able to address the comments here?

Frankly, i don't feel i have enough skill of writing to address the comments here(especially as i'm not a English-speaker). I can deal with "add XXX after YYY"-style of comments well, but i can't deal with the "This should mention" and "worth stating explicitly" style of comments.

crlf0710:
Help wanted: It would be nice if someone could post a new PR addressing comments above and superseding this PR.

traviscross:
We discussed this in the lang-docs call today.  It seems this stabilization still needs to happen, though the way for it has been mostly cleared with this PR being merged:

- https://github.com/rust-lang/rust/pull/120248

..and the documentation here likely needs to be updated accordingly.

cc @WaffleLapkin

crlf0710:
Superseded by #1622 :)

PR #1260, opened 2022-09-04
Clarify that 0 is a valid multiple of a type's alignment

joshlf:
The reference currently states that "[t]he size of a value is always a multiple of its alignment." This might be taken to imply that a type's size is greater than or equal to its alignment. This commit clarifies that 0 is a valid multiple; zero-sized types have 0 size and nonzero alignment.


ehuss (approved):
Thanks!  Yea, it does seem like `0` is considered a factor. I've always felt this paragraph is a bit awkward in how it uses array layout to define *size*.  

I'm going to go ahead and merge even though I'm not 100% confident in the wording here. If someone wants to rework it to be clearer, feel free to follow up.

PR #1261, opened 2022-09-05
add section about implied bounds

lcnr:
cc https://github.com/rust-lang/rust/pull/100747/files#r962488178

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -158,7 +158,7 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
 
 ## Implied bounds
 
-Lifetime bounds required for types to be well-formed are sometimes inferred by the compiler.
+Lifetime bounds required for types to be well-formed are sometimes inferred.
 
 ```rust
 fn requires_t_outlives_a<'a, T>(x: &'a T) {}
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -156,6 +156,35 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
 }
 ```
 
+## Implied bounds
+
+Rust sometimes infers some bounds the user would have otherwise been required to write.
+
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {}
+```
+While this function requires `t` to outlive `'a`, this is inferred as the function signature
+contains the type `&'a T` which is only valid if `T: 'a` holds.
+
+Rust adds implied bounds for all inputs and outputs of functions. Inside of `requires_t_outlives_a`
+you can assume `T: 'a` to hold even if you don't explicitly specify this:
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {
+    requires_t_outlives_a_not_implied::<'a, T>();
+}
+
+fn requires_t_outlives_a_not_implied<'a, T: 'a>() {}
+```
+
+Only lifetime bounds are implied, trait bounds still have to be explicitly added.
+This behavior may change in the future however. The following example still causes an error:
+```rust,compile_fail
+use std::fmt::Debug;
+struct IsDebug<T: Debug>(T);
+// error[E0277]: `T` doesn't implement `Debug`
+fn doesnt_specify_t_debug<T>(x: IsDebug<T>) {}
+```
+
 [LIFETIME_OR_LABEL]: tokens.md#lifetimes-and-loop-labels
 [_GenericParams_]: items/generics.md
 [_TypePath_]: paths.md#paths-in-types
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -221,7 +221,7 @@ trait Trait<'a, T: 'a> {}
 // in the impl header.
 //     impl<'a, T> Trait<'a, T> for () {}
 
-// This compiles as `T: 'a` is implied by the self type `&'a ()`.
+// This compiles as `T: 'a` is implied by the self type `&'a T`.
 impl<'a, T> Trait<'a, T> for &'a T {}
 ```
 
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -158,40 +158,73 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
 
 ## Implied bounds
 
-Rust sometimes infers some bounds the user would have otherwise been required to write.
+Lifetime bounds required for types to be well-formed are sometimes inferred by the compiler.
 
 ```rust
 fn requires_t_outlives_a<'a, T>(x: &'a T) {}
 ```
-While this function requires `T` to outlive `'a`, this is inferred because the function signature
-contains the type `&'a T` which is only valid if `T: 'a` holds.
+The type parameter `T` is required to outlive `'a` for the type `&'a T` to be well-formed.
+This is inferred because the function signature contains the type `&'a T` which is
+only valid if `T: 'a` holds.
 
 Rust adds implied bounds for all inputs and outputs of functions. Inside of `requires_t_outlives_a`
 you can assume `T: 'a` to hold even if you don't explicitly specify this:
-```rust,compile_fail
+```rust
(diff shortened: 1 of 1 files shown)


src/trait-bounds.md
```diff
@@ -156,6 +156,35 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
 ```
 
+## Implied bounds
+
+Rust sometimes infers some bounds the user would have otherwise been required to write.
+
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {}
+```
+While this function requires `t` to outlive `'a`, this is inferred as the function signature
```
MatthewPeterKelly, line 166: Should `t` here be `T`?

MatthewPeterKelly: Also,  I think that `inferred because the function` is a bit easier to read than `inferred as the function`.

src/trait-bounds.md
```diff
@@ -156,6 +156,35 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+contains the type `&'a T` which is only valid if `T: 'a` holds.
+
+Rust adds implied bounds for all inputs and outputs of functions. Inside of `requires_t_outlives_a`
+you can assume `T: 'a` to hold even if you don't explicitly specify this:
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {
+    requires_t_outlives_a_not_implied::<'a, T>();
+}
+
+fn requires_t_outlives_a_not_implied<'a, T: 'a>() {}
```
MatthewPeterKelly, line 176: In this case, would it be correct to add a statement like:

"Note that we are require to add a lifetime bound on `T` in `requires_t_outlives_a_not_implied()` because it cannot be inferred from the argument (like it was in `requires_t_outlives_a()`)."

src/trait-bounds.md
```diff
@@ -156,6 +156,35 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+```
+
+Only lifetime bounds are implied, trait bounds still have to be explicitly added.
+This behavior may change in the future however. The following example still causes an error:
+```rust,compile_fail
+use std::fmt::Debug;
+struct IsDebug<T: Debug>(T);
+// error[E0277]: `T` doesn't implement `Debug`
+fn doesnt_specify_t_debug<T>(x: IsDebug<T>) {}
+```
```
MatthewPeterKelly, line 186: This provides helpful context - good idea to add it here.

MatthewPeterKelly (commented):
Thanks for adding this documentation to support https://github.com/rust-lang/rust/pull/100747. I'll plan to hold off merging that PR until this one lands (so that I can add a good reference).

MatthewPeterKelly (approved):
Looks good to me!

MatthewPeterKelly:
(I don't have write access to the repo, so I think you'll need someone else to rubber stamp the PR).

MatthewPeterKelly:
@Havvy - would you be able to do a quick review for this PR?

src/trait-bounds.md
```diff
@@ -156,6 +156,43 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+
+Rust sometimes infers some bounds the user would have otherwise been required to write.
+
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {}
+```
+While this function requires `T` to outlive `'a`, this is inferred because the function signature
+contains the type `&'a T` which is only valid if `T: 'a` holds.
+
+Rust adds implied bounds for all inputs and outputs of functions. Inside of `requires_t_outlives_a`
```
Havvy, line 169: The inputs and called parameters.

Havvy (changes requested):
Reading through this, I feel like the prose is unnecessarily vague. And in fact, it fails to answer two key questions:

1. What are all the places that lifetime bounds are inferred? As written, it looks like it is only for generics in functions. Are there others?
2. How are the bounds chosen by looking at the types of the function parameters and output? In the simplest case, as shown by the example, a top level reference to a generic type gives said generic type a lifetime bound of the lifetime of the reference. Are there more complicated cases? If so, what are they?

--

The examples are good.

src/trait-bounds.md
lcnr: how would you write this sentence then?

Havvy: Just replace "inputs" with "parameters"

src/trait-bounds.md
```diff
@@ -156,6 +156,76 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
 }
 ```
 
+## Implied bounds
+
+Lifetime bounds required for types to be well-formed are sometimes inferred by the compiler.
```
Havvy, line 161:
```suggestion
Lifetime bounds required for types to be well-formed are sometimes inferred.
```

src/trait-bounds.md
```diff
@@ -156,6 +156,76 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {}
+```
+The type parameter `T` is required to outlive `'a` for the type `&'a T` to be well-formed.
+This is inferred because the function signature contains the type `&'a T` which is
+only valid if `T: 'a` holds.
+
+Rust adds implied bounds for all inputs and outputs of functions. Inside of `requires_t_outlives_a`
```
Havvy, line 170: When does it add implied bounds? You have a lot of good examples, but the actual rules for how the bounds are inferred are missing. I don't actually know the rules, so I can't be helpful here. If you don't either, I guess we can merge this and open an issue or we can find somebody who does know the rules. I do imagine it would be something like "For each parameter type and output type, if the type contains a reference to a generic type `T` (declared by the function item?), then the bounds with a lifetime `'a`, then `T: 'a` is inferred."

Havvy (changes requested):

src/trait-bounds.md
lcnr: > I guess we can merge this and open an issue

let's do that then. I think explaining this would require the reference to have a section about well-formedness requirements and I don't have the time to write that one rn. Lets open an issue and i can try to look into that one at a later point.

src/trait-bounds.md
```diff
@@ -156,6 +156,76 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+Lifetime bounds required for types to be well-formed are sometimes inferred.
+
+```rust
+fn requires_t_outlives_a<'a, T>(x: &'a T) {}
+```
+The type parameter `T` is required to outlive `'a` for the type `&'a T` to be well-formed.
+This is inferred because the function signature contains the type `&'a T` which is
+only valid if `T: 'a` holds.
+
+Rust adds implied bounds for all parameters and outputs of functions. Inside of `requires_t_outlives_a`
```
Havvy, line 170: Still need to write this without referring to Rust directly. Everything is about Rust.

src/trait-bounds.md
```diff
@@ -156,6 +156,76 @@ fn call_on_ref_zero<F>(f: F) where F: for<'a> Fn(&'a i32) {
...
+```rust,compile_fail
+# fn requires_t_outlives_a_not_implied<'a, T: 'a>() {}
+fn not_implied<'a, T>() {
+    // This errors, because `T: 'a` is not implied by
+    // the function signature.
+    requires_t_outlives_a_not_implied::<'a, T>();
+}
+```
+
+Only lifetime bounds are implied, trait bounds still have to be explicitly added. The following example therefore causes an error:
```
Havvy, line 191: Missing newline between sentences.

Havvy (changes requested):
Couple nits.

lcnr:
done :+1:

lcnr:
:wave: do you think we can go ahead and merge this?

ehuss (approved):
Sure, I think this can go ahead and get merged. I'd agree with Havvy's comments that this seems a little informal, but I can understand how being more precise would take a ton more work.

PR #1262, opened 2022-09-05
Clarify reference on async blocks

matklad:
* fix "typo" when declaring semantics of `return` from within async block
* remove confusing false comment from the example


ehuss (approved):
Thanks!

PR #1263, opened 2022-09-06
Document `label_break_value` in the reference

jyn514:
r? @ehuss

### DIFF
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -7,6 +7,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_PredicateLoopExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_PredicatePatternLoopExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_IteratorLoopExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_LabelBlockExpression_]\
 > &nbsp;&nbsp; )
 
 [_LoopLabel_]: #loop-labels
@@ -14,16 +15,19 @@
 [_PredicateLoopExpression_]: #predicate-loops
 [_PredicatePatternLoopExpression_]: #predicate-pattern-loops
 [_IteratorLoopExpression_]: #iterator-loops
+[_LabelBlockExpression_]: #labelled-block-expressions
 
-Rust supports four loop expressions:
+Rust supports five loop expressions:
 
 *   A [`loop` expression](#infinite-loops) denotes an infinite loop.
 *   A [`while` expression](#predicate-loops) loops until a predicate is false.
 *   A [`while let` expression](#predicate-pattern-loops) tests a pattern.
 *   A [`for` expression](#iterator-loops) extracts values from an iterator, looping until the iterator is empty.
+*   A [labelled block expression](#labelled-block-expressions) runs a loop exactly once, but allows exiting the loop early with `break`.
 
-All four types of loop support [`break` expressions](#break-expressions), [`continue` expressions](#continue-expressions), and [labels](#loop-labels).
-Only `loop` supports [evaluation to non-trivial values](#break-and-loop-values).
+All five types of loop support [`break` expressions](#break-expressions), and [labels](#loop-labels).
+All except labelled block expressions support [`continue` expressions](#continue-expressions).
+Only `loop` and labelled block expressions support [evaluation to non-trivial values](#break-and-loop-values).
 
 ## Infinite loops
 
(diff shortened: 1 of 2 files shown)


nbdd0121:
Shouldn't this be called labelled block expression instead of labelled break expression?

ciphergoth:
Is this really a loop expression? It doesn't loop.

Havvy:
I take it that it is documented in loop expressions because it can only be used in loop expression context. Which I think is wrong. I don't see label-break-value as introducing a new kind of expression but rather modifies the semantics of just a block expression and break expressions. If I'm not too tired, I think I might create a competing pull request with how I think it should be documented; explaining it fully would be too much work.

Also, please oh please don't call this a `labeled expression` if we end up keeping this as its own expression kind. At least call it a `labeled block expression` or something.

Havvy:
I, uh, didn't write the PR today. I did think about perspectives some more, and now there's too many perspectives in my head. I wrote about them at https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/Perspectives.20on.20Break.20Label.20Value

Havvy:
I'm nominating this issue as per the question raised on Zulip.

src/expressions/loop-expr.md
```diff
@@ -226,6 +241,15 @@ Example:
 
 A `break` expression is only permitted in the body of a loop, and has one of the forms `break`, `break 'label` or ([see below](#break-and-loop-values)) `break EXPR` or `break 'label EXPR`.
 
+## Label expressions
+
+> **<sup>Syntax</sup>**\
+> _LabelBreakExpression_ :\
+> &nbsp;&nbsp; [LIFETIME_OR_LABEL] [_BlockExpression_]
```
ehuss, line 248: I might choose a different name for this, since I think "label break expression" is a little too similar _BreakExpression_. To me, it makes it sound like `'label break expr`. What do you think of renaming it to something like _LabelBlockExpression_?

Also I think it would be good to add this to the _LoopExpression_ definition at the top of the page and drop the `LIFETIME_OR_LABEL`. That way it can share the definition for _LoopLabel_ (and, btw, this is missing the `:`).

So, this might be something like:

```suggestion
> _LabelBlockExpression_ :\
> &nbsp;&nbsp; [_BlockExpression_]
```

src/expressions/loop-expr.md
```diff
@@ -226,6 +241,15 @@ Example:
 
 A `break` expression is only permitted in the body of a loop, and has one of the forms `break`, `break 'label` or ([see below](#break-and-loop-values)) `break EXPR` or `break 'label EXPR`.
 
+## Label expressions
```
ehuss, line 244: Similarly, I might make this a little clearer by calling them "Labeled block expressions".

ehuss (commented):
Thanks for posting this!

Posting a review now, though I know @Havvy was interested in exploring other options. Feel free to ignore until there is more consensus on how to approach this.

src/expressions/loop-expr.md
```diff
@@ -14,15 +14,18 @@
...
 *   A [`loop` expression](#infinite-loops) denotes an infinite loop.
 *   A [`while` expression](#predicate-loops) loops until a predicate is false.
 *   A [`while let` expression](#predicate-pattern-loops) tests a pattern.
 *   A [`for` expression](#iterator-loops) extracts values from an iterator, looping until the iterator is empty.
+*   A [labelled `break` expression](#label-expressions) runs a loop exactly once, but allows exiting the loop early.
 
-All four types of loop support [`break` expressions](#break-expressions), [`continue` expressions](#continue-expressions), and [labels](#loop-labels).
+All five types of loop support [`break` expressions](#break-expressions), and [labels](#loop-labels).
+All except labelled `break` expressions support [`continue` expressions](#continue-expressions). 
 Only `loop` supports [evaluation to non-trivial values](#break-and-loop-values).
```
Tyilo, line 29: Labelled break expressions also support "evaluation to non-trivial values".

nikomatsakis:
My two cents--

I think the approach in this PR seems *fine*, but I think we should have a link from wherever blocks are documented saying "labeled blocks are described under the loop section".

jyn514:
I tried to address @ehuss and @nikomatsakis's concerns. I renamed "label expression" to "label block expression".

pnkfelix:
@jyn514 It want to echo the [comments of others](https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/Perspectives.20on.20Break.20Label.20Value/near/298043558) that it *feels* really weird to document labelled blocks in a section named "Loop expressions"

Could we rename the "Loop expressions" chapter to something that doesn't indirectly imply that a labelled block somehow is a looping construct?

E.g. "Loop expressions and non-local jumps" or "Loop expressions and non-local control flow"? (Or even just "Complex control flow expressions"?)

nbdd0121:
Or "breakable expressions"?

pnkfelix:
I like "Loops and breakable expressions", that rolls of the tongue nicely.

jyn514:
I don't have objections to @pnkfelix's suggestion, but I don't know when I'll have time to apply them. Happy for someone else to push to my branch.

pnkfelix:
> Happy for someone else to push to my branch.

Done!

joshtriplett (approved):

nikomatsakis (approved):

PR #1264, opened 2022-09-06
Use actual `ne` method in derive reference

BD103:
I noticed the expanded derive example was a lot more complicated than [what actually happens](https://doc.rust-lang.org/stable/src/core/cmp.rs.html#219-234), so I made this change to clarify.

### DIFF
--- a/src/attributes/derive.md
+++ b/src/attributes/derive.md
@@ -26,7 +26,7 @@ impl<T: PartialEq> PartialEq for Foo<T> {
     }
 
     fn ne(&self, other: &Foo<T>) -> bool {
-        self.a != other.a || self.b != other.b
+        !self.eq(other)
     }
 }
 ```


memoryruins:
The code in that link is not the generated derive code. It's the provided method for `ne` when it's not overridden by implementations.

That said, the derive did change a few months ago and no longer derives `ne` <https://github.com/rust-lang/rust/pull/98655>. This can be checked the [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021) by clicking Tools > `Expand Macros`. If we want to match that here, then we can remove the function altogether.
```diff
- fn ne(&self, other: &Foo<T>) -> bool {
-     self.a != other.a || self.b != other.b
- }
```

BD103:
True, writing the `ne` implementation would just be superfluous. I'll remove it then.

ehuss:
@BD103 FYI, it looks like you pushed your change to your `master` branch instead of the `patch-1` branch, so it doesn't show up here.

BD103:
Whoops! My bad 😅

ehuss (approved):
Thanks!

PR #1265, opened 2022-09-09
Add basic GATs reference information

jackh726:
There really isn't a *ton* to add here. Almost all of the syntax related to GATs falls out from existing syntax.

### DIFF
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -224,7 +224,7 @@ The identifier is the name of the declared type alias. The optional trait bounds
 must be fulfilled by the implementations of the type alias.
 There is an implicit [`Sized`] bound on associated types that can be relaxed using the special `?Sized` bound.
 
-An *associated type definition* defines a type alias on for the implementation
+An *associated type definition* defines a type alias for the implementation
 of a trait on a type. They are written similarly to an *associated type declaration*,
 but cannot contain `Bounds`, but instead must contain a `Type`:
 
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -211,6 +211,7 @@ lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
 on the associated type, and `WhereBounds` is a comma-separated list of bounds on
 parameters:
 
+<!-- ignore: illustrative example forms -->
 ```rust,ignore
 type Assoc;
 type Assoc: Bounds;
@@ -228,11 +229,12 @@ An *associated type definition* defines a type alias for the implementation
 of a trait on a type. They are written similarly to an *associated type declaration*,
 but cannot contain `Bounds`, but instead must contain a `Type`:
 
+<!-- ignore: illustrative example forms -->
 ```rust,ignore
 type Assoc = Type;
 type Assoc<Params> = Type; // the type `Type` here may reference `Params`
-type Assoc<Params> where WhereBounds = Type;
 type Assoc<Params> = Type where WhereBounds;
+type Assoc<Params> where WhereBounds = Type; // deprecated, prefer the form above
 ```
 
 If a type `Item` has an associated type `Assoc` from a trait `Trait`, then
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -335,6 +335,68 @@ impl<T> Container for Vec<T> {
 }
 ```
 
+### Required where clauses on generic associated types
+
+Generic associated type declarations on traits currently may require a list of
+where clauses, dependent on functions in the trait and how the GAT is used.
+
+In a few words, these where clauses are required in order to maximize the allowed
+definitions of the associated type in impls. To do this, any clauses that *can be
+proven to hold* on functions (using the parameters of the function or trait)
+where a GAT appears as an input or output must also be written on the GAT itself.
+
+```rust
+trait LendingIterator {
+    type Item<'x> where Self: 'x;
+    fn next<'a>(&'a mut self) -> Self::Item<'a>;
+}
+```
+
+In the above, on the `next` function, we can prove that `Self: 'a`, because of
+the implied bounds from `&'a mut self`; therefore, we must write the equivalent
+bound on the GAT itself: `where Self: 'x`.
+
+When there are multiple functions in a trait that use the GAT, then the
+*intersection* of the bounds from the different functions are used, rather than
+the union.
+
+```rust
+trait Check<T> {
+    type Checker<'x>;
+    fn create_checker<'a>(item: &'a T) -> Self::Checker<'a>;
+    fn do_check(checker: Self::Checker<'a>);
+}
+```
+
+In this example, no bounds are required on the `type Checker<'a>;`. While we
+know that `T: 'a` on `create_checker`, we do not know that on `do_check`. However,
+if `do_check` was commented out, then the `where T: 'x` bound would be required
+on `Checker`.
+
+The bounds on associated types also propagate required where clauses.
+
+```rust
+trait Iterable {
+    type Item<'a> where Self: 'a;
+    type Iterator<'a>: Iterator<Item = Self::Item<'a>> where Self: 'a;
+    fn iter<'a>(&'a self) -> Self::Iterator<'a>;
+}
+```
+
+Here, `where Self: 'a` is required on `Item` because of `iter`. However, `Item`
+is used in the bounds of `Iterator`, the `where Self: 'a` clause is also required
+there.
+
+Finally, any explicit uses of `'static` on GATs in the trait do not count towards
+the required bounds.
+
+```rust
+trait StaticReturn {
+    type Y<'a>;
+    fn foo(&self) -> Self::Y<'static>;
+}
+```
+
 ## Associated Constants
 
 *Associated constants* are [constants] associated with a type.
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -364,7 +364,7 @@ the union.
 trait Check<T> {
     type Checker<'x>;
     fn create_checker<'a>(item: &'a T) -> Self::Checker<'a>;
-    fn do_check(checker: Self::Checker<'a>);
+    fn do_check(checker: Self::Checker<'_>);
 }
 ```
 
--- a/src/items/associated-items.md
+++ b/src/items/associated-items.md
@@ -205,26 +205,46 @@ types cannot be defined in [inherent implementations] nor can they be given a
 default implementation in traits.
 
 An *associated type declaration* declares a signature for associated type
-definitions. It is written as `type`, then an [identifier], and
-finally an optional list of trait bounds.
+definitions. It is written in one of the following forms, where `Assoc` is the
+name of the associated type, `Params` is a comma-separated list of type,
+lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
+on the associated type, and `WhereBounds` is a comma-separated list of bounds on
+parameters:
+
+```rust,ignore
+type Assoc;
+type Assoc: Bounds;
+type Assoc<Params>;
+type Assoc<Params>: Bounds;
+type Assoc<Params> where WhereBounds;
+type Assoc<Params>: Bounds where WhereBounds;
+```
 
 The identifier is the name of the declared type alias. The optional trait bounds
 must be fulfilled by the implementations of the type alias.
 There is an implicit [`Sized`] bound on associated types that can be relaxed using the special `?Sized` bound.
 
-An *associated type definition* defines a type alias on another type. It is
-written as `type`, then an [identifier], then an `=`, and finally a [type].
+An *associated type definition* defines a type alias on for the implementation
+of a trait on a type. They are written similarly to an *associated type declaration*,
+but cannot contain `Bounds`, but instead must contain a `Type`:
+
+```rust,ignore
+type Assoc = Type;
+type Assoc<Params> = Type<Params>;
+type Assoc<Params> where WhereBounds = Type;
+type Assoc<Params> = Type where WhereBounds;
+```
 
 If a type `Item` has an associated type `Assoc` from a trait `Trait`, then
 `<Item as Trait>::Assoc` is a type that is an alias of the type specified in the
 associated type definition. Furthermore, if `Item` is a type parameter, then
 `Item::Assoc` can be used in type parameters.
 
-Associated types may include [generic parameters] or [where clauses]; these may
-be referred to as generic associated types, or GATs. If the type `Thing` has an
-associated type `Item` from a trait `Trait` with the generics `<'a>` , the type
-can be named like `<Thing as Trait>::Item<'x>`, where `'x` is some lifetime in
-scope. In this case, `'x` will be used wherever `'a` appears in the associated
+Associated types may include [generic parameters] and [where clauses]; these are
+often referred to as *generic associated types*, or *GATs*. If the type `Thing`
+has an associated type `Item` from a trait `Trait` with the generics `<'a>` , the
+type can be named like `<Thing as Trait>::Item<'x>`, where `'x` is some lifetime
+in scope. In this case, `'x` will be used wherever `'a` appears in the associated
 type definitions on impls.
 
 ```rust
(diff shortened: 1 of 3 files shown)


src/items/associated-items.md
```diff
@@ -220,7 +220,12 @@ If a type `Item` has an associated type `Assoc` from a trait `Trait`, then
 associated type definition. Furthermore, if `Item` is a type parameter, then
 `Item::Assoc` can be used in type parameters.
 
-Associated types must not include [generic parameters] or [where clauses].
+Associated types may include [generic parameters] or [where clauses]; these may
+be referred to generic associated types, or GATs. If the type `Thing` has an
```
jackh726, line 224:
```suggestion
be referred to as generic associated types, or GATs. If the type `Thing` has an
```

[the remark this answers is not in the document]
jackh726: Thanks! Good catch.

src/items/associated-items.md
```diff
@@ -220,7 +220,12 @@ If a type `Item` has an associated type `Assoc` from a trait `Trait`, then
 associated type definition. Furthermore, if `Item` is a type parameter, then
 `Item::Assoc` can be used in type parameters.
 
-Associated types must not include [generic parameters] or [where clauses].
+Associated types may include [generic parameters] or [where clauses]; these may
+be referred to as generic associated types, or GATs. If the type `Thing` has an
```
Havvy, line 224:
```suggestion
be referred to as *generic associated types*, or *GATs*. If the type `Thing` has an
```

I also think "may be referred to" should just say "these are called".

src/items/associated-items.md
```diff
@@ -220,7 +220,12 @@ If a type `Item` has an associated type `Assoc` from a trait `Trait`, then
 associated type definition. Furthermore, if `Item` is a type parameter, then
 `Item::Assoc` can be used in type parameters.
 
-Associated types must not include [generic parameters] or [where clauses].
+Associated types may include [generic parameters] or [where clauses]; these may
```
dhardy, line 223: 'and', not 'or'

src/items/type-aliases.md
```diff
@@ -4,7 +4,7 @@
 > _TypeAlias_ :\
 > &nbsp;&nbsp; `type` [IDENTIFIER]&nbsp;[_GenericParams_]<sup>?</sup>
 >              ( `:` [_TypeParamBounds_] )<sup>?</sup>
->              [_WhereClause_]<sup>?</sup> ( `=` [_Type_] )<sup>?</sup> `;`
+>              [_WhereClause_]<sup>?</sup> ( `=` [_Type_] )<sup>?</sup> [_WhereClause_]<sup>?</sup> `;`
```
dhardy, line 7: Why include `_WhereClause_` twice? The first placement is deprecated.

dhardy (commented):
Also in `src/items/associated-items.md:207`:

This paragraph gives a wordy description of syntax for associated type declarations. This isn't very easy to follow and isn't updated regarding GATs. I suggest something like the following:

> An *associated type declaration* declares a signature for associated type definitions. It is written in one of the following forms, where `Assoc` is the name of the associated type, `Params` is a comma-separated list of type, lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds on the associated type, and `WhereBounds` is a comma-separated list of bounds on parameters:
>
> ```
> type Assoc;
> type Assoc: Bounds;
> type Assoc<Params>;
> type Assoc<Params>: Bounds;
> type Assoc<Params> where WhereBounds;
> type Assoc<Params>: Bounds where WhereBounds;
> ```

Similarly for assoc. type definitions, para from line 215.

BurntSushi (commented):
Does anything about `where` clauses like, e.g., `Self: 'a`, need to be added here? I'm thinking about this specifically: https://rust-lang.github.io/generic-associated-types-initiative/explainer/required_bounds.html

src/items/type-aliases.md
```diff
@@ -37,6 +37,9 @@ A type alias without the [_Type_] specification may only appear as an
 A type alias with [_TypeParamBounds_] may only specified when used as
 an [associated type] in a [trait].
 
+A type alias with where clauses after the equals sign may only appear as an
+[associated type] in a [trait] or a [trait impl].
```
ehuss, line 41: Hm, so I may be a little confused here. IIUC, `associated_type_defaults` is not being stabilized? If so, this seems to imply that `=` can be used in a trait definition. Due to #24 (type defaults are not documented at all), it makes it a bit awkward to talk about this. 

If I am understanding this correctly, I'm not sure how to handle it. Perhaps add a sentence that type alias defaults are not allowed in trait definitions?

src/items/type-aliases.md
ehuss: Since it is supported in both positions, it needs to be included in both places.

However, this grammar does seem incorrect to me.  It seems to allow `type Assoc<'a, T> where T: 'a where T 'a;`.  I think the second where clause needs to be inside the parentheses for the `=`.

It might be good to have a sentence somewhere on this page that discusses the deprecated form. That might be a bit subtle to express clearly, since there are different forms, such as with and without `=`, and bounds on (non-associated) type aliases are ignored (`type_alias_bounds` warning).

ehuss (commented):
Thanks for posting this!

Unfortunately I don't know enough to even get started on what to cover.  I see things like the [Object safety](https://github.com/rust-lang/reference/blob/master/src/items/traits.md#object-safety) list needs to be updated?

src/items/type-aliases.md
jackh726: Yes, it's *soft* deprecated, but it's still valid syntax. I think for now it makes sense to keep it.

jackh726:
> Does anything about `where` clauses like, e.g., `Self: 'a`, need to be added here? I'm thinking about this specifically: https://rust-lang.github.io/generic-associated-types-initiative/explainer/required_bounds.html

I can try to add a short explanation

src/items/type-aliases.md
jackh726: Yeah, oops. This is unnecessary.

src/items/type-aliases.md
jackh726: I guess not "soft" deprecated, since we warn. Added a sentence.

src/items/type-aliases.md
```diff
@@ -34,11 +34,15 @@ let _ = TypeAlias(5); // Doesn't work
...
 
+A type alias with the [_Type_] specification may only appear as an
+[associated type] in a [trait impl].
+
 A type alias with [_TypeParamBounds_] may only specified when used as
 an [associated type] in a [trait].
 
-A type alias with where clauses after the equals sign may only appear as an
-[associated type] in a [trait] or a [trait impl].
+Where clauses before the equals sign on a type alias in a [trait impl] (like
```
Havvy, line 43: You have the same example of code in both parenthesis. Also, there's no reason to use parentheis; just use commas.

jackh726: I fixed the second example. Parentheses make more sense to me, since they are somewhat of an aside and aren't needed to understand the sentence, only add additional clarification.

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
 
-An *associated type definition* defines a type alias on another type. It is
-written as `type`, then an [identifier], then an `=`, and finally a [type].
+An *associated type definition* defines a type alias on for the implementation
+of a trait on a type. They are written similarly to an *associated type declaration*,
+but cannot contain `Bounds`, but instead must contain a `Type`:
+
+```rust,ignore
+type Assoc = Type;
+type Assoc<Params> = Type<Params>;
```
nikomatsakis, line 233:
```suggestion
type Assoc<Params> = Type;
```

This is inconsistent with the other examples, no?

src/items/associated-items.md
```diff
@@ -279,6 +335,68 @@ impl<T> Container for Vec<T> {
 }
 ```
 
+### Required where clauses on generic associated types
+
+Generic associated type declarations on traits currently may require a list of
+where clauses, dependent on functions in the trait and how the GAT is used.
```
nikomatsakis, line 341: I feel we should indicate somewhere that these rules are "in flux". Perhaps we could link to [the page on the generic-associated-types initiative repository](https://rust-lang.github.io/generic-associated-types-initiative/explainer/required_bounds.html)? Alternatively, isn't there an issue that we redirect people to?

nikomatsakis (changes requested):
This looks good. I have a few minor suggestions, @jackh726.

src/items/associated-items.md
jackh726: Yeah, I specifically wanted to highlight that you can use those `Params` in the defined type. But maybe it's just too confusing

src/items/associated-items.md
jackh726: good idea

src/items/associated-items.md
nikomatsakis: Yeah, I can see that. I could also see people interpreting it to mean that the type *has* to have that form (e.g., `Vec<T>` or something).

nikomatsakis: Could do 

```rust
type Assoc<Params> = Type; // the type `Type` here may reference `Params`
```

if you want to make it extra clear.

nikomatsakis (approved):

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
+type Assoc<Params>: Bounds where WhereBounds;
+```
 
 The identifier is the name of the declared type alias. The optional trait bounds
 must be fulfilled by the implementations of the type alias.
 There is an implicit [`Sized`] bound on associated types that can be relaxed using the special `?Sized` bound.
 
-An *associated type definition* defines a type alias on another type. It is
-written as `type`, then an [identifier], then an `=`, and finally a [type].
+An *associated type definition* defines a type alias on for the implementation
```
simonbuchan, line 227: Small copy note: "on for" isn't grammatical here, probably just meant to pick one.

nikomatsakis, line 227:
```suggestion
An *associated type definition* defines a type alias for the implementation
```

nikomatsakis: good catch

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
-written as `type`, then an [identifier], then an `=`, and finally a [type].
+An *associated type definition* defines a type alias for the implementation
+of a trait on a type. They are written similarly to an *associated type declaration*,
+but cannot contain `Bounds`, but instead must contain a `Type`:
+
+```rust,ignore
+type Assoc = Type;
+type Assoc<Params> = Type; // the type `Type` here may reference `Params`
+type Assoc<Params> where WhereBounds = Type;
+type Assoc<Params> = Type where WhereBounds;
```
nikomatsakis, line 235:
```suggestion
type Assoc<Params> = Type where WhereBounds;
type Assoc<Params> where WhereBounds = Type; // deprecated, prefer the form above
```

Nit, but it makes sense to let people know that this where clause placement is not the one to use.

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
+name of the associated type, `Params` is a comma-separated list of type,
+lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
+on the associated type, and `WhereBounds` is a comma-separated list of bounds on
+parameters:
+
+```rust,ignore
+type Assoc;
+type Assoc: Bounds;
+type Assoc<Params>;
+type Assoc<Params>: Bounds;
```
nikomatsakis, line 218: It occurs to me that it would be worth clarifying the distinction between `Bounds` and `WhereBounds`!

nikomatsakis (changes requested):

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
 An *associated type declaration* declares a signature for associated type
-definitions. It is written as `type`, then an [identifier], and
-finally an optional list of trait bounds.
+definitions. It is written in one of the following forms, where `Assoc` is the
+name of the associated type, `Params` is a comma-separated list of type,
+lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
+on the associated type, and `WhereBounds` is a comma-separated list of bounds on
+parameters:
+
+```rust,ignore
```
ehuss, line 214: Can you include a comment explaining why it is ignored? We try to avoid `ignore` as much as possible.

```suggestion
<!-- ignore: illustrative example forms -->
```rust,ignore
```

src/items/associated-items.md
```diff
@@ -205,22 +205,47 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
 must be fulfilled by the implementations of the type alias.
 There is an implicit [`Sized`] bound on associated types that can be relaxed using the special `?Sized` bound.
 
-An *associated type definition* defines a type alias on another type. It is
-written as `type`, then an [identifier], then an `=`, and finally a [type].
+An *associated type definition* defines a type alias for the implementation
+of a trait on a type. They are written similarly to an *associated type declaration*,
+but cannot contain `Bounds`, but instead must contain a `Type`:
+
+```rust,ignore
```
ehuss, line 231:
```suggestion
<!-- ignore: illustrative example forms -->
```rust,ignore
```

src/items/type-aliases.md
```diff
@@ -34,9 +34,16 @@ let _ = TypeAlias(5); // Doesn't work
 A type alias without the [_Type_] specification may only appear as an
 [associated type] in a [trait].
 
+A type alias with the [_Type_] specification may only appear as an
+[associated type] in a [trait impl].
```
ehuss, line 38: This clause by itself doesn't seem to be correct to me.  It seems to imply that the following is not allowed:

```rust
type X = i32;
```

Perhaps this could be turned around so that it specifically says what a trait impl associated type needs? A sentence could be added to the previous paragraph, since they are talking about the same thing.  Maybe something like this:

A type alias without the [_Type_] specification may only appear as an [associated type] in a [trait]. 
Associated types in a [trait impl] must include the [_Type_] specification.

src/items/associated-items.md
jackh726: That is stated in the paragraph above? Should we be more detailed?

src/items/type-aliases.md
jackh726: I've made a separate paragraph for each for of type aliases, to be specific about what is allow, what is required, and what isn't allowed in each.

src/items/associated-items.md
```diff
@@ -205,22 +205,49 @@ types cannot be defined in [inherent implementations] nor can they be given a
 default implementation in traits.
 
 An *associated type declaration* declares a signature for associated type
-definitions. It is written as `type`, then an [identifier], and
-finally an optional list of trait bounds.
+definitions. It is written in one of the following forms, where `Assoc` is the
+name of the associated type, `Params` is a comma-separated list of type,
+lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
+on the associated type, and `WhereBounds` is a comma-separated list of bounds on
+parameters:
```
nikomatsakis, line 212:
```suggestion
lifetime or const parameters, `Bounds` is a plus-separated list of trait bounds
that the associated type must meet, and `WhereBounds` is a comma-separated list of bounds that the parameters must meet:
```

src/items/associated-items.md
```diff
@@ -205,22 +205,49 @@ types cannot be defined in [inherent implementations] nor can they be given a
...
+
+<!-- ignore: illustrative example forms -->
+```rust,ignore
+type Assoc;
+type Assoc: Bounds;
+type Assoc<Params>;
+type Assoc<Params>: Bounds;
+type Assoc<Params> where WhereBounds;
+type Assoc<Params>: Bounds where WhereBounds;
+```
```
nikomatsakis, line 222: Maybe something like

### Relationship between `Bounds` and `WhereBounds`

In this example:

```rust
trait Example {
    type Output<T>: Ord where T: Debug
}
```

Given a reference to the associated type like `<X as Example>::Output<Y>`, the associated type itself must be `Ord`, and the type `Y` must be `Debug`.

src/items/associated-items.md
nikomatsakis: I think we could be more explicit. Left some suggestions. WDYT?

nikomatsakis (approved):

PR #1268, opened 2022-09-11
Classify AsyncBlockExpression as ExpressionWithoutBlock

### LINKED ISSUE #1267: AsyncBlockExpression is misclassified as an ExpressionWithBlock
According to [section 8.2 *Expressions*](https://doc.rust-lang.org/nightly/reference/expressions.html), *AsyncBlockExpression*s are classified as *ExpressionWithBlock* while `rustc` parses them as if they were an *ExpressionWithoutBlock*.

If *AsyncBlockExpression* was an *ExpressionWithBlock*, the following code **would be syntactically malformed**. In reality however, `rustc` successfully parses it!

```rust
fn main() {
    #[cfg(FALSE)]
    match () { () => async { () }[..] }
}
```

You can experiment with the code above by replacing `async` with `unsafe` / `loop` (true *ExpressionWithBlock*s) or an empty string (making it a *BlockExpression* which is also a true *ExpressionWithBlock*) and observing that the compiler rightfully reports a syntax error.

Explanation of the test snippet:
For an *ExpressionWithBlock*, the block ends the expression and the comma at the end of a match arm is optional. Meaning if there isn't one, we immediately expect a pattern (in this case we find `[..]` – a slice pattern) followed by `=>` (here we find `}` and emit a syntax error). For an *ExpressionWithoutBlock* on the other hand, we continue to parse an expression after the end of the block, interpret the `[..]` as being part of an *IndexExpression* and succeed.

The reference and the (reference) compiler are in disagreement here. I am not sure who is right. I couldn't find any mention of this in [RFC 2394](https://github.com/rust-lang/rfcs/pull/2394). Should I instead / additionally create an issue at the rust-lang/rust repo?

This issue was first [discovered on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/const.20blocks.20differ.20from.20normal.20and.20from.20unsafe.20blocks/near/290232262) when investigating the same issue for the experimental [inline consts](https://github.com/rust-lang/rfcs/pull/2920) (*ConstBlockExpression*s). For the latter, the issue was acknowledged but no straightforward answer could be found and a note was added to its [tracking issue](https://github.com/rust-lang/rust/issues/76001).

@rustbot label Bug A-async A-grammar
<sup>(hmm, labeling does not work despite the existence of `triagebot.toml` in the root folder and the rule set `relabel.allow-unauthenticated` within it)</sup>

fmease:
If rustc's behavior regarding async blocks is intentional, then this fixes #1267.

We might want to consult with the Async WG on this first. Edit: [Zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/187312-wg-async/topic/syntax.20of.20async.20blocks/near/298263840).

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -22,6 +22,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_MethodCallExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_FieldExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ClosureExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_AsyncBlockExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ContinueExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_BreakExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_RangeExpression_]\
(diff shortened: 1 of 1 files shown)


fmease:
Now that #1267 is confirmed to be a bug by **eholk**,

r? @ehuss

PR #1270, opened 2022-09-16
Add `sym` operands for inline assembly

Amanieu:
Tracking issue: https://github.com/rust-lang/rust/issues/93333


Amanieu:
This has now been stabilized.

ehuss (approved):
Thanks!

PR #1272, opened 2022-09-24
Use semver-compliant example version

Smittyvb:
Clippy warns if the `since` version in `#[deprecated]` isn't semver-compliant. This PR updates an example using `#[deprecated]` to make the version semver-compliant.


ehuss (approved):
Thanks!

PR #1273, opened 2022-09-25
mention the extra const UB

RalfJung:
Cc @rust-lang/wg-const-eval @rust-lang/wg-unsafe-code-guidelines  -- I think that is the only extra const UB but let me know if I missed anything. :)

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -60,6 +60,11 @@ code.
     > `rustc_layout_scalar_valid_range_*` attributes.
 * Incorrect use of inline assembly. For more details, refer to the [rules] to
   follow when writing code that uses inline assembly.
+* **In [const context](const_eval.md#const-context)**: transmuting or otherwise
+  reinterpreting a pointer (reference, raw pointer, or function pointer) into
+  some allocated object as a non-pointer type (such as integers).
+  'Reinterpreting' refers to loading the pointer value at integer type without a
+  cast, e.g. by doing raw pointer casts or using a union.
 
 **Note:** Uninitialized memory is also implicitly invalid for any type that has
 a restricted set of valid values. In other words, the only cases in which


src/behavior-considered-undefined.md
```diff
@@ -60,6 +60,10 @@ code.
     > `rustc_layout_scalar_valid_range_*` attributes.
 * Incorrect use of inline assembly. For more details, refer to the [rules] to
   follow when writing code that uses inline assembly.
+* **In `const` context** (i.e., during the evaluation of a `const`/`static`
+  initializer, array length, enum discriminant, or const generic value):
+  transmuting or otherwise reinterpreting a pointer into some allocated object
+  as an integer or other non-pointer type.
```
Havvy, line 66: Pointer type should ~~probably~~ maybe link to https://doc.rust-lang.org/reference/types/pointer.html

I say maybe because are those three types (`&T`, `&mut T`, `*T`) the only pointer types, or are function pointers also included? If so, we should probably reorganize the types chapter so that all pointer types as per this definition are in the `types/pointer.html` chapter.

Havvy (dismissed):
This is an editorial review. I have not verified the technical content.

src/behavior-considered-undefined.md
RalfJung: For the purpose of CTFE, function pointers are definitely pointers.

But for the semantics in general this is less clear so I would hold off on changing the overall organization.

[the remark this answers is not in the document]
RalfJung: Note that this is about transmutation/reinterpretation, not casts. Casts are simply disallowed in const context so they cannot cause UB.

RalfJung:
@Havvy @ehuss what is needed to move this PR forwards?

ehuss:
Sorry for the delay. This looks good to me, but I don't actually know the specifics of what is UB, or how we should decide that. I trust that you are correct, but it might help to get one other person who is familiar with it to approve it.

Perhaps @oli-obk or someone on wg-unsafe-code-guidelines, can you find someone to take a look?

RalfJung:
@rust-lang/wg-const-eval @rust-lang/opsem please have a look at this PR. :)

digama0:
Is there a reason this needs to be UB and not simply an error? A more advanced implementation of const-eval could potentially handle this in limited situations.

CAD97:
I believe this is an accurate description of the current state of affairs.

The one thing to pay attention to is that this implicitly confirms that ptr-int transmutes are permitted outside of const contexts (by stating it's UB specifically in const context). I believe we've decided this is both fine and desired (semantics: strip provenance when loading as integer (`ptr::addr`)), loading a pointer without provenance loads an integer then `as` casts to pointer (`ptr::from_exposed_addr`)), but it's probably still a good idea to have an FCP for that.

> Is there a reason this needs to be UB and not simply an error?

A not insignificant reason is that we don't really have a way to say / precedent for an operation which is an error at const time but just permitted at runtime. The closest would I guess be panics, but those errors would be better understood as being an unwind, and were `catch_unwind` to be const, it should probably catch const panics as well.

There's also the fact that it decidedly _was_ UB previously (exposing nondeterminism into const) before it was turned into an error.

The [original unsafe in const RFC](https://github.com/rust-lang/rust/issues/55607) covers more motivation on why we chose to treat const UB the way we do. (Not the least being continuing confusion between lib and lang UB.) Rustc goes to decent lengths to diagnose unambiguous lang UB in const, and we should expect it and any alternative frontends to continue to do so, but from the opsem perspective const evaluation errors are basically UB by definition (they're raised by operations for which the evaluator defines no semantics).

(IIRC: const UB is permitted to fail compilation and/or compromise the output artifact arbitrarily, but must not result in the compiler doing anything it couldn't otherwise do. Rustc provides an additional QOI guarantee that when no error is produced, compiler behavior is further bound to that reachable by selecting arbitrary byte-valid results for the const context (e.g. won't cause a linked but never mentioned proc macro to be run).)

digama0:
Right, I am aware that our const UB semantics are both much more constrained than regular UB, but also quite nondeterministic, which is why I think if at all possible we should opt for something more well defined than that.

> A not insignificant reason is that we don't really have a way to say / precedent for an operation which is an error at const time but just permitted at runtime.

You could say likewise that we don't have a precedent for an operation which is UB at const time and permitted at runtime. If the RFC https://github.com/rust-lang/rfcs/pull/3352 regarding const time / runtime divergence goes through (which I expect to happen), then having an operation that is an error at const time and works at runtime seems quite reasonable and tame to work with, while an operation that is UB at const time (and is allowed to corrupt any constant it is used in) couldn't really be part of a library at all.

RalfJung:
The main reason is that there is no such thing as a const eval error. Also the current implementation doesn't error at transmute time, it errors later or not at all depending on what is done with the result. Detecting illegal transmutes is non-trivial as they can occur in many syntactic positions (e.g. any time a load from a raw pointer or reference occurs). So the status quo really is const UB, and it is unclear whether we can do qualitatively better in the given performance budget.

> There's also the fact that it decidedly was UB previously (exposing nondeterminism into const) before it was turned into an error.

We never exposed non-det to const. The implementation has always been very much like what it is now, just less documented.

RalfJung:
Note that this UB went through t-lang FCP [here](https://github.com/rust-lang/rust/pull/85769#issuecomment-873168242), 2 years ago. We just forgot to add it to the reference. So I think that should be sufficient justification to merge as-is.

@digama0 if you want to change this to a const-eval error and have a good idea for how to realize this, I'm not opposed, but I don't think that should block updating the reference to a consensus that was formed 2 years ago.

digama0:
I'll retract my concern.

lcnr (approved):

PR #1278, opened 2022-09-29
update 'unsafe'

RalfJung:
I started equipping this with some examples for the more subtle situations (in particular `fn` in an `unsafe trait` vs an `unsafe fn` in a safe `trait`), but then I realized we have https://doc.rust-lang.org/nightly/std/keyword.unsafe.html which looks like a better place for the examples. So now https://github.com/rust-lang/rust/pull/102475 adds the examples and this here just updates the reference documentation and adds the missing parts (`unsafe trait` and `unsafe impl`).

### DIFF
--- a/src/unsafe-keyword.md
+++ b/src/unsafe-keyword.md
@@ -1,4 +1,12 @@
-# Unsafe blocks
+# The `unsafe` keyword
+
+## Unsafe functions
+
+Unsafe functions are functions that are not safe in all contexts and/or for all
+possible inputs. Such a function must be prefixed with the keyword `unsafe` and
+can only be called from an `unsafe` block or another `unsafe` function.
+
+## Unsafe blocks
 
 A block of code can be prefixed with the `unsafe` keyword, to permit calling
 `unsafe` functions or dereferencing raw pointers within a safe function.
--- a/src/unsafe-keyword.md
+++ b/src/unsafe-keyword.md
@@ -4,27 +4,25 @@ The `unsafe` keyword can occur in several different contexts:
 unsafe functions (`unsafe fn`), unsafe blocks (`unsafe {}`), unsafe traits (`unsafe trait`), and unsafe trait implementations (`unsafe impl`).
 It plays several different roles, depending on where it is used and whether the `unsafe_op_in_unsafe_fn` lint is enabled:
 - it is used to mark code that *defines* extra safety conditions (`unsafe fn`, `unsafe trait`)
-- it is used to mark code that needs to *satisfy* extra safety conditions (`unsafe {}`, `unsafe impl`, `unsafe fn` without `unsafe_op_in_unsafe_fn`)
+- it is used to mark code that needs to *satisfy* extra safety conditions (`unsafe {}`, `unsafe impl`, `unsafe fn` without [`unsafe_op_in_unsafe_fn`])
 
 The following discusses each of these cases.
 See the [keyword documentation][keyword] for some illustrative examples.
-
-[keyword]: ../std/keyword.unsafe.html
 
 ## Unsafe functions (`unsafe fn`)
 
 Unsafe functions are functions that are not safe in all contexts and/or for all possible inputs.
 We say they have *extra safety conditions*, which are requirements that must be upheld by all callers and that the compiler does not check.
-For example, `get_unchecked` has the extra safety condition that the index must be in-bounds.
-The module defining an unsafe function is responsible for documenting what those extra safety conditions are.
+For example, [`get_unchecked`] has the extra safety condition that the index must be in-bounds.
+The unsafe function should come with documentation explaining what those extra safety conditions are.
 
-Such a function must be prefixed with the keyword `unsafe` and can only be called from inside an `unsafe` block.
+Such a function must be prefixed with the keyword `unsafe` and can only be called from inside an `unsafe` block, or inside `unsafe fn` without the [`unsafe_op_in_unsafe_fn`] lint.
 
 ## Unsafe blocks (`unsafe {}`)
 
 A block of code can be prefixed with the `unsafe` keyword, to permit calling `unsafe` functions or dereferencing raw pointers.
 By default, the body of an unsafe function is also considered to be an unsafe block;
-this can be changed by enabling the `unsafe_op_in_unsafe_fn` lint.
+this can be changed by enabling the [`unsafe_op_in_unsafe_fn`] lint.
 
 By putting operations into an unsafe block, the programmer states that they have taken care of satisfying the extra safety conditions of all operations inside that block.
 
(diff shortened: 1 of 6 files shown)


src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
+# The `unsafe` keyword
+
+The `unsafe` keyword in Rust can occur in several different context: unsafe
```
ehuss, line 3: We try to avoid "in Rust" since everything in the reference is about Rust. I also think this seems like it should be plural.

```suggestion
The `unsafe` keyword can occur in several different contexts: unsafe
```

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+[keyword documentation][keyword] for some illustrative examples.
+
+[keyword]: ../std/keyword.unsafe.html
+
+## Unsafe functions (`unsafe fn`)
+
+Unsafe functions are functions that are not safe in all contexts and/or for all
+possible inputs. We say they have *extra safety conditions*, which are
+requirements that must be upheld by all callers and that the compiler does not
+check. For example, `get_unchecked` has the extra safety
```
ehuss, line 23: I think it would be good to link this for context.

```suggestion
check. For example, [`get_unchecked`] has the extra safety
```

```
[`get_unchecked`]: ../std/primitive.slice.html#method.get_unchecked
```

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+condition that the index must be in-bounds. The module defining an unsafe
+function is responsible for documenting what those extra safety conditions are.
```
ehuss, line 25: This seems rather specific to say it must go in the *module* documentation. For example, the aforementioned `get_unchecked` is just documented on the method itself, not in the module docs. Can this be reworded to be a little more general?

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+
+Such a function must be prefixed with the keyword `unsafe` and can only be
+called from inside an `unsafe` block.
```
ehuss, line 28:
```suggestion
called from inside an `unsafe` function or block.
```

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
+# The `unsafe` keyword
+
+The `unsafe` keyword in Rust can occur in several different context: unsafe
+functions (`unsafe fn`), unsafe blocks (`unsafe {}`), unsafe traits (`unsafe
+trait`), and unsafe trait implementations (`unsafe impl`). It plays several
+different roles, depending on where it is used and whether the
+`unsafe_op_in_unsafe_fn` lint is enabled:
```
ehuss, line 7: I'm always on the fence on referring to rustc's lints in the reference since they aren't exactly part of the language. However, in this case it seems very relevant here. I think we'll need to figure out a better policy in the future.

It might help to link this for someone to learn more about the lint:

```suggestion
[`unsafe_op_in_unsafe_fn`] lint is enabled:
```

```
[`unsafe_op_in_unsafe_fn`]: ../rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn
```

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+For example, a doubly-linked list is not a tree structure and can only be
+represented with reference-counted pointers in safe code. By using `unsafe`
+blocks to represent the reverse links as raw pointers, it can be implemented
+with only boxes.
+
+## Unsafe traits (`unsafe trait`)
+
+An unsafe trait is a trait that comes with extra safety conditions that must be
+upheld by *implementations* of the trait. The module defining an unsafe trait is
+responsible for documenting what those extra safety conditions are.
```
ehuss, line 66: Ditto about being specific about the docs being in the module.

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+Such a trait must be prefixed with the keyword `unsafe` and can only be
+implemented by `unsafe impl` blocks.
+
+## Unsafe trait implementations (`unsafe impl`)
+
+When implementing an unsafe trait, the implementation needs to be prefixed with
+the `unsafe` keyword. By writing `unsafe impl`, the programmer states that they
+have taken care of satisfying the extra safety conditions required by the trait.
+
+Unsafe traits are the logical dual to unsafe traits: where unsafe traits define
```
ehuss, line 77: ???

```suggestion
Unsafe traits are the logical dual to unsafe implementations: where unsafe traits define
```

ehuss (commented):
Thanks so much for adding this!

Since this is mostly new content, can you wrap the lines on sentences (aka [semantic line breaks](https://sembr.org/))? For example, it helps keep diffs cleaner instead of rewrapping lines. We've been trying to move more of the reference to this style. There is also a [style guide](https://github.com/rust-lang/reference/blob/master/STYLE.md) which has some of our guidelines.

src/unsafe-keyword.md
Havvy: I'm going to say that we should point out that this isn't an actual requirement of the language, but rather expected practice. There's nothing in `rustc` that will (or could) stop your code from compiling because you failed to explain what extra conditions you are imposing.

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+or the unsafe block could be inside an `unsafe fn` and use its own proof
+obligations to discharge the proof obligations of its callees.
+
+Unsafe blocks are used to wrap foreign libraries, make direct use of hardware
+or implement features not directly present in the language. For example, Rust
+provides the language features necessary to implement memory-safe concurrency
+in the language but the implementation of threads and message passing in the
+standard library uses unsafe blocks.
+
+Rust's type system is a conservative approximation of the dynamic safety
```
Havvy, line 55: Small wording quibble. Is it an approximation, or is it a subset?

(You can also just change "Rust's" to "The"

src/unsafe-keyword.md
```diff
@@ -0,0 +1,79 @@
...
+requirements, so in some cases there is a performance cost to using safe code.
+For example, a doubly-linked list is not a tree structure and can only be
+represented with reference-counted pointers in safe code. By using `unsafe`
+blocks to represent the reverse links as raw pointers, it can be implemented
+with only boxes.
```
Havvy, line 60: I have a feeling you are purposely hinting at https://rust-unofficial.github.io/too-many-lists/

I wonder if we could explicitly just tell people they can read that book.

RalfJung: This is old text, I didn't write this ;)

src/unsafe-keyword.md
RalfJung: :joy:

RalfJung:
> can you wrap the lines on sentences

That's how I wrote it, in fact; I forgot that the reference is switching to my preferred style ;)

src/unsafe-keyword.md
RalfJung: Oh I didn't know each lint has a page like that. :)

src/unsafe-keyword.md
RalfJung: "conservative approximation" means subset or superset, depending on which direction is the safe one.

I didn't write this part, but I think I prefer "conservative approximation" over subset or superset, since with subset/superset it is often ambiguous what the polarity of the order is.

src/unsafe-keyword.md
RalfJung: I didn't mean this to imply it must be the module-level docs, just 'somewhere inside the module'.

src/unsafe-keyword.md
RalfJung: I have adjusted the wording, please let me know what you think.

ehuss (approved):
Looks great, thanks!

PR #1281, opened 2022-10-10
Fix grammar for TypePathFn

ehuss:
The grammar for _TypePathFn_ was changed in https://github.com/rust-lang/rust/pull/45294 (Rust 1.25) to require parenthesis if there is more than one bound. [Here](https://github.com/ehuss/rust/blob/be1c7aad723126b2ea65543b4ceed54167b841a2/compiler/rustc_parse/src/parser/path.rs#L297-L298) is where the return type is parsed (notice that it has the `AllowPlus::No` restriction).


JohnTitor (approved):
LGTM (thanks for the reference)!

PR #1282, opened 2022-10-11
Denazification

sam0x2b:


src/identifiers.md
```diff
@@ -18,7 +18,7 @@ Identifiers follow the specification in [Unicode Standard Annex #31][UAX31] for
 * `foo`
 * `_identifier`
 * `r#true`
-* `Москва`
+* `Київ`
```
bjorn3, line 21: Rather than replacing Moscow with Kyiv, I think it should be replaced with an entirely neutral term.

sam0x2b: ok, lets make it "Lemon"

bjorn3 (approved):
Lemon is a good neutral term. I don't believe I have write permission, so someone else will have to merge.

ehuss:
I'm going to close this PR. I am very sympathetic for the events going on, and the Rust lang org has publicly stated its support for the people of Ukraine. However, this kind of PR and language isn't appropriate, and this isn't a forum for making a statement.

PR #1283, opened 2022-10-13
Clarify that #[repr(transparent)] is valid without a non-ZST field

kupiakos:
The following is valid, but the prior language suggests it is not:

```rust
#[repr(transparent)]
enum Banana {
    Phone,
}

#[repr(transparent)]
struct Blooey(());

#[repr(transparent)]
struct Orange;
```

Of note: it's still invalid to have a zero-sized field in a transparent
struct that has an alignment requirement higher than 1:
```rust
#[repr(align(2))]
struct Inner;

// error[E0691]: zero-sized field in transparent struct has alignment
//               larger than 1
#[repr(transparent)]
struct Fizz(Inner);
```

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -562,11 +562,11 @@ Like all ways to create undefined behavior in safe Rust, this is a bug.
 The `transparent` representation can only be used on a [`struct`][structs]
 or an [`enum`][enumerations] with a single variant that has:
 
-- a single field with non-zero size, and
+- a single field with non-zero size, and/or
 - any number of fields with size 0 and alignment 1 (e.g. [`PhantomData<T>`]).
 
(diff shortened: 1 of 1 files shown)


kupiakos:
An extension on that align-2 weirdness: it's valid to express in generic code, but not concretely. Rustc bug?

```rust
// This is just fine, because it's generic?
#[repr(transparent)]
struct Fizz<T = Inner>(T, [(); 10]);
```

kupiakos:
Looks like https://github.com/rust-lang/rust/issues/100954 is already tracking the inconsistencies with `#[repr(transparent)]`

traviscross:
Thanks @kupiakos for this.  We're going to close as we merged https://github.com/rust-lang/reference/pull/1568 instead.

PR #1285, opened 2022-10-19
Document that type parameter `Self` is unsized by default

fmease:
The (implicit) type parameter `Self` in traits is not `Sized` by default but the current documentation is not very clear about it.

> All type parameters have an implicit bound of `Sized`. […] The one exception is the implicit `Self` type of a trait.

Source: [Documentation of `std::marker::Sized`](https://doc.rust-lang.org/nightly/std/marker/trait.Sized.html):

It's *except `Self` **in traits*** specifically since e.g. the “sizedness” of `Self` in *structs* depends on the sizedness of the last field.

---

As an aside, in general not much is written about `Self` in the reference. That's why I chose this specific location in the docs.

### DIFF
--- a/src/special-types-and-traits.md
+++ b/src/special-types-and-traits.md
@@ -135,7 +135,7 @@ UnwindSafe>` is a valid type.
 ## `Sized`
 
 The [`Sized`] trait indicates that the size of this type is known at compile-time; that is, it's not a [dynamically sized type].
-[Type parameters] are `Sized` by default, as are [associated types].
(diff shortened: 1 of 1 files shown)


fmease:
Is there anything I can do to push this PR forward? Do you generally like or dislike this clarification? Do you think this is the right place?

As for a wider context, when working on fixing rust-lang/rust#24183, I took a look into the Reference to make sure I was doing the right thing but I left a bit confused since I couldn't find anything about the relation between `Self` and `Sized`. That's why I created patch.

ehuss (approved):
Thanks!

PR #1286, opened 2022-10-19
multiple examples, not just one example

amab8901:


ehuss:
Thanks for the PR! Can you say more about why it seems like there are multiple examples? From what I can see, the example has one parenthesized expression: `(2 + 3)`.

amab8901:
oh! You're right. A better idea would then be to add comments saying that one is grouped and the other is not

PR #1288, opened 2022-10-26
Fix a minor typo in the "Higher-ranked trait bounds" section

ohno418:


ehuss (approved):
Thanks for the PR!

I'm not sure I would consider this a "typo".  "type bound" and "trait bound" are essentially synonymous, and both terms are used extensively (particularly in the compiler).

There is a minor difference in the meaning of the terms. I would almost say that "higher-ranked trait bounds" is misnamed, because there can technically be a lone outlives bound such as `for<'a> Ty: 'a`, which does not involve traits at all.

However, since this section is using the "trait bounds" terminology, and "type bounds" aren't mentioned elsewhere in the reference, I figure this change should be fine to make.

ohno418:
>  "type bound" and "trait bound" are essentially synonymous

Yeah, I've considered this, but I googled "type bounds" and didn't come up with anything in particular. And that's why I submitted this PR.

> both terms are used extensively (particularly in the compiler).

I now searched and found a lot of "type bounds" in rustc source code. Indeed, both seems to be used as synonyms. Thanks for letting me know!

PR #1290, opened 2022-10-31
update aliasing rules section of the reference

RalfJung:
This makes explicit some de-facto rules that existed at least since Rust 1.0:
- Boxes also have aliasing requirements.
- References must outlive functions they are passed to. (It is not *entirely* clear that this was intended when the `dereferenceable` attribute was added, given that the attribute was also added to `Box` which can definitely be deallocated by functions they are passed to. But when that got fixed in https://github.com/rust-lang/rust/issues/66600, nobody suggested to also remove the attribute from references.)

Irrespective of whether we actually *want* these rules, they *do* reflect our current requirements and code not following them risks miscompilations. As such I think we should document them here and consider them the baseline for changes to the aliasing model. It's not great to have a baseline that was never carefully reviewed or designed, but for better or worse that's where we are at.

Also incorporates https://github.com/rust-lang/rust/pull/98017.

[Rendered](https://github.com/RalfJung/reference/blob/aliasing/src/behavior-considered-undefined.md)

Cc @rust-lang/lang

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -28,7 +28,7 @@ code.
   (e.g. `addr_of!(&*expr)`).
 * Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
   LLVM’s scoped [noalias] model, except if the `&T` contains an
-  [`UnsafeCell<U>`]. References and boxes must not be dangling while they are
+  [`UnsafeCell<U>`]. References and boxes must not be [dangling] while they are
   live. The exact liveness duration is not specified, but some bounds exist:
   * For references, the liveness duration is upper-bounded by the syntactic
     lifetime assigned by the borrow checker; it cannot be live any *longer* than
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -31,7 +31,8 @@ code.
   [`UnsafeCell<U>`]. References and boxes must not be dangling while they are
   live. The exact liveness duration is not specified, but some bounds exist:
   * For references, the liveness duration is upper-bounded by the syntactic
-    lifetime assigned by the borrow checker.
+    lifetime assigned by the borrow checker; it cannot be live any *longer* than
+    that lifetime.
   * Each time a reference or box is passed to or returned from a function, it is
     considered live.
   * When a reference (but not a `Box`!) is passed to a function, it is live at
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -26,8 +26,15 @@ code.
 * Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow LLVM’s
+  scoped noalias model, except if the `&T` contains an [`UnsafeCell<U>`].
+  References must not be dangling while they are live. (The exact liveness
+  duration is not specified, but it is certainly upper-bounded by the syntactic
+  lifetime assigned by the borrow checker. When a reference is passed to a
+  function, it is live at least as long as that function call, again except if
+  the `&T` contains an [`UnsafeCell<U>`].) All this also applies when values of
+  these types are passed in a (nested) field of a compound type, but not behind
+  pointer indirections.
 * Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
   data reached through a shared reference or data owned by an immutable binding
   is immutable, unless that data is contained within an [`UnsafeCell<U>`].
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -27,7 +27,7 @@ code.
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
 * Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow LLVM’s
-  scoped noalias model, except if the `&T` contains an [`UnsafeCell<U>`].
+  scoped [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
   References must not be dangling while they are live. (The exact liveness
   duration is not specified, but it is certainly upper-bounded by the syntactic
   lifetime assigned by the borrow checker. When a reference is passed to a
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -26,14 +26,20 @@ code.
 * Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow LLVM’s
-  scoped [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
-  References must not be dangling while they are live. (The exact liveness
-  duration is not specified, but it is certainly upper-bounded by the syntactic
-  lifetime assigned by the borrow checker. When a reference is passed to a
-  function, it is live at least as long as that function call, again except if
-  the `&T` contains an [`UnsafeCell<U>`].) All this also applies when values of
-  these types are passed in a (nested) field of a compound type, but not behind
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
+  LLVM’s scoped [noalias] model, except if the `&T` contains an
+  [`UnsafeCell<U>`]. References and boxes must not be dangling while they are
+  live. The exact liveness duration is not specified, but some bounds exist:
+  * For references, the liveness duration is upper-bounded by the syntactic
+    lifetime assigned by the borrow checker.
+  * Each time a reference or box is passed to or returned from a function, it is
+    considered live.
+  * When a reference (but not a `Box`!) is passed to a function, it is live at
+    least as long as that function call, again except if the `&T` contains an
+    [`UnsafeCell<U>`].
+
+  All this also applies when values of these
+  types are passed in a (nested) field of a compound type, but not behind
   pointer indirections.
 * Mutating immutable data. All data inside a [`const`] item is immutable. Moreover, all
   data reached through a shared reference or data owned by an immutable binding


src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,15 @@ code.
 * Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow LLVM’s
+  scoped noalias model, except if the `&T` contains an [`UnsafeCell<U>`].
```
ehuss, line 30: I'm curious why remove the link to LLVM's docs, even though the sentence is still referring to "LLVM's scoped noalias model"? Without the link, how can a reader follow up on what that means?

If you do intend to remove the link to llvm's "noalias", can you also remove the link reference to http://llvm.org/docs/LangRef.html#noalias at the bottom of the page?

RalfJung: Sorry, that was a mistake because this round-tripped through another document...

src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,15 @@ code.
...
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow LLVM’s
+  scoped [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+  References must not be dangling while they are live. (The exact liveness
+  duration is not specified, but it is certainly upper-bounded by the syntactic
+  lifetime assigned by the borrow checker. When a reference is passed to a
+  function, it is live at least as long as that function call, again except if
```
CAD97, line 34: ... Does this apply to `Box`? (The intent is *obviously* **no**, but what about for what we're telling LLVM?)

<details><summary>Because it's obviously safe to deallocate a <code>Box</code> allocation within a function,</summary>

```rust
fn example(x: Box<Data>) {
    drop(x);
    // here -- is `x` still considered `dereferencable`?
}
```

</details>

does this mean that `Box` needs to be marked with the hypothetical LLVM-`dereferenceable_on_entry` instead of just LLVM-`dereferenceable`? I suspect this is a nonissue in practice (memory allocation is either a very side-effectful built-in or an opaque environment call preventing code motion over it) but it's definitely *relevant*.

CAD97 (approved):
cc also @rust-lang/wg-unsafe-code-guidelines

Just to log it, I'm *weakly* against merging this without also providing a solution like rust-lang/rfcs#3336 to the question of `Box`-just-as-a-`Drop`-token. Even though the `struct { mem: Box<T>, ptr: &'self.mem T }` pattern is currently UB, it is in practice used for lack of a reasonable[^into-raw] alternative.

[^into-raw]: `into_raw`ing the owner is possible but makes interacting with the type much more difficult.

At this point, though, I do strongly agree that this is the correct way to go. The optimizations that these requirements provide are significant.

<details><summary>musings about the future</summary>

I also think that `Vec` should get the same treatment as `Box` (though noting so can be deferred until later) even though IIRC there is more code misusing `Vec` (including stdlib) than `Box` (in part because miri doesn't complain about `Vec` (yet)), although I no longer think this is a hard requirement. (However, if we do make `Vec` enforce uniqueness even when not reallocating, then I expect we'll need to expose some form of `RawVec` which doesn't in order to cover the gap.)

This definitely does make Rust `unsafe` significantly more treacherous to use than C, but `noalias` analysis is getting better to the point where Rust could be *more performant* than C (without `restrict`) while also being safer. If we manage to make working with raw pointers more ergonomic to the point where it's not significantly worse than C, then a simple first-order rule of "if you're managing an object with raw pointers, only use raw pointers" prevents the issues of mixing strong references and permissive pointers.

RalfJung:
> I'm weakly against merging this without also providing a solution like https://github.com/rust-lang/rfcs/pull/3336

So now we have a circular dependency. :joy: I opened this because that RFC has a diff for what it does to the reference, and the diff used to include the clarifications of this PR, which is misleading since the RFC doesn't actually change them. So I think this here needs to be merged for that RFC to even be able to discuss the aliasing rules in the level of detail that is needed to say what `MaybeDangling` does. We need a baseline to describe what that RFC changes.

Also as I said above, for better or worse, what this PR describes *is* the baseline that future changes will have to compare against. Not merging this PR won't make any code less likely to break.

Basically, this PR is descriptive in nature, not prescriptive. It describes the world as it is, whether we like it or not. We can change the world later, but first we need to agree on where we are.

src/behavior-considered-undefined.md
RalfJung: See https://github.com/rust-lang/rust/issues/66600 (linked in the PR description) :)

src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,21 @@ code.
...
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
+  LLVM’s scoped [noalias] model, except if the `&T` contains an
+  [`UnsafeCell<U>`]. References and boxes must not be dangling while they are
+  live. The exact liveness duration is not specified, but some bounds exist:
+  * For references, the liveness duration is upper-bounded by the syntactic
+    lifetime assigned by the borrow checker.
```
joshtriplett, line 34:
```suggestion
    lifetime assigned by the borrow checker; it cannot be live any *longer* than
    that lifetime.
```

joshtriplett:
Would it be legal to take a `&mut`, turn it into a `*mut`, free it, reallocate something at the *same* address, and then use the `&mut`?

src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,21 @@ code.
 * Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
+  LLVM’s scoped [noalias] model, except if the `&T` contains an
+  [`UnsafeCell<U>`]. References and boxes must not be dangling while they are
```
joshtriplett, line 31: What *exactly* does "dangling" mean here? What does this allow and disallow?

JakobDegen:
No. LLVM recognizes allocation level provenance, meaning the resulting `&mut` would not have provenance for the new allocation, regardless of address

src/behavior-considered-undefined.md
joshtriplett: Could this please forward-reference `#dangling-pointers` with a link?

nikomatsakis:
@rfcbot fcp merge

We discussed this in the @rust-lang/lang meeting today and arrived at the consensus that it was a good idea to describe current behavior, with the caveat that some of what we define as UB here may become legal behavior in the future (perhaps via language changes, or as part of adopting a more precise operational semantics).

joshtriplett:
Checking my box for documenting the current behavior. Not committing to wanting this to be the long-term behavior; I expect to have further thoughts on the associated RFC.

src/behavior-considered-undefined.md
RalfJung: Good point, done.

RalfJung:
@joshtriplett your box isn't checked though?

> Not committing to wanting this to be the long-term behavior

I assume you specifically mean the "outlives the function call" part? Yes I do expect that to be an important point in a future discussion, when we are ready to start deciding on actual aliasing model details. This is usually a knob that is fairly easy to turn (we can just have or not have that specific rule without surprising interactions elsewhere), but it has a lot of consequences for LLVM attributes and MIR optimizations.

Havvy:
@joshtriplett Your box is still unchecked. Is that intentional?

src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,22 @@ code.
 * Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
   [dangling] or unaligned, even in [place expression context]
   (e.g. `addr_of!(&*expr)`).
-* Breaking the [pointer aliasing rules]. `&mut T` and `&T` follow LLVM’s scoped
-  [noalias] model, except if the `&T` contains an [`UnsafeCell<U>`].
+* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
+  LLVM’s scoped [noalias] model, except if the `&T` contains an
+  [`UnsafeCell<U>`]. References and boxes must not be [dangling] while they are
```
saethlin, line 31:
```suggestion
  `UnsafeCell<U>`. References and boxes must not be [dangling] while they are
```

src/behavior-considered-undefined.md
```diff
@@ -26,8 +26,22 @@ code.
...
+  live. The exact liveness duration is not specified, but some bounds exist:
+  * For references, the liveness duration is upper-bounded by the syntactic
+    lifetime assigned by the borrow checker; it cannot be live any *longer* than
+    that lifetime.
+  * Each time a reference or box is passed to or returned from a function, it is
+    considered live.
+  * When a reference (but not a `Box`!) is passed to a function, it is live at
+    least as long as that function call, again except if the `&T` contains an
+    [`UnsafeCell<U>`].
```
saethlin, line 40:
```suggestion
    `UnsafeCell<U>`.
```
(these render as broken hyperlinks, I'm guessing the square brackets weren't intentional because the other types don't have them?)

RalfJung: There's already a ``[`UnsafeCell<U>`]`` in the code now so I just kept them. And it works correctly on https://doc.rust-lang.org/reference/behavior-considered-undefined.html?

saethlin: Hunh. It's unfortunate that they don't work on your branch. Ah well.

joshtriplett:
Not sure why my checkbox didn't stick. Transient network issue? Checked now, in any case.

RalfJung:
So FCP should start now, right? "Once a majority of reviewers approve (and at most 2 approvals are outstanding)"... let's see if the bot jumps into action...

ehuss (approved):
Thanks!

PR #1292, opened 2022-11-02
Primitive bit validity and padding guarantees

### LINKED ISSUE #1291: Describe bit validity and padding for primitive types
Though it's generally understood to be the case, the [type layout](https://doc.rust-lang.org/reference/type-layout.html) page of the reference doesn't explicitly guarantee that, for each primitive type, T:
- All bit patterns are valid instances of T
- All of the bytes of T are guaranteed to be initialized (none of them are padding)

joshlf:
Also fix a few typos.

Resolves #1291

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -58,13 +58,19 @@ aligned to 32 bits.
 
 ### Primitive Bit Validity and Padding
 
-For each primitive type, `T`, in the preceding table, any sequence of
-`size_of::<T>()` bytes is a valid instance of `T`. In other words,
+For each primitive type, `T`, in the preceding table other than `char`, any
+sequence of `size_of::<T>()` bytes is a valid instance of `T`. In other words,
 `transmute::<[u8; size_of::<T>()], T>(...)` is guaranteed to be sound.
 
-Similarly, for each primitive type, `T`, in the preceding table, `T` contains
-no padding or otherwise uninitialized bytes. In other words,
-`transmute::<T, [u8; size_of::<T>()]>(...)` is guaranteed to be sound.
+Similarly, for each primitive type, `T`, in the preceding table (including
+`char`), `T` contains no padding or otherwise uninitialized bytes. In other
+words, `transmute::<T, [u8; size_of::<T>()]>(...)` is guaranteed to be sound.
+
+#### `char` Bit Validity
+
+A `char`'s numerical value is guaranteed to be in the range 0 to 0x10FFFF,
+inclusive. It is undefined behavior to construct a `char` with a value outside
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  Can you say more about why it seems that this should be part of the layout chapter? My perspective is that the layout chapter is relevant to the alignment and padding of fields within a nominal type (and a few other circumstances), but the internal representation of a type is not relevant here. The types chapters defines what the valid bit patterns are (and by extension, the UB chapter explains invalid values and safety).

I'm reluctant to define this multiple times, or go into details about what kind of transmutes are sound within the layout chapter.

src/type-layout.md
```diff
@@ -583,6 +599,7 @@ used with any other representation.
 [`Sized`]: ../std/marker/trait.Sized.html
 [`Copy`]: ../std/marker/trait.Copy.html
 [dynamically sized types]: dynamically-sized-types.md
+[char-docs]: ../std/primitive.char.html
```
jswrenn, line 602:
```suggestion
[bool-docs]: ../std/primitive.bool.html
[char-docs]: ../std/primitive.char.html
```

src/type-layout.md
```diff
@@ -53,9 +53,25 @@ target platform. For example, on a 32 bit target, this is 4 bytes and on a 64
 bit target, this is 8 bytes.
 
 Most primitives are generally aligned to their size, although this is
-platform-specific behavior. In particular, on x86 u64 and f64 are only
+platform-specific behavior. In particular, on x86, `u64` and `f64` are only
```
jswrenn, line 56: Hm, this phrasing struck me as potentially confusing, since "x86" is often used as a shorthand to refer to "x86-64". (I did a brief double take, at least.) Per Wikipedia's [chronology](https://en.wikipedia.org/wiki/X86#Chronology), "IA-32" might be a more appropriate term: 
```suggestion
platform-specific behavior. In particular, on IA-32, `u64` and `f64` are only
```

...but I think you're referring to [`target_arch`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch) value "x86" (which is distinct from `x86_64`). Rust [inherits this shorthand from LLVM](https://llvm.org/doxygen/Triple_8h_source.html#l00085): 
```text
    x86,            // X86: i[3-9]86
```
...so you could instead disambiguate this by writing:
```suggestion
platform-specific behavior. In particular, on `target_arch = x86` (which has a word size of 32 bits), `u64` and `f64` are only
```

joshlf:
@ehuss I didn't realize that validity was documented elsewhere! I've put up #1293, which adds a bit of documentation to the pages in the "Types" section. Thanks for the feedback!

PR #1294, opened 2022-11-04
Disallow newline directly following `//`

autumnull:
Currently this reference marks up e.g.
```rust
fn main() { //
    println!("Hello, world!");
}
```
as

```
fn main() { <comment-start>//
    println!("Hello, world!");<comment-end>
}
```
because the ~\[`/` `!`] rule allows a newline to sneak in before the ~`\n`*

### DIFF
--- a/src/comments.md
+++ b/src/comments.md
@@ -2,7 +2,7 @@
 
 > **<sup>Lexer</sup>**\
 > LINE_COMMENT :\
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR!  It looks like CI is a little broken. I posted https://github.com/rust-lang/rust/pull/103952 which should fix it.

ehuss (approved):
Thanks!

PR #1295, opened 2022-11-07
Document inline const/const block expression

nbdd0121:
Stabilisation PR rust-lang/rust#104087

### DIFF
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -117,6 +117,37 @@ loop {
 }
 ```
 
+## `const` blocks
+
+> **<sup>Syntax</sup>**\
+> _ConstBlockExpression_ :\
+> &nbsp;&nbsp; `const` _BlockExpression_
+
+A *const block* is a variant of a block expression which evaluates in the compile time instead of in the run time.
+
+A `const` block allows you to define a constant value without having to define a new `const` item, and thus is also sometimes called as inline `const` or anonymous `const`.
+
+For example, this code:
+
+```rust
+# fn foo(x: i32) -> i32 { x }
+fn main() {
+    let x = foo(const { 1 + 1 });
+}
+```
+
+is equivalent to:
+
+```rust
+# fn foo(x: i32) -> i32 { x }
+fn main() {
+    const FOO: i32 = 1 + 1;
+    let x = foo(FOO);
+}
+```
+
+A `const` block supports type inference so you do not have to write the type of the constant value like a `const` item. `const` blocks are allowed to reference general parameters in their scope.
+
 ## `unsafe` blocks
 
 > **<sup>Syntax</sup>**\
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -23,6 +23,7 @@
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_FieldExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ClosureExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_AsyncBlockExpression_]\
+> &nbsp;&nbsp; &nbsp;&nbsp; | [_ConstBlockExpression_]\
 > &nbsp;&nbsp; &nbsp;&nbsp; | [_ContinueExpression_]\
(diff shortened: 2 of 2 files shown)


src/expressions/block-expr.md
```diff
@@ -117,6 +117,37 @@ loop {
 }
 ```
 
+## `const` blocks
+
+> **<sup>Syntax</sup>**\
+> _ConstBlockExpression_ :\
+> &nbsp;&nbsp; `const` _BlockExpression_
+
+A *const block* is a variant of a block expression which evaluates in the compile time instead of in the run time.
```
Havvy, line 126: Is it a requirement for other compilers to actually evaulate the expression at compile time?

I would think the primary definition would be that it is a block that's a const context.

src/expressions/block-expr.md
```diff
@@ -117,6 +117,37 @@ loop {
...
+
+A `const` block allows you to define a constant value without having to define a new `const` item, and thus is also sometimes called as inline `const` or anonymous `const`.
```
Havvy, line 128: Given the term defined is "const block", the quote ticks on subsequent usage is wrong. You may also consider giving it a name like "constant block" instead.

Same with const item. They are called "constant item" on their chapter.

src/expressions/block-expr.md
```diff
@@ -117,6 +117,37 @@ loop {
...
+
+```rust
+# fn foo(x: i32) -> i32 { x }
+fn main() {
+    const FOO: i32 = 1 + 1;
+    let x = foo(FOO);
+}
+```
+
+A `const` block supports type inference so you do not have to write the type of the constant value like a `const` item. `const` blocks are allowed to reference general parameters in their scope.
```
Havvy, line 149: Since this is a contrast to constant items, it should say so. Otherwise these just look like random facts.

Also, one sentence per line. There's two sentences here.

Havvy (changes requested):
You also need to update the chapter on constant evaluation to include these as a const context.

ehuss (commented):
@nbdd0121 Sorry I dropped the ball of the review here.

Can you update or answer these questions?
* Can you respond to the review comments above?
* It looks like this changed during the stabilization effort, such as changing to _ExpressionWithBlock_. Can you update the text for any changes that happened?
* Do all the rules and restrictions of a constant item hold for a const block? If so, can it say that? If not, can it explain the differences and similarities. For example, are they always [evalutated](https://doc.rust-lang.org/nightly/reference/items/constant-items.html#evaluation)?
* Should const blocks be added to the list of [constant expressions](https://doc.rust-lang.org/nightly/reference/const_eval.html#constant-expressions)?
* Is it also a [const context](https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context)? If so, can it be added there.
* My understanding is that this does not match `expr` in a macro. If that is correct, can you update the [`expr`](https://doc.rust-lang.org/nightly/reference/macros-by-example.html#metavariables) docs to mention that limitation? I understand this might change for the edition, but it doesn't look like that work has finished.

src/expressions/block-expr.md
nbdd0121: It seems that we have "constant item" but in the const eval chapter we use "const function", "const generic parameters" etc. I'll just keep it `const block` here.

The quote ticks are removed after term definition.

src/expressions/block-expr.md
nbdd0121: The const eval doc does mention that exprs in const context are evaluated at compile time, so I don't see an issue here.

src/expressions/block-expr.md
```diff
@@ -117,6 +117,37 @@ loop {
...
+```
+
+is equivalent to:
+
+```rust
+# fn foo(x: i32) -> i32 { x }
+fn main() {
+    const FOO: i32 = 1 + 1;
+    let x = foo(FOO);
+}
```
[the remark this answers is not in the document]
nbdd0121: Updated.

ehuss (approved):
Thanks!

RalfJung:
Auto-merge is enabled but nothing is happening...?

ehuss:
Thanks, I didn't see that it was blocked.

PR #1299, opened 2022-11-13
Document native library modifier `verbatim`

petrochenkov:
This is an accompanying PR for `verbatim` stabilization in https://github.com/rust-lang/rust/pull/104360.


ehuss (approved):
Looks like this has landed on nightly, so I'm going to go ahead and merge. Thanks!

PR #1300, opened 2022-11-15
Add an anchor to the "forwarding macro fragments" paragraph

danielhenrymantilla:
Since this is a subtle idiosyncracy into which macro users sometimes bump, it can be incredibly helpful to give them a pin-pointed link to the exact paragraph talking about this. Hence the idea of adding some kind of header, and thus an anchor, to it. Since it's just a paragraph, I've gone for a very low header, such as h4.

cc @nilstrieb

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -76,7 +76,7 @@ delimiters for the matcher will match any pair of delimiters. Thus, for
 instance, the matcher `(())` will match `{()}` but not `{{}}`. The character
 `$` cannot be matched or transcribed literally.
 
-#### Forwarding a matched fragment
+### Forwarding a matched fragment
 
 When forwarding a matched fragment to another macro-by-example, matchers in
 the second macro will see an opaque AST of the fragment type. The second macro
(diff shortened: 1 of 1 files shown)


Nilstrieb:
For context, I'm planning on implementing such a diagnostic

src/macros-by-example.md
```diff
@@ -76,6 +76,8 @@ delimiters for the matcher will match any pair of delimiters. Thus, for
 instance, the matcher `(())` will match `{()}` but not `{{}}`. The character
 `$` cannot be matched or transcribed literally.
 
+#### Forwarding a matched fragment
```
ehuss, line 79: I think it's fine to use a 3rd level header. We generally try to keep the headers nested in a natural order where h4 is nested in an h3 which is nested in an h2 which is nested in an h1. Jumping the order can break that hierarchy.

```suggestion
### Forwarding a matched fragment
```

ehuss (commented):
Thanks, seems like a great addition!

ehuss (approved):
Thanks!

PR #1307, opened 2022-11-23
Clearly specify the `instruction_set` effects

Lokathor:
~~This updates the `instruction_set` attribute entry to more clearly explain the additional restrictions that the attribute places on inlining. Since Rust performance heavily relies on inlining, it's actually a fairly big deal to document it.~~

~~Pinging @pnkfelix as the liaison for the `instruction_set` attribute during its development.~~

~~This explanation is actually slightly looser than the explanation given in the RFC: this explanation allows inlining between no-attribute code without inline asm and other code. So, T-lang should likely approve the update with an FCP or similar process. The intent is that once this is approved as the documentation for how the attribute works then https://github.com/rust-lang/rust/pull/104121 (which updates the MIR inliner to match these rules) will be easier to land.~~

Updates the attributes page to specify what the `instruction_set` attribute does and where you can use it.

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -361,6 +361,10 @@ It uses the [_MetaListPath_] syntax, and a path comprised of the architecture fa
 
 [_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
 
+It is a compilation error to use the `instruction_set` attribute on a target that does not support it.
+
+### On ARM
+
 For the `ARMv4T` and `ARMv5te` architectures, the following are supported:
 
 * `arm::a32` - Generate the function as A32 "ARM" code.
@@ -375,9 +379,7 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-If your function has neither the `instruction_set` attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
-This ends up creating a limitation to how often code is inlined:
+Using the `instruction_set` attribute has the following effects:
 
-* If a function has an `instruction_set` attribute it won't inline into a function of another instruction set.
-* If a function does not have an `instruction_set` attribute but *does* contain inline assembly, then the inline assembly is assumed to require the default instruction set of the build target, and so inlining between different instruction sets won't happen.
-* Otherwise, inlining happens normally.
+* If the address of the function is taken as a function pointer, the low bit of the address will be set to 0 (arm) or 1 (thumb) depending on the instruction set.
+* Any inline assembly in the function must use the specified instruction set instead of the target default.
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -355,15 +355,16 @@ trait object whose methods are attributed.
 
 ## The `instruction_set` attribute
 
-The *`instruction_set` attribute* may be applied to a function to enable code generation for a specific
-instruction set supported by the target architecture. It uses the [_MetaListPath_] syntax and a path
-comprised of the architecture and instruction set to specify how to generate the code for
-architectures where a single program may utilize multiple instruction sets.
+On some CPU architectures it is possible to mix more than one instruction set into a single program.
+The `instruction_set` attribute lets you control which instruction set a particular function will be generated for.
+It uses the [_MetaListPath_] syntax, and a path comprised of the architecture family name and instruction set name.
 
-The following values are available on targets for the `ARMv4` and `ARMv5te` architectures:
+[_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
 
-* `arm::a32` - Uses ARM code.
-* `arm::t32` - Uses Thumb code.
+For the `ARMv4T` and `ARMv5te` architectures, the following are supported:
+
+* `arm::a32` - Generate the function as A32 "ARM" code.
+* `arm::t32` - Generate the function as T32 "Thumb" code.
 
 <!-- ignore: arm-only -->
 ```rust,ignore
@@ -374,4 +375,8 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-[_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
+The rules for inlining functions using the `instruction_set` attribute are slightly more strict than normal:
+
+* If a function has an `instruction_set` attribute, then the function is assumed to require the given instruction set and it won't inline into a function of another instruction set.
+* If a function does not have an `instruction_set` attribute but *does* contain inline assembly, then the inline assembly is assumed to require the default instruction set of the build target and so inlining between instruction sets won't happen.
+* Otherwise, a function is assumed to not rely on a particular instruction set and the function *may* be inlined into any calling function (all other restrictions on inlining still apply).
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -375,8 +375,9 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-The rules for inlining functions using the `instruction_set` attribute are slightly more strict than normal:
+If your function has neither the instruction_set attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
+This ends up creating a limitation to how often code is inlined:
 
-* If a function has an `instruction_set` attribute, then the function is assumed to require the given instruction set and it won't inline into a function of another instruction set.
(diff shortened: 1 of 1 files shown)


pnkfelix:
Nominating for a quick skim from T-lang. In particular, the rules around inlining, starting here:

https://github.com/rust-lang/reference/pull/1307/files#diff-11f03f772eabdb1a762b90f1ce58afd34c7938b20beae226e451da44d4fcd73aR378

are we okay with this set of rules? There are three basic rules there, the first two dictating the cases where inlining will be disallowed, and the third dictating when inlining will be allowed (and thus the calling function must not rely on details of the instruction set, or at least not details that would be in conflict with the instruction set of the callee that is being inlined).

The two specific questions are:

 1. Are we all okay with the first two rules, in terms of saying "these are cases where it is safe to rely on the instruction set"?
     * (I personally think they seem to be very clear and intuitive rules, a kind of case where I expect people to say "well yes, how could it be any other way?")
 2. Are we okay with the third rule, in terms of saying "in all other cases the code of a fn must *not* rely on the (implicit) instruction set"?
     * (I personally am fine with it, and its hard for me to imagine a scenario where we would want to allow a fn to have such an implicit reliance without any marker via the `instruction_set` attribute. But, having said that ... I assume it would then be a source of potential undefined behavior, so we should still take care in making a decision here.)

Lokathor:
I should perhaps note that if a function is inlined in some places as well as being fully generated to call over to from other places, the generated version will be of the target's default instruction set as one would expect.

The goal here is to carve out a little wiggle room for the MIR inliner to do more work. Without the ability to inline no-attribute code into attribute'd code it makes any use of the attribute a complete drop in performance.

For more background on how we got here you can also read the zulip topic: https://rust-lang.zulipchat.com/#narrow/stream/189540-t-compiler.2Fwg-mir-opt/topic/What.20exactly.20does.20the.20MIR.20optimizer.20do.3F

pnkfelix:
Hey @Lokathor , sorry for the long delay in response here

The T-lang design team [talked about this PR](https://hackmd.io/DjG4XJY6SJGWrFWtPtKkvQ?both#%E2%80%9CClearly-specify-the-instruction_set-inlining-restrictions%E2%80%9D-reference1307) in its triage meeting yesterday

No one at the meeting expressed any problem with the rules as stated.

There was a slight concern about whether the language here is potentially misleading: we already don't say what inlining exactly "means", and all discussions of inlining attributes repeatedly say that they are all merely *hints*, with no requirements that inlining actually happen, or actually be strictly prohibited.

So, there was some feedback about whether this text could try incorporate that notion, albeit redundant, to make it clear that this text is non-normative, and is just meant to set users expectations about what might happen, without making strict guarantees in one direction or the other.

pnkfelix:
Also, there was some followup discussion on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/inlining.20and.20instruction_set.20reference.231307/near/319431370)

in that conversation, while attempting to summarize the meaning of the third rule, I wrote the something like the following:

> If your function has neither the instruction_set attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set

and there was some feedback indicating that this is a better way to write the third rule in question, with the belief that the resulting effects on whether inlining is allowed or not falls out naturally from that phrasing.

Maybe the third rule can instead be pulled out and phrased in something like the above form, and then you can keep the discussion of inlining, including all three parts, but reframe it all as *consequences* of the new phrasing?

nikomatsakis:

Removing nomination as this is waiting on @Lokathor

Lokathor:
Ah, slipped my mind, I'll try to get to this soon

Lokathor:
I think the new wording is sufficiently improved?

src/attributes/codegen.md
```diff
@@ -355,15 +355,16 @@ trait object whose methods are attributed.
 
 ## The `instruction_set` attribute
 
-The *`instruction_set` attribute* may be applied to a function to enable code generation for a specific
-instruction set supported by the target architecture. It uses the [_MetaListPath_] syntax and a path
-comprised of the architecture and instruction set to specify how to generate the code for
-architectures where a single program may utilize multiple instruction sets.
+On some CPU architectures it is possible to mix more than one instruction set into a single program.
+The `instruction_set` attribute lets you control which instruction set a particular function will be generated for.
```
ehuss, line 359: Can this be kept to something closer to the original? The reference uses a specific template for the definition of an attribute. That is, the first sentence starts with:

```
The *`foo` [attribute]* <explanation of what it does, and where it can be applied>
```

perhaps something like:

```suggestion
The *`instruction_set` [attribute]* may be applied to a function to control which instruction set the function will be generated for.
This allows mixing more than one instruction set in a single program on CPU architectures that support it.
```

src/attributes/codegen.md
```diff
@@ -374,4 +375,9 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-[_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
+If your function has neither the instruction_set attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
```
ehuss, line 378:
```suggestion
If your function has neither the `instruction_set` attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
```

ehuss (commented):
Just some editorial nits.

ehuss:
@pnkfelix Did you have any comments on the updated text? Do you want to start an FCP to formally approve the changes, or was the team discussion sufficient?

nikomatsakis:
@rustbot labels +I-lang-nominated

Raising for team discussion. I think that this text is fine as is. This whole feature is on that "grey line" where it doesn't theoretically affect language semantics etc etc.

src/attributes/codegen.md
```diff
@@ -374,4 +375,9 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-[_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
+If your function has neither the `instruction_set` attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
```
joshtriplett, line 378:
```suggestion
If your function has neither the `instruction_set` attribute nor inline assembly, and you're compiling for a target with multiple instruction sets, the code you write within that function should not presume any particular instruction set.
```

src/attributes/codegen.md
```diff
@@ -374,4 +375,9 @@ fn foo_arm_code() {}
...
+This ends up creating a limitation to how often code is inlined:
+
+* If a function has an `instruction_set` attribute it won't inline into a function of another instruction set.
+* If a function does not have an `instruction_set` attribute but *does* contain inline assembly, then the inline assembly is assumed to require the default instruction set of the build target, and so inlining between different instruction sets won't happen.
+* Otherwise, inlining happens normally.
```
joshtriplett, line 383: This seems like it's specific to ARM and Thumb; this may not necessarily *universally* be true for all future uses of `instruction_set`.

src/attributes/codegen.md
```diff
@@ -374,4 +375,9 @@ fn foo_arm_code() {}
 fn bar_thumb_code() {}
 ```
 
-[_MetaListPath_]: ../attributes.md#meta-item-attribute-syntax
+If your function has neither the `instruction_set` attribute nor inline assembly, then the code you write within that function should not presume any particular instruction set.
+This ends up creating a limitation to how often code is inlined:
```
joshtriplett, line 379:
```suggestion
For ARM in particular, `instruction_set` adds a limitation to where code gets inlined:
```

Amanieu:
I don't think the reference should make any mention of inlining, since that is an implementation detail. Instead the text should clearly document the *actual* guaranteed effects of `#[instruction_set]`:
- If the address of the function is taken as a function pointer. The low bit will be set to 0 or 1 depending on the instruction set.
- Any inline assembly in the function will use the specified instruction set instead of the target default.

The restrictions on inlining are mainly an artifact of how inline assembly is compiled with LLVM: it uses the instruction set of the containing function *after ininling*. As such we need to ensure that any functions containing inline assembly are not inlined into functions with a different instruction set.

Lokathor:
@joshtriplett updated per the 2023-06-27 T-lang meeting, ready for FCP if the text seems fine.

joshtriplett:
@rfcbot merge

RalfJung:
The PR description and title still need to be updated to the new content.

Lokathor:
should be fixed

nikomatsakis:
@rfcbot reviewed

ehuss (approved):
Thanks everyone!

PR #1309, opened 2022-11-27
Document the `efiapi` ABI

nicholasbishop:
Tracking issue for this feature: https://github.com/rust-lang/rust/issues/65815


ehuss (approved):
Thanks!  I missed that this got stabilized so the docs will be delayed a release.

PR #1310, opened 2022-11-30
replace `crateid` term with `crate_name`

crabrs:
The `crateid` term or attribute is replaced with `crate_name` in [chapter crates and source files](https://doc.rust-lang.org/reference/crates-and-source-files.html#the-crate_name-attribute). It was originally proposed in [RFC0109](https://rust-lang.github.io/rfcs/0109-remove-crate-id.html) and merged in [PR109](https://github.com/rust-lang/rfcs/pull/109)

### DIFF
--- a/src/items/extern-crates.md
+++ b/src/items/extern-crates.md
@@ -20,9 +20,9 @@ clause can be used to bind the imported crate to a different name.
 The external crate is resolved to a specific `soname` at compile time, and a
 runtime linkage requirement to that `soname` is passed to the linker for
 loading at runtime. The `soname` is resolved at compile time by scanning the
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):

ehuss (approved):
Oh my, that is some old pre-1.0 cruft. Thanks!

I think long term the language here could be cleaned up quite a bit, as it is veering into specific compiler behavior ("scanning the compiler's library path"), and the `soname` stuff seems more relevant to shared libraries. It also doesn't really define what a "`name` attribute" means, or what happens if both `extern crate` and `#[crate_name]` are not used.

PR #1313, opened 2022-12-27
Fix parsing of doc comments which are turned into raw strings

kraktus:
PR prompted by https://github.com/dtolnay/syn/issues/771

```rust
macro_rules! m {
    (#[doc = $tt:tt]) => { stringify!($tt) };
}

fn main() {
    println!(m! {
        ///test
    });
   // output: r"test"
}
```

Will also open a PR for rustdoc

### DIFF
--- a/src/comments.md
+++ b/src/comments.md
@@ -44,8 +44,8 @@ Non-doc comments are interpreted as a form of whitespace.
 Line doc comments beginning with exactly _three_ slashes (`///`), and block
 doc comments (`/** ... */`), both inner doc comments, are interpreted as a
 special syntax for [`doc` attributes]. That is, they are equivalent to writing
-`#[doc="..."]` around the body of the comment, i.e., `/// Foo` turns into
-`#[doc="Foo"]` and `/** Bar */` turns into `#[doc="Bar"]`.
+`#[doc=r"..."]` around the body of the comment, i.e., `/// Foo` turns into
+`#[doc=r"Foo"]` and `/** Bar */` turns into `#[doc=r"Bar"]`.
 
 Line comments beginning with `//!` and block comments `/*! ... */` are
 doc comments that apply to the parent of the comment, rather than the item


ehuss:
Thanks for the PR!  I don't think this is entirely accurate. For example, a doc-comment passed to a proc-macro will receive an escaped string, such as `#[doc = "comment with  \"quotes\" inside"]` (see [`proc_macro_server.rs`](https://github.com/rust-lang/rust/blob/29d76cc6f5064e393440019198328b4424302633/compiler/rustc_expand/src/proc_macro_server.rs#L201-L223)).  Also, the encoding form for a token in a macro_rules `tt` will use sufficient `#` to encode properly (see [`desugar`](https://github.com/rust-lang/rust/blob/29d76cc6f5064e393440019198328b4424302633/compiler/rustc_parse/src/parser/mod.rs#L315-L371)). 

I'm not sure if these exact behaviors are something we want to enshrine in the documentation just yet, as I suspect they have not had much focus. Perhaps you can engage with the lang team to see if this difference in behavior is something they want to commit to?

Given that these are mostly just illustrations of how a doc comment gets converted to an attribute, I don't think it is trying to exhaustively specify the exact method of that conversion. I think it would be best to leave it as-is, and have a separate paragraph discussing the conversions and how they interact with macros of different kinds.

kraktus:
Thanks for the insight. The issue with the reference as is, is that user could think unicode encoding would be interpreted while it's not.


Maybe just make it clear they're turned into raw-literal strings without weighting on the implementation?

ehuss:
Sorry, I'm a bit confused. I'm not sure what is meant by "unicode encoding". All Rust strings are unicode. 

I'm also not sure what is meant by waiting (I assume) on the implementation. Waiting for what implementation?

As I mentioned, I don't think it is correct to say they are raw string literals, since in some circumstances (proc-macros) they are not, so it would be incorrect for that case.

kraktus:
Sorry, I meant that a user writing `/// foo\u{2122}` could expect from reading the reference that `\u{2122}` would be turned into `™`

traviscross:
Thanks for the PR.  Probably the text could be improved to clarify what's happening exactly, but simply adding `r` here doesn't quite capture this, so we'll go ahead and just close this PR.

PR #1315, opened 2023-01-07
Add links to definitions of terminology ...

steffahn:
... in the definition for what an expression that “has an extended temporary scope” is.

Improvement discussed in and motivated by the question in this [thread on URLO].

[thread on URLO]: https://users.rust-lang.org/t/how-to-understand-extending-based-on-expressions/87068/16?u=steffahn


ehuss (approved):
Thanks!

PR #1317, opened 2023-01-10
Document type alias impl trait

oli-obk:
This feature is on track to be stabilized soon: https://github.com/rust-lang/rust/issues/63063

### DIFF
--- a/src/types/impl-trait.md
+++ b/src/types/impl-trait.md
@@ -7,9 +7,15 @@
 
 `impl Trait` provides ways to specify unnamed but concrete types that
 implement a specific trait.
-It can appear in two sorts of places: argument position (where it can act as an anonymous type parameter to functions), and return position (where it can act as an abstract return type).
+It can appear in a few places:
+
+* argument position (where it can act as an anonymous type parameter to functions),
+* return position (where it can act as an abstract return type),
+* in type aliases (where it acts as an abstract type), and
+* in associated types (where it acts as an abstract type).
 
 ```rust
+#![feature(type_alias_impl_trait)]
 trait Trait {}
 # impl Trait for () {}
 
@@ -19,6 +25,12 @@ fn foo(arg: impl Trait) {
 
 // return position: abstract return type
 fn bar() -> impl Trait {
+}
+
+// type alias: abstract type
+type Foo = impl Trait;
+fn baz() -> Foo {
+    let x: Foo = ();
 }
 ```
 ## Anonymous type parameters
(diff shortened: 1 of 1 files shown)


src/types/impl-trait.md
```diff
@@ -111,10 +123,189 @@ fn foo() -> impl Trait {
...
+// The above function can be changed to the
+// following without a change in behaviour for callers.
+
+type Foo = impl Trait;
+fn foo() -> Foo {
+    value_of_type_that_implements_Trait
+}
+```
+
+When such a type alias is used as the return type of multiple functions, all functions must use the same hidden type.
```
Mark-Simulacrum, line 145: I think it would be good to start this section with something that defines "hidden type" as the real type behind a TAIT.

src/types/impl-trait.md
```diff
@@ -111,10 +123,189 @@ fn foo() -> impl Trait {
...
+
+```rust,ignore
+fn bop(x: Bar) {
+    println!("{x:?}");
+}
+```
+
+Note that this is very different from using `impl Trait` in argument position, as there is no anonymous generic parameter introduced.
+
+Binding a hidden type works in both directions, not just assigning a hidden type value to the opaque type, but also reading an opaque type into a hidden type value:
```
Mark-Simulacrum, line 177: I think "Binding a hidden type" is confusing to me here. Can we rephrase to avoid "binding"? Or explain what we mean by that? Perhaps something like: "The hidden type can be defined by any usage of it. This includes returning a concrete type from a function with the opaque type and assigning a value with the opaque type to a variable with a concrete type."

src/types/impl-trait.md
```diff
@@ -111,10 +123,189 @@ fn foo() -> impl Trait {
...
+impl Trait for () {}
+type Foo = impl Trait;
+fn foo() -> Foo {}
+
+trait Bar {}
+impl Bar for Foo {}
+impl Bar for i32 {}
+```
+
+This is legal, because `i32` could not possibly be a hidden type of `Foo`, because it doesn't implement `Trait` wich is a requirement for all hypothetical hidden types of `Foo`.
```
Mark-Simulacrum, line 263: I find this pretty surprising -- it means adding a trait impl is a direct breaking change, not just because of changes to e.g. method call resolution or similar. Is this limited to the same crate perhaps?

src/types/impl-trait.md
```diff
@@ -111,10 +123,189 @@ fn foo() -> impl Trait {
...
+    }
+}
+```
+
+This way you do not need to write burdensome `Future` impls yourself. Similarly with complex `Iterator` implementations.
+
+#### Defining scope
+
+Similar to return-position-impl-trait, you can only bind a hidden type of a type-alias-impl-trait within a specific "scope" (henceforth called "defining scope").
+The defining scope of a return-position-impl-trait is the function's body, excluding other items nested within that function's body (we may want to relax that restriction on return-position-impl-trait in the future).
```
Mark-Simulacrum, line 297: The note around future relaxation seems a little atypical for the reference, but maybe we can keep it.

src/types/impl-trait.md
oli-obk: hmm... RPITs have the same thing, so I could bring it up where "abstract types" are explained?

src/types/impl-trait.md
oli-obk: Yes, this is limited to impls for traits in the same crate.

src/types/impl-trait.md
```diff
@@ -53,6 +65,16 @@ This includes generic arguments for the return type or any const generics.
 >
 > Therefore, changing the function signature from either one to the other can constitute a breaking change for the callers of a function.
 
+## Abstract types
+
+> Note: these are also called "existential types" or "opaque types".
```
ehuss, line 70: Style-wise, I suggest using italics instead of quotes when introducing terms or marking key phrases. I realize the rest of the text isn't always consistent with that, but I think that is the generally accepted way to format these. (There's a few other examples below that could be changed.)

```suggestion
> Note: these are also called *existential types* or *opaque types*.
```

src/types/impl-trait.md
```diff
@@ -53,6 +65,16 @@ This includes generic arguments for the return type or any const generics.
...
+
+Abstract types are backed by a "hidden type", but only expose certain traits of that hidden type.
+Every abstract type has exactly one hidden type, and the hidden type is inaccessible (it could even be a private
+type or an unnameable type like a closure). It is required that every abstract type gets a hidden type
+assigned/constrained/bound within its "defining scope" (more to that later). It may get constrained multiple times,
```
ehuss, line 75: I think it would be good to link instead of vaguely referring to some other text.

```suggestion
assigned/constrained/bound within its [*defining scope*](#defining-scope). It may get constrained multiple times,
```

src/types/impl-trait.md
```diff
@@ -53,6 +65,16 @@ This includes generic arguments for the return type or any const generics.
 >
 > Therefore, changing the function signature from either one to the other can constitute a breaking change for the callers of a function.
 
+## Abstract types
+
+> Note: these are also called "existential types" or "opaque types".
+
+Abstract types are backed by a "hidden type", but only expose certain traits of that hidden type.
```
ehuss, line 72:
```suggestion
Abstract types are backed by a *hidden type*, but only expose certain traits of that hidden type.
```

src/types/impl-trait.md
```diff
@@ -111,10 +133,193 @@ fn foo() -> impl Trait {
 doesn't allow the caller to determine the return type.
 Instead, the function chooses the return type, but only promises that it will implement `Trait`.
 
+
+## Type Alias Impl Trait
+
+In contrast to `impl Trait`s in function return types, type aliases can be used in more places than just return types and the same type alias can be used multiple times.
+
+```rust,ignore
```
ehuss, line 141: I'd generally prefer to avoid using `ignore` whenever possible, as it has caused problems in the past. Can you try to remove `ignore` from all of the examples? If there is some irrelevant parts needed, they can be prefixed with `#` to hide them. For example:

```
# trait Trait {}
# struct ImplementsTrait;
# impl Trait for ImplementsTrait {}

fn foo() -> impl Trait {
    # let value_of_type_that_implements_Trait = ImplementsTrait;
    value_of_type_that_implements_Trait
}
```

src/types/impl-trait.md
```diff
@@ -111,10 +133,193 @@ fn foo() -> impl Trait {
...
+type Foo = impl Trait;
+fn foo() -> Foo {
+    value_of_type_that_implements_Trait
+}
+```
+
+When such a type alias is used as the return type of multiple functions, all functions must use the same hidden type.
+This is similar to how all code paths in a function with a return-position-impl-trait must return the same type:
+
+```rust,ignore
```
ehuss, line 158: When an example explicitly illustrates a failure, they should be marked with `compile_fail`.  Even better is to include the error code, which will be validated that the correct error is reported.

```suggestion
```rust,compile_fail,E0404
```

src/types/impl-trait.md
```diff
@@ -111,10 +133,193 @@ fn foo() -> impl Trait {
...
+fn burp(x: Bar) -> Bar {
+    x
+}
+```
+
+### Binding types
+
+You can also use type-alias-impl-trait for the type
+of local variables, constants, statics, ...
```
ehuss, line 209: Is `...` here to be completed?

Also, when possible it is preferred to link to concepts (even if they seem painfully obvious to you).

```suggestion
of [local variables], [constants], [statics], ...

[local variables]: ../variables.md
[constants]: ../items/constant-items.md
[statics]: ../items/static-items.md
```

src/types/impl-trait.md
```diff
@@ -53,6 +65,16 @@ This includes generic arguments for the return type or any const generics.
...
 
+## Abstract types
+
+> Note: these are also called "existential types" or "opaque types".
+
+Abstract types are backed by a "hidden type", but only expose certain traits of that hidden type.
+Every abstract type has exactly one hidden type, and the hidden type is inaccessible (it could even be a private
+type or an unnameable type like a closure). It is required that every abstract type gets a hidden type
+assigned/constrained/bound within its "defining scope" (more to that later). It may get constrained multiple times,
+but each such constraining site must be using the exact same type (to ensure the abstract type has exactly one hidden type).
```
ehuss, line 76: This mentions *constraining* but doesn't directly define what it means. Would it be possible to add some kind of definition for that? For example, the [generic implementations](https://doc.rust-lang.org/nightly/reference/items/implementations.html#generic-implementations) section has a definition of *constrain*. Perhaps those are similar enough they can be shared, or perhaps they are different enough that another explicit statement here would be good?

src/types/impl-trait.md
```diff
@@ -111,10 +133,193 @@ fn foo() -> impl Trait {
...
+This way you do not need to write burdensome `Future` impls yourself. Similarly with complex `Iterator` implementations.
+
+### Defining scope
+
+Similar to return-position-impl-trait, you can only bind a hidden type of a type-alias-impl-trait within a specific "scope" (henceforth called "defining scope").
+The defining scope of a return-position-impl-trait is the function's body, excluding other items nested within that function's body.
+
+The defining scope of a type-alias-impl-trait is the scope in which it was defined. So usually a module and all its child items, but it can also be a function body, const initializer and similar scopes that can define items.
+
+Any use of the type-alias-impl-trait within the defining scope will become a **defining use** (meaning it binds a hidden type), if the type is coerced to or from, equated with, or subtyped with any other concrete type.
```
ehuss, line 315: Can this maybe add an example that illustrates the consequence of a defining scope of tait? I think there is a connection that I am missing from binding of the hidden type in a defining scope to the limitations now implied of using that type alias outside of that defining scope. (Or perhaps those limitations are the same within and without the defining scope, I'm not clear.)

ehuss (commented):
Thank you for preparing this! Unfortunately I have mostly editorial nits instead of substantive feedback right now.

PR #1319, opened 2023-01-20
Syntax correction for trait-bounds.md

jamesreprise:
```rs
fn f<A> where A: Copy () {} // ->
fn f<A>() where A: Copy {}
```

The former isn't valid Rust as far as I can tell, though I may be missing something.

```
   Compiling playground v0.0.1 (/playground)
error: expected `(`, found keyword `where`
 --> src/lib.rs:1:9
  |
1 | fn f<A> where A: Copy () {}
  |         ^^^^^ expected `(`

error: could not compile `playground` due to previous error
```


ehuss (approved):
Thanks!

PR #1320, opened 2023-01-21
make more easy to read

tshepang:


traviscross:
Thanks for the PR.  We reviewed and decided that, in this case, the extra comma doesn't make it easier to read for us.

PR #1323, opened 2023-01-27
Relax ordering rules for `asm!` operands

Amanieu:
This updates the reference to match the changes in rust-lang/rust#105798

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -43,13 +43,13 @@ format_string := STRING_LITERAL / RAW_STRING_LITERAL
 dir_spec := "in" / "out" / "lateout" / "inout" / "inlateout"
 reg_spec := <register class> / "\"" <explicit register> "\""
 operand_expr := expr / "_" / expr "=>" expr / expr "=>" "_"
-reg_operand := dir_spec "(" reg_spec ")" operand_expr
-operand := reg_operand
+reg_operand := [ident "="] dir_spec "(" reg_spec ")" operand_expr
 clobber_abi := "clobber_abi(" <abi> *("," <abi>) [","] ")"
 option := "pure" / "nomem" / "readonly" / "preserves_flags" / "noreturn" / "nostack" / "att_syntax" / "raw"
 options := "options(" option *("," option) [","] ")"
-asm := "asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," clobber_abi) *("," options) [","] ")"
-global_asm := "global_asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," options) [","] ")"
+operand := reg_operand / clobber_abi / options
+asm := "asm!(" format_string *("," format_string) *("," operand) [","] ")"
+global_asm := "global_asm!(" format_string *("," format_string) *("," operand) [","] ")"
 ```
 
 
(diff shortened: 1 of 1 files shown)


src/inline-assembly.md
```diff
@@ -43,13 +43,13 @@ format_string := STRING_LITERAL / RAW_STRING_LITERAL
...
-reg_operand := dir_spec "(" reg_spec ")" operand_expr
-operand := reg_operand
+reg_operand := [ident "="] dir_spec "(" reg_spec ")" operand_expr
 clobber_abi := "clobber_abi(" <abi> *("," <abi>) [","] ")"
 option := "pure" / "nomem" / "readonly" / "preserves_flags" / "noreturn" / "nostack" / "att_syntax" / "raw"
 options := "options(" option *("," option) [","] ")"
-asm := "asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," clobber_abi) *("," options) [","] ")"
-global_asm := "global_asm!(" format_string *("," format_string) *("," [ident "="] operand) *("," options) [","] ")"
+operand := reg_operand / clobber_abi / options
+asm := "asm!(" format_string *("," format_string) *("," operand) [","] ")"
```
ehuss, line 51: This grammar doesn't seem to capture the order requirements for how it is parsed. Am I missing something for how this is defined?  That is, why is it not something closer to:

```
asm := "asm!(" format_string *("," format_string) *("," operand) *("," ident "=" operand) *("," options) [","] ")"
```

nikomatsakis:
Discussed in lang-team meeting: matching the implementation seems good, we don't need further review, but it'd be good @Amanieu to figure out whether @ehuss is correct.

src/inline-assembly.md
Amanieu: The new rules allow options and operands to be freely mixed, so there are no order requirements anymore (except that the template strings must come first).

ehuss: I think what I'm getting after is to capture the rule that positional operands must come before named operands. This grammar seems to allow:

```rust
let x: i64;
asm!("add {}, {foo}, {}", out(reg) x, foo=in(reg) 2, in(reg) 1);
```

But that generates an error "positional arguments cannot follow named arguments or explicit register arguments".

Amanieu: Right, but then again options can appear in any position between named & positional arguments. It's a bit difficult to express all the details in the grammar.

ehuss: If I understand how it parses correctly, the following doesn't seem too unwieldy:

```
asm := "asm!(" format_string *("," format_string) operands ")"
operands := *("," reg_operand / clobber_abi / options) *("," ident "=" reg_operand) *("," clobber_abi / options)
```

Does that seem like it would be OK to have?

Amanieu: I gave this a try but I think it just makes the grammar harder to read overall. I prefer leaving it as it currently is and describing the operand ordering constraints in the text.

Amanieu: The example you gave isn't actually correct since it doesn't cover explicit register operands which don't have names and can be mixed with named operands.

ehuss: That's fair. Thanks for taking a look!

ehuss (approved):
Thanks!

PR #1324, opened 2023-01-31
remove confusing words

tshepang:
I thought it wanted to be "at least once", but it could also have wanted to be "immediately". Either way, it seems not important.


ehuss (approved):
Thanks!

PR #1325, opened 2023-01-31
Reword "expression for a match arm"

tmandry:
It's unclear what this bullet means. If it means a match guard, that is covered in the line above. If it referred to the scrutinee, that would be wrong. I don't think it's possible to embed an expression inside of a match arm _pattern_.

This came up in the [lang team triage meeting][meeting].

[meeting]: https://hackmd.io/JACzq-ssT4qtH0DS3B8nQQ?both#%E2%80%9CIntroduce-terminating-scope-for-tail-expressions-of-breakable-scopes%E2%80%9D-rust106493

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -162,7 +162,6 @@ smallest scope that contains the expression and is one of the following:
 * The `else` block of an `if` expression.
 * The condition expression of an `if` or `while` expression, or a `match`
   guard.
-* The expression for a match arm.
 * The second operand of a [lazy boolean expression].
 
 > **Notes**:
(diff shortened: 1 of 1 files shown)


ehuss:
I believe it is referring to the expression to the right of `=>`.  That is, the terminating scope defined [here](https://github.com/rust-lang/rust/blob/f361413cbf44ce2f144df59fc440cd484af4a56e/compiler/rustc_hir_analysis/src/check/region.rs#L181).

tmandry:
Ah, thank you @ehuss! I've put the bullet back and added the word "body", so hopefully it's clearer now.

ehuss (approved):
Thanks!  I suppose we don't have a name for that expression outside of the compiler. The term "body" seems to stem back from https://github.com/rust-lang/rust/commit/d2f9b150529384e29db337bcdd37a72b797f6aae, but I'm not sure if I've seen it commonly used that way. Anyways, maybe something to consider in the future.

PR #1327, opened 2023-02-02
fix place expression context example

RalfJung:
`addr_of!(&_)` is a double-address-of... no idea what I was thinking when writing this, probably it's just a typo.


ehuss (approved):
Thanks!

PR #1328, opened 2023-02-03
Clarify that free constants are always evaluated at compile time

XrXr:
See https://github.com/rust-lang/rust/issues/93838
It seems like everyone is onboard with blessing this (useful) behavior as stable.

### DIFF
--- a/src/items/constant-items.md
+++ b/src/items/constant-items.md
@@ -89,10 +89,15 @@ m!(const _: () = (););
 // const _: () = ();
 ```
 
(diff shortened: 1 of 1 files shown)


RalfJung:
I don't think we want to do any special-casing based on the name (or lack thereof) of a constant. This rule applies to all non-associated constants and should be documented as such.

scottmcm (approved):

nikomatsakis:
Discussed in the @rust-lang/lang meeting and concluded that this represents existing consensus. Ship it!

PR #1330, opened 2023-02-08
Eliminate 'half open' terminology from range pattern grammar

### LINKED ISSUE #1329: RangeLimits naming: inconsistent meaning of "half open"
In https://github.com/dtolnay/syn/issues/1373 I noticed an inconsistency between how ranges are described in the reference vs how the compiler uses the same terminology.

Rustc has the following enum which distinguishes `..` vs `..=` in range patterns and expressions:
https://github.com/rust-lang/rust/blob/1.67.0/compiler/rustc_ast/src/ast.rs#L1323-L1328

```rust
pub enum RangeLimits {
    /// Inclusive at the beginning, exclusive at the end (..)
    HalfOpen,
    /// Inclusive at the beginning and end (..=)
    Closed,
}
```

Ranges in rustc are effectively `(Option<Expr>, RangeLimits, Option<Expr>)`.

Meanwhile a different approach is currently taken by the Rust Reference: https://doc.rust-lang.org/1.67.0/reference/patterns.html#range-patterns

```abnf
RangePattern :
    InclusiveRangePattern | HalfOpenRangePattern

InclusiveRangePattern :
    RangePatternBound `..=` RangePatternBound

HalfOpenRangePattern :
    RangePatternBound `..` | `..=` RangePatternBound
```

Notice how it has a `..=` pattern which it refers to as half open, even though `..=` is never considered half open in rustc.

| range | rustc terminology | reference terminology |
|---|---|---|
| `lo..` | half open | half open |
| `..=hi` | closed (**opposite of half open**) | half open (**!**) |
| `lo..=hi` | closed | inclusive |

Do we know whether one of these uses of the "half open" terminology is more preferred? Might we reach agreement about which of rustc or the reference should be changed?

dtolnay:
As described in #1329, the Reference's current use of this terminology does not match rustc's use.

Following the core::ops trait terminology (as is already done for [RangeExpression](https://doc.rust-lang.org/1.67.0/reference/expressions/range-expr.html)) results in a clearer grammar.

Closes #1329.

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -380,16 +380,19 @@ match tuple {
 
 > **<sup>Syntax</sup>**\
 > _RangePattern_ :\
-> &nbsp;&nbsp; &nbsp;&nbsp; _InclusiveRangePattern_\
-> &nbsp;&nbsp; | _HalfOpenRangePattern_\
+> &nbsp;&nbsp; &nbsp;&nbsp; _RangeInclusivePattern_\
+> &nbsp;&nbsp; | _RangeFromPattern_\
+> &nbsp;&nbsp; | _RangeToInclusivePattern_\
 > &nbsp;&nbsp; | _ObsoleteRangePattern_
 >
-> _InclusiveRangePattern_ :\
+> _RangeInclusivePattern_ :\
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1331, opened 2023-02-16
Document `cmpxchg16b` target feature

Nugine:
Stabilization PR:
+ https://github.com/rust-lang/rust/pull/106774


ehuss (approved):
Looks like https://github.com/rust-lang/rust/pull/106774 has merged, so looks like this is ready to go. Thanks!

PR #1332, opened 2023-02-20
Add `\r\n` to string_continue grammar

ModProg:
Not sure if this is necessary, but I wasn't 100% sure that crlf line endings would be supported (though in hindsight would be strange if they weren't)

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -148,7 +148,7 @@ which must be _escaped_ by a preceding `U+005C` character (`\`).
 > &nbsp;&nbsp; )<sup>\*</sup> `"` SUFFIX<sup>?</sup>
 >
 > STRING_CONTINUE :\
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! I think I would prefer to add a holistic solution for CRLF normalization instead of trying to address individual places. There are various points noted in #626 about CRLF normalization (and bare CR handling) that should probably be added as a whole (I believe per https://github.com/rust-lang/rust/issues/62865 that it is translated globally, but I haven't investigated in a long while).

ModProg:
makes sense, it was just that having only the `\n` here sounded incomplete. But as there is probably no case where CRLF will be preserved in source code there is a better solution than listing it.

ehuss:
Closing as superseded by #1459.  Thanks!

PR #1338, opened 2023-03-06
fix: Typo/reference

alexpovel:
Hi! Found a tiny typo. Although not a 100% certain this fix is correct, the sentence reads much clearer this way.


ehuss (approved):
Thanks! I'm not sure if that particular wording was intentional, but "as" seems to read better to me, too.

PR #1341, opened 2023-03-19
Inline assembly: Fix repeated and unordered items in guaranteed directives

chirsz-ever:
`.balign`, `.balignl` and `.balignw` are repeated, some other directives are not placed correctly in lexicographical order.


ehuss:
@Amanieu, I assume this is ok?

Amanieu (approved):

ehuss (approved):
Thanks!

PR #1342, opened 2023-03-19
Improve labeled blocks documentation

est31:
* list labeled blocks in the namespaces page
* add an example
* improve wording

### DIFF
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -249,8 +249,27 @@ A `break` expression is only permitted in the body of a loop, and has one of the
 > &nbsp;&nbsp; [_BlockExpression_]
 
 Labelled block expressions are exactly like block expressions, except that they allow using `break` expressions within the block.
-Unlike other loops, `break` expressions within a label expression *must* have a label (i.e. the label is not optional).
-Unlike other loops, labelled block expressions *must* begin with a label.
+Unlike loops, `break` expressions within a labelled block expression *must* have a label (i.e. the label is not optional).
+Similarly, labelled block expressions *must* begin with a label.
+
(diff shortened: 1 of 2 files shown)


src/expressions/loop-expr.md
```diff
@@ -249,8 +249,16 @@ A `break` expression is only permitted in the body of a loop, and has one of the
...
-Unlike other loops, labelled block expressions *must* begin with a label.
+Unlike with loops, `break` expressions within a labelled block expression *must* have a label (i.e. the label is not optional).
+Unlike loops, labelled block expressions *must* begin with a label.
+
+```rust
+'a: {
+    loop {
+        break 'a;
+    }
+}
```
ehuss, line 260: Thanks for adding an example! I was thinking that it might be a little confusing to also have a loop here, since this is in the loop chapter. The break isn't necessary to be in a loop. What do you think of using the example from the RFC?

```suggestion
# fn do_thing() {}
# fn condition_not_met() -> bool { true }
# fn do_next_thing() {}
# fn do_last_thing() {}
'block: {
    do_thing();
    if condition_not_met() {
        break 'block;
    }
    do_next_thing();
    if condition_not_met() {
        break 'block;
    }
    do_last_thing();
}
```

est31: Maybe one could use the one from the [release announcement](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#break-from-labeled-blocks)? It is a modified version that contains a returned value.

ehuss: That works, too!

ehuss (approved):
Thanks!

PR #1344, opened 2023-03-27
Explain typos in `asm!` can be unsound

workingjubilee:
This sort of thing may seem "obvious", but in practice people often have high expectations of Rust. They expect safety even in highly `unsafe` circumstances, as Rust does in fact offer that. This feature of the language is an exception, as it amounts to allowing the programmer to talk as directly to the machine as possible. So, it is time to explain that we ask for nothing from the assembler, only correct assembling of correct assembly.

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -52,6 +52,16 @@ asm := "asm!(" format_string *("," format_string) *("," operand) [","] ")"
 global_asm := "global_asm!(" format_string *("," format_string) *("," operand) [","] ")"
 ```
 
+## Correctness and Validity
+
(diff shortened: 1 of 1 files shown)


workingjubilee:
cc @Amanieu

Amanieu:
This feels a bit out of place. Could this be moved lower to just before the "rules" section?

workingjubilee:
Sure!

workingjubilee:
Moved and reflowed things a little to be more consistent with the style guide.

Amanieu (approved):

ehuss (approved):
Thanks!

PR #1356, opened 2023-05-06
Add an entry for macro_rules in the "Weak keywords" lexer block

### LINKED ISSUE #1348: Missing lexer token for weak keyword macro_rules
macro_rules is listed as a weak keyword but has no entry in the lexer token table.

mattheww:
Closes #1348

There are no other two-word keywords, so I'm not sure whether `KW_MACRO_RULES` or `KW_MACRORULES` is better; I used the former.


ehuss (approved):
Thanks!

PR #1358, opened 2023-05-10
fixed minor error in a comment that annotated output of a proc macro

Oxyn4:
A comment in the procedural macros chapter had a space that meant it was wrongly annotating what a piece of code would output.


ehuss (approved):
Thanks!

PR #1360, opened 2023-05-17
Fix example code in `impl` docs

eegli:
Tiny fixes for invalid code examples in the [`impl` docs](https://doc.rust-lang.org/reference/types/impl-trait.html#differences-between-generics-and-impl-trait-in-return-position).


ehuss (approved):
Thanks!  Generally we shouldn't be using `ignore` in the reference unless absolutely necessary. I have removed the use of `ignore` and adjusted the examples to match. We also use comments to show elided sections of code.

PR #1361, opened 2023-05-17
linkage: Add a note about symbol visibility and unused sections to the paragraph about staticlibs

sdroege:
Fixes https://github.com/rust-lang/rust/issues/111593

----

CC @bjorn3

### DIFF
--- a/src/linkage.md
+++ b/src/linkage.md
@@ -47,6 +47,16 @@ be ignored in favor of only building the artifacts specified by command line.
   linking Rust code into an existing non-Rust application
   because it will not have dynamic dependencies on other Rust code.
 
+  Note that, because the resulting static library contains all of the
+  dependencies, including the standard library, and also exports all public
+  symbols of them, linking the static library into an executable or shared
+  library will need special care. In case of a shared library the list of
+  exported symbols will have to be limited via e.g. a linker or symbol version
+  script, exported symbols list (macOS) or module definition file (Windows).
+  Additionally, unused sections can be removed to remove all code of
+  dependencies that is not actually used (e.g. `--gc-sections` or
+  `-dead_strip` for macOS).
+
 * `--crate-type=cdylib`, `#![crate_type = "cdylib"]` - A dynamic system
   library will be produced.  This is used when compiling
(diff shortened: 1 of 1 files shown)


ehuss:
@bjorn3 or @petrochenkov  Does the information here look correct to you?

petrochenkov:
Yes, but I'm not sure this belongs to the language reference?
It's all about implementation details.

src/linkage.md
```diff
@@ -47,6 +47,16 @@ be ignored in favor of only building the artifacts specified by command line.
   linking Rust code into an existing non-Rust application
   because it will not have dynamic dependencies on other Rust code.
 
+  Note that, because the resulting static library contains all of the
+  dependencies, including the standard library, and also exports all public
+  symbols of them, linking the static library into an executable or shared
+  library will need special care. In case of a shared library the list of
+  exported symbols will have to be limited via e.g. a linker or symbol version
```
petrochenkov, line 54: I wouldn't say that they *have to* or that special care is *needed*.
It's more like "if you want to optimize the output in a similar way to rustc".

ehuss:
> Yes, but I'm not sure this belongs to the language reference?

Yea, there is a gray area where it is unclear how much regarding crate types and linkage are implementation-specific, and what should be considered part of the language. We haven't really nailed down how to draw those lines (#599 discusses this a bit). I seem to recall thinking that most of this chapter should be moved to the `rustc` guide (at least the compiler-specific parts), but I'm not sure exactly how to make that separation. We also don't have a home for guide-style material like this.

I would love to see a section of the rustc guide (or an entire book!) on linking with Rust. This seems like the kind of content that would live there.

I think for now, without a clear home for this, leaving it here would be fine. It can be moved to a better home in the future.

sdroege:
I agree that the reference seems like the wrong place, but this specific chapter seems to be the best currently existing place for this :)

src/linkage.md
sdroege: In case of a shared library it seems more than an optimization as you're otherwise risking symbol conflicts, but yes.

How about softening it from the (current) "must" to "should"?

traviscross (approved):

PR #1366, opened 2023-06-08
Fix inconsistent formatting of Disambiguating Function Calls example

foresterre:
Before:

inconsistency in formatting before the change

After:

inconsistency in formatting after the change


ehuss (approved):
Thanks!

Issue #1326, opened 2023-02-01
Document unaligned reference restrictions


ehuss:
cc https://github.com/rust-lang/rust/pull/102513

dtolnay:
As part of this, please also delete this outdated warning:

https://github.com/rust-lang/reference/blob/b9ccb0960e5e98154020d4c02a09cc3901bc2500/src/type-layout.md?plain=1#L552-L558

https://github.com/rust-lang/reference/blob/b9ccb0960e5e98154020d4c02a09cc3901bc2500/src/type-layout.md?plain=1#L590

ehuss:
#1367 has removed the warning, but there still needs to be something added about the restriction of creating unaligned references (probably in the pointers page?). Also, I don't think "unaligned" is ever explicitly defined, which probably should be covered.

PR #1367, opened 2023-06-14
Remove obsolete note about soundness hole in type-layout.md

yjhn:
Issue https://github.com/rust-lang/rust/issues/27060 has been resolved, so it is no longer possible to safely create unaligned pointers to packed struct fields.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -549,7 +549,28 @@ The `align` modifier can also be applied on an `enum`.
 When it is, the effect on the `enum`'s alignment is the same as if the `enum`
 was wrapped in a newtype `struct` with the same `align` modifier.
 
-Dereferencing an unaligned pointer is [undefined behavior].
+> Note: References to unaligned fields are not allowed because it is [undefined behavior].
+> When fields are unaligned due to an alignment modifier, consider the following options for using references and dereferences:
+> 
+> ```rust
+> #[repr(packed)]
+> struct Packed {
+>     f1: u8,
+>     f2: u16,
+> }
+> let mut e = Packed { f1: 1, f2: 2 };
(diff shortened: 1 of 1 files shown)


ehuss (commented):
This for opening the PR! It would be good to get this cleaned up.

I was wondering if instead it would be good to include a note here that is more of a reminder about the things to be concerned about? Perhaps something along the following:

----

> **Note**: References to unaligned fields are not allowed because it is [undefined behavior].
> When fields are unaligned due to an alignment modifier, consider the following options for using references and dereferences:
>
> ```rust
> #[repr(packed)]
> struct Packed {
>     f1: u8,
>     f2: u16,
> }
> let mut e = Packed { f1: 1, f2: 2 };
> // Instead of creating a reference to a field, copy the value to a local variable.
> let x = e.f2;
> // Or in situations like `println!` which creates a reference, use braces
> // to change it to a copy of the value.
> println!("{}", {e.f2});
> // Or if you need a pointer, use the unaligned methods for reading and writing
> // instead of dereferencing the pointer directly.
> let ptr: *const u16 = std::ptr::addr_of!(e.f2);
> let value = unsafe { ptr.read_unaligned() };
> let mut_ptr: *mut u16 = std::ptr::addr_of_mut!(e.f2);
> unsafe { mut_ptr.write_unaligned(3) }
> ```

As part of #1326, we also probably need to eventually specify the restrictions about creating unaligned references (probably in the pointers page?). Also, I don't think we ever define what "unaligned" means, which probably should be eventually covered.

I'm not sure if this particular place is the right place to get into all the specifics of what the rules are, though I haven't thought too much about how to organize it.

Also there is a link to 27060 at the bottom of the page that should be removed.

ehuss (approved):
Thanks!

PR #1370, opened 2023-06-21
fix incorrect syntax for type-paths

Equilibris:
The current syntax allows for invalid paths, such as:

- `hello::world::` as the trailing `::` would be caught by the incorrectly mapped option
- `hello::::world` as the first `::` would be caught by the incorrectly mapped option

This PR fixes these ambiguities.


ehuss (approved):
Thanks!

PR #1378, opened 2023-07-04
fix incorrect pattern syntax

Equilibris:
The current reference does not allow some valid tuple patterns. Here are all the patterns that have been added as valid according to the new syntax:

```rust
struct A(u8, u8, u8, u8);
struct B();

match A(1, 1, 1, 1) {
    A(10, 1, ..) => (),
    A(10, .., 10) => (),
    A(..) => (),
}
match B() {
    B() => (),
}

match (1, 1, 1, 1) {
    (10, 1, ..) => (),
    (10, .., 10) => (),
    (..) => (),
}
match (1, 1) {
    (..,) => (), // NOTE: Rust fmt removes the comma
}
``

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -659,7 +659,8 @@ A struct pattern is refutable when one of its subpatterns is refutable.
 > &nbsp;&nbsp; [_PathInExpression_] `(` _TupleStructItems_<sup>?</sup> `)`
 >
 > _TupleStructItems_ :\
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! Can you say more about why it appears that the current grammar isn't handling those cases? _RestPattern_ is a part of the _Pattern_ production, so it should be valid within the existing _TuplePattern_ when substituted for the positions where there is a _Pattern_.

Equilibris:
Ah yes. I missed that. Thanks for informing me.

PR #1379, opened 2023-07-09
Change reference to 'callees' to 'callers'.

gregschmit:
This sentence stuck out to me as odd (emphasis mine):

> ...or the unsafe block could be inside an `unsafe fn` and use its own proof obligations to discharge the proof obligations of its *callees*.

I think it means to say that it would use its own proof obligations to discharge the proof obligations of its *callers*, because it would be the calling functions ("callers") that normally have the proof obligation that the callee has discharged by using an unsafe block.

Am I missing something?


src/unsafe-keyword.md
```diff
@@ -29,7 +29,7 @@ By putting operations into an unsafe block, the programmer states that they have
 Unsafe blocks are the logical dual to unsafe functions:
 where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations have been discharged.
 There are many ways to discharge proof obligations;
-for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and use its own proof obligations to discharge the proof obligations of its callees.
+for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and use its own proof obligations to discharge the proof obligations of its callers.
```
ehuss, line 32: Hm, yea, the use of "callee" doesn't seem right, at least with my understanding in a caller/callee relationship of a function, the callee is the function itself. 

However, I've had to re-read this sentence about a dozen times to try to make sure it is clear in my head and I'm thinking some of my struggle is the use of "its" twice. That makes me need to think a little harder to understand which "it" is being referred to, and the relationship.

I'm wondering, does the following seem clearer to you?

```suggestion
for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and the unsafe block can use the safety requirements the unsafe function imposes on its callers to discharge the proof obligations of the unsafe block.
```

gregschmit: I'm not totally sure that's better, because the last part references `proof obligations of the unsafe block`, when really I believe the sentence is trying to address the `proof obligations of the callers`, and saying that the unsafe block discharges **their** proof obligations. In other words, normally you have (with an `unsafe fn`):

1. Safety requirements of `unsafe fn`, which normally are met by the:
2. Proof obligation of the caller(s)

In the case of using an unsafe block inside an `unsafe fn` we have an additional:

3. Proof obligation of the unsafe block, which discharges the proof obligation of the callers (2).

Also, the unsafe block would not be using `the safety requirements the unsafe function imposes on its callers`; rather, it's using its own proof obligations.

We could replace a couple of the "its" like this:

```suggestion
for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and use the block's own proof obligations to discharge the proof obligations of the `unsafe fn`'s callers.
```

mattheww:
I think this paragraph is talking about the case where the unsafe block contains calls to unsafe functions, and when it says "callees" it's talking about those unsafe functions.

If that's right, "callees" was correct but "its own proof obligations" should instead have referred to the block's parent function's proof obligations.

I suggest writing something like

"the unsafe block could be inside an unsafe fn, in which case the block can use the proof obligations of that function to discharge the proof obligations of any unsafe functions called in the block."

ehuss:
Oh, that does seem to make more sense to me. FWIW, I am not well-versed in formal methods.

cc @RalfJung in case you have a chance to clarify. Does @mattheww's interpretation seem correct?

RalfJung:
> I think it means to say that it would use its own proof obligations to discharge the proof obligations of its callers, because it would be the calling functions ("callers") that normally have the proof obligation that the callee has discharged by using an unsafe block.

How could a block discharge it's callers proof obligations? I don't understand what that would even mean.

`get_unchecked` has a proof obligation: the index must be in-bounds. When you write an `unsafe fn` like
```rust
/// Unsafe precondition: `idx < 4`
unsafe fn silly_example(slice: &[u8], idx: usize) {
  assert!(slice.len() >= 4);
  unsafe {
    slice.get_unchecked(idx);
  }
}
```
then you are using the `idx < 4` (the proof obligation of the surrounding `unsafe fn`) to discharge the "`idx` in bounds" (the proof obligation of the callee).

What @mattheww wrote sounds good.

src/unsafe-keyword.md
```diff
@@ -29,7 +29,7 @@ By putting operations into an unsafe block, the programmer states that they have
 Unsafe blocks are the logical dual to unsafe functions:
 where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations have been discharged.
```
RalfJung, line 30:
```suggestion
where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations of functions or operations called inside the block have been discharged.
```
Might be good to add something here as well?

gregschmit:
> > I think it means to say that it would use its own proof obligations to discharge the proof obligations of its callers, because it would be the calling functions ("callers") that normally have the proof obligation that the callee has discharged by using an unsafe block.
> 
> How could a block discharge it's callers proof obligations? I don't understand what that would even mean. callers discharge callee's proof obligations, not vice versa. (At least when talking about preconditions, which are usually the proof obligations we talk about here.)
> 
> `get_unchecked` has a proof obligation: the index must be in-bounds. When you write an `unsafe fn` like
> 
> ```rust
> /// Unsafe precondition: `idx < 4`
> unsafe fn silly_example(slice: &[u8], idx: usize) {
>   assert!(slice.len() >= 4);
>   unsafe {
>     slice.get_unchecked(idx);
>   }
> }
> ```
> 
> then you are using the `idx < 4` (the proof obligation of the surrounding `unsafe fn`) to discharge the "`idx` in bounds" (the proof obligation of the callee).
> 
> What @mattheww wrote sounds good.

Ohh! Then in that case I totally misunderstood the sentence. I was mistaken, thinking it was trying to communicate that the unsafe block within the `unsafe fn` was discharging the proof obligations of the callers. Yeah your last edit makes it clearer IMO.

src/unsafe-keyword.md
```diff
@@ -27,9 +27,9 @@ this can be changed by enabling the [`unsafe_op_in_unsafe_fn`] lint.
...
 
 Unsafe blocks are the logical dual to unsafe functions:
-where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations have been discharged.
+where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations of functions or operations called inside the block have been discharged.
 There are many ways to discharge proof obligations;
-for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and use its own proof obligations to discharge the proof obligations of its callees.
+for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn`, in which case the block can discharge the proof obligations of any unsafe functions called inside the block.
```
gregschmit, line 32: Ok, so now when I read it now after I integrated those changes, it's much more clear to me what proof obligations are being met with an unsafe block inside an `unsafe fn`. The only thing I still see what stands out is that it seems like those proof obligations would exist whether or not the unsafe block was inside an unsafe function, so what really oriented my thinking was the prefix phrase `or the unsafe block could be inside an unsafe fn...`, which seemed to indicate that when the unsafe block is inside an unsafe fn that it has a special role in discharging certain proof obligations, and that's why my mind went to "oh maybe it means that the unsafe block can discharge some or all of the proof obligations of the `unsafe fn` it is residing in". I think I understand that to NOT be the case, so I'm wondering if the prefix phrase should be removed so it would say:

```suggestion
for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or an unsafe block can discharge the proof obligations of any unsafe functions called inside the block.
```

gregschmit: Or maybe, in other words: does the requirement that an unsafe block discharge any proof obligations of unsafe functions that it calls change if the block exists within a safe or unsafe function?

mattheww: No, that requirement doesn't change.

If you put the words "use the proof obligations of that function to" back then I think the reference to the parent unsafe function will make more sense.

RalfJung: >  so what really oriented my thinking was the prefix phrase or the unsafe block could be inside an unsafe fn..., which seemed to indicate that when the unsafe block is inside an unsafe fn that it has a special role in discharging certain proof obligations

The point was that inside an unsafe fn, there are extra preconditions we can use. When the unsafe block is in a regular safe fn, this is not the case -- the callee's preconditions have to be discharged based on whatever the function does itself, and private invariants, but it cannot "propagate" the requirements further outwards.

> an unsafe block can discharge the proof obligations of any unsafe functions called inside the block

It *must* discharge those proof obligations! The question is where the information to discharge the obligations comes from:
- local checks
- data structure invariants
- (only possible if this is inside an `unsafe fn`) preconditions of the current function

RalfJung: You removed the crucial part from @mattheww's proposal! "use the proof obligations of that function" was load-bearing, after removing it this sentence now doesn't make a ton of sense any more.

gregschmit: Ok, I think I understand better now. It's using the proof obligations of the `unsafe fn` to discharge the proof obligations of other unsafe functions called in an unsafe block within the original `unsafe fn`, which wouldn't make any sense if the original function was safe, because then it would have no proof obligations. I understand why that part needs to be there now. I've added that phrase back in.

src/unsafe-keyword.md
```diff
@@ -27,9 +27,9 @@ this can be changed by enabling the [`unsafe_op_in_unsafe_fn`] lint.
...
 
 Unsafe blocks are the logical dual to unsafe functions:
-where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations have been discharged.
+where unsafe functions define a proof obligation that callers must uphold, unsafe blocks state that all relevant proof obligations of functions or operations called inside the block have been discharged.
 There are many ways to discharge proof obligations;
-for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn` and use its own proof obligations to discharge the proof obligations of its callees.
+for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn`, in which case the block can use the proof obligations of that function to discharge the proof obligations of any unsafe functions called inside the block.
```
RalfJung, line 32:
```suggestion
for example, there could be run-time checks or data structure invariants that guarantee that certain properties are definitely true, or the unsafe block could be inside an `unsafe fn`, in which case the block can use the proof obligations of that function to discharge the proof obligations arising inside the block.
```

RalfJung:
Looking good, thanks for bringing this up and working with us on a less confusing wording!

@bors r+ rollup

RalfJung:
... oh, this is the reference. Let's wait for @ehuss to merge this then. :)

Issue #579, opened 2019-04-21
Improve panic documentation.


ehuss:
Panicking is almost completely undocumented. How much of this is relevant to the reference and how much is implementation-specific? At the very least, something should define what "panic" is, since it is referred to in a few places. I feel like there is lots to be said about panicking, but it's hard for me to say specifically.

One example is to document panic behavior with extern functions: https://github.com/rust-lang/rust/pull/55982/.

PR #1382, opened 2023-07-17
Say that division by zero for primitive types panics

### LINKED ISSUE #1375: Specify that integer division by zero panics
While there is a [section on overflow behavior](https://github.com/rust-lang/reference/blob/3b6313a2141fcbbf9d47788cef6006995b753543/src/expressions/operator-expr.md#overflow), which discusses that integer overflows will trigger a panic in debug mode, the [section "Arithmetic and Logical Binary Operators"](https://github.com/rust-lang/reference/blob/3b6313a2141fcbbf9d47788cef6006995b753543/src/expressions/operator-expr.md#arithmetic-and-logical-binary-operators) doesn't specify what happens when an integer division by zero is encountered.

There is a note in the standard library documentation regarding the `impl std::ops::Div<i32> for i32` ([source](https://github.com/rust-lang/rust/blob/de223888735d79a77b38c74c52bde64add244212/library/core/src/ops/arith.rs#L464-L487)), which states that a division by zero panics (irregardless of whether being in debug mode or not):

> This operation will panic if `other == 0` or the division results in overflow.

However, as [discussed on URLO](https://users.rust-lang.org/t/guarantee-that-0i32-0-panics/96190/19?u=jbe), this only seems to guarantee that `0.div(0)` panics, but it (formally) doesn't make any statement regarding the primitive operation `0 / 0`.

The section "Arithmetic and Logical Binary Operators" should be extended and state that integer division by zero panics (irregardless of whether being compiled with debug mode).

mattheww:
Closes #1375

Explicitly say that division by zero for primitive types panics (for both `/` and `%`).

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -243,8 +243,8 @@ The operands of all of these operators are evaluated in [value expression contex
 | `+`    | Addition                |               | Addition       | `std::ops::Add`    | `std::ops::AddAssign`                 |
 | `-`    | Subtraction             |               | Subtraction    | `std::ops::Sub`    | `std::ops::SubAssign`                 |
 | `*`    | Multiplication          |               | Multiplication | `std::ops::Mul`    | `std::ops::MulAssign`                 |
-| `/`    | Division*               |               | Division       | `std::ops::Div`    | `std::ops::DivAssign`                 |
-| `%`    | Remainder**             |               | Remainder      | `std::ops::Rem`    | `std::ops::RemAssign`                 |
+| `/`    | Division*†              |               | Division       | `std::ops::Div`    | `std::ops::DivAssign`                 |
+| `%`    | Remainder**†            |               | Remainder      | `std::ops::Rem`    | `std::ops::RemAssign`                 |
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks! 

Just a random thought is that const-eval works differently (it is a compile-time error, though I'm not sure if that is always the case), so that would need to be documented differently, though I'm not sure how.

mattheww:
Treat that as part of #579?

ehuss:
I was roughly thinking that it could go into const-eval chapter, though I guess it already discusses this ("Behaviors such as out of bounds array indexing or overflow are compiler errors").  There's probably more rigor that could be applied to the const-eval chapter to exhaustively cover its specific behavior, but I think for now it is probably sufficient.

JanBeh:
Not all primitive type panic on division by zero.

```
fn main() {
    println!("{}", 1.0 / 0.0);
}
```

This gives infinity. ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c92466c532da7ae9b37fa196a35e9aa5))

PR #1384, opened 2023-07-19
Operator expressions: make the note about division by zero clearer.

mattheww:
Say that the note applies to integer division only (compare the note about rounding to zero).

Addresses #1375


ehuss (approved):
Thanks!

PR #1387, opened 2023-07-31
replace 'UB on raw ptr deref' with UB on place projection/access

RalfJung:
As discussed by @rust-lang/opsem in https://github.com/rust-lang/opsem-team/issues/9: instead of saying that `*ptr` is insta-UB for "bad" pointers, only have UB when the place is actually accessed, and when place projections go out-of-bounds.

This makes the following code (which currently has UB) legal:
```rust
let ptr = 23 as *const i32;
// `ptr` is not aligned and doesn't point to an actual allocation.
// Still, this is allowed:
let ptr = addr_of!(*ptr);
// Even with a null pointer, this is allowed:
let ptr = addr_of!(*(0 as *const i32));
```
However, the following code is still UB:
```rust
// Accessing (loading from or storing to) a place that is dangling or unaligned
let mem = 0u32;
let odd_ptr = addr_of!(mem).byte_add(1).cast::<u16>();
let _val = *odd_ptr; // pointer is 1-aligned but must be 2-aligned
```
```rust
// Performing a place projection that violates the requirements of in-bounds pointer arithmetic
let ptr = 23 as *const (i32, i32);
let ptr = addr_of!((*ptr).1); // pointer arithmetic not inside the bounds of an allocation
```

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -23,7 +23,8 @@ code.
 </div>
 
 * Data races.
-* Accessing (loading from or storing to) a place that is [dangling] or unaligned.
+* Accessing (loading from or storing to) a place that is [dangling] or [based on
+  a misaligned pointer].
 * Performing a place projection that violates the requirements of [in-bounds
   pointer arithmetic][offset]. A place projection is a [field
   expression][project-field], a [tuple index expression][project-tuple], or an
@@ -66,7 +67,7 @@ code.
   * A `!` (all values are invalid for this type).
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
-  * A reference or `Box<T>` that is [dangling], unaligned, or points to an invalid value.
+  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
   * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
     * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
       `Trait` that matches the actual dynamic trait the pointer or reference points to.
@@ -99,6 +100,36 @@ reading uninitialized memory is permitted are inside `union`s and in "padding"
 ### Pointed-to bytes
 
 The span of bytes a pointer or reference "points to" is determined by the pointer value and the size of the pointee type (using `size_of_val`).
+
+### Places based on misaligned pointers
+[based on a misaligned pointer]: #places-based-on-misaligned-pointers
+
+A place is said to be "based on a misaligned pointer" if the last `*` projection
+during place computation was performed on a pointer that was not aligned for its
+type. (If there is no `*` projection in the place expression, then this is
+accessing the field of a local and rustc will guarantee proper alignment. If
+there are multiple `*` projection, then each of them incurs a load of the
+pointer-to-be-dereferenced itself from memory, and each of these loads is
+subject to the alignment constraint. Note that some `*` projections can be
+omitted in surface Rust syntax due to automatic dereferencing; we are
+considering the fully expanded place expression here.)
+
+For instance, if `ptr` has type `*const S` where `S` has an alignment of 8, then
+`ptr` must be 8-aligned or else `(*ptr).f` is "based on an misaligned pointer".
+This is true even if the type of the field `f` is `u8` (i.e., a type with
+alignment 1). In other words, the alignment requirement derives from the type of
+the pointer that was dereferenced, *not* the type of the field that is being
+accessed.
+
+Note that a place based on a misaligned pointer only leads to Undefined Behavior
+when it is loaded from or stored to. `addr_of!`/`addr_of_mut!` on such a place
+is allowed. `&`/`&mut` on a place requires the alignment of the field type (or
+else the program would be "producing an invalid value"), which generally is a
+less restrictive requirement than being based on an aligned pointer. Taking a
+reference will lead to a compiler error in cases where the field type might be
+more aligned than the type that contains it, i.e., `repr(packed)`. This means
+that being based on an aligned pointer is always sufficient to ensure that the
+new reference is aligned, but it is not always necessary.
 
 ### Dangling pointers
 [dangling]: #dangling-pointers
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -23,9 +23,11 @@ code.
 </div>
 
 * Data races.
-* Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
-  [dangling] or unaligned, even in [place expression context]
-  (e.g. `addr_of!(*expr)`).
+* Accessing (loading from or storing to) a place that is [dangling] or unaligned.
+* Performing a place projection that violates the requirements of [in-bounds
+  pointer arithmetic][offset]. A place projection is a [field
+  expression][project-field], a [tuple index expression][project-tuple], or an
+  [array/slice index expression][project-slice].
 * Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
   LLVM’s scoped [noalias] model, except if the `&T` contains an
   [`UnsafeCell<U>`]. References and boxes must not be [dangling] while they are
@@ -124,8 +126,11 @@ must never exceed `isize::MAX`.
 [Rustonomicon]: ../nomicon/index.html
 [`NonNull<T>`]: ../core/ptr/struct.NonNull.html
 [`NonZero*`]: ../core/num/index.html
-[dereference expression]: expressions/operator-expr.md#the-dereference-operator
 [place expression context]: expressions.md#place-expressions-and-value-expressions
 [rules]: inline-assembly.md#rules-for-inline-assembly
 [points to]: #pointed-to-bytes
 [pointed to]: #pointed-to-bytes
+[offset]: ../std/primitive.pointer.html#method.offset
+[project-field]: expressions/field-expr.md
+[project-tuple]: expressions/tuple-expr.md#tuple-indexing-expressions
+[project-slice]: expressions/array-expr.md#array-and-slice-indexing-expressions


RalfJung:
I am not sure if this counts as a "small" decision that t-opsem can just make without t-lang involvement, so I'll add both teams. This is an irreversible decision after all and a change to the core "list of UB", that seems quite fundamental.

The justification for why we think this first example should not be UB is mostly that many people are confused about the entire place expression business, so they don't realize `addr_of!(*ptr)` is "dereferencing" a pointer. There's also not a clear benefit from making it UB -- our current codegen is already completely compatible with this change, the UB we are removing is only recognized by Miri but never exploited by the compiler.

The justification for keeping the later 2 examples UB is that our current codegen actually *does* exploit both of them, so allowing them comes at a cost. The first example should be rather uncontroversial; the fact that every access has to be in-bounds and aligned has never been up for debate. The second example is probably less clear, but we do want to keep using `getelementptr inbounds` for place projections, so we have to make this UB. (Maybe we will come back to this later and weaken place projections from "inbounds" to "nowrap". But for now we don't see a good motivation for doing this, and without a good motivation it is unlikely someone is going to put in the work of implementing those semantics in the codegen backend.)

@rfcbot fcp merge

JakobDegen:
Fwiw I don't expect this example to have needed T-lang signoff, but there's no harm.

@rfcbot reviewed

saethlin:
@rfcbot reviewed

RalfJung:
https://github.com/rust-lang/rust/pull/114330 implements this change in Miri. I found some other interesting cases that are allowed now.

A mere "place mention" of a dangling place is no longer UB -- that is in line with `addr_of!` no longer being UB here:
```rust
fn deref_invalid_underscore() {
    unsafe {
        let _ = *ptr::invalid::<i32>(0);
        let _ = *ptr::invalid::<i32>(1);
    }
}
```

If `xptr` only partially points to memory, and `(*xptr).1` accesses the inbounds part, then this is now allowed -- both the place projection (pointer arithmetic) and the load from that place are fully inbounds, so there's nothing to cause UB here:
```rust
fn deref_partially_dangling() {
    let x = (1, 13);
    let xptr = &x as *const _ as *const (i32, i32, i32);
    let val = unsafe { (*xptr).1 };
    assert_eq!(val, 13);
}
```

lukas-code:
> ```rust
> fn deref_partially_dangling() {
>     let x = (1, 13);
>     let xptr = &x as *const _ as *const (i32, i32, i32);
>     let val = unsafe { (*xptr).1 };
>     assert_eq!(val, 13);
> }
> ```

Have we decided that the order of the fields in the tuple `(i32, i32, i32)` is guaranteed? In the [UCG](https://rust-lang.github.io/unsafe-code-guidelines/layout/structs-and-tuples.html#tuple-types) it specifically says that tuple fields can be reordered, such that `.1` could actually be the third field in memory and out-of-bounds.

RalfJung:
No we have not decided that, the proper version of that test would use a `repr(C)` struct. I was just lazy...

scottmcm:
This makes good sense to me.  Having `addr_of!(*p)` be a no-op seems like what most people would expect.

digama0:
@rfcbot reviewed

CAD97:
@rfcbot reviewed

tmandry:
> This makes good sense to me. Having `addr_of!(*p)` be a no-op seems like what most people would expect.

I agree with this and don't have any blocking concerns over this change specifically. But I am concerned that this will lead users to a mental model where place projections inside `addr_of!` (the last example in the description) are also a no-op. I found it surprising that the latter example is UB if the first example is not.

At a high level I would like us to strongly prefer simple rules for what is and isn't UB, wherever possible, and justify complexity as either inherently necessary or temporary. "`*foo` is insta-UB for bad pointers" is quite simple, despite being overly restrictive. "Only if actually accessed" is quite simple. "Only if actually accessed or a place projection goes out of bounds" is less so.

From the description it sounded like this condition might be temporary (and is due to a current limitation in LLVM), but if we could wave a magic wand and fix the LLVM limitation, are we sure we would?

saethlin:
> I am concerned that this will lead users to a mental model...

I agree, generally, but I think this is a problem of documentation. Currently we have essentially no documentation for what valid uses of `addr_of!` are. Our documentation for this currently just punts to "the usual rules" while never defining that as a means to avoid specifying what uses are and aren't valid, because we haven't made decisions about the rules such as the one that this PR is about.

I believe the documentation for `addr_of!` is a good place to head off incorrect mental models, and that at the moment the documentation does precious little.

scottmcm:
> "Only if actually accessed or a place projection goes out of bounds" is less so.

With any luck `offset_of!` will help here, since we'll be able to talk about how
`addr_of!((*p).foo)` is `p.byte_add(offset_of!(Type, foo)).cast()`, and how `p.wrapping_byte_add(offset_of!(Type, foo)).cast()` is available for cases where you don't want the UB from `GEPi`.

RalfJung:
> At a high level I would like us to strongly prefer simple rules for what is and isn't UB, wherever possible, and justify complexity as either inherently necessary or temporary. "*foo is insta-UB for bad pointers" is quite simple, despite being overly restrictive. 

Those are the rules we currently have. I also think they are simpler. But many, many people have been caught off-guard by the fact that `*foo` itself can be UB, even in a context like `addr_of!(*foo)`. Yes, our docs are bad, but at this point I think we also have enough evidence that people strongly expect  `addr_of!(*foo)` to never be UB, and it might be better to align the model with people's expectations -- even if that nominally makes the model more complicated, and given in particular that the extra UB does not actually seem that useful.

> "Only if actually accessed" is quite simple. "Only if actually accessed or a place projection goes out of bounds" is less so.

You just have to separate alignment and inbounds:
- alignment is required when a load actually happens (place projections do not matter). What exactly the required alignment is is still being discussed, see https://github.com/rust-lang/opsem-team/issues/11.
- inbounds is required for every single place projection.

src/behavior-considered-undefined.md
```diff
@@ -23,9 +23,11 @@ code.
 </div>
 
 * Data races.
-* Evaluating a [dereference expression] (`*expr`) on a raw pointer that is
-  [dangling] or unaligned, even in [place expression context]
-  (e.g. `addr_of!(*expr)`).
+* Accessing (loading from or storing to) a place that is [dangling] or unaligned.
```
RalfJung, line 26: The "unaligned" here is doing unexpected amounts of work. Cc https://github.com/rust-lang/opsem-team/issues/11

RalfJung:
> From the description it sounded like this condition might be temporary (and is due to a current limitation in LLVM), but if we could wave a magic wand and fix the LLVM limitation, are we sure we would?

If this is talking about the requirement that field projections follow the rules of the `offset` method, then this is intended to be permanent. I can't imagine a good reason for why we'd want the two to behave differently.

We considered weakening the requirements of `offset` from `inbounds` to `nowrap`. However, we don't have any motivating usecase for this -- as in, we don't know any real code that people might write that would benefit from weakening the requirement. (The one exception is manual `offset_of` implementations but we consider that solved by the language-native macro.) Without a usecase, it's not worth trying to convince LLVM to add support for this attribute. Also, this weakening causes some trouble for `offset` as a `const fn` -- basically, in a `const fn` we cannot know if the pointer overflows or not since we don't know where in the address space it lies, so we have to still require `inbounds` there.

src/behavior-considered-undefined.md
RalfJung: Since it doesn't look like we will quickly get consensus to reduce the alignment requirement here, I have added an explanation to the PR:

> The alignment requirement of a place is computed based on the entire place expression since the last `*` (if any).
> If the place was computed via `(*ptr).field.field`, then the final place is considered aligned only if `ptr` is aligned *according to its type*.

This leaves us the option to keep rejecting all the examples in https://github.com/rust-lang/opsem-team/issues/11.

RalfJung:
@rfcbot fcp cancel

I was asked to re-start FCP since I made substantial changes to the PR text after starting initial FCP -- specifically, I originally forgot to define when a load from a place is considered misaligned, and the definition is not obvious at all.

RalfJung:
So, let's go again: @rfcbot fcp merge

The summary at the top is still accurate. For alignment, the text now says that it is UB to access (loading from or storing to) a place that is dangling or based on a misaligned pointer. The latter is defined as follows:

> A place is said to be "based on a misaligned pointer" if the last `*` projection
during place computation was performed on a pointer that was not aligned for its
type. (If there is no `*` projection in the place expression, then this is
accessing the field of a local and rustc will guarantee proper alignment.)
> 
> For instance, if `ptr` has struct type `S` with a field `f` of type `u8`, but
the alignment of `S` is 8, then `ptr` must be 8-aligned or else `(*ptr).f` is
"based on an misaligned pointer". In other words, the alignment requirement
derives from the type of the pointer that was dereferenced, *not* the type of
the field that is being accessed.

This effectively means that every load/store that is currently UB due to misalignment remains UB due to misalignment. (However, `addr_of!(...)` incurs no alignment requirement; the requirement only arises if an actual load/store happens.) If you had expected this to just say that we have UB if the load/store occurs from a place that does not match the alignment of its type, consider packed fields -- that definition would render loads from `struct.packed_field` UB, which we surely want to avoid. We could do something clever based on the projections that occurred since the "base", but in the t-opsem meeting we couldn't actually find a usecase for allowing any of the code that would allow, and it was pointed out that the extra alignment could be quite useful e.g. when accessing fields of `repr(aligned)` structs (imagine a case like https://github.com/rust-lang/rust/issues/54028 but using raw pointers). Additionally the proposed spec leads (in my opinion) to a simpler implementation in both MiniRust and Miri.

I also want to use this opportunity to clarify that there is *no* alignment constraint on offset projections. Those require the preconditions of `ptr::offset` but nothing else. Only loads/stores require alignment (and creating a reference requires the alignment of the pointee type).

scottmcm:
Triggering the bot on behalf of Ralf's summary https://github.com/rust-lang/reference/pull/1387#issuecomment-1693462092

@rfcbot fcp merge

RalfJung:
(The bot only checks this repo once an hour via a cronjob. I hope it won't get confused by the cancel + new-FCP...)

RalfJung:
Looks like it only got the first message...
@rfcbot fcp merge

digama0:
@rfcbot reviewed

JakobDegen:
Hmm, I'm not convinced that these are the semantics we want in the long term. But switching to the version discussed in the meeting on Tuesday is forwards compatible, so I'm fine with this FCP as it is now

@rfcbot reviewed

RalfJung:
(I see @scottmcm checked their box twice and the bot removed it again each time... I tried that again, but @scottmcm you might have to use the `rfcbot reviewed` command instead.)

RalfJung:
@CAD97 @joshtriplett @nikomatsakis @pnkfelix @saethlin friendly reminder that this FCP is waiting for your checkbox. :) Most of you have already approved a previous version of this FCP, but I unfortunately forgot to properly define when a load/store is sufficiently aligned -- and further discussion since the creation of this PR has lead to changes in what the proper definition is.  See [here](https://github.com/rust-lang/reference/pull/1387#issuecomment-1693462092) for more detail.

CAD97:
Missed the re-FCP, whoops. The current wording looks good and fairly straightforward (if loading from a dereference, the dereferenced pointer must be aligned for its type), keeping in mind that we do reserve the right to make more things defined in the future. If you want to only check the alignment of the final place instead, this will allow you to write `*addr_of!(expr)`, which is straightforward if a somewhat awkward seeming no-op (and problematic for `repr(packed)`).

It might be worth also calling out that `&(*ptr).field` only requires alignment of the field, not the pointer, given this formulation.

@rustbot reviewed

src/behavior-considered-undefined.md
```diff
@@ -98,6 +101,24 @@ reading uninitialized memory is permitted are inside `union`s and in "padding"
...
+### Places based on misaligned pointers
+[based on a misaligned pointer]: #places-based-on-misaligned-pointers
+
+A place is said to be "based on a misaligned pointer" if the last `*` projection
+during place computation was performed on a pointer that was not aligned for its
+type. (If there is no `*` projection in the place expression, then this is
+accessing the field of a local and rustc will guarantee proper alignment. If
+there are multiple `*` projection, then each of them incurs a load of the
+pointer-to-be-dereferenced itself from memory, and each of these loads is
+subject to the alignment constraint.)
```
CAD97, line 113: It might be worthwhile to add a note here that e.g.

> `*` projections can be implicitly introduced even without a literal source `*` when references are involved.

RalfJung: References are always aligned though so this cannot lead to UB.

RalfJung:
> It might be worth also calling out (for this thread, not necessarily in the book) that &(*ptr).field also only requires alignment of the field, not the pointer, given this formulation — the reference needs to be aligned for its type, but no load/store is done to check the pointer derived from being aligned. (The same also goes for (*ptr).field.method_by_ref().)

I've added this to the reference now, and only then saw you mean this as clarification for the PR only. I can remove it again in case people think it is a distraction.

> Note that a place based on a misaligned pointer only leads to Undefined Behavior
when it is loaded from or stored to. `addr_of!`/`addr_of_mut!` on such a place
is allowed. `&`/`&mut` on a place requires the alignment of the field type (or
else the program would be "producing an invalid value"), which generally is a
less restrictive requirement than being based on an aligned pointer. Taking a
reference will lead to a compiler error in cases where the field type might be
more aligned than the type that contains it, i.e., `repr(packed)`. This means
that being based on an aligned pointer is always sufficient to ensure that the
new reference is aligned, but it is not always necessary.

src/behavior-considered-undefined.md
CAD97: So if you have `(*misaligned).tref.field`, that's actually `(*(*misaligned).tref).field` after desugar due to reference autoderef, so it's the multiple `*` projections case and the misaligned pointer is loaded from, causing UB. It's not surprising after you consider the desugar, but I think it's relevant to note here since the condition of only being a single `*` projection is not purely syntactical in the surface language when references may be involved.

CAD97:
That added note is better worded than any I could think of, so I think it fits well in context. I only said clarification primarily for the PR because I didn't think the text *needed* the clarification, but I do think it's now better for it.

src/behavior-considered-undefined.md
RalfJung: `(*misaligned).tref.field` looks like it requires `misaligned` to be aligned. `(*(*misaligned).tref).field` looks like it requires `misaligned` and `tref` to be aligned, but `tref` is a reference so there is already very explicit documentation saying it has to always be aligned.

I don't see how there is a chance of missing UB here.

RalfJung: Are you maybe referring to a different part of the diff, where it says "Accessing (loading from or storing to) a place ..."? Indeed I haven't defined this entirely in terms of surface syntax, I am relying on other parts of the documentation saying when place accesses happen.

RalfJung:
> @rustbot reviewed

That was the wrong bot, I am taking the liberty to set the checkmark.

saethlin:
The change to the text now has a sort of delayed UB upon read/write if the pointer deref in the place was invalid due to alignment, but what about inbounds?

```rust
struct Thing {
    head: u8,
    tail: u8,
}

fn main() {
    let mem = 0u8;
    let ptr = &mem as *const u8 as *const Thing;
    unsafe {
        let head = (*ptr).head;
    }    
}
```
Is the read in this example UB? This code reads from a place which is "based on" an out-of-bounds pointer dereference, but it is out-of-bounds by the pointer dereference, not out-of-bounds by the place projection. The text change here makes it clear that it is UB to go out-of-bounds by a field projection, but not whether you can dereference out of bounds then by field projection end up back to something inbounds.

And then if that is UB, what if we do this:
```rust
        let head: &u8 = &(*ptr).head;
```

saethlin:
@rfcbot reviewed

RalfJung:
FCP passed, @ehuss let me know what needs to be dine for landing this. :)

ehuss (approved):
Thanks!

PR #1392, opened 2023-08-11
Specify bit validity and padding of some types

### LINKED ISSUE #1291: Describe bit validity and padding for primitive types

joshlf:
Specify the bit validity and padding of the primitive numeric  types, bool, char, and pointer and reference types.

Closes #1291

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -58,11 +58,11 @@ aligned to 32 bits.
 
 For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
 `i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every bit pattern
-represents a valid instance of the type (in other words,
-`transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the primitive
-numeric types and also for `bool` and `char`, every byte is guaranteed to be
-initialized (in other words, `transmute::<T, [u8; size_of::<T>()]>(...)` is always
-sound).
+represents a valid instance of the type (in other words, for every primitive numeric
+type, `T`, `transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the
+primitive numeric types and also for `bool` and `char`, every byte is guaranteed to be
+initialized (in other words, for every such type, `T`,
+`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound).
 
 ## Pointers and References Layout
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -64,9 +64,9 @@ a valid `u8`. A byte at any offset in a reference or pointer type may not be a
 valid `u8` (the semantics of transmuting a reference or pointer to a non-pointer
 type is currently undecided).
 
-For the primitive numeric types and also for `bool` and `char`, every byte is
+For `bool` and `char`, every byte is
 guaranteed to be initialized (in other words, for every such type, `T`,
-`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound).
+`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound -- but the inverse is not).
 
 ## Pointers and References Layout
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -60,9 +60,9 @@ For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
 `i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
 of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. `u8` has 256
 valid representations (namely, every 8-bit sequence). An uninitialized byte is not
-a valid u8. A byte at any offset in a reference or pointer type may not be a valid
-u8 (the semantics of transmuting a reference or pointer to a non-pointer type is
-currently undecided).
+a valid `u8`. A byte at any offset in a reference or pointer type may not be a
+valid `u8` (the semantics of transmuting a reference or pointer to a non-pointer
+type is currently undecided).
 
 For the primitive numeric types and also for `bool` and `char`, every byte is
 guaranteed to be initialized (in other words, for every such type, `T`,
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -61,7 +61,7 @@ For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
 represents a valid instance of the type (in other words,
 `transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the primitive
 numeric types and also for `bool` and `char`, every byte is guaranteed to be
-initialized (in other words, `transmute::<T, [u8; size_of::<T>()]>(...) is always
+initialized (in other words, `transmute::<T, [u8; size_of::<T>()]>(...)` is always
 sound).
 
 ## Pointers and References Layout
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -56,12 +56,16 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
-For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
-`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every bit pattern
-represents a valid instance of the type (in other words, for every primitive numeric
-type, `T`, `transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the
-primitive numeric types and also for `bool` and `char`, every byte is guaranteed to be
-initialized (in other words, for every such type, `T`,
+For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
+of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. `u8` has 256
+valid representations (namely, every 8-bit sequence). An uninitialized byte is not
+a valid u8. A byte at any offset in a reference or pointer type may not be a valid
+u8 (the semantics of transmuting a reference or pointer to a non-pointer type is
+currently undecided).
+
+For the primitive numeric types and also for `bool` and `char`, every byte is
+guaranteed to be initialized (in other words, for every such type, `T`,
 `transmute::<T, [u8; size_of::<T>()]>(...)` is always sound).
 
 ## Pointers and References Layout
@@ -69,10 +73,7 @@ initialized (in other words, for every such type, `T`,
 Pointers and references have the same layout. Mutability of the pointer or
 reference does not change the layout.
 
-Pointers to sized types have the same size and alignment as `usize`. Every
-byte of a pointer to a sized type and of a reference to a sized type is
-initialized (in other words, for such a pointer or reference type, `P`,
-`transmute::<P, [u8; size_of::<P>()]>(...)` is always sound).
+Pointers to sized types have the same size and alignment as `usize`.
 
 Pointers to unsized types are sized. The size and alignment is guaranteed to be
 at least equal to the size and alignment of a pointer.
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -56,17 +56,6 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
-For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
-`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
-of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. An
-uninitialized byte is not a valid `u8`. A byte at any offset in a reference or
-pointer type may not be a valid `u8` (the semantics of transmuting a reference or
-pointer to a non-pointer type is currently undecided).
-
-For `bool` and `char`, every byte is
-guaranteed to be initialized (in other words, for every such type, `T`,
-`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound -- but the inverse is not).
-
 ## Pointers and References Layout
 
 Pointers and references have the same layout. Mutability of the pointer or
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -58,11 +58,10 @@ aligned to 32 bits.
 
 For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
 `i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
-of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. `u8` has 256
-valid representations (namely, every 8-bit sequence). An uninitialized byte is not
-a valid `u8`. A byte at any offset in a reference or pointer type may not be a
-valid `u8` (the semantics of transmuting a reference or pointer to a non-pointer
-type is currently undecided).
+of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. An
+uninitialized byte is not a valid `u8`. A byte at any offset in a reference or
+pointer type may not be a valid `u8` (the semantics of transmuting a reference or
+pointer to a non-pointer type is currently undecided).
 
 For `bool` and `char`, every byte is
 guaranteed to be initialized (in other words, for every such type, `T`,
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -56,12 +56,23 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
+For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every bit pattern
+represents a valid instance of the type (in other words,
+`transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the primitive
+numeric types and also for `bool` and `char`, every byte is guaranteed to be
+initialized (in other words, `transmute::<T, [u8; size_of::<T>()]>(...) is always
+sound).
+
 ## Pointers and References Layout
 
 Pointers and references have the same layout. Mutability of the pointer or
 reference does not change the layout.
 
-Pointers to sized types have the same size and alignment as `usize`.
+Pointers to sized types have the same size and alignment as `usize`. Every
+byte of a pointer to a sized type and of a reference to a sized type is
+initialized (in other words, for such a pointer or reference type, `P`,
+`transmute::<P, [u8; size_of::<P>()]>(...)` is always sound).
 
 Pointers to unsized types are sized. The size and alignment is guaranteed to be
 at least equal to the size and alignment of a pointer.
--- a/src/types/pointer.md
+++ b/src/types/pointer.md
@@ -50,6 +50,12 @@ Raw pointers can be created directly using [`core::ptr::addr_of!`] for `*const` 
 
 The standard library contains additional 'smart pointer' types beyond references and raw pointers.
 
+## Bit validity
+
+Despite pointers and references being similar to `usize`s in the machine code emitted on most platforms,
+the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
+Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
+
 [`core::ptr::addr_of!`]: ../../core/ptr/macro.addr_of.html
 [`core::ptr::addr_of_mut!`]: ../../core/ptr/macro.addr_of_mut.html
 [Interior mutability]: ../interior-mutability.md
--- a/src/types/pointer.md
+++ b/src/types/pointer.md
@@ -56,6 +56,10 @@ Despite pointers and references being similar to `usize`s in the machine code em
 the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
 Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
 
+For thin raw pointers (i.e., for `<T as Pointee>::Metadata == ()` and `P = *const T` or `P = *mut T`),
+the inverse direction (transmuting from an integer or array of integers to `P`) is always valid.
+However, the pointer produced via such a transmutation may not be dereferenced (not even if `T` has size zero).
+
 [`core::ptr::addr_of!`]: ../../core/ptr/macro.addr_of.html
 [`core::ptr::addr_of_mut!`]: ../../core/ptr/macro.addr_of_mut.html
 [Interior mutability]: ../interior-mutability.md
--- a/src/types/pointer.md
+++ b/src/types/pointer.md
@@ -56,7 +56,7 @@ Despite pointers and references being similar to `usize`s in the machine code em
 the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
 Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
 
-For thin raw pointers (i.e., for `<T as Pointee>::Metadata == ()` and `P = *const T` or `P = *mut T`),
+For thin raw pointers (i.e., for `P = *const T` or `P = *mut T` for `T: Sized`),
 the inverse direction (transmuting from an integer or array of integers to `P`) is always valid.
 However, the pointer produced via such a transmutation may not be dereferenced (not even if `T` has size zero).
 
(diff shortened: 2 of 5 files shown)


src/type-layout.md
```diff
@@ -56,12 +56,23 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
+For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every bit pattern
+represents a valid instance of the type (in other words,
+`transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the primitive
+numeric types and also for `bool` and `char`, every byte is guaranteed to be
+initialized (in other words, `transmute::<T, [u8; size_of::<T>()]>(...)` is always
+sound).
```
the8472, line 65: `T` here appears unconstrained and could be misunderstood. Instead it could say something like this:

> For the primitive numeric types `NUM` [...] `transmute::<[u8; size_of::<NUM>()], NUM>(...)`

joshlf: I took a slightly different approach so it'd be consistent across the whole paragraph. How does that look?

the8472: Well, that leaves the first transmute.

joshlf: I updated the first one too:

> (in other words, for every primitive numeric
type, `T`, `transmute::<[u8; size_of::<T>()], T>(...)` is always sound)

the8472: Oh yeah, looks fine.

src/type-layout.md
```diff
@@ -56,12 +56,23 @@ Most primitives are generally aligned to their size, although this is
...
 ## Pointers and References Layout
 
 Pointers and references have the same layout. Mutability of the pointer or
 reference does not change the layout.
 
-Pointers to sized types have the same size and alignment as `usize`.
+Pointers to sized types have the same size and alignment as `usize`. Every
+byte of a pointer to a sized type and of a reference to a sized type is
+initialized (in other words, for such a pointer or reference type, `P`,
+`transmute::<P, [u8; size_of::<P>()]>(...)` is always sound).
```
RalfJung, line 75: Ah this is where ptr2int transmute is even directly mentioned. That's a huge topic for discussion, see https://github.com/rust-lang/unsafe-code-guidelines/issues/286. And it's far from an easy topic, see [this blog post](https://www.ralfj.de/blog/2022/04/11/provenance-exposed.html) for why it is tricky.

It is almost certainly the case that transmuting a pointer to usize and back will *not* give you back the original pointer, but something more like `ptr::invalid(ptr.addr())`. So even if we are willing to commit to ptr2int transmute not being insta-UB, we should caution against the caveats that come with it.

src/type-layout.md
```diff
@@ -56,12 +56,23 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
+For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every bit pattern
+represents a valid instance of the type (in other words, for every primitive numeric
+type, `T`, `transmute::<[u8; size_of::<T>()], T>(...)` is always sound). For the
+primitive numeric types and also for `bool` and `char`, every byte is guaranteed to be
+initialized (in other words, for every such type, `T`,
+`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound).
```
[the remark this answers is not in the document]
RalfJung: Also the properties you are talking about here are not "layout" properties any more, they go deep into the operational semantics / value representation. Not sure if this is the right place in the reference to put them. (But then, not sure where to put anything about the op.sem into the reference.)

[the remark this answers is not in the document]
the8472: > And (b) it presupposes that we allow pointer-to-integer transmutation, which is a subtle question and I am not sure if we're ready to commit to the answer yet.

I think pointers aren't covered by this paragraph since it says "primitive *numeric* type" which aren't part of the list above.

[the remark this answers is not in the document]
RalfJung: Yeah we can definitely commit to transmuting between all the integer and float types.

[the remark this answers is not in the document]
joshlf: Is there accepted nomenclature for "all bit patterns that don't carry extra information in the abstract machine" or something to that effect? In other words, is there nomenclature for what your average programmer means when they say "all bit patterns"? I agree that the transmutes are precise, but IMO the prose is more intuitive for most readers.

src/type-layout.md
joshlf: Do you have a sense of what we _can_ guarantee about ptr2int or int2ptr? If the answer is "nothing", I'd be happy to just remove this from the PR.

src/type-layout.md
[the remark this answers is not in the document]
RalfJung: I'm not aware of a standard term for that.

> In other words, is there nomenclature for what your average programmer means when they say "all bit patterns"?

I've seen plenty of programmers include uninitialized memory in "all bit patterns". So I don't think we can make the assumption that they mean what you say they mean.

src/type-layout.md
RalfJung: For ptr2int transmutes I don't think we are ready to guarantee anything yet (except for the case that is covered by https://github.com/rust-lang/rust/pull/113510).

int2ptr transmutes... I think we can commit to them being legal, but I don't know if we even can commit to the resulting pointer being valid for zero-sized accesses (Cc https://github.com/rust-lang/opsem-team/issues/12).

joshlf: Yeah I'd be fine just removing this paragraph entirely.

src/type-layout.md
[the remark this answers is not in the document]
joshlf: Gotcha. How does this sound? (note the addition of "initialized" in the first sentence)

> For the primitive numeric types (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), every initialized bit pattern
represents a valid instance of the type (in other words, for every primitive numeric
type, `T`, `transmute::<[u8; size_of::<T>()], T>(...)` is always sound). Note that this
does not include uninitialized bit patterns, so transmuting from a type with padding
or other uninitialized bytes is unsound.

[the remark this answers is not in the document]
RalfJung: "every initialized bit pattern represents a valid instance of the type" to me still sounds like "I can transmute pointers to the type".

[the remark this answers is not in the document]
joshlf: Yeah that makes sense. I don't have a good sense of what nomenclature is used for these different concepts; do you have a wording you'd suggest?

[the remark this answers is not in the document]
RalfJung: I'm not aware of a standard term for "bytes that are initialized and don't have provenance". We could call them "pure bytes" or so, but even if a short term existed, it wouldn't be helpful -- people don't expect all this trouble with provenance, so there is no short term we can use to alert them to it.

I think for now this should just say the for every primitive numeric type, transmuting between them is always sound. And it should probably also explicitly say that the semantics of transmuting a pointer to an integer is undecided and that operation should be avoided. (That's what the `transmute` docs say.)

I don't think anyone ever doubted that transmuting between integer types was sound, so I am also somewhat wondering what the goal of this PR is. And is this really the right place to document transmutation rules? Wouldn't that be in the `transmute` docs?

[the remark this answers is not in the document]
joshlf: > I think for now this should just say the for every primitive numeric type, transmuting between them is always sound. And it should probably also explicitly say that the semantics of transmuting a pointer to an integer is undecided and that operation should be avoided. (That's what the `transmute` docs say.)

Gotcha; how does this sound?

> For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. `u8` has 256 valid representations (namely, every 8-bit sequence). The following are not valid representations for `u8`:
> - A padding or uninitialized byte
> - A byte at any offset in a reference or pointer type (the semantics of transmuting a reference or pointer to a non-pointer type is currently undecided)

> I don't think anyone ever doubted that transmuting between integer types was sound, so I am also somewhat wondering what the goal of this PR is. Is this really the right place to document transmutation rules? Wouldn't that be in the `transmute` docs?

My specific use case is that I want to be able to write a safety comment on `unsafe impl FromBytes for T` (where `T` is one of the primitive types described in this paragraph) that proves the soundness of that impl. While I agree that everybody agrees that this is sound, I try to hold myself to a higher standard for safety comments - in particular, I want all of my safety comments to be proofs whose validity logically follows from text in the reference or the standard library documentation. My intention with [PRs](https://github.com/rust-lang/rust/pull/104081) [like](https://github.com/rust-lang/rust/pull/88915) [these](https://github.com/rust-lang/rust/pull/94786) is to codify this understanding so that even people who are completely unaware of the current Rust zeitgeist can still reason about the soundness of their code (namely, based purely on what's guaranteed by official docs).

Another reason that I call out `unsafe impl FromBytes` is that I don't actually think `transmute` is the right place for these things. At least in my mind, `transmute` is a function whose soundness is based on what facts are guaranteed about type layout and validity. There are ways to do what is effectively a transmute without using the `transmute` function itself. IMO the soundness of `transmute` ought to be downstream of these guarantees, and those guarantees should be documented in a central location. My reading is that, when you use `transmute` as an anchor in your soundness proofs, it's really shorthand for a more verbose argument like:
- `transmute::<X, Y>()` is documented as sound.
- Lemma: <some claim about Rust's layout and bit validity guarantees>
  - Proof by contradiction: If <some claim about Rust's layout and bit validity guarantees> were false, `transmute::<X, Y>()` would be unsound.
- Proof: Because <some claim about Rust's layout and bit validity guarantees> is true, what we're doing in this code is sound.

I recognize the irony of making that point on a PR in which I've used `transmute` in the Rust reference itself; it's just that I don't think we have good language for these things right now, and using `transmute` is an understood shorthand. My hope is that at least putting this in the reference (rather than just in the `transmute` docs) is a step in the right direction.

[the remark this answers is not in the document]
Dante-Broggi: > I'm not aware of a standard term for "bytes that are initialized and don't have provenance".

Perhaps "sequence of octets"? IIUC, this is what data formats developed before cchar stabilized on 8-bits call 'normal' bytes.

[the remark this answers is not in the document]
RalfJung: > The following are not valid representations for u8:

We should make it clear that parts of this is undecided, I think. This sounds too definite for my taste.^^

> I want all of my safety comments to be proofs whose validity logically follows from text in the reference or the standard library documentation. 

I like it. :)

>  At least in my mind, transmute is a function whose soundness is based on what facts are guaranteed about type layout and validity. 

I fully agree. However, I certainly would not have thought to look in this place in the reference for such information. At the least, the library docs should then point to the reference docs -- for example [pages like these](https://doc.rust-lang.org/nightly/std/primitive.i32.html).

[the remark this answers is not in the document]
joshlf: > > The following are not valid representations for u8:
> 
> We should make it clear that parts of this is undecided, I think. This sounds too definite for my taste.^^

How about:

> A padding or uninitialized byte is not a valid `u8`. A byte at any offset in a reference or pointer type _may_ not be a valid `u8` (the semantics of transmuting a reference or pointer to a non-pointer type is currently undecided).

> > At least in my mind, transmute is a function whose soundness is based on what facts are guaranteed about type layout and validity.
> 
> I fully agree. However, I certainly would not have thought to look in this place in the reference for such information. And `transmute` is the established terminology for "re-interpreting memory from one type to another". At the least, the library docs should then point to the reference docs -- for example [pages like these](https://doc.rust-lang.org/nightly/std/primitive.i32.html), but also the docs for `transmute`.

Yeah, at some point when I have more time I'm hoping to propose a bigger overhaul of everything that includes a) putting all of these docs in a uniform place and putting links to those docs in appropriate places and, b) nailing down terminology like layout, bit validity, ABI, etc, and updating existing docs to use that terminology consistently. One of those never-the-most-important-thing-on-my-plate things, so it's never made it off the back burner, but I'm feeling very tempted to do it soon.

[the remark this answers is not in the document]
joshlf: > > I'm not aware of a standard term for "bytes that are initialized and don't have provenance".
> 
> Perhaps "sequence of octets"? IIUC, this is what data formats developed before cchar stabilized on 8-bits call 'normal' bytes.

Personally I think of this as a term used in contexts where bytes _can't_ have any extra semantic meaning (e.g., a network packet sent on the wire); not sure I've seen it used in the context of abstract machines.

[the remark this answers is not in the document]
RalfJung: Slightly edited:

> An uninitialized byte is not a valid u8. A byte at any offset in a reference or pointer type may not be a valid u8 (the semantics of transmuting a reference or pointer to a non-pointer type is currently undecided).

"uninitialized byte" is part of the memory content. "padding byte" is a completely different concept, it appears in questions such as "does type X have a padding byte at offset N". But a byte that just sits in memory doesn't have a type and hence it makes no sense to ask whether it is a padding byte.

---

> nailing down terminology like layout, bit validity, ABI, etc, and updating existing docs to use that terminology consistently

Yeah, that makes sense. FWIW I think bit validity itself isn't even the core concept, the core concept is the representation relation -- which has more information than bit validity, for instance for integers if we do allow ptr2int transmutes the representation relation will say that the provenance of the bytes is ignored when computing the value representation of the integer. See [this](https://github.com/RalfJung/minirust/blob/master/spec/lang/representation.md) for an example of how that can be spelled out in full precision.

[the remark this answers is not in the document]
joshlf: > Slightly edited:
> 
> > An uninitialized byte is not a valid u8. A byte at any offset in a reference or pointer type may not be a valid u8 (the semantics of transmuting a reference or pointer to a non-pointer type is currently undecided).
> 
> "uninitialized byte" is part of the memory content. "padding byte" is a completely different concept, it appears in questions such as "does type X have a padding byte at offset N". But a byte that just sits in memory doesn't have a type and hence it makes no sense to ask whether it is a padding byte.

Okay, updated. How does the new text look?

--

> > nailing down terminology like layout, bit validity, ABI, etc, and updating existing docs to use that terminology consistently
> 
> Yeah, that makes sense. FWIW I think bit validity itself isn't even the core concept, the core concept is the representation relation -- which has more information than bit validity, for instance for integers if we do allow ptr2int transmutes the representation relation will say that the provenance of the bytes is ignored when computing the value representation of the integer. See [this](https://github.com/RalfJung/minirust/blob/master/spec/lang/representation.md) for an example of how that can be spelled out in full precision.

Cool, thanks for that link!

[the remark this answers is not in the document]
RalfJung: Can you remove that part?

> `u8` has 256 valid representations (namely, every 8-bit sequence).

We didn't define "8-bit sequence", but this is too close for comfort to "every bit pattern". ;) Also, if a pointer bytes can be valid then u8 has infinitely many representations -- still only 256 abstract values, but infinitely many ways to represent each of them.

src/type-layout.md
```diff
@@ -56,6 +56,18 @@ Most primitives are generally aligned to their size, although this is
...
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
+of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. `u8` has 256
+valid representations (namely, every 8-bit sequence). An uninitialized byte is not
+a valid `u8`. A byte at any offset in a reference or pointer type may not be a
+valid `u8` (the semantics of transmuting a reference or pointer to a non-pointer
+type is currently undecided).
+
+For the primitive numeric types and also for `bool` and `char`, every byte is
+guaranteed to be initialized (in other words, for every such type, `T`,
+`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound).
```
RalfJung, line 69:
```suggestion
For `bool` and `char`, every byte is
guaranteed to be initialized (in other words, for every such type, `T`,
`transmute::<T, [u8; size_of::<T>()]>(...)` is always sound -- but the inverse is not).
```
The other part seems redundant with the previous paragraph? I always find it confusing when there is redundancy that is not explicitly called out as redundancy, since the reader might think this is supposed to be new information.

src/type-layout.md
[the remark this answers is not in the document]
joshlf: Hmmm good point. Is there something else we can say instead of just removing it? Worst comes to worst, removing it is still fine - most code that needs to justify its soundness in terms of this paragraph is probably "bottoming out" at `u8` anyway, so defining everything in terms of `u8` and not fully nailing down `u8` itself is probably actually fine. But it'd be good to be able to make guarantees about what `u8` is instead of just what it isn't (uninitialized, pointer).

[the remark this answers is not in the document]
RalfJung: Without having laid down the terminology of values and representation, it's tricky. We could say that every u8 represents a mathematical integer in `0..256`. But how high is the risk that people will skip the "represents" and get this wrong? Also, it seems really strange to say that only about u8. Unlike C, Rust doesn't have a magic 1-byte integer types. All our `uN` types represent a mathematical integer in `0..(2.pow(N))`.

I feel if we want to say something like this, we need some text somewhere about how each type has two sides: a set of abstract values it can represent, and a set of low-level representations each representing one such value (with potentially more than one representation per value). Then this text here can link to those terms and therefore make it clear what they mean. We can nail down the set of abstract values for the integer types, `bool`, `char`, but we cannot fully lay down their sets of valid representations -- but we *can* say that they are sufficiently compatible, in the sense that a valid representation of any of them is also a valid representation of `[u8; size_of::<T>()]` (and for the integer types, this is an if-and-only-if).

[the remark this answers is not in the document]
joshlf: Yeah, that sounds reasonable. I'll just remove the "all 256 8-bit sequences" text for now, and we can follow up with that at some point in the future.

RalfJung (approved):
LGTM! I'm confident this is covered by the FCP in https://github.com/rust-lang/unsafe-code-guidelines/issues/439.

joshlf:
Thanks for all of the discussion on this, @RalfJung!

RalfJung:
I don't think so, but review capacity on reference PRs is very low.

RalfJung:
With the reference I am never sure.^^ @ehuss is our main editor here.

ehuss:
Can you say why this is placed in the type-layout chapter? In general, that chapter is focused on the size and alignment of types, and not what constitutes and valid type. Valid bit patterns for types are generally documented in their corresponding type pages (like bool specifies that only 0 and 1 are valid bytes).

src/type-layout.md
```diff
@@ -56,6 +56,17 @@ Most primitives are generally aligned to their size, although this is
 platform-specific behavior. In particular, on x86 u64 and f64 are only
 aligned to 32 bits.
 
+For every primitive numeric type (`u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`,
+`i64`, `u128`, `i128`, `usize`, `isize`, `f32`, and `f64`), `T`, the bit validity
+of `T` is equivalent to the bit validity of `[u8; size_of::<T>()]`. An
+uninitialized byte is not a valid `u8`. A byte at any offset in a reference or
+pointer type may not be a valid `u8` (the semantics of transmuting a reference or
+pointer to a non-pointer type is currently undecided).
```
ehuss, line 64: This "may" seems confusing to me. When is it valid or not valid?

RalfJung: Those details are undecided as of now.

ehuss: Oh, I see, thanks!

joshlf:
> Can you say why this is placed in the type-layout chapter? In general, that chapter is focused on the size and alignment of types, and not what constitutes and valid type. Valid bit patterns for types are generally documented in their corresponding type pages (like bool specifies that only 0 and 1 are valid bytes).

I can do that, it'd just require somewhat more text since the second paragraph (about going from `T` to `[u8; size_of::<T>()]`) would need to be duplicated on the pages for [bools](https://doc.rust-lang.org/stable/reference/types/boolean.html), [chars](https://doc.rust-lang.org/stable/reference/types/textual.html), and [numeric types](https://doc.rust-lang.org/stable/reference/types/numeric.html). If that's your preference, I'm happy to do that.

ehuss:
Yea, I'm fine with duplicating it. For the bool case, that could also be worded a little different since "every byte" is a little confusing since there is only one byte. Thanks!

joshlf:
Okay, I updated to this text in `pointer.md`, so the PR is ready for review. Happy to edit this as requested.

> ## Bit validity
> Despite pointers and references being similar to `usize`s in the machine code emitted on most platforms, the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided. Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.

joshlf:
One follow-up thought inspired by https://github.com/google/zerocopy/pull/294: @RalfJung and @ehuss, would you be on board with me also adding the following to `pointer.md`? If this is at all controversial, I'll just follow up in a separate PR to avoid holding this one up.

> It is always sound to produce a thin null pointer (i.e., for `T: Sized` and `P = *const T` or `P = *mut T`, `transmute::<_, P>([0u8; size_of::<P>()])`). Note that this can be done without `unsafe` code using the [`ptr::null`](https://doc.rust-lang.org/core/ptr/fn.null.html) or [`ptr::null_mut`](https://doc.rust-lang.org/core/ptr/fn.null_mut.html) functions.

src/types/pointer.md
```diff
@@ -50,6 +50,12 @@ Raw pointers can be created directly using [`core::ptr::addr_of!`] for `*const`
 
 The standard library contains additional 'smart pointer' types beyond references and raw pointers.
 
+## Bit validity
+
+Despite pointers and references being similar to `usize`s in the machine code emitted on most platforms,
+the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
+Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
```
RalfJung, line 57:
```suggestion
Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.

For thin raw pointers (i.e., for `T: Sized` and `P = *const T` or `P = *mut T`), the inverse direction is always valid.
However, the resulting pointer may not be dereferenced (not even if the pointee type has size zero).
```
Note however that there are thin pointers where `T: !Sized` (due to extern types). So maybe instead of `T: Sized` we should say `<T as Pointee>::Metadata == ()`.

joshlf: Sounds good, but do we want to refer to unstable features (the ptr_metadata feature for `Pointee`) in the reference? Maybe we could instead say `size_of::<P>() == size_of::<usize>()`?

RalfJung: IMO using `Metadata` is more clear even if it isn't stable; I will defer to @ehuss here.

joshlf: @ehuss Any preference on the wording here? I'm happy to go with whatever you recommend.

ehuss: We generally try to avoid referring to any unstable elements in the Reference, so if you can avoid it, it would be appreciated. If @RalfJung is ok with that wording as a compromise, it seems good to me.

Unfortunately "thin" and "fat" pointers currently aren't defined in the reference, so it is a little awkward to start using terms that aren't defined. I don't think that necessarily needs to be addressed in this PR, but it is something to be aware of (and eventually resolved).

joshlf: Sounds good, updated. Should be ready to go!

RalfJung:
Yeah this is fine for me, it just will affect people working with extern types (but those are still unstable anyway).

ehuss (approved):
Thanks!

src/types/pointer.md
```diff
@@ -50,6 +50,16 @@ Raw pointers can be created directly using [`core::ptr::addr_of!`] for `*const`
...
```
joshlf, line 57: @RalfJung For my own edification, how does this interact with pointer-to-integer `as` casts? I can imagine a few possibilities (not mutually exclusive):
- `as` casts are sound, but transmutations are not necessarily sound (even though you can emulate a transmutation by doing an `as` cast followed by a transmutation from `usize`)
- the first bullet is true _and_ pointer-to-integer `as` casts probably shouldn't have been allowed in safe Rust, but it's too late

RalfJung: The current working model is that `as` casts are special:
- `ptr as usize` is like https://doc.rust-lang.org/nightly/std/primitive.pointer.html#method.expose_addr
- `usize as ptr` is like https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html

In contrast:
- "transmute ptr to usize" is like https://doc.rust-lang.org/nightly/std/primitive.pointer.html#method.addr
- "transmute usize to ptr" is like https://doc.rust-lang.org/nightly/std/ptr/fn.invalid.html

All of these operations are safe and sound, but as their docs show, they behave quite differently and that can affect the soundness of later code that uses the values they produce.

Note that this is a prototype model, which has some unresolved problems and which is very much not stabilized.

joshlf: Okay that all makes sense, thanks!

> Note that this is a prototype model, which has some unresolved problems and which is very much not stabilized.

I always assume that of everything in this space unless documented otherwise :)

PR #1394, opened 2023-08-15
Document early exit of method candidate search

Urhengulas:
Before it was unclear that Rust does not search through all candidates types, but exits the search as soon as it found a fitting method.


traviscross:
Thanks @Urhengulas for submitting this.  We agree this could be improved a bit, but on review, we weren't happy enough with the language here to merge it.  It's hard to get this really right!  We'll close this in favor of #1534.  See also the other linked issues there.

PR #1395, opened 2023-08-16
conditional-compilation.md: Mention the "none" target_os value

evanperrygrove-somewear:
As an embedded dev who uses Windows at work, it physically pains me when people use `#[cfg(target_os = "unix")]` when they actually mean `#[cfg(not(target_os = "none"))]`. Hopefully this will increase awareness that "none" is a valid `target_os`.


ehuss (approved):
Thanks!

PR #1397, opened 2023-08-19
we reserve the right to reduce our amount of UB

RalfJung:
It was pointed out to me that the docs currently say that the things on the UB list are definitely UB. That's not true, we do reserve the right to have less UB in the future. (Exactly that is happening e.g. in https://github.com/rust-lang/reference/pull/1387.)

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -14,10 +14,14 @@ undefined behavior, it is *unsound*.
 
 <div class="warning">
 
-***Warning:*** The following list is not exhaustive. There is no formal model of
-Rust's semantics for what is and is not allowed in unsafe code, so there may be
-more behavior considered unsafe. We also reserve the right to make some of the
-behavior in that list defined in the future.
+***Warning:*** The following list is not exhaustive; it may grow or shrink.
+There is no formal model of Rust's semantics for what is and is not allowed in
+unsafe code, so there may be more behavior considered unsafe. We also reserve
+the right to make some of the behavior in that list defined in the future. In
+other words, this list does not say that anything will *definitely* always be
+undefined in all future Rust version (but we might make such commitments for
+some list items in the future).
+
 Please read the [Rustonomicon] before writing unsafe code.
 
 </div>
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -16,9 +16,9 @@ undefined behavior, it is *unsound*.
 
 ***Warning:*** The following list is not exhaustive. There is no formal model of
 Rust's semantics for what is and is not allowed in unsafe code, so there may be
-more behavior considered unsafe. The following list is just what we know for
-sure is undefined behavior. Please read the [Rustonomicon] before writing unsafe
-code.
+more behavior considered unsafe. We also reserve the right to make some of the
+behavior in that list defined in the future.
+Please read the [Rustonomicon] before writing unsafe code.
 
 </div>
 


saethlin (approved):

pnkfelix:
@rustbot label: +I-lang-nominated

I want to discuss this at the @rust-lang/lang design team meeting. In particular, on the [associated zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/we.20reserve.20the.20right.20to.20reduce.20our.20amount.20of.20UB/near/387027064) where I asked why someone would *want* the prior interpretation of this text, a couple interesting cases were described:
 * I hypothesized, as a [strawman](https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/we.20reserve.20the.20right.20to.20reduce.20our.20amount.20of.20UB/near/387048107), that compiler optimization authors want it so they don't have to update their code when the language spec changes
 * @chorman0773 [pointed out](https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/we.20reserve.20the.20right.20to.20reduce.20our.20amount.20of.20UB/near/387069939) that some API's might be written in a way where they just say "precondition: `*x` is a legal operation", and then treat the implications of that according to this list **at the time when they are writing the API** as something they can assume for all time.
 * @RalfJung pointed out their own motivating case, which is verification tools that want to similarly make [assumptions  in their reasoning](https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc/topic/we.20reserve.20the.20right.20to.20reduce.20our.20amount.20of.20UB/near/387149252) based on the UB list.

src/behavior-considered-undefined.md
```diff
@@ -16,9 +16,9 @@ undefined behavior, it is *unsound*.
 
 ***Warning:*** The following list is not exhaustive. There is no formal model of
 Rust's semantics for what is and is not allowed in unsafe code, so there may be
-more behavior considered unsafe. The following list is just what we know for
-sure is undefined behavior. Please read the [Rustonomicon] before writing unsafe
-code.
+more behavior considered unsafe. We also reserve the right to make some of the
+behavior in that list defined in the future.
```
joshtriplett, line 20:
```suggestion
behavior in that list defined in the future, or conversely, to guarantee that it will always be undefined.
```

nikomatsakis:
I think we should make this change, but I just want to raise the point I always raise, that simply "removing UB" isn't "always" a fine change to make. My classic example is that rayon (and other libraries) assume that destructors will run for their local variables before the stack frame is popped. I think we eventually will want to be clear about those things we are guaranteeing (which implies a certain amount of operations that will always be UB) versus things that are UB now but which we may change.

nikomatsakis:
@rfcbot fcp merge

I think we should make this change!

pnkfelix:
> @pnkfelix points out this is similar to the "assume `*x` is legal to deref" example that @chorman0773 raised.

In particular, @chorman0773 's example was a specific "the given UB list means that 'expression A is UB implies expression B is also UB', so if my method precondition includes 'expression B is *well-defined*', then my code can assume A is likewise well-defined." (i.e. an instance of using the contrapositive reasoning).

In the meeting, I waffled on whether @nikomatsakis 's example is similar or not. I think I'm still waffling now, but I'm coming back around to thinking that it indeed is such an example.

pnkfelix:
@rfcbot fcp merge

I think we should make this change!

pnkfelix:
(does rfcbot not monitor rust-lang/reference ...?)

RalfJung:
> I think we should make this change, but I just want to raise the point I always raise, that simply "removing UB" isn't "always" a fine change to make, because it can make a sound API become unsound. My classic example is that rayon (and other libraries) assume that destructors will run for their local variables before the stack frame is popped. I think we eventually will want to be clear about those things we are guaranteeing (which implies a certain amount of operations that will always be UB) versus things that are UB now but which we may change.

I agree in general, yes.
The rayon case wouldn't just remove UB though, that would significantly alter the lowering from surface Rust to MIR/MiniRust.

> (does rfcbot not monitor rust-lang/reference ...?)

It checks in once an hour.

src/behavior-considered-undefined.md
RalfJung: I am confused, this now says "we reserve the right to [...] guarantee that it will always be undefined"?

tmandry: I think Josh is trying to capture the point that Niko raised below: We may want to _promise_ that certain operations will always be UB, so that unsafe code can rely on that.

RalfJung: I don't understand the point of saying "in the future we may want to promise that something will always be UB". On its own, how is that statement useful for anyone? I would understand the point of marking particular items in the UB list as "this will definitely always be UB", but we don't currently have any such items, or do we?

joshtriplett: @RalfJung The wording could absolutely change. The wording currently in the PR says that we could *define* it in the future. The point Niko made, which I wanted to capture, was that we could also commit to *not* defining something (or for that matter commit to not defining it in particular ways), which is something that some crates may need us to do.

RalfJung: Hm, I still don't understand whom this comment is serving/helping. People will be asking for us to commit to something like that no matter what we write here. But sure I can try to put in a phrase along those lines. I just worry it will cause a lot more confusion than it will help people. It would certainly confuse the heck out of me without a multi-paragraph explanation.

RalfJung: I tried my hand at it, but honestly I don't like it. I think the text would be semantically equivalent but less confusing without that new sentence.

ehuss (approved):
Thanks!

PR #1400, opened 2023-09-10
fix(trait docs): remove unnecessary `where Self: Sized` clause

poliorcetics:
This is probably the result of a copy-paste but it's unnecessary for the purpose of the example and leads to the error returned by the playground to be about `where Self: Sized` first instead of the generic type `T`, making the example confusing for this error.

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -116,7 +116,7 @@ trait NonDispatchable {
     // `other` may be a different concrete type of the receiver.
     fn param(&self, other: Self) where Self: Sized {}
     // Generics are not compatible with vtables.
-    fn typed<T>(&self, x: T) where Self: Sized {}
+    fn typed<T>(&self, x: T) {}
 }
 
 struct S;


ehuss:
I believe the original is correct. The example is trying to show that the given trait is valid, it's just that the given associated functions cannot be dispatched on. In other words, if you remove the code that actually tries to dispatch on the functions, it should compile successfully. If the `Self: Sized` bound is removed, then the trait can't be compiled at all which isn't want the example is trying to illustrate.

poliorcetics:
The example can be compiled fully when removing the dispatch calls, even with my change: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fc746026dbcce1c39efaf5257d67b5c0 

The function is taking `&self` so `where Self: Sized` is unnecessary here

ehuss:
But it will fail to compile if you also include the line:

```rust
let obj: Box<dyn NonDispatchable> = Box::new(S);
```

That's what it is trying to illustrate. That you *can* create a trait object, but you can't dispatch on those specific functions. Without the `Self: Sized`, it prevents it from being able to create a trait object at all. The illustration of non-object safety is below in `NotObjectSafe`, which shows the generics without `Self: Sized` prevents being able to create an object.

poliorcetics:
Ah sorry, you're right, I misunderstood the example and surrounding text

PR #1401, opened 2023-09-11
Guarantee `char` layout

joshlf:


joshlf:
Ping 🙂

ehuss:
Nominated to get lang's attention.

src/types/textual.md
```diff
@@ -17,7 +17,9 @@ is valid UTF-8. Calling a `str` method with a non-UTF-8 buffer can cause
 Since `str` is a [dynamically sized type], it can only be instantiated through a
 pointer type, such as `&str`.
 
-## Bit validity
+## Layout and bit validity
+
+`char` is guaranteed to have the same size and alignment as `u32` on all platforms.
```
joshtriplett, line 22:
```suggestion
`char` is guaranteed to have the same size and alignment as `u32` on all platforms.

`Option<char>` is guaranteed to have the same size and alignment as `char` on all platforms.
```

(Unless we want to put this somewhere else.)

joshtriplett: This suggestion should probably be handled separately, not in this PR.

joshtriplett:
Consensus from lang is to merge.

std docs should also be updated.

src/types/textual.md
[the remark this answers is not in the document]
joshlf: No opposition to this going in `core` as well, but we already have text in the Reference about the sizes of primitive types: https://doc.rust-lang.org/reference/type-layout.html#primitive-data-layout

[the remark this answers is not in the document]
joshlf: Put up a PR to add the same text to `core`: https://github.com/rust-lang/rust/pull/116894

WaffleLapkin:
What's the status here?

src/types/textual.md
joshlf: Filed https://github.com/rust-lang/reference/issues/1426 to track this.

PR #1402, opened 2023-09-18
Update conditional-compilation.md

klu235:
* Clarify
* Reduce redundancy

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -22,43 +22,31 @@
 > _ConfigurationPredicateList_\
 > &nbsp;&nbsp; _ConfigurationPredicate_ (`,` _ConfigurationPredicate_)<sup>\*</sup> `,`<sup>?</sup>
 
-*Conditionally compiled source code* is source code that may or may not be
-considered a part of the source code depending on certain conditions. <!-- This
-definition is sort of vacuous --> Source code can be conditionally compiled
-using the [attributes] [`cfg`] and [`cfg_attr`] and the built-in [`cfg` macro].
-These conditions are based on the target architecture of the compiled crate,
-arbitrary values passed to the compiler, and a few other miscellaneous things
-further described below in detail.
+*Conditionally compiled source code* is source code that is compiled only under certain conditions. Source code can be made conditionally compiled using the [`cfg`] and [`cfg_attr`] [attributes] and the built-in [`cfg` macro]. Whether to compile can depend on the target architecture of the compiled crate, arbitrary values passed to the compiler, and other things further described below.
 
 Each form of conditional compilation takes a _configuration predicate_ that
 evaluates to true or false. The predicate is one of the following:
 
-* A configuration option. It is true if the option is set and false if it is
+* A configuration option. The predicate is true if the option is set, and false if it is
   unset.
-* `all()` with a comma separated list of configuration predicates. It is false
-  if at least one predicate is false. If there are no predicates, it is true.
-* `any()` with a comma separated list of configuration predicates. It is true
-  if at least one predicate is true. If there are no predicates, it is false.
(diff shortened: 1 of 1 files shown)


src/conditional-compilation.md
```diff
@@ -248,9 +232,7 @@ Example values:
 <!-- should we say they're active attributes here? -->
 
 The `cfg` [attribute] conditionally includes the thing it is attached to based
-on a configuration predicate.
-
-It is written as `cfg`, `(`, a configuration predicate, and finally `)`.
+on the given configuration predicate (_ConfigurationPredicate_ in the syntax above).
```
ehuss, line 235: The original is written intentionally that way, since we agreed on a style of describing the syntax in prose in most cases, and to not defer to the grammar. Can this change be reverted?

src/conditional-compilation.md
```diff
@@ -22,43 +22,31 @@
...
-* A configuration option. It is true if the option is set and false if it is
+* A configuration option. The predicate is true if the option is set, and false if it is
   unset.
-* `all()` with a comma separated list of configuration predicates. It is false
-  if at least one predicate is false. If there are no predicates, it is true.
-* `any()` with a comma separated list of configuration predicates. It is true
-  if at least one predicate is true. If there are no predicates, it is false.
-* `not()` with a configuration predicate. It is true if its predicate is false
-  and false if its predicate is true.
+* `all()` with a comma-separated list of configuration predicates. It is true precisely if all of the given predicates are true.
```
ehuss, line 32: This doesn't seem to be quite correct, since it doesn't seem to cover the empty list case (there are no true predicates in that case). Perhaps something like:

```suggestion
* `all()` with a comma-separated list of configuration predicates. It is true if all of the given predicates are true, or if the list is empty.
```

src/conditional-compilation.md
```diff
@@ -22,43 +22,31 @@
...
+* `any()` with a comma-separated list of configuration predicates. It is true precisely if at least one of the given predicates is true.
```
ehuss, line 33: Similarly:

```suggestion
* `any()` with a comma-separated list of configuration predicates. It is true precisely if at least one of the given predicates is true. If there are no predicates, it is false.
```

src/conditional-compilation.md
```diff
@@ -22,43 +22,31 @@
...
+* `not()` with a configuration predicate. It is true precisely if the given predicate is false.
```
ehuss, line 34: I don't feel like brevity of words here is helping. The original seemed sufficiently brief, and seemed a little clearer to me. I would suggest keeping this as it was before.

```suggestion
* * `not()` with a configuration predicate. It is true if its predicate is false
  and false if its predicate is true.
```

src/conditional-compilation.md
```diff
@@ -22,43 +22,31 @@
...
 
-_Configuration options_ are names and key-value pairs that are either set or
-unset. Names are written as a single identifier such as, for example, `unix`.
-Key-value pairs are written as an identifier, `=`, and then a string. For
-example, `target_arch = "x86_64"` is a configuration option.
+_Configuration options_ are either names or key-value pairs, and are either set or
+unset. Configuration options that are names, such as `unix`, are specified by simply writing them.
+Configuration options that are key-value pairs are written in the form `key = "value"`, such as `target_arch = "x86_64"`.
```
ehuss, line 38: The change here seems to remove some explicitness about the details of the format. That is, the words "written as a single identifier" and "Key-value pairs are written as an identifier, `=`, and then a string.".  The text needs to explicitly say what is the syntax and form. Just mentioning an example like `key = "value"` isn't sufficient to define what the syntax is.

ehuss (commented):
Thanks, I appreciate the improvements!

ehuss (approved):
Thanks! I went ahead and pushed the suggestions.

PR #1408, opened 2023-09-26
docs: Fix links to ECMA standards in `attributes.md`

ghoseb:
The links to the ECMA-334 and ECMA-335 were broken, I have updated those to working links.


ehuss (approved):
Thanks!

PR #1409, opened 2023-09-26
Adjust reference for return-position `impl Trait` in trait and `async fn` in trait

compiler-errors:
The section on RPITIT is kind of bare, but I'm happy to flesh it out if we want to specify how it works more.

cc https://github.com/rust-lang/rust/pull/115822#issuecomment-1731889250

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -43,7 +43,7 @@ trait Example {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks! Yea, this is an area that is not well documented. I don't think I'll have a chance to review the implementation, but this seems to match my limited understanding.

PR #1413, opened 2023-10-09
some asm block flags also mean there can be no fences

RalfJung:
@Amanieu I wasn't sure about the `readonly` part of this. An acquire read is not entirely a read-only operation in terms of how it affects the machine state, e.g. it cannot be arbitrarily reordered with other reads. So I assume `readonly` blocks cannot really have "acquire" effects?

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -416,8 +416,10 @@ Currently the following options are defined:
   This allows the compiler to execute the `asm!` block fewer times than specified in the program (e.g. by hoisting it out of a loop) or even eliminate it entirely if the outputs are not used.
 - `nomem`: The `asm!` blocks does not read or write to any memory.
   This allows the compiler to cache the values of modified global variables in registers across the `asm!` block since it knows that they are not read or written to by the `asm!`.
+  The compiler also assumes that this `asm!` block does not perform any kind of synchronization with other threads, e.g. via fences.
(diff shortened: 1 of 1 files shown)


RalfJung:
I realized `pure` probably also needs this comment about not having fences.

Are there any others? Or can everything else still do synchronization?

RalfJung:
Oh I see. That's not mentioned in the docs either, so I added it.

josephlr:
I like this! It definitely cleared some stuff up for me.   

> Oh I see. That's not mentioned in the docs either, so I added it.

It seems like it's mentioned further down:

> The `pure` option must be combined with either the `nomem` or `readonly` options, otherwise a compile-time error is emitted. 

But I think also having it in the `pure` section makes the options easier to understand.

RalfJung:
Oh I see, so now it's redundant.

I'm fine either way, @ehuss @Amanieu let me know which way you prefer. :)

Amanieu:
I think it's probably better to move it next to the definition of `pure`.

RalfJung:
All right, done that.

RalfJung:
@ehuss as far as I can see, this is ready to land.

ehuss (approved):
Thanks! I was waiting for @Amanieu, but it seems like they are ok with it. If there are any changes needed, someone can follow up with another PR.

PR #1414, opened 2023-10-10
Fix the syntax of the last expression of a block.

### LINKED ISSUE #762: Flaw/inconsistency between block expression description and rustc.
Currently this compiles:
```rust
fn foo() -> i32 {
    {3}
}
```
According to https://doc.rust-lang.org/nightly/reference/expressions/block-expr.html  , the {3} part cannot be parsed as tail expression, since it's not `ExpressionWithoutBlock`. If it is parsed as `Statement`, then there should be a compile error, because its type is not `()` and the semicolon could not be not eliminated.

@Centril

tczajka:
The last expression is no longer a "Statement", even when it has a block.

Fixes #762.

### DIFF
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -4,13 +4,8 @@
 > _BlockExpression_ :\
 > &nbsp;&nbsp; `{`\
 > &nbsp;&nbsp; &nbsp;&nbsp; [_InnerAttribute_]<sup>\*</sup>\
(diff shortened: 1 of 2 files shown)


traviscross:
Thanks @tczajka for this PR.  There is a lot of good here.  But we actually want to go in a different direction on this, as described here:

- https://github.com/rust-lang/reference/issues/1533

That is, we want to express a block as a sequence of statements.

We'll close this PR, but feel free to open another one along these lines.

PR #1415, opened 2023-10-10
Doc: Add the RISC-V stabilized target features

coastalwhite:
This PR adds the stabilized target features as part of rust-lang/rust#116485.


ehuss (approved):
Thanks!

PR #1416, opened 2023-10-11
Fix temporary drop scope for last expression.

tczajka:
The outermost scope is "the entire function", not "the entire function body" which is different.


tczajka:
This was pointed out in this [URLO thread](https://github.com/rust-lang/reference/pull/1416).

PR #1417, opened 2023-10-11
Guarantee that raw pointer conversions preserve slice element count

joshlf:
This encodes the behavior agreed upon in https://github.com/rust-lang/unsafe-code-guidelines/issues/288.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -480,7 +480,7 @@ assert_eq!(values[1], 3);
 
 #### Slice DST pointer to pointer cast
 
-When `T` and `U` are both "slice DSTs" - ie, slice types or types whose trailing field
+When `T` and `U` are both "slice DSTs" - i.e., slice types or types whose trailing field
 is a slice type - the raw pointer types `*const T`, `*mut T`, `*const U`, and `*mut U`
 encode the number of elements in this slice. Casts between these raw pointer types
 preserve the number of elements. Note that, as a consequence, such casts do *not*
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -480,12 +480,15 @@ assert_eq!(values[1], 3);
 
 #### Slice DST pointer to pointer cast
 
-For slice types like `[T]` and `[U]`, the raw pointer types `*const [T]`, `*mut [T]`, `*const [U]`, and `*mut [U]`
-encode the number of elements in this slice. Casts between these raw pointer types
-preserve the number of elements. Note that, as a consequence, such casts do *not*
-necessarily preserve the size of the pointer's referent (e.g., casting `*const [u16]`
-to `*const [u8]` will result in a raw pointer which refers to an object of half the size
-of the original).
+For slice types like `[T]` and `[U]`, the raw pointer types `*const [T]`, `*mut [T]`,
+`*const [U]`, and `*mut [U]` encode the number of elements in this slice. Casts between
+these raw pointer types preserve the number of elements. Note that, as a consequence,
+such casts do *not* necessarily preserve the size of the pointer's referent (e.g.,
+casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an
+object of half the size of the original). The same holds for `str` and any compound type
+whose unsized tail is a slice type, such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+
+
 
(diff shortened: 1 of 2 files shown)


src/type-coercions.md
```diff
@@ -156,6 +156,16 @@ Coercion is allowed between the following types:
 
 * `!` to any `T`
 
+#### Slice DST raw pointer conversions
+
+When `T` and `U` are both "slice DSTs" - ie, slice types or types whose trailing field
+is a slice type - the raw pointer types `*const T`, `*mut T`, `*const U`, and `*mut U`
+encode the number of elements in this slice. Coercions between these raw pointer types
```
kupiakos, line 163: This is not a coercion: this is a cast. Casts are explicit and use `as`, coercions are implicit.

joshlf: Hmmm yeah looks like I misread the cast docs; I've moved this text to a more appropriate location.

src/expressions/operator-expr.md
```diff
@@ -478,6 +478,16 @@ unsafe {
 assert_eq!(values[1], 3);
 ```
 
+#### Slice DST pointer to pointer cast
+
+When `T` and `U` are both "slice DSTs" - ie, slice types or types whose trailing field
```
kupiakos, line 483: nit: i.e.

joshlf: Done.

joshlf:
Friendly ping 🙂

nikomatsakis (approved):
Substance LGTM, I have a nitpick on the wording.

src/expressions/operator-expr.md
```diff
@@ -478,6 +478,16 @@ unsafe {
...
 
+#### Slice DST pointer to pointer cast
+
+When `T` and `U` are both "slice DSTs" - i.e., slice types or types whose trailing field
+is a slice type - the raw pointer types `*const T`, `*mut T`, `*const U`, and `*mut U`
+encode the number of elements in this slice. Casts between these raw pointer types
+preserve the number of elements. Note that, as a consequence, such casts do *not*
+necessarily preserve the size of the pointer's referent (e.g., casting `*const [u16]`
+to `*const [u8]` will result in a raw pointer which refers to an object of half the size
+of the original).
```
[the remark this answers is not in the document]
joshlf: I'm okay with this approach, but I believe we'd need to add text to make it clear that this covers arbitrary slice DSTs such as:

```rust
struct Foo {
    u: u16,
    trailing: [u8],
}

struct Bar {
    u: u32,
    foo: Foo,
}
```

nikomatsakis:
@rfcbot fcp merge

We discussed in the @rust-lang/lang meeting today and meeting consensus was to merge this. Documenting that.

joshlf:
> I think this is often sketchy, and it might overall be better to encourage `ptr::slice_from_raw_parts` instead of `as` for this, but regardless this is what it _does_ on stable so 👍 we should have that recorded in the reference.
> 
> @rfcbot reviewed

Also worth noting that `ptr::slice_from_raw_parts` does not work for custom DSTs (eg, `struct Foo { u: u8, trailing: [u8] }`).

traviscross:

This was discussed today and is now in FCP.  We'll remove the nomination.

src/expressions/operator-expr.md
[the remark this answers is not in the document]
CAD97: That should be as simple as adding one more sentence that

> The same holds for `str` and any compound type whose unsized tail is a slice type, such as `struct Foo(i32, [u8])` or `(u64, Foo)`.

Given that, I do think it's clearer to show this in terms of normal slices then say it applies to compound types as well than it is to define it directly in terms of potentially compound types.

I know the concept of a type's "unsized tail" is defined somewhere, though I don't recall exactly where... I believe it's the rules around when a field is allowed to be `?Sized`.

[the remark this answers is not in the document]
joshlf: > That should be as simple as adding one more sentence that
> 
> > The same holds for `str` and any compound type whose unsized tail is a slice type, such as `struct Foo(i32, [u8])` or `(u64, Foo)`.
> 
> Given that, I do think it's clearer to show this in terms of normal slices then say it applies to compound types as well than it is to define it directly in terms of potentially compound types.

Ah I like that wording, done!

ehuss (approved):
Thanks!

PR #1424, opened 2023-11-01
Document that <- is a single token

### LINKED ISSUE #583: Maybe document emplacement syntax `<-`
I think the `<-` token might need to be included in tokens.md because it is accepted as a token in macros.

Additionally, `if x<-1 {}` is not valid syntax. It requires a space before `-`.  See https://github.com/rust-lang/rust/pull/51883  That should be specified somewhere.

mattheww:
`<-` was the "move operator", removed in Rust 0.5, but is still lexed as a single token.

So I think it should be listed in the Punctuation table (like the tilde token is).

Originally reported as ferrocene/specification#452

Fixes #583

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -608,6 +608,7 @@ usages and meanings are defined in the linked pages.
 | `::`   | PathSep     | [Path separator][paths]
 | `->`   | RArrow      | [Function return type][functions], [Closure return type][closures], [Function pointer type]
 | `=>`   | FatArrow    | [Match arms][match], [Macros]
+| `<-`   | LArrow      | The left arrow symbol has been unused since before Rust 1.0, but it is still treated as a single token
 | `#`    | Pound       | [Attributes]
 | `$`    | Dollar      | [Macros]
 | `?`    | Question    | [Question mark operator][question], [Questionably sized][sized], [Macro Kleene Matcher][macros]


ehuss:
Thanks, yea, I noted this in #583.

Since there was an attempt to remove this that was then later backed out, I'm assuming the language team agrees that the emplacement operator has to stay permanently. However, I'm not 100% sure based on https://github.com/rust-lang/rust/issues/50832#issuecomment-389884751. 

Since I'm not completely confident that the language team doesn't want to try to remove this in the future, I'm going to nominate this for their attention for approval. I think this PR probably should be merged. I wouldn't be surprised if there were more macros using it since the last time it was attempted to be removed.

mattheww:
It looks like the main question in rust-lang/rust#50832 is whether to remove this feature from the parser.

Perhaps the question of whether it will be removed from the lexer is easier.

In particular, if the answer to that one is "perhaps, but if so it would happen in a future edition", then I think there'd be no reason to delay documenting the current behaviour any further.

tmandry:
> Since there was an attempt to remove this that was then later backed out, I'm assuming the language team agrees that the emplacement operator has to stay permanently. However, I'm not 100% sure based on [rust-lang/rust#50832 (comment)](https://github.com/rust-lang/rust/issues/50832#issuecomment-389884751).

It isn't clear to me _why_ the token removal was backed out; the crater run above https://github.com/rust-lang/rust/pull/60803#issuecomment-494661527 didn't find any actual regressions and the comment didn't leave a rationale.

It seems like it would have been possible to remove the token then, but now it might be harder. It seems worth running crater again to check, if only to make `x<-1` parse correctly. Though if someone knows of crates that will definitely break or wants to argue for why it's a bad idea, we can skip it and I'll resolve the concern.

@rfcbot concern can we still remove the token

Otherwise, I'm not opposed to documenting the current behavior.

@rfcbot reviewed

ehuss:
It sounds like from https://github.com/rust-lang/rust/pull/60803 that Centril and petrochenkov wanted to keep the `<-` token for "future-proofing" and "keeping options open". Does the lang team want to close that door? 

Perhaps it could be removed, and then added again later in an edition (though I don't know how feasible that is)?

tmandry:
Discussed in t-lang and we agreed to unblock documenting this for now. @compiler-errors also wants to do a crater run so we can consider removing it, but that doesn't need to block this PR.

@rfcbot resolve can we still remove the token

traviscross:

As mentioned above, we discussed this today and the consensus was that we should separate the question of documenting the current behavior from maybe changing that behavior.  This can proceed as soon as enough people review and approve.

Once the crater run is analyzed, if someone wants to make a case for changing this behavior, please discuss the pros/cons and nominate for T-lang over in rust-lang/rust#118920 or in a new issue.

ehuss (approved):

PR #1427, opened 2023-11-19
Lone `self` in a method body resolves to the self parameter

mattheww:
In a method body, a path expression consisting only of the keyword `self` resolves to the method's 'self parameter'.

(I checked PR #1052 and I think it doesn't add this case.)


ehuss (approved):
Thanks!

PR #1428, opened 2023-11-20
Reference wildcard patterns from underscore expr

riking:
These appear in nearly identical locations in source code and have subtly different semantics, so cross link them.

I'm not sure if this is the correct link path?


ehuss (approved):
Thanks!

We generally use `.md` extensions so that they are browsable on GitHub. mdBook automatically converts the extension. I have updated it to use `.md`.

PR #1434, opened 2023-12-06
Lifetime and scope clarification in `Closure types` reference.

stegrams:
As I understood so far, lifetime is affecting the borrowing rules enforcement, scope on the other hand, has to do with memory deallocation. 
The variables' lifetime ends the last time the compiler locates its usage, whereas the scope ends with a distinct symbol, i.e. right bracket.

### DIFF
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -461,7 +461,8 @@ let z = &x;
 ```
 
 In this case, borrowing `x` mutably is not possible, because `x` is not `mut`.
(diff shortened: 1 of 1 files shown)


traviscross:
The changes in the following PR may address the motivation for the change in this one:

- https://github.com/rust-lang/reference/pull/1521

stegrams:
I tried to resolve the conflicts two times, but my changes go to /dev/null. 

I reverted to the example of the master branch (conflict 1), but I kept the clarification in the explanation that follows the example (conflict 2), because it's the main purpose of this PR.

Furthermore, I reverted to the one-line sentence style of the master branch.

PR #1435, opened 2023-12-08
Document object unsafety of async-fn-in-trait 

kpreid:
`async fn` in traits, and return-position `impl Trait` in traits, were stabilized by <https://github.com/rust-lang/rust/pull/115822>. The object-safety error E0038 points to the Reference, so the Reference should mention this new criterion.


compiler-errors (approved):

compiler-errors:
lgtm

ehuss (approved):
Thanks!

PR #1442, opened 2023-12-26
pure asm blocks must terminate

RalfJung:
@Amanieu am I correct in assuming that all other asm blocks (i.e, not marked `pure`), may loop infinitely?

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -412,8 +412,7 @@ The list of clobbered registers for each ABI is updated in rustc as architecture
 
 Flags are used to further influence the behavior of the inline assembly block.
 Currently the following options are defined:
-- `pure`: The `asm!` block has no side effects, and its outputs depend only on its direct inputs (i.e. the values themselves, not what they point to) or values read from memory (unless the `nomem` options is also set).
-  The `asm!` block must always terminate.
+- `pure`: The `asm!` block has no side effects, must eventually return, and its outputs depend only on its direct inputs (i.e. the values themselves, not what they point to) or values read from memory (unless the `nomem` options is also set).
   This allows the compiler to execute the `asm!` block fewer times than specified in the program (e.g. by hoisting it out of a loop) or even eliminate it entirely if the outputs are not used.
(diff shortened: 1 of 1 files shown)


Amanieu:
Since non-pure asm blocks can do anything C FFI can do, yes, it's valid.

However I'm not sure that `pure` should be prevented from looping indefinitely. How is that different from it just taking a very long time to run?

RalfJung:
If they loop infinitely they can cannot be optimized away.
```rust
asm!("<infinite loop>", pure);
unreachable_unchecked();
```
It is my understanding that the compiler can remove this asm block, so this program must have UB. Our current docs do not explain where that UB comes from.

src/inline-assembly.md
```diff
@@ -413,6 +413,7 @@ The list of clobbered registers for each ABI is updated in rustc as architecture
 Flags are used to further influence the behavior of the inline assembly block.
 Currently the following options are defined:
 - `pure`: The `asm!` block has no side effects, and its outputs depend only on its direct inputs (i.e. the values themselves, not what they point to) or values read from memory (unless the `nomem` options is also set).
+  The `asm!` block must always terminate.
```
Amanieu, line 416:
```suggestion
- `pure`: The `asm!` block has no side effects, must eventually return, and its outputs depend only on its direct inputs (i.e. the values themselves, not what they point to) or values read from memory (unless the `nomem` options is also set).
```
This reads better when placed with the other requirements in the first sentence. Feel free to adjust the exact wording though.

Amanieu (approved):

Amanieu (approved):

ehuss (approved):
Thanks!

I'm going to go ahead and approve, but I'm a little uncomfortable with the process we've been using for defining assembly and its behavior. It feels very ad-hoc, without any oversight other than Amanieu. It's not that I don't trust @Amanieu (I trust him more than anyone else, and am very appreciative of his help), but that I think that having a very large surface area of Rust be defined by essentially one person without any discussion or broad feedback can lead to mistakes or disagreements or misunderstandings.

I don't know how else to handle this since I have difficulty getting the lang team to participate.

PR #1444, opened 2023-12-29
Use proper footnote for number literals `_` separator note

Jules-Bertholet:


ehuss (approved):
Thanks! I also pushed a change to add a footnote for another one on the same page.

PR #1447, opened 2024-01-13
fix: Add explanation for counterintuitiveness of lifetime subtyping

gabyx:
- Copied from Rust 1.25 manual where this essiential paragraph to clear things up has unfortunately vanished. Reintroduce it with a refurbished version.

Source: https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/nomicon/subtyping.html


traviscross:
Thanks for the suggestion, but we're going to go ahead and close this.  The text here would need to be substantially reworked to fit correctly in this spot in the Reference.

PR #1449, opened 2024-01-18
Add the `#[diagnostic]` attribute namespace and the `#[diagnostic::on_unimplemented]` feature to the reference

weiznich:
As requested [here](https://github.com/rust-lang/rust/pull/119888#issuecomment-1889492631)


weiznich:
I'm not really sure what needs to be done to address the CI errors as the name space is in fact currently unstable but it was requested during the stablization PR to fill this PR.

ehuss:
The CI errors are fine, this won't be merged until the stabilization hits nightly which should resolve them.

ehuss (commented):
Thanks!

compiler-errors:
This can/should get merged now

src/attributes/diagnostics.md
```diff
@@ -301,6 +301,72 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+* `label` — The text for the label shown inline in the broken code in the error message.
+* `note` — Provides additional notes.
+
+The `note` option can appear several times, which results in several note messages being emitted.
+If any of the other options appears several times the first occurrence of the relevant option specifies the actually used value.
+Any other occurrence generates an lint warning.
+For any other non-existing option a lint-warning is generated.
+
+All three options accept a string as an argument.
+The text in the string may contain the following format parameters which provide substitutions in the generated message:
```
ehuss, line 329: If this is the actual intention here, then I think an edit like the following would make sense:

```suggestion
All three options accept a string as an argument, interpreted using the same formatting as a [`std::fmt`] string.
Format parameters with the given named parameter will be replaced with the following text:
```

src/attributes/diagnostics.md
```diff
@@ -301,6 +301,72 @@ When used on a function in a trait implementation, the attribute does nothing.
...
+
+* `{Self}` — The name of the type implementing the trait.
+* `{` *GenericParameterName* `}` — The name of the generic argument's type for the given generic parameter.
+
```
ehuss, line 333:
```suggestion

Invalid format strings may generate a warning, but are otherwise allowed, but may not display as intended.
Format specifiers may generate a warning, but are otherwise ignored.

```

src/attributes/diagnostics.md
weiznich: That sound reasonable for me and that mirrors the actual implementation after the fix implemented in https://github.com/rust-lang/rust/pull/122402

src/attributes/diagnostics.md
weiznich: That sound reasonable for me and that mirrors the actual implementation after the fix implemented in https://github.com/rust-lang/rust/pull/122402

ehuss (approved):
Thanks!

Issue #1448, opened 2024-01-17
Document behavior of wide to thin pointer casts for DSTs


udoprog:
Related #66, rust-lang/rust#81513

The following casts are considered ["true casts" per the reference](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions):
* `*const [T] as *const U`
* `*const dyn T as *const U`

Today they are implemented so that the *data pointer* is returned and *metadata discarded*. However, this behavior does not appear to be documented anywhere authoritative.

In the [rfc to `ptr_metadata`](https://rust-lang.github.io/rfcs/2580-ptr-meta.html#reference-level-explanation) (Tracking issue: rust-lang/rust#81513) the following text was proposed, but as of today hasn't made it into the unstable Pointee trait:
> Pointer metadata can be extracted from a pointer or reference with the `metadata` function. The data pointer can be extracted by casting a (fat) pointer to a (thin) pointer to a `Sized` type with the `as` operator, for example `(x: &dyn SomeTrait) as *const SomeTrait as *const ()` or `(x: *const dyn SomeTrait).cast::<()>()`.

Should this be documented in the reference or is the semantics contentious?

RalfJung:
This is not contentious at all. But indeed it would be better to state this explicitly.

PR #1451, opened 2024-01-22
Clarify semantics of the various pointer to pointer casts

udoprog:
Relates #1448

I've tried to the best of my ability to reproduce the verbiage of the reference, some parts feel a bit wordy to me so any feedback would be appreciated.

To fully describe the semantics of unsized to sized pointer casts, I felt it was necessary to document that the reinterpreting sized to sized casts produce identical values. I couldn't immediately find this documented.

This is because unsized to sized pointer casts also permits a reintepreting cast in the same sweep (such as `*const [u8] as *const u32`).

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -695,3 +695,17 @@ See [this test] for an example of using this dependency.
 [_TypeNoBounds_]: ../types.md#type-expressions
 [_RangeExpression_]: ./range-expr.md
 [_UnderscoreExpression_]: ./underscore-expr.md
+
+<script>
+(function() {
+    var fragments = {
+        "#slice-dst-pointer-to-pointer-cast": "operator-expr.html#pointer-to-pointer-cast",
+    };
+    var target = fragments[window.location.hash];
+    if (target) {
+        var url = window.location.toString();
+        var base = url.substring(0, url.lastIndexOf('/'));
+        window.location.replace(base + "/" + target);
+    }
+})();
+</script>
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -478,15 +478,23 @@ unsafe {
 assert_eq!(values[1], 3);
 ```
 
-#### Slice DST pointer to pointer cast
+#### Pointer-to-pointer cast
 
-For slice types like `[T]` and `[U]`, the raw pointer types `*const [T]`, `*mut [T]`,
-`*const [U]`, and `*mut [U]` encode the number of elements in this slice. Casts between
-these raw pointer types preserve the number of elements. Note that, as a consequence,
-such casts do *not* necessarily preserve the size of the pointer's referent (e.g.,
-casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an
-object of half the size of the original). The same holds for `str` and any compound type
-whose unsized tail is a slice type, such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+`*const T` / `*mut T` can be cast to `*const U` / `*mut U` with the following behavior:
+
+- If `T` and `U` are both sized, the pointer is returned unchanged.
+- If `T` and `U` are both unsized, the pointer is also returned unchanged. In particular,
+  the metadata is preserved exactly.
+
+    For instance, a cast from `*const [T]` to `*const [U]` preserves the number of elements.
+  Note that, as a consequence, such casts do not necessarily preserve the size of the
+  pointer's referent (e.g., casting `*const [u16]` to `*const [u8]` will result in a raw
+  pointer which refers to an object of half the size of the original). The same
+  holds for `str` and any compound type whose unsized tail is a slice type, such
+  as `struct Foo(i32, [u8])` or `(u64, Foo)`.
+- If `T` is unsized and `U` is sized, the cast discards all metadata that
+  completes the wide pointer `T` and produces a thin pointer `U` consisting
+  of the data part of the unsized pointer.
 
 ## Assignment expressions
 
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -483,18 +483,15 @@ assert_eq!(values[1], 3);
 `*const T` / `*mut T` can be cast to `*const U` / `*mut U` with the following behavior:
 
 - If `T` and `U` are both sized, the pointer is returned unchanged.
-- If `T` and `U` are both unsized, the pointer is also returned unchanged. In particular,
-  the metadata is preserved exactly.
+- If `T` and `U` are both unsized, the pointer is also returned unchanged.
+  In particular, the metadata is preserved exactly.
 
-    For instance, a cast from `*const [T]` to `*const [U]` preserves the number of elements.
-  Note that, as a consequence, such casts do not necessarily preserve the size of the
-  pointer's referent (e.g., casting `*const [u16]` to `*const [u8]` will result in a raw
-  pointer which refers to an object of half the size of the original). The same
-  holds for `str` and any compound type whose unsized tail is a slice type, such
-  as `struct Foo(i32, [u8])` or `(u64, Foo)`.
-- If `T` is unsized and `U` is sized, the cast discards all metadata that
-  completes the wide pointer `T` and produces a thin pointer `U` consisting
-  of the data part of the unsized pointer.
+  For instance, a cast from `*const [T]` to `*const [U]` preserves the number of elements.
+  Note that, as a consequence, such casts do not necessarily preserve the size of the pointer's referent
+  (e.g., casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an object of half the size of the original).
+  The same holds for `str` and any compound type whose unsized tail is a slice type,
+  such as `struct Foo(i32, [u8])` or `(u64, Foo)`.
+- If `T` is unsized and `U` is sized, the cast discards all metadata that completes the wide pointer `T` and produces a thin pointer `U` consisting of the data part of the unsized pointer.
 
 ## Assignment expressions
 


ehuss:
@RalfJung Can you review this?

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,48 @@ unsafe {
...
+Casts between these raw pointer types preserve the number of elements. Note
+that, as a consequence, such casts do *not* necessarily preserve the size of the
+pointer's referent (e.g., casting `*const [u16]` to `*const [u8]` will result in
+a raw pointer which refers to an object of half the size of the original). The
+same holds for `str` and any compound type whose unsized tail is a slice type,
+such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+
+#### Unsized to sized pointer casts
+
+Any wide to narrow pointer cast discards all metadata that completes the wide
```
ehuss, line 504: The word "metadata" isn't defined anywhere. Can you add a description here that defines what it means?

udoprog: I don't know if I'm just promoting kicking the can down the road, but I opted not to define metadata since:
* It's not necessary to define the semantics of this cast, the only distinction is that unsized pointers have metadata and this cast removes it.
* Another part of the reference "behaviors considered undefined" uses metadata without defining it.

That being said, if I were to define it, where do you think I should do it and to what extent?

> The representation of pointers and references to DSTs is made up of two components; A base pointer and metadata. The metadata is the necessary extra data that has to be attached to pointers and references of DSTs which completes their representation. This is stored adjacent to the base pointer causing them to be wider than sized pointers earning them the name "wide" or "fat" pointers. The base pointer can be extracted through an unsized to sized pointer cast.

I'm a bit concerned that I'll end up having to define everything metadata adjacent 😓, so at least I'd want to avoid that as much as possible.

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,48 @@ unsafe {
...
-For slice types like `[T]` and `[U]`, the raw pointer types `*const [T]`, `*mut [T]`,
-`*const [U]`, and `*mut [U]` encode the number of elements in this slice. Casts between
-these raw pointer types preserve the number of elements. Note that, as a consequence,
-such casts do *not* necessarily preserve the size of the pointer's referent (e.g.,
-casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an
-object of half the size of the original). The same holds for `str` and any compound type
-whose unsized tail is a slice type, such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+A sized pointer to pointer cast `*const T as *const U` or `*mut T as *mut U`
+where `T: Sized` and `U: Sized` produces a value identical to the original
+pointer.
```
RalfJung, line 485: We also allow const-to-mut and mut-to-const

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,48 @@ unsafe {
...
+
+#### Unsized to unsized pointer casts
+
+Any unsized to unsized pointer cast `*const T as *const U` or `*mut T as *mut U`
+produces a value identical to the original pointer if `T` and `U` are compatible
+unsized types.
```
RalfJung, line 491: This then raises the question, what does it do if they are not compatible?

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,48 @@ unsafe {
...
+
+The following casts are well-defined:
+
+* `*const [T] to *const U` or `*mut [T] to *mut U` discards the length of the
+  slice and produces a pointer to the first element. The produced pointer is
+  casted to `U` in the same manner as a sized pointer to pointer cast.
+* `*const dyn T to *const U` or `*mut dyn T to *mut U` discards all metadata
+  associated with the [trait object] and produces a pointer to the instance from
+  which the trait object was constructed. The produced pointer is casted to `U`
+  in the same manner as a sized pointer to pointer cast.
```
RalfJung, line 515: This list is not exhaustive, e.g. it misses this case:
```rust
fn test(x: *const (u8, str)) -> *const u16 { x as _ }
```
Also, it seems strange to list both of these only to then say they do basically the same. The enumeration just kind of repeats what the text above says?

RalfJung:
I think this might be better structured in a different way... I'd make a single big section for raw-to-raw casts, which also handles once and for all the mut/const part: 
```
#### Pointer-to-pointer cast

`*const T`/`*mut T` can be cast to `*const U`/`*mut U` under the following conditions:

- If both `T` and `U` are sized, [...]
- If both `T` and `U` are unsized, [...]
- If `T` is unsized and `U` is sized, [...]

File: src/expressions/operator-expr.md (hunk 2 above)
Comment (udoprog) [reply src/expressions/operator-expr.md:485]: What's a clear way to enumerate all permutations concisely?

File: src/expressions/operator-expr.md (hunk 3 above)
Comment (udoprog) [reply src/expressions/operator-expr.md:491]: Do we have to answer that now? I'd like to avoid pulling as many threads as possible. 😅

File: src/expressions/operator-expr.md (hunk 4 above)
Comment (udoprog) [reply src/expressions/operator-expr.md:515]: So you either want to also enumerate unsized tuples and presumably unsized custom types or collapse all of them into one?

It would be nice to list the common instances (`*const [T]` / `*const dyn T`) somehow, but maybe not necessary?

I would like to defer to your preference on this matter, anything works for me as long as the unsized pointer cast semantics are documented.

Comment (udoprog) [reply src/expressions/operator-expr.md:515]: Ah, saw the last comment now.

File: src/expressions/operator-expr.md (hunk 2 above)
Comment (udoprog) [reply src/expressions/operator-expr.md:485]: Saw the last comment.

File: src/expressions/operator-expr.md (hunk 3 above)
Comment (RalfJung) [reply src/expressions/operator-expr.md:491]: The PR in its current form introduces a confusing new sentence. So yes this needs to be fixed.

But I think the structure I propose will also handle this pretty naturally.

File: src/expressions/operator-expr.md (hunk 4 above)
Comment (RalfJung) [reply src/expressions/operator-expr.md:515]: > It would be nice to list the common instances (*const [T] / *const dyn T) somehow, but maybe not necessary?

This is a reference, not tutorial material. So no, common cases should not be listed. Every special case listed should only be listed because it actually is a special case that behaves different from the general case.

Comment (RalfJung) [reply src/expressions/operator-expr.md:515]: Fair. If you feel it's useful, add something like that -- but the "for example" wording is crucial here.

However, I'm not convinced that an example is actually helpful here. This will be easier to discuss once there is concrete text for the revised version. :)

File: src/expressions/operator-expr.md (hunk 3 above)
Comment (udoprog) [reply src/expressions/operator-expr.md]: Is the sentence "compatible unsized types"? If so, that's not my phrasing. It's borrowed from the table just above.

Comment (RalfJung) [reply src/expressions/operator-expr.md]: Ah, I see. But in the table it is clear that if the types are not compatible, then the cast is simply not permitted.

Comment (udoprog) [reply src/expressions/operator-expr.md]: Ok, so I've removed the double reference to validity since it's already covered by the table and rewrote it to only focus on behavior. Please check it out!

File: src/expressions/operator-expr.md [hunk 5]
```diff
@@ -483,7 +483,8 @@ assert_eq!(values[1], 3);
 `*const T` / `*mut T` can be cast to `*const U` / `*mut U` with the following behavior:
 
 - If `T` is unsized and `U` is sized, the cast discards all metadata that
-  completes the wide pointer `T` and produces a thin base pointer `U`.
+  completes the wide pointer `T` and produces a thin pointer `U` to the
+  value of the unsized pointer.
```
udoprog, line 487: I'm struggling to find a phrasing here for to describe what "value" is which applies to all unsized pointers and also conveys *what exactly it points to*. In the reference for trait objects "instance" is used, but that would be a bit unorthodox for slices.

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,15 @@ unsafe {
...
-casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an
-object of half the size of the original). The same holds for `str` and any compound type
-whose unsized tail is a slice type, such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+#### Pointer-to-pointer cast
+
+`*const T` / `*mut T` can be cast to `*const U` / `*mut U` with the following behavior:
+
+- If `T` is unsized and `U` is sized, the cast discards all metadata that
+  completes the wide pointer `T` and produces a thin pointer `U` to the
+  value of the unsized pointer.
```
RalfJung, line 487:
```suggestion
- If `T` is unsized and `U` is sized, the cast discards all metadata that
  completes the wide pointer `T` and produces a thin pointer `U` consisting
  of the data part of the unsized pointer.
```

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,15 @@ unsafe {
...
-these raw pointer types preserve the number of elements. Note that, as a consequence,
-such casts do *not* necessarily preserve the size of the pointer's referent (e.g.,
-casting `*const [u16]` to `*const [u8]` will result in a raw pointer which refers to an
-object of half the size of the original). The same holds for `str` and any compound type
-whose unsized tail is a slice type, such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+#### Pointer-to-pointer cast
+
+`*const T` / `*mut T` can be cast to `*const U` / `*mut U` with the following behavior:
+
+- If `T` is unsized and `U` is sized, the cast discards all metadata that
```
RalfJung, line 485:
```suggestion
- If `T` and `U` are both sized, the pointer is returned unchanged.
- If `T` and `U` are both unsized, the pointer is also returned unchanged. In particular,
  the metadata is preserved exactly.

    For instance, a cast from `*const [T]` to `*const [U]` preserves the number of elements.
  Note that, as a consequence, such casts do not necessarily preserve the size of the
  pointer's referent (e.g., casting `*const [u16]` to `*const [u8]` will result in a raw
  pointer which refers to an object of half the size of the original). The same
  holds for `str` and any compound type whose unsized tail is a slice type, such
  as `struct Foo(i32, [u8])` or `(u64, Foo)`.
- If `T` is unsized and `U` is sized, the cast discards all metadata that
```

src/expressions/operator-expr.md
```diff
@@ -478,15 +478,25 @@ unsafe {
...
+  Note that, as a consequence, such casts do not necessarily preserve the size of the
+  pointer's referent (e.g., casting `*const [u16]` to `*const [u8]` will result in a raw
+  pointer which refers to an object of half the size of the original). The same
+  holds for `str` and any compound type whose unsized tail is a slice type, such
+  as `struct Foo(i32, [u8])` or `(u64, Foo)`.
+- If `T` is unsized and `U` is sized, the cast discards all metadata that
+  completes the wide pointer `T` and produces a thin pointer `U` consisting
+  of the data part of the unsized pointer.
+- Otherwise a cast from `T` to `U` produces a value identical to the original
+  pointer.
```
RalfJung, line 499:
```suggestion
```
This is now covered by the first two cases.

RalfJung (commented):
LGTM, but since I now largely rewrote this it should be reviewed by someone else.

udoprog:
For what it's worth, you have my +1.

udoprog:
This seems a bit stalled. Since I'm pretty much OK with this change, can we say that I wrote it and move forward? 😄

I think this would pretty much emulate the same scenario as if this was fully an external change proposed by me which was reviewed by Ralf. Or maybe @ehuss can give this a quick look over?

RalfJung:
@udoprog please squash the PR into a single commit.

But yeah, for review would be good if @ehuss took a look. Maybe someone from @rust-lang/opsem can help as well.

digama0 (approved):

udoprog:
Done

ehuss (approved):
Thanks!

PR #1452, opened 2024-01-22
String literal expressions

mattheww:
This fills in most of what is still missing for #939.

I haven't tried to cover C string literals at this stage, as they haven't quite settled down yet.

I haven't removed much of the text in `tokens.md` which is discussing what string literal tokens "represent" or "denote", although I think strictly speaking that applies to literal expressions rather than literal tokens.

I think the best course would be think about that after, or together with, making the Tokens chapter correct (which is #626).

### DIFF
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -87,6 +87,25 @@ For this purpose, the whitespace characters are `U+0009` (HT), `U+000A` (LF), `U
 
 The escaped value is an empty sequence of characters.
 
+> **Note**: The effect of this form of escape is that a string continuation skips following whitespace, including additional newlines.
+> Thus `a`, `b` and `c` are equal:
+> ```rust
+> let a = "foobar";
+> let b = "foo\
+>          bar";
+> let c = "foo\
+>
+>      bar";
+>
+> assert_eq!(a, b);
+> assert_eq!(b, c);
+> ```
+>
+> Skipping additional newlines (as in example c) is potentially confusing and unexpected.
+> This behavior may be adjusted in the future.
+> Until a decision is made, it is recommended to avoid relying on skipping multiple newlines with line continuations.
+> See [this issue](https://github.com/rust-lang/reference/pull/1042) for more information.
+
 ## Character literal expressions
 
 A character literal expression consists of a single [CHAR_LITERAL] token.
--- a/src/expressions/literal-expr.md
+++ b/src/expressions/literal-expr.md
@@ -26,29 +26,208 @@ Each of the lexical [literal][literal tokens] forms described earlier can make u
 5;         // integer type
 ```
 
+In the descriptions below, the _string representation_ of a token is the sequence of characters from the input which matched the token's production in a *Lexer* grammar snippet.
+
+> **Note**: this string representation never includes a character `U+000D` (CR) immediately followed by `U+000A` (LF): this pair would have been previously transformed into a single `U+000A` (LF).
+
+## Escapes
+
+The descriptions of textual literal expressions below make use of several forms of _escape_.
+
+Each form of escape is characterised by:
+ * an _escape sequence_: a sequence of characters, which always begins with `U+005C` (`\`)
+ * an _escaped value_: either a single character or an empty sequence of characters
+
+In the definitions of escapes below:
+ * An _octal digit_ is any of the characters in the range \[`0`-`7`].
+ * A _hexadecimal digit_ is any of the characters in the ranges \[`0`-`9`], \[`a`-`f`], or \[`A`-`F`].
+
+### Simple escapes
+
+Each sequence of characters occurring in the first column of the following table is an escape sequence.
+
+In each case, the escaped value is the character given in the corresponding entry in the second column.
+
+| Escape sequence | Escaped value            |
+|-----------------|--------------------------|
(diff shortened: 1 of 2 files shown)


src/expressions/literal-expr.md
```diff
@@ -26,29 +26,227 @@ Each of the lexical [literal][literal tokens] forms described earlier can make u
...
+The expression's type is a shared reference (with `static` lifetime) to the primitive [`str`][textual types] type.
+That is, the type is `&'static str`.
+
+The token must not have a suffix.
+
+The token's _literal content_ is the sequence of characters following the first `U+0022` (`"`) and preceding the last `U+0022` (`"`) in the string representation of the token.
+
+The literal expression's _represented string_ is a sequence of characters derived from the literal content as follows:
+
+* If the token is a [STRING_LITERAL], each escape sequence of any of the following forms occuring in the literal content is replaced by the escape sequence's escaped value.
```
ehuss, line 154:
```suggestion
* If the token is a [STRING_LITERAL], each escape sequence of any of the following forms occurring in the literal content is replaced by the escape sequence's escaped value.
```

src/expressions/literal-expr.md
```diff
@@ -26,29 +26,227 @@ Each of the lexical [literal][literal tokens] forms described earlier can make u
...
+The expression's type is a shared reference (with `static` lifetime) to an array whose element type is [`u8`][numeric types].
+That is, the type is `&'static [u8; N]`, where `N` is the number of characters in the represented string described below.
+
+The token must not have a suffix.
+
+The token's _literal content_ is the sequence of characters following the first `U+0022` (`"`) and preceding the last `U+0022` (`"`) in the string representation of the token.
+
+The literal expression's _represented string_ is a sequence of characters derived from the literal content as follows:
+
+* If the token is a [BYTE_STRING_LITERAL], each escape sequence of any of the following forms occuring in the literal content is replaced by the escape sequence's escaped value.
```
ehuss, line 224:
```suggestion
* If the token is a [BYTE_STRING_LITERAL], each escape sequence of any of the following forms occurring in the literal content is replaced by the escape sequence's escaped value.
```

src/expressions/literal-expr.md
```diff
@@ -26,29 +26,227 @@ Each of the lexical [literal][literal tokens] forms described earlier can make u
...
+```
 
 ## Byte string literal expressions
 
-A string literal expression consists of a single [BYTE_STRING_LITERAL] or [RAW_BYTE_STRING_LITERAL] token.
+A byte string literal expression consists of a single [BYTE_STRING_LITERAL] or [RAW_BYTE_STRING_LITERAL] token.
 
-> **Note**: This section is incomplete.
+The expression's type is a shared reference (with `static` lifetime) to an array whose element type is [`u8`][numeric types].
+That is, the type is `&'static [u8; N]`, where `N` is the number of characters in the represented string described below.
```
ehuss, line 216: Perhaps to avoid confusion here?

```suggestion
That is, the type is `&'static [u8; N]`, where `N` is the number of bytes in the represented string described below.
```

Also, this may become a little more confusing after [RFC 3349](https://rust-lang.github.io/rfcs/3349-mixed-utf8-literals.html) is implemented/stabilized.

src/tokens.md
```diff
@@ -156,30 +156,13 @@ A _string literal_ is a sequence of any Unicode characters enclosed within two
...
-Line-breaks are allowed in string literals. A line-break is either a newline
-(`U+000A`) or a pair of carriage return and newline (`U+000D`, `U+000A`). Both
-byte sequences are normally translated to `U+000A`, but as a special exception,
-when an unescaped `U+005C` character (`\`) occurs immediately before a line
-break, then the line break character(s), and all immediately following
-` ` (`U+0020`), `\t` (`U+0009`), `\n` (`U+000A`) and `\r` (`U+0000D`) characters
-are ignored. Thus `a`, `b` and `c` are equal:
+Line-breaks are allowed in string literals.
+A line-break is either a newline (`U+000A`) or a pair of carriage return and newline (`U+000D`, `U+000A`).
+Both byte sequences are translated to `U+000A`.
```
ehuss, line 161: I think this is fine to keep. Just a note that my preference longer-term would be to have some global explanation of the CRLF→LF conversion in a single place so it doesn't need to be repeated in multiple places. For example, this is relevant for all string types (raw strings, byte strings, c-strings, etc.), but they don't currently mention this in the same way.

If there is a section somewhere in the lexing chapter that talks about this translation, I think these sentences could be removed.

ehuss:
Thanks!

I think it would be fine to also include the c-string section if you want. I believe the discussion has settled on that.

mattheww:
I've force-pushed with your suggested changes.

I think I'd better look at C strings separately. I think I'm going to need some help with the "The expression's value is …" part.

src/expressions/literal-expr.md
mattheww: After RFC 3349 I think the byte and byte-string literals will need a different approach, with a separate version of the whole "Escapes" section. But that will be easier to describe when the new forms of escape are there.

src/tokens.md
mattheww: I'm thinking of looking at documenting CRLF conversion properly next (rather than trying to fight all of #626 in one go).

ehuss (approved):
Thanks!

PR #1453, opened 2024-01-24
'.inst' in inline-assembly changed to '.insn'

CyoeeA9e:
The correct directive name should be `.insn`. [Here is the issue in rust repository.](https://github.com/rust-lang/rust/issues/90558)

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -570,6 +570,7 @@ The following directives are guaranteed to be supported by the assembler:
 - `.global`
 - `.globl`
 - `.inst`
+- `.insn`
 - `.lcomm`
 - `.long`
 - `.octa`


ehuss:
I'm a bit confused, as I believe `.inst` is a valid directive still. What is the difference between them? Is there some more information about this change?

@Amanieu Can you provide some information here?

Amanieu:
Both are correct: `.inst` is used on ARM, `.insn` is used in RISC-V.

ehuss:
@joshtriplett Do you want to *add* `.insn` to the guaranteed directive list? (Not what this PR does, it should be *added* not replaced.)

Looks like .insn was added to gnu-as in 2018 in https://github.com/bminor/binutils-gdb/commit/0e35537d754f1c687850d1caccb2d78d2e418391 and in llvm in llvm-14 in https://github.com/llvm/llvm-project/commit/283879793dc787225992496587581ec77b6b0610 (I think, from some quick googling).

Amanieu:
I'm happy to add it in addition to `.inst`.

joshtriplett:
@ehuss Yeah, that seems reasonable.

ehuss (approved):
Pushed an update so that it is adding instead of replacing. Thanks!

PR #1454, opened 2024-01-25
type-layout.md: Guarantee that repr(C) enum discriminants are stable …

joshlf:
…within a compilation

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -304,6 +304,11 @@ the default `enum` size and alignment for the target platform's C ABI.
 > Note: The enum representation in C is implementation defined, so this is
 > really a "best guess". In particular, this may be incorrect when the C code
 > of interest is compiled with certain flags.
+>
+> However, it is guaranteed that, for any two `repr(C)` enum types, their
+> discriminants will have the same size and alignment *within a given
+> compilation of the program.* This is true for both field-less enums and
+> for enums with fields.
 
 <div class="warning">
 


joshlf:
cc @djkoloski @jswrenn

joshlf:
Friendly ping 🙂

joshlf:
Friendly ping :)

ehuss:
Is there some context for this change? Has it been discussed by the lang team?

src/type-layout.md
```diff
@@ -304,6 +304,11 @@ the default `enum` size and alignment for the target platform's C ABI.
 > Note: The enum representation in C is implementation defined, so this is
 > really a "best guess". In particular, this may be incorrect when the C code
 > of interest is compiled with certain flags.
+>
+> However, it is guaranteed that, for any two `repr(C)` enum types, their
+> discriminants will have the same size and alignment *within a given
+> compilation of the program.* This is true for both field-less enums and
+> for enums with fields.
```
ehuss, line 311: "within a given compilation of the program" doesn't seem clear to me what that means. Certainly all types have the same size and alignment during compilation. I'm not sure I can guess what this is trying to say, can you clarify?

joshlf: Maybe I should phrase this differently. The idea is that, whatever choice rustc makes about the size and alignment of the discriminant of a `repr(C)` enum, it makes the same choice for all enums within a program.

The reason this is important is that the size and alignment aren't guaranteed in any way other than to say "rustc will choose some size and alignment." This puts unsafe code in a bind: how do you know what the size and alignment of the discriminant is if you're trying to work with the layout of an enum type? With this added text, you can construct a "dummy"  field-less `repr(C)` enum, query for *its* size and alignment using `size_of` and `align_of`, and take these to be equal to the size and alignment of the discriminant of the enum you're working with.

The specific context is that zerocopy's [`TryFromBytes`](https://docs.rs/zerocopy/0.8.0-alpha.14/zerocopy/trait.TryFromBytes.html) trait can currently be derived on field-less enums, and we want to add support for data-ful enums, but we can't without this guarantee (or some other way of querying for the size and alignment of `repr(C)` enum discriminants).

joshlf:
> Has it been discussed by the lang team?

No, but I figured it'd be pretty uncontroversial. Happy to CC them here if you'd like.

src/type-layout.md
traviscross: @joshlf: We discussed this in the rustdocs call today.  This is going to need to be lang nominated, but before we do that, we'd like this language to be as clear as possible.  Do you want to take another pass at stating this more clearly as discussed?

traviscross:
@joshlf: Before we lang nominate it, could you update the PR description with a summary of the change being proposed here and describe in some detail the motivations for this change.  E.g., if you could go into more detail about the situation you have with the `zerocopy` library and what the possible solutions for that problem are (especially if there are any other ones than the one proposed here), that would be helpful.

src/type-layout.md
bjorn3, line 311: You probably mean the tag, not the discriminant. The discriminant is the value you get when you cast an enum to an integer or call `std::mem::discriminant` on it. The tag is the actual representation in memory. It may be smaller or larger than the discriminant. Without `repr(C)` or `repr(iN)` the tag may be a different integer than the discriminant or the discriminant may be represented using niches instead.

Issue #1445, opened 2024-01-01
`Self` receiver does *not* imply `Self: Sized`


QuineDot:
[The object safety page says](https://doc.rust-lang.org/nightly/reference/items/traits.html#object-safety)

> * All associated functions must either be dispatchable from a trait object or be explicitly non-dispatchable:
>     * Dispatchable functions must:
>         * [...]
>         * Not have a `where Self: Sized` bound (receiver type of `Self` (i.e. `self`) implies this).
>     * Explicitly non-dispatchable functions require:
>         * Have a `where Self: Sized` bound (receiver type of `Self` (i.e. `self`) implies this).

However, [it is not true that a receiver type of `Self` implies `Self: Sized`.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d0fee50714d4b9c7890fe44d7ac0df38)
```rust
// Minimal example; it also can be demonstrated with bounds that
// require `Sized`, like `<G: From<Self>>` say, even without a
// function body
trait Trait {
    // Fails without an explicit `where Self: Sized` bound
    fn method(self) {}
}
```

Accepting `self`-receiving methods as object-safe methods was [a separate consideration from a `Self: Sized` bound in the RFC.](https://rust-lang.github.io/rfcs/0255-object-safety.html#detailed-design)

> This is why all methods that require Self:Sized are exempt from the typical rules. This is also why by-value self methods are permitted, since currently one cannot invoke pass an unsized type by-value (though we consider that a useful future extension).

ehuss:
Thanks for the report! I believe this is a duplicate of #1247, so closing in favor of that.

PR #1455, opened 2024-01-26
Document how `Self` receiver types *actually* interact with object safety

### LINKED ISSUE #1247: Rule around non-dispatchable functions doesn't match the compiler behavior

steffahn:
Fixes #1247

The description here on how this works is based purely on experiments on what the compiler does or does not accept (without unstable features), not on any inspection of the implementation in `rustc`. The term of “implicitly non-dispatchable functions” is one I just made up, as far as I’m aware, because it makes sense in this context. The motivation of future-compatibility is just my guess, too.

I might want to read up some actual sources around relevant RFCs and/or PRs, for precise (intended) behavior, terminology, and motivation of the behavior of `self: Self` but not `Self: Sized` methods in traits.

For review, I’d thus also appreciate links to relevant discussions that I haven’t found yet.


steffahn:
cc @QuineDot since you recently opened #1445, so you might have some ideas or feedback here

QuineDot:
See [this PR](https://github.com/rust-lang/rust/pull/125380) -- it's not just `Self: Sized` that makes a method non-dispatchable, it's `TypeInvolvingSelf: NonAutoTrait`... plus some others.  [See also.](https://github.com/rust-lang/rust/issues/50781#issuecomment-400997165)

I haven't experimented enough to feel :100: confident in what is and isn't accepted, and there is some wonkiness around the check.  Such as the "empty bounds" in the examples below...

Example `where` clauses which are fine (do not make the method non-dispatchable):
- `Self: Send`
- `Self: 'a`
- `Self: `
- `Self::AssociatedType: Debug`
- `Box<Self>: 'a`

Example `where` clauses which make the method non-dispatchable:
- `Self: Debug`
- `Box<Self>: Debug`
- `Box<Self>: Send`
- `Box<Self>:`  :shrug: 
- `dyn Trait<Assoc = Box<Self>>: Debug`

Tangential aside: `Self: Sized` in an implied bound (e.g. `where Self: Clone`) also counts as explicitly non-dispatchable.

traviscross:
We reviewed this PR in the rustdocs call, and we've asked for review of this by the types team in:

- https://github.com/rust-lang/rust/issues/127251

src/items/traits.md
```diff
@@ -71,7 +71,8 @@ Object safe traits can be the base trait of a [trait object]. A trait is
 * `Sized` must not be a [supertrait][supertraits]. In other words, it must not require `Self: Sized`.
 * It must not have any associated constants.
 * It must not have any associated types with generics.
-* All associated functions must either be dispatchable from a trait object or be explicitly non-dispatchable:
+* All associated functions must either be dispatchable from a trait object,
+  or be explicitly or implicitly non-dispatchable
```
compiler-errors, line 75:
```suggestion
  or be explicitly or implicitly non-dispatchable.
```

src/items/traits.md
```diff
@@ -71,7 +71,8 @@ Object safe traits can be the base trait of a [trait object]. A trait is
 * `Sized` must not be a [supertrait][supertraits]. In other words, it must not require `Self: Sized`.
 * It must not have any associated constants.
 * It must not have any associated types with generics.
-* All associated functions must either be dispatchable from a trait object or be explicitly non-dispatchable:
+* All associated functions must either be dispatchable from a trait object,
+  or be explicitly or implicitly non-dispatchable.
```
lcnr, line 75: that terminology feels weird, `fn foo<T>` is also "implicitly non-dispatchable", isn't it?

All associated functions must either be dispatchable from a trait object or must not be applicable to trait objects
```suggestion
or must not be applicable to trait objects.
```

src/items/traits.md
```diff
@@ -81,13 +82,26 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
             * Not have a return position `impl Trait` type (`fn example(&self) -> impl Trait`).
-        * Not have a `where Self: Sized` bound (receiver type of `Self` (i.e. `self`) implies this).
-    * Explicitly non-dispatchable functions require:
-        * Have a `where Self: Sized` bound (receiver type of `Self` (i.e. `self`) implies this).
+        * Not have a `where Self: Sized` bound.
+    * Explicitly non-dispatchable functions must:
+        * Have a `where Self: Sized` bound.
+    * Implicitly non-dispatchable functions must:
+        * Have the receiver type `Self` (i.e. `self`)
+        * Fulfill all the conditions of dispatchable functions, except for the receiver type.
```
lcnr, line 94:
```suggestion
    * Non-applicable functions have either:
        * a `where Self: Sized` bound.
        * the receiver type `Self` (i.e. `self`) and fulfill all the conditions of dispatchable functions, except for the receiver type.
```

src/items/traits.md
```diff
@@ -81,13 +82,26 @@ Object safe traits can be the base trait of a [trait object]. A trait is
...
+
+Methods with `Self` receiver type are implicitly non-dispatchable
+(“non-dispatchable” means that they cannot be called on trait object types), but
+[might become dispatchable in future versions of Rust](https://github.com/rust-lang/rust/issues/48055).
```
lcnr, line 98:
```suggestion
Methods with `Self` receiver type are currently non-dispatchable
("non-dispatchable" means that they cannot be called on trait object types), but
[might become dispatchable in future versions of Rust](https://github.com/rust-lang/rust/issues/48055).
```

ehuss:
We pinged on https://rust-lang.zulipchat.com/#narrow/channel/326866-t-types.2Fnominated/topic/.23127251.3A.20Review.20of.20Reference.20change.20regarding.20receives.20and.20.E2.80.A6/with/505464002, and @steffahn indicated they can take a look.

PR #1456, opened 2024-01-26
update patterns.md for const pattern RFC

RalfJung:
Updates the reference for https://github.com/rust-lang/rfcs/pull/3535.
Tracking issue: https://github.com/rust-lang/rust/issues/120362.

This is not 100% implemented yet:
- parts of it are pending on https://github.com/rust-lang/rust/pull/116284
- also the (recursive) structural equality checker in rustc currently sometimes issues a future-compat lint instead of a hard error; this may still take some tome to phase out (we need to first bump that to a "future-compat lint that is also shown in dependencies")
- violations of the `T: PartialEq` currently also just give a future-compat lint, not a hard error

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -776,7 +776,7 @@ They are refutable when they refer to refutable constants or enum variants for e
 When a constant `C` of type `T` is used as a pattern, we first check that `T: PartialEq`.
 Furthermore we require that the value of `C` *has (recursive) structural equality*, which is defined recursively as follows:
 
-- Integers as well as `bool` and `char` values always have structural equality.
+- Integers as well as `str`, `bool` and `char` values always have structural equality.
 - Tuples, arrays, and slices have structural equality if all their fields/elements have structural equality.
   (In particular, `()` and `[]` always have structural equality.)
 - References have structural equality if the value they point to has structural equality.
@@ -787,7 +787,7 @@ Furthermore we require that the value of `C` *has (recursive) structural equalit
 - Nothing else has structural equality.
 
 In particular, the value of `C` must be known at pattern-building time (which is pre-monomorphization).
-This means that associated consts cannot be used as patterns.
+This means that associated consts that involve generic parameters cannot be used as patterns.
 
 After ensuring all conditions are met, the constant value is translated into a pattern, and now behaves exactly as-if that pattern had been written directly.
 In particular, it fully participates in exhaustiveness checking.
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -144,12 +144,6 @@ Since negative numbers are not [literals], literal patterns also accept an optio
 
 <div class="warning">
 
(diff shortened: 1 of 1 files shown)


ehuss:
@RalfJung Is this ready to merge? Were there any changes since this was written?

RalfJung:
We are slowly staging the deployment of this in rustc. Not sure how closely the reference wants to match that.

If you are okay with the reference stating things as required that are just future-compat lints in rustc (rather than hard errors), then this is ready to go, yes.

RalfJung:
FYI, when https://github.com/rust-lang/rust/pull/124661 lands rustc will fully behave as documented here. I hope to land that in this release cycle.

src/patterns.md
```diff
@@ -776,12 +768,30 @@ Unqualified path patterns can refer to:
...
+- Tuples, arrays, and slices have structural equality if all their fields/elements have structural equality.
+  (In particular, `()` and `[]` always have structural equality.)
+- References have structural equality if the value they point to has structural equality.
+- A value of `struct` or `enum` type has structural equality if its `PartialEq` instance is derived via `#[derive(PartialEq)]`,
+  and all fields (for enums: of the active variant) have structural equality.
+- A raw pointer has structural equality if it was defined as a constant integer (and then cast/transmuted).
+- A float value has structural equality if it is not a `NaN`.
+- Nothing else has structural equality.
+
+In particular, the value of `C` must be known at pattern-building time (which is pre-monomorphization).
```
RalfJung, line 789:
```suggestion
In particular, the value of `C` must be known at pattern-building time (which is pre-monomorphization).
This means that associated consts cannot be used as patterns.
```

src/patterns.md
```diff
@@ -776,12 +768,31 @@ Unqualified path patterns can refer to:
...
+This means that associated consts cannot be used as patterns.
```
ehuss, line 790: I don't understand this part. Path patterns can be associated consts. Can you clarify what it is referring to?

src/patterns.md
```diff
@@ -776,12 +768,31 @@ Unqualified path patterns can refer to:
...
-
 Path patterns are irrefutable when they refer to structs or an enum variant when the enum has only one variant or a constant whose type is irrefutable.
 They are refutable when they refer to refutable constants or enum variants for enums with multiple variants.
 
+### Constant patterns
+
+When a constant `C` of type `T` is used as a pattern, we first check that `T: PartialEq`.
+Furthermore we require that the value of `C` *has (recursive) structural equality*, which is defined recursively as follows:
+
+- Integers as well as `bool` and `char` values always have structural equality.
```
ehuss, line 779: And strings?

src/patterns.md
RalfJung: This is intended to explain why this code does not compile:
```rust
trait Trait {
    const C: i32;
}

fn f<T: Trait>(x: i32) -> bool {
    matches!(x, T::C)
}
```

ehuss (approved):
Thanks!

PR #1458, opened 2024-01-28
Improvements to `items/functions.md`

steffahn:
See the commit messages for details.


ehuss (approved):
Thanks!

We decided that we liked the commit about the block terminology, so we are going to merge that. For the wrapping, the style guide for the reference says lines shouldn't be wrapped. For the example, we weren't entirely certain about moving forward with that for now, though we might consider it later.

PR #1459, opened 2024-01-28
Input format

### LINKED ISSUE #1252: Contradictory definition of 'STRING_LITERAL'
[STRING_LITERAL](https://doc.rust-lang.org/reference/tokens.html#string-literals) definition says:
Screenshot from 2022-08-16 22-34-51
Where [_IsolatedCR_](https://doc.rust-lang.org/reference/comments.html#comments) is:
Screenshot from 2022-08-16 22-43-45
And  [ASCII_ESCAPE](https://doc.rust-lang.org/reference/tokens.html#ascii-escapes) is:
Screenshot from 2022-08-16 22-50-17

But look, there is no compile time error:
Screenshot from 2022-08-16 23-16-45

My question:
Is this a human error or some feature of the lexer implementation which I don't know about?

mattheww:
Moves the descriptions of BOM removal, CRLF normalisation, and shebang removal to the "Input format" page.

So the rest of the _Lexical structure_ chapter now assumes CRLF normalisation has already taken place.

Covers some of #626 and #527.

Closes #1252

I think this supersedes #1332

### DIFF
--- a/src/comments.md
+++ b/src/comments.md
@@ -30,7 +30,7 @@
 > &nbsp;&nbsp; | INNER_BLOCK_DOC
 >
 > _IsolatedCR_ :\
-> &nbsp;&nbsp; _A `\r` not followed by a `\n`_
+> &nbsp;&nbsp; \\r
 
 ## Non-doc comments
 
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Fantastic!

TIL that SHEBANG could go on any source file, not just the root crate.

You linked to #527, but I don't think that is the right number. Do you happen to remember which one you were thinking of?

mattheww:
I meant #567, sorry

PR #1460, opened 2024-01-29
Patterns and enums

mattheww:
The Patterns chapter didn't say that struct patterns can match enum values, and didn't document how refutability works for enums.

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -688,7 +688,7 @@ A struct pattern is refutable if the _PathInExpression_ resolves to a constructo
 Tuple struct patterns match tuple struct and enum values that match all criteria defined by its subpatterns.
 They are also used to [destructure](#destructuring) a tuple struct or enum value.
 
-A tuple struct pattern is refutable when one of its subpatterns is refutable.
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -597,8 +597,8 @@ Reference patterns are always irrefutable.
 [_OuterAttribute_]: attributes.md
 [TUPLE_INDEX]: tokens.md#tuple-index
 
-Struct patterns match struct values that match all criteria defined by its subpatterns.
-They are also used to [destructure](#destructuring) a struct.
+Struct patterns match struct and enum values that match all criteria defined by its subpatterns.
+They are also used to [destructure](#destructuring) a struct or enum value.
```
ehuss, line 601: Can this also include `union`?

ehuss (commented):
Thanks! Sorry, this fell off my radar.

src/patterns.md
mattheww: I've added that

ehuss (approved):
Thanks!

PR #1461, opened 2024-01-30
Document importing `main`

clubby789:
Tracking issue: https://github.com/rust-lang/rust/issues/28937

I'd like to push for stabilizing this, so writing a documentation PR to go along with it.

@rustbot label +S-waiting-on-stabilization

### DIFF
--- a/src/crates-and-source-files.md
+++ b/src/crates-and-source-files.md
@@ -119,6 +119,17 @@ fn main() -> impl std::process::Termination {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


clubby789:
Changed to use `an import` rather than `re-export` as it matches the current feature name and not requiring a `pub use`.

ehuss (approved):
Thanks! Merging now that https://github.com/rust-lang/rust/pull/122060 is out.

PR #1465, opened 2024-02-09
runtime.md: The `#[unix_sigpipe = "sig_dfl"]` attribute

Enselic:
This documents the `#[unix_sigpipe = "sig_dfl"]` attribute that I propose to stabilize in https://github.com/rust-lang/rust/pull/120832.

[Rendered](https://github.com/Enselic/reference/blob/unix_sigpipe/src/runtime.md#the-unix_sigpipe--sig_dfl-attribute).

The tracking issue is https://github.com/rust-lang/rust/issues/97889.

### DIFF
--- a/src/runtime.md
+++ b/src/runtime.md
@@ -55,6 +55,60 @@ runtime with the [set_hook] function.
 The *`global_allocator` attribute* is used on a [static item] implementing the
 [`GlobalAlloc`] trait to set the global allocator.
 
+## The `#[unix_sigpipe = "sig_dfl"]` attribute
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! Can you also add a link to the attribute from the [built-in attribute list](https://github.com/rust-lang/reference/blob/master/src/attributes.md#built-in-attributes-index)?

Enselic:
I closed the stabilization PR so let's close this too for now. I will re-open when we are closer to stabilization.

PR #1471, opened 2024-02-17
Update destructors.md

Tbkhi:
Based on QA examples provided here: https://users.rust-lang.org/t/why-the-parameter-of-a-function-is-droped-afer-the-temporary-variable-in-return-of-the-function/106430/4

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -105,20 +105,21 @@ dropped after any bindings introduced in that parameter's pattern.
 
 ```rust
 // entire function scope begins
-fn fun(a: A)
+fn fun(a:i8) -> i8
 { // function body block begins
-    let b: A;
+    let b:i8 = 0;
+    return b;
 } // function body block ends. `b` is dropped here.
 // entire function scope ends. `a` is dropped here.
 
 // entire function begins.
-fn fun(a:A) // `a` associated to entire function.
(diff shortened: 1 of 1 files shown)


Tbkhi:
CC: @mbrubeck

Tbkhi:
CC; @ehuss

Tbkhi:
Oops this will need to be fixed up.

src/destructors.md
```diff
@@ -104,6 +104,26 @@ dropped last when evaluating the function. Each actual function parameter is
 dropped after any bindings introduced in that parameter's pattern.
 
 ```rust
+// entire function scope begins
```
jdahlstrom, line 107: First letter should be capitalized, for correctness and consistency with other Rust documentation (including this file). Ditto for all the other comments.

traviscross:
We're discussing this in the rustdocs call.  We're a bit unclear on the motivation for this change.  If you could describe here in more detail what should be changed here, why, and why the change proposed is the right way to address it, that would be helpful.

Also, we note that the PR should be rebased and squashed.

ehuss:
I'm going to close. I think we will eventually put some more work into the examples on drop scopes, but we're unsure exactly what it might look like. Thanks for the contribution!

PR #1472, opened 2024-02-17
Add note about static libraries not linking their dependencies

madsmtm:
This confused me for a while, but this is just fundamentally how static libraries work.

Wading through some history on this, I think when specifying `--crate-type=staticlib`, rustc used to output the `--print=native-static-libs` note by default, but it doesn't anymore, so let's document it instead!

### DIFF
--- a/src/linkage.md
+++ b/src/linkage.md
@@ -50,7 +50,7 @@ be ignored in favor of only building the artifacts specified by command line.
   Note that any dynamic dependencies that the static library may have (such as
   dependencies on system libraries, or dependencies on Rust libraries that are
   compiled as dynamic libraries) will have to be specified manually when
-  linking. The `--print=native-static-libs` flag may help with this.
+  linking that static library from somewhere. The `--print=native-static-libs` flag may help with this.
 
 * `--crate-type=cdylib`, `#![crate_type = "cdylib"]` - A dynamic system
   library will be produced.  This is used when compiling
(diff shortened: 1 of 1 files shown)


traviscross:
We're reviewing this in the the rustdocs call, and we're not sure about it.

@petrochenkov: What do you think about this one?

src/linkage.md
```diff
@@ -44,8 +44,13 @@ be ignored in favor of only building the artifacts specified by command line.
...
   on Windows (MSVC). This format is recommended for use in situations such as
-  linking Rust code into an existing non-Rust application
-  because it will not have dynamic dependencies on other Rust code.
+  linking Rust code into an existing non-Rust application because it will not
+  have dynamic dependencies on other Rust code.
+
+  Note that any dynamic dependencies that the static library may have (such as
+  dependencies on system libraries, or dependencies on Rust libraries that are
+  compiled as dynamic libraries) will have to be specified manually when
+  linking. The `--print=native-static-libs` flag may help with this.
```
petrochenkov, line 53:
```suggestion
  linking that static library from somewhere. The `--print=native-static-libs` flag may help with this.
```

petrochenkov (approved):
The new paragraph is correct.

traviscross (approved):
Thanks @petrochenkov for having a look.  If it looks good to you, then it looks good to me, and we'll merge this.

PR #1474, opened 2024-02-19
Add `rust-toolchain.toml` and revise README

traviscross:
We spent a lot of text in the README describing how to use the nightly Rust compiler with this project.  Let's instead add a `rust-toolchain.toml` file so that the nightly compiler is used automatically, and let's redraft the README in that light.

With this revision, we'll add a section on installing dependencies in which we'll describe how to ensure that the nightly compiler is installed and up to date and how to install `mdbook`.

When describing how to clone the project, we'll suggest that people clone the URL ending in `.git`.  This is what GitHub recommends, and it removes a level of redirection which sometimes matters.

For people who want to download a ZIP file rather than using `git`, we'll provide a direct link rather than simply referencing the GitHub page.

When describing how to build the book, we'll suggesting passing `--open` to `mdbook`, as this is usually what people will want, and it will save a step.

Along with these changes, we include some general copyediting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,58 +10,54 @@ what we have for now.
 
 ## Dependencies
 
-- rustc (the Rust compiler).
-- [mdbook](https://rust-lang.github.io/mdBook/) (use `cargo install
-  mdbook` to install it).
-- rust nightly (you would be required to set your Rust version to the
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

PR #1475, opened 2024-02-22
Update enumerations.md remove Visibility on Enum items

luposmi:
As far as I know, enums can only have a visibility as a whole, but not for single items.


ehuss:
Thanks for the PR! The current grammar is correct, in that it describes the AST of the language. This is the syntax of the language as understood when processing macros. The [Variant visibility](https://github.com/rust-lang/reference/blob/5f28ec3b0377c884cc92ee3ad593d80701646d28/src/items/enumerations.md#variant-visibility) section below explains why this is.

luposmi:
ahhh, thank you for your answer!

PR #1476, opened 2024-02-29
Update behavior-considered-undefined.md

zylthinking:


src/behavior-considered-undefined.md
```diff
@@ -119,7 +119,7 @@ omitted in surface Rust syntax due to automatic dereferencing; we are
 considering the fully expanded place expression here.)
 
 For instance, if `ptr` has type `*const S` where `S` has an alignment of 8, then
-`ptr` must be 8-aligned or else `(*ptr).f` is "based on an misaligned pointer".
+`*ptr` must be 8-aligned or else `(*ptr).f` is "based on an misaligned pointer".
```
traviscross, line 122:
```suggestion
the address contained in `ptr` must be cleanly divisible by 8 or else `(*ptr).f` is "based on an misaligned pointer".
```

On the rustdocs call today we spent a lot of time trying to determine whether or not this was correct or an improvement.  Maybe there's a better way to say this more explicitly that relies less on notation?

The ambiguity all comes from whether we're talking about where the pointer is stored or whether we're talking about the address contained in the pointer.

PR #1478, opened 2024-03-03
underscore-expr: add more examples

japm48:
Add examples of `_`-expressions.
In particular, assignment to `_` is something I constantly use while prototyping.


traviscross (approved):
Thanks @japm48 for this PR.  We discussed this on the rustdocs call and this looks good to us.

PR #1480, opened 2024-03-05
Update macros-by-example.md

JoergBrueggmann:
correct syntax '_MacroRulesDef_' 
NOTE: I tried to use round and square brackets. It seems that only curly brackets are allowed.

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -5,9 +5,7 @@
 > &nbsp;&nbsp; `macro_rules` `!` [IDENTIFIER] _MacroRulesDef_
 >
 > _MacroRulesDef_ :\
-> &nbsp;&nbsp; &nbsp;&nbsp; `(` _MacroRules_ `)` `;`\
-> &nbsp;&nbsp; | `[` _MacroRules_ `]` `;`\
-> &nbsp;&nbsp; | `{` _MacroRules_ `}`
+> &nbsp;&nbsp; `{` _MacroRules_ `}`
 >
 > _MacroRules_ :\
 > &nbsp;&nbsp; _MacroRule_ ( `;` _MacroRule_ )<sup>\*</sup> `;`<sup>?</sup>


JoergBrueggmann:
Also NOTE: Later on, the MacroMatcher, can have round, square and curly brackets.

ehuss:
Thanks for the PR! However, I don't quite understand the change. Parentheses and square brackets are valid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4cb725835420bff6d2cd198e8c847581

Perhaps you are missing the semicolon?

JoergBrueggmann:
> Thanks for the PR! However, I don't quite understand the change. Parentheses and square brackets are valid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4cb725835420bff6d2cd198e8c847581
> 
> Perhaps you are missing the semicolon?

Yes, later on you can use all three, but directly after the IDENTFIER the marco rules (symbol "_MacroRules_") can not be put e.g. into square brackets. The marco rules can only be put into curly braces.

Look at the compiler error and it's suggestion:

```
error: macros that expand to items must be delimited with braces or followed by a semicolon
  --> src/lib_example/sub_example.rs:32:25
   |
32 |   macro_rules! test_macro [
   |  _________________________^
33 | |     { $( $x:literal ).* } => {
34 | |         {
35 | |             $(
...  |
39 | |     };
40 | | ]
   | |_^
   |
help: change the delimiters to curly braces
   |
32 ~ macro_rules! test_macro {
33 |     { $( $x:literal ).* } => {
 ...
39 |     };
40 + }
```

ehuss:
In the given play link, it uses square or round brackets directly after the IDENTIFIER. The given error message explains the options:

> must be delimited with braces or followed by a semicolon

It looks like from your error message, it is missing a semicolon.

JoergBrueggmann:
> must be delimited with braces or followed by a semicolon

OK, you are right. According to the syntax it has to end with semicolon when round or square brackets are used. When you use curly brackets then it must NOT use a semicolon at the end. I overlooked that may be such syntax looks wierd to me. Why the exception for round or square brackets? May be there is a reason that I just don't understand.


However, the syntax is documented correctly.


I withdraw my PR. Sorry, for the inconvinence.


Best regards

Jörg

PR #1481, opened 2024-03-07
add grammar for `associated_type_bounds` in reference

compiler-errors:
This also edits the grammar to capture the fact that associated type bounds (both `:` and `=`) may have GAT parameters, i.e. `T: Trait<Foo<'a> = i32>`.

Not sure if I should also add a section on associated type bounds, and if so, where. Maybe `trait-bounds.md`?

cc rust-lang/rust#122055.

### DIFF
--- a/src/paths.md
+++ b/src/paths.md
@@ -53,7 +53,7 @@ mod m {
 > &nbsp;&nbsp; | `<` ( _GenericArg_ `,` )<sup>\*</sup> _GenericArg_ `,`<sup>?</sup> `>`
 >
 > _GenericArg_ :\
-> &nbsp;&nbsp; [_Lifetime_] | [_Type_] | _GenericArgsConst_ | _GenericArgsBinding_
+> &nbsp;&nbsp; [_Lifetime_] | [_Type_] | _GenericArgsConst_ | _GenericArgsBinding_ | _GenericArgsBounds_
 >
 > _GenericArgsConst_ :\
 > &nbsp;&nbsp; &nbsp;&nbsp; [_BlockExpression_]\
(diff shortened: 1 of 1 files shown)


src/paths.md
```diff
@@ -62,7 +62,10 @@ mod m {
 > &nbsp;&nbsp; | [_SimplePathSegment_]
 >
 > _GenericArgsBinding_ :\
-> &nbsp;&nbsp; [IDENTIFIER] `=` [_Type_]
+> &nbsp;&nbsp; [IDENTIFIER] _GenericArgs_<sup>?</sup> `=` [_TypeParamBounds_]
```
ehuss, line 65: I'm a little confused by this change to GenericArgsBinding. Why is it changing from `=Type` to `=TypeParamBounds`?

From what I can tell, this is in [`parse_assoc_equality_term`](https://github.com/rust-lang/rust/blob/42825768b103c28b10ce0407749acb21d32abeec/compiler/rustc_parse/src/parser/path.rs#L668-L676), which looks like it is expecting a type.

Can you walk me through what I'm missing? For example, `x::<foo<X>=impl Trait + ?Foo>`. The `impl Trait` part is the where I'm getting lost. If the `=` part is supposed to be a `TypeParamBounds`, how does that parse? If it is `TypeParamBounds`, then it looks like that follows to `TraitBound` which is essentially a `TypePath`, and `TypePath` does not include `Type` (it is mostly just paths with identifiers).

compiler-errors: This is a copy/paste error, lol

compiler-errors: This should definitely just be `= TYPE`

src/paths.md
```diff
@@ -62,7 +62,10 @@ mod m {
 > &nbsp;&nbsp; | [_SimplePathSegment_]
 >
 > _GenericArgsBinding_ :\
-> &nbsp;&nbsp; [IDENTIFIER] `=` [_Type_]
+> &nbsp;&nbsp; [IDENTIFIER] _GenericArgs_<sup>?</sup> `=` [_Type_]
```
compiler-errors, line 65: For the record, only thing this adds is GenericArgs<sup>?</sup>, so you can write `T: Trait<Assoc<i32> = Ty>`

jackh726:
Associated type bounds should be trait bounds, since they can contain a `for`.

compiler-errors:
https://doc.rust-lang.org/reference/trait-bounds.html -- `TypeParamBound` is correct here.

ehuss (approved):
Thanks! Merging now that https://github.com/rust-lang/rust/pull/122055 is out.

PR #1482, opened 2024-03-11
[type-layout] Document minimum size and alignment

joshlf:
This is a draft implementation of https://github.com/rust-lang/rust/pull/117474#issuecomment-1989349525. I'm happy to make any suggested edits.

cc @RalfJung

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -39,9 +39,9 @@ alignment." Every value of a type must have a size at least as large as the type
 minimum size and an alignment at least as large as the type's minimum alignment.
 Some notable cases of minimum size and alignment are:
 - For [`Sized`] types, the type's minimum size and minimum alignment
-  are always equal to the type's size and alignment
+  are always equal to the type's size and alignment, respectively
 - For a [slice type](#slice-layout), `[T]`, the minimum size is 0 bytes (corresponding
-  a slice with 0 elements), and the minimum alignment is the alignment of `T`
+  to a slice with 0 elements), and the minimum alignment is the alignment of `T`
 - For a [trait object](#trait-object-layout), the minimum size is 0 and the minimum
   alignment is 1
 - For a struct type with a dynamically-sized field, the minimum size is taken to be
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -32,6 +32,28 @@ as [dynamically sized types]. Since all values of a `Sized` type share the same
 size and alignment, we refer to those shared values as the size of the type and
 the alignment of the type respectively.
 
+### Minimum Size and Alignment
+
+Every type, including [dynamically sized types], has a "minimum size" and "minimum
+alignment." Every value of a type must have a size at least as large as the type's
+minimum size and an alignment at least as large as the type's minimum alignment.
+Some notable cases of minimum size and alignment are:
+- For [`Sized`] types, the type's minimum size and minimum alignment
+  are always equal to the type's size and alignment
+- For a [slice type](#slice-layout), `[T]`, the minimum size is 0 bytes (corresponding
+  a slice with 0 elements), and the minimum alignment is the alignment of `T`
+- For a [trait object](#trait-object-layout), the minimum size is 0 and the minimum
+  alignment is 1
+- For a struct type with a dynamically-sized field, the minimum size is taken to be
+  the size of the struct when the dynamically-sized field has *its* minimum size.
+  The struct's minimum alignment is taken to be the alignment of the struct when the
+  dynamically-sized field has *its* minimum alignment. When relying on this property
+  of dynamically-sized structs, be careful not to assume too much! For example,
+  this property provides *no* guarantees about [`repr(Rust)`](#the-rust-representation)
+  types, as such a type can have arbitrarily large size and alignment regardless of
+  the sizes and alignments of its fields
+- Every type's minimum size is less than or equal to `isize::MAX`
+
 ## Primitive Data Layout
 
 The size of most primitives is given in this table.
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -45,14 +45,15 @@ Some notable cases of minimum size and alignment are:
 - For a [trait object](#trait-object-layout), the minimum size is 0 and the minimum
   alignment is 1
 - For a struct type with a dynamically-sized field, the minimum size is taken to be
-  the size of the struct when the dynamically-sized field has *its* minimum size.
-  The struct's minimum alignment is taken to be the alignment of the struct when the
-  dynamically-sized field has *its* minimum alignment. When relying on this property
-  of dynamically-sized structs, be careful not to assume too much! For example,
-  this property provides *no* guarantees about [`repr(Rust)`](#the-rust-representation)
-  types, as such a type can have arbitrarily large size and alignment regardless of
-  the sizes and alignments of its fields
-- Every type's minimum size is less than or equal to `isize::MAX`
+  the size of the struct when the dynamically-sized field has *its* minimum size and
+  its minimum alignment. The struct's minimum alignment is taken to be the alignment
+  of the struct when the dynamically-sized field has *its* minimum alignment. When
+  relying on this property of dynamically-sized structs, be careful not to assume too
+  much! For example, this property provides *no* guarantees about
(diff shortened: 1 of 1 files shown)


src/type-layout.md
```diff
@@ -32,6 +32,28 @@ as [dynamically sized types]. Since all values of a `Sized` type share the same
...
+minimum size and an alignment at least as large as the type's minimum alignment.
+Some notable cases of minimum size and alignment are:
+- For [`Sized`] types, the type's minimum size and minimum alignment
+  are always equal to the type's size and alignment, respectively
+- For a [slice type](#slice-layout), `[T]`, the minimum size is 0 bytes (corresponding
+  to a slice with 0 elements), and the minimum alignment is the alignment of `T`
+- For a [trait object](#trait-object-layout), the minimum size is 0 and the minimum
+  alignment is 1
+- For a struct type with a dynamically-sized field, the minimum size is taken to be
+  the size of the struct when the dynamically-sized field has *its* minimum size.
```
RalfJung, line 48: Alignment affects size, so discussing the two separately here doesn't quite work.

src/type-layout.md
```diff
@@ -32,6 +32,28 @@ as [dynamically sized types]. Since all values of a `Sized` type share the same
...
+  The struct's minimum alignment is taken to be the alignment of the struct when the
+  dynamically-sized field has *its* minimum alignment. When relying on this property
+  of dynamically-sized structs, be careful not to assume too much! For example,
+  this property provides *no* guarantees about [`repr(Rust)`](#the-rust-representation)
+  types, as such a type can have arbitrarily large size and alignment regardless of
+  the sizes and alignments of its fields
+- Every type's minimum size is less than or equal to `isize::MAX`
```
RalfJung, line 55: This bullet doesn't belong into the list of "notable cases of minimum size and alignment are", it belongs before or after the list. Also it should say what happens when one declares a type whose minimum size exceeds this limit.

RalfJung:
There's a potential issue here -- not all types mentioned in the source code actually make it to codegen, so sometimes types that are "too big" do not lead to errors.

This example fails to build in a debug build but succeeds as a release build:
```rust
#![crate_type = "lib"]

pub fn f() {
    g::<u16>();
}
pub fn g<T>() -> std::mem::MaybeUninit<[T; 1 << 47]> {
    std::mem::MaybeUninit::uninit()
}
```

src/type-layout.md
joshlf: Are you suggesting just adjusting it to say this?

> For a struct type with a dynamically-sized field, the minimum size is taken to be the size of the struct when the dynamically-sized field has *its* minimum size and alignment.

(Done. I can revert if I misunderstood your suggestion.)

src/type-layout.md
joshlf: > This bullet doesn't belong into the list of "notable cases of minimum size and alignment are", it belongs before or after the list.

Done.

> Also it should say what happens when one declares a type whose minimum size exceeds this limit.

Do we need to specify that? Presumably the implication is just that such a program would not conform to the spec, and so we couldn't permit it to compile, but a) that's a logical implication of the text here and, b) I don't believe we need to specify the *manner in which* it fails to compile.

I'm happy to also specify that such programs won't compile - I just want to make sure I'm understanding correctly that this is just a nice-to-have rather than something that affects the semantics of this text.

joshlf:
> There's a potential issue here -- not all types mentioned in the source code actually make it to codegen, so sometimes types that are "too big" do not lead to errors.
> 
> This example fails to build in a debug build but succeeds as a release build:
> 
> ```rust
> #![crate_type = "lib"]
> 
> pub fn f() {
>     g::<u16>();
> }
> pub fn g<T>() -> std::mem::MaybeUninit<[T; 1 << 47]> {
>     std::mem::MaybeUninit::uninit()
> }
> ```

Good call. This shouldn't be a problem in our [motivating use case](https://github.com/rust-lang/rust/issues/69835#issuecomment-1782076799), but obviously we need to be careful with what we promise.

Intuitively, the guarantee is something like "a type which you can do anything with" or "a type which you can create a pointer to". But I'm not sure how to formalize that idea.

For the motivating use case, we need to guarantee that this holds of `*const T` if I can actually execute code that synthesizes a `*const T`.

RalfJung:
The new wording sounds good to me.

However, the issue with "which types actually end up being size-checked" remains. I don't have a good idea for how to express that.

@rust-lang/opsem @rust-lang/lang: any thoughts? There's tow questions; the easy one is what you think about the text of the PR, but the harder one is [this concern](https://github.com/rust-lang/reference/pull/1482#issuecomment-2036255754). Specifically, the question is for which types we guarantee that their size is less than `isize::MAX`. You may think the answer is "all of them", but that's [not true](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=b5537881c367b8e68d0b3f2716f5c76d) -- if a type gets optimized away before codegen, it may never be size-checked.

I don't have a good idea for how to even say which types are *guaranteed* to be size-checked in the current implementation. Similar to what we did for const-eval failures, we could try to have a notion of "required types" such that if a function runs, these types are definitely size-checked -- but given the huge number of types in a typical program, that sounds like it could be quite expensive.

src/type-layout.md
joshlf: Sounds reasonable. Presumably resolving https://github.com/rust-lang/reference/pull/1482#issuecomment-2036255754 is required in order for us to agree on a wording for this?

joshlf:
Is there some notion of "observability" we could use here? As in, if a type's existence is "observable" in some way? Intuitively, I'd imagine this just means that if the execution of a program in any way depends on the existence of the type (calling `size_of::<T>()`, constructing a `*const T`, etc), although I'm not sure if that notion is formalized anywhere.

src/type-layout.md
RalfJung: The questions of "which types are checked" and "what happens when a type fails the check" seem orthogonal to me, so I don't see an overlap with the other discussion.

RalfJung:
The only things that are "observable" about a program are its IO behavior (and volatile accesses). So I don't think that will help us here.

src/type-layout.md
```diff
@@ -32,6 +32,29 @@ as [dynamically sized types]. Since all values of a `Sized` type share the same
...
+minimum size and an alignment at least as large as the type's minimum alignment.
+Some notable cases of minimum size and alignment are:
+- For [`Sized`] types, the type's minimum size and minimum alignment
+  are always equal to the type's size and alignment, respectively
+- For a [slice type](#slice-layout), `[T]`, the minimum size is 0 bytes (corresponding
+  to a slice with 0 elements), and the minimum alignment is the alignment of `T`
+- For a [trait object](#trait-object-layout), the minimum size is 0 and the minimum
+  alignment is 1
+- For a struct type with a dynamically-sized field, the minimum size is taken to be
+  the size of the struct when the dynamically-sized field has *its* minimum size and
```
nikomatsakis, line 48: I am not clear on what "it" refers to in this sentence

nikomatsakis: I *think* the answer is value-dependent, e.g., for `Foo<dyn Blah>` the alignment depends on the value stored in the vtable -- put another way, it's determined by the "erased information"

RalfJung: You mean "its"? It is straight-forward recursion. If a struct has field types `T, U, V` where `V` is dyn-sized, then the min size of the struct is computed using the min size of `V`.

RalfJung:
FWIW after further discussion, my conclusion is that all we really need for @joshlf's usecase right now is https://github.com/rust-lang/rust/pull/126152. That avoids having to document when exactly compilation fails due to types being bigger than `isize::MAX`.

IOW, the entire point of this PR is to be able to say "Every type's minimum size is less than or equal to `isize::MAX`", but that statement has [problems](https://github.com/rust-lang/reference/pull/1482#issuecomment-2036255754).

src/type-layout.md
nikomatsakis: That formulation seems much clearer :)

RalfJung: The minimum alignment is not value-dependent, no. Minimum size and minimum alignment are static properties that *every* type possesses.

nikomatsakis: I see.

PR #1483, opened 2024-03-19
Literal expressions: fix mistake in the definition of unicode escapes

mattheww:
Write `u32::from_str_radix` rather than `u8::from_str_radix` for unicode escapes.


ehuss (approved):
Thanks!

PR #1485, opened 2024-03-23
Add documentation for meta variable expressions

c410-f3r:
Requested in https://github.com/rust-lang/rust/pull/122808

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -196,6 +196,276 @@ compiler knows how to expand them properly:
     not have the same number. This requirement applies to every layer of nested
     repetitions.
 
+## Metavariable expressions
+
+Metavariable expressions live within a special braced syntax `${...}` and are function-like operators used to access "meta" information about how macros get matched and expanded, such as how many items get matched in a repetition group. This information is typically difficult or even
+impossible to otherwise obtain.
+
+### count($ident, depth=0)
+
+Expands to an unsuffixed integer literal representing the number of times a ***repetition*** repeats in total. That is, this will be the number of times that a repetition would be expanded.
+
+The output of `count` depends on where it is placed as well the provided index. If no index is provided, then it will always start at the innermost level.
+
+```rust
+macro_rules! count_idents {
+    ( $( $i:ident ),* ) => {
+        ${count($i)}
+    };
+}
+
+fn main() {
+    assert_eq!(count_idents!(a, b, c), 3);
+}
+```
+
+If repetitions are nested, then the optional depth parameter can be used to
+count the repetitions of a parent repetition group.
+
+Let `M` be the number of repetition groups in the matcher, `T` the depth placement in the transcriber and `C` the written depth in `count`, then it is possible to infer that `M > T + C`.
+
+We are counting how often the repetition group `C` up from a metavariable in the matcher occurs inside the current iteration of the `T` outer repetitions in the transcriber, effectively making the metavariable a reference or starting point.
+
+```rust
+macro_rules! count_value {
+    ( $( $name:ident: $( $value:literal ),* );+ ) => {
+        // Count the total number of times that the (innermost) group
+        // containing `$value` gets matched.
+        ${count($value)}
+        // This is the same as `${count($value, 0)}`
+    };
+}
+macro_rules! count_name1 {
+    ( $( $name:ident: $( $value:literal ),* );+ ) => {
+        // This is one way to get the number of times that the group
+        // containing `$name` gets matched. Alternatively...
+        ${count($name)}
+    };
+}
+macro_rules! count_name2 {
+    ( $( $name:ident: $( $value:literal ),* );+ ) => {
+        // ...`$value` can be used again with a depth specifier of 1,
+        // indicating that repetitions of the 1st parent group of
+        // `$value` should be counted, rather than `$value`'s innermost group.
+        //
+        // `1` is the maximum value here since `$value`'s group has a single
+        // parent.
+        ${count($value, 1)}
+    };
+}
+macro_rules! count_value_nested {
+    ( $( $name:ident: $( $value:literal ),* );+ ) => {
+        [ $(
+            // using `count` within a repetition group will return the number
+            // of times `$value` is matched _within that group_.
+            ${count($value)},
+        )+ ]
+    };
+}
+fn main() {
+    // all instances of `$value` counted: count(1, 2, 3, 4, 5) = 5
+    assert_eq!(count_value!(a: 1, 2, 3; b: 4, 5), 5);
+    // count(1, 2, 3, ... 11) = 11
+    assert_eq!(
+        count_value!(a: 1, 2, 3; b: 4, 5; c: 6, 7; d: 8, 9, 10, 11),
+        11
+    );
+    // `$value` is never matched; count() = 0
+    assert_eq!(count_value!(a:), 0);
+    // count(a, b) = 2 matches
+    assert_eq!(count_name1!(a: 1, 2, 3; b: 4, 5), 2);
+    // count(a, b, c, d) = 4 matches
+    assert_eq!(
+        count_name1!(a: 1, 2, 3; b: 4, 5; c: 6, 7; d: 8, 9, 10, 11),
+        4
+    );
+    // count(a) = 1 match
+    assert_eq!(count_name1!(a:), 1);
+    // These have the same results as the above
+    assert_eq!(count_name2!(a: 1, 2, 3; b: 4, 5), 2);
+    assert_eq!(
+        count_name2!(a: 1, 2, 3; b: 4, 5; c: 6, 7; d: 8, 9, 10, 11),
+        4
+    );
+    assert_eq!(count_name2!(a:), 1);
+    // first match: count(1, 2, 3) = 3. second match: count(4, 5) = 2.
+    assert_eq!(count_value_nested!(a: 1, 2, 3; b: 4, 5), [3, 2]);
+    // first match: count(1, 2, 3) = 3. second match: count(4, 5) = 2.
+    // third match: count(6, 7) = 4. fourth match: count(8, 9, 10, 11) = 4.
+    assert_eq!(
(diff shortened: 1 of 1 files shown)


src/macros-by-example.md
```diff
@@ -195,6 +195,220 @@ compiler knows how to expand them properly:
     not have the same number. This requirement applies to every layer of nested
     repetitions.
 
+## Meta variable expressions
```
jdahlstrom, line 198: "Meta variable" should be "metavariable" for consistency with everything else. Ditto for all other appearances of the phrase.

src/macros-by-example.md
```diff
@@ -195,6 +195,220 @@ compiler knows how to expand them properly:
...
+
+Give access to additional metadata about meta variables that otherwise would be difficult or even impossible to get through elements such as `count` or `index`.
```
jdahlstrom, line 200: > elements such as `count` or `index`.

The wording implies that `count` and `index` are "elements" through which it would be difficult or impossible to get metadata, but presumably that's not the intention.

src/macros-by-example.md
```diff
@@ -195,6 +195,220 @@ compiler knows how to expand them properly:
...
+    assert_eq!(innermost1!(a: 1, 2, 3; b: 4, 5), [0, 0, 0, 1, 1]);
+    
+    // a b = 2 elements
+    //
+    // Increasing list from the outermost loop refering outermost indexes
+    assert_eq!(outermost!(a: 1, 2, 3; b: 4, 5), [0, 1]);
+}
+```
+
+### length(depth)
```
jdahlstrom, line 352: Bikeshedding, but should consider changing the name to "len". Length is abbreviated to "len" everywhere in `std`; there is only a single, unstable, API that uses "length".

tgross35: @jdahlstrom you may want to bring that up on the PR, currently under FCP proposal https://github.com/rust-lang/rust/pull/122808

jdahlstrom: @tgross35 Indeed, done!

ehuss:
@c410-f3r Can you update this from the changes in https://github.com/rust-lang/rust/pull/124987?

c410-f3r:
Updated

tgross35 (commented):
Added some suggestions for wording that I think is a bit easier to understand. I think this section could really get its own page since "Macros By Example" is already huge, and metavariable expressions will likely have more entries in the future.

Also, most of this md doc is wrapped to 80 chars, probably good to keep that consistent rather than switching wrapping styles

src/macros-by-example.md
```diff
@@ -196,6 +196,220 @@ compiler knows how to expand them properly:
...
+    // 1 2 3 4 5 = 5 elements
+    assert_eq!(no_repetition0!(a: 1, 2, 3; b: 4, 5), [5]);
+    
+    // a b = 2 elements
+    assert_eq!(no_repetition1!(a: 1, 2, 3; b: 4, 5), [2]);
+    
+    // 1 2 3 = 3 elements
+    // 4 5 = 2 elements
+    assert_eq!(outermost!(a: 1, 2, 3; b: 4, 5), [3, 2]);
+}
```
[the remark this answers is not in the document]
RalfJung: > using `count` within a repetition group will return the number>
> of times `$value` is matched _within that group_.

This is a key part of the spec. It should be in the reference text above, not just in the examples.

src/macros-by-example.md
```diff
@@ -196,6 +196,220 @@ compiler knows how to expand them properly:
...
+```
+
+`count` can not be placed inside the innermost repetition.
```
RalfJung, line 255: Would be good to have an example of that; I don't know what is meant here.

Is this even about the innermost repetition? In `( $( $name:ident: $( $value:literal ),* );+ )`, can I do `$(  ${count($name) }`? That is not the innermost repetition, as `value` is even-more-inner -- but I am trying to count `name` inside a repetition for `name` which will always return 1.

src/macros-by-example.md
```diff
@@ -196,6 +196,220 @@ compiler knows how to expand them properly:
...
 
+## Metavariable expressions
+
+Gives access to additional information about metavariables that otherwise would be difficult or even impossible to obtain manually.
+
+### count($ident, depth)
+
+Expands to an unsuffixed integer literal representing the number of times a ***metavariable*** repeats in total.
+
+The output of `count` depends on where it is placed as well the provided index. If no index is provided, then it will always start at the innermost level.
```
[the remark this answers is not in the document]
RalfJung: I think this shouldn't be removed, rather it should be expanded to fully describe the exact algorithm used to determine the result. I assume it's something like [this](https://github.com/rust-lang/rust/pull/122808#issuecomment-2138788702) but I don't know the right terms for all the concepts involved (and I only experimented with `depth=0`).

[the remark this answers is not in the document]
RalfJung: In general, it seems to be something like
- let's say `$ident` occurs under N repetition groups in the matcher
- and now we use `${count($ident, D)}` under M repetition groups in the transcriber
- then we must have M+D < N
- and we are counting how often the repetition group D up from `Ident` in the matcher, occurs inside the current instance of the M outer repetitions in the transcriber

traviscross:
@c410-f3r: We looked at this in the rustdocs call today.  Do you have any thoughts on the feedback given above?  If so, that could help us in reviewing this.

c410-f3r:
Feedback looks good but I personally prefer the wait for the decision about https://github.com/rust-lang/rust/pull/122808#issuecomment-2156454878.

Otherwise this could be my third reverted PR. See https://github.com/rust-lang/reference/pull/1179, https://github.com/rust-lang/reference/pull/1251, https://github.com/rust-lang/reference/pull/1192 and https://github.com/rust-lang/rust/pull/99435.

RalfJung:
> Feedback looks good but I personally prefer to wait for the decision about https://github.com/rust-lang/rust/pull/122808#issuecomment-2156454878.

The problem is that currently we are [guessing](https://github.com/rust-lang/rust/pull/122808#issuecomment-2181569717) what the proposed semantics even are. So someone who knows the implementation needs to write up-to-date information on that before a decision can be made.

c410-f3r:
I will try to address each concern or question this weekend in my free time.

For what it is worth, here goes another illustrative example in the meanwhile.

meta

```rust
#![feature(macro_metavar_expr)]

macro_rules! no_repetition {
    ( $( [ $( ( $($i:ident)* ) )* ] )* ) => {
        (
            // ****** 1 `ident` count *****
            //
            // Guided by 6 innermost `ident` from `$($i:ident)*`
            //
            // 6 (a b c d e f)
            [${count($i)}],

            // ****** 1 `ident` count *****
            //
            // Guided by 5 middle `(...)` from `$( ( $($i:ident)* ) )*`
            //
            // 5 (() () () () ())
            [${count($i, 1)}],

            // ****** 1 `ident` count *****
            //
            // Guided by 3 outermost `[...]` from `$( [ $( ( $($i:ident)* ) )* ] )*`
            //
            // 3 ([] [] [])
            [${count($i, 2)}],
        )
    }
}

macro_rules! one_repetition {
    ( $( [ $( ( $($i:ident)* ) )* ] )* ) => {
        (
            // ****** 3 `[...]` counts *****
            //
            // Guided by 6 innermost `ident` from `$($i:ident)*`
            //
            // 2 (a b)
            // 4 (c d e f)
            // 0
            [$( ${count($i)}, )*],

            // ****** 3 `[...]` counts *****
            //
            // Guided by 5 middle `(...)` from `$( ( $($i:ident)* ) )*`
            //
            // 2 (() ())
            // 3 (() () ())
            // 0
            [$( ${count($i, 1)}, )*],
        )
    }
}

macro_rules! two_repetitions {
    ( $( [ $( ( $($i:ident)* ) )* ] )* ) => {
        (
            // ****** 5 `(...)` counts *****
            //
            // Guided by 6 innermost `ident` from `$($i:ident)*`
            //
            // 2 (a b)
            // 0
            // 1 (c)
            // 0
            // 3 (d e f)
            [$( $( ${count($i)}, )* )*],
        )
    }
}

fn main() {
    let (innermost, middle, outermost) = no_repetition!([(a b) ()] [(c) () (d e f)] []);
    assert_eq!(innermost, [6]);
    assert_eq!(middle, [5]);
    assert_eq!(outermost, [3]);

    let (innermost, middle) = one_repetition!([(a b) ()] [(c) () (d e f)] []);
    dbg!(innermost, middle);

    let (innermost,) = two_repetitions!([(a b) ()] [(c) () (d e f)] []);
    assert_eq!(innermost, [2, 0, 1, 0, 3]);
}
```

src/macros-by-example.md
c410-f3r: > Is this even about the innermost repetition?

That is correct, `count` can not be placed inside the repetition depth of its referenced meta-variable. So `innermost` in regards to the macro declaration is inaccurate.

> can I do $(  ${count($name) }?

No because of the above explanation.

> but I am trying to count name inside a repetition for name which will always return 1.

That is the reason behind such constraint. It doesn't make sense to count something that will always yield `1`.

Nevertheless, it can be relaxed in the future if so desired.

src/macros-by-example.md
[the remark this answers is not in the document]
c410-f3r: > In general, it seems to be something like
> 
>     * let's say `$ident` occurs under N repetition groups in the matcher
> 
>     * and now we use `${count($ident, D)}` under M repetition groups in the transcriber
> 
>     * then we must have M+D < N
> 
>     * and we are counting how often the repetition group D up from `$ident` in the matcher occurs inside the current iteration of the M outer repetitions in the transcriber. IOW, for the case where D=0 (IMO the only one we should stabilize for now), we are counting how many times `$ident` would expand  if we surrounded `$ident` in enough repetition groups to make it legal to occur at this point.

That is a good explanation. Specially in the "`counting how often the repetition group D up from $ident in the matcher occurs`" highlighting that `$ident` is just a matcher reference when using indexes.

PR #1488, opened 2024-03-30
Fix clippy warning in procedural macro example

yotamofek:
I copy+pasted this example into my code and the `clippy::to_string_in_format_args` lint fired.


ehuss (approved):
Thanks!

PR #1491, opened 2024-04-14
Document how `non_exhaustive` interacts with tuple and unit-like structs.

### LINKED ISSUE #1097: Clarify non_exhaustive behavior with unit structs
The current page on [`#[non_exhaustive]`](https://github.com/rust-lang/reference/blob/b5c68b02984f74e99d1f1b332029e05f607e2660/src/attributes/type_system.md#the-non_exhaustive-attribute) doesn't really say much about unit structs. I think it could use some clarification. For example, the visibility of the unit struct const is lowered to `pub(crate)` like tuple struct constructors.

The [RFC](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html#unit-structs) has more detail.  It might be good to also comb the RFC and think about anything else missing from the non_exhaustive docs.

kpreid:
I was surprised by this behavior and it led me to believe [Clippy was issuing a false positive](https://github.com/rust-lang/rust-clippy/issues/12675) because I thought that `#[non_exhaustive]` was largely ineffective on unit structs. But in fact it does have notable effects, and these effects were not previously documented in the reference except in the matching section.

Closes #1097

### DIFF
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -72,10 +72,12 @@ Non-exhaustive types cannot be constructed outside of the defining crate:
 
 - Non-exhaustive variants ([`struct`][struct] or [`enum` variant][enum]) cannot be constructed
   with a [_StructExpression_] \(including with [functional update syntax]).
-- The visibility of the same-named constant of a [unit-like struct][struct]
-  is lowered to `min($vis, pub(crate))`.
-- The visibility of the same-named constructor function of a [tuple struct][struct]
-  is lowered to `min($vis, pub(crate))`.
+- The implicitly defined same-named constant of a [unit-like struct][struct],
+  or the same-named constructor function of a [tuple struct][struct],
+  has a [visibility] no greater than `pub(crate)`.
+  That is, if the struct’s visibility is `pub`, then the constant or constructor’s visibility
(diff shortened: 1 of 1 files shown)


src/attributes/type_system.md
```diff
@@ -64,30 +72,47 @@ Non-exhaustive types cannot be constructed outside of the defining crate:
 
 - Non-exhaustive variants ([`struct`][struct] or [`enum` variant][enum]) cannot be constructed
   with a [_StructExpression_] \(including with [functional update syntax]).
+- The visibility of the same-named constant of a [unit-like struct][struct]
+  is lowered to `min($vis, pub(crate))`.
+- The visibility of the same-named constructor function of a [tuple struct][struct]
+  is lowered to `min($vis, pub(crate))`.
```
ehuss, line 78: `min($vis, pub(crate))` doesn't really define what `min` or `$vis` means. I imagine most people can suss out what it means, but I would be more comfortable expressing it without inventing new syntax. Perhaps it could say something like:

> If the visibility of … is `pub`, its visibility is lowered to `pub(crate)`.

kpreid: I was copying the existing text from the matching section, but I agree it's not ideal. Edited both.

ehuss (approved):
Thanks!

PR #1492, opened 2024-04-16
Update clone reference to include closures

conradludgate:


conradludgate:
> Thanks! I pushed a small change to also mention if it doesn't capture anything.

Thanks, that's fair. I didn't include it originally as it was already mentioned under the "All copy types" from above, but I have no complaints in being extra explicit

PR #1493, opened 2024-04-16
Expand and clarify primitive alignment

kpreid:
These changes are intended to make the section more informative and readable, without making any new normative claims.

* Specify that the alignment might be _less_ than the size, rather than just that it might be different. This is mandatory and stated in the previous section, but I think it's useful to reiterate here.
* Mention `u128`/`i128` as another example of alignment less than size, so that this doesn't sound like a mainly 32-bit thing.
* Add `usize`/`isize` to the size table, so it can be spotted at a glance.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -44,17 +44,20 @@ The size of most primitives is given in this table.
 | `u32` / `i32`     | 4                  |
 | `u64` / `i64`     | 8                  |
 | `u128` / `i128`   | 16                 |
+| `usize` / `isize` | See below          |
 | `f32`             | 4                  |
 | `f64`             | 8                  |
 | `char`            | 4                  |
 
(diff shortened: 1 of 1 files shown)


ehuss:
> Specify that the alignment might be less than the size, rather than just that it might be different. This is mandatory and stated in the previous section, but I think it's useful to reiterate here.

Sorry, I'm not quite following this. Where is it stated that the alignment must be `<=` to their size?

kpreid:
> Where is it stated that the alignment must be `<=` to their size?

The previous section, “Size and Alignment”:

> … The size of a value is always a multiple of its alignment. …

This means that either the alignment is less than or equal to the size, or the size is zero (which does not apply to any of the types in this list).

ehuss (approved):
Thanks!

PR #1499, opened 2024-05-23
elaborate on slice wide pointer metadata

RalfJung:
As decided in https://github.com/rust-lang/unsafe-code-guidelines/issues/510

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -72,11 +72,14 @@ Please read the [Rustonomicon] before writing unsafe code.
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
   * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
-  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
+  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
+    for the metadata is determined by the type of the unsized tail:
     * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
       `Trait` that matches the actual dynamic trait the pointer or reference points to.
     * Slice metadata is invalid if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
+      Furthermore, for wide references and `Box<T>`, slice metadata is invalid
+      if it makes the total size of the pointed-to value bigger than `isize::MAX`.
   * Invalid values for a type with a custom definition of invalid values.
     In the standard library, this affects [`NonNull<T>`] and [`NonZero*`].
 
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -71,12 +71,13 @@ Please read the [Rustonomicon] before writing unsafe code.
   * A `!` (all values are invalid for this type).
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
-  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
+  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value
+    (in case of dynamically sized types, using the actual dynamic type of the
+    pointee as determined by the metadata).
   * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
     for the metadata is determined by the type of the unsized tail:
-    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the pointer or reference points to.
-    * Slice metadata is invalid if the length is not a valid `usize`
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for `Trait`.
+    * Slice (`[T]`) metadata is invalid if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
       Furthermore, for wide references and `Box<T>`, slice metadata is invalid
       if it makes the total size of the pointed-to value bigger than `isize::MAX`.


src/behavior-considered-undefined.md
```diff
@@ -72,11 +72,14 @@ Please read the [Rustonomicon] before writing unsafe code.
...
-  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
+  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
+    for the metadata is determined by the type of the unsized tail:
     * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the pointer or reference points to.
+      `Trait` that matches the actual dynamic type the pointer or reference points to.
     * Slice metadata is invalid if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
+      Furthermore, for wide references and `Box<T>`, slice metadata is invalid
+      if it makes the total size of the type bigger than `isize::MAX`.
```
joshlf, line 82:
```suggestion
      if it makes the total size of the value bigger than `isize::MAX`.
```

(Alternatively, "referent")

src/behavior-considered-undefined.md
```diff
@@ -72,11 +72,14 @@ Please read the [Rustonomicon] before writing unsafe code.
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
   * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
-  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
+  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
+    for the metadata is determined by the type of the unsized tail:
     * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the pointer or reference points to.
+      `Trait` that matches the actual dynamic type the pointer or reference points to.
```
tmandry, line 78: Preexisting but how can this be a requirement for wide _pointers_, whose data portion is allowed to dangle, I assume?

RalfJung: Hm... yeah the actually implemented requirement is that the `trait` matches the trait given in the pointer type, i.e. a `*const dyn Debug` that points to a vtable for `Display` is UB.

traviscross:
@rust-lang/opsem / @RalfJung: In looking carefully at this on the rustdocs call, we realized it might be helpful, both to us and when the lang teams takes this up, if someone might be able annotate the changes here (e.g. using the GH review features), with some explanation for each of these (e.g. "this was incorrect because...", "this was already true because we had said...", "we're adding this guarantee here because...").

src/behavior-considered-undefined.md
```diff
@@ -71,12 +71,17 @@ Please read the [Rustonomicon] before writing unsafe code.
   * A `!` (all values are invalid for this type).
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
-  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
-  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
+  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value
+    (using the actual dynamic type of the pointee as determined by the vtable in
+    the metadata in case of dynamically sized types).
```
RalfJung, line 76: This does not change anything, it clarifies what the pointed-to value is in case of e.g. `&dyn Trait`.

src/behavior-considered-undefined.md
```diff
@@ -71,12 +71,17 @@ Please read the [Rustonomicon] before writing unsafe code.
...
+  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
+    for the metadata is determined by the type of the unsized tail:
```
RalfJung, line 78: Another piece of clarification, previously we were not very clear on what the metadata requirements are for types like `(u32, [u32])`: we spoke about "slice metadata" and the fact that this covered all types whose unsized tail is a slice was left implicit. Now it is explicit.

src/behavior-considered-undefined.md
```diff
@@ -71,12 +71,16 @@ Please read the [Rustonomicon] before writing unsafe code.
   * A `!` (all values are invalid for this type).
   * An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained
     from [uninitialized memory][undef], or uninitialized memory in a `str`.
-  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value.
-  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer:
-    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
-      `Trait` that matches the actual dynamic trait the pointer or reference points to.
```
RalfJung, line 77: "that matches the actual dynamic trait" was clearly nonsense. Also this is now moved to the "points to an invalid" point since it's not about the metadata, it's about using the metadata to keep going recursively through the reference.

src/behavior-considered-undefined.md
```diff
@@ -71,12 +71,16 @@ Please read the [Rustonomicon] before writing unsafe code.
...
+  * A reference or `Box<T>` that is [dangling], misaligned, or points to an invalid value
+    (using the actual dynamic type of the pointee as determined by the vtable in
+    the metadata in case of dynamically sized types).
+  * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement
+    for the metadata is determined by the type of the unsized tail:
+    * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for `Trait`.
+    * Slice (`[T]`) metadata is invalid if the length is not a valid `usize`
       (i.e., it must not be read from uninitialized memory).
+      Furthermore, for wide references and `Box<T>`, slice metadata is invalid
+      if it makes the total size of the pointed-to value bigger than `isize::MAX`.
```
RalfJung, line 83: This is the only actual change, decided in https://github.com/rust-lang/unsafe-code-guidelines/issues/510.

RalfJung:
Note that the PR has two commits which also separate functional changes from clarifications and typo fixes.

nikomatsakis (approved):
+1

tmandry (approved):

traviscross:
@rfcbot reviewed

traviscross:
@rustbot labels -I-lang-nominated +final-comment-period

This is now in FCP so we can unnominate.

ehuss (approved):
Thanks, looks like FCP has finished.

PR #1500, opened 2024-05-23
On non-pointer types `*x` is not always equivalent to its de-sugar form

NichtsHsu:
`*x` is allowed to temporarily extend the lifetime, but its de-sugar form is not.

For example:

```rust
use std::ops::Deref;

struct Num(i32);
struct NumRef<'a>(&'a i32);

impl Num {
    fn get_ref(&self) -> NumRef<'_> {
        NumRef(&self.0)
    }
}

impl Deref for NumRef<'_> {
    type Target = i32;
    fn deref(&self) -> &Self::Target {
        self.0
    }
}

fn main() {
    let num = Num(1);
    
    // Temporary lifetime extension
    let r = &*num.get_ref();
    println!("{r}");
    
    // Failed to compile
    let r = &*Deref::deref(&num.get_ref());
    println!("{r}");
}
```

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -134,7 +134,7 @@ When applied to a [pointer](../types/pointer.md) it denotes the pointed-to locat
 If the expression is of type `&mut T` or `*mut T`, and is either a local variable, a (nested) field of a local variable or is a mutable [place expression], then the resulting memory location can be assigned to.
 Dereferencing a raw pointer requires `unsafe`.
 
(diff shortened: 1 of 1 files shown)


traviscross:
We discussed this in the rustdocs call.  Thanks @NichtsHsu for proposing this.  It's an interesting point, because whether or not the text is correct depends on what we consider `x` to be in the text.  We think of it as referring to some existing binding, in which case, it's already correct.  So it's on that basis that we're going to close this, but it is an interesting point that we'll keep in mind for future work.

PR #1501, opened 2024-05-24
Document `gen` keyword as reserved in Rust 2024

traviscross:
Following RFC 3513 and subsequent work, the `gen` keyword is reserved in the Rust 2024 edition.  Let's document that.

See tracking issue:

- https://github.com/rust-lang/rust/issues/123904


ehuss:
Thanks! Will merge with the 2024 stabilization.

ehuss (approved):

PR #1504, opened 2024-05-29
Update procedural-macros.md adding context for use of `extern crate p…

mav3ri3k:
Use of `extern crate` is generally stopped and not recommended [source](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2018/path-changes.md#an-exception). Its use here was confusing for me. Based on the config provided here for the `Cargo.toml` file, use of `extern crate proc_macro;` in following examples is not required. 

The snippets were last updated 6 years ago. I would personally remove use of `extern crate` crate from following examples and add one with it to show how it can be done when using separate build system which would likely be niche case.

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -21,6 +21,11 @@ Procedural macros must be defined in a crate with the [crate type] of
 > [lib]
 > proc-macro = true
 > ```
+> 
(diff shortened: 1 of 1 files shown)


traviscross:
We discussed this in the rustdocs call.  We appreciate the PR; thanks @mav3ri3k for proposing it.  However, we're going to close it as it goes a bit too far in discussing details of Cargo that aren't really appropriate for the Reference.

PR #1505, opened 2024-05-30
Correct the remark regarding the C Representation dual purpose

jjerphan:
Thank you for documenting the different representations available in Rust.

I think the following sentence has to be reworded.


traviscross:
Thanks for making a suggestion here.  However, my read is that the sentence is correct as written.  It's saying that it's possible to create a type that is not useful for interfacing with C (but which might be useful e.g. so that one can rely on the layout), and this is true.

While it is also true that one could create a type that is "not only" useful for interfacing with C, that's not what this sentence is trying to convey.  This sentence is trying to convey the existence of *disjointness*.

PR #1506, opened 2024-05-31
this needs a space

correabuscar:
else it [looks](https://doc.rust-lang.org/reference/items/generics.html) kinda off


ehuss (approved):
Thanks!

PR #1507, opened 2024-06-02
Add Apple `target_abi` values to the example values

madsmtm:


ehuss:
Can you say more about why this is suggesting to add these values? The list here isn't intended to be exhaustive, but just a small sample of some representative values.

madsmtm:
Yeah, I know the list isn't exhaustive, but looking at it from a user perspective, I would've been unsure whether these were in `target_env` or in `target_abi`

ehuss (approved):
Seems fine to add, I just want to be cautious since people can make the argument about being uncertain of the values for any target, and we don't want to list the values for every target here. 

Perhaps in the future we can document every target's cfg values in the rustc book, which can help people with that confusion.

PR #1509, opened 2024-06-07
union syntax fix for empty field list

nefelitav:
Macros can accept and generate unions with empty field lists, which means the rule about prohibiting empty union field lists is more about semantics rather than syntax. This can be seen by compiling [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a3c7c2e0cafb924ead10ded4dd93a4b1).


ehuss (approved):
Thanks!

PR #1512, opened 2024-06-16
Reword "object safety" as "dyn compatibility"

kornelski:
"Object safe" is not a self-explanatory term. 

The the word "object" doesn't have a clear definition in Rust's context, and to users it may not be obvious that the term refers to *trait* objects specifically.

The "safe" part is not good either, because it could be misunderstood as relating to "safe Rust" or not-`unsafe`.

However, the term "object safe" is used in a lot of places, so I propose a minimal backwards-compatible fix by calling this concept "trait-object safety". Addition of the word "trait" adds context, and clearly connects it to trait objects.

Users may only know this concept from `Box<dyn Trait>` types, so I'm also emphasising `dyn Trait` compatibility.

More in the forum thread:
https://internals.rust-lang.org/t/object-safety-is-a-terrible-term/21025

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -175,10 +175,13 @@ the hierarchy has its own collection of named entities.
 Types that can be referred to by a path directly. Specifically [enums],
 [structs], [unions], and [trait objects].
 
-### Object safe traits
+### Trait-object-safe traits
 
-[Traits] that can be used as [trait objects]. Only traits that follow specific
-[rules][object safety] are object safe.
+[Traits] that can be used in `dyn Trait` types, and allow creation of
+[trait objects]. Only traits that follow specific [rules][object safety] are
+trait-object safe.
+
+This is also called *object safety*.
 
 ### Path
 
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -62,12 +62,12 @@ trait Seq<T> {
 }
 ```
 
-## Object Safety
+## Trait-Object Safety {#object-safety}
 
-Object safe traits can be the base trait of a [trait object]. A trait is
-*object safe* if it has the following qualities (defined in [RFC 255]):
+An object-safe trait can be used in a `dyn Trait` type, and be the base trait of a [trait object].
+A trait is *object safe* if it has the following qualities (defined in [RFC 255]):
 
-* All [supertraits] must also be object safe.
+* All [supertraits] must also be trait-object safe.
 * `Sized` must not be a [supertrait][supertraits]. In other words, it must not require `Self: Sized`.
 * It must not have any associated constants.
 * It must not have any associated types with generics.
@@ -93,7 +93,7 @@ Object safe traits can be the base trait of a [trait object]. A trait is
 # use std::rc::Rc;
 # use std::sync::Arc;
 # use std::pin::Pin;
-// Examples of object safe methods.
+// Examples of trait-object safe methods.
 trait TraitMethods {
     fn by_ref(self: &Self) {}
     fn by_ref_mut(self: &mut Self) {}
@@ -110,7 +110,7 @@ trait TraitMethods {
 ```
 
 ```rust,compile_fail
-// This trait is object-safe, but these methods cannot be dispatched on a trait object.
+// This trait is object-safe, but these methods cannot be dispatched on a trait object (such as `&dyn NonDispatchable`).
 trait NonDispatchable {
     // Non-methods cannot be dispatched.
     fn foo() where Self: Sized {}
@@ -135,7 +135,7 @@ obj.typed(1);  // ERROR: cannot call with generic type
 ```rust,compile_fail
 # use std::rc::Rc;
 // Examples of non-object safe traits.
-trait NotObjectSafe {
+trait NotDynCompatible {
     const CONST: i32 = 1;  // ERROR: cannot have associated const
 
     fn foo() {}  // ERROR: associated function without Sized
@@ -145,14 +145,14 @@ trait NotObjectSafe {
 }
 
 struct S;
-impl NotObjectSafe for S {
+impl NotDynCompatible for S {
     fn returns(&self) -> Self { S }
 }
-let obj: Box<dyn NotObjectSafe> = Box::new(S); // ERROR
+let obj: Box<dyn NotDynCompatible> = Box::new(S); // ERROR
 ```
 
 ```rust,compile_fail
-// Self: Sized traits are not object-safe.
+// Self: Sized traits are not trait-object-safe.
 trait TraitWithSize where Self: Sized {}
 
 struct S;
@@ -161,7 +161,7 @@ let obj: Box<dyn TraitWithSize> = Box::new(S); // ERROR
 ```
 
 ```rust,compile_fail
-// Not object safe if `Self` is a type argument.
+// Not trait-object safe if `Self` is a type argument.
 trait Super<A> {}
 trait WithSelf: Super<Self> where Self: Sized {}
 
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -62,7 +62,7 @@ trait Seq<T> {
 }
 ```
 
-## Trait-Object Safety {#object-safety}
+## Trait-Object Safety <a id="object-safety"></a>
 
 An object-safe trait can be used in a `dyn Trait` type, and be the base trait of a [trait object].
 A trait is *object safe* if it has the following qualities (defined in [RFC 255]):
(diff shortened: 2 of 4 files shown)


kornelski:
It's odd that the link checker doesn't get the right anchor, even tough when I run `mdbook build` locally, it writes the correct one. It seems like `{#anchor}` syntax isn't always supported?

ehuss:
To support heading attributes, the version of mdbook in https://github.com/rust-lang/reference/blob/0b805c65804019b0ac8f2fe3117afad82a6069b8/.github/workflows/main.yml#L22 needs to be updated to the latest version (0.4.40).

traviscross:
We discussed this in a rustdocs call.  We're going to let the discussion on IRLO play out a bit further, then we may lang nominate this.

joshtriplett:
> However, the term "object safe" is used in a lot of places, so I propose a minimal backwards-compatible fix by calling this concept "trait-object safety".

We've definitely been using the term "object safe" or "object safety" for a long time in many places, so we don't want to leave people unable to look up that term. But I don't think that means we need to leave the term substantially unmodified; I think that means we need to leave plenty of pointers to what we used to call it.

I would suggest, instead, that we fully rebrand this as "`dyn Trait` compatible" (or, when appropriate, "compatible with `dyn Trait`" or "`dyn`-compatible"), and then include notes saying things like "This is sometimes also called 'trait-object safety', 'object safety', or 'object safe'." With that, plus keeping the old link anchors working (though we should add new ones as well to avoid making the old ones look like the canonical names), I think that'll be sufficient to not abandon people who already know the existing term.

traviscross:
@rustbot labels -S-waiting-on-team

When [this ongoing FCP](https://github.com/rust-lang/lang-team/issues/286) completes, we will have decided to call traits that are compatible with `dyn` "dyn compatible" rather than "object safe".

src/glossary.md
```diff
@@ -175,10 +175,13 @@ the hierarchy has its own collection of named entities.
...
-### Object safe traits
+### Trait-object-safe traits
 
-[Traits] that can be used as [trait objects]. Only traits that follow specific
-[rules][object safety] are object safe.
+[Traits] that can be used in `dyn Trait` types, and allow creation of
+[trait objects]. Only traits that follow specific [rules][object safety] are
+trait-object safe.
+
+This is also called *object safety*.
```
fmease, line 184:
```suggestion
[Traits] that can be used in [trait object types][trait objects] (`dyn Trait`). Only traits that follow specific [rules][dyn-compatibility] are dyn compatible.

Formerly known as *object safety*.
```

\+ updating the second part of the reference-style MarkDown link.

TC, let me know if you agree with my suggestion.

src/items/traits.md
```diff
@@ -62,12 +62,12 @@ trait Seq<T> {
 }
 ```
 
-## Object Safety
+## Trait-Object Safety <a id="object-safety"></a>
```
fmease, line 65:
```suggestion
## Dyn Compatibility
```

NB: I'd love the URL fragment (named anchor) to get changed to `dyn-compatibility` and have an HTTP redirect from `object-safety`. Although I haven't checked if the Reference is set up to support redirects. I know that the rustc-dev-guide is.

src/items/traits.md
```diff
@@ -93,7 +93,7 @@ Object safe traits can be the base trait of a [trait object]. A trait is
 # use std::rc::Rc;
 # use std::sync::Arc;
 # use std::pin::Pin;
-// Examples of object safe methods.
+// Examples of trait-object safe methods.
```
fmease, line 96:
```suggestion
// Examples of dyn-compatible methods.
```

Though strictly speaking we haven't defined "dyn compatible" for methods, only for traits ("type mismatch"). We could be more explicit.

src/items/traits.md
```diff
@@ -62,12 +62,12 @@ trait Seq<T> {
 }
 ```
 
-## Object Safety
+## Trait-Object Safety <a id="object-safety"></a>
 
-Object safe traits can be the base trait of a [trait object]. A trait is
-*object safe* if it has the following qualities (defined in [RFC 255]):
+An object-safe trait can be used in a `dyn Trait` type, and be the base trait of a [trait object].
+A trait is *object safe* if it has the following qualities (defined in [RFC 255]):
```
fmease, line 68:
```suggestion
A dyn-compatible trait can be used in a trait object type (`dyn Trait`), and be the base trait of a [trait object].
A trait is *dyn compatible* if it has the following qualities (defined in [RFC 255]):
```

src/items/traits.md
fmease: We should add a paragraph or footnote pointing out that this concept is "formerly known as object safety".

src/items/traits.md
```diff
@@ -62,12 +62,12 @@ trait Seq<T> {
...
 
-* All [supertraits] must also be object safe.
+* All [supertraits] must also be trait-object safe.
```
fmease, line 70:
```suggestion
* All [supertraits] must also be dyn compatible.
```

src/items/traits.md
fmease: Unrelated to the renaming: The dyn-compatibility rules have been updated over time (for example, "dyn-incompatible" associated *types* marked with `where Self: Sized`  no longer render the trait dyn-incompatible). So stating that the rules were defined in RFC 255 is misleading or at the very least outdated. That should probably be changed as part of a separate PR though.

src/items/traits.md
```diff
@@ -110,7 +110,7 @@ trait TraitMethods {
 ```
 
 ```rust,compile_fail
-// This trait is object-safe, but these methods cannot be dispatched on a trait object.
```
fmease, line 113: You should double-check if there are any other mentions of `/(non[\- _])?obj(ect)?[\- _](un)?safe(ty)?/i`. and update them accordingly.

src/items/traits.md
```diff
@@ -135,7 +135,7 @@ obj.typed(1);  // ERROR: cannot call with generic type
 ```rust,compile_fail
 # use std::rc::Rc;
 // Examples of non-object safe traits.
-trait NotObjectSafe {
+trait NotDynCompatible {
```
fmease, line 138:
```suggestion
trait DynIncompatible {
```

\+ usage sites

src/items/traits.md
```diff
@@ -135,7 +135,7 @@ obj.typed(1);  // ERROR: cannot call with generic type
 ```rust,compile_fail
 # use std::rc::Rc;
 // Examples of non-object safe traits.
```
fmease, line 137:
```suggestion
 // Examples of dyn-incompatible traits.
```

\+ all other occurrences of this phrasing in this repo.

src/types/trait-object.md
```diff
@@ -7,8 +7,8 @@
 > _TraitObjectTypeOneBound_ :\
 > &nbsp;&nbsp; `dyn`<sup>?</sup> [_TraitBound_]
 
-A *trait object* is an opaque value of another type that implements a set of
-traits. The set of traits is made up of an [object safe] *base trait* plus any
+A *trait object* is an opaque value of a `dyn`-prefixed type that implements a set of
```
fmease, line 10: Minor nitpick: "Dyn-prefixed" is not edition-aware.

src/types/trait-object.md
fmease, line 1: update analogously

fmease (changes requested):

src/items/traits.md
fmease: > NB: I'd love the URL fragment (named anchor) to get changed to dyn-compatibility and have an HTTP redirect from object-safety

This would be quite nice because we show this URL literally in rustc's diagnostics. As a result of that we currently have a mismatch of terminology (diagnostic text vs. URL). Of course, that's expected but I'd like this to be temporary only (namely until this PR hits stable (rustc's diagnostics usually link to the stable channel of the Reference)).

ehuss: When renaming sections, you need to use a `<script>` tag to redirect the fragment, see https://github.com/rust-lang/reference/blob/c64e52a3d306eac0129f3ad6c6d8806ab99ae2e9/src/expressions/operator-expr.md?plain=1#L687-L699 as an example.

src/glossary.md
traviscross: Looks about right.  Probably worth making a full sentence out of the last line, e.g. "These were formerly known as object safe traits".

fmease:
Competing PR: #1666.

ehuss:
Closing in favor of #1666.

PR #1514, opened 2024-06-21
Add example of 1-ary tuple type

### LINKED ISSUE #1511: suggestion: add an example of a 1-ary tuple type in section 10.1.5
Here's the current list of examples in section 10.1.5 (https://doc.rust-lang.org/reference/types/tuple.html):

> Some examples of tuple types:
>
> () (unit)
> (f64, f64)
> (String, i32)
> (i32, String) (different type from the previous example)
> (i32, f64, Vec<String>, Option<bool>)

Curiously, there's no example of a 1-ary tuple type. I suggest adding one to the list, like this:
> (i32,)

This would make the list more complete and consistent, and would also help to illustrate the obligatory comma, not just in the *tuple expressions* for 1-ary tuple values, but also in the *type expressions* for 1-ary tuple types.

Suguis:
To illustrate the existence of 1-ary tuple types with the obligatory comma, as mentioned here #1511

Fixes #1511

### DIFF
--- a/src/types/tuple.md
+++ b/src/types/tuple.md
@@ -27,7 +27,7 @@ Its one value is also called *unit* or *the unit value*.
 Some examples of tuple types:
 
 * `()` (unit)
-* `(i32, )` (1-ary tuple)
+* `(i32,)` (1-ary tuple)
 * `(f64, f64)`
 * `(String, i32)`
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1516, opened 2024-06-24
Enable mdbook smart-punctuation.

ehuss:
This enables mdBook's [Smart Punctuation](https://rust-lang.github.io/mdBook/format/markdown.html#smart-punctuation) feature. This can help with entering special characters that can be difficult to enter or see in a text editor.


src/tokens.md
```diff
@@ -93,7 +93,7 @@ A suffix is a sequence of characters following the primary part of a literal (wi
 
 > **<sup>Lexer</sup>**\
 > SUFFIX : IDENTIFIER_OR_KEYWORD\
-> SUFFIX_NO_E : SUFFIX <sub>_not beginning with `e` or `E`_</sub>
+> SUFFIX_NO_E : SUFFIX <sub>_not beginning with `e` or `E`_</sub>
```
JoelMarcey, line 96: I don't see the change here. Tab/space?

JoelMarcey (approved):

src/tokens.md
ehuss: It is the no-break-space character (0xa0). I do not think it is particularly necessary here. If we really want to use no-break-space, then the invisible characters should be replaced with `&nbsp;` so it is visible in an editor.

PR #1520, opened 2024-07-01
Add mdbook-spec

ehuss:
This adds the use of mdbook-spec for the reference. The intent is to support some extensions that will help with the formatting that we eventually want to add, and to make authoring easier. Most of the changes are described in `authoring.md`.

This includes several changes split into separate commits. This includes a sync of the authoring document which was over in the rust-lang/spec repo.

This includes some tentative example changes in the CSS, but those will certainly want to change over time (particularly to support different theme colors). These aren't quite being used, yet. The current formatting of the reference should be identical as to before this PR. I will follow up with future PRs to do additional changes, such as removing explicit std links.

This will require some changes in rust-lang/rust in order to work properly, which I will handle before the next book sync.


ehuss:
For more context: It is noted that it is not optimal to have `mdbook-spec` in a separate repo. The reasons are complicated, due to the way the documentation is published. This is just intended as a short-term solution to unblock progress. Later I will try to figure out something so that we can move it to this repo and make it easier to develop and use.

traviscross (approved):

PR #1523, opened 2024-07-09
Reformat inline-assembly chapter as a spec chapter.

chorman0773:
This rewrites the inline-assembly chapter to use the agreed upon style and mdbook-spec extensions. It also more formally defines certain portions and adds several more examples.


src/inline-assembly.md
```diff
@@ -37,515 +50,878 @@ assert_eq!(x, 4 * 6);
...
+// The following lines are equivalent
+core::arch::asm!("mov rax, 5", "mov rcx, rax", out("rax") x, out("rcx") _); 
+core::arch::asm!("mov rax, 5\nmov rcx, rax", out("rax") x, out("rcx") _);  
+# }}
+```
+
+r[asm.invocation.operands]
+Each operand, other than an explicit register operand ([asm.operands.register]) shall be mentioned by at least one format_specifier in the *joined asm-string*. Explicit registers may not be referred to be a format_specifier.
+
+```rust,compile_fail
```
ehuss, line 136: These compile_fail annotations won't work with non-x86_64 architectures (since it won't fail to compile).

src/inline-assembly.md
```diff
@@ -37,515 +50,878 @@ assert_eq!(x, 4 * 6);
...
+asm_string_content := [*asm_string_piece]
+```
+
+## Invocation [asm.invocation]
+
+r[asm.invocation.asm]
+The [`core::arch::asm!`] macro shall be expanded in an expression context only. The input tokens shall match the `asm_inner` production. The expansion is [`unsafe`][static.expr.safety] and has type `()`, unless the option `noreturn` is specified, in which case it has type `!`.
+
+```rust
+# #[cfg(target_arch = "x86_64")]
```
ehuss, line 89: This method of using `cfg` along with an `fn main` confuses rustdoc and causes this to fail on non-x86_64 architectures. Can you look into some way to work around that issue?

For this particular example, does it need the cfg?

src/inline-assembly.md
```diff
@@ -37,515 +50,878 @@ assert_eq!(x, 4 * 6);
...
+```
+
+## Invocation [asm.invocation]
+
+r[asm.invocation.asm]
+The [`core::arch::asm!`] macro shall be expanded in an expression context only. The input tokens shall match the `asm_inner` production. The expansion is [`unsafe`][static.expr.safety] and has type `()`, unless the option `noreturn` is specified, in which case it has type `!`.
+
+```rust
+# #[cfg(target_arch = "x86_64")] 
+pub fn main(){
```
ehuss, line 90:
```suggestion
pub fn main() {
```

It would be good to generally try to ensure that the examples are formatted in a way that is pretty close to how rustfmt works.

src/inline-assembly.md
chorman0773: Yeah, I'll add a
```
# #[cfg(not(target_arch = "x86_64")]
# core::compile_error!("Tests currently require x86_64. Please add per-architecture tests to run off of x86_64");

File: src/inline-assembly.md (hunk 2 above)
Comment (chorman0773) [reply src/inline-assembly.md:89]: It does at least for the 5 architectures that asm is valid on (since if it gets tested off of those architectures, it will just error out)

File: theme/reference.css [hunk 4]
```diff
@@ -187,6 +195,25 @@ blockquote > p {
...
+    font-size: 0.9em;
+}
+
+.ayu  .target-specific blockquote,
+.navy .target-specific blockquote,
+.coal  .target-specific blockquote {
+    background: #008f2e;
+    box-shadow: inset -1px -1px 0 #03a430;
+}
+
```
JoelMarcey, line 216: Maybe we can get rid of any color-based changes for now and deal with look and feel in separate PRs. I wonder if you could for now add a new column to the table with a `*` or some other marker to make the representation instead?

JoelMarcey (changes requested):
The content changes look good to me, as much as I can gather not being a complete expert in this area.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
 # }
 ```
 
-## Syntax
+## Syntax [asm.syntax]
```
ehuss, line 53: It's a little unclear to me what these bracketed things are in the headers. Can you say more about why you added those, or what their intent is?

Like here, I would expect this to be something like:

```
## Syntax

r[asm.syntax]
```

chorman0773: That seems to me like it would be formatted weird. The Rule would appear dangling, not attached to anything.

The rules in the header are also supposed to be the anchor for the entire header section (although I think I got the cmark syntax wrong for that).

JoelMarcey: What does the `r` in front of the bracket refer to? Rule? Render? Where in the md-book documentation can I find that - I tried to search, but couldn't find it?

I am pretty indifferent on which way we go here. The anchor will still basically go to the top even if we format it as @ehuss suggests. I could almost argue another way that it should *above* the header similar to how we do other things.

```
r[asm.syntax]
## Syntax
```

But I don't feel strongly. Let's just pick something and go with it for now.

JoelMarcey (commented):
I am almost ready to approve this.

src/inline-assembly.md
chorman0773: Rule I believe - it's in mdbook-spec not mdbook proper.

Putting the anchor above doesn't quite do what I want, though - I want to replace the default anchor for the header, and I also don't really want the rendering of the rule in this case - it seems like it's supposed to have normal text below it (not a header or a code block), but we either hard switch the context with a header or we put another rule in immediately for the  next actual text entry. That's what I meant by "The Rule would appear dangling".

traviscross:
@Amanieu: We on the spec team would appreciate if you could have a look at this PR.

The context here is that we're reformatting and reworking spec chapters to make it more "spec like".  Principally, that means that we're adding an identifier to each block.  Since each identifier has a name, that means reworking things a bit such that "one idea" falls within each such block.

We 1) want to be sure you know this is going on, and 2) are happy with how it turned out, especially as there was some reworking, rewording, and the addition of some new content.

We're using this chapter as a model for reworking other ones (so please bear with us a bit as this is the first).

JoelMarcey:
Approved review. I say we aim for getting a 50% or more checkbox approval here before merging.

- [ ] @JoelMarcey 
- [ ] @ehuss 
- [ ] @traviscross 
- [ ] @m-ou-se 
- [ ] @pnkfelix 
- [ ] @Amanieu

src/inline-assembly.md
ehuss: > Where in the md-book documentation can I find that - I tried to search, but couldn't find it?

BTW, it is documented here: https://github.com/rust-lang/reference/blob/master/docs/authoring.md#rules

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+// The following snippet may have unpredictable results
+static mut FOO: i32 = 5;
+# #[cfg(target_arch = "x86_64")] { unsafe{
+let x: i32;
+core::arch::asm!("mov {:e}, dword ptr [{}+rip]", out(reg) x, sym FOO, options(nomem));
+# }}
+```
+
+r[asm.options.readonly]
+The `readonly` option may be specified. The behaviour is undefined if the assembly block modifies any allocation or activates any tag.
```
Amanieu, line 820: `readonly` also isn't allowed to synchronize with anything. For example you can't use `readonly` for a fence instruction.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+
+r[asm.invocation.expansion]
+The *joined asm-string* is expanded as defined in [asm.operands.expansion], replacing each `format_specifier` with the appropriate expansion for the operand. The resulting string is called the *expanded asm-string*.
+
+r[asm.invocation.syntax]
+The syntax of the *expanded asm-string* is a subset of the GNU AS syntax for the target. Invoking the macro with a *expanded asm-string* that does not match syntax requirements is *conditionally supported* and has *assembler dependent behaviour*. Invoking a directive that is not specified by [asm.directives] is *conditionally supported* and has *assembler dependent behaviour*.
+
+> [!TARGET-SPECIFIC]
+> On x86 and x86_64 targets, the syntax of the *expanded asm-string* acts as though the directive `.intel_syntax noprefix` is issued before parsing the *expanded asm-string*, except that the `option(att_syntax)` causes the syntax to act as though the directive `.att_syntax prefix` is issued before parsing the *expanded asm-string* instead.
+> On ARM and Aarch64 targets, the syntax of the *expanded asm-string* acts as though the directive `.syntax unified` is issued before parsing the *expanded asm-string*.
```
Amanieu, line 184: That's only on ARM, not AArch64.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+```rust,compile_fail
+# #[cfg(target_arch = "x86_64")] { unsafe{
+let x: i32;
+core::arch::asm!("xor edi, edi", "call exit@plt", out("edi") x, options(noreturn));
+# }}
+# #[cfg(not(target_arch = "x86_64"))] compile_error!("Inline Assembly Tests are not supported off of x86_64");
+```
+
+r[asm.options.global]
+A program shall not specify an option, other than the `att_syntax` option, in an invocation of the [`core::arch::global_asm!`] macro.
```
Amanieu, line 926: You can also use the `raw` option with global_asm. In fact it seems the `raw` option was omitted entirely in the spec?

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+# }}
+```
+
+> [!NOTE]
+> In particular, an asm block may be duplicated, for example if the containing function is inlined, or omitted from the output entirely.
+> As a consequence, asm blocks should not use directives that have non-idempotent non-local effects, or named labels and symbol definitions. 
+> Additionally, two asm blocks may not rely upon being adjacent in executable memory, even if they are adjacent in the source.
+
+> [!NOTE]
+> Local Labels (a decimal literal that doesn't solely consist of 0s and 1s) may be used freely if the asm block needs to define a label.
```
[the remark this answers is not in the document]
chorman0773: I think it should be noted so that people don't rely on the statement to say "But it says 'a decimal literal'", but it could be mentioned as being a bug. We can always loosen the specification, but we can't tighten it later after everything is all stabilized.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+
+r[asm.invocation.global-section]
+The *expanded asm-string* of a [`core::arch::global_asm!`] invocation acts as though a `.section` directive is issued before the *expanded asm-string*  which causes code to be generated in the default section on the target for executable code.
+
+> [!NOTE]
+> This section is typically named `.text`. 
+
+
+r[asm.invocation.prefix-instr]
+An *expanded asm-string* shall not end with an instruction that is interpreted as a prefix on the architecture. No Diagnostic is required
```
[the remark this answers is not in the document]
chorman0773: I didn't see a complete resolution in that thread (@joshtriplett seemed skeptical that it should be considered DB to have it in dead code), so I'm leaving it as the most restrictive rule. We can change the guarantees here if we find it handled correctly (or fixable), but this reformatting isn't intended to change the rules, just change how they are expressed.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+let x: i32;
+core::arch::asm!("mov eax, ecx", in("rcx") 5i64, out("eax") x);
+# }}
+```
+
+> [!NOTE]
+> Explicit Register Operands have no `ident` name and cannot be referred to by an operand specifier
+
+r[asm.operands.types]
+Each operand, other than a placeholder expression shall be of an integer type, floating-point type, function pointer type, pointer type, or target-specific vector type. These types are collectively called *asm operand types*. A pointer type is an *asm operand type* only if the pointee type has no metadata-type.
```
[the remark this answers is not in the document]
chorman0773: I can add that `input_expr`s have to be coerced to an *asm operand type*.

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+```
+
+```rust,ignore
+static FOO: () = ();
+# #[cfg(target_arch = "x86_64")]
+core::arch::global_asm!("/*{}*/", sym FOO);
+```
+
+r[asm.operands.clobbers_abi]
+A special operand `clobbers_abi` may be specified. If the `clobers_abi` operand is specified, the no reg_operand, other than an *explicit register operand*, shall be specified. When specified, it accepts a string literal which shall belong to a subset of the string literals accepted for an `extern` calling convention specification. The `clobbers_abi` special operand acts as though it is replaced by a `lateout` operand with an out-expr of `_` for each register considered by the specified calling convention to not be preserved by a function call.
```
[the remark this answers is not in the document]
chorman0773: I mention that it belongs to a subset, not any valid extern string already.

> shall belong to a subset of the string literals accepted for an `extern` calling convention specification

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
-| ARM | `qreg` | None | `q0` | `q` |
-| ARM | `qreg` | `e` / `f` | `d0` / `d1` | `e` / `f` |
-| RISC-V | `reg` | None | `x1` | None |
-| RISC-V | `freg` | None | `f0` | None |
-| LoongArch | `reg` | None | `$r1` | None |
-| LoongArch | `freg` | None | `$f0` | None |
 
-> **Notes**:
+r[asm.register.clobbers_only]
+Certain registers and register classes are *clobbers only*. Such register names or register classes shall not be specified by an operand, other than a clobbers output.
```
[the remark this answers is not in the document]
chorman0773: It's the register class that is specified as "Clobbers Only". Although I probably should restrict clobbers output in the operands section to "Shall be an *explicit register operand*"

src/inline-assembly.md
```diff
@@ -37,515 +50,892 @@ assert_eq!(x, 4 * 6);
...
+
+
+r[asm.evaluation.register-value]
+The behaviour is undefined upon exiting an asm block unless the stack pointer register and each operand-usable register not mentioned by an `out` , `lateout`, `inout`, or `inlateout` operand has the value the register held upon entry to the asm block.
+
+> [!TARGET-SPECIFIC]
+> In addition to operand-usable registers, certain other registers on a target may require being preserved, or have specific rules regarding the value at exit.
+> On x86 and x86-64 targets:
+> * The Direction flag (`flags.DF`) is clear upon entry and must be clear upon exit
+> * The x87 Stack (that is the `TOP` field of the floating-point status word, and each bit in the floating-point tag word) must be preserved and restored upon exit. If all x87 `st` registers are marked as clobbered, the stack is guaranteed to be empty on entry to the asm block (that is, `TOP` is set to `0x7` and the `ftw` is set to `0xFFFF`).
```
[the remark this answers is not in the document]
chorman0773: This is specifically for architectural register state. Other state is/would be handled by other rules.

src/inline-assembly.md
chorman0773: Are `acquire` operations not allowed for `readonly`? That seems odd to me.

src/inline-assembly.md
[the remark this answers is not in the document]
chorman0773: (Aside, is that a way I can force a coercion to a function pointer without knowing the signature :eyes:)

src/inline-assembly.md
Amanieu: I'm not 100% sure if it's legal for the compile to eliminate an acquire load if its result is unused. In fact LLVM currently doesn't do this even for relaxed loads.

src/inline-assembly.md
```diff
@@ -37,515 +50,901 @@ assert_eq!(x, 4 * 6);
...
+```
+
+```rust,ignore
+static FOO: () = ();
+# #[cfg(target_arch = "x86_64")]
+core::arch::global_asm!("/*{}*/", sym FOO);
+```
+
+r[asm.operands.clobbers_abi]
+A special operand `clobbers_abi` may be specified. If the `clobers_abi` operand is specified, the no `out`, `lateout`, `inout`, or `inlateout` reg_operand, other than an *explicit register operand*, shall be specified. When specified, it accepts a string literal which shall belong to a subset of the string literals accepted for an `extern` calling convention specification. The `clobbers_abi` special operand acts as though it is replaced by a `lateout` operand with an out-expr of `_` for each register considered by the specified calling convention to not be preserved by a function call.
```
Amanieu, line 418:
```suggestion
A special operand `clobbers_abi` may be specified. If the `clobers_abi` operand is specified, then no `out`, `lateout`, `inout`, or `inlateout` reg_operand, other than an *explicit register operand*, shall be specified. When specified, it accepts a string literal which shall belong to a subset of the string literals accepted for an `extern` calling convention specification. The `clobbers_abi` special operand acts as though it is replaced by a `lateout` operand with an out-expr of `_` for each register considered by the specified calling convention to not be preserved by a function call. 
```

src/inline-assembly.md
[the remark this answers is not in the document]
Amanieu: At the moment we specifically only perform the 2 coercions I mentioned: https://github.com/rust-lang/rust/blob/f00f85091904530ee8f5b29251c53571e0062a76/compiler/rustc_hir_typeck/src/expr.rs#L3253-L3272

Also we accept `MaybeUninit` versions of all allowed types: https://github.com/rust-lang/rust/pull/114790

src/inline-assembly.md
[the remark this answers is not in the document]
Amanieu: That's fair, we can update it in a later PR.

Amanieu (approved):

src/inline-assembly.md
chorman0773: I've changed this to have the rule after. I'd like to have a more extended discussion about this, but it's a mechanical change to revert it anyways so I'm not going to block this PR on it.  I'll open a policy issue for it though, and would like to put it on the agenda for tomorrow. Should that issue go on rust-lang/reference or rust-lang/spec?

JoelMarcey (approved):

src/inline-assembly.md
chorman0773: File [spec#71](https://github.com/rust-lang/spec/issues/71)

traviscross:
Let's use rfcbot here.  We should FCP this since it's the model for other chapters.

I'm going to propose FCP merge, even though I still need to review carefully, because I certainly want this in *some* form.  And I'll check Joel's box as he's checked off above.

(I'll file a concern to hold space for that review.)

(We'll see if rfcbot works with T-spec here.  It's the first go of this.)

@rfcbot fcp merge

traviscross:
@rfcbot concern tc-wants-to-review-carefully

src/inline-assembly.md
```diff
@@ -594,74 +1031,67 @@ The following directives are guaranteed to be supported by the assembler:
...
+> - `.nops`
+> - `.code16`
+> - `.code32`
+> - `.code64`
+> Use of `.code16`, `.code32`, and `.code64` directives are only supported if the state is reset to the default before exiting the assembly block.
+> 32-bit x86 uses `.code32` by default, and x86_64 uses `.code64` by default.
+
+r[asm.directives.arm]
+
+> The following directives are guaranteed to be supported on 32-bit ARM platforms
```
JoelMarcey, line 1089: Isn't this `[!TARGET-SPECIFIC]` too?

chorman0773:
Discussion re. https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/.60.2Esection.60.20directives.2E

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+
+r[asm.invocation.global-section]
+The *expanded asm-string* of a [`core::arch::global_asm!`] invocation acts as though a `.section` directive is issued before the *expanded asm-string*  which causes code to be generated in the default section on the target for executable code.
+
+> [!NOTE]
+> This section is typically named `.text`. 
+
+
+r[asm.invocation.prefix-instr]
+An *expanded asm-string* shall not end with an instruction that is interpreted as a prefix on the architecture. No Diagnostic is required
```
ehuss, line 241:
```suggestion
An *expanded asm-string* shall not end with an instruction that is interpreted as a prefix on the architecture. No diagnostic is required.
```

Also, I'm not sure why it is mentioning whether a diagnostic is required. Can you say more why it says that?

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+asm_string_piece := non_format_char / format_specifier / format_escape
+asm_string_content := [*asm_string_piece]
+```
+
+## Invocation 
+
+r[asm.invocation]
+
+r[asm.invocation.asm]
+The [`core::arch::asm!`] macro shall be expanded in an expression context only. The input tokens shall match the `asm_inner` production. The expansion is [`unsafe`][static.expr.safety] and has type `()`, unless the option `noreturn` is specified, in which case it has type `!`.
```
ehuss, line 92: This link to `unsafe` is broken since there aren't any rules for that, yet. Can you make sure to link directly until those sections are ready? In this case,

```
[`unsafe`]: unsafety.md
```

src/inline-assembly.md
```diff
@@ -1,21 +1,34 @@
...
+The macros [`core::arch::asm!`] and [`core::arch::global_asm!`] expand to inline assembly syntax when used in the expression position and item position respectively. The macros shall not be expanded in any other context.
 
-[`asm!`]: ../core/arch/macro.asm.html
-[`global_asm!`]: ../core/arch/macro.global_asm.html
+> [!NOTE]
+> The expansion of the macros has no stable syntax equivalent. This section will refer to the expansion of the macro, rather than the surface syntax.
 
-Support for inline assembly is stable on the following architectures:
+r[asm.safety]
+The macro [`core::arch::asm!`] shall be expanded only within an `unsafe` block.
```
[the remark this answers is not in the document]
chorman0773: Does it? Shall is a static constraint on the program, so program is ill-formed if violated.

src/inline-assembly.md
chorman0773: It is, because it can just take the value from the last store that already *happens-before* so it doesn't make any new side effects visible. LLVM is just understably conservative arround optimizing atomics.

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+> [!TARGET-SPECIFIC]
+> On x86 and x86_64 targets, the syntax of the *expanded asm-string* acts as though the directive `.intel_syntax noprefix` is issued before parsing the *expanded asm-string*, except that the `option(att_syntax)` causes the syntax to act as though the directive `.att_syntax prefix` is issued before parsing the *expanded asm-string* instead.
+> On ARM targets, the syntax of the *expanded asm-string* acts as though the directive `.syntax unified` is issued before parsing the *expanded asm-string*.
+
+r[asm.invocation.duplication]
+The number of times, locations, and the order in which a given invocation of [`core::arch::asm!`] is expanded is unspecified.
+
+```rust,ignore
+// The following code may have suprising results, and may fail to compile or link. 
+// The results, including whether it succesfully compiles, may depend on non-local use sites of the function, and on optimization settings.
```
ehuss, line 199: Can you make sure that the code examples are wrapped roughly at 80 columns? The way it gets presented in the book requires horizontal scrolling which can be awkward.

Also, there are some misspellings here.

```suggestion
// The following code may have surprising results, and may fail to compile or
// link. The results, including whether it successfully compiles, may depend on
// non-local use sites of the function, and on optimization settings.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+>   - Floating-point condition flags in `$fcc[0-7]`.
+
+```rust,ignore
+# #[cfg(target_arch = "x86_64")] { unsafe{
+core::arch::asm!("cmp eax, eax", in("eax") 5, options(preserve_flags));
+# }}
+```
+
+r[asm.options.noreturn]
+The `noreturn` option may be specifed. An invocation of the [`core::arch::asm!`] macro that specifies the `noreturn` option expands to an expression of type `!`. The behaviour is undefined if an evaluation of the assembly block exits. The program shall not specify the `clobber_abi` specification, or an operand that is an `out`, `lateout`, `inout`, or `inlateout` operand.
```
ehuss, line 944:
```suggestion
The `noreturn` option may be specified. An invocation of the [`core::arch::asm!`] macro that specifies the `noreturn` option expands to an expression of type `!`. The behaviour is undefined if an evaluation of the assembly block exits. The program shall not specify the `clobber_abi` specification, or an operand that is an `out`, `lateout`, `inout`, or `inlateout` operand. 
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+
+## Options 
+
+r[asm.options]
+
+r[asm.options.general]
+An options-spec provided in the asm invocation places constraints on the assembly block. 
+
+r[asm.options.att_syntax]
+The `att_syntax` option may be specfied on the x86 and x86_64 target. The program shall not specify the `att_syntax` option on any other target.
```
ehuss, line 824:
```suggestion
The `att_syntax` option may be specified on the x86 and x86_64 target. The program shall not specify the `att_syntax` option on any other target.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+
+```rust,compile_fail
+# #[cfg(target_arch = "x86_64")] { unsafe{
+core::arch::asm!("mov dword ptr [FOO+rip], 3", options(readonly, nomem));
+# }}
+# #[cfg(not(target_arch = "x86_64"))] compile_error!("Inline Assembly Tests are not supported off of x86_64");
+```
+
+r[asm.options.pure]
+The `pure` option may be specfied. The evaluation of the assembly block shall not produce any observable behaviour, consume input, or terminate execution, and the implementation may assume that the outputs of the assembly block depends only on the inputs and the contents of any allocation. If the program specifies the `pure` option, it shall specify either the `nomem` or `readonly` option.
```
ehuss, line 886:
```suggestion
The `pure` option may be specified. The evaluation of the assembly block shall not produce any observable behaviour, consume input, or terminate execution, and the implementation may assume that the outputs of the assembly block depends only on the inputs and the contents of any allocation. If the program specifies the `pure` option, it shall specify either the `nomem` or `readonly` option.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
-| x86-64 | `"C"`, `"system"` (on Windows), `"efiapi"`, `"win64"` | `ax`, `cx`, `dx`, `r[8-11]`, `xmm[0-31]`, `mm[0-7]`, `k[0-7]`, `st([0-7])`, `tmm[0-7]` |
-| x86-64 | `"C"`, `"system"` (on non-Windows), `"sysv64"` | `ax`, `cx`, `dx`, `si`, `di`, `r[8-11]`, `xmm[0-31]`, `mm[0-7]`, `k[0-7]`, `st([0-7])`, `tmm[0-7]` |
-| AArch64 | `"C"`, `"system"`, `"efiapi"` | `x[0-17]`, `x18`\*, `x30`, `v[0-31]`, `p[0-15]`, `ffr` |
-| ARM | `"C"`, `"system"`, `"efiapi"`, `"aapcs"` | `r[0-3]`, `r12`, `r14`, `s[0-15]`, `d[0-7]`, `d[16-31]` |
-| RISC-V | `"C"`, `"system"`, `"efiapi"` | `x1`, `x[5-7]`, `x[10-17]`, `x[28-31]`, `f[0-7]`, `f[10-17]`, `f[28-31]`, `v[0-31]` |
-| LoongArch | `"C"`, `"system"`, `"efiapi"` | `$r1`, `$r[4-20]`, `$f[0-23]` |
+> [!TARGET-SPECIFIC]
+> The correspondance between the operation performed by the asm block is target-dependant and implementation-dependant, subject to the rules set in [asm.operands]. Unless the program modifies the execution state, the basic operation performed by the asm block is the one performed by executing the sequence of instructions specified in the *expanded asm-string* starting with the first instruction.
```
ehuss, line 777:
```suggestion
> The correspondence between the operation performed by the asm block is target-dependant and implementation-dependant, subject to the rules set in [asm.operands]. Unless the program modifies the execution state, the basic operation performed by the asm block is the one performed by executing the sequence of instructions specified in the *expanded asm-string* starting with the first instruction.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
-> Notes:
-> - On AArch64 `x18` only included in the clobber list if it is not considered as a reserved register on the target.
+r[asm.evaluation.reg-values]
+The value of each register mentioned in an input operand is set according to [asm.operands] before evaluating any instructions in the asm block. The value of each other *operand-usable register* is unspecified. The value of all other registers is target-dependant.
+
+> [!NOTE]
+> The target may define that the register value (or some portion thereof) is undefined.
+
+r[asm.evaluation.constraints]
+Certain constraints may be placed on the asm block, and on the requirements of the correspondance, by default or by an option explicitly specified on the asm block. The behaviour is undefined if any such constraint is violated.
```
ehuss, line 786:
```suggestion
Certain constraints may be placed on the asm block, and on the requirements of the correspondence, by default or by an option explicitly specified on the asm block. The behaviour is undefined if any such constraint is violated.
```

Also, in general, I don't understand what "requirements of the correspondance" means.

Also, I'm in general not understanding this rule. What constraints is it referring to? Can you point to the original chapter that this is referring to?

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+let x: i32;
+core::arch::asm!("mov {output}, {input}", input = in(reg) 5i64, output = out(reg) x);
+# }}
+```
+
+r[asm.invocation.expansion]
+If the `raw` option is not specified, the *joined asm-string* is expanded as defined in [asm.operands.expansion], replacing each `format_specifier` with the appropriate expansion for the operand. The resulting string is called the *expanded asm-string*. If the `raw` option is specified, the *expanded asm-string* is the *joined asm-string* verbatim. 
+
+r[asm.invocation.syntax]
+The syntax of the *expanded asm-string* is a subset of the GNU AS syntax for the target. Invoking the macro with a *expanded asm-string* that does not match syntax requirements is *conditionally supported* and has *assembler dependent behaviour*. Invoking a directive that is not specified by [asm.directives] is *conditionally supported* and has *assembler dependent behaviour*.
```
ehuss, line 188: Can you make sure to use American English? (Here and elsewhere in the document.)

```suggestion
The syntax of the *expanded asm-string* is a subset of the GNU AS syntax for the target. Invoking the macro with a *expanded asm-string* that does not match syntax requirements is *conditionally supported* and has *assembler dependent behavior*. Invoking a directive that is not specified by [asm.directives] is *conditionally supported* and has *assembler dependent behavior*.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
-## ABI clobbers
+## Behaviour of an asm block 
 
-The `clobber_abi` keyword can be used to apply a default set of clobbers to an `asm!` block.
-This will automatically insert the necessary clobber constraints as needed for calling a function with a particular calling convention: if the calling convention does not fully preserve the value of a register across a call then `lateout("...") _` is implicitly added to the operands list (where the `...` is replaced by the register's name).
+r[asm.evaluation]
 
-`clobber_abi` may be specified any number of times. It will insert a clobber for all unique registers in the union of all specified calling conventions.
+r[asm.evaluation.general]
+Each evaluation of an asm block (invocation of [`core::arch::asm!`]) shall perform an operation that correpsonds to the result of a valid sequence of operations on the Minirust Abstract Machine. The behaviour is undefined if the operations performed by the asm block do not validly correspond to a valid sequence of Minirust operations.
```
ehuss, line 771:
```suggestion
Each evaluation of an asm block (invocation of [`core::arch::asm!`]) shall perform an operation that corresponds to the result of a valid sequence of operations on the Minirust Abstract Machine. The behavior is undefined if the operations performed by the asm block do not validly correspond to a valid sequence of Minirust operations.
```

src/inline-assembly.md
```diff
@@ -37,515 +52,937 @@ assert_eq!(x, 4 * 6);
...
+#}}
+# #[cfg(not(target_arch = "x86_64"))] compile_error!("Inline Assembly Tests are not supported off of x86_64");
+```
+
+
+r[asm.operands.in-expr]
+An `input_expr` shall be a value expression of an *asm operand type* or an *input coerceable type*. If the expression is of an *input coerceable type*, it is coerced to an *asm operand type*. 
+
+r[asm.operands.out-expr]
+An `output_expr` shall be the placeholder expression `_` or a (potentially unitialized) place expression of an *asm operand type*. If the place expression is initialized, it shall be a mutable place.
```
ehuss, line 353:
```suggestion
An `output_expr` shall be the placeholder expression `_` or a (potentially uninitialized) place expression of an *asm operand type*. If the place expression is initialized, it shall be a mutable place.
```

ehuss (commented):
This seems to be missing the changes from https://github.com/rust-lang/reference/pull/1453.

In general, can you be sure to spell-check your work?

src/inline-assembly.md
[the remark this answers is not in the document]
Monadic-Cat: As a casual reader, I agree that it could stand to be rephrased to mention a compiler error, or state some similarity to calling an `unsafe fn`.

Though the below compile_fail test, which you've asked to have stylistically called out as "this will fail to compile", might serve to alleviate the concern a little bit here. But it would be nice if this sentence could avoid being misleading in the first place.

src/inline-assembly.md
chorman0773: The presumption is that a diagnostic is required, and that the compiler is forbidden to compile an ill-formed program (one that violates a constraint like that one). That overrides the rule by telling the compiler that it's free to ignore the violation and translate the program notwithstanding, but that the result of the program is allowed to be anything.

chorman0773:
I'm quickly adding "Terms used in the Specification" to the glossary in order to resolve some of the above wording concerns. We'll need the terms defined anyways, and we might as well have them sooner rather than later.

src/inline-assembly.md
```diff
@@ -39,513 +54,964 @@ assert_eq!(x, 4 * 6);
...
+
+> [!NOTE]
+> This section is typically named `.text`.
+
+
+r[asm.invocation.prefix-instr]
+An *expanded asm-string* shall not end with an instruction that is interpreted as a prefix on the architecture. No diagnostic is required.
+
+> [!TARGET-SPECIFIC]
+> On x86 and x86-64, the `lock`, `repnz`, `rep`, `repz`, as well as GNU AS specific address-size, data-size, and explicit rex, vex, and evex prefixes.
```
ehuss, line 253: This seems to be missing a verb. Is this saying something like:

```suggestion
> On x86 and x86-64, prefix instructions include the `lock`, `repnz`, `rep`, `repz`, as well as GNU AS specific address-size, data-size, and explicit rex, vex, and evex prefixes.
```

src/inline-assembly.md
```diff
@@ -39,513 +54,964 @@ assert_eq!(x, 4 * 6);
...
-## ABI clobbers
+## Behaviour of an asm block
 
-The `clobber_abi` keyword can be used to apply a default set of clobbers to an `asm!` block.
-This will automatically insert the necessary clobber constraints as needed for calling a function with a particular calling convention: if the calling convention does not fully preserve the value of a register across a call then `lateout("...") _` is implicitly added to the operands list (where the `...` is replaced by the register's name).
+r[asm.evaluation]
 
-`clobber_abi` may be specified any number of times. It will insert a clobber for all unique registers in the union of all specified calling conventions.
+r[asm.evaluation.general]
+Each evaluation of an *asm block* shall perform an operation that correpsonds to the result of a valid sequence of operations on the Minirust Abstract Machine. The behaviour is undefined if the operations performed by the asm block do not validly correspond to a valid sequence of Minirust operations.
```
ehuss, line 795: This seems like a new claim that we will be defining things in terms of Minirust. Is that something we have decided to do?

chorman0773: So there are two separate "new claims" here - one is that the operational semantics will be defined in terms of Minirust in general (that had a whole meeting on it, where we took an hour and a half to say "Yes"... actually, no, we took 5 minutes to say "Yes" and then spent the other hour 25 minutes debating "to what extent"), and the other is that inline assembly must affect the Rust AM in a way that Rust Code could possibly have done.

I believe the latter has been the opinion of the opsem team since before it was the opsem team, and is more generally, makes the claim that compilers can optimize Rust in the face of unanalyzed code (which has been assumed as true since we started plugging Rust into llvm).

chorman0773: In practice, this means that this code has UB, because it doesn't correspond back to a valid sequence of Minirust operations (that do not themselves have undefined behaviour):
```rust
let x: i32 = 0;
asm!("mov dword ptr [{}], 5", in(reg) &x);
println!("{x}"); // May be turned into `println!("0");`
```

The code writes to a `Frozen` tag, which is UB. But if inline assembly isn't constrained to simply doing things (to the AM) Rust code itself can do, then it could circumvent that rule.

ehuss: > that had a whole meeting on it, where we took an hour and a half to say "Yes"... actually, no, we took 5 minutes to say "Yes" and then spent the other hour 25 minutes debating "to what extent")

I don't recall having a team decision on this point, and I don't see any discussions about that decision in the [meeting notes](https://hackmd.io/team/rust-spec-team?nav=overview). 

I'm not opposed to Minirust as an eventual basis to use for defining things, but I'm not yet convinced that we are in a position where we are ready to do that *today*.

I'd also say that adding things like this are pretty heavy decisions that are probably best not to bundle in one large update. This makes it harder to review and make progress when there are many new changes included in one PR. Is it possible to work in smaller batches, and to keep new additions separate from things that are more editorially oriented?

src/inline-assembly.md
chorman0773: > Can you point to the original chapter that this is referring to?

This is the formalization of the "Rule of Inline Assembly" section's preamble. The constraints mentioned there are then elaborated in the rest of the section (for default constraints) and in [asm.options] (for option specific constraints). 

The correspondence is the correspondence between the abstract machine state and the physical machine state affected by the inline assembly - defined in the previous clause and target-specific admonition, and constrained in various sections (such as in [asm.operands] and [asm.registers]). IE. when you do X in assembly, what effect does that have on the Rust/Minirust AM, and how do we overall map both incoming state and outgoing state. The constraints then limit the results of the correspondence. For example, if the correspondence says that the operation wrote to some allocation, but the `nomem` option was set, you've violated that constraint on the asm block. Another example is the fact that if a register isn't mentioned by an `out`, `lateout`, `inout`, or `inlateout` operand, it must be preserved by the asm block. An asm block that then clobbers a register that is required to be preserved violates that constraint.

src/inline-assembly.md
chorman0773: I think we need to be able to say what an inline assembly block does, and I'm not sure there's any other model we can use. We know that it can do more than Rust code can.  For example, an asm block can yield a nondeterministic value, directly accessing the `pick` Minirust operation.

traviscross: @ehuss is right here.  Maybe we do...

> need to be able to say what an inline assembly block does

...in more detail than what the Reference chapter says today.  But that doesn't need to all happen in this PR.  And trying to do that is making it difficult to review this and make progress here.

Amanieu: FWIW this paragraph is not specific to inline assembly, it generally applies to any kind of FFI call.

chorman0773: Clarifying my statement at the T-spec meeting:
* I think we need *a* claim here, for the rest of the section and the following section ([asm.options]) to make sense.
* I don't think it necessarily needs to make a claim about Minirust.
* We know that the claim should allow *something* more than what Rust itself allows, but less than what the hardware allows.

traviscross:
@rfcbot concern decide-on-issue-1550-first

Let's decide, first, on the FCP in:

- https://github.com/rust-lang/reference/pull/1550

chorman0773:
This can be closed now, correct?

PR #1527, opened 2024-07-14
Suppress type length limit test and note that it is not enforced

compiler-errors:
Bit of a cop out, but I don't think we should actually do anything more here.

I think deprecating the limit attribute is a bit overkill, since users who add it to their code because nightly *just* started telling them it's necessary shouldn't be punished imo.

https://github.com/rust-lang/rust/pull/127670#issuecomment-2227383876

### DIFF
--- a/src/attributes/limits.md
+++ b/src/attributes/limits.md
@@ -35,9 +35,7 @@ a!{}
 
 ## The `type_length_limit` attribute
 
-> [!NOTE]
-> This limit is only enforced when the nightly `-Zenforce-type-length-limit`
-> flag is active.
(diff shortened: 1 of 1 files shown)


compiler-errors:
cc @ehuss

ehuss (approved):
Thanks! I pushed a small change to remove the custom admonition syntax. We have an extension to try to support that, but we aren't quite ready to use it, yet.

I don't think this is ultimately the way we'll want to document this attribute. Generally we try to avoid talking too much about the specifics of `rustc`, particularly nightly-specific things. I think for now it is OK to move forward with this (in particular, to fix toolstate).

Moving forward, we should probably consider this attribute to be deprecated if it isn't going to be doing anything useful. (Unless this change is temporary.)

PR #1528, opened 2024-07-15
Ignore admonitions in the broken link checker.

ehuss:
This fixes a problem where the broken reference link checker was triggering for `[!FOO]` admonitions used in mdbook-spec. The pulldown_cmark parser doesn't know about this custom syntax. Ideally this broken link check should be moved into mdbook itself, which can run *after* the preprocessors.


chorman0773:
Rule names are also triggering the checker.

PR #1530, opened 2024-07-16
add missing associated type link in type-aliases.md

cafkafk:
The other preceding paragraphs all link to `associated type` in the same place, for consistency, I think it makes sense to add a link here as well.


ehuss (approved):
Thanks!

PR #1538, opened 2024-07-20
Describe all cases for which the weak `dyn` keyword is parsed as a keyword

Veykril:
Had to figure this out for rust-analyzer and noticed that the reference is lacking here (arguably not an important detail as 2015 is rather old :)

Sourced from rustc here: https://github.com/rust-lang/rust/blob/339015920d0787ea14aee05fca2c205683c525f9/compiler/rustc_parse/src/parser/ty.rs#L722-L731 and https://github.com/rust-lang/rust/blob/339015920d0787ea14aee05fca2c205683c525f9/compiler/rustc_parse/src/parser/ty.rs#L84-L99


traviscross (approved):
Thanks @Veykril.  Looks good.

PR #1539, opened 2024-07-21
Add some basic docs for unsafe attrs

carbotaniuman:
Reference docs for https://github.com/rust-lang/rust/issues/123757

I'm not sure how much detail I need to go into for these changes, or how the reference itself is laid out, or even if I'm missing anything.

### DIFF
--- a/src/abi.md
+++ b/src/abi.md
@@ -66,11 +66,18 @@ item's name.
 Additionally, the item will be publicly exported from the produced library or
 object file, similar to the [`used` attribute](#the-used-attribute).
 
+This attribute is unsafe as an unmangled symbol may collide with another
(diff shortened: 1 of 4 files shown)


traviscross (approved):
Looks right to me.  We'll merge this when the stabilization PR for unsafe attributes is merged.

ehuss:
I pushed some fixes here.

The grammar didn't look correct to me. The _Attr_ rule is used in the `meta` fragment specifier, and the `cfg_attr` attribute, and it needs to include the new `unsafe` syntax.

Additionally, I updated the examples so they use the new form.

ehuss (approved):
Looks like https://github.com/rust-lang/rust/pull/128771 has merged, thanks!

PR #1540, opened 2024-07-22
re-structure 'invalid values' enumeration to instead define what is *valid*

RalfJung:
I've always been bothered by the fact that was are listing what is *invalid*, which just feels all backwards. So this PR changes that list to instead enumerate what is valid. We also move that list into its own subsection to give a bit more structure to this entire page.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -150,11 +150,11 @@ Whether a value is valid depends on the type:
 * A reference or [`Box<T>`] must be aligned, it cannot be [dangling], and it must point to a valid value
   (in case of dynamically sized types, using the actual dynamic type of the
   pointee as determined by the metadata).
-  Note that the last point (about pointing to a valid value) is still subject of debate.
+  Note that the last point (about pointing to a valid value) remains a subject of some debate.
 * The metadata of a wide reference, [`Box<T>`], or raw pointer must match
   the type of the unsized tail:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`.
-    (For raw pointers, this requirement is still subject of debate.)
+    (For raw pointers, this requirement remains a subject of some debate.)
   * Slice (`[T]`) metadata must be a valid `usize`.
     Furthermore, for wide references and [`Box<T>`], slice metadata is invalid
     if it makes the total size of the pointed-to value bigger than `isize::MAX`.
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -143,15 +143,18 @@ Whether a value is valid depends on the type:
 * A `str` value is treated like `[u8]`, i.e. it must be initialized.
 * An `enum` must have a valid discriminant, and all fields of the variant indicated by that discriminant must be valid at their respective type.
 * A `struct`, tuple, and array requires all fields/elements to be valid at their respective type.
-* For a `union`, the exact validity requirements are not decided yet. The following is certain:
-  * If the `union` has a zero-sized field, then all values are valid.
-  * If a value is valid for a particular `union` field, then it is valid for the union.
+* For a `union`, the exact validity requirements are not decided yet.
+  Obviously, all values that can be created entirely in safe code are valid.
+  If the union has a zero-sized field, then every possible value is valid.
+  Further details are [still being debated](https://github.com/rust-lang/unsafe-code-guidelines/issues/438).
 * A reference or [`Box<T>`] must be aligned, it cannot be [dangling], and it must point to a valid value
   (in case of dynamically sized types, using the actual dynamic type of the
   pointee as determined by the metadata).
+  Note that the last point (about pointing to a valid value) is still subject of debate.
 * The metadata of a wide reference, [`Box<T>`], or raw pointer must match
   the type of the unsized tail:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`.
+    (For raw pointers, this requirement is still subject of debate.)
   * Slice (`[T]`) metadata must be a valid `usize`.
     Furthermore, for wide references and [`Box<T>`], slice metadata is invalid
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -155,6 +127,46 @@ entire range, so it is important that the length metadata is never too large. In
...
+  (in case of dynamically sized types, using the actual dynamic type of the
+  pointee as determined by the metadata).
+* The metadata of a wide reference, `Box<T>`, or raw pointer must match
+  the type of the unsized tail:
+  * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`.
+  * Slice (`[T]`) metadata must be a valid `usize`.
+    Furthermore, for wide references and `Box<T>`, slice metadata is invalid
+    if it makes the total size of the pointed-to value bigger than `isize::MAX`.
+* If a type has a custom range of a valid values, then a valid value must be in that range.
+  In the standard library, this affects [`NonNull<T>`] and [`NonZero*`].
```
chorman0773, line 159: This should mention [`NonZero<T>`](https://doc.rust-lang.org/stable/core/num/struct.NonZero.html) now. Rather than [`NonZero*`], which is a broken link.
```suggestion
* If a type has a custom range of a valid values, then a valid value must be in that range.
  In the standard library, this affects [`NonNull<T>`] and [`NonZero<T>`].
```

src/behavior-considered-undefined.md
```diff
@@ -155,6 +127,46 @@ entire range, so it is important that the length metadata is never too large. In
...
+* A `char` value must not be a surrogate (i.e., must not be in the range `0xD800..=0xDFFF`) and must be below `char::MAX`.
+* A `!` value must never exist.
+* An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer must be
+  initialized, i.e., must not be obtained from [uninitialized memory][undef].
+* A `str` value is treated like `[u8]`, i.e. it must be initialized.
+* An `enum` must have a valid discriminant, and all fields of the variant indicated by that discriminant must be valid at their respective type.
+* A `struct`, tuple, and array requires all fields/elements to be valid at their respective type.
+* For a `union`, the exact validity requirements are not decided yet. The following is certain:
+  * If the `union` has a zero-sized field, then all values are valid.
+  * If a value is valid for a particular `union` field, then it is valid for the union.
```
chorman0773, line 148: This isn't exhaustive on what is possible with safe code. For example, it implicitly excludes the value `[0,0]` for the union:
```rust
#[repr(C)]
pub union Foo{
    x: (NonZero<u8>, u8),
    y: (u8, NonZero<u8>),
}
```
which can be created using the purely safe code:
```rust
let mut x = Foo{x: NonZero::new(1).unwrap(), y: 0};
x.y.0 = 0;
let x = x;
``` 

I'm not sure how to modify this to support that, but I'd expect we shouldn't exclude values that are definitely valid (because they are constructable in safe code) if we can avoid it.

src/behavior-considered-undefined.md
```diff
@@ -155,6 +127,46 @@ entire range, so it is important that the length metadata is never too large. In
...
+* A reference or `Box<T>` must be aligned, it cannot be [dangling], and it must point to a valid value
+  (in case of dynamically sized types, using the actual dynamic type of the
+  pointee as determined by the metadata).
+* The metadata of a wide reference, `Box<T>`, or raw pointer must match
+  the type of the unsized tail:
```
chorman0773, line 153: References to `Box<T>` should be links so they'll autolink later (currently I believe we need a reference link, but these will get fixed with mdbook-spec)
```suggestion
* A reference or [`Box<T>`] must be aligned, it cannot be [dangling], and it must point to a valid value
  (in case of dynamically sized types, using the actual dynamic type of the
  pointee as determined by the metadata).
* The metadata of a wide reference, [`Box<T>`], or raw pointer must match
  the type of the unsized tail:
```

src/behavior-considered-undefined.md
```diff
@@ -155,6 +127,46 @@ entire range, so it is important that the length metadata is never too large. In
...
+### Invalid values
+[invalid-values]: #invalid-values
+
+The Rust compiler assumes that all values produced during program execution are
+"valid", and producing an invalid value is hence immediate UB.
+
+Whether a value is valid depends on the type:
+* A [`bool`] value must be `false` (`0`) or `true` (`1`).
+* A `fn` pointer value must be non-null.
+* A `char` value must not be a surrogate (i.e., must not be in the range `0xD800..=0xDFFF`) and must be below `char::MAX`.
```
chorman0773, line 139:
```suggestion
* A `char` value must not be a surrogate (i.e., must not be in the range `0xD800..=0xDFFF`) and must be equal to below `char::MAX`.
```
Current wording suggests exclusion, IE. that `char::MAX` is an invalid `char`.

chorman0773 (changes requested):
Other than the things listed in the review, seems to be correct and editorially sound.

src/behavior-considered-undefined.md
RalfJung: I know that this doesn't cover everything one can do in safe code. It doesn't claim to do that.

RalfJung: If we want this to be able to capture everything safe code can do, we could try something like "if a value satisfies the validity invariant, then picking any (nested) field of an enum and writing a value valid at that field's type to that field yields a resulting union value that is still valid", or so. I think most readers would probably be more confused by this than it helps them...

This isn't meant to summarize all we know about unions, just some basic facts. Currently the page states nothing about unions. So another alternative would be to just say "is not decided yet" or "is subject to active discussion" and link to the [UCG issue](https://github.com/rust-lang/unsafe-code-guidelines/issues/438).

chorman0773: Maybe it could be put into a sub bullet or a note.

I just don't think it's a good idea to omit values that are possible to create in safe code.

traviscross (approved):

PR #1545, opened 2024-07-24
Spec abi chapter

chorman0773:
This rewrites the `abi` chapter, and adds [call compatibility](https://doc.rust-lang.org/core/primitive.fn.html#abi-compatibility) to the chapter.

### DIFF
--- a/src/abi.md
+++ b/src/abi.md
@@ -26,10 +26,13 @@ Two integer types are *abi compatible* if they have the same size and the same s
 > Two integer types with different signedness, such as `u8` and `i8` are not *abi compatible*.
 
 r[abi.compatibility.char]
-The type `char`is *abi compatible* with the type `u32`. 
+The type `char` is *abi compatible* with the type `u32`. 
 
 r[abi.compatibility.pointer]
 Two pointer types, `*mut T` and `*const U`, are *abi compatible* if the *metadata type*s of `T` and `U` are the same type. 
+
+> [!NOTE]
+> [`Sized`] types have a *metadata type* of `()`. 
 
 > [!NOTE]
 > With transitivity, this applies regardless of the mutability of either pointer type
@@ -38,14 +41,13 @@ r[abi.compatibility.reference-box]
 The types `&T`, `&mut T`, [`Box<T>`][core::boxed::Box], and [`NonNull<T>`][core::ptr::NonNull], are *abi compatible* with `*const T`
 
 > [!NOTE]
-> With transitivity,t hey are also *abi compatible** with each other, and with `*mut T`, as well as references/`Box` to different types that have the same *metadata type*.
-
+> With transitivity, they are also *abi compatible* with each other, and with `*mut T`, as well as references/`Box` to different types that have the same *metadata type*.
 
 r[abi.compatibility.core]
 The types [`MaybeUninit<T>`][core::mem::MaybeUninit], [`UnsafeCell<T>`][core::cell::UnsafeCell], and [`NonZero<T>`][core::num::NonZero], are *abi compatible* with `T`.
 
 r[abi.compatibility.transparent]
-A `struct` declared with the `transparent` representation is *abi compatible* with its field that does not have size 0 and alignment 1, if such a field exists
+A `struct` declared with the `transparent` representation is *abi compatible* with its field that does not have size 0 and alignment 1, if such a field exists.
 
 r[abi.compatibilty.zst]
 Two types, `T` and `U`, are *abi compatible* if both have size 0 and alignment 1.
@@ -68,30 +70,35 @@ Two function signatures are compatible if:
 * Each parameter of both signatures, in order, are *abi compatible*, and
 * Either both signatures have C-varargs, or neither signature does.
 
+> [!NOTE]
+> A signature is compatible with itself.
+
 r[abi.compatibility.simd-abi]
 A type has *simd abi requirements* if:
 * It is a type declared with the standard-library repr-attrbute `simd`,
 * It is a aggregate type, which has a type with *simd abi requirements* as a field.
+
+> [!NOTE]
+> The `repr(simd)` attribute cannot be used by Rust code, only by the standard library.
 
 r[abi.compatibility.simd-target-feature]
 A type with *simd abi requirements* may have one or more [*salient target features*][target_feature] . In the case of an aggregate type, the set of [*salient target features*][target_feature] is the union of the set of [*salient target features*][target_feature] of each field with *simd abi requirements*.
 
 > [!TARGET-SPECIFIC]
 > On x86 and x86-64, the [*salient target features*][target_feature] of the `simd` types are:
-> * [`__m128`], [`__m128i`], [`__m128f`], and [`__m128d`]: `sse`
-> * [`__m256`], [`__m256i`], [`__m256f`], and [`__m256d`]: `avx`
-> * [`__m512`], [`__m512i`], [`__m512f`], and [`__m512d`]: `avx512f` and `avx512vl`
+> * [`__m128`], [`__m128i`], [`__m128f`], and [`__m128d`] (128-bit vector types): `sse`
+> * [`__m256`], [`__m256i`], [`__m256f`], and [`__m256d`] (256-bit vector types): `avx`
+> * [`__m512`], [`__m512i`], [`__m512f`], and [`__m512d`] (512-bit vector types): `avx512f` and `avx512vl`
 
 r[abi.compatibility.call]
-A call to a function `f` via a function item or function pointer with a given signature `S` is valid only if the signature of `f` is *compatible* with the signature `S`, and:
-* The ABI tag of the function is `extern "Rust"`, or
-* If the type of any parameter, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each [*salient target features*][target_feature]of that type is either set at both the definition site of the function, and at the call site, or is set at neither site.
+A call to a function `f` via a function item or function pointer with a given signature `S` is valid if and only if the signature of the definition `f` is *compatible* with the signature `S`, and:
+* The ABI tag of the signature is `extern "Rust"`, or
+* If any parameter type, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each [*salient target features*][target_feature] of that type is either set at both the definition site of the function, and at the call site, or is set at neither site.
 
 The behaviour a call that is not valid is undefined.
 
 > [!NOTE]
 > the ABI tag `extern "Rust"` is the default when the `extern` keyword is not used (either to declare the function within an [`extern` block], or as a [function qualifier][extern functions]). Thus it is safe to call most functions that use simd types.
-
 
 [`__m128`]: https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m128.html
 [`__m128i`]: https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m128i.html
@@ -192,8 +199,6 @@ The *`no_mangle` attribute* may be specified as a built-in attribute, using the 
 r[abi.symbol-name.export_name]
 The *`export_name` attribute* may be specified as a built-in attribute, using the [_MetaNameValueStr_] syntax. The *export name* of an item with the *`no_mangle` attribute* is the content of `STRING_LITERAL`. 
 
-
-
 ## The `link_section` attribute
 
 r[abi.link_section]
@@ -234,10 +239,6 @@ pub static VAR1: u32 = 1;
 > * `.tbss`: Readable and Writable - Uninitialized and Thread-local.
 > 
 > This is not an exhaustive list, and generally extended versions of these section names such as `.text.foo`, are also defined with the same properties as the base section.
-> 
-> 
-
-
 
 [_MetaWord_]: attributes.md#meta-item-attribute-syntax
 [_MetaNameValueStr_]: attributes.md#meta-item-attribute-syntax
--- a/src/abi.md
+++ b/src/abi.md
@@ -74,16 +74,23 @@ A type has *simd abi requirements* if:
 * It is a aggregate type, which has a type with *simd abi requirements* as a field.
 
 r[abi.compatibility.simd-target-feature]
-A type with *simd abi requirements* may have one or more *salient target features* . In the case of an aggregate type, the set of *salient target features* is the union of the set of *salient target features* of each field with *simd abi requirements*.
+A type with *simd abi requirements* may have one or more [*salient target features*][target_feature] . In the case of an aggregate type, the set of [*salient target features*][target_feature] is the union of the set of [*salient target features*][target_feature] of each field with *simd abi requirements*.
 
 > [!TARGET-SPECIFIC]
-> On x86 and x86-64, the *salient target features* of the `simd` types are:
+> On x86 and x86-64, the [*salient target features*][target_feature] of the `simd` types are:
 > * [`__m128`], [`__m128i`], [`__m128f`], and [`__m128d`]: `sse`
 > * [`__m256`], [`__m256i`], [`__m256f`], and [`__m256d`]: `avx`
 > * [`__m512`], [`__m512i`], [`__m512f`], and [`__m512d`]: `avx512f` and `avx512vl`
 
 r[abi.compatibility.call]
-A call to a function `f` via a function item or function pointer with a given signature `S` is valid only if the signature of `f` is *compatible* with the signature `S`, and, if the type of any parameter, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each *salient target feature* of that type is either set at both the definition site of the function, and at the call site, or is set at neither site. The behaviour a call that is not valid is undefined.
+A call to a function `f` via a function item or function pointer with a given signature `S` is valid only if the signature of `f` is *compatible* with the signature `S`, and:
+* The ABI tag of the function is `extern "Rust"`, or
+* If the type of any parameter, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each [*salient target features*][target_feature]of that type is either set at both the definition site of the function, and at the call site, or is set at neither site.
+
+The behaviour a call that is not valid is undefined.
+
+> [!NOTE]
+> the ABI tag `extern "Rust"` is the default when the `extern` keyword is not used (either to declare the function within an [`extern` block], or as a [function qualifier][extern functions]). Thus it is safe to call most functions that use simd types.
 
 
 [`__m128`]: https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m128.html
@@ -100,7 +107,6 @@ A call to a function `f` via a function item or function pointer with a given si
 [`__m512d`]: https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m512d.html
 
 ## The `used` attribute
-
 
 r[abi.used]
 
@@ -231,6 +237,8 @@ pub static VAR1: u32 = 1;
 > 
 > 
 
+
+
 [_MetaWord_]: attributes.md#meta-item-attribute-syntax
 [_MetaNameValueStr_]: attributes.md#meta-item-attribute-syntax
 [`static` items]: items/static-items.md
@@ -240,3 +248,4 @@ pub static VAR1: u32 = 1;
 [function]: items/functions.md
 [item]: items.md
 [static]: items/static-items.md
+[target_feature]: attributes/codegen.md#the-target_feature-attribute
--- a/src/abi.md
+++ b/src/abi.md
@@ -59,13 +59,13 @@ r[abi.compatibility.fn-ptr]
 An `fn`-ptr type `T` is compatible with an `fn`-ptr type `U` if `T` and `U` have *abi compatible* tags.
 
 r[abi.compatibility.extern-tag]
-Two abi tags are *abi compatible* if:
+Two [abi tags][abi] are *abi compatible* if:
 * They are the same string, or
 * One tag is `"X"`, and the other is `"X-unwind"`
 
 r[abi.compatibility.signature]
 Two function signatures are compatible if:
-* The abi tags of both signatures are *abi compatible*,
+* The [abi tags][abi] of both signatures are *abi compatible*,
 * They have the same number of parameters, excluding C-varargs,
 * Each parameter of both signatures, in order, are *abi compatible*, and
 * Either both signatures have C-varargs, or neither signature does.
@@ -246,6 +246,7 @@ pub static VAR1: u32 = 1;
 [attribute]: attributes.md
 [extern functions]: items/functions.md#extern-function-qualifier
 [`extern` block]: items/external-blocks.md
+[abi]: items/external-blocks.md#abi
 [function]: items/functions.md
 [item]: items.md
 [static]: items/static-items.md
--- a/src/abi.md
+++ b/src/abi.md
@@ -19,53 +19,53 @@ Two types `T` and `U` are *abi compatible* if:
 > These properties ensure that *abi compatibility* is an equivalence relation.
 
 r[abi.compatibility.integer]
-Two integer types are *abi compatible* if they have the same size and the same signednes
+Two [integer types] are *abi compatible* if they have the same size and the same signednes
 
 > [!NOTE]
 > In particular, `usize` is *abi compatible* with `uN`, and `isize` is *abi compatible* with `iN` where `N` is the target_pointer_width. 
 > Two integer types with different signedness, such as `u8` and `i8` are not *abi compatible*.
 
 r[abi.compatibility.char]
-The type `char` is *abi compatible* with the type `u32`. 
+The type [`char`] is *abi compatible* with the type [`u32`][integer types]. 
 
 r[abi.compatibility.pointer]
-Two pointer types, `*mut T` and `*const U`, are *abi compatible* if the *metadata type*s of `T` and `U` are the same type. 
+Two [pointer types], `*mut T` and `*const U`, are *abi compatible* if the *metadata type*s of `T` and `U` are the same type. 
 
 > [!NOTE]
-> [`Sized`] types have a *metadata type* of `()`. 
+> [`core::marker::Sized`] types have a *metadata type* of `()`. 
 
 > [!NOTE]
 > With transitivity, this applies regardless of the mutability of either pointer type
 
 r[abi.compatibility.reference-box]
-The types `&T`, `&mut T`, [`Box<T>`][core::boxed::Box], and [`NonNull<T>`][core::ptr::NonNull], are *abi compatible* with `*const T`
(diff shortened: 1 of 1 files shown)


src/abi.md
```diff
@@ -1,21 +1,209 @@
...
-linking external libraries.
+## ABI Compatibility
+
+r[abi.compatibility]
+
+r[abi.compatibilty.type]
+Two types, `T` and `U`, can be *abi compatible*.
+
+r[abi.compatibility.equivalence]
+Two types `T` and `U` are *abi compatible* if:
```
pnkfelix, line 13: My first reaction upon seeing these three bullets was to say "yes this yields an equivalence relation, but not an interesting one". Skimming down I was reminded that there were going to be other base cases in the inductive definition, which is what makes it interesting.

So: Is there a (concise) way to emphasize that this is just an "if" and not an "if and only if" ?

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+* They are the same type,
+* `U` is *abi compatible* with `T`, or
+* There exists a type `V`, such that `T` is *abi compatible* with `V` an `V` is *abi compatuble* with `U`,
+
+> [!NOTE]
+> These properties ensure that *abi compatibility* is an equivalence relation.
+
+r[abi.compatibility.integer]
+Two [integer types] are *abi compatible* if they have the same size and the same signednes
```
bjorn3, line 22:
```suggestion
Two [integer types] are *abi compatible* if they have the same size and the same signedness
```

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+> With transitivity, this applies regardless of the mutability of either pointer type
+
+```rust
+unsafe fn foo(x: *mut i32){
+   unsafe{x.write(5);}
+}
+
+fn main(){
+    let f: unsafe fn(*mut ()) = unsafe{core::mem::transmute(foo as unsafe fn(_))}; // Type Erase the function
+    let mut val = 0;
```
pnkfelix, line 78: this example is relying on `i32` being the fallback integer type when there is no other alternative being imposed, right?

(E.g., if one had written `5_i8` down below, then that ends up affecting the type inferred for `val`, and yields UB overall since now the write will be out-of-bounds, at least according to miri.)

I am wondering whether it would be better, for purposes of this example, to explicitly assign the `i32` type via `let val: i32 = 0;` and then you avoid discussion of how integer type fallback is handled in this part of the spec.

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+r[abi.compatibility]
+
+r[abi.compatibilty.type]
+Two types, `T` and `U`, can be *abi compatible*.
+
+r[abi.compatibility.equivalence]
+Two types `T` and `U` are *abi compatible* if:
+* They are the same type,
+* `U` is *abi compatible* with `T`, or
+* There exists a type `V`, such that `T` is *abi compatible* with `V` an `V` is *abi compatuble* with `U`,
```
bjorn3, line 16:
```suggestion
* There exists a type `V`, such that `T` is *abi compatible* with `V` an `V` is *abi compatible* with `U`,
```

src/abi.md
chorman0773: Probably, yeah. I guess my brain was just on autopilot writing that test.

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+
+r[abi.compatibilty.zst]
+Two types, `T` and `U`, are *abi compatible* if both have size 0 and alignment 1.
+
+r[abi.compatibility.option]
+If `T` is a type listed in [layout.enum.option](https://doc.rust-lang.org/stable/core/option/index.html#representation), then given `S` is a type with size 0 and alignment 1, `T` is *abi compatible* with the types [`core::option::Option<T>`], [`core::result::Result<T,S>`], and [`core::result::Result<S,T>`].
+
+
+r[abi.compatibility.fn-ptr]
+An [`fn`-ptr type] `T` is compatible with an [`fn`-ptr type] `U` if `T` and `U` have *abi compatible* tags.
```
pnkfelix, line 119: is this supposed to say "*abi compatible*", or just "compatible" ?

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+Two types, `T` and `U`, can be *abi compatible*.
+
+r[abi.compatibility.equivalence]
+Two types `T` and `U` are *abi compatible* if:
+* They are the same type,
+* `U` is *abi compatible* with `T`, or
+* There exists a type `V`, such that `T` is *abi compatible* with `V` an `V` is *abi compatuble* with `U`,
+
+> [!NOTE]
+> These properties ensure that *abi compatibility* is an equivalence relation.
```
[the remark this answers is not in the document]
chorman0773: Probably a good idea. I wouldn't want to put it on the bullets though, maybe add it to the note.

> These properties are respectively called "reflexivity", "symmetry", and "transitivity". They ensure that *abi compatibility* is an equivalence relation.

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+* Each parameter of both signatures, in order, are *abi compatible*, and
+* Either both signatures have C-varargs, or neither signature does.
+
+> [!NOTE]
+> A signature is compatible with itself.
+
+r[abi.compatibility.simd-abi]
+A type has *simd abi requirements* if:
+* It is a type declared with the standard-library repr-attrbute `simd`,
+* It is a aggregate type[^1], which has a type with *simd abi requirements* as a field.
```
ehuss, line 142: Can you use labels instead of numbers for footnotes? It makes it a little easier to avoid collisions, and makes it easier to see semantically what a footnote is referring to.

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+* The [abi tags][abi tag] of both signatures are *abi compatible*,
+* They have the same number of parameters, excluding C-varargs,
+* Each parameter of both signatures, in order, are *abi compatible*, and
+* Either both signatures have C-varargs, or neither signature does.
+
+> [!NOTE]
+> A signature is compatible with itself.
+
+r[abi.compatibility.simd-abi]
+A type has *simd abi requirements* if:
```
ehuss, line 140: Can you add something here that explains what it means to have *simd abi requirements*? It seems to immediately jump in to the rules for what has them, but doesn't seem to really explain what the consequence is or what it is used for. I think a few sentences here could help a lot.

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+> * [`__m128`], [`__m128i`], [`__m128f`], and [`__m128d`] (128-bit vector types): `sse`
+> * [`__m256`], [`__m256i`], [`__m256f`], and [`__m256d`] (256-bit vector types): `avx`
+> * [`__m512`], [`__m512i`], [`__m512f`], and [`__m512d`] (512-bit vector types): `avx512f` and `avx512vl`
+
+r[abi.compatibility.call]
+A call to a function `f` via a function item or function pointer with a given signature `S` is valid if and only if the signature of the definition `f` is *compatible* with the signature `S`, and:
+* The ABI tag of the signature is `extern "Rust"`, or
+* If any parameter type, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each [*salient target features*][target_feature] of that type is either set at both the definition site of the function, and at the call site, or is set at neither site.
+
+The behaviour a call that is not valid is undefined.
```
ehuss, line 161: Can you say where you got these rules from?

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+
+> [!NOTE]
+> When parameter/return types do not exactly match, they are converted as though by calling [`core::mem::transmute`]. The representation and validity requirements of the type in the definition/return site still apply, for example, passing `0` to a function pointer `fn(u32)` that points to a function declared as `fn foo(x: NonZeroU32)` is undefined behaviour.
+
+> [!NOTE]
+> the ABI tag `extern "Rust"` is the default when the `extern` keyword is not used (either to declare the function within an [`extern` block], or as a [function qualifier][extern functions]). Thus it is safe to call most functions that use simd types.
```
ehuss, line 167:
```suggestion
> The ABI tag `extern "Rust"` is the default when the `extern` keyword is not used (either to declare the function within an [`extern` block], or as a [function qualifier][extern functions]). Thus it is safe to call most functions that use simd types.
```

chorman0773:
We need to make mdbook-spec linkify `Result<T,S>`

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
 
-The *`used` attribute* can only be applied to [`static` items]. This [attribute] forces the
-compiler to keep the variable in the output object file (.o, .rlib, etc. excluding final binaries)
-even if the variable is not used, or referenced, by any other item in the crate.
-However, the linker is still free to remove such an item.
+r[abi.used]
+
+```abnf
+MetaItemUsed := "used"
+```
```
ehuss, line 190: Can you stick with the abnf style used in the rest of the reference? I realize it's not the easiest thing to write, but I would prefer to avoid having various different styles.

```suggestion
> **<sup>Attribute Syntax</sup>**\
> _MetaItemUsed_ :\
> &nbsp;&nbsp; `used`
```

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+
```
ehuss, line 191: Can you add a sentence or two here introducing the attribute and briefly explaining what it is for. Generally the template for attributes that we follow is:

The *`NAME` attribute* ...explain what it is for.

Having introductory material is really helpful to give context and immediately understand what the following rules are talking about. It can be helpful to know the purpose up-front, and let the following rules further refine the details.

src/abi.md
```diff
@@ -57,20 +245,97 @@ $ nm -C foo.o
...
+r[abi.symbol-name]
+
+```abnf
+MetaItemNoMangle := "no_mangle"
+MetaItemExportName := "export_name" "=" ([STRING_LITERAL] | [RAW_STRING_LITERAL])
+```
+
+r[abi.symbol-name.names]
+The *`no_mangle` attribute* and the *`export_name` attribute* shall only be applied to a `static` or `fn` item. The *`export_name` attribute* shall not be applied to an item declared within an [`extern` block].
```
ehuss, line 261: Can't no_mangle be applied to all other items?

src/abi.md
```diff
@@ -57,20 +245,97 @@ $ nm -C foo.o
...
+#[export_name = "bar"]
+extern "C" fn baz(x: i32) -> i32 {
+    x + 2
+}
+```
+
+```rust,compile_fail
+extern "C" {
+    #[export_name = "foo"]
+    fn __foo(x: i32) -> i32;
```
ehuss, line 278: It can be helpful for error examples to highlight where the error is.

```suggestion
    fn __foo(x: i32) -> i32;  // error: not a free function, impl method, or static
```

src/abi.md
```diff
@@ -79,22 +344,46 @@ The *`link_section` attribute* specifies the section of the object file that a
 pub static VAR1: u32 = 1;
 ```
 
-## The `export_name` attribute
+r[abi.link_section.def]
+An item with the *`link_section` attribute* is placed in the specified section when linking. The section specified shall not violate the constraints on section names on the target, and shall not be invalid for the item type, no diagnostic is required.
```
ehuss, line 351: Similar to our conversation before about ndr, I think it would be best to avoid this for now until we have a decision on it.

src/abi.md
```diff
@@ -79,22 +344,46 @@ The *`link_section` attribute* specifies the section of the object file that a
...
 
-The *`export_name` attribute* specifies the name of the symbol that will be
-exported on a [function] or [static]. It uses the [_MetaNameValueStr_] syntax
-to specify the symbol name.
+> [!NOTE]
+> A section name may be invalid if it violates the requirements for the item type, for example, an `fn` item must be placed in an executable section, and a mutable static item (`static mut` or one containing an `UnsafeCell`) must be placed in a writable section.
```
ehuss, line 354:
```suggestion
> A section name may be invalid if it violates the requirements for the item type. For example, an `fn` item must be placed in an executable section, and a mutable static item (`static mut` or one containing an `UnsafeCell`) must be placed in a writable section.
```

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
-This section documents features that affect the ABI of the compiled output of
-a crate.
+r[abi]
 
-See *[extern functions]* for information on specifying the ABI for exporting
-functions. See *[external blocks]* for information on specifying the ABI for
-linking external libraries.
+## ABI Compatibility
+
+r[abi.compatibility]
```
ehuss, line 7: Can you add a few sentences explaining what *abi compatibility* is? Such as the consequence of what it means for types to be abi compatible? In general, I felt a little uncertain what I was reading, since it immediately jumps into a list of rules without giving context or explaining what this section is about.

ehuss (commented):
Thanks @chorman0773! I really appreciate this!

src/abi.md
```diff
@@ -1,21 +1,209 @@
...
+fn main(){
+    let f: fn(u32)->char = unsafe{core::mem::transmute(foo as fn(_)->_)};
+    let x = b'A' as u32; // ascii character indecies are the same as Unicode character indecies
+    let y = f(x);
+    assert_eq!(y, 'A');
+}
+```
+
+r[abi.compatibility.pointer]
+Two [pointer types], `*mut T` and `*const U`, are *abi compatible* if the *metadata type*s of `T` and `U` are the same type.
```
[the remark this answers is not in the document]
chorman0773: Will be defined as part of layout.  The term is used in the docs.rs page.

src/abi.md
chorman0773: From: https://doc.rust-lang.org/core/primitive.fn.html#abi-compatibility

ehuss:
> We need to make mdbook-spec linkify `Result<T,S>`

For now, you'll need to use a fully qualified path like ```[`Result<T, S>`](std::result::Result)```. It's not really possible to support prelude items, since it needs some kind of token (like `std::`) to know if it is a standard library link.

chorman0773:
That's current what I'm using - the issue is on the `<T,S>`. It seems to handle 1 generic parameter fine, but not 2.

ehuss:
Ah, I see! Posted https://github.com/rust-lang/reference/pull/1549 with a fix.

ehuss:
A concern that I think we should consider is that this seems to duplicate content from the core documentation. I think this is an important question about how we want to handle that. There are a few options:

- Move everything to the reference, and remove from the core docs and link to the reference.
- Keep everything in the core docs, and *don't* add it to the reference.
- Duplicate in both places.

I think if it is duplicated, it will get out of sync, which I think will contribute to confusion, and cause more work.

I lean towards moving it to the reference, but there are some considerations of it being very useful to be in the core docs.

traviscross:
cc @RalfJung

src/abi.md
chorman0773: It is stated in the following note.

workingjubilee:
> I lean towards moving it to the reference, but there are some considerations of it being very useful to be in the core docs.

I think that where it is relevant, some things should be documented in the standard library even at the cost of duplication and even at the cost of desync.

Tersely, and then immediately (ahem) reference the Reference.

workingjubilee:
This probably could afford splitting the attributes and argument/return type equivalence into different files?

ehuss:
cc https://github.com/rust-lang/rust/pull/130653

traviscross:
cc @rust-lang/opsem

ehuss:
Can you update to also include https://github.com/rust-lang/rust/pull/128784 (assuming that isn't already covered, and should go here and not elsewhere)?

ehuss:
Does this also need to be updated for https://github.com/rust-lang/rust/issues/110503?

chorman0773:
https://github.com/rust-lang/rust/pull/128784 seems more like something for the `items/extern-block.md`, as the list of supported ABI strings are documented in that chapter.

> Does this also need to be updated for https://github.com/rust-lang/rust/issues/110503?

Hmm... I can't tell whether the FCP in https://github.com/rust-lang/rust/pull/130628 provides a just the Layout guarantee for all types or an ABI guarantee, so I'd like for T-lang to clarify that.

@rustbot label: +I-lang-nominated

Question for T-lang: In rust#130628, T-lang FCP-stabilized a layout guarantee for any `enum` type that matches the specific form documented there in https://github.com/rust-lang/rust/pull/130628#issuecomment-2402761599. Does this layout guarantee also carry an ABI compatibility guarantee that should be documented here?

traviscross:

Good catch that we didn't do a great job of being clear about that in writing up the FCP.  We did mean, though, for this to be an ABI guarantee (e.g., we were building on @RalfJung's comment [here](https://github.com/rust-lang/rust/pull/130628#issuecomment-2365492463)).

We're documenting this also in:

- https://github.com/rust-lang/rust/pull/132003

chorman0773:
The new section points to Ralf's UCG chapter temporarily, but the link can be removed and pointed to the id defined by #1654.

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
 a crate.
 
 See *[extern functions]* for information on specifying the ABI for exporting
-functions. See *[external blocks]* for information on specifying the ABI for
+functions. See *[`extern` block]s* for information on specifying the ABI for
 linking external libraries.
 
+## ABI compatibility
+
+r[abi.compatibility]
```
RalfJung, line 14: There's no paragraph after this marker... is this the normal syntax?

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+
+r[abi.compatibility.intro]
+Function calls pass parameters and return values between the caller and the callee function. This requires the caller and callee to agree on an ABI for those parameters and return values. This is typically only guaranteed when the same type is used in both the call site and the definition site of the callee, but certain other types may be *abi compatible*. This can appear when transmuting a [function pointer] or using an [`extern` block] to call a function. When the parameters or return types differ between the call site and def site, assuming they are *abi compatible*, the parameters or return types are transmuted to the type at the def site or call site respectively.
```
RalfJung, line 17:
```suggestion
Function calls pass parameters and return values between the caller and the callee function. This requires the caller and callee to agree on an ABI for those parameters and return values. This is typically only guaranteed when the same type is used in both the call site and the definition site of the callee. However, differences between caller and callee types can appear when transmuting a [function pointer] or using an [`extern` block] to call a function. This is permitted if the caller and callee types are *abi compatible* (otherwise, it is undefined behavior). If the types are abi compatible, the parameters are transmuted to the callee type as part of the call and the return value is transmuted to the caller type upon return.
```

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+
+> [!NOTE]
+> This can include calls to functions defined outside of rust, or built using a different Rust compiler version.
```
RalfJung, line 20:
```suggestion
> This can include calls to functions defined outside of Rust, or built using a different Rust compiler version.
```

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+> Additional guarantees will apply in this case for "FFI Safe" types, which match up with the platform C ABI in well-defined ways.
+> These are not fully documented here currently.
```
RalfJung, line 22: Shouldn't this say that additional *requirements* apply? The document right now only describes the rules for Rust-to-Rust calls, as far as I can see.

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+> [!WARNING]
+> Two types that are ABI Compatible may not allow the same set of values (e.g. [`*const T`] and [`core::ptr::NonNull<T>`]).
+> If an invalid value is passed as a parameter or returned from a function, the result is immediate undefined behaviour, even if the parameter or return value is never used.
+> For example, passing a null pointer to a function that accepts a `NonNull<i32>` parameter via `fn(*const i32)` function pointer caused undefined behaviour.
+
+r[abi.compatibility.equivalence]
+Two types `T` and `U` are *abi compatible* if:
+* They are the same type,
+* `U` is *abi compatible* with `T`, or
+* There exists a type `V`, such that `T` is *abi compatible* with `V` an `V` is *abi compatible* with `U`,
```
RalfJung, line 33: IMO this is better stated after all the other clauses, since really what we are doing is we are generating a relation by having some "seeds" and then we are upclosing it to become an equivalence relation. It is quite confusing to read this before we added any of the elements to the relation that make it interesting.

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+    let f: unsafe fn(*mut ()) = unsafe{core::mem::transmute(foo as fn(_))}; // Type Erase the function
+    let mut val = 0;
+    let ptr = core::ptr::addr_of_mut!(val).cast::<()>(); // Get Opaque Userdata from somewhere
+    unsafe{f(ptr);}
+    assert_eq!(val, 5);
+}
+```
+
+r[abi.compatibility.core]
+The types [`core::mem::MaybeUninit<T>`], [`core::cell::UnsafeCell<T>`], and [`core::num::NonZero<T>`], are *abi compatible* with `T`.
```
RalfJung, line 127: This is odd... why do we list these types and not others?

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+r[abi.compatibility.discriminant]
+If `T` is an a type listed in [layout.repr.rust.option.elision], and `U` is the type of the *elision candidate field*, then `T` is layout compatible with `U`.
+
+> [!NOTE]
+> `Option<U>`, `Result<U,Z>`, or `Result<Z,U>` are such types, when `U` are *elision candidate type*s, and `Z` is a 1-ZST type.
+>
+> Due to transitivity, two such types are *abi compatible* with each other if their *elision candidate field*s are *abi comaptible*
+
+r[abi.compatibility.fn-ptr]
+An [function pointer] type `T` is *abi compatible* with an [function pointer] type `U` if `T` and `U` have *abi compatible* tags.
```
RalfJung, line 144:
```suggestion
A [function pointer] type `T` is *abi compatible* with an [function pointer] type `U` if `T` and `U` have *abi compatible* tags.
```

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+Two function signatures are compatible if:
+* The [abi tags][abi tag] of both signatures are *abi compatible*,
+* They have the same number of parameters, excluding C-varargs,
+* Each parameter of both signatures, in order, are *abi compatible*, and
+* Either both signatures have C-varargs, or neither signature does.
+
+> [!NOTE]
+> A signature is compatible with itself.
+
+r[abi.compatibility.simd-abi]
```
RalfJung, line 164: Given that we consider this a bug and are very close to fixing the bug to at least emit warnings, I don't think we should list this here.

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+> * [`__m128`], [`__m128i`], [`__m128f`], and [`__m128d`] (128-bit vector types): `sse`
+> * [`__m256`], [`__m256i`], [`__m256f`], and [`__m256d`] (256-bit vector types): `avx`
+> * [`__m512`], [`__m512i`], [`__m512f`], and [`__m512d`] (512-bit vector types): `avx512f` and `avx512vl`
+
+r[abi.compatibility.call]
+A call to a function `f` via a function item or function pointer with a given signature `S` is valid if and only if the signature of the definition `f` is *compatible* with the signature `S`, and:
+* The ABI tag of the signature is `extern "Rust"`, or
+* If any parameter type, the return type, or the type of any argument passed via C-varargs has *simd abi requirements*, each [*salient target features*][target_feature] of that type is either set at both the definition site of the function, and at the call site, or is set at neither site.
+
+The behavior of a call that is not valid is undefined.
```
RalfJung, line 194: Should this say that a valid call transmutes the arguments / return value?

src/abi.md
```diff
@@ -1,21 +1,246 @@
...
+The types [`core::mem::MaybeUninit<T>`], [`core::cell::UnsafeCell<T>`], and [`core::num::NonZero<T>`], are *abi compatible* with `T`.
+
+r[abi.compatibility.transparent]
+A [`struct`] declared with the `transparent` representation is *abi compatible* with its field that does not have size 0 and alignment 1, if such a field exists.
+
+r[abi.compatibility.zst]
+Two types, `T` and `U`, are *abi compatible* if both have size 0 and alignment 1.
+
+r[abi.compatibility.discriminant]
+If `T` is an a type listed in [layout.repr.rust.option.elision], and `U` is the type of the *elision candidate field*, then `T` is layout compatible with `U`.
```
RalfJung, line 136: The referenced section here doesn't exist yet, does it?

RalfJung: Oh this is just a normal markdown link for now, I see.

src/abi.md
RalfJung: This clause no longer exists in the library docs, so it should be removed here as well.

chorman0773: Yep, was waiting for the stdlib to fix to land. Will fix this tomorrow.

src/abi.md
chorman0773: Yes, it's an identifier for the whole section.

src/abi.md
chorman0773: These types are magic lang items. `MaybeUninit<T>` is a transparent union (which isn't otherwise a thing that can exist), `UnsafeCell<T>` doesn't fully qualify as normal `repr(transparent)` due to excluding niches (including guaranteed ones that would normally be inherited by `repr(transparent)`, and `NonZero<T>` has GDE.

src/abi.md
chorman0773: Getting rid of the hard dependency on #1654.
If #1654 lands first, then I'll remove the link. If this lands before #1654 I'll make #1645 remove the link.

src/abi.md
chorman0773: No, because the note is talking about the fact that "FFI Safe" types have a defined mapping to the platform C abi. 
You can technically pass a non-"FFI Safe" type to an FFI call, there's just no guarantee you can match it on the other side.

RalfJung: Oh so this is about guarantees of the form "This Rust type has an ABI matching this C type". Would be good to clarify the comment here, I currently find it a bit confusing -- everything else here is about guarantees of the form "these two Rust types have the same ABI", after all.

src/abi.md
```diff
@@ -1,21 +1,201 @@
...
+
+> [!NOTE]
+> This can include calls to functions defined outside of Rust, or built using a different Rust compiler version.
+> Additional guarantees will apply in this case for "FFI Safe" types, which match up with the platform C ABI in well-defined ways.
+> These are not fully documented here currently.
+
+> [!WARNING]
+> Two types that are ABI Compatible may not allow the same set of values (e.g. [`*const T`] and [`core::ptr::NonNull<T>`]).
+> If an invalid value is passed as a parameter or returned from a function, the result is immediate undefined behaviour, even if the parameter or return value is never used.
+> For example, passing a null pointer to a function that accepts a `NonNull<i32>` parameter via `fn(*const i32)` function pointer caused undefined behaviour.
```
bluebear94, line 27:
```suggestion
> For example, passing a null pointer to a function that accepts a `NonNull<i32>` parameter via `fn(*const i32)` function pointer causes undefined behavior.
```

PR #1556, opened 2024-08-02
add the `const` operand to docs for inline assembly

folkertdev:
in order to stabilize https://github.com/rust-lang/rust/issues/93332.

cc @Amanieu


ehuss (approved):
Looks like https://github.com/rust-lang/rust/pull/128570 has merged, thanks!

PR #1559, opened 2024-08-04
Say that `pub(in path)` can't depend on `use` statements

mattheww:
Arbitrary path expressions aren't allowed here (in particular, the path can't rely on another `use` statement).

### DIFF
--- a/src/visibility-and-privacy.md
+++ b/src/visibility-and-privacy.md
@@ -147,8 +147,9 @@ expressions, types, etc.
 In addition to public and private, Rust allows users to declare an item as
 visible only within a given scope. The rules for `pub` restrictions are as
 follows:
-- `pub(in path)` makes an item visible within the provided `path`. `path` must
-be an ancestor module of the item whose visibility is being declared.
+- `pub(in path)` makes an item visible within the provided `path`.
+  `path` must be a simple path which resolves to an ancestor module of the item whose visibility is being declared.
+  Each identifier in `path` must refer directly to a module (not to a name introduced by a `use` statement).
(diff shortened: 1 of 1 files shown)


mattheww:
Sorry, saying "canonical path" isn't right because things like `pub (in super::m)` are allowed.

So instead, just say explicitly that using names from `use` statements isn't allowed.

traviscross:
@petrochenkov: We talked about this in the lang-docs call today, and we were uncertain whether or not this behavior is intentional.  What do you think?

This is about cases like this:

```rust
mod m {
    use crate::m as n;
    pub(in n) fn f() {}
    //~^ ERROR relative paths are not supported in visibilities in 2018 edition or later
}
```

mattheww:
I think this is a better example, because (as the error message above says) relative paths aren't allowed for `pub(in path)` these days anyway.

```rust
mod m {
    use crate::m as n;
    pub(in self::n) fn f() {}
    //           ^ could not find `n` in `self`
}
```

petrochenkov:
I explained the situation in https://github.com/rust-lang/rust/issues/109657#issuecomment-1485357968.

Basically supporting arbitrary paths in `pub` is both nontrivial and has very little value, so it's easier to say that only parent paths without imports are supported.

I suspect that in practice imports may be supported in some cases, if they are ready early enough, but it's not guaranteed.

traviscross (approved):

traviscross:
Thanks @petrochenkov for that background.  Wild example there; thanks for sharing that.

Given that, the change seems fine.  Thanks @mattheww.

PR #1560, opened 2024-08-06
Add identifiers to attributes.md and its subchapters

chorman0773:
This adds identifiers to the attributes.md file and its subchapters, being:
* attributes/codegen.md
* attributes/debugging.md
* attributes/derive.md
* attributes/diagnostics.md
* attributes/limits.md
* attributes/testing.md
* attributes/type_system.md

### DIFF
--- a/src/attributes/debugger.md
+++ b/src/attributes/debugger.md
@@ -1,20 +1,32 @@
 # Debugger attributes
 
+r[attributes.debugger]
+
 The following [attributes] are used for enhancing the debugging experience when using third-party debuggers like GDB or WinDbg.
 
 ## The `debugger_visualizer` attribute
 
+r[attributes.debugger.debugger_visualizer]
+
+r[attributes.debugger.debugger_visualizer.general]
 The *`debugger_visualizer` attribute* can be used to embed a debugger visualizer file into the debug information.
 This enables an improved debugger experience for displaying values in the debugger.
+
+r[attributes.debugger.debugger_visualizer.restriction]
 It uses the [_MetaListNameValueStr_] syntax to specify its inputs, and must be specified as a crate attribute.
 
 ### Using `debugger_visualizer` with Natvis
 
+r[attributes.debugger.debugger_visualizer.natvis]
+
+r[attributes.debugger.debugger_visualizer.natvis-general]
 Natvis is an XML-based framework for Microsoft debuggers (such as Visual Studio and WinDbg) that uses declarative rules to customize the display of types.
 For detailed information on the Natvis format, refer to Microsoft's [Natvis documentation].
 
+r[attributes.debugger.debugger_visualizer.natvis-restrictions]
 This attribute only supports embedding Natvis files on `-windows-msvc` targets.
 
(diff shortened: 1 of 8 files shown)


src/attributes/debugger.md
```diff
@@ -1,20 +1,32 @@
...
 This enables an improved debugger experience for displaying values in the debugger.
+
+r[attributes.debugger.debugger_visualizer.restriction]
 It uses the [_MetaListNameValueStr_] syntax to specify its inputs, and must be specified as a crate attribute.
 
 ### Using `debugger_visualizer` with Natvis
 
+r[attributes.debugger.debugger_visualizer.natvis]
+
+r[attributes.debugger.debugger_visualizer.natvis-general]
```
ehuss, line 22: Any particular reason to use `-` and not `.` for `nativis-general`?

src/attributes/derive.md
```diff
@@ -1,7 +1,14 @@
 # Derive
 
+r[attributes.derive]
+
+
+r[attributes.derived.general]
```
ehuss, line 6:
```suggestion
r[attributes.derive.general]
```

src/attributes/diagnostics.md
```diff
@@ -390,6 +432,7 @@ impl Trait for i32 {
 5i32.use_me();
 ```
 
+r[attributes.diagnostics.must_use.impl-function]
```
ehuss, line 435: To distinguish from other non-trait impl, maybe?

```suggestion
r[attributes.diagnostics.must_use.trait-impl-function]
```

src/attributes/diagnostics.md
```diff
@@ -424,36 +467,59 @@ When used on a function in a trait implementation, the attribute does nothing.
 
 ## The `diagnostic` tool attribute namespace
 
+r[attributes.diagnostics.namespace]
```
ehuss, line 470: Hm, I'm on the fence here, since the namespace is singular `diagnostic`. I realize this is in the "diagnostics" section, and being inconsistent with the rest of the section could be more trouble. I'm just a little concerned about seeing `diagnostics.namespace` could confuse a reader, since it is a subtle distinction. WDYT?

src/attributes/limits.md
```diff
@@ -1,12 +1,20 @@
 # Limits
 
+r[attributes.limits]
+
 The following [attributes] affect compile-time limits.
 
 ## The `recursion_limit` attribute
 
+r[attributes.limits.recursion_limits]
```
ehuss, line 9: Here and elsewhere.

```suggestion
r[attributes.limits.recursion_limit]
```

src/attributes/diagnostics.md
chorman0773: I'm thinking the section should be changed to `diagnostic` because I've been otherwise favouring the singular form except for lists and when the construct is plural in the language.

traviscross:
@rustbot labels +S-waiting-on-author -S-waiting-on-review

The CI is failing on this one.

ehuss (approved):
Thanks!

PR #1566, opened 2024-08-10
const_eval: update for const-fn float stabilization

RalfJung:
Accompanying reference update for https://github.com/rust-lang/rust/pull/128596.

The float semantics RFC decided that this is *not* a guarantee we actually provide (and I am surprised to see that this was ever documented here, since we have [other plans](https://github.com/rust-lang/rust/issues/124625) that will make this even less true):

> Calling a const function at compile-time
will always yield the same result as calling it at runtime, even when called
multiple times.

Therefore, I removed this paragraph. I am not sure what its purpose was.

Also fixes https://github.com/rust-lang/reference/issues/1581.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -63,7 +63,9 @@ A _const context_ is one of the following:
 
 Const contexts that are used as parts of types (array type and repeat length
 expressions as well as const generic arguments) can only make restricted use of
-surrounding generic type and lifetime parameters.
+surrounding generic parameters: such an expression must either be a single bare
+const generic parameter, or an arbitrary expression not making use of any
+generics.
 
 ## Const Functions
 
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -61,39 +61,21 @@ A _const context_ is one of the following:
 * A [const generic argument]
 * A [const block]
 
+Const contexts that are used as parts of types (array type and repeat length
+expressions as well as const generic arguments) can only make restricted use of
+surrounding generic type and lifetime parameters.
+
 ## Const Functions
 
 A _const fn_ is a function that one is permitted to call from a const context. Declaring a function
 `const` has no effect on any existing uses, it only restricts the types that arguments and the
-return type may use, as well as prevent various expressions from being used within it. You can freely
-do anything with a const function that you can do with a regular function.
+return type may use, and restricts the function body to constant expressions.
 
 When called from a const context, the function is interpreted by the
 compiler at compile time. The interpretation happens in the
 environment of the compilation target and not the host. So `usize` is
 `32` bits if you are compiling against a `32` bit system, irrelevant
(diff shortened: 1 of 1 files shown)


RalfJung:
Seems like this guarantee was added a long time ago, in https://github.com/rust-lang/reference/pull/440. Cc @oli-obk

oli-obk:
> I am not sure what its purpose was.

the purpose was to be able to land the reference change and doing any work, as there was resistance to breaking that "guarantee" that we accidentally had

src/const_eval.md
```diff
@@ -61,40 +61,22 @@ A _const context_ is one of the following:
 * A [const generic argument]
 * A [const block]
 
+Const contexts that are used as parts of types (array type and repeat length
+expressions as well as const generic arguments) can only make restricted use of
+surrounding generic type and lifetime parameters.
```
ehuss, line 66: Do you happen to know what the restrictions are?

Also, just to verify, there are no restrictions for const generic parameters now?

RalfJung: I don't know, I just copied this text to a place where I think it makes more sense.
Cc @BoxyUwU

joshtriplett:
> Accompanying reference update for [rust-lang/rust#128596](https://github.com/rust-lang/rust/pull/128596).
> 
> The float semantics RFC decided that this is _not_ a guarantee we actually provide (and I am surprised to see that this was ever documented here, since we have [other plans](https://github.com/rust-lang/rust/issues/124625) that will make this even less true):
> 
> > Calling a const function at compile-time
> > will always yield the same result as calling it at runtime, even when called
> > multiple times.
> 
> Therefore, I removed this paragraph. I am not sure what its purpose was.

Would it be feasible, rather than deleting this paragraph, to document exceptions to it (e.g. NaN bit patterns, calls to some floating-point functions)? It still seems to *generally* be true with the current set of things we've accepted.

(This is not an objection, just a question.)

RalfJung:
Yes, it is feasible. Currently, the NaN bit patterns returned by some float operations are the only case. However, I'm not very fond of the idea that we'll have to update the reference each time we add a new source of potential non-determinism or other kinds of difference between compile and runtime behavior...

chorman0773:
Also sounds like a stability nightmare.

traviscross (approved):

src/const_eval.md
BoxyUwU: type system const expressions can either be arbitrary expressions not making using of any generics, or a single bare const parameter, e.g. `{ N }`

RalfJung: Thanks! I've added this to the PR.

It's not related to "const fn float" at all, but I don't want to make this a separate PR as those would conflict...

ehuss (approved):
Thanks!

PR #1567, opened 2024-08-10
operator expressions: add &raw

RalfJung:
Accompanying PR for https://github.com/rust-lang/rust/pull/127679.
Fixes https://github.com/rust-lang/rust/issues/64490.

I reworded things to avoid the "address-of" terminology, since a pointer is not just an address.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -146,7 +146,7 @@ A *value expression* is an expression that represents an actual value.
 The following contexts are *place expression* contexts:
 
 * The left operand of a [compound assignment] expression.
-* The operand of a unary [borrow], [address-of][addr-of] or [dereference][deref] operator.
+* The operand of a unary [borrow], [raw borrow][raw-borrow] or [dereference][deref] operator.
 * The operand of a field expression.
 * The indexed operand of an array indexing expression.
 * The operand of any [implicit borrow].
@@ -276,7 +276,7 @@ They are never allowed before:
 
 [assign]:               expressions/operator-expr.md#assignment-expressions
 [borrow]:               expressions/operator-expr.md#borrow-operators
-[addr-of]:              expressions/operator-expr.md#raw-address-of-operators
+[raw-borrow]:           expressions/operator-expr.md#raw-borrow-operators
 [comparison]:           expressions/operator-expr.md#comparison-operators
 [compound assignment]:  expressions/operator-expr.md#compound-assignment-expressions
 [deref]:                expressions/operator-expr.md#the-dereference-operator
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -146,7 +146,7 @@ A *value expression* is an expression that represents an actual value.
 The following contexts are *place expression* contexts:
 
 * The left operand of a [compound assignment] expression.
-* The operand of a unary [borrow], [raw borrow][raw-borrow] or [dereference][deref] operator.
+* The operand of a unary [borrow], [raw borrow] or [dereference][deref] operator.
 * The operand of a field expression.
 * The indexed operand of an array indexing expression.
 * The operand of any [implicit borrow].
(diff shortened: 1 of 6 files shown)


src/expressions/operator-expr.md
```diff
@@ -101,14 +101,14 @@ struct Packed {
 
 let packed = Packed { f1: 1, f2: 2 };
 // `&packed.f2` would create an unaligned reference, and thus be Undefined Behavior!
-let raw_f2 = ptr::addr_of!(packed.f2);
+let raw_f2 = &raw const packed.f2;
```
RalfJung, line 104: I assume this will fail in CI since the operator is not stable yet. Not sure how you usually stage this.

src/expressions/operator-expr.md
```diff
@@ -41,7 +41,9 @@ The following things are considered to be overflow:
 > **<sup>Syntax</sup>**\
 > _BorrowExpression_ :\
 > &nbsp;&nbsp; &nbsp;&nbsp; (`&`|`&&`) [_Expression_]\
-> &nbsp;&nbsp; | (`&`|`&&`) `mut` [_Expression_]
+> &nbsp;&nbsp; | (`&`|`&&`) `mut` [_Expression_]\
+> &nbsp;&nbsp; | (`&`|`&&`) `raw` `const [_Expression_]\
```
ehuss, line 45: Looks to be missing a back tick.

```suggestion
> &nbsp;&nbsp; | (`&`|`&&`) `raw` `const` [_Expression_]\
```

src/expressions/operator-expr.md
```diff
@@ -79,20 +81,18 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
-### Raw address-of operators
+### Raw pointer operators
 
-Related to the borrow operators are the *raw address-of operators*, which do not have first-class syntax, but are exposed via the macros [`ptr::addr_of!(expr)`][addr_of] and [`ptr::addr_of_mut!(expr)`][addr_of_mut].
+Related to the borrow operators are the *raw pointer operators*, `&raw const` and `&raw mut`.
 The expression `expr` is evaluated in place expression context.
```
ehuss, line 87: What `expr` is hasn't been specified (as the old text it was referring to has been removed).

```suggestion
The operand expression of the raw pointer operator is evaluated in place expression context.
```

ehuss (commented):
Thanks!

Should the last sentence of the section https://github.com/rust-lang/reference/blob/master/src/types/pointer.md#raw-pointers-const-and-mut be updated to use the new syntax instead of addr_of?

And in general, should all other uses of addr_of in the reference be changed to use the new syntax? My understanding is that is now the preferred style, is that correct?

src/expressions/operator-expr.md
ehuss: Yea, CI will just be red until the stabilization PR hits nightly, after which point we can merge this.

src/expressions/operator-expr.md
```diff
@@ -79,20 +81,18 @@ let a = && && mut 10;
 let a = & & & & mut 10;
 ```
 
-### Raw address-of operators
+### Raw pointer operators
```
RalfJung, line 84: I am not sure what is the best thing to call it. Another possible name here might be "Raw borrow expression/operator", to make it clear that this is very similar to reference-creating operations. After all this is in a section titled "borrow operators".

traviscross: "Raw borrow" sounds right to me.  A "raw borrow" creates a "raw reference", which, yes, is also a "raw pointer".

RalfJung: All right, I have switched the PR to use "raw borrow operator".

src/expressions.md
```diff
@@ -276,7 +276,7 @@ They are never allowed before:
 
 [assign]:               expressions/operator-expr.md#assignment-expressions
 [borrow]:               expressions/operator-expr.md#borrow-operators
-[addr-of]:              expressions/operator-expr.md#raw-address-of-operators
+[raw-borrow]:           expressions/operator-expr.md#raw-borrow-operators
```
traviscross, line 279: If you would, just make the anchor "raw borrow" and correspondingly simplify the use of it above, and move this line to its alphabetized spot (even though it muddies up the diff).

traviscross (approved):
Otherwise this looks good to me.  We'll merge this once the stabilization PR is merged.

PR #1571, opened 2024-08-13
Add spec identifier syntax to destructors.md

chorman0773:
This adds spec identifier syntax for destructors.md

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -1,8 +1,6 @@
 # Destructors
-r[destructors]
 
-
-r[destructors.general]
+r[destructors.intro]
 When an [initialized]&#32;[variable] or [temporary] goes out of
 [scope](#drop-scopes), its *destructor* is run, or it is *dropped*. [Assignment]
 also runs the destructor of its left-hand operand, if it's initialized. If a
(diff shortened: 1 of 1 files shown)


src/destructors.md
```diff
@@ -57,48 +62,85 @@ core::mem::forget(partial_move.1);
 
 ## Drop scopes
 
+r[destructor.scope]
```
traviscross, line 65:
```suggestion
r[destructors.scope]
```

The document goes back and forth between the plural and singular.  Seems these should all be plural.

src/destructors.md
```diff
@@ -57,48 +62,85 @@ core::mem::forget(partial_move.1);
...
+
```
traviscross, line 66:
```suggestion
```

Let's remove the double line breaks.

src/destructors.md
```diff
@@ -1,11 +1,15 @@
 # Destructors
+r[destructors]
```
traviscross, line 2: In another PR, where the `r[..]` had the same name as the primary header, it was elided.  Do we want to elide these or include them?

src/destructors.md
```diff
@@ -346,6 +413,8 @@ let x = (&temp()).use_temp();  // ERROR
 
 ## Not running destructors
 
+r[destructors.forget]
+
```
traviscross, line 417: Let's be consistent on whether or not there's a line break between the `r[..]` tag and the content that follows when it's ahead of a paragraph like this.

traviscross (changes requested):

src/destructors.md
```diff
@@ -1,11 +1,15 @@
 # Destructors
+r[destructors]
 
+
+r[destructors.general]
```
traviscross, line 5:
```suggestion
r[destructors.intro]
```

In the other chapter, we went with `.intro` for cases like these.  The particular name probably doesn't matter, but we should be consistent.  `intro` is shorter, so maybe let's do that.

src/destructors.md
```diff
@@ -57,48 +62,85 @@ core::mem::forget(partial_move.1);
 
 ## Drop scopes
 
+r[destructor.scope]
+
+
```
traviscross, line 67:
```suggestion

```

Let's remove the double line breaks here and elsewhere.

traviscross:
We talked about this in our lang-docs call.  Other than as mentioned here, these changes seem OK to us.

src/destructors.md
chorman0773: The tag here is supposed to be attached to the heading. When the content in the section is small enough to not require more than one of its own tags, I've just omitted the tags on the paragraphs themselves.

src/destructors.md
```diff
@@ -256,9 +311,13 @@ always has the type `&'static Option<_>`, as it contains nothing disallowed).
 
 ### Temporary lifetime extension
 
+r[destructors.scope.lifetime-extension]
+
 > **Note**: The exact rules for temporary lifetime extension are subject to
 > change. This is describing the current behavior only.
 
+r[destructors.scope.lifetime-extension.let]
+
```
traviscross, line 320: Is there a reason for the line break here?

src/destructors.md
traviscross: Makes sense.  Seems reasonable.

src/destructors.md
chorman0773: I'd like to have the rendered syntax there, but link-check says no.

traviscross (approved):

PR #1575, opened 2024-08-15
don't capitalize Undefined Behavior

RalfJung:
Usually we use "undefined behavior", but in a few cases it's "Undefined Behavior" instead. Let's make this consistent at least across the reference.


ehuss (approved):
Thanks!

PR #1576, opened 2024-08-16
Fix wildcard pattern documentation

### LINKED ISSUE #848: Wildcard pattern in function parameters.
I noticed that the section on [wildcard patterns](https://doc.rust-lang.org/reference/patterns.html#wildcard-pattern) says that:

> Unlike identifier patterns, it does not copy, move or borrow the value it matches.

However for a function parameter, any arguments that match the wildcard are still consumed by the function:

    fn main() {
        let s = String::from("Hello");
        f(s);

        // Won't compile since `s` was moved. 
        println!("{}", s);
    }
    
    fn f(_: String) {}

Should the description be changed to mention this behavior?

Tomer-Eliahu:
The current [wildcard pattern documentation](https://doc.rust-lang.org/reference/patterns.html#wildcard-pattern) has the following as an example of the wildcard pattern:

```rust
// ignore a function/closure param
let real_part = |a: f64, _: f64| { a };
```
The problem is that this is not a correct example of the wildcard pattern. The wildcard pattern does not move or copy the value it matches, whereas '_' in closure (or function) parameters _does_ move (or copy) the value. One notable example is the toilet closure:
```
let s = String::from("Hello");
let toilet = |_| ();
toilet(s);
println!("{}", s); //ERROR--- s was moved when we called the closure
```

This pull request fixes and clarifies the documentation. This also closes #848.

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -300,35 +300,53 @@ It is used to ignore values when they don't matter.
 Inside other patterns it matches a single data field (as opposed to the `..` which matches the remaining fields).
 Unlike identifier patterns, it does not copy, move or borrow the value it matches.
 
+The wildcard pattern is always irrefutable.
+
 Examples:
 
 ```rust
-# let x = 20;
-let (a, _) = (10, x);   // the x is always matched by _
+let x = String::from("Hello");
+let _ = x;
+
+let y = (10, String::from("World"));
+let (a, _) = y;
+
+println!("{} {}", x, y.1); //x and y.1 were NOT moved
 # assert_eq!(a, 10);
 
-// ignore a function/closure param
-let real_part = |a: f64, _: f64| { a };
-
+struct Person {
+    name: String,
+    age: u8,
+}
+let person = Person {
+    name: String::from("John"),
+    age: 23,
+};
 // ignore a field from a struct
-# struct RGBA {
-#    r: f32,
-#    g: f32,
-#    b: f32,
-#    a: f32,
-# }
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -300,35 +300,54 @@ It is used to ignore values when they don't matter.
...
 
+let x = Some(String::from("Value"));
 // accept any Some, with any value
-# let x = Some(10);
 if let Some(_) = x {}
+println!("{}", x.unwrap()); //the String in x was NOT moved
 ```
 
-The wildcard pattern is always irrefutable.
+Note, however, that `_` in function or closure parameters is _not_ a part of the wildcard pattern and _does_ move the value it matches:
```
traviscross, line 337: This sentence is incorrect.  The `_` is still a wildcard pattern.

traviscross:
We talked about this in the lang-docs call today.  Our feeling overall is there's an interesting question hiding in here about how to describe that the value is moved as part of the call and that the pattern doesn't matter for this.

But at the same time, this PR seems to be build around a mistake -- the idea that the `_` is not a wildcard pattern in these circumstances -- so we're going to close it.

Tomer-Eliahu:
Fair enough. 
I'll be happy to revise the pull request to point out that while the `_` is still a wildcard pattern in function/closure calls, that the value is still moved. 
For the explanation as to why that is, I think that reproducing the quote from the [Function body](https://doc.rust-lang.org/nightly/reference/items/functions.html#function-body) documentation in this [comment](https://github.com/rust-lang/reference/issues/848#issuecomment-2274044966) is enough.

PR #1577, opened 2024-08-16
Add capturing and precise capturing rules

traviscross:
The Reference didn't include any description of capturing or capturing behavior for `impl Trait` opaque types.  Let's describe briefly what capturing is and what the currently-stable automatic capturing rules are.  Then let's describe the syntax and behavior of RFC 3617 precise capturing.

Precise capturing is currently undergoing stabilization in:

- https://github.com/rust-lang/rust/pull/127672

Tracking issue:

- https://github.com/rust-lang/rust/issues/123432

cc @compiler-errors @ehuss

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -5,7 +5,7 @@
 > &nbsp;&nbsp; _TypeParamBound_ ( `+` _TypeParamBound_ )<sup>\*</sup> `+`<sup>?</sup>
 >
 > _TypeParamBound_ :\
-> &nbsp;&nbsp; &nbsp;&nbsp; _Lifetime_ | _TraitBound_
+> &nbsp;&nbsp; &nbsp;&nbsp; _Lifetime_ | _TraitBound_ | _UseBound_
 >
 > _TraitBound_ :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `?`<sup>?</sup>
@@ -19,6 +19,21 @@
 > _Lifetime_ :\
 > &nbsp;&nbsp; &nbsp;&nbsp; [LIFETIME_OR_LABEL]\
 > &nbsp;&nbsp; | `'static`
+>
+> _UseBound_ :\
+> &nbsp;&nbsp; `use` _UseBoundGenericArgs_
+>
+> _UseBoundGenericArgs_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `<` `>` \
+> &nbsp;&nbsp; | `<` \
+> &nbsp;&nbsp; &nbsp;&nbsp; ( _UseBoundGenericArg_ `,`)<sup>\*</sup> \
+> &nbsp;&nbsp; &nbsp;&nbsp; _UseBoundGenericArg_ `,`<sup>?</sup> \
+> &nbsp;&nbsp; &nbsp;&nbsp; `>`
+>
+> _UseBoundGenericArg_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _Lifetime_ \
+> &nbsp;&nbsp; | [IDENTIFIER][] \
+> &nbsp;&nbsp; | `Self`
 
 [Trait] and lifetime bounds provide a way for [generic items][generic] to
 restrict which types and lifetimes are used as their parameters. Bounds can be
@@ -227,7 +242,11 @@ trait Trait<'a, T: 'a> {}
 impl<'a, T> Trait<'a, T> for &'a T {}
 ```
 
+## Use bounds
 
+Certain bounds lists may include a `use<..>` bound to control which generic parameters are captured by the `impl Trait` [abstract return type].  See [precise capturing] for more details.
+
+[IDENTIFIER]: identifiers.html
 [LIFETIME_OR_LABEL]: tokens.md#lifetimes-and-loop-labels
 [_GenericParams_]: items/generics.md
 [_TypePath_]: paths.md#paths-in-types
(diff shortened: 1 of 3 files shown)


src/types/impl-trait.md
```diff
@@ -91,7 +91,33 @@ Functions in traits may also use `impl Trait` as a syntax for an anonymous assoc
...
-### Differences between generics and `impl Trait` in return position
+## Capturing
+
+Behind each return-position `impl Trait` abstract return type is some hidden concrete type.  For this concrete type to use a generic parameter, that generic parameter must be *captured* by the abstract type.
+
+## Automatic capturing
+
+Return-position `impl Trait` abstract return types automatically capture certain of the in-scope generic parameters.  Everywhere, these automatically capture all in-scope type and const generic parameters.
+
+On items of trait impls and trait definitions, these types additionally automatically capture all in-scope generic lifetime parameters, including higher-ranked ones.
```
compiler-errors, line 102: What about non-RPITITs?

compiler-errors (approved):
probably should document the stable behavior for free fns/inherent impls

src/types/impl-trait.md
traviscross: Thanks.  Spelled out the negative case.

src/types/impl-trait.md
```diff
@@ -91,7 +91,33 @@ Functions in traits may also use `impl Trait` as a syntax for an anonymous assoc
...
-### Differences between generics and `impl Trait` in return position
+## Capturing
+
+Behind each return-position `impl Trait` abstract return type is some hidden concrete type.  For this concrete type to use a generic parameter, that generic parameter must be *captured* by the abstract type.
+
+## Automatic capturing
+
+Return-position `impl Trait` abstract return types automatically capture certain of the in-scope generic parameters.  Everywhere, these automatically capture all in-scope type and const generic parameters.
+
+On items of trait impls and trait definitions, these types additionally automatically capture all in-scope generic lifetime parameters, including higher-ranked ones.  On free functions and on associated functions and methods of inherent impls, they do not.
```
compiler-errors, line 102: When I asked for non-RPITIT case, I meant that we should probably specify what happens:

```suggestion
On items of trait impls and trait definitions, these types additionally automatically capture all in-scope generic lifetime parameters, including higher-ranked ones.  On free functions and on associated functions and methods of inherent impls, only the lifetimes that are mentioned in the return-position `impl Trait` abstract return type's bounds are captured.
```

traviscross: Got it.  Thanks for clarifying.

compiler-errors (approved):

src/trait-bounds.md
```diff
@@ -19,6 +19,20 @@
...
+> _UseBoundGenericArgs_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `<` `>` \
+> &nbsp;&nbsp; | `<` \
+> &nbsp;&nbsp; &nbsp;&nbsp; ( _UseBoundGenericArg_ `,`)<sup>\*</sup> \
+> &nbsp;&nbsp; &nbsp;&nbsp; _UseBoundGenericArg_ `,`<sup>?</sup> \
+> &nbsp;&nbsp; &nbsp;&nbsp; `>`
+>
+> _UseBoundGenericArg_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [LIFETIME_OR_LABEL][] \
+> &nbsp;&nbsp; | [IDENTIFIER][]
```
ehuss, line 35: What's the story about `Self` here? It looks like the compiler accepts it, but it is then rejected?

src/types/impl-trait.md
```diff
@@ -91,7 +91,33 @@ Functions in traits may also use `impl Trait` as a syntax for an anonymous assoc
...
+  //                                      ~~~~~~~~~~~~~~~~~~~~~~~
+  //                                     Captures `'a` and `T` only.
+  (x, y)
+}
+```
+
+Currently, only one `use<..>` bound may be present in a bounds list, such bounds are not allowed in the signature of items of a trait definition, all in-scope type and const generic parameters must be included, and all lifetime parameters that appear in other bounds of the abstract type must be included.  If the elided lifetime (`'_`) is otherwise allowed to appear within the `impl Trait` return type, it may be named within the `use<..>` bound.
+
+Because all in-scope type parameters must be included by name, a `use<..>` bound may not be used in the signature of items that use argument-position `impl Trait`, as those items have anonymous type parameters in scope.
+
```
ehuss, line 119: Perhaps include the same restriction that generic params has: "The order of generic parameters is restricted to lifetime parameters and then type and const parameters intermixed."

I assume the ordering restriction is the same.

src/trait-bounds.md
compiler-errors: `Self` is allowed semantically when it's a *parameter*, but the only case where that is true is when it's in an RPITIT, which we don't allow `use<>` on yet.

compiler-errors: Syntactically `Self` is just an identifier, so it should be allowed. But that's the same as non-param lifetimes like `'static`, which are allowed syntactically but not semantically.

ehuss: The reference excludes keywords from IDENTIFIER, and `Self` is a keyword.

I guess similarly for `'static`, that would mean this accepts _Lifetime_, not LIFETIME_OR_LABEL.

Is there a timeline for supporting RPITIT?

compiler-errors: Lifetime_or_label is a subset of lifetime? Confusing

> Is there a timeline for supporting RPITIT?

Nope

compiler-errors: > The reference excludes keywords from IDENTIFIER, and Self is a keyword.

Well, so the only keyword that we allow here is `Self`, so let's make it `IDENTIFIER | Self | LIFETIME`.

traviscross: Thanks; updated;

ehuss (approved):
Thanks!

PR #1578, opened 2024-08-17
Rewrite the automatic std link translation, and switch to automatic links

ehuss:
This rewrites the automatic standard link translation code. The old code didn't work very well, and this switches it to a new approach which should work better (see individual commits for more information).

This also switches the content to start using the automatic link conversion. From what I can tell from running diffs before and after, there shouldn't be any significant changes. The main change is that rustdoc links re-exports to the original definition instead of the re-export site. For example, a bunch of things that link to `std` now end up linking to `core` or `alloc`.


traviscross (approved):
Looks right to me.  I pushed the things noted and a couple others to the branch.  I'll let you have a look before we merge it.

PR #1579, opened 2024-08-19
2024: Add unsafe attribute differences

ehuss:
Draft pending stabilization of 2024.

### DIFF
--- a/src/abi.md
+++ b/src/abi.md
@@ -74,6 +74,8 @@ with the same name (or with a well-known symbol), leading to undefined behavior.
 extern "C" fn foo() {}
 ```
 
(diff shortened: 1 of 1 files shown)


traviscross (approved):

src/abi.md
```diff
@@ -74,6 +74,8 @@ with the same name (or with a well-known symbol), leading to undefined behavior.
 extern "C" fn foo() {}
 ```
 
+> **Edition Differences**: Before the 2024 edition it is allowed to use the `no_mangle` attribute without the `unsafe` qualification.
```
traviscross, line 77: The Reference is split between "Edition Differences" and "Edition differences".  The former is in the majority, but I think the latter actually makes more sense given that we use sentence-case for headers, and this is a kind of header.

Otherwise, the changes here look right.

PR #1588, opened 2024-08-26
Clarify requires of Prefix/Directive validity requirements

chorman0773:
This alters the text of the inline-assembly chapter to clarify that ending an inline-assembly block with a prefix or using a "stateful" assembly directive without reverting it results in undefined results whenever the program is executed at all - even if it doesn't eventually evaluate (let alone return from) the assembly block.  It also notes that the compiler may (though does not currently) diagnose improper uses of stateful assembly directives that would violate that point. 

This also removes the specific application of the prefix to x86 - I do not know whether this applies to any existing supported architectures, but it future proofs adding new such architectures.

Editorially, it also moves the aforementioned rules into the `[asm.validity]` section, and renames the identifiers accordingly.


src/inline-assembly.md
```diff
@@ -684,6 +680,14 @@ and the rules for assembly may include thousands of pages of architectural refer
 Programmers should exercise appropriate care, as invoking this `unsafe` capability comes with
 assuming the responsibility of not violating rules of both the compiler or the architecture.
 
+r[asm.validity.stateful]
+If inline assembly includes any "stateful" directive that modifies how subsequent assembly is processed, the block must undo the effects of any such directives before the inline assembly ends. If this constraint is violated, an error may be issued if it can be detected, but otherwise the result of executing the program that contains it is undefined - regardless of whether or not the inline assembly block that contains is eventually evaluated by that execution.
+
+r[asm.validity.prefix-restriction]
+Inline assembly must not end with an instruction prefix (such as `LOCK`) that would apply to instructions generated by the compiler. If this constraint is violated, an error may be issued if it can be detected, but otherwise the result of executing the program that contains it is undefined - regardless of whether or not the inline assembly block that contains is eventually evaluated by that execution.
```
Amanieu, line 687: I would generalize this even more: at the end of the asm block, the CPU must be in a state where it is ready to start a new instruction. Alternatively, the asm block must end at an instruction boundary. This covers cases where you use `.byte` to only emit a part of an instruction, or leaving the assembler at an unaligned address for architectures that require aligned instructions (e.g. aarch64).

ehuss:
@chorman0773 Just checking in, is this something you want to continue working on?

chorman0773:
Yes, I just haven't had much time to work on it.

PR #1590, opened 2024-08-27
const_eval: update for const_mut_refs and const_refs_to_cell stabilization

RalfJung:
This is the reference PR accompanying https://github.com/rust-lang/rust/pull/129195.

Fixes https://github.com/rust-lang/rust/issues/57349
Fixes https://github.com/rust-lang/rust/issues/80384

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -41,7 +41,7 @@ to be run.
 * All forms of [borrow]s, including raw borrows, with one limitation:
   mutable borrows and shared borrows to values with interior mutability
   are only allowed to refer to *transient* places. A place is *transient*
-  if it will be deallocated before the end of evaluating the current constant item.
+  if its lifetime is strictly contained inside the current [const context].
 * The [dereference operator].
 * [Grouped] expressions.
 * [Cast] expressions, except
@@ -52,6 +52,7 @@ to be run.
 * [if], [`if let`] and [match] expressions.
 
 ## Const context
+[const context]: #const-context
 
 A _const context_ is one of the following:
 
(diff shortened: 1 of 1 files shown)


src/const_eval.md
```diff
@@ -38,8 +38,11 @@ to be run.
 * [Closure expressions] which don't capture variables from the environment.
 * Built-in [negation], [arithmetic], [logical], [comparison] or [lazy boolean]
   operators used on integer and floating point types, `bool`, and `char`.
-* Shared [borrow]s, except if applied to a type with [interior mutability].
-* The [dereference operator] except for raw pointers.
+* All forms of [borrow]s, including raw borrows, with one limitation:
+  mutable borrows and shared borrows to values with interior mutability
+  are only allowed to refer to *transient* places. A place is *transient*
+  if it will be deallocated before the end of evaluating the current constant item.
```
traviscross, line 44: How do we think about what a "constant item" means with respect to `const { .. }` blocks?  Would it be better here to say "end of evaluating the current [constant context](https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context)"?

RalfJung: If "evaluating a const context" is meaningful, then yeah that would work. It sounds a bit odd to me, how can you evaluate a "context"? The context is what the evaluation occurs in.

No strong preference either way.

traviscross: Maybe "evaluating the contents of the current const context"?  Open to other ideas.  Saying either "evaluating a context" or "evaluating an item" sound about equally odd to me.

The main thrust is that [const context](https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context) is something we define and seems to be closer to what we mean here.

RalfJung: "evaluating the item" is how we call it in the compiler so I got used to that one. ;)

traviscross (approved):

ehuss (approved):
Thanks!

PR #1596, opened 2024-08-29
stabilize `const_extern_fn`

folkertdev:
to stabilize https://github.com/rust-lang/rust/issues/64926.

This feels a little bare-bones, but I guess that is sort of what the spec is for: just factual statements.


traviscross (approved):
This looks right.  This is pending the stabilization in:

- https://github.com/rust-lang/rust/pull/129753

ehuss (approved):
Thanks!

PR #1599, opened 2024-08-30
Add identifier syntax to items.md and subchapters

chorman0773:
This adds identifier syntax to items.md and its subchapters


ehuss (approved):
Thanks! I pushed some fixes and resolved some merge conflicts.

PR #1600, opened 2024-09-03
dont use stdcall on x86_64 where it is not a valid ABI

RalfJung:
Unblocks https://github.com/rust-lang/rust/pull/129935

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -195,7 +195,7 @@ called from other programming languages like C:
 extern "C" fn new_i32() -> i32 { 0 }
 
 // Declares a function with the "stdcall" ABI
-# #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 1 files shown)


RalfJung:
@ehuss any preference which ABI to use here? stdcall is 32bit-only, vectorcall is unstable. We could use "system"?

ehuss:
As an alternative, you can just change the `cfg` to be something like `cfg(windows)` (assuming that is "compatible" for stdcall).

RalfJung:
That would make it dead code for CI, but yeah that also works.

ehuss:
> That would make it dead code for CI, but yeah that also works.

I think it's fine. It gets tested in rust-lang/rust under windows.

ehuss (approved):
Thanks!

PR #1601, opened 2024-09-03
Add Lifetime Capture Rules 2024

traviscross:
The rules for which generic lifetime parameters are automatically captured in an RPIT-like opaque type are changing in Rust 2024 according to RFC 3498.  Let's document these changes.


ehuss (approved):
Thanks!

PR #1602, opened 2024-09-07
Document limitations on block doc comments

GKFX:
It is unintuitive that the Markdown contained within doc comments is not actually part of their syntax, as evidenced in the linked issue. This can result in certain block doc comments parsing (or rather, failing to parse) in an unintuitive way.

Closes rust-lang/rust#129685.


ehuss (approved):
Thanks!

PR #1603, opened 2024-09-08
Raw lifetimes

compiler-errors:
See https://github.com/rust-lang/rust/pull/126452

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -767,7 +767,7 @@ r[lex.token.life.syntax]
 > &nbsp;&nbsp; | RAW_LIFETIME
 >
 > RAW_LIFETIME :\
-> &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
+> &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier] <sub>*Except `crate`, `self`, `super`, `Self`*</sub>
 >   _(not immediately followed by `'`)_
 
 r[lex.token.life.intro]
@@ -779,7 +779,7 @@ r[lex.token.life.raw.intro]
 A raw lifetime is like a normal lifetime, but its identifier is prefixed by `r#`. (Note that the `r#` prefix is not included as part of the actual lifetime.)
 
 r[lex.token.life.raw.allowed]
-Unlike a normal lifetime, a raw lifetime may be any strict or reserved keyword.
+Unlike a normal lifetime, a raw lifetime may be any strict or reserved keyword except the ones listed above for `RAW_LIFETIME`.
 
 > **Edition differences**: Raw lifetimes are accepted in the 2021
 > edition or later. In earlier additions the token `'r#lt` is lexed as `'r # lt`.
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -781,6 +781,7 @@ A raw lifetime is like a normal lifetime, but its identifier is prefixed by `r#`
 r[lex.token.life.raw.allowed]
 Unlike a normal lifetime, a raw lifetime may be any strict or reserved keyword except the ones listed above for `RAW_LIFETIME`.
 
+r[lex.token.life.raw.edition2021]
 > **Edition differences**: Raw lifetimes are accepted in the 2021
 > edition or later. In earlier additions the token `'r#lt` is lexed as `'r # lt`.
 
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -768,6 +768,7 @@ r[lex.token.life.syntax]
 >
 > RAW_LIFETIME :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
+>   _(not immediately followed by `'`)_
 
 r[lex.token.life.intro]
 Lifetime parameters and [loop labels] use LIFETIME_OR_LABEL tokens. Any
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -769,6 +769,9 @@ r[lex.token.life.syntax]
 > RAW_LIFETIME :\
 > &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier] <sub>*Except `crate`, `self`, `super`, `Self`*</sub>
 >   _(not immediately followed by `'`)_
+>
+> RESERVED_RAW_LIFETIME : `'r#_`
+>   _(not immediately followed by `'`)_
 
 r[lex.token.life.intro]
 Lifetime parameters and [loop labels] use LIFETIME_OR_LABEL tokens. Any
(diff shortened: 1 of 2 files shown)


src/tokens.md
```diff
@@ -642,12 +642,17 @@ Examples of reserved forms:
...
 >
 > LIFETIME_OR_LABEL :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `'` [NON_KEYWORD_IDENTIFIER][identifier]
 >   _(not immediately followed by `'`)_\
 > &nbsp;&nbsp; | `'_`
 >   _(not immediately followed by `'`)_
+> &nbsp;&nbsp; | RAW_LIFETIME
+>
+> RAW_LIFETIME :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
```
compiler-errors, line 655:
```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
>   _(not immediately followed by `'`)_
```

mattheww:
The

> (not immediately followed by `'`)

bit is there in `LIFETIME_OR_LABEL` as a way to say that `'xxx'yyy` is rejected (rather than being interpreted as two lifetimes).

But (after rust-lang/rust#126452) `'r#xxx'yyy` *is* interpreted as two lifetimes. So I think that bit should be left out of the `RAW_LIFETIME` rule.

(There's no need to say anything special to indicate that `'r#kw'` is rejected: that's true because `'` on its own isn't a token, and character literals can't have more than one character between the quotes.)


Alternatively, it might be worth considering changing the implementation to reject `'r#xxx'yyy` for consistency with the non-raw case.

Note that allowing that form would close the door to delaying the rejection of overlong character literals to post-expansion, which was being considered late last year at rust-lang/rust#118699 .

ehuss (commented):
Can you update https://doc.rust-lang.org/nightly/reference/tokens.html#reserved-prefixes to include a new rule for the reserved prefix lifetime? I would assume it is something like 
```
`'` (IDENTIFIER_OR_KEYWORD | _) `#`
```

And update the examples in the "Edition differences" in that section to include reserved lifetime prefixes.

Can you add a "Raw lifetime or label" section similar to https://doc.rust-lang.org/nightly/reference/identifiers.html#raw-identifiers that explains what the raw lifetime means?

Can you add an "Edition differences" block in the "Lifetimes and loop labels" section that explains that raw lifetimes are not supported before 2021? There are some examples within the tokens.md chapter of "Edition differences" blocks for the kind of wording to use.

ehuss:
@compiler-errors Just checking if you'll be able to look at the requested changes?

compiler-errors:
I believe I addressed the comments, but please look closely because markdown is not my thing.

compiler-errors:
Well, except for:

> Alternatively, it might be worth considering changing the implementation to reject 'r#xxx'yyy for consistency with the non-raw case.

Which I think I will open a rustc PR to implement.

src/tokens.md
```diff
@@ -758,19 +758,34 @@ r[lex.token.life.syntax]
...
 >
 > LIFETIME_OR_LABEL :\
 > &nbsp;&nbsp; &nbsp;&nbsp; `'` [NON_KEYWORD_IDENTIFIER][identifier]
 >   _(not immediately followed by `'`)_\
 > &nbsp;&nbsp; | `'_`
+>   _(not immediately followed by `'`)_\
+> &nbsp;&nbsp; | RAW_LIFETIME
+>
+> RAW_LIFETIME :\
+> &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
```
ehuss, line 772: I believe this should also include `_`, right? That is, `'r#_` is valid, correct?

I also just want to double-check, it is intentional that `'r#_` is allowed as a loop label? (I only ask because `'_` is explicitly not allowed).

```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; `'r#` [IDENTIFIER_OR_KEYWORD][identifier]
>   _(not immediately followed by `'`)_\
> &nbsp;&nbsp; | `'r#_`
```

compiler-errors: Actually this is likely something *else* we want to deny here.  I feel like `'r#_` is invalid, or at least not something we need to support initially.

mattheww:
Do we need changes outside the _Lexical structure_ chapter?

As far as I can see nothing is saying that `r#a` and `'a` are treated as equivalent when used as a lifetime or label.

That is, I don't think the Reference is saying that these are now accepted:

```rust
fn foo<'r#a>(s: &'a str) {}
```

```rust
    'r#a: { break 'a; }
```

src/tokens.md
```diff
@@ -851,7 +866,8 @@ r[lex.token.reserved-prefix.syntax]
 > **<sup>Lexer 2021+</sup>**\
 > RESERVED_TOKEN_DOUBLE_QUOTE : ( IDENTIFIER_OR_KEYWORD <sub>_Except `b` or `c` or `r` or `br` or `cr`_</sub> | `_` ) `"`\
 > RESERVED_TOKEN_SINGLE_QUOTE : ( IDENTIFIER_OR_KEYWORD <sub>_Except `b`_</sub> | `_` ) `'`\
-> RESERVED_TOKEN_POUND : ( IDENTIFIER_OR_KEYWORD <sub>_Except `r` or `br` or `cr`_</sub> | `_` ) `#`
+> RESERVED_TOKEN_POUND : ( IDENTIFIER_OR_KEYWORD <sub>_Except `r` or `br` or `cr`_</sub> | `_` ) `#`\
+> RESERVED_TOKEN_LIFETIME : `'` (IDENTIFIER_OR_KEYWORD <sub>_Except `r`_</sub> | _) `#`
```
mattheww, line 870: I think this is isn't reserving `'r#` on its own, which rustc rejects but could lex as `'r` followed by `#`.

But I'm not sure it's worth trying to get all the `RESERVED_` rules exactly right at this stage.

ehuss:
Posted https://github.com/rust-lang/edition-guide/pull/330 as the companion for the edition.

src/tokens.md
ehuss: https://github.com/rust-lang/rust/pull/132363

Hopefully resolved by https://github.com/rust-lang/reference/pull/1603/commits/912a6d5c0c04ea77a90fcdbb5ca8fe7a5f12f6b9

ehuss:
@compiler-errors I pushed a change for the validation, and also rebased since we made a slight change in the definition of the lifetime token in https://github.com/rust-lang/reference/pull/1668.

Please let me know if you think the current version looks good.

ehuss:
I pushed a commit specifying that `'r#_` will generate an error (and similarly for `r#_`).

This is a bit of an awkward issue around whether or not `_` is an identifier. Currently the reference does not define it that way (intentionally), even though internally rustc allows that (but quickly rejects it). IIRC, there are some issues around that and proc-macros, but I did not look into that.

I did not want to define an identifier token that allowed `_`, because then I would have to define separate identifier tokens that exclude it, and this `_` identifier would only be used by these raw tokens.

compiler-errors:
LGTM!

ehuss (approved):
Thanks!

PR #1606, opened 2024-09-09
Places based on misaligned pointers: also mention 'static's as a possible base for place projections

RalfJung:
In MIR, locals are the only place "roots" since statics get turned into `*PTR` where `PTR` is a constant pointer pointing to that static... but in the surface language, `STATIC.foo.bar` arguably has `STATIC` as its root and contains no `*`, so let's add that as a possible case here.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -78,7 +78,7 @@ The span of bytes a pointer or reference "points to" is determined by the pointe
 A place is said to be "based on a misaligned pointer" if the last `*` projection
 during place computation was performed on a pointer that was not aligned for its
 type. (If there is no `*` projection in the place expression, then this is
-accessing the field of a local and rustc will guarantee proper alignment. If
+accessing the field of a local or `static` and rustc will guarantee proper alignment. If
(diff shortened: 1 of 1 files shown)


ehuss:
@RalfJung Just curious, this came up in discussing https://github.com/rust-lang/reference/pull/1610 where it seems like statics get converted to implicit dereferences as you mentioned. Is that something we need to explicitly say somewhere? Is this a property that should be stated, or is it just a consequence of how MIR works and there could be other ways it was implemented? Is there existing terminology for what this `static` translation is called?

traviscross (approved):

RalfJung:
> statics get converted to implicit dereferences as you mentioned. Is that something we need to explicitly say somewhere? Is this a property that should be stated, or is it just a consequence of how MIR works and there could be other ways it was implemented?

I consider this to be a MIR implement detail that should not be mentioned in the reference.

Cc @rust-lang/opsem

digama0:
Agreed that from a language perspective we should treat locals and statics on similar footing and not expose the pointer desugaring.

PR #1607, opened 2024-09-09
Fix improper documentation on casting non_exhaustive enums

mixy1:
Added a more complete example that explains what is allowed and what's not allowed when the non_exhaustive attribute is in use.

### DIFF
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -163,7 +163,7 @@ match message {
 }
 ```
 
-However, casting non-exhaustive types from foreign crates is generally disallowed, except when dealing with enums that have no non-exhaustive variants.
+It's also not allowed to use numeric casts (`as`) on enums that contain any non-exhaustive variants.
 
 For example, the following enum can be cast because it doesn't contain any non-exhaustive variants:
 ```rust, ignore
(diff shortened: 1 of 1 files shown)


src/attributes/type_system.md
```diff
@@ -163,12 +163,33 @@ match message {
 }
 ```
 
-It's also not allowed to cast non-exhaustive types from foreign crates.
+However, casting non-exhaustive types from foreign crates is generally disallowed, except when dealing with enums that have no non-exhaustive variants.
```
obi1kenobi, line 166: I'm not a maintainer of the reference, so this will need a review from someone else as well.

I'm not sure if casting any other type other than an enum is ever allowed in Rust. I can't think of any, and if there are some, it might be good to mention them.

A quick suggestion to avoid starting with `However` in two nearby paragraphs:
```suggestion
It's also not allowed to use numeric casts (`as`) on enums that contain any non-exhaustive variants.
```

mixy1: whoops completely missed the duplication of "However" 😬. Yeah I don't think anything other than an enum can be cast as well. I'll commit the suggestion.

ehuss (approved):
Thanks! Not sure how I missed this in the original PR.

I pushed some tweaks to the examples just to clarify some things.

PR #1609, opened 2024-09-10
Explain how to name rule identifiers

chorman0773:
This adds guidelines for using spec identifier syntax in the reference.

It is also intended to aid review in the PRs for adding spec identifier syntax.

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -78,12 +78,12 @@ In the HTML, the rules are clickable just like headers.
 When assigning rules to new paragraphs, or when modifying rule names, use the following guidelines:
 1. A rule applies to one core idea, which should be easily determined when reading the paragraph it is applied to,
 2. Other than the "intro" paragraph, purely explanatory, expository, or exemplary content does not need a rule. If the expository paragraph isn't directly related to the previous, separate it with a hard (rendered) line break
-3. Rust Code examples and tests do not need their own rules
+3. Rust code examples and tests do not need their own rules
 4. Notes do not need rules. For other admonition types, use the following guidelines:
     * Warning: Omit the rule if and only if the warning follows from the previous paragraph.
-    * Target Specific Behaviour: Always include the rule
-    * Edition Differences: Always include the rule
-    * Version History: Omit the rule if the present behaviour is explained in the immediately preceeding rule.
+    * Target specific behaviour: Always include the rule
+    * Edition differences: Always include the rule
+    * Version history: Omit the rule if the present behaviour is explained in the immediately preceeding rule.
 4. The following keywords should be used to identify paragraphs when unambiguous:
     * `intro`: The beginning paragraph of each section - should explain the construct being defined overall.
     * `syntax`: Syntax definitions or explanations when BNF syntax definitions are not used
@@ -95,7 +95,7 @@ When assigning rules to new paragraphs, or when modifying rule names, use the fo
     * `preconditions`: Conditions which must be satisfied for the evaluation of the construct to be well-defined
     * `namespace`: For items only, specifies the namespace(s) the item introduces a name in. May also be used elsewhere when defining a namespace (e.g. `r[attribute.diagnostic.namespace]`)
 5. When a rule doesn't fall under the above keywords. or for section rule ids, name the subrule as follows:
-    * If the rule is naming a specific Rust language construct (IE. an attribute, standard library type/function, or keyword-introduced concept), use the construct as named in the language, appropriately case-adjusted (but do not replace `_`s with `-`s)
+    * If the rule is naming a specific Rust language construct (e.g. an attribute, standard library type/function, or keyword-introduced concept), use the construct as named in the language, appropriately case-adjusted (but do not replace `_`s with `-`s)
     * Other than rust language concepts with `_`s in the name, use `-` characters to separate words within a "subrule"
     * Whenever possible, do not repeat previous components of the rule
     * Prefer using singular forms of words over plural unless the rule applies to a list or the construct is named as plural in the language (e.g. `r[attribute.diagnostic.lint.group])
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -94,8 +94,10 @@ When assigning rules to new paragraphs, or when modifying rule names, use the fo
     * Other than rust language concepts with `_`s in the name, use `-` characters to separate words within a "subrule"
     * Whenever possible, do not repeat previous components of the rule
     * Prefer using singular forms of words over plural unless the rule applies to a list or the construct is named as plural in the language (e.g. `r[attribute.diagnostic.lint.group])
-    * Whenever possible, don't use a name that conflicts with one of the above keywords, even if this violates the first bullet. 
+    * Edition differences admonitions should typically be named by the edition referenced directly by the rule. If multiple editions are named, use the one for which the behaviour is defined by the admonition, and not by a previous paragraph
+    * Target Specific admonitions should typically be named by the least specific target property to which they apply (e.g. if a rule affects all x86 cpus, the rule name should include `x86` rather than separately listing `i586`, `i686` and `x86_64`, and if a rule applies to all ELF platforms, it should be named `elf` rather than listing every ELF os) 
     * Use an appropriately discriptive, but short, name if the language does not provide one.
+    
 
 ### Standard library links
 
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
 
 In the HTML, the rules are clickable just like headers.
 
+When assigning rules to new paragraphs, or when modifying rule names, use the following guidelines:
+1. A rule applies to one core idea, which should be easily determined when reading the paragraph it is applied to,
+2. Other than the "intro" paragraph, purely explanatory, expository, or exemplary content does not need a rule. If the expository paragraph isn't directly related to the previous, separate it with a hard (rendered) line break
+3. Rust Code examples and tests do not need their own rules
+4. Notes do not need rules. For other admonition types, use the following guidelines:
+    * Warning: Omit the rule if and only if the warning follows from the previous paragraph.
+    * Target Specific Behaviour: Always include the rule
+    * Edition Differences: Always include the rule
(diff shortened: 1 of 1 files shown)


docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
 
 In the HTML, the rules are clickable just like headers.
 
+When assigning rules to new paragraphs, or when modifying rule names, use the following guidelines:
+1. A rule applies to one core idea, which should be easily determined when reading the paragraph it is applied to,
+2. Other than the "intro" paragraph, purely explanatory, expository, or exemplary content does not need a rule. If the expository paragraph isn't directly related to the previous, separate it with a hard (rendered) line break
+3. Rust Code examples and tests do not need their own rules
```
traviscross, line 81:
```suggestion
3. Rust code examples and tests do not need their own rules
```

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+4. Notes do not need rules. For other admonition types, use the following guidelines:
+    * Warning: Omit the rule if and only if the warning follows from the previous paragraph.
+    * Target Specific Behaviour: Always include the rule
```
traviscross, line 84:
```suggestion
    * Target specific behaviour: Always include the rule
```

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * Edition Differences: Always include the rule
```
traviscross, line 85:
```suggestion
    * Edition differences: Always include the rule
```

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * Version History: Omit the rule if the present behaviour is explained in the immediately preceeding rule.
```
traviscross, line 86:
```suggestion
    * Version history: Omit the rule if the present behaviour is explained in the immediately preceeding rule.
```

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+4. The following keywords should be used to identify paragraphs when unambiguous:
+    * `intro`: The beginning paragraph of each section - should explain the construct being defined overall.
+    * `syntax`: Syntax definitions or explanations when BNF syntax definitions are not used
+    * `restriction`: Syntactic (parsing) requirements on the construct
+    * `constraint`: Semantic (type checking) requirements on the construct
```
traviscross, line 91: @ehuss and I talked about this part.  We're a but uncertain about the strategy here.  It feels like there are going to be a ton of `restriction-this` and `constraint-that` rules, and we're not sure that it carries its weight.

We're also probably unconvinced on this means of separating syntactic and type checking requirements.  I.e., it doesn't immediately speak to us that "restriction" would refer to parsing and "constraint" would refer to type checking and that these should be separated in this manner.

This probably also applies to some of the others, such as "preconditions".

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * `safety` (instead of restriction): Stating that an operation is `unsafe` or the conditions under which it is `unsafe`
+    * `behavior`: Runtime effects of evaluating the construct in a well-defined manner
+    * `panics`: Conditions under which evaluating the construct causes a runtime panic
+    * `preconditions`: Conditions which must be satisfied for the evaluation of the construct to be well-defined
+    * `namespace`: For items only, specifies the namespace(s) the item introduces a name in. May also be used elsewhere when defining a namespace (e.g. `r[attribute.diagnostic.namespace]`)
+5. When a rule doesn't fall under the above keywords. or for section rule ids, name the subrule as follows:
+    * If the rule is naming a specific Rust language construct (IE. an attribute, standard library type/function, or keyword-introduced concept), use the construct as named in the language, appropriately case-adjusted (but do not replace `_`s with `-`s)
```
traviscross, line 98:
```suggestion
    * If the rule is naming a specific Rust language construct (e.g. an attribute, standard library type/function, or keyword-introduced concept), use the construct as named in the language, appropriately case-adjusted (but do not replace `_`s with `-`s)
```

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * Other than rust language concepts with `_`s in the name, use `-` characters to separate words within a "subrule"
+    * Whenever possible, do not repeat previous components of the rule
+    * Prefer using singular forms of words over plural unless the rule applies to a list or the construct is named as plural in the language (e.g. `r[attribute.diagnostic.lint.group])
```
traviscross, line 101: I'm not sure this is true everywhere.  Plural often makes sense.  I'd expect chapter titles like "Items" (rather than "Item") or "Lints" (rather than "Lint"), and I'd expect the rule names to follow suit here.

Often, what's going to come up, as it did with the diagnostic namespace, is not the difference between plural and singular, but the difference between nouns and adjectives.  As a noun, it makes sense for the chapter title (and rules that follow that) to be "Diagnostics", but when used with "diagnostic namespace", it's an adjective, and so it doesn't have the "s".

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * Whenever possible, don't use a name that conflicts with one of the above keywords, even if this violates the first bullet.
```
traviscross, line 102: Can we think of any examples of where this would come up?

docs/authoring.md
chorman0773: `item.fn`

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+5. When a rule doesn't fall under the above keywords. or for section rule ids, name the subrule as follows:
+    * If the rule is naming a specific Rust language construct (IE. an attribute, standard library type/function, or keyword-introduced concept), use the construct as named in the language, appropriately case-adjusted (but do not replace `_`s with `-`s)
+    * Other than rust language concepts with `_`s in the name, use `-` characters to separate words within a "subrule"
+    * Whenever possible, do not repeat previous components of the rule
+    * Prefer using singular forms of words over plural unless the rule applies to a list or the construct is named as plural in the language (e.g. `r[attribute.diagnostic.lint.group])
+    * Whenever possible, don't use a name that conflicts with one of the above keywords, even if this violates the first bullet. 
+    * Use an appropriately discriptive, but short, name if the language does not provide one.
+6. When a keyword applies, but multiple different rules in the same section would use the same keyword, prefix or suffix the rule with a descriptive id given above, separated with a `-`
```
traviscross, line 104: Are there some cases where we'd want to prefix and others where we'd want to suffix?

traviscross: Nevermind; this is answered in the bullet points that follow.

docs/authoring.md
```diff
@@ -75,6 +75,36 @@ Rules can be linked to by their ID using markdown such as `[foo.bar]`. There are
...
+    * When the paragraph modifies a specific named construct or applies to a specific named construct only, prefix the rule with the name of the construct (e.g. `r[items.fn.params.self-constraint]`).
+    * When the paragraph refers to a specific named construct that applies the particular keyword behaviour, suffix the rule with the name of the construct
```
traviscross, line 106: In discussion we were a bit unclear about the distinction being drawn here.  Perhaps more examples would help.

docs/authoring.md
chorman0773: `AssertUnwindSafe` is probably close (`assert-unwind-safe` if we ever refer to it in the reference). I can't think of any specific examples, but that doesn't mean it won't happen (especially if we expand the list of keywords).

docs/authoring.md
chorman0773: I actually did this a couple times working on #1618. `r[type.trait-object.syntax-edition2021]` - it refers to Edition 2021 and modifies the syntax rule accordingly, so we use a suffix here. `r[type.text.char-precondition]` is a precondition that applies to `char`, rather than modifying a general precondition because of the use of `char`.

As another example,  `r[items.fn.params.self-constraint]`, is talking about the functions that can use a `self` param... which probably should be a suffix here as well, thinking about it. `self-constraint` probably should be for talking about what types a `self` param could have.

traviscross:
On the lang-docs call, we had a long discussion about this.  The general thrust is that we'd like in general to be descriptive rather than prescriptive and to follow patterns that emerge from a recurring need in the document.

To that end, we want to remove the specific keywords noted in the document, other than "intro", "syntax", and "namespace".

We want to remove the prescriptive note about singular versus plural.  And let's remove this rule:

> * Whenever possible, don't use a name that conflicts with one of the above keywords, even if this violates the first bullet.

Let's remove item 6 entirely also for the same reason.

And then, for this rule:

> 2. Other than the "intro" paragraph, purely explanatory, expository, or exemplary content does not need a rule. If the expository paragraph isn't directly related to the previous, separate it with a hard (rendered) line break

We didn't want to use hard line breaks for this.  Instead, we're interested in moving all of the expository material into delimited notes sections (this applies also to rule 3).

chorman0773:
> Instead, we're interested in moving all of the expository material into delimited notes sections (this applies also to rule 3).

That was my thought as well, but that's a later pass to be honest. I'd rather have rules that keep things consistent now.

ehuss (approved):
Thanks! I'm going to go ahead and approve, since I think we have agreement on what I've edited, and I don't want to block progress over specific points we haven't agreed on. Specifically:

- Version history: I don't think we've made an agreement if or how we're going to be documenting any historical information.
- Singular forms: I think we are using plurals almost all over (the existing rule PRs, section headings, etc.). There may be some subtle details about what guidance we want to give here, but I don't know if we can hammer them out in a very short term. For now, I think the identifiers should usually be plural to match the style in the section headers, and we can figure out what kind of refinement we want to give on that guidance in a followup. For example, we have some abbreviated or shortened forms that aren't plural, which seem fine to me. There are other factors, such as whether or not something is categorical, an initialism, an adjective, etc.

PR #1610, opened 2024-09-11
`const` expression can borrow static items

### LINKED ISSUE #1690: The reference is wrong about constants referencing statics since v1.83.0
Hello! I've noticed that the Rust 1.83.0 announcement [mentions](https://blog.rust-lang.org/2024/11/28/Rust-1.83.0.html#new-const-capabilities) that it is now possible to reference `static` items in `const` contexts:
```Rust
static S: i32 = 25;
const C: &i32 = &S;
```

But the Rust Reference still [says](https://doc.rust-lang.org/reference/items/static-items.html) it is not:
> All access to a static is safe, but there are a number of restrictions on statics:
> - The type must have the `Sync` trait bound to allow thread-safe access.
> - **Constants cannot refer to statics.**

I am willing to write a PR for this but I am not sure what the best course of action is. Simply removing that bullet-point would mean having to rephrase the text above it, but I don't know what would sound best (also, a single bullet-point may look weird).

What should I do?

dingxiangfei2009:
Tracked by rust-lang/rust#119618

Stabilization report: https://github.com/rust-lang/rust/issues/128183

Fix #1690

This PR updates the documentation to mention that now `static` items can be used for borrowing, explicitly or implicitly, which permits general use of them in `const` context.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -42,7 +42,12 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions:
+  * Writes to `static` items are not allowed in any constant evaluation context.
+  * Reads from `extern` statics are not allowed in any constant evaluation context.
+  * If the evaluation is *not* carried out in an initializer of a `static` item, then reads from any mutable `static` are not allowed. A mutable `static` is a `static mut` item, or a `static` item with an interior-mutable type.
+
+These requirements are checked only when the constant is evaluated. In other words, having such accesses syntactically occur in const contexts is allowed as long as they never get executed.
 
 r[const-eval.const-expr.tuple]
 * [Tuple expressions].
@@ -177,6 +182,7 @@ of whether you are building on a `64` bit or a `32` bit system.
 [enum discriminants]:   items/enumerations.md#discriminants
 [expression statements]: statements.md#expression-statements
 [expressions]:          expressions.md
+[`extern` statics]:     items/external-blocks.md#statics
 [field]:                expressions/field-expr.md
 [functions]:            items/functions.md
 [grouped]:              expressions/grouped-expr.md
--- a/src/items/constant-items.md
+++ b/src/items/constant-items.md
@@ -48,6 +48,9 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
 };
 ```
 
+r[items.const.final-value-immutable]
+The final value of a `const` item cannot contain references to anything mutable.
+
 r[items.const.expr-omission]
 The constant expression may only be omitted in a [trait definition].
 
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -37,10 +37,7 @@ All access to a static is safe, but there are a number of restrictions on
 statics:
 
 r[items.static.sync]
-* The type must have the `Sync` trait bound to allow thread-safe access.
-
-r[items.static.const]
-* Constants cannot refer to statics.
+* The type must have the [`Sync`](std::marker::Sync) trait bound to allow thread-safe access.
 
 r[items.static.init.omission]
 The initializer expression must be omitted in an [external block], and must be
@@ -159,7 +156,7 @@ It can be confusing whether or not you should use a constant item or a static
 item. Constants should, in general, be preferred over statics unless one of the
 following are true:
 
-* Large amounts of data are being stored
+* Large amounts of data are being stored.
 * The single-address property of statics is required.
 * Interior mutability is required.
 


ehuss (commented):
Can you also update https://github.com/rust-lang/reference/blob/master/src/items/static-items.md which mentions this isn't possible?

I'm wondering if this might be confusion (and perhaps my understanding is missing) with regards to what counts as a borrow. IIUC, statics have an implied `&`, is that correct? For example, if you see `const C: i32 = SOME_STATIC;`, there's no obvious borrow there.

dingxiangfei2009:
@ehuss I updated the wording to include simply "use of statics".

I have updated `static-items` page as well with tweaks on wording.

traviscross:
We probably want to update the `constant-items` page that currently says:

> Constants may refer to the address of other constants...

Since it'd now make sense for that to also address the fact that constants may refer to statics with some limitations.

traviscross:
Also in `const_eval.md`, where it says:

> Paths to [statics](https://github.com/rust-lang/reference/pull/items/static-items.html). These are only allowed within the initializer of a static.

...that seems to need to change also.

traviscross:
More generally, it seems what we need to do here is to loosen the various places in the Reference that have these restrictions, and in loosening them, add the necessary caveats, e.g. with respect to `Freeze` as described in #128183.

RalfJung:
> For example, if you see const C: i32 = SOME_STATIC;, there's no obvious borrow there.

There's no borrow at all there, obvious or otherwise.

src/items/constant-items.md
```diff
@@ -91,6 +91,32 @@ m!(const _: () = (););
 // const _: () = ();
 ```
 
+## Use and reference to `static` items
+
+When a constant item or constant block is defined, [`static` items] can be used, borrowed or taken address of.
+By extension, you are allowed to call methods that immutably borrows the `static` items as receivers.
```
RalfJung, line 97: Yeah this should definitely explain that it is not allowed to read from or write to any mutable static (`static mut`  and `static !Freeze`).

RalfJung:
> Constants may refer to the address of other constants...

That's a somewhat nonsensical statements as constants are values, they don't have an address.

src/items/constant-items.md
dingxiangfei2009: Do we already have a link into the reference to explain the `static !Freeze` items? It makes sense to me to make a reference here.

dingxiangfei2009: Oh this is largely missing. I guess I have to expand this section in `const_eval.md`. I can do that.

RalfJung:
I opened https://github.com/rust-lang/reference/pull/1624 to fix the part about "address of a constant".

dingxiangfei2009:
@rustbot ready

- I sieved through the document and I believe these are the places where use of `static` is mentioned.
- I added some sort of document about `Freeze` without mentioning it, given that it is not stabilized. In case rust-lang/rfcs#3633 lands and stabilization of the trait actually happens, I will follow up with the edits.

src/const_eval.md
```diff
@@ -67,6 +67,30 @@ surrounding generic parameters: such an expression must either be a single bare
 const generic parameter, or an arbitrary expression not making use of any
 generics.
 
+## Permitted use of `static`s
+
+Within the constant evaluation context, items that ultimately contain no mutable memory or data that is
```
ehuss, line 72: I'm struggling to follow this section. It seems like you are allowed to borrow static mut items, as in:

```rust
const fn f() -> &'static u32 {
    static mut VALUE: u32 = 0;
    unsafe { &VALUE }
}

fn main() {
    let y = const { f() };
}
```

I can't really tell why it is not allowed in the initializer of a `const` as shown below. What is the rule there, and why is it different?

(I'm probably very confused, but the example above also seems to go against what the stabilization report says.)

Additionally, the stabilization report seems to make a distinction of *resulting* in a value that contains a reference to something mutable. (As opposed to merely accessing mutable statics within the course of evaluation, which seems to be allowed?)

src/const_eval.md
```diff
@@ -67,6 +67,30 @@ surrounding generic parameters: such an expression must either be a single bare
...
+const fn allowed() -> &'static u32 {
+    static VALUE: u32 = 0;
+    &VALUE
+}
+const A_CONST: &'static u32 = allowed();
+```
+
+On the contrary, for example, `static mut` and types embedding an `UnsafeCell` is not allowed.
+
+```rust,edition2021,compile_fail
```
ehuss, line 86: `compile_fail` should usually have an E code to ensure the correct error is being emitted.

```suggestion
```rust,edition2021,compile_fail,E0080
```

ehuss (commented):
Is there anywhere that mentions the `extern static` restrictions mentioned in the stabilization report?

src/const_eval.md
```diff
@@ -23,7 +23,7 @@ to be run.
 * [Const parameters].
 * [Paths] to [functions] and [constants].
   Recursively defining constants is not allowed.
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to immutable [statics].
```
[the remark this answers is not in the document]
RalfJung: You can use the path but evaluation will fail. Any attempt to read from a mutable or interior mutable static is an error.

However, if this is dead code, it will get accepted. So this is different from the prior check that just rejected even syntactically mentioning statics.

src/const_eval.md
RalfJung: > It seems like you are allowed to borrow static mut items, as in:

Yes, but you can't read from them. (It's also questionable whether this may be UB, but probably not worth explicitly making UB.)

> I can't really tell why it is not allowed in the initializer of a const as shown below. What is the rule there, and why is it different?

It is not allowed if the reference ends up in the final value of the const *as a safely readable reference*. (If it ends up as a raw pointer, or just inside a union, it's fine.) This is because such consts can be used as patterns, and patterns depending on mutable data would be Bad News.

src/const_eval.md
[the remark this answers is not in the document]
ehuss: Oh, sorry, my testing is wrong because these errors don't happen with `cargo check` 😦 .

src/const_eval.md
ehuss: Yea, sorry, I deleted my comment since I was more confused than I thought, and misunderstood how it worked.

src/const_eval.md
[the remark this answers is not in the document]
RalfJung: `const { ... }` blocks don't get evaluated in `cargo check`, yeah. `const` *items* do.

dingxiangfei2009:
@ehuss I added a section on `extern { static }`.

ehuss:
I pushed a commit to simplify things. After reading more carefully, I was feeling like this was duplicating content that is already specified in the const_eval chapter. I'm reluctant to duplicate things since it makes it more challenging to maintain. I'm also a little reluctant to add deeper "exploration" style content which is essentially restating other rules. There is more specific reasoning in the commit message. If you disagree or think there is something that is now not documented, please let me know!

I realize this removes a lot of content from this PR, and I apologize for that. I probably should have been paying more attention earlier on and provided some guidance.

As for having more examples that illustrate specific const-eval limitations, I expect those to be added later over time (perhaps through the test suite links).

@traviscross Would you be willing to give this a final look? I think this is ready to go from my perspective.

src/items/static-items.md
```diff
@@ -22,11 +22,8 @@ Static initializers may refer to other statics.
 Non-`mut` static items that contain a type that is not [interior mutable] may
 be placed in read-only memory.
 
-All access to a static is safe, but there are a number of restrictions on
-statics:
-
-* The type must have the `Sync` trait bound to allow thread-safe access.
-* Constants cannot refer to statics.
+All access to a static is safe,
+provided that the type must have the `Sync` trait bound to allow thread-safe access.
```
RalfJung, line 26: This makes it sound like a `!Sync` static would be unsafe to access. But actually it's just rejected.

src/const_eval.md
```diff
@@ -23,7 +23,8 @@ to be run.
...
+    * Reads of [`extern` statics] are not allowed.
```
RalfJung, line 27: Reads from and writes to `static mut` and `static` with interior mutability are also not allowed.

And then there are restrictions on references that escape into the final value of the const/static. More precisely, this is about expressions that have been subject to lifetime extension, also see https://github.com/rust-lang/reference/pull/1637. Those references must also be of type `&T` and point to a value without an `UnsafeCell`.

And finally, specifically for `const` there also can't be any references to anything mutable in the final value.

src/items/static-items.md
ehuss: Thanks! I have rewritten this to use more direct language. I'm not sure why the original was written to say "it is safe to ...", since there isn't anything to imply that it was unsafe.

src/const_eval.md
ehuss: Hm, I'm trying to unpack what you are saying here and align that with the current rules.

> Reads from and writes to `static mut` …

Isn't this implied by the fact that you are not allowed to refer to `static mut` at all by the rule that says paths only to *immutable* statics are allowed?

> Reads from and writes to [..] `static` with interior mutability are also not allowed.

Thanks, I have added that.

> And then there are restrictions on references that escape into the final value of the const/static.

Isn't this specified in the later rule that limits borrows of `UnsafeCell` that are transient? (That was my reading of it, and why I didn't say it explicitly.)

> And finally, specifically for `const` there also can't be any references to anything mutable in the final value.

Thanks, I have added that to the constant-items chapter.

dingxiangfei2009: I actually have found the points mentioned down in the list here, in a bullet point starting with `All forms of [borrow]s, including raw borrows, with one limitation ...`. I think we have it covered.

src/items/static-items.md
dingxiangfei2009: I fixed the wording, to mention that a type with `!Sync` is rejected.

dingxiangfei2009: I picked the changeset from @ehuss actually. I think those wordings are better.

src/const_eval.md
RalfJung: > Isn't this specified in the later rule that limits borrows of UnsafeCell that are transient? (That was my reading of it, and why I didn't say it explicitly.)

Ah right, I forgot that those are already documented.

ehuss: > That's not correct, referring to static mut is permitted.

OK, then I go back to my comment [earlier](https://github.com/rust-lang/reference/pull/1610#discussion_r1779733601) about why the word "immutable" is being added here.

I would think this PR shouldn't add the word *immutable*, and the edit @dingxiangfei2009 just added about not being allowed to read or write from `static mut` should cover everything?

(Because as written now, it says you can only do paths to immutable statics, but that does not seem to be true based on the playground above.)

RalfJung: Yeah you are right, `immutable` should not be mentioned here.

src/const_eval.md
```diff
@@ -23,7 +23,10 @@ to be run.
 * [Const parameters].
 * [Paths] to [functions] and [constants].
   Recursively defining constants is not allowed.
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to immutable [statics] with these exception with these restrictions.
```
ehuss, line 26:
```suggestion
* Paths to [statics] with the following restrictions:
```

ehuss (commented):
I think this should be the last edit, do you agree?

src/const_eval.md
```diff
@@ -23,7 +23,10 @@ to be run.
...
+    * Reads out of and writes into [`extern` statics] are not allowed.
+    * Reads out of and writes into either a `static` with data equipped with interior mutability,
+      or a whole or parts of `static mut`, are not allowed.
```
RalfJung, line 29: I am confused by the wording here. Why does `static mut` emphasize this "whole or part of", but interior mutable `static` does not? The two are treated exactly the same so they should also be described the same.

The actual underlying check is simply "is this global mutable memory -- if yes, reject read", and "if this memory global -- if yes, reject write".

chorman0773: Is this code accepted?
```rust
static FOO: (u8, UnsafeCell<u8>) = (0, UnsafeCell::new(1));

const BAR: u8 = FOO.0;
```

traviscross:
@rustbot author

We reviewed this in the call today.  This still needs the edits about to cover the case of accesses to mutable statics being allowed.

src/const_eval.md
```diff
@@ -23,7 +23,10 @@ to be run.
 * [Const parameters].
 * [Paths] to [functions] and [constants].
   Recursively defining constants is not allowed.
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to immutable [statics] with these restrictions and observations.
```
ehuss, line 26: I think this might still need to be updated, since it allows `static mut`.

```suggestion
* Paths to [statics] with these restrictions and observations:
```

dingxiangfei2009: Applied

dingxiangfei2009:
@rustbot ready

- Rebased with merge conflict resolutions
- Clarify the accepted use of mutable `static` items.

ehuss:
@RalfJung How do you feel about the final wording here?

src/const_eval.md
```diff
@@ -42,7 +42,10 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions and observations.
+    * In particular, reads and writes to any `static`, `static mut` or [`extern` statics] is not allowed.
```
RalfJung, line 46: Reads from immutable statics are allowed.

src/const_eval.md
```diff
@@ -42,7 +42,10 @@ r[const-eval.const-expr.path-item]
...
+    * Immutable borrows and pointers into immutable part of a `static` are allowed and observes the same restriction
+      on all other forms of [borrow]s as mentioned below.
```
RalfJung, line 48: This doesn't quite parse for me. I'm also not sure what the point of this bullet is -- is there any reason to not just entirely remove it?

src/const_eval.md
```diff
@@ -42,7 +42,8 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions and observations.
+  * In particular, reads and writes to any `static mut` or [`extern` statics] is not allowed.
```
traviscross, line 46: So over in #1657, there is proposed language:

> Static initializers may refer to and even read from other statics.
> When reading from mutable statics, they read the initial value of that static.

The language in this PR seems contradictory to the language in that one, no?

RalfJung: Yeah this already works on stable
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=59b79563db4fe5d5df26ab3f27cd4f1c

ehuss: @RalfJung Can you write down exactly what we should say here? It would be helpful since we don't know exactly what the rules are. From looking at the implementation, I'm guessing, is it the `CanAccessMutGlobal` setting, which appears to be set only when in a static initializer? Does this need to say that you cannot read from `static mut` unless inside a static initializer, and that you can never write to a `static mut`, and that reading *and* writing `extern` statics always disallowed?

RalfJung: Yes:
- reads from mutable statics (`static mut`, and interior mutable `static`) are forbidden exception inside the initializer expression of a `static`/`static mut`
- writes to mutable statics are forbidden everywhere
- reads from and writes to extern static are forbidden everywhere

dingxiangfei2009: I have updated the text just now, so that the cases for the two evaluation contexts are clearer.

src/items/constant-items.md
```diff
@@ -48,6 +48,8 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
 };
 ```
 
+The final value of a `const` item cannot contain references to anything mutable.
```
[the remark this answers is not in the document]
dingxiangfei2009: Applied

src/const_eval.md
```diff
@@ -42,7 +42,9 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions and observations.
+  * If the evaluation is carried out in an initializer of a `const` item, or inside a `const` block, reads and writes to any `static mut` or [`extern` statics] are not allowed.
```
RalfJung, line 46: This is missing some things, like array lengths or enum discriminants. I wouldn't try to list them all. Instead, I would say "If the evaluation is *not* carried out in an initializer of a `static` item, ...".

Also, saying `static mut` is wrong since it fails to mention interior mutability. By summary carefully mentioned interior mutability. Please ensure your text matches my [summary](https://github.com/rust-lang/reference/pull/1610#discussion_r1830928182).

```suggestion
  * If the evaluation is *not* carried out in an initializer of a `static` item, then reads and writes to any mutable static (`static mut` or `static` with an interior mutable type) or [`extern` statics] are not allowed.
```

src/const_eval.md
```diff
@@ -42,7 +42,9 @@ r[const-eval.const-expr.path-item]
...
+  * If the evaluation is carried out in an initializer of a `static` item, reads into a `static`, or inside `unsafe` blocks reads into `static mut` are allowed. Reads and writes to `extern` statics are still not allowed.
```
RalfJung, line 47:
```suggestion
  * If the evaluation is carried out in an initializer of a `static` item, *writes* to a mutable `static` are not allowed, but reads are allowed. Reads and writes to `extern` statics are still not allowed.
```

src/const_eval.md
dingxiangfei2009: Applied with edits. I am also reordering the bullets around.

src/const_eval.md
dingxiangfei2009: Applied

src/const_eval.md
```diff
@@ -42,7 +42,10 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions and observations.
+  * Reads and writes to `extern` statics are not allowed in any constant evaluation context.
+  * If the evaluation is carried out in an initializer of a `static` item, then *writes* into mutable `static` are not allowed, such as `static mut`s and `static`s containing an interior-mutable type.
+  * If the evaluation is *not* carried out in an initializer of a `static` item, then reads and writes to any mutable `static`, such as `static mut`s and `static`s containing an interior-mutable type.
```
ehuss, line 48: Something (a verb?) seems to have been lost in this sentence. Maybe this?

```suggestion
  * If the evaluation is *not* carried out in an initializer of a `static` item, then reads and writes to any mutable `static`, such as `static mut`s and `static`s containing an interior-mutable type, are not allowed.
```

src/const_eval.md
```diff
@@ -43,8 +43,9 @@ r[const-eval.const-expr.path-item]
 
 r[const-eval.const-expr.path-static]
 * Paths to [statics] with these restrictions and observations.
-  * If the evaluation is carried out in an initializer of a `const` item, or inside a `const` block, reads and writes to any `static mut` or [`extern` statics] are not allowed.
-  * If the evaluation is carried out in an initializer of a `static` item, reads into a `static`, or inside `unsafe` blocks reads into `static mut` are allowed. Reads and writes to `extern` statics are still not allowed.
+  * Reads and writes to `extern` statics are not allowed in any constant evaluation context.
+  * If the evaluation is carried out in an initializer of a `static` item, then *writes* into mutable `static` are not allowed, such as `static mut`s and `static`s containing an interior-mutable type.
```
RalfJung, line 47:
```suggestion
  * Writes to `static` items are not allowed in any constant evaluation context.
  * Reads from `extern` statics are not allowed in any constant evaluation context.
```
I think that is simpler than what you have.

src/const_eval.md
RalfJung: With writes being handled separately by my suggestion above, this can just say "reads from any mutable `static".

src/const_eval.md
```diff
@@ -43,9 +43,11 @@ r[const-eval.const-expr.path-item]
...
 r[const-eval.const-expr.path-static]
 * Paths to [statics] with these restrictions and observations.
-  * Reads and writes to `extern` statics are not allowed in any constant evaluation context.
-  * If the evaluation is carried out in an initializer of a `static` item, then *writes* into mutable `static` are not allowed, such as `static mut`s and `static`s containing an interior-mutable type.
-  * If the evaluation is *not* carried out in an initializer of a `static` item, then reads and writes to any mutable `static`, such as `static mut`s and `static`s containing an interior-mutable type.
+  * Writes to `static` items are not allowed in any constant evaluation context.
+  * Reads from `extern` statics are not allowed in any constant evaluation context.
+  * If the evaluation is *not* carried out in an initializer of a `static` item,
+  then reads from any mutable `static` are not allowed.
+  By mutable `static` it includes `static mut`s and `static`s containing an interior-mutable type.
```
RalfJung, line 50:
```suggestion
  * If the evaluation is *not* carried out in an initializer of a `static` item,
    then reads from any mutable `static` are not allowed.
    A mutable `static` is a `static mut` item, or a `static` item with an interior-mutable type.
```

src/const_eval.md
```diff
@@ -42,7 +42,12 @@ r[const-eval.const-expr.path-item]
   Recursively defining constants is not allowed.
 
 r[const-eval.const-expr.path-static]
-* Paths to [statics]. These are only allowed within the initializer of a static.
+* Paths to [statics] with these restrictions and observations.
```
RalfJung, line 45:
```suggestion
* Paths to [statics] with these restrictions:
```
Sorry, just noticed this now. What are the "observations"?

src/const_eval.md
```diff
@@ -42,7 +42,12 @@ r[const-eval.const-expr.path-item]
...
+  * Writes to `static` items are not allowed in any constant evaluation context.
+  * Reads from `extern` statics are not allowed in any constant evaluation context.
+  * If the evaluation is *not* carried out in an initializer of a `static` item,
+    then reads from any mutable `static` are not allowed.
+    A mutable `static` is a `static mut` item, or a `static` item with an interior-mutable type.
```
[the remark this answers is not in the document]
dingxiangfei2009: Applied with edits.

traviscross (approved):

src/const_eval.md
RalfJung: I had a `:` in my suggestion here, which you turned back into a `.` for some reason -- seems odd to introduce a list with a `.` like this. It's not a big deal, I just wonder why you didn't simply click the button to apply my suggestion.

src/const_eval.md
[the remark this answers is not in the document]
RalfJung: Ah unfortunately you copied my typo.^^ "accessed"...

src/const_eval.md
dingxiangfei2009: Oh sorry. I did not notice that there was a change suggestion when reading the email. I would have just switch to colon.

RalfJung:
Ah right they all got fixed, great. Thanks all!

PR #1614, opened 2024-09-16
Add identifier syntax to type-layout.md

chorman0773:


src/type-layout.md
```diff
@@ -75,49 +93,70 @@ at least equal to the size and alignment of a pointer.
...
+r[layout.trait-object]
+
 Trait objects have the same layout as the value the trait object is of.
 
 > Note: This is about the raw trait object types, not pointers (`&dyn Trait`,
 > `Box<dyn Trait>`, etc.) to trait objects.
 
 ## Closure Layout
 
+r[kayout.closure]
```
ehuss, line 139:
```suggestion
r[layout.closure]
```

src/type-layout.md
```diff
@@ -199,13 +252,18 @@ layout such as reinterpreting values as a different type.
...
 
+r[layout.repr.c.constraint]
 This representation can be applied to structs, unions, and enums. The exception
 is [zero-variant enums] for which the `C` representation is an error.
 
 #### `#[repr(C)]` Structs
 
+r[layout.repr.c.struct]
+
+r[layour.repr.c.struct.align]
```
ehuss, line 263:
```suggestion
r[layout.repr.c.struct.align]
```

src/type-layout.md
```diff
@@ -374,24 +445,32 @@ struct MyDFields;
...
+r[layout.repr.primitive.constraint]
 Primitive representations can only be applied to enumerations and have
 different behavior whether the enum has fields or no fields. It is an error
 for [zero-variant enums] to have a primitive representation. Combining
 two primitive representations together is an error.
 
 #### Primitive Representation of Field-less Enums
 
+r[layout.repr.primitive.enum]
+
```
ehuss, line 464: Throughout the changes here it seems to be inconsistent about whether or not a blank line is added after the rule.

I would say the predominate style seems to not include blank lines. Can we stay consistent and avoid those extra blank lines?

chorman0773: The blank line here is because it's a section rule-id not a paragraph rule-id. There's no paragraph rule-id because the current text would only have an intro paragraph. This style (blank line after rules for sections, and omitting a lone `.intro` id) has been common throughout the PRs.

JoelMarcey: Could the section rule-id go above the main header?

ehuss: OK, I'm still not sure I agree with this, since the paragraph below does not look like an introduction to me. It seems to define the specific layout of these types of enums. I'm not going to block here, but I have opened #1651 to further discuss how these things interact.

ehuss (approved):
Thanks!

PR #1622, opened 2024-09-20
Add trait_upcasting related languages changes

WaffleLapkin:
I applied review comments from #1259 and added updates from https://github.com/rust-lang/rust/pull/120248.

Tracking issue: https://github.com/rust-lang/rust/issues/65991
Stabilization proposal: https://github.com/rust-lang/rust/pull/134367

cc @crlf0710 @compiler-errors @RalfJung

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -476,7 +476,7 @@ reference types and `mut` or `const` in pointer types.
 
 * Both slice metadata (`*[u16]` -> `*[u8]`, `*str` -> `*(u8, [u32])`).
 * Both the same trait object metadata, modulo dropping auto traits (`*dyn Debug` -> `*(u16, dyn Debug)`, `*dyn Debug + Send` -> `*dyn Debug`).
-    * **Note**: Adding auto traits is not allowed (`*dyn Debug` -> `*dyn Debug + Send` is invalid).
+    * **Note**: Adding auto traits is only allowed if the principal trait has the auto trait as a super trait (given `trait T: Send {}`, `*dyn T` -> `*dyn T + Send` is valid, but `*dyn Debug` -> `*dyn Debug + Send` is not).
     * **Note**: Generics (including lifetimes) must match (`*dyn T<'a, A>` -> `*dyn T<'b, B>` requires `'a = 'b` and `A = B`).
 
 \*\* only when `m₁` is `mut` or `m₂` is `const`. Casting `mut` reference to
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -472,7 +472,12 @@ reference types and `mut` or `const` in pointer types.
 | [Function pointer]    | Integer               | Function pointer to address cast |
 | Closure \*\*\*        | Function pointer      | Closure to function pointer cast |
 
-\* or `T` and `V` are compatible unsized types, e.g., both slices, both the same trait object.
+\* or `T` and `V` are unsized types with compatible metadata:
+
+* Both slice metadata (`*[u16]` -> `*[u8]`, `*str` -> `*(u8, [u32])`).
+* Both the same trait object metadata, modulo dropping auto traits (`*dyn Debug` -> `*(u16, dyn Debug)`, `*dyn Debug + Send` -> `*dyn Debug`).
+    * **Note**: Adding auto traits is not allowed (`*dyn Debug` -> `*dyn Debug + Send` is invalid).
+    * **Note**: Generics (including lifetimes) must match (`*dyn T<'a, A>` -> `*dyn T<'b, B>` requires `'a = 'b` and `A = B`).
 
 \*\* only when `m₁` is `mut` or `m₂` is `const`. Casting `mut` reference to
 `const` pointer is allowed.
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -191,7 +191,7 @@ r[coerce.unsize]
 
 r[coerce.unsize.intro]
 The following coercions are called `unsized coercions`, since they
-relate to converting sized types to unsized types, and are permitted in a few
+relate to converting types to unsized types, and are permitted in a few
 cases where other coercions are not, as described above. They can still happen
 anywhere else a coercion can occur.
 
@@ -206,6 +206,9 @@ r[coerce.unsize.slice]
 
 r[coerce.unsize.trait-object]
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
+
+r[coerce.unsize.trait-upcast]
+* `dyn T` to `dyn U`, when `T` has `U` as one of its ancestor trait.
 
 r[coerce.unsized.composite]
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -210,6 +210,7 @@ r[coerce.unsize.trait-object]
 r[coerce.unsize.trait-upcast]
 * `dyn T` to `dyn U`, when `U` is one of `T`'s [supertraits].
     * This allows dropping auto traits, i.e. `dyn T + Auto` to `dyn U` is allowed.
+    * This allows adding auto traits if the principal trait has the auto trait as a super trait, i.e. given `trait T: U + Send {}`, `dyn T` to `dyn T + Send` or to `dyn U + Send` coercions are allowed.
 
 r[coerce.unsized.composite]
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -208,7 +208,8 @@ r[coerce.unsize.trait-object]
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
 
 r[coerce.unsize.trait-upcast]
-* `dyn T` to `dyn U`, when `T` has `U` as one of its ancestor trait.
+* `dyn T` to `dyn U`, when `U` is one of `T`'s [supertraits].
+    * This allows dropping auto traits, i.e. `dyn T + Auto` to `dyn U` is allowed.
 
 r[coerce.unsized.composite]
 * `Foo<..., T, ...>` to `Foo<..., U, ...>`, when:
@@ -325,3 +326,4 @@ precisely.
 [`Unsize`]: std::marker::Unsize
 [`CoerceUnsized`]: std::ops::CoerceUnsized
 [method-call expressions]: expressions/method-call-expr.md
+[supertraits]: items/traits.md#supertraits


WaffleLapkin:
https://github.com/rust-lang/rust/issues/101336 is still open, did we actually end up committing to something?

compiler-errors:
@WaffleLapkin: Yes, it's just pending documentation https://github.com/rust-lang/rust/issues/101336#issuecomment-1251216579

src/type-coercions.md
```diff
@@ -172,6 +175,9 @@ an implementation of `Unsize<U>` for `T` will be provided:
 
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [object safe].
 
+* `dyn T` to `dyn U` when `U` is one of `T`'s [supertraits].
+    * This allows dropping auto traits, i.e. `dyn T + Auto` to `dyn U` is allowed.
```
compiler-errors, line 179: This also allows adding auto traits, i.e. `trait Sub: Sup + Auto` you can do `dyn Sub` -> `dyn Sup + Auto`. https://github.com/rust-lang/rust/pull/119338

src/expressions/operator-expr.md
```diff
@@ -385,7 +385,11 @@ reference types and `mut` or `const` in pointer types.
 | [Function pointer]    | Integer               | Function pointer to address cast |
 | Closure \*\*\*        | Function pointer      | Closure to function pointer cast |
 
-\* or `T` and `V` are compatible unsized types, e.g., both slices, both the same trait object.
+\* or `T` and `V` are unsized types with compatible metadata:
+* both slice metadata (`*[u16]` -> `*[u8]`, `*str` -> `*(u8, [u32])`).
+* both the same trait object metadata, modulo dropping auto traits (`*dyn Debug` -> `*(u16, dyn Debug)`, `*dyn Debug + Send` -> `*dyn Debug`).
+    * **Note**: *adding* auto traits is not allowed (`*dyn Debug` -> `*dyn Debug + Send` is invalid).
```
compiler-errors, line 391: Unless it's a supertrait of the principal trait: https://github.com/rust-lang/rust/pull/119338

compiler-errors (commented):
You may also want to edit https://doc.rust-lang.org/reference/behavior-considered-undefined.html as Niko mentioned in the issue you linked; you may want to re-read the FCP, but I believe it's UB to conjure up a vtable now since it must be valid for upcasting.

RalfJung:
> You may also want to edit https://doc.rust-lang.org/reference/behavior-considered-undefined.html as Niko mentioned in the issue you linked; you may want to re-read the FCP, but I believe it's UB to conjure up a vtable now since it must be valid for upcasting.

What exactly are you referring to here? I assume the "issue" is https://github.com/rust-lang/rust/issues/101336 (a link would have been good, it took clicking on 4 links to track this down :D ). Niko says a lot of things there, though. :)

WaffleLapkin:
@compiler-errors [doc.rust-lang.org/reference/behavior-considered-undefined.html](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) already mentions invalid metadata in any kind of pointer makes an invalid value, so I don't think we need to add anything on top of that?

> - Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement for the metadata is determined by the type of the unsized tail:
>    - `dyn Trait` metadata is invalid if it is not a pointer to a vtable for `Trait`.
>    - `Slice` (`[T]`) metadata is invalid if the length is not a valid `usize` (i.e., it must not be read from uninitialized memory). Furthermore, for wide references and `Box<T>`, slice metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.

src/type-coercions.md
WaffleLapkin: Added a mention of that

src/expressions/operator-expr.md
WaffleLapkin: Added a mention of that

compiler-errors (approved):
I believe this is ready and consistent with what is implemented on master. I think all that needs to be done w.r.t. trait upcasting is a fresh stabilization PR?

PoignardAzur:
Any movement on this?

WaffleLapkin:
@PoignardAzur I'm working on a stabilization PR (after which this can be merged)

WaffleLapkin:
@compiler-errors could you take another look, do the changes still make sense?

src/type-coercions.md
```diff
@@ -211,6 +215,13 @@ r[coerce.unsize.slice]
 r[coerce.unsize.trait-object]
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
 
+r[coerce.unsize.trait-upcast]
+* `dyn T` to `dyn U`, when `U` is one of `T`'s supertraits.
+* `dyn T` to `dyn U` when `U` is one of `T`'s [supertraits].
```
traviscross, line 220: These lines seem duplicates.

traviscross:
In the original PR for this...

- https://github.com/rust-lang/reference/pull/1259

...there's a [discussion](https://github.com/rust-lang/reference/pull/1259/files#r1404647526) about adding a line to the behavior considered undefined of:

> - Performing non-nop coercion on a dangling or unaligned raw pointer.

There was some back and forth and it's not clear where that landed.  Thoughts on that?

@compiler-errors @RalfJung @WaffleLapkin

compiler-errors:
The comment here is the answer https://github.com/rust-lang/reference/pull/1259/files#r1404804424. I don't think this needs further changing. It doesn't really matter if the pointer is bad; upcasting only cares about the vtable metadata.

src/type-coercions.md
```diff
@@ -195,10 +195,14 @@ r[coerce.unsize]
...
 r[coerce.unsize.intro]
 The following coercions are called `unsized coercions`, since they
-relate to converting sized types to unsized types, and are permitted in a few
+relate to converting types to unsized types, and are permitted in a few
 cases where other coercions are not, as described above. They can still happen
 anywhere else a coercion can occur.
 
+r[coerce.unsize.misnomer]
+> Note: "unsizing" is a bit of a misnomer,
+> since this covers unsized->unsized coercions too.
```
traviscross, line 204:
```suggestion
> Note: These are sometimes called "unsizing" coercions, but that can be a bit of a misnomer, since these cover unsized->unsized coercions too.
```

Not sure if something like this is what you meant to say or not.  The Reference doesn't refer to "unsizing" anywhere -- even though, you're right, we often do in conversation -- so we'd need to introduce that somehow in this note.

src/expressions/operator-expr.md
```diff
@@ -385,7 +385,12 @@ reference types and `mut` or `const` in pointer types.
 | [Function pointer]    | Integer               | Function pointer to address cast |
 | Closure \*\*\*        | Function pointer      | Closure to function pointer cast |
 
-\* or `T` and `V` are compatible unsized types, e.g., both slices, both the same trait object.
+\* or `T` and `V` are unsized types with compatible metadata:
+* both slice metadata (`*[u16]` -> `*[u8]`, `*str` -> `*(u8, [u32])`).
+* both the same trait object metadata, modulo dropping auto traits (`*dyn Debug` -> `*(u16, dyn Debug)`, `*dyn Debug + Send` -> `*dyn Debug`).
+    * **Note**: *adding* auto traits is only allowed if the principal trait has the auto trait as a super trait
```
traviscross, line 391:
```suggestion
    * **Note**: *adding* auto traits is only allowed if the principal trait has the auto trait as a super trait.
```

src/type-coercions.md
```diff
@@ -195,10 +195,14 @@ r[coerce.unsize]
 
 r[coerce.unsize.intro]
 The following coercions are called `unsized coercions`, since they
-relate to converting sized types to unsized types, and are permitted in a few
+relate to converting types to unsized types, and are permitted in a few
```
nikomatsakis, line 198:
```suggestion
relate to converting sized types to unsized types or converting from one unsized type to another, and are permitted in a few
```

(more clarity, same meaning)

src/expressions/operator-expr.md
```diff
@@ -385,7 +385,12 @@ reference types and `mut` or `const` in pointer types.
...
+      (given `trait T: Send {}`, `*dyn T` -> `*dyn T + Send` is valid, but `*dyn Debug` -> `*dyn Debug + Send` is not).
+    * **Note**: generics (including lifetimes) must match (`*dyn T<'a, A>` -> `*dyn T<'b, B>` requires `'a = 'b` and `A = B`).
```
nikomatsakis, line 393: Does `as` not permit upcasting?

nikomatsakis (dismissed):

nikomatsakis:
> @compiler-errors [doc.rust-lang.org/reference/behavior-considered-undefined.html](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) already mentions invalid metadata in any kind of pointer makes an invalid value, so I don't think we need to add anything on top of that?
> 
> > * Invalid metadata in a wide reference, `Box<T>`, or raw pointer. The requirement for the metadata is determined by the type of the unsized tail:
> >   
> >   * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for `Trait`.
> >   * `Slice` (`[T]`) metadata is invalid if the length is not a valid `usize` (i.e., it must not be read from uninitialized memory). Furthermore, for wide references and `Box<T>`, slice metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.

@WaffleLapkin I think this isn't *quite* right. How I read the [Behaviors Considered Undefined](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) section, it says...

> Producing an [invalid value](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#invalid-values). “Producing” a value happens any time a value is assigned to or read from a place, passed to a function/primitive operation or returned from a function/primitive operation.

where ["producing an invalid value is hence immediate UB"](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#invalid-values). But the consensus from https://github.com/rust-lang/rust/issues/101336 was that having an invalid vtable is not a failure of the **validity invariant**. Rather, the UB occurs when a `dyn` pointer (raw, reference, whatever) with a malformed vtable is upcast or has a method invoked on it. Because safe code can perform those operations (including on raw pointers, in the case of upcast and potentially method calls in the future), this implies that the **safety invariant** (the conditions required to release such a value to safe code) requires a valid vtable. But if you can keep the `*const dyn Foo` confined to unsafe code and be absolutely sure nobody upcasts it, it doesn't require a valid vtable. Quoting https://github.com/rust-lang/rust/issues/101336:

> The proposal is as follows:
>
> * Vtable-adjusting upcasts are **safe operations**. The upcast is UB if performed on a value without a valid vtable
> * As such, the "safety invariant" requires a **fully valid vtable**.
> * The "validity invariant" requires `*dyn` metadata to be **word-aligned and non-null**.
>
> Vtable-adjusting upcasts are defined as:
>
> * Trait upcasts that alter the set of methods that can be invoked on the resulting value at runtime (e.g., `dyn Bar` to `dyn Foo` from the introduction). In particular, upcasts that simply add or remove auto-traits are not vtable-adjusting (e.g., `dyn Debug + Send` to `dyn Debug`).

RalfJung:
> But the consensus from https://github.com/rust-lang/rust/issues/101336 was that having an invalid vtable is not a failure of the validity invariant.

To be clear, consensus was that we are not ruling on the validity invariant at this point. The validity invariant for vtable metadata is an open question, discussed at https://github.com/rust-lang/unsafe-code-guidelines/issues/516. Meanwhile, the reference says that the vtable pointer must point to a vtable of the right trait and Miri enforces this; this is following our usual strategy of having a bit more UB for now and possibly relaxing this in the future.

nikomatsakis:
Hmm, OK, that's not what I thought the consensus was. =) Maybe I'm misremembering, though when I read the text I quoted it seems clear-ish. We need to find some clearer way though to describe the "upper/lower bounds" of what is and is not UB, I think it's going to be very hard to keep things in sync over time the way it is now (where it is e.g. difficult to tell if the reference is out of date or over approximating).

src/expressions/operator-expr.md
WaffleLapkin: `as` permits upcasting.

src/type-coercions.md
WaffleLapkin: Hm. There may not be the literal word "unsizing", but we do refer to this coercion as "unsize coercion". 

> The following coercions are called `unsized coercions`, since they

I do feel like it's clear what this means, given that this note is surrounded by mentions of unsize coercions. (but maybe not see you seem to be confused?)

src/type-coercions.md
WaffleLapkin: For me your proposed change kinda flushes the cache, I had to re-read it 4 times before I figured out what it means 💀 (then again, maybe I shouldn't be reading it at 3am)

Since it's the same meaning I'd think that more concise is better? Do you disagree?

src/type-coercions.md
WaffleLapkin: Although I'd also say that replacing "unsized coercion" to "unsizing coercion" everywhere in the reference would be a good idea. It sounds better and more closely matches how people actually talk about this feature.

src/type-coercions.md
```diff
@@ -211,6 +215,13 @@ r[coerce.unsize.slice]
 r[coerce.unsize.trait-object]
 * `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
 
+r[coerce.unsize.trait-upcast]
```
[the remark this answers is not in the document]
WaffleLapkin: I think this whole section might need a retractor :p

Specifically I think it should more clearly state the difference between `Unsize` and `CoerceUnsize`, mention that the former is not actually a coercion, and the latter is only for pointers. 

Then it would also make sense to add a note that unsizing coercion changes the metadata of a pointer, while all other coercions/casts don't.

But, if you don't mind I'll leave this as a follow up ^^'

src/type-coercions.md
ehuss: In conversation with @traviscross, I was one of the confused ones. TC pointed out that "unsized coercion" can be interpreted as the codomain so to speak. That is, it doesn't seem like a misnomer to me because it is only referring to coercions *to* something not sized, and thus seemed accurate.

It is true the word "unsizing" implies "to remove the size of", that is the *domain* is sized, and thus is not always an accurate term. However, we don't use "unsizing" here.

Probably a bit of a muddled interpretation on my part, but does that make sense?

RalfJung: "unsized coercion" sounds like the coercion itself is unsized. I don't even know what that should mean.

RalfJung:
We currently say this above the UB list

> The following list is not exhaustive; it may grow or shrink. 

So the entire thing should be considered as approximating. Maybe we should start slowly widdling this down to just a few items in the list, though I think we should always reserve the right to remove UB.

Anyway this is getting off-topic for this PR. The validity invariant for vtables doesn't have to be settled for this one.

src/expressions/operator-expr.md
WaffleLapkin: Why is the dot needed here? The dot is currently after the parenthesis. Should it be moved?

traviscross: Got it, thanks; didn't see the continuation after the line break.

traviscross:
@rustbot labels +I-lang-nominated

When we handled the nomination for starting the FCP on #134367, we reviewed this text, leading to some back and forth.  Let's nominate to double check we're OK with where that landed.

WaffleLapkin:
As a note I'm working on rewriting most of `src/type-coercions.md`, when reading it carefully I found a lot of outdated/incorrect/misleading info. I'd like to land that as a follow up to this.

nikomatsakis:
OK, so I re-read https://github.com/rust-lang/rust/issues/101336 and in particular found [this comment](https://github.com/rust-lang/rust/issues/101336#issuecomment-1251216579) which was asking whether the validity invariant for vtables was a "decision" or just a "strong recommendation". I guess @RalfJung interpreted it as the latter, which seems fine. I have edited the issue to clarify. I agree then that no edits are needed to the list of UB, as no new validity invariants have been decided.

nikomatsakis:
Wearing lang-team hat: +1

src/expressions/operator-expr.md
nikomatsakis: To clarify my question, this text as written suggests `as` is only permitted between the same trait modulo auto-traits. But I think as is more general now. Shouldn't this test say something like

* `T` is a subtrait of `V` with compatible metadata, e.g., 
  * `*dyn (Debug + Send) -> `*dyn Debug`
  * `*dyn Subtrait -> `*dyn Supertrait`

or perhaps simply refer to the definition of upcasts that is given in the unsizing section?

WaffleLapkin: This section only explains \[lacking a better word\] _proper_ pointer casts. The reason why you can `*dyn Sub -> *dyn Super` via an `as` cast is that there is a _coercion_ `*dyn Sub -> *dyn Super` and `as` casts can be used to explicitly trigger an implicit coercion. The coercion is described in `type-coercions.md`, so I think everything is good on this front.

We might add a note about this though, since it might be confusing otherwise.

Also, I feel like this does not precisely capture the behavior wrt wrapper types and dropping the principal trait, I'll have to fix that.

traviscross (approved):

WaffleLapkin:
@rustbot label -I-lang-nominated

I've also prepared a refactor of the unsizing section: de3ffc3e19c141ac7d6cf08beba4545a3f4e0a9f. I'll open a PR once this is merged.

WaffleLapkin:
https://github.com/rust-lang/rust/pull/134367#event-16230091099 has been merged, so this can be merged as well :)

src/type-coercions.md
WaffleLapkin: @traviscross you've applied the suggestion without addressing my concern, so I reverted the commit (5dbb2de1f49758d50eeae21865fe75148c88b15f). I don't think this change improves this paragraph -- it changes a general statements into a union of concrete, which just makes it harder to read. Moreover, my refactor of the unsizing section also changes this paragraph anyway, so I'd wait for the refactor PR to be open before trying to fix anything.

traviscross: I had talked with Niko awhile back, and he had wanted to keep it, and it seems clearer to me as well.  What if were to just say both, e.g.?:

```suggestion
relate to converting types to unsized types (that is, converting sized types to unsized types or converting from one unsized type to another), and are permitted in a few
```

WaffleLapkin: That might be fine, but again, my refactor already contains similar explanations, so I'd suggest discussing this once it's open.

src/type-coercions.md
traviscross: I'm just removing this note section for now.  We can handle this in later work.

src/type-coercions.md
traviscross: OK.  We'll handle this is your later PR.

src/expressions/operator-expr.md
traviscross: > Also, I feel like this does not precisely capture the behavior wrt wrapper types and dropping the principal trait, I'll have to fix that.

I gather you want to fix this in the later PR you've mentioned, which is fine.

src/type-coercions.md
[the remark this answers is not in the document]
traviscross: OK, we'll leave this for the follow-up PR you have in progress.

traviscross:
I rebased to collapse some back and forth, to make some minor stylistic tweaks (capitalization and line unwrapping), and to remove the note section mentioned above.

WaffleLapkin:
Follow-up: #1731

PR #1625, opened 2024-09-23
Add spec identifier syntax to macro subchapters

chorman0773:
This adds identifiers to macros-by-example.md and proc-macro.md.

macro-ambiguity.md is intentionally excluded due to complexity.


ehuss (approved):
Thanks!

PR #1628, opened 2024-09-23
`coverage` attribute

clarfonthey:
Since rust-lang/rust#84605 passed FCP, this adds the `coverage` attribute to the reference. Coverage instrumentation isn't mentioned in the reference elsewhere, as most of that information lies in the `rustc` book, but it feels like it should at least go here.


clarfonthey:
Somehow this ended up not having anyone assigned for review:

r? lang

clarfonthey:
(no longer waiting on stabilisation; can be merged now)

ehuss (approved):
Thanks!

Issue #1616, opened 2024-09-18
Clarify "immutable bytes" definition


RalfJung:
Currently, we define "immutable bytes" as

> All bytes inside a [const](https://doc.rust-lang.org/nightly/reference/items/constant-items.html) item or within an implicitly [const-promoted](https://doc.rust-lang.org/nightly/reference/destructors.html#constant-promotion) expression are immutable. The bytes owned by an immutable binding or immutable static are immutable, unless those bytes are part of an [UnsafeCell<U>](https://doc.rust-lang.org/nightly/core/cell/struct.UnsafeCell.html).

However, there are some issue with this. First of all, a const item can only be used as a value, not as a place, so it doesn't really "contain" bytes that could be (im)mutable -- except if it is subject to lifetime extension, as in
```rust
const C: &Vec<i32> = &Vec::new();
```

Secondly, for `static`, the actual rule is that for bytes inside lifetime-extended values, they are *all* immutable. `UnsafeCell` makes no difference.

chorman0773:
I have a few ideas about how to write it, for varying levels of rigor. I think that in terms of lifetime extension, const-promotion occurs for both `static` and `const` items, correct? I'd expect that the const-promotion rule would apply there as well.

RalfJung:
Const promotion and lifetime extension are largely orthogonal mechanisms. I was talking about lifetime extension above.

chorman0773:
Ok, then I'd probably say
> All bytes within a const-promoted expression, or a lifetime-extended expression in the initializer of a `static` or `const` item are immutable. All bytes owned (directly) by an immutable binding or immutable `static` are immutable, except bytes within an [`core::cell::UnsafeCell`].

RalfJung:
Yeah, that sounds good.

PR #1637, opened 2024-09-29
Clarify definition of "immutable bytes"

RalfJung:
Fixes https://github.com/rust-lang/reference/issues/1616

This requires talking about lifetime extension to `'static` scope, which doesn't seem mentioned at all so I added a sentence and example for that.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -325,6 +325,18 @@ let x = &mut 0;
 println!("{}", x);
 ```
 
+r[destructors.scope.lifetime-extension.static]
+Lifetime extension also applies to `static` and `const` items, where it
+makes temporaries live until the end of the program. For example:
+
+```rust
+const C: &Vec<i32> = &Vec::new();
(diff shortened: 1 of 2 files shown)


src/destructors.md
```diff
@@ -325,6 +325,18 @@ let x = &mut 0;
 println!("{}", x);
 ```
 
+r[destructors.scope.lifetime-extension.const]
```
RalfJung, line 328: I'm not sure if I am supposed to just come up with some name for this paragraph or what...

Cc @chorman0773

chorman0773: Yeah, we're adding paragraph identifiers to the reference. If the file already has them, adding them in is good. Otherwise, you don't need to and they can be taken care of in the respective file PR.

chorman0773 (approved):
Information here is correct and not ambiguously presented.

ehuss (approved):
Thanks!

PR #1638, opened 2024-10-01
Add three more WebAssembly features to the list of accepted ones

alexcrichton:
This is intended to be a sibling PR to rust-lang/rust#131080 to update the reference documentation for wasm supporting two more features.


ehuss (approved):
Thanks!

PR #1644, opened 2024-10-04
Fix inline-assembly documentation for LoongArch

taiki-e:
Currently, clobber_abi doc lists `"efiapi"` as a supported ABI of LoongArch:

https://github.com/rust-lang/reference/blob/e61a925ee051c75c0dd71035c8e4e6b3383a0f3d/src/inline-assembly.md#L487

But only `"C"` and `"system"` are actually supported: https://github.com/rust-lang/rust/blob/267cf8d3b20e98315c8e0ddf1d7020f60f664249/compiler/rustc_target/src/asm/mod.rs#L942

r? @Amanieu


Amanieu (approved):
I don't have merge permissions on the reference.

chorman0773 (approved):
Seems correct, given the implementation.

ehuss (approved):
Thanks!

PR #1646, opened 2024-10-08
Add test linking

ehuss:
This adds the ability to link `rust-lang/rust` tests to reference rule annotations. Rules now have a link that pops up a list of tests that are associated with that rule. Additionally, there is a new appendix that lists the number of rules in each chapter, how many tests are associated, and various summaries.

This requires a local checkout of rust-lang/rust which is pointed to by the `SPEC_RUST_ROOT` environment variable.


chorman0773 (approved):
Looks to be working. The summary appendex is decent.

All of the rules were listed as not having tests, but I don't know whether that's because there aren't any yet or because it wasn't picking them up.

ehuss:
Thanks! Yea, there aren't any tests linked, yet. I was thinking about adding a few from one of the simpler chapters as a demonstration.

I'm going to go ahead and merge. I expect this to go through some iterations over time.

PR #1647, opened 2024-10-08
Add preview artifacts in CI

ehuss:
This adds a CI job which will generate a preview of the reference when a PR is posted/updated.

This does not yet offer a way to view it directly. I've been looking at various options for supporting that, but that might take a while to get that ready. This particular job is a requirement of that anyways, and since it provides a benefit now, I think it would be useful to add it.

This works by uploading to a GitHub Artifact. Artifacts persist for 90 days. You can download an artifact with:

`gh run download -R rust-lang/reference -n preview-1234`

where 1234 is the PR number. And then open `preview-1234/index.html` in your browser.

GitHub may or may not make this easier in the future (https://github.com/actions/upload-artifact/issues/14).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,6 +98,34 @@ jobs:
       working-directory: ./mdbook-spec
       run: cargo fmt --check
 
+  preview:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


chorman0773 (approved):
GHA workflow looks correct. 

Getting this set up with pages might be a good next step, but it's probably not something to do this PR.

PR #1648, opened 2024-10-08
Spec inline assembly tests

chorman0773:
This adds inline tests for the inline-assembly chapter. These serve as both examples for users reading the text and a micro test suite that can be run against a rust compiler.

Most `r[]` delimited rules have at least one test. Ones that don't are ones that can easily be tested (such as many of the `r[asm.rules]`).

Identifiers are not updated to match #1635, but this should be a merge-compatible change. I'd recommend merging that one before this one however, unless there are major changes required on it.

A few tests are `no_run`, they have been explained directly with html comments. If mdbook test can be improved to test for certain conditions (not full on ui_test), that would be useful, and the tests can be modified accordingly


ehuss (approved):
Thanks!

PR #1650, opened 2024-10-11
trait bounds grammar: make `?` and `for<>` mutually exclusive

lukas-code:
After https://github.com/rust-lang/rust/pull/127054, the `T: ?Trait` and `T: for<...> Trait` syntax is mutually exclusive, and this code no longer compiles:

```rust
fn foo<T: ?for<> Sized>() {}
```

Update the reference to reflect that.

|before|after|
|:-:|:-:|
|||

r? @compiler-errors


compiler-errors (approved):

ehuss (approved):
Thanks!

PR #1654, opened 2024-10-17
Spec: Improve Definition of Type Layout

chorman0773:
This improves the definition of Type Layout to be more formal. The following changes are made:
* The guarantees of repr(Rust) structs and unions are defined in a separate section, as they apply to all structs and unions, even when they have another representation.
* The layout of `repr(Rust)` enums is documented, with some permissions granted conservatively to implementations (like not including uninhabited variants in the size/alignment of the enum), pending approval either way. 
    * I'm not too attached to including these, but omitting them could be taken as a prohibition against them in the future.
 * Discriminant Elision (now guaranteed by https://github.com/rust-lang/rust/pull/130628) is documented. 
     * This is intended the supersede https://github.com/rust-lang/unsafe-code-guidelines/pull/538. 
 * The `repr(C)` layouts are consolidated, though the section on `struct`s is left presently. I'm not sure whether to get rid of it or replace it with a large note admonition, because it isn't needed as normative text, but the definition of `repr(C)` as applied to `struct`s can be confusing on its own. 

I think there's some more work to be done on the chapter, but I think this is a reasonable set of stuff to do in one PR, specifically targetted at the `repr` attribute.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -76,6 +76,9 @@ In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though
 their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only
 aligned to 4 bytes, not 8.
 
+r[layout.primitive.align-char]
+The alignment of `char` is the same as the alignment of `u32`.
+
 ## Pointers and References Layout
 
 r[layout.pointer]


src/type-layout.md
```diff
@@ -217,31 +248,65 @@ without a `repr` attribute. Using this representation explicitly through a
...
+r[layout.repr.rust.option]
 
-r[layout.repr.rust.field-storage]
-The second guarantee means that the fields can be
-ordered such that the offset plus the size of any field is less than or equal to
-the offset of the next field in the ordering. The ordering does not have to be
-the same as the order in which the fields are specified in the declaration of
-the type.
+r[layout.repr.rust.option.intro]
+Certain `repr(Rust)` enums are specially layed out when used with certain types. This layout process is known as discriminant elision.
```
workingjubilee, line 283:
```suggestion
Certain `repr(Rust)` enums are specially laid out when used with certain types. This layout process is known as discriminant elision.
```

src/type-layout.md
```diff
@@ -217,31 +248,65 @@ without a `repr` attribute. Using this representation explicitly through a
...
 
-Be aware that the second guarantee does not imply that the fields have distinct
-addresses: zero-sized types may have the same address as other fields in the
-same struct.
+r[layout.repr.rust.option.elligible]
```
workingjubilee, line 285:
```suggestion
r[layout.repr.rust.option.eligible]
```

src/type-layout.md
```diff
@@ -217,31 +248,65 @@ without a `repr` attribute. Using this representation explicitly through a
...
+An `enum` type is a *discriminant elision eligible* enum if:
+* It has exactly two variants,
+* One variant has exactly one field, known as the *elision candidate field*, and
+* The other variant has no fields that have size greater than 0 or alignment greater than 1, known as the *elided variant*.
```
workingjubilee, line 289: It is slightly confusing when this suggests the variant is "elided". It is not: it requires a concrete, physical encoding of _some_ kind for the discriminant. Instead, the rule here imposes a constraint on how that encoding is performed, which forces it to be encoded in a way that the resulting enum matches the ABI of another type.

The discriminant's encoding, ultimately, will still *occupy* that physical bitpattern.

src/type-layout.md
```diff
@@ -217,31 +248,65 @@ without a `repr` attribute. Using this representation explicitly through a
...
+* [`&U`][type.pointer.reference.shared], if `U` is a `Sized` type
+* [`&mut U`][type.pointer.reference.mut], if `U` is a `Sized` type
+* [`core::ptr::NonNull<U>`], if `U` is a `Sized` type
+* [`alloc::boxed::Box<U>`], if `U` is a `Sized` type
+* [`core::num::NonZero<T>`] if `T` is an integer type
+* A [function pointer][type.fn-pointer] type
+* A `struct`, other than [`core::cell::UnsafeCell<T>`], defined using [`repr(transparent)`][layout.repr.transparent], which has a field that is an *elision candidate*, and all other fields have size 0 and alignment 1.
+
+r[layout.repr.rust.option.elision]
+If the *elision candidate field* of a *discriminant elision eligible* `enum` has an *elision candidate type*, then the `enum` has the same layout as that field, except that the value consisting of all `0` bytes represents the *elided variant* of the `enum`.
```
workingjubilee, line 306: This part is strictly incorrect. It will have the _associated value_, determined by the "inner type" `T`, as T-lang has already FCPed a type that is all-`0xFF` bytes.

workingjubilee (changes requested):

PR #1656, opened 2024-10-19
Document mixed-site hygiene

GoldsteinE:
Current documentation for declarative macros seems to state that macros-by-example have call-site hygiene:

> By default, all identifiers referred to in a macro are expanded as-is, and are looked up at the macro’s invocation site.

This PR documents actual mixed-site hygiene rules and adds some examples.


traviscross (approved):

traviscross:
Looks good.  Thanks @GoldsteinE.

PR #1657, opened 2024-10-20
distinct 'static' items never overlap

RalfJung:
It seems like so far we did not actually guarantee this.

While we are at it, also clarify that static initializers can read even mutable statics, and what happens in that case.

### DIFF
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -21,9 +21,10 @@ Static items have the `static` lifetime, which outlives all other lifetimes in a
 Static items do not call [`drop`] at the end of the program.
 
 r[items.static.storage-disjointness]
-This allocated object is disjoint from all other `static` objects as well as heap-allocated and
-stack-allocated variables. However, the storage of immutable `static` items can overlap with objects
-that do not themselves have a unique address, such as [promoteds] and [`const` items][constant].
+If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other
+`static` objects as well as heap-allocated and stack-allocated variables. However, the storage of
+immutable `static` items can overlap with objects that do not themselves have a unique address, such
+as [promoteds] and [`const` items][constant].
 
 r[items.static.namespace]
 The static declaration defines a static value in the [value namespace] of the module or block where it is located.
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -21,7 +21,7 @@ Static items have the `static` lifetime, which outlives all other lifetimes in a
 Static items do not call [`drop`] at the end of the program.
 
 r[items.static.storage-disjointness]
-If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other
+If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other such
 `static` objects as well as heap-allocated and stack-allocated variables. However, the storage of
 immutable `static` items can overlap with objects that do not themselves have a unique address, such
 as [promoteds] and [`const` items][constant].
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -12,21 +12,26 @@ r[items.static.syntax]
 >   allowed semantically within `extern` blocks.
 
 r[items.static.intro]
-A *static item* is similar to a [constant], except that it represents a precise
-memory location in the program. All references to the static refer to the same
-memory location.
+A *static item* is similar to a [constant], except that it represents an allocated object in the
+program that is initialized with the initializer expression. All references and raw pointers to the
+static refer to the same allocated object.
 
 r[items.static.lifetime]
-Static items have the `static` lifetime, which outlives all
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+Static items have the `static` lifetime, which outlives all other lifetimes in a Rust program.
+Static items do not call [`drop`] at the end of the program.
+
+r[items.static.storage-disjointness]
+This allocated object is disjoint from all other `static` objects as well as heap-allocated and
+stack-allocated variables. However, the storage of immutable `static` items can overlap with objects
+that do not themselves have a unique address, such as [promoteds] and [`const` items][constant].
 
 r[items.static.namespace]
 The static declaration defines a static value in the [value namespace] of the module or block where it is located.
 
 r[items.static.init]
 The static initializer is a [constant expression] evaluated at compile time.
-Static initializers may refer to other statics.
+Static initializers may refer to and read from other statics.
+When reading from mutable statics, they read the initial value of that static.
 
 r[items.static.read-only]
 Non-`mut` static items that contain a type that is not [interior mutable] may
@@ -173,3 +178,4 @@ following are true:
 [_Expression_]: ../expressions.md
 [value namespace]: ../names/namespaces.md
 [_ItemSafety_]: functions.md
+[promoteds]: ../destructors.md#constant-promotion
--- a/src/items/static-items.md
+++ b/src/items/static-items.md
@@ -22,7 +22,7 @@ Static items do not call [`drop`] at the end of the program.
 
 r[items.static.storage-disjointness]
 If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other such
-`static` objects as well as heap-allocated and stack-allocated variables. However, the storage of
+`static` objects as well as heap allocations and stack-allocated variables. However, the storage of
 immutable `static` items can overlap with objects that do not themselves have a unique address, such
 as [promoteds] and [`const` items][constant].
 


workingjubilee:
I don't think you understood my example? I was positing that this seems to be a legal interpretation of the reference's current text:
```rust
static ZEE: u8 = 0;
static ZED: u8 = 0;
assert_eq!(&raw const ZEE, &raw const ZEE);
assert_eq!(&raw const ZED, &raw const ZED);
assert_eq!(&raw const ZEE, &raw const ZED);
```

workingjubilee:
Because this
> I would say if a static has a wibbly wobbly address not equal to itself, that's not a "precise memory location". We don't even have addresses that are not equal to themselves.

cannot be a response to what I actually said if it is said with an understanding of what I was trying to say. :/

RalfJung:
I understood the example. I don't understand how that can be a valid interpretation of the text. A static has a location, `&raw const` gives you a pointer pointing there, location comparison is reflexive.

RalfJung:
Oh, maybe I didn't quite understand the example.

But statics are certainly intended to be unique. That's their point -- they describe a place, distinct from all other places.

More specifically, statics form their own allocated objects that don't overlap with any other allocated object. So in fact ZST statics are not quite unique -- but statics of type `i32` are guaranteed to be at least 4 apart.

JakobDegen:
Lgtm, with the note that it's important this language remains restricted to static items, and not other language constructs that produce statics - const promotion, vtables, functions, etc. Will check my box when I'm off mobile, as apparently the GH app doesn't let you edit things anymore (or someone can do it for me :) )

RalfJung:
You probably don't have edit rights here. I don't.

You can use @`rfcbot reviewed` to check your box.

> with the note that it's important this language remains restricted to static items, and not other language constructs that produce statics - const promotion, vtables, functions, etc. 

I would argue those aren't statics, they are other kinds of global allocations -- exactly because of this fundamental difference.

JakobDegen:
@rfcbot reviewed

workingjubilee:
hmm. how must the addressing work for this, then?
```rust
static BLAH: &str = "blah";
static ALSO_BLAH: &str = "blah";
```

RalfJung:
> Are these potentially two different pointers to the same string literal?

Yes.

saethlin:
@rfcbot reviewed

digama0:
@rfcbot reviewed

src/items/static-items.md
```diff
@@ -8,16 +8,17 @@
...
 >   allowed semantically within `extern` blocks.
 
-A *static item* is similar to a [constant], except that it represents a precise
-memory location in the program. All references to the static refer to the same
-memory location. Static items have the `static` lifetime, which outlives all
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+A *static item* is similar to a [constant], except that it represents an allocated object in the
+program that is initialized with the initializer expression. This allocated object is disjoint from
+all other allocated objects. All references and raw pointers to the static refer to the same
```
chorman0773, line 13: Do we want to guarantee that they are disjoint from other allocated objects? Or just other statics. IE. if I have:
```rust
static FOO: i32 = 0;
const BAR: i32 = 0;

fn foo(){
    assert!(!core::ptr::eq(&FOO, &BAR));
}
```

Should we guarantee the assertion will always pass?

chorman0773: I don't have concern permissions, but I think this should be addresses by T-opsem before the end of the FCP.

chorman0773 (changes requested):

src/items/static-items.md
RalfJung: Yeah I think we should guarantee that, and it's what the text already says, isn't it?

CAD97:
@rfcbot reviewed

src/items/static-items.md
workingjubilee: That prevents emitting `unnamed_addr` on `const` items.

workingjubilee: `unnamed_addr` specifically *does* coalescing of non-significant-address-items into each other, and potentially into a significant-address item.

workingjubilee:
I have a concern: this proposed definition prevents emitting optimization annotations on **non**-`static` items that we may wish to have coalesced via optimizations that exploit the fact that `const` items have a non-significant address. Even if we wish to guarantee `static` unicity, it seems pointlessly penalizing to make this guarantee affect the ability to optimize, quite literally, the items that **don't** have unique addresses.

src/items/static-items.md
RalfJung: Wait, LLVM will merge `unnamed_addr` allocations with "named addr" (non-`unnamed_addr`) allocations? That's... surprising.

Cc @nikic

saethlin: The LangRef says:
> Note that a constant with significant address can be merged with a unnamed_addr constant, the result being a constant whose address is significant.

RalfJung: Ah, that is unfortunate.

chorman0773: > Yeah I think we should guarantee that, and it's what the text already says, isn't it?

That's what the proposed text says, my question is whether that's what we want it to say.

workingjubilee: I... honestly don't think that's surprising at all?

That's exactly what I would expect in a case of
```rust
const BIG_CONST: BigFrozen = BigFrozen::big_init();
static BIG_STATIC: BigFrozen = BIG_CONST;
```

That's the precise situation where the `const` will be unified "into" the `static`.

saethlin:
@rfcbot concern clarify-optimization-of-consts

I think the proposed text and the previous text have the same meaning, which is unfortunate because I think that we'd already specified that consts are not merged into statics. But given the participation on this PR, and its title which seems to be about overlap of statics, and the fact that I think rustc currently does not implement what is documented here, I would like the implications for consts to be spelled out clearly.

src/items/static-items.md
RalfJung: > I... honestly don't think that's surprising at all?

I guess we have different intuitions then.

But reality clearly works one way here, so I have updated the text to match reality. @chorman0773 please have a look.

RalfJung:
> But given the participation on this PR, and its title which seems to be about overlap of statics, and the fact that I think rustc currently does not implement what is documented here, I would like the implications for consts to be spelled out clearly.

I have that a shot, please have a look.

src/items/static-items.md
```diff
@@ -8,16 +8,20 @@
...
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+A *static item* is similar to a [constant], except that it represents an allocated object in the
+program that is initialized with the initializer expression. All references and raw pointers to the
+static refer to the same allocated object. Static items have the `static` lifetime, which outlives
+all other lifetimes in a Rust program. Static items do not call [`drop`] at the end of the program.
+
+This allocated object is disjoint from all other `static` objects as well as heap-allocated and
+stack-allocated variables. However, the storage of immutable `static` items can overlap with objects
+that do not themselves have a unique address, such as promoteds and `const` items.
```
workingjubilee, line 18: this seems basically good to me, thank you!

though I wonder if https://github.com/rust-lang/reference/blob/23ce619966541bf2c80d45fdfeecf3393e360a13/src/destructors.md?plain=1#L295 should be linked from [promoteds]?

chorman0773 (commented):
Given the concern is now resolved, I see no further issues with this.

RalfJung:
> @rustbot resolve clarify-optimization-of-consts

That was the wrong bot. :)

Let's see if I can also do this...

@rfcbot resolve clarify-optimization-of-consts

RalfJung:
@saethlin maybe you have to issue that command.

saethlin:
🤦 

@rfcbot resolve clarify-optimization-of-consts

src/items/static-items.md
```diff
@@ -12,21 +12,26 @@ r[items.static.syntax]
...
+stack-allocated variables. However, the storage of immutable `static` items can overlap with objects
+that do not themselves have a unique address, such as [promoteds] and [`const` items][constant].
 
 r[items.static.namespace]
 The static declaration defines a static value in the [value namespace] of the module or block where it is located.
 
 r[items.static.init]
 The static initializer is a [constant expression] evaluated at compile time.
-Static initializers may refer to other statics.
+Static initializers may refer to and even read from other statics.
```
nikomatsakis, line 33:
```suggestion
Static initializers may refer to and read from other statics.
```

This feels out of place.

nikomatsakis (approved):

joshtriplett:
A concern that came up in the @rust-lang/lang meeting: what is our definition of "overlap"? How does it handle ZSTs? There are multiple possible definitions we could use.

In particular, I don't think it should be possible for a ZST to be *in the middle* of an unrelated object (e.g. an unrelated array of non-ZSTs).

@nikomatsakis proposed defining overlap as "the start of one object lies between the start and end of the other, exclusive". That would prevent that property.

Another possible definition would be "has one or more bytes in common", but that *would* allow placing a ZST in the middle of an unrelated array.

traviscross:
@rfcbot fcp cancel

traviscross:
We'll repropose this as a dual FCP and rechecking the existing checked boxes from above.

nikomatsakis:
Discussed in the lang-team design meeting. We are aligned for non-ZST statics but the text seems underspecified for ZSTs. We suggest you either

1. clarify that the text does not apply to ZSTs or else
2. address the questions below.

Questions to be clarified:

* What is the definition of "overlap" (see options below)?

which in turn answers scenarios like this

* Can the address of a ZST static appear "in the middle" of another (non-ZST) static?
    * For example, if `static x: (u32, u32)` is at address X, can `static y: ()` be at address `X + 4`?
* Can two ZST statics have the same address?

Considerations:

* If we assert the first property, it implies that non-ZST statics cannot be placed at low addresses like 1 or 2, since e.g. a ZST value with alignment 2 will be assigned address 2. This might be something an embedded platform wants to do, for example.
* If we do not assert the second property, then users need to ensure that their statics have at least one byte if they want to rely on the address.

We would like op-sem to give recommendations. We considered two definitions for overlap:

```rust
fn overlap_a(start1: usize, end1: usize, start2: usize, end2: usize) -> bool {
    // Overlap iff the start of one (potentially trivial) range falls within another non-trivial range.
    //
    // Put another way: no byte of one static can be stored
    // at the starting address of another.
    start1 <= start2 < end1 ||
    start2 <= start1 < end2
}

fn overlap_b(start1: usize, end1: usize, start2: usize, end2: usize) -> bool {
    // Overlap iff they have bytes in common:
    overlap_a(start1, end1, start2, end2)
    && start1 != end1 // <-- empty ranges cannot overlap
    && start2 != end2
}
```

RalfJung:
> what is our definition of "overlap"? 

I would say two ranges do *not* overlap if `end1 <= start2 || end2 <= start1`. So indeed, a ZST static is not allowed to be "in the middle of" another static.

> If we assert the first property, it implies that non-ZST statics cannot be placed at low addresses like 1 or 2, since e.g. a ZST value with alignment 2 will be assigned address 2. 

I don't think it implies this. There's no ZST *allocation* created at address 2. It's just a pointer without provenance that has address 2, that's not in conflict with anything.

RalfJung:
> If we assert the first property, it implies that non-ZST statics cannot be placed at low addresses like 1 or 2, since e.g. a ZST value with alignment 2 will be assigned address 2.

Or... do we actually have codegen synthesize addresses like this now? And not just the library by calling `NonNull::dangling`?

RalfJung:
> Can two ZST statics have the same address?

Yes, definitely. This is desirable, it means one can soundly use the pattern of "my linker defines two symbols that mark the beginning and end of a relevant range of memory" via two ZST statics.

traviscross:
> I would say two ranges do _not_ overlap if `end1 <= start2 || end2 <= start1`: one range has to end before the other begins. Therefore, they overlap if `end1 > start2 && end2 > start1`.

I thought about this for awhile.  I don't have an answer yet, but I did decide this logic as stated doesn't hold.  That is, I agree that two ranges do not overlap if `p` is true for `p == end1 <= start2 || end2 <= start1`.  But `p => !overlap` does not imply `!p => overlap`.  That would be affirming the consequent.

workingjubilee:
> In particular, I don't think it should be possible for a ZST to be in the middle of an unrelated object (e.g. an unrelated array of non-ZSTs).

I don't think that's a coherent model. As far as I have understood, we have said that ZSTs can overlap with non-ZSTs because the overlap is of 0 bytes, each time, and an overlap of 0 bytes is not an overlap. I think going down the route of "an allocation 0 bytes cannot 'overlap' with an existing allocation" is a bad idea, because we would be:
- inviting self-contradiction by complicating the model
- implicitly saying ZSTs cannot actually exist in Rust by imposing constraints that say ZSTs can no longer be at certain locations, when programmers actually do rely on the ability to allocate ZSTs at certain locations!
- ultimately, trying to promise programmers things we cannot actually offer, instead of just admitting there is a sharp edge around 1ZSTs being slightly nonintuitive, because the idea of _zero_ is not perfectly intuitive in general

Can T-lang explain the actual utility of this prohibition?

Why should there be a statement that a 1ZST cannot be allocated somewhere I could absolutely create a pointer to a zero-sized byte-slice? Why should the reasoning be fundamentally different?

traviscross:
> we have said that ZSTs can overlap with non-ZSTs because the overlap is of 0 bytes, each time, and an overlap of 0 bytes is not an overlap.... there is a sharp edge around 1ZSTs being slightly nonintuitive, because the idea of _zero_ is not perfectly intuitive in general

This I find persuasive.

chorman0773:
I'm not sure a prohibition here would do anything to users, only to implementations - it would simply prohibit placing a ZST static at an address that lies inside the storage of another static. We're not talking about allocations generally either, just `static` items in particular. 
Frankly, I think any reasonable definition of "overlap" not explicitly written to permit this would prohibit doing that. I also think this is true of aggregate layout - any meaningful definition of "overlap" would prohibit the compiler from placing a ZST field at an offset that leaves it "inside" a non-zero sized field. 

I'm not sure there's substantial programmer benefit from this prohibition (as this doesn't present any opsem concerns like non-zero sized allocations overlapping can), but I don't think there's substantial implementation benefit from explicitly authorizing it either (I'm not sure why a compiler would want to place a ZST static arbitrarily within some other static, but I doubt it helps any optimizations), so I think the best way forward is the intuitive definition that would disallow it.

chorman0773:
With the above being said, I think that a definition of "Range R overlaps Range Q if Q.start <= R.start && R.end < Q.end`" is the most reasonable and intuitive definition (also used elsewhere in mathematics when talking about numeric ranges). As mentioned above, I don't see a benefit to us defining this in a way that exempts an empty range R.

traviscross:
> I think that a definition of "Range R overlaps Range Q if Q.start <= R.start && R.end < Q.end`" is the most reasonable and intuitive definition

For ranges:

- `R: [1, 3)`
- `Q: [1, 3)`

This rule would not imply that the ranges overlap (since it is false that `R.end < Q.end`).

The rule @RalfJung gave above for overlap is correct (strengthened to "if and only if"):

> I would say two ranges do _not_ overlap if `end1 <= start2 || end2 <= start1`: one range has to end before the other begins. Therefore, they overlap if `end1 > start2 && end2 > start1`.

Put another way, two ranges overlap if and only if there is some value that might appear in both ranges.  To prove that impossible, it's necessary and sufficient to prove that the last included value of each range cannot contain the first included value of the other range.

workingjubilee:
> I'm not sure a prohibition here would do anything to users, only to implementations

Hmm. One of my preferred mental models for Rust is one where the compiler is a very trustworthy Rust programmer who writes an unnerving amount of `unsafe` code, and especially `asm!`. Likewise, I sometimes explain that the rules of `unsafe` can often be thought of as manually upholding the rules the compiler also obeys.

And I would find it hard to explain to people who enjoy playing linker tricks with `static` items why they shouldn't arrange for a given `static ZST` to be found at an address that might be within a given object, considering it occupies zero bytes. Occasionally they actually link `static` objects of non-zero size to be overlapping, and then I explain that bytes can't overlap like that, but they need the ability to point at addresses somehow. Then I explain they should use `extern static`, but the conversation usually arrives at something like "but linking in a separate crate just to make these statics `extern` is super annoying!"

chorman0773 (approved):
Late on this, but my concern earlier was resolved, so approving now.

workingjubilee:
Note I am happy to be corrected on this if all uses of ZST markers by people engaging in linker-script-shenanigans are compatible with them being ZSTs that are never "within" another object, but my memory was that trick was used in a few ways and that they were always at risk of this being a problem, i.e. they might want to put one or both of those markers inside a larger `static`.

CAD97:
I think that allowing zero-sized `static` places to have an address strictly inside the address range of a different `static` is the slightly better choice. Because users can place `static`s at custom addresses by utilizing linker scripts, this isn't solely a question of restricting the implementation, but also one of restricting users. If implementation considerations are otherwise immaterial, the option with less potential UB should be preferred.

Zero-sized accesses/places are already special-cased, so special casing zero-sized `static` in the rule to explicitly exempt them from overlap rules doesn't seem unreasonable. But there are two very straightforward formulations that don't need to do so:

- `static` items cannot *alias* other `static` items (using the definition of *aliasing* from the reference aliasing model).
- No byte of memory is contained in more than one `static` item.

nikomatsakis:
My two cents:

I agree with @workingjubilee's model of the compiler as "just another Rust programmer", that's how I like to think of it too (and I like to think of the type system as "guardrails" that keep you on the straight-and-narrow, but which you can choose to take off).

I don't know of specific utility from defining overlap based on the "start address", though I will say that I found it surprising to think of the address of a ZST-static falling inside another one.

I am curious what people think about the uniqueness of ZST-statics as well. I think if I just consider the pointer value of a ZST-static to be effectively meaningless and arbitrary, which seems to be implied by definitions of overlap that focus on actual *bytes*, then it all makes a lot more sense to me and I agree seems consistent.

pnkfelix:
@nikomatsakis  [wrote](https://github.com/rust-lang/reference/pull/1657#issuecomment-2461871561):
> The actual [comment in question](https://github.com/rust-lang/reference/pull/1657#issuecomment-2460271116) said that the text as written was ambiguous and asked for clarification. Certainly some people in the discussion felt that having the start address of a ZST static lie within another static ought not to be allowed but others did not, and hence we decided to ask that the text either _not_ answer the question or provide an answer.

Just to emphasize this point further: Niko's [original comment](https://github.com/rust-lang/reference/pull/1657#issuecomment-2460271116) omitted one point that I think all of T-lang agreed upon during our meeting yesterday: During our conversation, I believe the whole team agreed that we would support allowing this PR to make progress if it simply restricted the guarantee be solely regarding positively-sized statics.

I.e., that would allow us leave the question of whether ZSTs also participated in the guarantee to be a matter we can resolve later.

(Obviously this would have the short-term effect of making the guarantees of zero-sized static items an implementation-specific detail that one should not rely upon, which is not an ideal state of affairs for the long-term; but it seemed to me from skimming the conversation in this thread that the participants here mostly cared about strengthening the guarantees for sets of positively-sized static items, and thus this compromise would be a reasonable way to make progress in the short term.)

nikomatsakis:
Thinking on this a bit more, what my message could've made clearer is that we were actively looking for a recommendation here (along with rationale) because it seemed unclear what was best and we recognize others are such closer to the discussion. *If there was not consensus amongst those participants (or if we find we don't agree with the rationale) another option is to defer the question. 

The motivation then to include the questions for consideration is that these are things that came up in the meeting so if the rationale or decision does not address this then we will likely just have the same questions again.

nikomatsakis:
Thanks for the clarification, @workingjubilee, I appreciate your message. I believe @joshtriplett's comment was meant to convey his personal opinion, but it can be difficult to tell the difference sometimes, and I already said how I felt my message could've been clearer.

I have a few takeaways from this:

* When writing for the team, be explicit about when the team has a consensus position and when it does not.
* When writing for oneself, it never hurts to be extra explicit about whether you are speaking for yourself or the team.

In general, I like the pattern of teams like opsem (or types) providing recommendations and rationale, and lang having the role of "double checking" and raising questions. In some sense, I view lang as playing the role of the non-expert, whereas the teams are the experts in the domain. The result has to make sense to both.

RalfJung:
>  I think going down the route of "an allocation 0 bytes cannot 'overlap' with an existing allocation" is a bad idea, because we would be:
> [...] inviting self-contradiction by complicating the model

I don't think so, the definition of "disjointness" that I suggested above is not particularly complicated.

> implicitly saying ZSTs cannot actually exist in Rust in a way that is useful, by imposing constraints that say ZSTs can no longer be at certain locations, when programmers actually do rely on the ability to allocate ZSTs at certain locations!

That is the question, isn't it? Do they rely on this? Generally programmers can't allocate things at certain locations, the allocator picks the location.

> Let's back up to first principles. Two ranges overlap if and only if there is some value that might appear in both ranges.

No, I don't think that has to be the usual definition of overlap of a *range*. Ranges are not sets.

I don't know which definition is more widely use in mathematics. @chorman0773 said my definition is used there; do you have a source for that? @digama0 do you have any idea?

> To extend this principle to when the ranges might be empty, we need to handle that case and say that they cannot overlap if and only if:

No, my definition already covered the "empty range" case in the intended way. Sorry for not being clear about that.

---

Anyway, I'm also fine with just punting on the question for ZST for now. :shrug: It just pains me a little since I think many of the ways in which people think ZST are special aren't actually anything special, they fall out of the same general principles. For instances, people say things like "two `&mut T` have different addresses except when `T` is a ZST"... which is true but (a) unnecessarily complicated, and (b) unnecessarily weak. Instead we can say "two `&mut T` point to disjoint ranges of memory", and this is (a) stronger (e.g. of `T` is `i32`, it tells us they are at least 4 bytes apart), and (b) doesn't need an "except".

But in the interest of documenting what we have consensus on and not closing the door on future possibilites, I can try to adjust the PR to only be about non-ZST statics.

digama0:
> No, I don't think that has to be the usual definition of overlap of a _range_. Ranges are not sets.
> 
> I don't know which definition is more widely used in mathematics. @chorman0773 said my definition is used there; do you have a source for that? @digama0 do you have any idea?

I think that the definition of overlap of sets is clearly that they have empty intersection, but I think your definition makes more sense in this context. The difference of course is whether we want to say that `[0, 2)` overlaps `[1, 1)`, which is false for sets because the latter set is empty, but is true under the endpoint-based definition. Off the top of my head I don't know a formalization of this interval order, I think it's not likely to come up except in CS-like contexts and in that case the definitions are usually tailored to the application anyway.

RalfJung:
> I recall a related version of this issue: We currently allow ZST's to be magicked up anywhere, such that <*const ()>::dangling(n) produces a valid pointer for any n. But if so, then that implies that we can also create ZST allocations in the middle of other allocations, 

As mentioned above, "maigcking up" a ZST like that does not create a ZST *allocation*. It just creates a pointer/reference without provenance. Those are not the same thing.

---

I wonder if we are constrained by LLVM here... @nikic do you know if there will be trouble with LLVM when we have a zero-sized static located "inside" another static (or inside a stack/heap allocation)? When I do `x = malloc(10)`, is LLVM then allowed to optimize `static_addr <= x || static_addr >= x+10` to `true`?

workingjubilee:
I went surveying various embedded software implementations using Rust today. Many of them do in fact use patterns that would prefer to be able to place `static` ZSTs at fairly arbitrary locations (including inside another `static`) and use those markers to generate slices in the ways we've discussed, because they need to somehow reason about fixed ranges of memory in slice-like manners. Ideally we would have some sort of pattern that enables writing this, and if not by using the setoid-based rule here, then we'd want something else.

One or two do in fact do everything in the most tedious way (entirely `asm!` and linker script), but not many seem to tolerate slogging through that mire (...they still gotta suffer through the linker script though).

RalfJung:
Note that using the *addresses* of those statics as markers can be fine, but under no circumstances is it fine to access the same underlying global memory with pointers derived from different `static` or `extern static` declarations.

nikic:
> I went surveying various embedded software implementations using Rust today. Many of them do in fact use patterns that would prefer to be able to place `static` ZSTs at fairly arbitrary locations (including inside another `static`) and use those markers to generate slices in the ways we've discussed, because they need to somehow reason about fixed ranges of memory in slice-like manners. Ideally we would have some sort of pattern that enables writing this, and if not by using the setoid-based rule here, then we'd want something else.
> 
> One or two do in fact do everything in the most tedious way (entirely `asm!` and linker script), but not many seem to tolerate slogging through that mire (...they still gotta suffer through the linker script though).

Do the embedded use-cases just need ZST markers that are directly adjacent to other statics (i.e. they share a start/end), or are they genuinely nested inside them? If the latter, do you have any example to share?

nikic:
> I wonder if we are constrained by LLVM here... @nikic do you know if there will be trouble with LLVM when we have a zero-sized static located "inside" another static (or inside a stack/heap allocation)? When I do `x = malloc(10)`, is LLVM then allowed to optimize `static_addr <= x || static_addr >= x+10` to `true`?

After some experimentation, I got this somewhat concerning result: https://llvm.godbolt.org/z/6qTvx3qxd

I guess test1 answers your question in terms of what LLVM currently assumes at least, while test0 looks like an outright miscompile to me. (Alive seems to be struggling with zero-sized globals and thinks that everything is UB: https://alive2.llvm.org/ce/z/qFySyd)

workingjubilee:
I can go into some more detail after I get some sleep but a simplified example would look like this, so yes, strict inclusion: https://godbolt.org/z/eE6Wdjvnb

Most of the ones I looked at were from fairly diligent and clued-in folks, so even the cases that wrote code that is _probably_ still UB tbh went to fairly extreme lengths to write contorted code that they clearly are hoping will avoid the notice of an optimizer.

...uh, that's fairly alarming and we should probably bring this up to Alive.

RalfJung:
> After some experimentation, I got this somewhat concerning result: https://llvm.godbolt.org/z/6qTvx3qxd

Could you add some comments that  aid in interpreting the results? :)

nikic:
> > After some experimentation, I got this somewhat concerning result: https://llvm.godbolt.org/z/6qTvx3qxd
> 
> Could you add some comments that aid in interpreting the results? :)

test0: zero-size global == start of alloca -> false
test1: zero-size global == middle of alloca -> false
test2: zero-size global == end of alloca -> unknown

test0 is a miscompile, and test1 depends on whether a zero-size global can be in the middle of an alloca or not. LLVM currently assumes it can't, but given that it also assumes it can't be at the *start* of one, it's hard to distinguish whether that's a bug or a feature :)

I think if there is a consensus that we really want to allow zero-size statics that arbitrarily overlap with other allocations, we probably could get through a LangRef change to that effect and adjust InstSimplify accordingly.

Though I think in this context, it's also important to distinguish between a) what addresses an extern static may be placed at (e.g. via linker script) and b) what addresses Rust itself can place a static. Even if we allow extern statics to overlap other statics, if you define a static directly in Rust, is Rust allowed to place it at an overlapping address?

RalfJung:
That helps, thanks!

I don't have a strong opinion on whether should allow zero-size statics that arbitrarily overlap with other allocations. If it can be done in LLVM without impacting relevant optimizations, I would generally err on the side of having less UB.

workingjubilee:
> Though I think in this context, it's also important to distinguish between a) what addresses an extern static may be placed at (e.g. via linker script) and b) what addresses Rust itself can place a static. Even if we allow extern statics to overlap other statics, if you define a static directly in Rust, is Rust allowed to place it at an overlapping address?

My impression is that from the perspective of embedded developers, the artificial distinction we draw between `extern "C" static` and "native" Rust `static` is not very strong to them. Many do avoid it, but especially in code that I see when someone is asking for help debugging stuff,  they don't bother with the `extern "C"` linkage. Partly because it works: as the example I show demonstrates, all you need to is `#[link_section]` and then you can start picking addresses.

And as I've mentioned, what's more important is having _a_ blessed pattern that we can recommend to this community, rather than telling them what _not_ to do, which mostly leads to them producing obfuscated code:

> Ideally we would have some sort of pattern that enables writing this, and if not by using the setoid-based rule here, then we'd want something else.

workingjubilee:
I have opened https://github.com/rust-lang/unsafe-code-guidelines/issues/546 to try to capture some of the discussion from here about the ZST issue and will try to expand on the embedded use-cases in that issue or the also-relevant https://github.com/rust-lang/unsafe-code-guidelines/issues/545

RalfJung:
@scottmcm okay, so that would place ZST constant allocations at `NonNull::dangling` addresses. That's already covered by the wording of this PR even without the ZST exception.

Do zero-sized local variables get an `alloca`, or what do we do about them?

joshtriplett:
> What I was responding to was specifically @joshtriplett's comment:
> 
> > In particular, I don't think it should be possible for a ZST to be in the middle of an unrelated object (e.g. an unrelated array of non-ZSTs).
> > @nikomatsakis proposed defining overlap as "the start of one object lies between the start and end of the other, exclusive". That would prevent that property.
> 
> Which doesn't seem to have a clear motivation behind _why_ preventing that property is desirable, and it sounded pretty definite, at least? Perhaps I am attaching too much finality to "I don't think it should be possible"?
> 
> I'm aware I can come across as needling, and I don't particularly enjoy it either. I feel I get into situations like the one that started with this PR often enough, where everyone signs off on something within 2 hours and then suddenly raising my concern and making sure people understand it is immediately on a timer. Then I feel rushed and have to fumble together an explanation before people have Made Up Their Minds.

That comment, written during the meeting in the course of discussion, was very much meant to be an early indicator of "this raised some eyebrows in a meeting around what seemed like an unaddressed corner case; here were some various thoughts that came up". It was not meant to convey any kind of definitive conclusion on What Behavior The Team Wants, just to start a discussion. That said, I could absolutely have spelled that out more explicitly. Sorry if it came across as a Decision Being Made rather than a Discussion Being Had.

digama0:
For what it's worth, my answer above was given without having read too much of the backscroll, and now that I have I'm leaning more toward allowing ZST allocations (or "allocations") to be anywhere, including strictly inside another allocation, consistently for all purposes. But for the purpose of this RFC, I'm also fine with just punting on the question for now.

nikomatsakis:
Jubilee, can you clarify what you are saying about embedded uses-- are you saying that the projects are relying on a "bytes in common" definition (i.e., they have ZSTs whose pointer value is a byte that is a member of another static)?

nikic:
The LLVM issue is fixed by https://github.com/llvm/llvm-project/pull/115728.

src/items/static-items.md
```diff
@@ -12,21 +12,27 @@ r[items.static.syntax]
...
 
 r[items.static.lifetime]
-Static items have the `static` lifetime, which outlives all
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+Static items have the `static` lifetime, which outlives all other lifetimes in a Rust program.
+Static items do not call [`drop`] at the end of the program.
+
+r[items.static.storage-disjointness]
+If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other
```
joshtriplett, line 24:
```suggestion
If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other such
```

To make it clear that this applies to other statics with a size of at least one byte.

traviscross:
Given the change in 8dde3eba9aadaa2c0a054da731a91f2ef55c6a33 that excludes the challenging bit under discussion here with respect to ZSTs, let's...

@rfcbot fcp merge

traviscross:
Checking off the boxes for members of T-opsem as these were checked off in the earlier FCP.

tmandry:
I think the clarification proposed by @joshtriplett (or something more explicit) should be included, but otherwise I'm happy to go ahead with this.

@rfcbot reviewed

ehuss:
@joshtriplett @nikomatsakis @scottmcm Do you have a chance to look at this and check your box or register a concern?

src/items/static-items.md
```diff
@@ -12,21 +12,27 @@ r[items.static.syntax]
...
-Static items have the `static` lifetime, which outlives all
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+Static items have the `static` lifetime, which outlives all other lifetimes in a Rust program.
+Static items do not call [`drop`] at the end of the program.
+
+r[items.static.storage-disjointness]
+If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other such
+`static` objects as well as heap-allocated and stack-allocated variables. However, the storage of
+immutable `static` items can overlap with objects that do not themselves have a unique address, such
```
kpreid, line 26: “Immutable `static` item” could be interpreted as meaning a static item without the `mut` keyword, but it must also exclude interior-mutable memory. Perhaps this should be clarified (or perhaps it doesn’t matter since this is a “can”, not “must”).

src/items/static-items.md
```diff
@@ -12,21 +12,27 @@ r[items.static.syntax]
...
 r[items.static.lifetime]
-Static items have the `static` lifetime, which outlives all
-other lifetimes in a Rust program. Static items do not call [`drop`] at the
-end of the program.
+Static items have the `static` lifetime, which outlives all other lifetimes in a Rust program.
+Static items do not call [`drop`] at the end of the program.
+
+r[items.static.storage-disjointness]
+If the `static` has a size of at least 1 byte, this allocated object is disjoint from all other such
+`static` objects as well as heap-allocated and stack-allocated variables. However, the storage of
```
kpreid, line 25: What’s a “heap-allocated variable”? Presumably, things like the contents of `Box`es, but calling that a variable seems too informal and imprecise to be appropriate for the reference — especially as the reference already [defines a variable as](https://doc.rust-lang.org/reference/variables.html) “a component of a stack frame”.

Perhaps something like:

```suggestion
`static` objects as well as heap allocations and stack allocations. However, the storage of
```

kpreid (commented):
I have a couple of thoughts on clarity and precision of the wording.

src/items/static-items.md
RalfJung: Yes we should have this as a shared definition somewhere that we can reference.

ehuss (approved):
Thanks @RalfJung!

PR #1663, opened 2024-10-24
Add documentation for `#[diagnostic::do_not_recommend]`

weiznich:
Documentation for https://github.com/rust-lang/rust/pull/132056


src/attributes/diagnostics.md
```diff
@@ -492,6 +492,47 @@ error[E0277]: My Message for `ImportantTrait<i32>` implemented for `String`
    = note: Note 2
 ```
 
+### The `diagnostic::do_not_recommend` attribute
+
+The `#[diagnostic::on_unimplemented]` attribute is a hint to the compiler to not show a certain trait implementation as part of the error message.
```
ehuss, line 497: Can this explain a little more what error this is referring to, and what circumstances it will avoid showing an error?

weiznich: > Can this explain a little more what error this is referring to

Currently this is mostly relevant for E0277, although that might change in the future?

> and what circumstances it will avoid showing an error?

It never avoids showing an error, it just changes the content of the error message to not mention/consider certain trait implementations if somewhat reasonable possible.

ehuss:
@weiznich I was wondering if you could help me come up with a slightly more realistic example that demonstrates an improvement in the diagnostics? I feel like the example in the RFC at https://github.com/rust-lang/rfcs/blob/master/text/2397-do-not-recommend.md#guide-level-explanation is really close. It mentions "Perhaps when a function expects Bar and it's not implemented, it would never make sense to implement Foo for that type.", but I can't immediately think of an example where "it would never make sense to implement Foo". Is there maybe a way to relate the example to a realistic scenario?

traviscross:
Specifically also, it would help to have a minimal self-contained example.

weiznich:
The rustc test suite has several more relatistic examples:

* [`AsExpression`](https://github.com/rust-lang/rust/blob/master/tests/ui/diagnostic_namespace/do_not_recommend/as_expression.rs) which was minimized from diesel
* [The tuple pyramid example](https://github.com/rust-lang/rust/blob/master/tests/ui/diagnostic_namespace/do_not_recommend/supress_suggestions_in_help.rs) which closely aligns with what bevy and also diesel want to do with quite a lot of "internal" implementations.

Both examples are quite a bit larger than he trivial one, that's the reason why I've not used the here.

> It mentions "Perhaps when a function expects Bar and it's not implemented, it would never make sense to implement Foo for that type.", but I can't immediately think of an example where "it would never make sense to implement Foo". Is there maybe a way to relate the example to a realistic scenario?

I think the wording is a bit too absolute here, in both cases linked above it's not that the suggested bound is **never** that one that the user want's. It's only that in almost all cases (for diesel something like >95%) the wrong suggestion that points users into the wrong direction. But there are edge cases where the current (without `#[diagnostic::do_not_recommend]`) suggestions are valid.

ehuss:
@weiznich Thanks! Both of those examples look pretty good, so I picked one that seemed like a pretty good demonstration. 

Does the current wording seem accurate to you?

weiznich (commented):
Thanks for adding the more complex example. This sounds better now. I've left a minor comment about the wording otherwise this sounds good to me.

ehuss (approved):
Thanks!

PR #1664, opened 2024-10-25
Add Values and Representation chapter

chorman0773:
This documents the values for most types (where it has been decided), as well as the representation of these values in memory. It also defines what a byte is in Rust (including initialized and uninitialized memory). 

The chapter does not define what Provenance carries in Rust. `repr(Rust)` enums are also not fully elaborated, as there are things undecided. 

This makes a normative reference to <https://ieeexplore.ieee.org/document/8766229> for floating-point format.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -99,6 +99,7 @@
     - [Type coercions](type-coercions.md)
     - [Destructors](destructors.md)
     - [Lifetime elision](lifetime-elision.md)
+    - [Values and Representation](values.md)
 
 - [Special types and traits](special-types-and-traits.md)
 
--- a/src/values.md
+++ b/src/values.md
@@ -1,0 +1,175 @@
+# Values and Representation
+
+r[value]
+
+## Bytes
+
+r[value.byte]
+
+r[value.byte.intro]
+The Most basic unit of Memory in Rust is a Byte. All values in Rust are computed from 0 or more bytes read from an allocation.
+
+> [!NOTE]
+> While bytes in Rust are typically lowered to hardware bytes, they may contain additional values,
+> such as being uninitialized, or storing part of a pointer.
+
+r[value.byte.init]
+Each byte may be initialized, and contain a value of type `u8`, as well as an optional pointer fragment.
+
+r[value.byte.uninit]
+Each byte may be uninitialized.
+
+> [!NOTE]
+> Uninitialized bytes do not have a value and do not have a pointer fragment.
+
+## Value Encoding
+
+r[value.encoding]
+
+r[value.encoding.intro]
+Each type in Rust has 0 or more values, which can have operations performed on them
+
+> [!NOTE]
+> `0u8`, `1337i16`, and `Foo{bar: "baz"}` are all values
+
+r[value.encoding.op]
+Each value of a type can be encoded into a sequence of bytes, and decoded from a sequence of bytes, which has a length equal to the size of the type.
+The operation to encode or decode a value is determined by the representation of the type.
+
+> [!NOTE]
+> Representation is related to, but is not the same property as, the layout of the type.
+
+r[value.encoding.decode]
+If a value of type `T` is decoded from a sequence of bytes that does not correspond to a defined value, the behavior is undefined. If a value of type `T` is decoded from a sequence of bytes that contain pointer fragments, which are not used to represent the value, the pointer fragments are ignored.
+
+## Primitive Values
+
+r[value.primitive]
+
+r[value.primitive.integer]
+Each value of an integer type is a whole number. For unsigned types, this is a positive integer or `0`. For signed types, this can either be a positive integer, negative integer, or `0`.
+
+r[value.primtive.integer-width]
+The range of values an integer type can represent depends on its signedness and its width, in bits. The width of type `uN` or `iN` is `N`. The width of type `usize` or `isize` is the value of the `target_pointer_width` property.
+
+r[value.primitive.integer-range]
+The range of an unsigned integer type of width `N` is between `0` and `1<<N - 1` inclusive. The range of a signed integer type of width `N` is between `-(1<<(N-1)` and `1<<(N-1) - 1` inclusive.
+
+> [!NOTE]
+> There are exactly `1<<N` unique values of an integer type of width `N`.
+
+r[value.primitive.unsigned-repr]
+A value `i` of an unsigned integer type `U` is represented by a sequence of initialized bytes, where the `m`th successive byte according to the byte order of the platform is `(i >> (m*8)) as u8`, where `m` is between `0` and the size of `U`. None of the bytes produced by encoding an unsigned integer has a pointer fragment.
+
+> [!NOTE]
+> The two primary byte orders are `little` endian, where the bytes are ordered from lowest memory address to highest, and `big` endian, where the bytes are ordered from highest memory address to lowest.
+> The `cfg` predicate `target_endian` indicates the byte order
+
+> [!WARN]
+> On `little` endian, the order of bytes used to decode an integer type is the same as the natural order of a `u8` array - that is, the `m` value corresponds with the `m` index into a same-sized `u8` array. On `big` endian, however, the order is the opposite of this order - that is, the `m` value corresponds with the `size_of::<T>() - m` index in that array.
+
+r[value.primitive.signed-repr]
+A value `i` of a signed integer type with width `N` is represented the same as the corresponding value of the unsigned counterpart type which is congruent modulo `2^N`.
+
+r[value.primitive.char]
+Each value of type `char` is a Unicode Scalar Value, between `U+0000` and `U+10FFFF` (excluding the surrogate range `U+D800` through `U+DFFF`).
+
+r[value.primitive.char-repr]
+The representation of type `char` is the same as the representation of the `u32` corresponding to the Code Point Number encoding by the `char`.
+
+r[value.primitive.bool]
+The two values of type `bool` are `true` and `false`. The representation of `true` is an initialized byte with value `0x01`, and the representation of `false` is an initialized  byte with value `0x00`. Neither value is represented with a pointer fragment.
+
+## Pointer Value
+
+r[value.pointer]
+
+r[value.pointer.thin]
+Each thin pointer consists of an address and an optional provenance. The address refers to which byte the pointer points to. The provenance refers to which bytes the pointer is allowed to access, and the allocation those bytes are within.
+
+> [!NOTE]
+> A pointer that does not have a provenance may be called an invalid or dangling pointer.
+
+r[value.pointer.thin-repr]
+The representation of a value of a thin pointer is a sequence of initialized bytes with `u8` values given by the representation of its address as a value of type `usize`, and pointer fragments corresponding to its provenance, if present.
+
+r[value.pointer.thin-ref]
+A thin reference to `T` consists of a non-null, well aligned address, and provenance for `size_of::<T>()` bytes starting from that address. The representation of a thin reference to `T` is the same as the pointer with the same address and provenance.
+
+> [!NOTE]
+> This is true for both shared and mutable references. There are additional constraints enforced by the aliasing model.
+
+r[value.pointer.fat]
+A fat pointer or reference consists of a data pointer or reference, and a pointee-specific metadata value.
+
+r[value.pointer.fat-reference]
+The data pointer of a fat reference has a non-null address, well aligned for `align_of_val(self)`, and with provenance for `size_of_val(self)` bytes.
+
+r[value.pointer.fat-representation]
+A fat pointer or reference is represented the same as `struct FatPointer<M>{data: *mut (), metadata: M}` where `M` is the pointee metadata type, and the `data` and `metadata` fields are the corresponding parts of the pointer.
+
+> [!NOTE]
+> The `FatPointer` struct has no guarantees about layout, and has the default representation.
+
+r[value.pointer.fn]
+A value of a function pointer type consists of an non-null address. A function pointer value is represented the same as an address represented as an unsigned integer type with the same width as the function pointer.
+
+> [!NOTE]
+> Whether or not a function pointer value has provenance, and whether or not this provenance is represented as pointer fragments, is not yet decided.
+
+## Aggregate Values
+
+r[value.aggregate]
+
+r[value.aggregate.value-bytes]
+A byte `b` in the representation of an aggregate is a value byte if there exists a field of that aggregate such that:
+* The field has some type `T`,
+* The offset of that field `o` is such that `b` falls at an offset in `o..(o+size_of::<T>())`,
+* Either `T` is a primitive type or the offset of `b` within the field is a value byte in the representation of `T`.
+
+> [!NOTE]
+> A byte in a union is a value byte if it is a value byte in *any* field.
+
+r[value.aggregate.padding]
+Every byte in an aggregate which is not a value byte is a padding byte.
+
+r[value.aggregate.struct]
+A value of a struct type consists of the values of each of its fields.
+The representation of such a struct contains the representation of the value of each field at its corresponding offset.
+
+r[value.aggregate.union]
+A value of a union type consists of a sequence of bytes, corresponding to each value byte. The value bytes of a union are represented exactly.
+
+> [!NOTE]
+> When a union value is constructed or a field is read/written to, the value of that field is encoded or decoded appropriately.
+
+r[value.aggregate.padding-uninit]
+When a value of an aggregate is encoded, each padding byte is left as uninit
+
+> [!NOTE]
+> It is valid for padding bytes to hold a value other than uninit when decoded, and these bytes are ignored when decoding an aggregate.
+
+r[value.aggregate.tuple-array]
+The fields of a tuple or an array are the elements of that tuple or array.
+
+## Enum Values
+
+r[value.enum]
+
+r[value.enum.intro]
+An enum value corresponds to exactly one variant of the enum, and consists of the fields of that variant
+
+> [!NOTE]
+> An enum with no variants therefore has no values.
+
+r[value.enum.variant-padding]
+A byte is a padding byte in a variant `V` if the byte is not used for computing the discriminant, and the byte would be a padding byte in a struct consisting of the fields of the variant at the same offsets.
+
+r[value.enum.value-padding]
+A byte is a padding byte of an enum if it is a padding byte in each variant of the enum. A byte that is not a padding byte of an enum is a value byte.
+
+r[value.enum.repr]
+The representation of a value of an enum type includes the representation of each field of the variant at the appropriate offsets. When encoding a value of an enum type, each byte which is a padding byte in the variant is set to uninit. In the case of a [`repr(C)`][layout.repr.c.adt] or a [primitive-repr][layout.repr.primitive.adt] enum, the discriminant of the variant is represented as though by the appropriate integer type stored at offset 0.
+
+> [!NOTE]
+> Most `repr(Rust)` enums will also store a discriminant in the representation of the enum, but the exact placement or type of the discriminant is unspecified, as is the value that represents each variant.
--- a/src/values.md
+++ b/src/values.md
@@ -80,6 +80,17 @@ The representation of type `char` is the same as the representation of the `u32`
 r[value.primitive.bool]
 The two values of type `bool` are `true` and `false`. The representation of `true` is an initialized byte with value `0x01`, and the representation of `false` is an initialized  byte with value `0x00`. Neither value is represented with a pointer fragment.
 
+r[value.primitive.float]
+A floating-point value consists of either a rational number, which is within the range and precision dictated by the type, an infinity, or a NaN value.
+
+r[value.primitive.float-repr]
+A floating-point value is represented the same as a value of the unsigned integer type with the same width given by its [IEEE 754-2019] encoding. 
+
+r[value.primitive.float-format]
+The [IEEE 754-2019] `binary32` format is used for `f32`, and the `binary64` format is used for `f64`.
+
+[IEEE 754-2019]: https://ieeexplore.ieee.org/document/8766229
+
 ## Pointer Value
 
 r[value.pointer]


src/values.md
```diff
@@ -0,0 +1,186 @@
+# Values and Representation
+
+r[value]
+
+## Bytes
+
+r[value.byte]
+
+r[value.byte.intro]
+The Most basic unit of Memory in Rust is a Byte. All values in Rust are computed from 0 or more bytes read from an allocation.
```
PatchMixolydic, line 10: Nit:
```suggestion
The most basic unit of memory in Rust is a byte. All values in Rust are computed from 0 or more bytes read from an allocation.
```

src/values.md
```diff
@@ -0,0 +1,186 @@
...
+The representation of type `char` is the same as the representation of the `u32` corresponding to the Code Point Number encoding by the `char`.
+
+r[value.primitive.bool]
+The two values of type `bool` are `true` and `false`. The representation of `true` is an initialized byte with value `0x01`, and the representation of `false` is an initialized  byte with value `0x00`. Neither value is represented with a pointer fragment.
+
+r[value.primitive.float]
+A floating-point value consists of either a rational number, which is within the range and precision dictated by the type, an infinity, or a NaN value.
+
+r[value.primitive.float-repr]
+A floating-point value is represented the same as a value of the unsigned integer type with the same width given by its [IEEE 754-2019] encoding.
```
traviscross, line 87: We looked at this one sentence in particular in our lang-docs call today, and we were having a lot of trouble parsing it.  Perhaps you could say this another way for us to better understand the intent here.  Why does this need to reference "an unsigned integer type of the same width" at all, e.g.?

traviscross:
@rust-lang/opsem: We're interested in your review on this.  From the lang-docs side, we're particularly interested to confirm that this text is both correct from your perspective and is not making an new guarantees about the language.

traviscross:
cc @rust-lang/lang

src/values.md
```diff
@@ -0,0 +1,186 @@
...
+The [IEEE 754-2019] `binary32` format is used for `f32`, and the `binary64` format is used for `f64`.
+
+[IEEE 754-2019]: https://ieeexplore.ieee.org/document/8766229
+
+## Pointer Value
+
+r[value.pointer]
+
+r[value.pointer.thin]
+Each thin pointer consists of an address and an optional provenance. The address refers to which byte the pointer points to. The provenance refers to which bytes the pointer is allowed to access, and the allocation those bytes are within.
```
ehuss, line 99: This starts using the term *provenance* without defining what it means. Would it be possible to at least start that in the glossary? I'm also not sure if we'll need a more elaborate introduction, since rfc#3559 is quite weighty, so maybe there will need to be a more dedicated section for that?

src/values.md
```diff
@@ -0,0 +1,186 @@
+# Values and Representation
```
traviscross, line 1:
```suggestion
# Values and representation
```

src/SUMMARY.md
```diff
@@ -99,6 +99,7 @@
     - [Type coercions](type-coercions.md)
     - [Destructors](destructors.md)
     - [Lifetime elision](lifetime-elision.md)
+    - [Values and Representation](values.md)
```
traviscross, line 102:
```suggestion
    - [Values and representation](values.md)
```

ehuss (commented):
I'm feeling uncertain about this approach of having a chapter specifically for these representations. In the past, we have placed these definitions in the chapters for the types they are defining (for example, the char chapter defines the char). Can we move these rules into those chapters (and avoid any duplication with things that are already there)?

src/values.md
chorman0773: It's to inherit the definition of endianess from unsigned integer types (where its the easiest to define). Signed integers also do the same thing.

src/values.md
chorman0773: There will be eventually, but it also an incredibly weakly defined topic right now. The main definition we have today is frankly, "It exists". Anything more requires a ton more litigation from T-opsem.  Defining a byte requires referring to provenance, though, at least in the most abstract sense.

chorman0773:
> Can we move these rules into those chapters (and avoid any duplication with things that are already there)?

We still have to define what a byte is, which is relatively small for its own chapter, but doesn't have any other chapter to go in (glossary shouldn't provide normative definitions). It's also going to duplicate content excessively for aggregate types, since tuples and structs have different chapters, but have the exact same values and representation (according to a particular layout).

src/values.md
```diff
@@ -0,0 +1,186 @@
...
+The representation of a value of a thin pointer is a sequence of initialized bytes with `u8` values given by the representation of its address as a value of type `usize`, and pointer fragments corresponding to its provenance, if present.
+
+r[value.pointer.thin-ref]
+A thin reference to `T` consists of a non-null, well aligned address, and provenance for `size_of::<T>()` bytes starting from that address. The representation of a thin reference to `T` is the same as the pointer with the same address and provenance.
+
+> [!NOTE]
+> This is true for both shared and mutable references. There are additional constraints enforced by the aliasing model.
+
+r[value.pointer.fat]
+A fat pointer or reference consists of a data pointer or reference, and a pointee-specific metadata value.
```
joshtriplett, line 114: The Rust reference currently uses the term "wide" in many places, e.g. "wide reference", "wide pointer". The compiler does as well, such as in the lint `ambiguous_wide_pointer_comparisons` and in many places in the source code. We should be consistent, and use `valid.pointer.wide` here, unless there's some strong reason to do otherwise.

(In addition to consistency, "wide" also seems like more inclusive terminology.)

traviscross: As Josh noted, this was decided by FCP here:

- https://github.com/rust-lang/rust/pull/117758#issuecomment-1832381065

chorman0773:
Need to figure out how to disentangle the aggregate representation rules to move them into a separate chapter.

chorman0773:
Just talked about this on the Community Discord, apparently there is one guarantee this makes that is "new". Namely, it guarantees that wide pointers are represented like some repr(Rust) struct of the data pointer and its metadata. While this seems like it's doing nothing, one thing it does say is that the fields exist *somewhere* in the representation (and the compiler isn't doing far more interesting shenanigans than it might do to a tuple or a struct). 
Is this a problematic guarantee to make?

traviscross:
cc @rust-lang/opsem

src/values.md
Noratrieb: When the entire lang team is confused by a sentence, I think it's fair to say that it needs to be rewritten.

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+## Bytes
+
+r[memory.byte]
+
+r[memory.byte.intro]
+The most basic unit of memory in Rust is a byte. All values in Rust are computed from 0 or more bytes read from an allocation.
+
+> [!NOTE]
+> While bytes in Rust are typically lowered to hardware bytes, they may contain additional values,
+> such as being uninitialized, or storing part of a pointer.
```
RalfJung, line 16:
```suggestion
> While bytes in Rust are typically lowered to hardware bytes, Rust uses an "abstract"
> notion of bytes that can make distinctions which are absent in hardware,
> such as being uninitialized, or storing part of a pointer.
```

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+r[memory]
+
+The Memory Model of Rust is incomplete and not fully decided. The following is some of the detail worked out so far.
+
+## Bytes
+
+r[memory.byte]
+
+r[memory.byte.intro]
+The most basic unit of memory in Rust is a byte. All values in Rust are computed from 0 or more bytes read from an allocation.
```
RalfJung, line 12: Is "value" defined anywhere?

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+
+> [!NOTE]
+> While bytes in Rust are typically lowered to hardware bytes, they may contain additional values,
+> such as being uninitialized, or storing part of a pointer.
+
+r[memory.byte.init]
+Each byte may be initialized, and contain a value of type `u8`, as well as an optional pointer fragment. When present, the pointer fragment carries [provenance][type.pointer.provenance] information.
+
+r[memory.byte.uninit]
+Each byte may be uninitialized.
```
RalfJung, line 22: It seems quite odd to split the definition of a "byte" over two separate items. It is very crucial that this list is exhaustive, and cannot be extended by some other clause somewhere else saying "a byte may also be X". So IMO this should be a single paragaph saying something like:

Each byte is described by one of the following cases:
- It is uninitialized.
- It is initialized, ...

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+
+> [!NOTE]
+> Uninitialized bytes do not have a value and do not have a pointer fragment.
+
+## Value Encoding
+
+r[memory.encoding]
+
+r[memory.encoding.intro]
+Each type in Rust has 0 or more values, which can have operations performed on them
```
RalfJung, line 32: This seems to be setting up for "memory.encoding", but really it defines the notion of a "value". I think that notion deserves a separate definition, in a `r[memory.value]`, which can then be referenced elsewhere. I am not fully convinced putting this in "memory" is the best choice, since one defining characteristic of values is that they do *not* exist in memory, but I also don't know the structure well enough to be able to suggest a better place.

This should then also give some sort of definition of a value. Something like:

A *value* in Rust describes a high-level, "mathematical" view on data at a given type. Examples of values include:
- mathematical integers
- Boolean truth values
- A tuple of values of potentially different types
- A homogeneous list of values of identical type

This list is not exhaustive; as Rust evolves, more kinds of values can be added.

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+
+> [!NOTE]
+> `0u8`, `1337i16`, and `Foo{bar: "baz"}` are all values
+
+r[memory.encoding.op]
+Each value of a type can be encoded into a sequence of bytes, and decoded from a sequence of bytes, which has a length equal to the size of the type.
+The operation to encode or decode a value is determined by the representation of the type.
```
RalfJung, line 39: The representation relation does not determine encode/decode, since there can be more than one representation of a value -- encode makes the choice of which representation to pick. So it's the other way around, encode/decode determine the representation relation.

I would phrase this to be more centered around types, not values. Something like:

Each type defines an operation to *decode* a sequence of bytes into a value, and to *encode* a value into a sequence of bytes. When a sequence of bytes decodes to a value, we say that it *represents* that value. A value that is valid for a given type may be represented by multiple different sequences of bytes, and a sequence of bytes may represent 0 or 1 values of any given type.

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+
+> [!NOTE]
+> Representation is related to, but is not the same property as, the layout of the type.
+
+r[memory.encoding.decode]
+If a value of type `T` is decoded from a sequence of bytes that does not correspond to a defined value, the behavior is undefined. If a value of type `T` is decoded from a sequence of bytes that contain pointer fragments, which are not used to represent the value, the pointer fragments are ignored.
```
RalfJung, line 45: The second sentence should IMO not be mentioned here. This is not a defining property of our representation relation, it is just the choice we made for some representations.

Put differently: "pointer fragment which are not used to represent the value" is not a defined notion, and actually precisely defining this is a lot more work than just stating the relevant part with each type.

src/types/boolean.md
```diff
@@ -21,9 +21,10 @@ r[type.bool.layout]
 An object with the boolean type has a [size and alignment] of 1 each.
 
 r[type.bool.repr]
-The value false has the bit pattern `0x00` and the value true has the bit pattern
-`0x01`. It is [undefined behavior] for an object with the boolean type to have
-any other bit pattern.
+A `bool` is represented as a single initialized byte with a value of `0x00` corresponding to `false` and a value of `0x01` corresponding to `true`. This byte does not have a pointer fragment.
+
+> [!NOTE]
+> No other representations are valid for `bool`. Undefined Behaviour occurs when any other byte is read as type `bool`.
```
RalfJung, line 27: I would suggest to state this more like:

A `bool` is represented as a single initialized byte with a value of `0x00` corresponding to `false` and a value of `0x01` corresponding to `true`. When decoding a `bool`, this byte can have an arbitrary pointer fragment. When encoding a `bool`, a byte without a pointer fragment is used.

src/types/enum.md
```diff
@@ -14,14 +14,32 @@ unit-like struct.
...
 r[type.enum.name]
 Enum types cannot be denoted *structurally* as types, but must be denoted by
 named reference to an [`enum` item].
 
+## Enum values and representation
+
+r[type.enum.value]
+
+r[type.enum.value.intro]
+An enum value corresponds to exactly one variant of the enum, and consists of the fields of that variant
```
RalfJung, line 26:
```suggestion
An enum value corresponds to exactly one variant of the enum, and consists of the fields of that variant.
```

src/types/enum.md
```diff
@@ -14,14 +14,32 @@ unit-like struct.
...
+A byte is a padding byte in a variant `V` if the byte is not used for computing the discriminant, and the byte would be a padding byte in a struct consisting of the fields of the variant at the same offsets.
+
+r[type.enum.value.value-padding]
+A byte is a padding byte of an enum if it is a padding byte in each variant of the enum. A byte that is not a padding byte of an enum is a value byte.
+
+r[type.enum.value.repr]
+The representation of a value of an enum type includes the representation of each field of the variant at the appropriate offsets. When encoding a value of an enum type, each byte which is a padding byte in the variant is set to uninit. In the case of a [`repr(C)`][layout.repr.c.adt] or a [primitive-repr][layout.repr.primitive.adt] enum, the discriminant of the variant is represented as though by the appropriate integer type stored at offset 0.
+
+> [!NOTE]
+> Most `repr(Rust)` enums will also store a discriminant in the representation of the enum, but the exact placement or type of the discriminant is unspecified, as is the value that represents each variant.
```
RalfJung, line 41:
```suggestion
> Most `repr(Rust)` enums will also store a discriminant in the representation of the enum, but the exact placement or type of the discriminant is unspecified, as is the byte sequence that represents each variant.
```

src/types/enum.md
```diff
@@ -14,14 +14,32 @@ unit-like struct.
...
+> An enum with no variants therefore has no values.
+
+r[type.enum.value.variant-padding]
+A byte is a padding byte in a variant `V` if the byte is not used for computing the discriminant, and the byte would be a padding byte in a struct consisting of the fields of the variant at the same offsets.
+
+r[type.enum.value.value-padding]
+A byte is a padding byte of an enum if it is a padding byte in each variant of the enum. A byte that is not a padding byte of an enum is a value byte.
+
+r[type.enum.value.repr]
+The representation of a value of an enum type includes the representation of each field of the variant at the appropriate offsets. When encoding a value of an enum type, each byte which is a padding byte in the variant is set to uninit. In the case of a [`repr(C)`][layout.repr.c.adt] or a [primitive-repr][layout.repr.primitive.adt] enum, the discriminant of the variant is represented as though by the appropriate integer type stored at offset 0.
```
RalfJung, line 38: This could be defined without having to define padding bytes, by saying something like:

When encoding an enum value, first the fields of the active variant are encoded at the respective offsets. Bytes which are not contained in any field of the active variant are left uninitialized. Then, the encoding of the discriminant of the active variant is applied to the byte sequence.

src/types/function-pointer.md
```diff
@@ -55,6 +55,12 @@ let bo: Binop = add;
 x = bo(5,7);
 ```
 
+r[type.fn-pointer.value]
+A value of a function pointer type consists of an non-null address. A function pointer value is represented the same as an address represented as an unsigned integer type with the same width as the function pointer.
```
RalfJung, line 59:
```suggestion
A value of a function pointer type consists of a non-null address. A function pointer value is represented the same as an address represented as an unsigned integer type with the same width as the function pointer.
```

src/types/numeric.md
```diff
@@ -59,9 +59,36 @@ r[type.numeric.int.size.minimum]
...
-r[type.numeric.validity]
+r[type.numeric.repr]
 
-For every numeric type, `T`, the bit validity of `T` is equivalent to the bit
-validity of `[u8; size_of::<T>()]`. An uninitialized byte is not a valid `u8`.
+r[type.numeric.repr.integer]
+Each value of an integer type is a whole number. For unsigned types, this is a positive integer or `0`. For signed types, this can either be a positive integer, negative integer, or `0`.
+
+r[type.numeric.repr.integer-width]
+The range of values an integer type can represent depends on its signedness and its width, in bits. The width of type `uN` or `iN` is `N`. The width of type `usize` or `isize` is the value of the `target_pointer_width` property.
```
RalfJung, line 70: Shouldn't this say *how* it depends on that?

Specifically, unsigned integers carry values in `0 .. 2^N`, and signed integers carry values in `-2^(N-1) .. 2^(N-1)` (left-inclusive, right-exclusive).

src/types/struct.md
```diff
@@ -29,6 +29,32 @@ A _unit-like struct_ type is like a struct type, except that it has no fields.
 The one value constructed by the associated [struct expression] is the only
 value that inhabits such a type.
 
+## Struct and aggregate values
```
RalfJung, line 32: What is an "aggregate" here? I would have expected that structs and enums are aggregates, but enums are defined elsewhere and structs seem to be a separate class as well, so I am confused.

src/types/struct.md
```diff
@@ -29,6 +29,32 @@ A _unit-like struct_ type is like a struct type, except that it has no fields.
...
+
+r[type.struct.value]
```
RalfJung, line 34: This should say somewhere that a value of strict type is a list of values with one value for each field of the struct.

src/types/struct.md
```diff
@@ -29,6 +29,32 @@ A _unit-like struct_ type is like a struct type, except that it has no fields.
...
+
+r[type.struct.value.padding]
+Every byte in an aggregate which is not a value byte is a padding byte.
+
+r[type.struct.value.struct]
+A value of a struct type consists of the values of each of its fields.
+The representation of such a struct contains the representation of the value of each field at its corresponding offset.
+
+r[type.struct.value.padding-uninit]
+When a value of an aggregate is encoded, each padding byte is left as uninit
```
RalfJung, line 53: As above, this can IMO be defined more nicely without having to talk about padding at all:

To decode a struct value, decode each field at its respective offset in the byte sequence, and use that to form the decoded value. To encode a struct value, encode each field, and place the encoding at the respective offset, leaving all bytes that are not in any field uninitialized.

NOTE: the bytes that are not in any field are also often called "padding bytes". The representation defined above implies that the contents of padding bytes are lost and reset to uninitialized each time a typed copy of a struct value is performed.


(Have we defined "typed copy" yet?)

src/types/union.md
```diff
@@ -24,5 +24,17 @@ The memory layout of a `union` is undefined by default (in particular, fields do
 *not* have to be at offset 0), but the `#[repr(...)]` attribute can be used to
 fix a layout.
 
+r[type.union.value]
+A value of a union type consists of a sequence of bytes, corresponding to each [value byte][type.struct.value.value-bytes]. The value bytes of a union are represented exactly. Each [padding byte][type.struct.value.padding] is set to uninit when encoded.
```
RalfJung, line 28: Given that the fields of a union do not have to be structs, it is odd to refer to `type.struct` here. This definition requires there to be a general notion of "padding bytes in any type", and each type needs to say what its padding bytes are.

src/memory-model.md
```diff
@@ -1,5 +1,45 @@
...
+
+r[memory.byte.intro]
+The most basic unit of memory in Rust is a byte. All values in Rust are computed from 0 or more bytes read from an allocation.
+
+> [!NOTE]
+> While bytes in Rust are typically lowered to hardware bytes, they may contain additional values,
+> such as being uninitialized, or storing part of a pointer.
+
+r[memory.byte.init]
+Each byte may be initialized, and contain a value of type `u8`, as well as an optional pointer fragment. When present, the pointer fragment carries [provenance][type.pointer.provenance] information.
```
RalfJung, line 19: What is the difference between a "pointer fragment" and "provenance"? Wouldn't it be easier to just say, "[...] as well as optional provenance information"?

RalfJung:
> We're interested in your review on this. From the lang-docs side, we're particularly interested to confirm that this text is both correct from your perspective and is not making an new guarantees about the language.

This makes several new guarantees:
- We have never defined what Rust's concept of a "byte" is. Uninitialized memory is a thing (though AFAIK this was never explicitly FCP'd either), provenance is a thing, but this PR basically guarantees that that's it, and there's nothing else "odd" about our bytes. I can't immediately think of anything else, but e.g. maybe we need more extra magic state to give an abstract definition of the behavior of coroutines?
- We have not defined that reading a pointer byte at integer type just ignores the provenance. It'd probably be better if this PR didn't make a commitment here.

src/types/enum.md
chorman0773: It also needs to define padding of the enum as a whole for the purposes of unions, though.

src/types/struct.md
chorman0773: Tuples are aggregates also. The current definition of a wide pointer also ends up using aggregate representation.

src/types/struct.md
chorman0773: We don't really need to define typed copy for this -  a typed copy is just a decoding of some memory to a value, then re-encoding that value elsewhere. It may be useful to define non-normatively, or elsewhere, but we don't need to define it for this purpose. 
As with enums, we need to define padding anyways b/c unions.

src/memory-model.md
chorman0773: This used to be one chapter that fully defined value. I was asked to move those definitions to the respective type chapters to avoid duplication.

src/memory-model.md
chorman0773: Values are defined constructively, with each "class" of types. As mentioned in another comment, these are present in different chapters at the request of T-lang-doc and T-spec.

RalfJung:
I should also add that overall I am very happy with the structure and style of the new text here. :) I can finally see how this could hold together as a proper spec also for the operational aspects of the language.

src/types/enum.md
RalfJung: Yeah, that is true. I'm not a big fan of the entire concept of "padding" being so pervasive, IMO it is unnecessary... but I guess it is common and many people coming from C/C++ can use prior knowledge to understand it.

src/types/struct.md
RalfJung: Tuples have their own section on this, though. So I find the current structure a bit confusing.

src/types/struct.md
RalfJung: > It may be useful to define non-normatively, or elsewhere, but we don't need to define it for this purpose.

Fair. I feel like (non-normatively) stating the fact that padding gets reset can be useful for understanding though. So maybe we can have something like

NOTE: As a consequence of this definition, if some sequence of bytes is decoded and then re-encoded, all information stored in padding bytes is lost (reset to uninitialized) as part of this round-trip.

src/memory-model.md
RalfJung: Having the encode/decode defined with each type makes sense IMO. But the concept of a value is a central concept that should be defined somewhere. Having the list of "what is a value" spread out is fine (after all, my suggestion already said that the list is non-exhaustive), but the central definition should give *some* examples of what values are, just to make it understandable and to clearly differentiate it from "sequence of bytes".

chorman0773:
> We have never defined what Rust's concept of a "byte" is. Uninitialized memory is a thing (though AFAIK this was never explicitly FCP'd either), provenance is a thing, but this PR basically guarantees that that's it, and there's nothing else "odd" about our bytes. I can't immediately think of anything else, but e.g. maybe we need more extra magic state to give an abstract definition of the behavior of coroutines?

I don't believe this does confine the definition of byte to the current list. As you pointed out, the current definition (over two separate clauses) is not clearly exhaustive. I don't think there's anything else that would clearly prohibit a new fancy type of byte from coming into existance (though once we do affirm that, I hope to reflect that in the text). 

> We have not defined that reading a pointer byte at integer type just ignores the provenance. It'd probably be better if this PR didn't make a commitment here.

This part is fair, though part of me wants to just ask for a T-opsem FCP on that (could be done over on ucg though), as I don't think there's any other rule that satisfies monotonicity (other than PVI which has been ruled out through other means).

src/types/struct.md
chorman0773: Tuples would have the exact same set of values and the exact same representation formula (given the same chosen field offsets, size, and alignment constraints) as a struct with the same type fields, though, so this would just end up duplicating the same text. The tuple chapter does have a clause that explicitly delegates to this section, though, so it's not implicit.

chorman0773: The array section also use to delegate in the same way, but I changed it in a later revision to simply do it manually since unlike structs or tuples, arrays can't shallowly have embedded padding bytes (only padding that exists inside of an element because of the elements representation).

RalfJung: I agree with the goal of not duplicating this for tuples. I just think the way it is currently done is confusing.

Even if this here just says "Struct values", we can still say in the tuple section that they use the same values and encode/deocde as structs, can't we?

src/memory-model.md
RalfJung: We do want "machine-readable" names for the variants, fair. But making them separate paragraphs just feels entirely wrong. This should be a list, formatted like a list. Can we have lists where each item has a "name"?

src/memory-model.md
chorman0773: I'm not sure that there is a good place for the definition of value.

src/memory-model.md
chorman0773: The only thing is that most of the other chapters will define just representation to get both at once. I'd like to preserve this ability.

chorman0773: Although in most cases, simply defining the decoding should suffice as `decode(T, encode(T, val))` is defined to round-trip `val` later.

src/values.md
chorman0773: Reworking this now.

@RalfJung would you know what the state of NaNs are? For example, could I say that rust has One NaN which is encoded non-deterministically?

chorman0773: I attempted to word the correspondance better. I can't get rid of it entirely without just copying the same encoding/decoding step (which IMO is just more confusing and redundant).

src/memory-model.md
RalfJung: We can put it in memory-model.md for now. It should go together with encode/decode and those already live there.

src/memory-model.md
RalfJung: > The only thing is that most of the other chapters will define just representation to get both at once. I'd like to preserve this ability.

Formally speaking, that's sloppy. If we say that every type has encode and decode, we have to say explicitly how they are defined anywhere.

> Although in most cases, simply defining the decoding should suffice as decode(T, encode(T, val)) is defined to round-trip val later.

So then I would suggest that these types have something like
```
r[<>.repr.decode]
...

r[<>.repr.encode]
The encoding of a value is the unique sequence of bytes that decodes back to the value
```

However, note that for most types there is no unique such sequence (since decode often ignores provenance). With more setup we could define the encoding as the unique minimal sequence of bytes that decodes back to the value, where "minimal" uses an ordering that sets "uninit < init(no-provenance) < init(with-provenance)".

src/values.md
RalfJung: NaNs are fully stable bit patterns, see https://rust-lang.github.io/rfcs/3514-float-semantics.html.

So we can either say that `f32` has the same values as `u32` and leave the "float" part to the higher layers, or we need a concept of "float value" that exactly preserves the sign, quiet bit, and payload of a NaN (and the sign of a 0).

src/memory-model.md
```diff
@@ -1,5 +1,66 @@
...
+> `0u8`, `1337i16`, and `Foo{bar: "baz"}` are all values
+
+r[memory.encoding.op]
+Each type defines a pair of properties which, together, define the representation of values of the type. The *encode* operation takes a value of the type and converts it into a sequence of bytes equal in length to the size of the type, and the *decode* operation takes such a sequence of bytes and optionally converts it into a value. Encoding occurs when a value is written to memory, and decoding occurs when a value is read from memory.
+
+> [!NOTE]
+> Only certain byte sequences may decode into a value of a given type. For example, a byte sequence consisting of all zeroes does not decode to a value of a reference type.
+
+r[memory.encoding.representation]
+A sequence of bytes is said to represent a value of a type, if the decode operation for that type produces that value from that sequence of bytes. The representation of a type is the partial relation between byte sequences and values those sequences represent.
```
RalfJung, line 50:
```suggestion
A sequence of bytes is said to represent a value of a type, if the decode operation for that type produces that value from that sequence of bytes. The representation of a type is the partial relation between byte sequences and values those sequences represent.
This relation is functional, i.e., a given byte sequence represents at most one value.
```

RalfJung: If we do decide to define decoding as the unique minium value, we could possibly even reduce this definition to just say, for each type there is a representation function, which is a partial function that decodes a sequence of bytes into a corresponding value. And then we once-and-forall define the corresponding encode operation across all types.

src/memory-model.md
```diff
@@ -1,5 +1,66 @@
...
+
+> [!NOTE]
+> Representation is related to, but is not the same property as, the layout of the type.
```
RalfJung, line 53: I find this more confusing than helpful. What even exactly is the "layout" of a type, i.e. what is the mathematical type of a "layout"?

I am not sure what you are trying to achieve with this sentence. Here's my attempt at rewording:
```suggestion
> The representation of a type is determined by its layout. For instance, the layout of a struct defines the offsets of all its fields, and this in turn defines the representation of struct values as sequences of bytes.
```

chorman0773: I'm attempting to explain someting to less mathetmatically-focused consumers of the reference. This is a non-normative note and therefore serves to explain or exemplify something about the preceeding paragraph.

RalfJung: What I wrote wasn't even very mathematical. ;)

I find your current note confusing. I don't know what you are trying to say, so I made a guess. I am also fine with removing it entirely, but if you want to keep it then please explain which point you are trying to make.

src/types/boolean.md
```diff
@@ -21,9 +21,10 @@ r[type.bool.layout]
 An object with the boolean type has a [size and alignment] of 1 each.
 
 r[type.bool.repr]
-The value false has the bit pattern `0x00` and the value true has the bit pattern
-`0x01`. It is [undefined behavior] for an object with the boolean type to have
-any other bit pattern.
+A `bool` is represented as a single initialized byte with a value of `0x00` corresponding to `false` and a value of `0x01` corresponding to `true`.
+
+> [!NOTE]
+> No other representations are valid for `bool`. Undefined Behaviour occurs when any other byte is read as type `bool`.
```
bluebear94, line 27:
```suggestion
> No other representations are valid for `bool`. Undefined behavior occurs when any other byte is read as type `bool`.
```

Use American spellings as prescribed in the [style guide](https://github.com/rust-lang/reference/blob/master/docs/authoring.md#style).

src/types/enum.md
```diff
@@ -14,14 +14,31 @@ unit-like struct.
...
+> [!NOTE]
+> An enum with no variants therefore has no values.
+
+r[type.enum.value.value-padding]
+A byte is a [padding][type.union.value.padding] byte of an enum if that byte is not part of the representation of the discriminant of the enum, and in each variant it either:
+* Does not overlap with a field of the variant, or
+* Overlaps with a padding byte in a field of that variant.
+
+r[type.enum.value.repr]
+The representation of a value of an enum type includes the representation of each field of the variant at the appropriate offsets. When encoding a value of an enum type, each byte which is not use d to store a field of the variant or the discriminant is . In the case of a [`repr(C)`][layout.repr.c.adt] or a [primitive-repr][layout.repr.primitive.adt] enum, the discriminant of the variant is represented as though by the appropriate integer type stored at offset 0.
```
bluebear94, line 37:
```suggestion
The representation of a value of an enum type includes the representation of each field of the variant at the appropriate offsets. When encoding a value of an enum type, each byte that is not used to store a field of the variant or the discriminant is unspecified. In the case of a [`repr(C)`][layout.repr.c.adt] or a [primitive-repr][layout.repr.primitive.adt] enum, the discriminant of the variant is represented as the appropriate integer type stored at offset 0.
```

src/types/numeric.md
```diff
@@ -59,9 +59,46 @@ r[type.numeric.int.size.minimum]
...
+> [!NOTE]
+> There are exactly `1<<N` unique values of an integer type of width `N`.
+> In particular, for an unsigned type, these values are in the range `0..(1<<N)` and for a signed type, are in the range `-(1<<(N-1))..(1<<(N-1))`, using rust range syntax.
+
+r[type.numeric.repr.unsigned]
+A value `i` of an unsigned integer type `U` is represented by a sequence of initialized bytes, where the `m`th successive byte according to the byte order of the platform is `(i >> (m*8)) as u8`, where `m` is between `0` and the size of `U`. None of the bytes produced by encoding an unsigned integer has a pointer fragment.
+
+> [!NOTE]
+> The two primary byte orders are `little` endian, where the bytes are ordered from lowest memory address to highest, and `big` endian, where the bytes are ordered from highest memory address to lowest.
+> The `cfg` predicate `target_endian` indicates the byte order
```
bluebear94, line 81:
```suggestion
> The `cfg` predicate `target_endian` indicates the byte order.
```

src/types/numeric.md
```diff
@@ -59,9 +59,46 @@ r[type.numeric.int.size.minimum]
...
+
+> [!WARN]
+> On `little` endian, the order of bytes used to decode an integer type is the same as the natural order of a `u8` array - that is, the `m` value corresponds with the `m` index into a same-sized `u8` array. On `big` endian, however, the order is the opposite of this order - that is, the `m` value corresponds with the `size_of::<T>() - m` index in that array.
```
bluebear94, line 84:
```suggestion
> On `little` endian, the order of bytes used to decode an integer type is the same as the natural order of a `u8` array - that is, the value `m` corresponds with the index `m` into a `u8` array of the same size. On `big` endian, however, the order is the opposite of this order – that is, the value `m` corresponds with the index `size_of::<T>() - m` into that array.
```

src/types/numeric.md
```diff
@@ -59,9 +59,46 @@ r[type.numeric.int.size.minimum]
...
+
+r[type.numeric.repr.signed]
+A value `i` of a signed integer type with width `N` is represented the same as the corresponding value of the unsigned counterpart type which is congruent modulo `2^N`.
+
+> [!NOTE]
+> This encoding of signed integers is known as the 2s complement encoding.
```
bluebear94, line 90:
```suggestion
> This encoding of signed integers is known as *two's complement* encoding.
```

src/types/pointer.md
```diff
@@ -81,19 +81,60 @@ r[type.pointer.smart]
...
-Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
+r[type.pointer.value.thin]
+Each thin pointer consists of an address and an optional [provenance][type.pointer.provenance]. The address refers to which byte the pointer points to. The provenance refers to which bytes the pointer is allowed to access, and the allocation those bytes are within.
 
-r[type.pointer.validity.raw]
-For thin raw pointers (i.e., for `P = *const T` or `P = *mut T` for `T: Sized`),
-the inverse direction (transmuting from an integer or array of integers to `P`) is always valid.
-However, the pointer produced via such a transmutation may not be dereferenced (not even if `T` has size zero).
+> [!NOTE]
+> A pointer that does not have a provenance may be called an invalid or dangling pointer.
```
bluebear94, line 92: *Dangling pointer* is used differently elsewhere in the reference: [link to the `undefined.dangling` section](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#dangling-pointers).

src/types/pointer.md
```diff
@@ -81,19 +81,60 @@ r[type.pointer.smart]
...
+Whenever a pointer to a particular allocation is produced by using the reference or raw reference operators, or when a pointer is returned from an allocation function, the resulting pointer has provenance that refers to that allocation.
+
+> [!NOTE]
+> There is additional information encoded by provenance, but the exact scope of this information is not yet decided.
+
+r[type.pointer.provenance.dangling]
+A pointer is dangling if it has no provenance, or if it has provenance to an allocation that has since been deallocated. An access, except for an access of size zero, using a dangling pointer, is undefined behavior.
+
+> [!NOTE]
+> Allocations include local and static variables, as well as temporaries. Local Variables and Temporaries are deallocated when they go out of scope.
```
bluebear94, line 133:
```suggestion
> Allocations include local and static variables, as well as temporaries. Local Variables and temporaries are deallocated when they go out of scope.
```

src/types/union.md
```diff
@@ -24,5 +24,36 @@ The memory layout of a `union` is undefined by default (in particular, fields do
...
 
+## Union Values
+
+r[type.union.value]
+
+r[type.union.value.value-bytes]
+A byte `b` in the representation of a struct or union is a value byte if there exists a field of that aggregate such that:
+* The field has some type `T`,
+* The offset of that field `o` is such that `b` falls at an offset in `o..(o+size_of::<T>())`,
+* Either `T` is a primitive type or the offset of `b` within the field is not a padding byte in the representation of `T`.
```
bluebear94, line 35: Couldn’t this be simplified to the latter condition only if we specify that primitive types have no padding bytes?

src/memory-model.md
ia0: Do we need the encoding function to be deterministic? In other words, do we have existing code that relies on a value being encoded with a specific sequence of bytes (when that sequence is not uniquely defined by the decoding function)?

If we don't need to provide such guarantee, it seems like defining the decoding function is enough for all intents and purposes. Whenever a value needs to be encoded, a sequence of bytes that would decode to that value is chosen. If the programmer needs a specific sequence of bytes, they should cast the place to a type where the decoding function is injective (at least for that sequence of bytes) like `MaybeUninit<[u8; N]>` and avoid typed copies by only working on the place.

Note on the code suggestion itself: Stating that the relation is functional when it's defined by a function (the decode function) feels confusing to me because it's redundant. Maybe wording it like "This relation is functional by definition, i.e., a given byte sequence represents at most one value, the one returned by the decode operation."

ia0 (commented):
Should we try to split this PR to merge it incrementally? (like it was done with #1772)

The first PR would probably be `memory-model.md` and `types/boolean.md`. Most of the discussion would be in `memory-model.md` (since the representation relation of `bool` is fairly uncontroversial). That's still a big piece, but ideally won't involve difficulties we could encounter for unions and pointers.

What do you think?

src/types/union.md
ia0: What about the following general notion of padding byte?

A byte offset is a _padding byte_ for a type, if it is smaller than the size of the type and for all sequences of bytes of that length, changing the byte in the sequence at that offset doesn't change the behavior of the decode function.

Mathematically:
```
is_padding(t: Type, i: Int) :=
    i < size(t) &&
    forall(S: List<AbsByte>, b: AbsByte)
        len(S) == size(t) =>
        decode(t, S[i := b]) == decode(t, S)
```

Or does this definition contradicts existing usage?

EDIT: I initially tried to cover unsized types but then bytes outside the size of the type were also padding bytes. I restricted the definition to sized types. Not sure if we want to extend the definition to unsized types.

PR #1666, opened 2024-10-28
Rename "object safe" to "dyn compatible"

fmease:
Supersedes #1512: In line with [T-lang's latest resolution](https://github.com/rust-lang/lang-team/issues/286#issuecomment-2338905118).

Part of rust-lang/rust#130852.

r? @traviscross

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -173,12 +173,14 @@ the hierarchy has its own collection of named entities.
 ### Nominal types
 
 Types that can be referred to by a path directly. Specifically [enums],
-[structs], [unions], and [trait objects].
+[structs], [unions], and [trait object types].
 
-### Object safe traits
+### Dyn-compatible traits
 
-[Traits] that can be used as [trait objects]. Only traits that follow specific
-[rules][object safety] are object safe.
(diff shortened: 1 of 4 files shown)


fmease:
> std/keyword.dyn.html:13: broken link fragment `#object-safety` pointing to `reference/items/traits.html`
> cargo/print.html:9654: broken link fragment `#object-safety` pointing to `reference/items/traits.html`
> cargo/reference/semver.html:1354: broken link fragment `#object-safety` pointing to `reference/items/traits.html`

Alright, so am I meant to update the links in std & cargo *before* this PR gets merged? Apparently, linkchecker doesn't like the JS URL rewrites. If I were to update std & cargo first, wouldn't this mean that these links would actually be broken until the next r-l/rust book sync? I intentionally left all `#object-safe` links intact everywhere to avoid this exact scenario.

ehuss (approved):
Thanks!

I pushed some small changes, in particular to fix the linkchecker. When there are other books that link to something, we have to temporarily include an HTML tag to hold the identifier. I will eventually try to get the other books updated.

Issue #392, opened 2018-08-07
What is a trait object, exactly?


ksqsf:
The current wording is pretty confusing and needs refining.

> A trait object is an opaque value of another type ...

This suggests a trait object is a value.

> ... that implements a set of traits

This suggests a trait object is a type.

> For example, given a trait Trait, the following are all trait objects:
> * Trait
> * dyn Trait
> * ...

This suggests "dyn Trait" itself is a trait object.

>  Syntax TraitObjectType :    dyn? TypeParamBounds

This suggests a trait object is a value of type "dyn Trait".

After reading this section, I still don't know which part in
```rust
// rfc1909. not exactly sure this is correct.
let reader: dyn Read = File::open(path)?;
```
is the trait object here. Is it `reader`, `dyn Read`, or `File`?

Havvy:
`reader` is a variable that holds the trait object. `dyn Read` is the type of the trait object. `File` is what the trait object points to. And then the trait object itself is (currently implemented as) a struct that is a pointer to the File struct and a vTable for the Read methods and the size and alignment of the File type.

Though you'd have to do `Box::new(File::open(path)? as dyn Read)` or some other type that can hold a trait object because they are dynamically sized types so variables cannot hold them directly.

mzji:
I've found some comfusing definitions about trait object in different parts of Rust document.
In the document of standard library, `&dyn Trait` is trait object:
![k46u yhx4 lc_3 s 8xejo](https://user-images.githubusercontent.com/23279426/45545327-8993f300-b84c-11e8-8c7d-757ae25ac2d4.png)
But in the Rust Reference, `dyn Trait` is trait object:
![23h va0 5tgf b8vdfgraex](https://user-images.githubusercontent.com/23279426/45545368-9e708680-b84c-11e8-9396-f19b0e3d6df4.png)
The definition of trait object is just a mess, we need to fix that.

alercah:
Also, the non-auto (principal) trait no longer needs to come first, so this page is out of date, too.

PR #1670, opened 2024-11-05
trait object constraint correction

pazmank:
The non-auto trait does not have to be in the first position in a trait object, also mentioned [here](https://github.com/rust-lang/reference/pull/255#discussion_r170752509) and [here](https://github.com/rust-lang/reference/issues/392#issuecomment-614408780),

### DIFF
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -24,7 +24,7 @@ Trait objects are written as the keyword `dyn` followed by a set of trait
 bounds, but with the following restrictions on the trait bounds.
 
 r[type.trait-object.constraint]
-All traits except one trait must be auto traits, there may not be more than one
+There may not be more than one non-auto trait, no more than one
 lifetime, and opt-out bounds (e.g. `?Sized`) are not allowed. Furthermore,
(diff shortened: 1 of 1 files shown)


src/types/trait-object.md
```diff
@@ -24,7 +24,7 @@ Trait objects are written as the keyword `dyn` followed by a set of trait
 bounds, but with the following restrictions on the trait bounds.
 
 r[type.trait-object.constraint]
-All traits except the first trait must be auto traits, there may not be more than one
+All traits except one trait must be auto traits, there may not be more than one
```
ehuss, line 27: I'm having a little bit of trouble clearly understanding what this is saying, because another interpretation is that it is required to have one non-auto trait. I'm wondering if this could be rewritten to be a little more explicit what it means. Perhaps something like:

```suggestion
There may not be more than one non-auto trait, no more than one
```

ehuss (commented):
Thanks!

Looks like this was changed in https://github.com/rust-lang/rust/pull/59445 and was noted in #392.

traviscross (approved):

PR #1678, opened 2024-11-19
Added an additional example of lifetime elision

alex:
The question of how did this behave showed up in a rust-for-linux code review, and it would have been easier to explain had this example been present.

### DIFF
--- a/src/lifetime-elision.md
+++ b/src/lifetime-elision.md
@@ -48,6 +48,9 @@ fn get_mut2<'a>(&'a mut self) -> &'a mut dyn T;       // expanded
 fn args1<T: ToCStr>(&mut self, args: &[T]) -> &mut Command;                  // elided
 fn args2<'a, 'b, T: ToCStr>(&'a mut self, args: &'b [T]) -> &'a mut Command; // expanded
 
+fn other_args<'a>(arg: &T) -> &'a str;                // elided
+fn other_args<'a, 'b>(arg: &'b T) -> &'a str;         // expanded
+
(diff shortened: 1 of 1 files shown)


frewsxcv (approved):

traviscross:
Interesting observation, in testing to verify this, we ran across the fact that this still compiles...

```rust
trait Tr {}
impl<'a> Tr for fn(&()) -> &'a () {}
impl<'a, 'b> Tr for fn(&'b ()) -> &'a () {}
```

...due to:

- https://github.com/rust-lang/rust/issues/56105

We can verify these lifetimes are not equal, though, with:

```rust
fn lt_eq<'a: 'a, T>(_: &'a T) {}
fn f<'a>(x: &()) -> &'a () {
    lt_eq::<'a>(x);
    //~^ ERROR explicit lifetime required in the type of `x`
    loop {}
}
```

The error message there is somewhat less than optimal, as the problem isn't that the lifetime isn't named, but instead that the anonymous lifetime doesn't outlive `'a`.  It seems it would be better if it gave a similar error message to:

```rust
fn lt_eq<'a: 'a, T>(_: &'a T) {}
fn f<'a, 'b>(x: &'b ()) -> &'a () {
    lt_eq::<'a>(x);
    //~^ ERROR lifetime may not live long enough
    loop {}
}
```

Regardless, this change to the Reference looks good.  Thanks for the PR.

traviscross (approved):

alex:
Thanks for the quick merge!

PR #1679, opened 2024-11-21
Explicitly specify syntax forms for each attribute

chorman0773:
This adds a syntax rule to each builtin macro explaining the valid syntax forms for the macro.


ehuss:
I'm going to go ahead and close as superseded by the work to migrate to the attribute template. Thanks, though!

PR #1681, opened 2024-11-22
Fix rule for lazy boolean temporary drop scope

ehuss:
This was changed in https://github.com/rust-lang/rust/pull/103293 (Rust 1.67).


chorman0773 (approved):
Appears to be correct.

PR #1686, opened 2024-11-27
fix inconsistent spacing in example

meadowsys:
I noticed this inconsistency here, I hope this is alright that I submit this tiny fix >~<


ehuss (approved):
Thanks!

PR #1688, opened 2024-11-30
Claim to follow Unicode 16 for lexing identifiers.

mattheww:
This change happened in Rust 1.83.

As I understand it, rustc's behaviour here is determined by the versions of the `unicode-xid` and `unicode-normalization` crates it's built with.

These were updated to 0.2.6 and 0.1.24 respectively in
https://github.com/rust-lang/rust/pull/129624

Both of those versions update their character databases to Unicode 16.0.0.

I've checked that rustc 1.83 accepts the following (which rustc 1.82 doesn't):

````rust
let 𐗉 = "Todhri Letter Ei";
dbg!(𐗉);
````

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -19,7 +19,7 @@ r[ident.syntax]
 
 <!-- When updating the version, update the UAX links, too. -->
 r[ident.unicode]
-Identifiers follow the specification in [Unicode Standard Annex #31][UAX31] for Unicode version 15.0, with the additions described below. Some examples of identifiers:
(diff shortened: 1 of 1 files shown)


traviscross (approved):

ehuss:
I posted https://github.com/rust-lang/rust/pull/134043 to give a mechanism to notify us when this changes. This has fallen out of date many times in the past, and I think it would be nice to avoid that.

I'm not sure if that's really the best approach (I thought about using cron jobs to do it instead), but it seemed relatively easy to just add a test.

PR #1153, opened 2022-02-01
Add documentation for the `naked` function attribute

bstrie:
This PR updates the reference to fulfill the documentation requirement of the stabilization checklist for RFC 2972, tracked at https://github.com/rust-lang/rust/issues/90957 .

PR #1689, opened 2024-12-01
naked functions

folkertdev:
tracking issue: https://github.com/rust-lang/rust/issues/90957
earlier version: https://github.com/rust-lang/reference/pull/1153

Pending on https://github.com/rust-lang/rust/pull/128004. Once merged, we'll move for stabliziation.

I'm working off of the original PR to the reference, written 2+ years ago when the reference was less strict. I suspect some refinement and editing will be needed here to satisfy the quality requirements of the spec, and update the text to accurately reflect the current state and the exact guarantees/requirements.

Stabilization PR:

- https://github.com/rust-lang/rust/pull/134213

cc @bstrie @traviscross @Amanieu @bjorn3


src/attributes/codegen.md
```diff
@@ -51,6 +51,53 @@ r[attributes.codegen.cold]
...
+be unnecessary, such as when the function is inlined.
+
+An implementation can fulfill this requirement by guaranteeing that naked functions
+are never inlined. However, implementations are not currently required to guarantee that
+naked functions are never inlined.
+
+In the future it may become a requirement for implementations to guarantee that
+naked functions are never inlined; users must not rely on any observable behavior
+that may result from inlining. according to the specified calling convention before
+executing a naked function,
```
Amanieu, line 82: I don't think we should be talking about inlining in the reference since that's a specific compiler optimization. Instead we should say what we guarantee:
- The asm code will have a valid call stack and register state on entry as per the signature and calling convention of the function.
- The asm code may not be duplicated by the compiler. This is important for naked functions that define symbols in the assembly code.

src/inline-assembly.md
```diff
@@ -578,9 +587,13 @@ r[asm.options.checks.pure]
 r[asm.options.checks.noreturn]
 - It is a compile-time error to specify `noreturn` on an asm block with outputs.
 
+r[asm.options.naked_asm-restriction]
+`global_asm!` only supports the `att_syntax` and `raw` options.
```
Amanieu, line 591:
```suggestion
`naked_asm!` only supports the `att_syntax` and `raw` options.
```

src/inline-assembly.md
```diff
@@ -693,6 +706,44 @@ r[asm.rules.x86-prefix-restriction]
...
+r[asm.naked-rules.intro]
+To avoid undefined behavior, these rules must be followed when using function-scope inline assembly in naked functions (`naked_asm!`):
+
+r[asm.naked-rules.reg-not-input]
+- Any registers not used for function inputs according to the calling convention and function signature will contain an undefined value on entry to the asm block.
+  - An "undefined value" in the context of inline assembly means that the register can (non-deterministically) have any one of the possible values allowed by the architecture.
+    Notably it is not the same as an LLVM `undef` which can have a different value every time you read it (since such a concept does not exist in assembly code).
+
+r[asm.naked-rules.reg-not-output]
+- Any callee-saved registers must have the same value upon exiting the asm block as they had on entry, otherwise behavior is undefined.
```
Amanieu, line 722: "exiting the asm block" is a bit vague here since you can exit by calling another function. This rule is specifically about the expected state when jumping to the return address provided by the calling convention.

src/inline-assembly.md
```diff
@@ -693,6 +706,44 @@ r[asm.rules.x86-prefix-restriction]
...
+  - Caller-saved registes may be used freely, even if they are not used for the return value.
+
+r[asm.naked-rules.unwind]
+- Behavior is undefined if execution unwinds out of an asm block.
+  - This also applies if the assembly code calls a function which then unwinds.
+
+r[asm.naked-rules.noreturn]
+- Behavior is undefined if execution falls through to the end of the asm block.
```
Amanieu, line 730: Maybe add a sub-point that the asm code is expected to provide its own return instruction.

src/inline-assembly.md
```diff
@@ -693,6 +706,44 @@ r[asm.rules.x86-prefix-restriction]
...
+r[asm.naked-rules.mem-same-as-ffi]
+- The set of memory locations that assembly code is allowed to read and write are the same as those allowed for an FFI function.
+  - Refer to the unsafe code guidelines for the exact rules.
+  - These rules do not apply to memory which is private to the asm code, such as stack space allocated within the asm block.
+
+r[asm.naked-rules.black-box]
+- The compiler cannot assume that the instructions in the asm are the ones that will actually end up executed.
+  - This effectively means that the compiler must treat the `naked_asm!` as a black box and only take the interface specification into account, not the instructions themselves.
+  - Runtime code patching is allowed, via target-specific mechanisms.
+  - However there is no guarantee that each `naked_asm!` directly corresponds to a single instance of instructions in the object file: the compiler is free to duplicate or deduplicate `naked_asm!` blocks.
```
Amanieu, line 741: This is not true, naked functions cannot be duplicated/merged, unlike `asm!`.

src/inline-assembly.md
```diff
@@ -693,6 +706,44 @@ r[asm.rules.x86-prefix-restriction]
...
+
+r[asm.naked-rules.not-exactly-once]
+- You cannot assume that an `naked_asm!` block will appear exactly once in the output binary.
+  The compiler is allowed to instantiate multiple copies of the `naked_asm!` block, for example when the function containing it is inlined in multiple places.
```
Amanieu, line 745: Same here.

src/inline-assembly.md
chorman0773: I think that poses problems on SPIR-V or NVPTX targets. I seem to recall that the loader just inlines the functions at the Shader Assembly level.

folkertdev: @chorman0773 just to be sure, you're referring to Amanieu's comment right?

What we do in rustc (or well, we will once that PR makes it through the queue) is that a naked function

```rust
#[naked]
extern "C" fn foo() {}
```

is emitted as something similar to

```rust
core::arch::global_asm!( 
    "foo:",
    "ret"
);

extern "C" {
    fn foo();
}
```

and we're relying on the compiler (rustc, llvm) not duplicating that bit of global assembly. I suspect we already rely on that in general, because global assembly can define symbols today.

If you still think this is problematic, how could we verify the behavior of this new codegen strategy for the targets you list?

src/inline-assembly.md
```diff
@@ -693,6 +706,40 @@ r[asm.rules.x86-prefix-restriction]
...
+r[asm.naked-rules.noreturn]
+- Behavior is undefined if execution falls through to the end of the asm block.
+    - the assembly code is expected to contain a return instruction or to diverge
+
+r[asm.naked-rules.mem-same-as-ffi]
+- The set of memory locations that assembly code is allowed to read and write are the same as those allowed for an FFI function.
+  - Refer to the unsafe code guidelines for the exact rules.
+  - These rules do not apply to memory which is private to the asm code, such as stack space allocated within the asm block.
+
+r[asm.rules.black-box]
```
bjorn3, line 738:
```suggestion
r[asm.naked-rules.black-box]
```

otherwise there are two rules with the same name. Also I'm surprised this rule isn't already specified globally for all kinds of asm rather than just for function local inline asm.

folkertdev: the impression I get is that `global_asm!` was never properly added: there is no example and the rules are specified for `asm!` and you have to read between the lines to figure out what is relevant for `global_asm!`.

src/inline-assembly.md
```diff
@@ -61,6 +62,7 @@ option := "pure" / "nomem" / "readonly" / "preserves_flags" / "noreturn" / "nost
 options := "options(" option *("," option) [","] ")"
 operand := reg_operand / clobber_abi / options
 asm := "asm!(" format_string *("," format_string) *("," operand) [","] ")"
+naked_asm := "asm!(" format_string *("," format_string) *("," operand) [","] ")"
```
pthariensflame, line 65:
```suggestion
naked_asm := "naked_asm!(" format_string *("," format_string) *("," operand) [","] ")"
```

src/inline-assembly.md
```diff
@@ -1331,6 +1344,70 @@ r[asm.rules.x86-prefix-restriction]
...
+            ".cfi_endproc",
+            function = sym function_that_panics,
+        )
+    }
+}
+
+extern "C-unwind" fn function_that_panics() {
+    panic!("unwind!");
+}
+# }
```
folkertdev, line 1408: I don't think I can currently run this because `#[naked]` is still unstable. So reminder to self: update this before merging.

cc @traviscross

traviscross: We won't merge this anyway until the stabilization lands, so you can just update it and leave the CI red until then.  That's what we normally do.

src/attributes/codegen.md
```diff
@@ -49,10 +49,48 @@ r[attributes.codegen.cold]
 The *`cold` [attribute]* suggests that the attributed function is unlikely to
 be called.
 
+r[attributes.codegen.naked]
+## The `naked` attribute
+
+r[attributes.codegen.naked.intro]
+The *`naked` [attribute]* prevents the compiler from emitting a function prologue and
+epilogue for the attributed function.
```
traviscross, line 57: We're unwrapping lines on new things added, so go ahead and unwrap these all if you would.

src/inline-assembly.md
```diff
@@ -2,10 +2,11 @@ r[asm]
 # Inline assembly
 
 r[asm.intro]
-Support for inline assembly is provided via the [`asm!`] and [`global_asm!`] macros.
+Support for inline assembly is provided via the [`asm!`], [`naked_asm!`] and [`global_asm!`] macros.
```
joshtriplett, line 5:
```suggestion
Support for inline assembly is provided via the [`asm!`], [`naked_asm!`], and [`global_asm!`] macros.
```

src/inline-assembly.md
```diff
@@ -384,8 +390,11 @@ assert_eq!(y, 1);
 # }
 ```
 
+r[asm.operand-type.naked_asm-restriction]
+Because `naked_asm!` defines a whole function body, it can only use `sym` and `const` operands.
```
joshtriplett, line 394:
```suggestion
Because `naked_asm!` defines a whole function body and the compiler cannot provide any additional code to handle operands, it can only use `sym` and `const` operands.
```

src/inline-assembly.md
folkertdev: Locally I still even get errors like this

```
Couldn't compile the test.
---- inline-assembly.md - Inline_assembly::Options (line 1360) stdout ----
error[E0658]: label operands for inline assembly are unstable
 --> inline-assembly.md:1364:32
  |
6 |       core::arch::asm!("jmp {}", label {
  |  ________________________________^
7 | |         println!();
8 | |     }, options(noreturn));
  | |_____^
  |
  = note: see issue #119364 <https://github.com/rust-lang/rust/issues/119364> for more information
```

with these versions in the `reference` folder (so I don't think there is a rustup override active)

```
> cargo --version
cargo 1.88.0-nightly (d811228b1 2025-04-15)
> rustc --version
rustc 1.88.0-nightly (191df20fc 2025-04-18)
```

That feature has been stable on nightly for a while, so I must be missing something. Anyway, I believe the current code to be correct, and CI will catch it before merge if there are minor issues.

traviscross: I've had trouble with overrides too on the Reference for reasons I haven't yet worked out.  Perhaps if you `rustup default nightly` it will work for you.

src/attributes/codegen.md
```diff
@@ -46,6 +46,39 @@ r[attributes.codegen.cold]
...
+The *`naked` [attribute]* prevents the compiler from emitting a function prologue and epilogue for the attributed function.
+
+r[attributes.codegen.naked.body]
+The [function body] must consist of exactly one [`naked_asm!`] macro invocation.
+
+r[attributes.codegen.naked.prologue-epilogue]
+No function prologue or epilogue are generated for the attributed function: the contents of the `naked_asm!` invocation make up the full body of a naked function.
+
+r[attributes.codegen.naked.unsafe-attribute]
+The `naked` attribute is an [unsafe attribute]. Annotating a function with `#[unsafe(naked)]` comes with the safety obligation that the body respects the function's calling convention, and that the body either returns or diverges.
```
traviscross, line 62: > ...and that the body either returns or diverges

Is it correct that this should be removed since we've determined that unwinding is OK?

src/inline-assembly.md
```diff
@@ -1362,6 +1384,67 @@ r[asm.rules.preserves_flags]
...
+To avoid undefined behavior, these rules must be followed when using function-scope inline assembly in naked functions (`naked_asm!`):
+
+r[asm.naked-rules.reg-not-input]
+- Any registers not used for function inputs according to the calling convention and function signature will contain an undefined value on entry to the `naked_asm!` block.
+  - An "undefined value" in the context of inline assembly means that the register can (non-deterministically) have any one of the possible values allowed by the architecture.
+    Notably it is not the same as an LLVM `undef` which can have a different value every time you read it (since such a concept does not exist in assembly code).
+
+r[asm.naked-rules.reg-not-output]
+- Any callee-saved registers must have the same value upon return as they had on entry, otherwise behavior is undefined.
+  - Caller-saved registes may be used freely, even if they are not used for the return value.
```
traviscross, line 1400: What's the meaning of the caveat here,

> even if they are not used for the return value.

i.e., why might someone expect that they could only be used freely only if they were used for the return value?

src/attributes/codegen.md
traviscross: Nevermind; we mean that it doesn't fall through.  I'm doing some editing; I'll clarify here.

src/inline-assembly.md
```diff
@@ -1362,6 +1384,67 @@ r[asm.rules.preserves_flags]
...
+
+r[asm.naked-rules.noreturn]
+- Behavior is undefined if execution falls through to the end of the `naked_asm!` block.
+    - every path through the assembly code is expected to terminate with a return instruction or to diverge
+
+r[asm.naked-rules.mem-same-as-ffi]
+- The set of memory locations that assembly code is allowed to read and write are the same as those allowed for an FFI function.
+  - Refer to the unsafe code guidelines for the exact rules.
```
traviscross, line 1408: I know this was preexisting in another section, but given the way the unsafe code guidelines document describes itself, it seems a bit strong to refer to it for "the exact rules".  I'm planning to just pull this out in an edit.

src/inline-assembly.md
```diff
@@ -1362,6 +1384,67 @@ r[asm.rules.preserves_flags]
...
+  - These rules do not apply to memory which is private to the assembly code, such as stack space allocated within the `naked_asm!` block.
```
traviscross, line 1409: I'm also deleting this line.  Where this was copied from, "these" meant rules about `readonly` and `nomem`, but here, "these" seems to refer to the UCG rules, and that doesn't make sense I think.

traviscross (approved):
OK, with the changes I've pushed, this look OK to me, so we can approve this to unblock the stabilization.

folkertdev:
Those final changes look good to me. The feature has been stabilized and CI passes here, so I think that should be everything.

src/inline-assembly.md
folkertdev: well if the register is used for the return value, any manipulations that you'd perform on that register before then would be overwritten, and hence not observable from the outside.

Changes to other caller-saved registers can be observed by the caller, but the calling convention says that the caller cannot rely on the value in the registers.

So I think this just re-states a property of caller-saved registers.

traviscross:
Thanks to @folkertdev for pushing this forward.

PR #1691, opened 2024-12-03
Fix placement of the weak keyword table

Veykril:


ehuss:
Thanks! I have a PR at #1685 which should fix the table, along with several other mistakes.

PR #1702, opened 2024-12-20
Document Rust 2024 match ergonomics reservations

traviscross:
We're adopting two changes to match ergonomics in Rust 2024 that have the effect of reserving language space:

- Rule 1C: When the DBM is not `move` (whether or not behind a reference), writing `mut`, `ref`, or `ref mut` on a binding is an error.

- Rule 2C: Reference patterns can only match against references in the scrutinee when the DBM is `move`.

Here, we document those changes in the Reference and describe the differences between editions.

cc @ehuss @Nadrieril @Jules-Bertholet

Tracking:

- https://github.com/rust-lang/rust/issues/131414


Nadrieril (approved):

ehuss (approved):
Thanks!

PR #1703, opened 2024-12-21
modules: describe both module filename styles without giving a clear preference

RalfJung:
There's currently no clear consensus on which filename style to use for modules, and many people have a style tha mixes both options. So let's make the reference just describe the facts, without saying which variant is preferred.

Also see the [discussion on IRLO](https://internals.rust-lang.org/t/the-module-scheme-module-rs-file-module-folder-instead-of-just-module-mod-rs-introduced-by-the-2018-edition-maybe-a-little-bit-more-confusing/21977).

### DIFF
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -79,7 +79,7 @@ Module Path               | Filesystem Path  | File Contents
 `crate::util`             | `util.rs` *or* `util/mod.rs` | `mod config;`
 `crate::util::config`     | `util/config.rs` |
 
-> **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the way to load
+> **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the only way to load
 > a module with nested children.
 
 ### The `path` attribute
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -80,7 +80,7 @@ Module Path               | Filesystem Path  | File Contents
 `crate::util::config`     | `util/config.rs` |
 
 > **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the only way to load
-> a module with nested children.
+> a module with nested children. Rust 2018 added the new naming convention to be more consistent with modules that don't have submodules, and to avoid having many files named `mod.rs`.
 
 ### The `path` attribute
 
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -65,27 +65,22 @@ not have a `path` attribute, the path to the file mirrors the logical [module
 path].
 
 r[items.mod.outlined.search]
-Ancestor module path components are directories, and the module's
-contents are in a file with the name of the module plus the `.rs` extension.
+Ancestor module path components are directories, and the module's contents are in a file with the
+name of the module plus the `.rs` extension. Alternatively, the module's contents can also be in a
+file called `mod.rs` in a directory with the name of the module. It is not allowed to have both
+`$name.rs` and `$name/mod.rs`.
+
 For example, the following module structure can have this corresponding
 filesystem structure:
 
 Module Path               | Filesystem Path  | File Contents
 ------------------------- | ---------------  | -------------
 `crate`                   | `lib.rs`         | `mod util;`
-`crate::util`             | `util.rs`        | `mod config;`
+`crate::util`             | `util.rs` *or* `util/mod.rs` | `mod config;`
 `crate::util::config`     | `util/config.rs` |
 
-r[items.mod.outlined.search-mod]
-Module filenames may also be the name of the module as a directory with the
-contents in a file named `mod.rs` within that directory. The above example can
-alternately be expressed with `crate::util`'s contents in a file named
-`util/mod.rs`. It is not allowed to have both `util.rs` and `util/mod.rs`.
-
 > **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the way to load
-> a module with nested children. It is encouraged to use the new naming
-> convention as it is more consistent, and avoids having many files named
-> `mod.rs` within a project.
+> a module with nested children.
 
 ### The `path` attribute
 
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -77,7 +77,7 @@ Module Path               | Filesystem Path  | File Contents
 ------------------------- | ---------------  | -------------
 `crate`                   | `lib.rs`         | `mod util;`
 `crate::util`             | `util.rs` *or* `util/mod.rs` | `mod config;`
-`crate::util::config`     | `util/config.rs` |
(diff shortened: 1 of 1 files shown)


joshtriplett:
I think changing the expressed preference would require a consensus to change it. There was an intentional decision in 2018 to try to migrate towards the Rust 2018 style.

RalfJung:
I didn't realize that this was an intentional decision. There's little other indication of a goal of pushing the ecosystem towards the 2018 style. Not even the compiler itself was ever systematically adjusted.

So, that would be a t-lang FCP then?

joshtriplett:
@RalfJung IIRC, in 2018 when we designed the new module system, it was at *that* time an intentional decision to pointedly present it as "the new way" by contrast with "the old way". We didn't particularly *push* people to switch.

I'm not suggesting that we would necessarily *object* to this proposal, but I do think it's fair to say that this proposal represents a change in recommendations, and it probably warrants at least a discussion.

traviscross:
As some archaeology, this went in with:

- https://github.com/rust-lang/reference/pull/450

At the time, @ehuss said:

> I wouldn't say it is deprecated. The RFC explicitly says "The use of mod.rs continues to be allowed without any deprecation." To me, deprecated usually implies a lint that triggers, but in this case not even the edition-idiom lints recommend this. I carefully chose to say it is "encouraged" though I don't think anyone has officially established even that.

The tracking issue, where the FCP happened, was:

- https://github.com/rust-lang/rust/issues/53125

There's some discussion between then-lang members there suggesting that deprecation would have run into blocking objections.

The RFC was:

- https://github.com/rust-lang/rfcs/pull/2126

What that RFC says specifically is:

> The use of `mod.rs` continues to be allowed without any deprecation. It is expected that tooling like Clippy will push for at least style consistency within a project, and perhaps eventually across the ecosystem.

src/items/modules.md
```diff
@@ -65,27 +65,22 @@ not have a `path` attribute, the path to the file mirrors the logical [module
...
+`crate::util`             | `util.rs` *or* `util/mod.rs` | `mod config;`
 `crate::util::config`     | `util/config.rs` |
 
-r[items.mod.outlined.search-mod]
-Module filenames may also be the name of the module as a directory with the
-contents in a file named `mod.rs` within that directory. The above example can
-alternately be expressed with `crate::util`'s contents in a file named
-`util/mod.rs`. It is not allowed to have both `util.rs` and `util/mod.rs`.
-
 > **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the way to load
```
joshtriplett, line 82:
```suggestion
> **Note**: Prior to `rustc` 1.30, using `mod.rs` files was the only way to load
```

src/items/modules.md
```diff
@@ -65,27 +65,22 @@ not have a `path` attribute, the path to the file mirrors the logical [module
...
-> a module with nested children. It is encouraged to use the new naming
-> convention as it is more consistent, and avoids having many files named
-> `mod.rs` within a project.
+> a module with nested children.
```
joshtriplett, line 83:
```suggestion
> a module with nested children. Rust 2018 added the new naming convention to be more consistent with modules that don't have submodules, and to avoid having many files named `mod.rs`.
```

This keeps the description of why the new convention was added, but frames it as "Rust 2018 added" rather than "It is encouraged".

RalfJung: Giving rationale for the new style without giving rationale for the old style would still constitute an implicit endorsement.

So, I prefer your wording over the status quo, but I don't think it is as impartial as the reference should be.

joshtriplett: I don't think the reference is required to be impartial between two ways of writing things.

In the future, when we have better alternatives for every use of `as`, I would want the reference to position `as` as the *old* way of doing things and to editorialize on the new ways being better.

ia0: What about using a `Note`? That way the reference stays descriptive, but still documents rationale or recommendation. Those notes are "useful but out of scope" information (as defined by the convention), thus giving the opportunity to be prescriptive for those who are interested.

traviscross:

We reviewed this in lang triage today.  We agreed that the 2018 lang consensus was indeed to encourage one way (`foo.rs`) over the other way (`foo/mod.rs`), while at the same time not deprecating the old way, and we did not have any consensus for changing that.

src/items/modules.md
RalfJung: If this proposed edit is the only way to get the PR in, I'm willing to accept that. I don't plan to spend a lot of energy pushing on this.

src/items/modules.md
```diff
@@ -65,27 +65,22 @@ not have a `path` attribute, the path to the file mirrors the logical [module
...
+
 For example, the following module structure can have this corresponding
 filesystem structure:
 
 Module Path               | Filesystem Path  | File Contents
 ------------------------- | ---------------  | -------------
 `crate`                   | `lib.rs`         | `mod util;`
-`crate::util`             | `util.rs`        | `mod config;`
+`crate::util`             | `util.rs` *or* `util/mod.rs` | `mod config;`
 `crate::util::config`     | `util/config.rs` |
```
joshtriplett, line 80: This says "can have", so I realize that it doesn't have to show every possibility. However, it seems like *either* it should consistently show only one case (dropping the addition of ``` *or* `util/mod.rs` ```, or it should show the second case for `crate::util::config` as well (adding ``` *or* `util/config/mod.rs` ```).

I would personally favor the former, but if you want to add the `mod.rs` version to this table, please add it to both lines.

ehuss:
> we did not have any consensus for changing that.

I'm uncertain how to interpret this. Are you saying that you do or do not want to proceed with this change? Or is there some other change you are suggesting?

My interpretation is that the current text encourages `foo.rs`, but does **not** imply a deprecation of `mod.rs`.

traviscross:
We did not have any current consensus to change the prior consensus, which was to encourage `foo.rs` as the "new" way while not deprecating `foo/mod.rs`.

That is, the lang team's standing normative (in the sense of setting norms, but not in a way encoded in the language) position is what is currently written in the Reference.

traviscross:
Thanks @RalfJung for putting this forward.  We talked about this on the lang-docs call, then on the spec call.  We're going to close this for a kind of stylistic reason.

In the Reference, we document the "most recent" Rust in the main flow of text, and then we document older things we support.  Most prominently, this is what we do for editions.  Here, it's a closer call, but given the standing lang consensus about issuing a normative (in the sense of setting norms) position on this, it falls in the same bucket for us, and so in that way, the existing text is OK.

Kinrany:
Instead of removing the encouragement altogether, perhaps it's possible to either amend it to mention significant controversy or to replace it with an indirect encouragement that states the motivation behind the current consensus and points to its discussion?

PR #1704, opened 2024-12-24
Correct repr(Rust) overlapping fields rule for enums and unions

GKFX:
It is currently specified that the fields of a `repr(Rust)` union do not overlap, and that all fields of an enum do not overlap, which is not true. It looks like this line in the reference was written for structs only, but it actually applies to all three kinds of type so needs adjustment.


chorman0773:
This is handled in #1654 btw.

PR #1709, opened 2025-01-06
Fix unclosed `<sup>` elements

max-heller:
Changes several instances of `<sup><sup>` (two unclosed `<sup>` elements) to `<sup></sup>` (one, properly closed `<sup>` element), matching the rest of the instances in the source e.g.: https://github.com/rust-lang/reference/blob/acd6794e712d5e2ef6f5c84fb95688d32a69b816/src/attributes.md?plain=1#L5

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -3,7 +3,7 @@
 r[ident]
 
 r[ident.syntax]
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Thanks!

PR #1710, opened 2025-01-08
move r[rules] to the left of the main body, using a grid

WaffleLapkin:
See discussion in zulip: <https://rust-lang.zulipchat.com/#narrow/channel/399173-t-spec/topic/Rule.20readability.20in.20the.20reference/near/492412172>

This requires some quite awkward CSS changes, but it _does_ work and does look a lot better...

### DIFF
--- a/mdbook-spec/src/rules.rs
+++ b/mdbook-spec/src/rules.rs
@@ -105,7 +105,7 @@ impl Spec {
                 }
                 format!(
                     "<div class=\"rule\" id=\"r-{rule_id}\">\
-                     <a class=\"rule-link\" href=\"#r-{rule_id}\">[{rule_id_broken}]</a>\
+                     <a class=\"rule-link\" href=\"#r-{rule_id}\"><span>[{rule_id_broken}]<span/></a>\
                      {test_html}\
                      </div>\n",
                      rule_id_broken = rule_id.replace(".", "<wbr>."),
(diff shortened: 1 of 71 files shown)


ehuss (approved):
Thanks, I appreciate your help with this!

I pushed an update to support h5 and h6 headers. I think I got it right, though I'm not sure. It at least looks ok.

I'm a little concerned about the complexity of the CSS here, but it seems to work pretty well. (I get jittery with any CSS beyond the most basics...)

WaffleLapkin:
`h5`/`h6` support looks good to me.

> I'm a little concerned about the complexity of the CSS here, but it seems to work pretty well. (I get jittery with any CSS beyond the most basics...)

Yea, I'm not very happy with the complexity either. But I don't really see easier ways to achieve the same.

(things would be much easier if we didn't try to display rule names though)

ehuss:
@WaffleLapkin This seems to have broken the rendering for the tests links. Is that something you can look at fixing?
See https://doc.rust-lang.org/1.84.0/reference/input-format.html#r-input.shebang.inner-attribute for how this used to work.

If you look at https://doc.rust-lang.org/nightly/reference/input-format.html on a narrow screen, it does weird wrapping (or even on a wide screen, it looks a little weird). The chevrons also seem to appear twice, and the links are underneath the left page navigation. The popup is also in the wrong location. 

I'm wondering if maybe when it is in wide mode, it could link underneath the rule name? Maybe something like:
image

And then have the popup show towards the right?

I'm not sure what to do in narrow mode. I can't think of a good way to link it. Do you have any ideas?

As a side note, in narrow mode, the `[*]` cuts off the leftmost `[`, and ends up look a little weird.
image

I'm wondering if there is a way to prevent that? I don't really understand how the layout works here.

WaffleLapkin:
@ehuss will look into this next week.

PR #1713, opened 2025-01-14
doc `cenum_impl_drop_cast`

ickk:
Add note about `as` casts being disallowed for enums that implement `Drop`.

The `cenum_impl_drop_cast` lint rust-lang/rust#73333 is already deny on stable rust, and I think eventually this will become a hard restriction for enum-to-numeric casts. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1db0b0e8e9efd94b60c06dd20b829aad)

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -586,6 +586,7 @@ halfway between two floating point numbers.
 r[expr.as.enum]
 #### Enum cast
 
+r[expr.as.enum.discriminant]
 Casts an enum to its discriminant, then uses a numeric cast if needed.
 Casting is limited to the following kinds of enumerations:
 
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! We generally are hesitant to document these future-incompat things until they are actually made into a hard error. In which case, it would be written as a normal rule and not a warning block. Is there a planned timeline for this?

ickk:
Based on the reference, it was a bit surprising to discover that the compiler rejects this (since rust-lang/rust#97652) without an explicit annotation, but I understand it's not technically a requirement. I wasn't sure what the reference guidelines were for such things, sorry.

> Is there a planned timeline for this?

On the lint's tracking issue page it appears to have been added to the "Lang Edition 2024" project, but doesn't seem to appear on the project board itself - I'm not sure if this means it was rejected from inclusion in the Edition, or somehow lost and forgotten? Either way, at this point I assume it won't be a hard error until at least the following edition.

ehuss:
I pushed an update as this is changing to a hard error via https://github.com/rust-lang/rust/pull/135964.

traviscross (approved):

ehuss:
https://github.com/rust-lang/rust/pull/135964 has merged, so I'm going to go ahead and merge this. Thanks for moving this along!

PR #1541, opened 2024-07-22
update 'dangling pointers' to new zero-sized rules

RalfJung:
This is the last step for https://github.com/rust-lang/rust/issues/117945. The corresponding change in rustc has already landed (https://github.com/rust-lang/rust/pull/117329).

Fixes https://github.com/rust-lang/rust/issues/117945

PR #1715, opened 2025-01-16
references and Box must be non-null

RalfJung:
I screwed this up in https://github.com/rust-lang/reference/pull/1541... "non-null" used to be implied by "non-dangling", and then when we made the null ptr non-dangling for size 0 we forgot to update the other place where "dangling" was used. Oopsie.


chorman0773 (approved):
Looks good, fairly simple diff.

ehuss (approved):
Thanks!

PR #1721, opened 2025-01-24
Change `'_static` to `'static` as an invalid lifetime parameter name

wkordalski:
I believe that `'static` is an invalid lifetime parameter name.
I've never head that `'_static` is also invalid, thus I suspect a typo here.


ehuss (approved):
Thanks! Not sure how that happened.

PR #1722, opened 2025-01-24
Another "object safety" breadcrumb

SamB:
Some readers my not have heard of the renaming


traviscross:
Thanks for the suggestion.  This is already mentioned further below.  We have some other changes in progress that may serve to make this more prominent.  It's something we'll keep in mind, but we're going to go ahead and close this particular PR.

PR #1727, opened 2025-02-02
Add note about `break rust;` easter egg

chorman0773:
This specifies the incredibly important feature from https://github.com/rust-lang/rust/issues/43162 (alas only as a non-normative note).


traviscross:
Thanks for the PR.  We talked about this in the lang-docs call.  While we do strive to document everything, documenting Easter eggs feels maybe like it kind of diminishes the joke, so we'll probably just leave this out.

PR #1728, opened 2025-02-04
Doc: add concrete decimal numbers to min/max

Tienisto:
We cannot expect people to know the exact number by heart.

Adds the concrete decimal number to improve usefulness.


traviscross:
Thanks for the PR.  We certainly see what you mean.  At the same time, we're not sure it really carries its weight in this case, and in particular, in all the other cases we might do this if we were to go down this road.  So we're going to close this.

PR #1731, opened 2025-02-09
Refactor unsizing coercion documentation

WaffleLapkin:
The old one was quite confusing and also incorrect in a few places. Follow-up to #1622.


nikomatsakis:
To add a bit more color, it'd be good to get a double check that the content is factually accurate before spec team takes it on for editorial review.

src/type-coercions.md
```diff
@@ -187,47 +182,168 @@ r[coerce.types.never]
...
+    * The last field is the *only* field which type involves `T`.
+    * The type of the last field implements `Unsize<F>` where `F` is the same type with `T` replaced by `U`. i.e. if the field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
+
+r[coerce.unsize.pointer]
+Additionally, a type `Foo<T>` can implement `CoerceUnsized<Foo<U>>` when `T` implements `Unsize<U>` or `CoerceUnsized<U>`. This allows it to provide an unsized coercion to `Foo<U>`.
+
+<!-- FIXME: are there more requirements for `CoerceUnsized`? -->
+
+r[coerce.unsize.coerce-unsized-impls]
+Types which currently implement `CoerceUnsized<_>` (assuming `T: Unsize<U>`, `'a: 'b`, `A: CoerceUnsized<B>`, `Al: Allocator`):
```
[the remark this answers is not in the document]
WaffleLapkin: So, mh. Yes, users can find them in the std docs (that's where I got the impls from!).

However, if we want the reference to be prescriptive, I think we should include this. While this is library code, it controls how a language feature behaves.

But I don't have a _strong_ opinion here, if you think that this is not useful I can remove the list.

ehuss:
@WaffleLapkin To move this along, would it be possible to split this PR up a little so it doesn't get hung up? In particular, I think it would move things faster if we remove the name change (unsized -> unsizing) to avoid getting stuck on bikeshedding (that doesn't mean we can't make that change separately, it just adds friction). And remove the list of `CoerceUnsized` impls since I think we don't want to go down that road of trying to duplicate the standard library docs (I understand that there is a close relationship of language features with certain parts of the standard library, but other than recognizing a trait or type is special, we usually don't go into details on the library side, and instead defer to those docs).

You can either just make this PR smaller, or open a new smaller PR, whichever is more convenient.

Assigning TC since he asked.

jackh726:
I can review this over the weekend

WaffleLapkin:
@ehuss I forgot to comment, but I removed the impls and split off the rename.

src/type-coercions.md
```diff
@@ -190,40 +185,89 @@ r[coerce.unsize]
...
+    * The type of the last field implements `Unsize<F>` where `F` is the same type with `T` replaced by `U`. i.e. if the field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
+
+r[coerce.unsize.pointer]
+Additionally, a type `Foo<T>` can implement `CoerceUnsized<Foo<U>>` when `T` implements `Unsize<U>` or `CoerceUnsized<U>`. This allows it to provide an unsized coercion to `Foo<U>`.
 
-r[coerce.unsized.pointer]
-Additionally, a type `Foo<T>` can implement `CoerceUnsized<Foo<U>>` when `T`
-implements `Unsize<U>` or `CoerceUnsized<Foo<U>>`. This allows it to provide an
-unsized coercion to `Foo<U>`.
+<!-- FIXME: are there more requirements for `CoerceUnsized`? -->
```
jackh726, line 270: So, I think another constraint here is that only one field can be coerced.

jackh726 (approved):
Seems good to me in terms of correctness.

WaffleLapkin:
Is there anything I can do to move this forward?

src/type-coercions.md
```diff
@@ -190,40 +185,92 @@ r[coerce.unsize]
...
+    // `&dyn A` can be coerced to `&dyn Super`,
+    // loosing the fact that the underlying type (unit) implements `A` too.
+    let _: &dyn Super = a;
+
+    // The same coercions work with other pointer-like types and wrappers over them:
+    let _: Box<[u8]> = Box::<[u8; 0]>::new([]);
+    let _: Cell<Box<[u8]>> = Cell::new(Box::<[u8; 0]>::new([]));
+
+    // The result of the coercion doesn't *have* to be the same pointer-like type,
+    // alhtough this is only allowed for certain pairs of pointer-like types.
```
lcnr, line 220:
```suggestion
    // although this is only allowed for certain pairs of pointer-like types.
```

src/type-coercions.md
```diff
@@ -190,40 +185,92 @@ r[coerce.unsize]
...
+> [!NOTE]
+> The term "unsized" might be quite confusing, since the coercion works on sized types (pointers) and the source pointer might point to an unsized type in the first place (`&dyn A -> &dyn Super` in the example above).
+>
+> "unsized" refers to the main purpose of these coercions --- converting (pointers to) sized types to (pointers to) unsized types. The pointers being not the focus, since unsized types can't exist without them.
+
+r[coerce.unsize.metadata]
+When performing unsized coercion, the pointer metadata type changes. For example, when unsized `&u32` to `&dyn Debug` metadate type changes from `()` to `DynMetadata<dyn Debug>` (note that exact metadata types are not yet stable). This can also lead to a change in the pointer size -- `&u32` is half the size of `&dyn Debug`.
+
+r[coerce.unsize.traits]
+Three traits, [`Unsize`], [`CoerceUnsized`], and [`PinCoerceUnsized`] are used to assist in this process and expose it for library use.
```
lcnr, line 235:
```suggestion
Three internal traits, [`Unsize`], [`CoerceUnsized`], and [`PinCoerceUnsized`] are
used to assist in this process and expose it for library use.
```

src/type-coercions.md
```diff
@@ -190,40 +185,92 @@ r[coerce.unsize]
...
+    * `S` is a struct.
     * `T` implements `Unsize<U>`.
-    * The last field of `Foo` has a type involving `T`.
-    * If that field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
-    * T is not part of the type of any other fields.
+    * The last field of `S` has a type involving `T`. i.e. it's either of `T` or `C<..., T, ...>` where `C` is a type constructor and `T` is only present in it once (`C<T, T>` is disallowed).
+    * The last field is the *only* field which type involves `T`.
+    * The type of the last field implements `Unsize<F>` where `F` is the same type with `T` replaced by `U`. i.e. if the field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
+
+r[coerce.unsize.pointer]
```
lcnr, line 267:
```suggestion
r[coerce.unsize.struct]
```
:thinking:

src/type-coercions.md
```diff
@@ -190,40 +185,92 @@ r[coerce.unsize]
...
 
 r[coerce.unsize.trait-object]
-* `T` to `dyn U`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
+* `T: Unsize<dyn U>`, when `T` implements `U + Sized`, and `U` is [dyn compatible].
 
 r[coerce.unsize.trait-upcast]
-* `dyn T` to `dyn U`, when `U` is one of `T`'s [supertraits].
+* `dyn T: Unsize<dyn U>`, when `U` is one of `T`'s [supertraits].
     * This allows dropping auto traits, i.e. `dyn T + Auto` to `dyn U` is allowed.
     * This allows adding auto traits if the principal trait has the auto trait as a super trait, i.e. given `trait T: U + Send {}`, `dyn T` to `dyn T + Send` or to `dyn U + Send` coercions are allowed.
```
lcnr, line 257: existing, but pls change this to use `TrA and `TrB` or `Foo` and `Bar` or `Trait` and `Super`

I personally tend to get very confused by single letter trait names, esp if we've just used `T` to mean "any type" 2 lines above

src/type-coercions.md
```diff
@@ -190,40 +185,92 @@ r[coerce.unsize]
...
     * `T` implements `Unsize<U>`.
-    * The last field of `Foo` has a type involving `T`.
-    * If that field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
-    * T is not part of the type of any other fields.
+    * The last field of `S` has a type involving `T`. i.e. it's either of `T` or `C<..., T, ...>` where `C` is a type constructor and `T` is only present in it once (`C<T, T>` is disallowed).
+    * The last field is the *only* field which type involves `T`.
+    * The type of the last field implements `Unsize<F>` where `F` is the same type with `T` replaced by `U`. i.e. if the field has type `Bar<T>`, then `Bar<T>` implements `Unsize<Bar<U>>`.
+
+r[coerce.unsize.pointer]
+A type `S<..., T, ...>` can implement `CoerceUnsized<S<..., U, ...>>` if
```
lcnr, line 268: see https://github.com/rust-lang/rust/blob/2cd37831b0706c9f2d6e7e20eb556dc7548bf732/compiler/rustc_next_trait_solver/src/solve/trait_goals.rs#L1048

unsizing can change an arbitrary number of struct type/const args as long all of them are only mentioned in the last field and that last field impls `CoerceUnsized`

> The last field of `S` has a type involving `T`. i.e. it's either of `T` or `C<..., T, ...>` where `C` is a type constructor and `T` is only present in it once (`C<T, T>` is disallowed).

Where do you get the "`T` is only present once" requirement from?

lcnr: wait, confused by the difference between `coerce.unsize.composite` and `coerce.unsize.pointer` :thinking: using

lcnr: an example which disagrees with the current docs
```rust
trait Trait<T, U> {
    type Assoc: ?Sized;
}

impl<T> Trait<T, u32> for T {
    type Assoc = [u8; 4];
}

impl<T> Trait<T, u16> for T {
    type Assoc = [u8];
}

struct Foo<T: Trait<T, U>, U>(<T as Trait<T, U>>::Assoc);

fn main() {
    let x: &Foo<(), u32> = &Foo([0; 4]);
    let y: &Foo<String, u16> = x;
}
```

lcnr (commented):
nits, then r=me from the "accuracy end". Unsure whether I've got the right to merge reference PRs myself

src/type-coercions.md
WaffleLapkin: "struct" is the one above, `coerce.unsize.composite`. This is for `CoerceUnsized` specifically, which is usually implemented for pointers.

But maybe `.pointer` is just confusing 🤔

src/type-coercions.md
WaffleLapkin: I fixed both of these to be in-line with the compiler. Could you check that I didn't mess it up again?

WaffleLapkin: > Where do you get the "T is only present once" requirement from?

I think this was an old limitation which isn't true anymore. (or I interpreted something incorrectly in the past?)

lcnr: could be, i've cleaned this up at some point during the last few years

src/type-coercions.md
```diff
@@ -190,40 +185,90 @@ r[coerce.unsize]
...
+    impl A for () {}
+
+    trait Super {}
+    impl Super for () {}
+
+    // `&()` can be coerced to `&dyn A`, losing the type information.
+    let a: &dyn A = &();
+
+    // `&dyn A` can be coerced to `&dyn Super`,
+    // loosing the fact that the underlying type (unit) implements `A` too.
```
lcnr:
```suggestion
    // losing the fact that the underlying type (unit) implements `A` too.
```

lcnr (commented):
r=me with nit, idk how to merge stuff

lcnr (approved):

src/type-coercions.md
```diff
@@ -190,40 +185,90 @@ r[coerce.unsize]
...
+
+r[coerce.unsize.traits.unsize]
+[`Unsize`] represents the fact that the target type is layout compatible with the source type and the pointer metadata of the target type can be derived from the metadata of the source, meaning that a pointer to the source type can be converted to a pointer to the target type. For example `[T; N]` implements `Unsize<[T]>` meaning that you can *unsize* former into the later, allowing coercions such as `&[T; N] -> &[T]`.
+
+r[coerce.unsize.traits.coerce-unsized]
+[`CoerceUnsized`] represents the fact that a pointer-like type can be coerced to another pointer-like type, due to `Unsize` being implemented for their pointees. For example, `&T` implements `CoerceUnsized<&U>` when `T: Unsize<U>`.
+
+r[coerce.unsize.traits.pin-coerce-unsized]
+[`PinCoerceUnsized`] is an unsafe marker trait for pointer-like types unsized coercion of which does not break [`Pin`] guarantees. It is a requirement of the `CoerceUnsized` implementation for `Pin`. That is, `&D: PinCoerceUnsized` implies `Pin<&T>: CoerceUnsized<Pin<&U>>`.
```
ehuss: Is "It" here meaning `PinCoerceUnsized`?

src/type-coercions.md
```diff
@@ -190,40 +185,90 @@ r[coerce.unsize]
...
+r[coerce.unsize.composite]
+* `S<T...>: Unsize<S<U...>>`, when:
+    * `S` is a struct.
+    * The type of the last field of `S<T...>` implements `Unsize<X>` where `X` is the type of the last field of `S<U...>`.
+    * All generic arguments of `S` that are involved in its last field are only involved there (and not in other fields).
+    * All generic arguments of `S` that are not involved in its last field are unchanged between `T...` and `U...`.
+
+r[coerce.unsize.pointer]
+A type `S<T...>` *can* implement `CoerceUnsized<S<U...>>` if
+1. Only one field of `S<T...>` has different type than the same field of `S<U...>` (ignoring fields of type `PhantomData<_>`).
```
ehuss: Just wanted to double check, was this phrasing intentional? It seemed a little awkward without the article.

```suggestion
1. Only one field of `S<T...>` has a different type than the same field of `S<U...>` (ignoring fields of type `PhantomData<_>`).
```

src/type-coercions.md
lcnr: yes. more accurately: "`PinCoerceUnsized` being implemented for the pointer is a requirement".
```rust
impl<T: CoerceUnsized<U>, U> CoerceUnsized<Pin<U>> for Pin<T>
where
    T: PinCoerceUnsized,
    U: PinCoerceUnsized,
```

src/type-coercions.md
lcnr: adding "a" seems fine :+1:

PR #1732, opened 2025-02-09
`as` cast improvements

WaffleLapkin:
Improve the description of `as` casts, see individual commits for more info.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -460,29 +460,29 @@ reference types and `mut` or `const` in pointer types.
 | Enumeration           | Integer type          | [Enum cast][expr.as.enum]                             |
 | `bool` or `char`      | Integer type          | [Primitive to integer cast][expr.as.bool-char-as-int] |
 | `u8`                  | `char`                | [`u8` to `char` cast][expr.as.u8-as-char]             |
-| `*T`                  | `*V` [^1]             | [Pointer to pointer cast][expr.as.pointer]            |
+| `*T`                  | `*V` [^meta-compat]   | [Pointer to pointer cast][expr.as.pointer]            |
 | `*T` where `T: Sized` | Integer type          | [Pointer to address cast][expr.as.pointer-as-int]     |
 | Integer type          | `*V` where `V: Sized` | [Address to pointer cast][expr.as.int-as-pointer]     |
-| `&m₁ [T; n]`          | `*m₂ T` [^2]          | Array to pointer cast                                 |
-| `*m₁ [T; n]`          | `*m₂ T` [^2]          | Array to pointer cast                                 |
+| `&m₁ [T; n]`          | `*m₂ T` [^lessmut]    | Array to pointer cast                                 |
+| `*m₁ [T; n]`          | `*m₂ T` [^lessmut]    | Array to pointer cast                                 |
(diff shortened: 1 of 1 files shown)


WaffleLapkin:
Am I reading the CI failure correctly? It seems to imply that footnotes are global and you can have `[^1]` in two different files. Why so?...

ehuss:
Because all pages are concatenated together to generate the print page. mdbook currently doesn't have the ability to rewrite the markdown to make the footnotes unique.

ehuss (approved):
Looks good. Will merge with the footnote labels fixed (probably can pick some short words for those?).

I suppose eventually we should be a little clearer about the normative nature of whether or not a footnote counts.

WaffleLapkin:
@ehuss fixed 👍🏻

src/expressions/operator-expr.md
```diff
@@ -454,35 +454,35 @@ Any cast that does not fit either a coercion rule or an entry in the table is a
...
+| `&m₁ [T; n]`          | `*m₂ T` [^lessmut]    | Array to pointer cast                                 |
+| `*m₁ [T; n]`          | `*m₂ T` [^lessmut]    | Array to pointer cast                                 |
+| [Function item]       | [Function pointer]    | Function item to function pointer cast                |
+| [Function item]       | `*V` where `V: Sized` | Function item to pointer cast                         |
+| [Function item]       | Integer               | Function item to address cast                         |
+| [Function pointer]    | `*V` where `V: Sized` | Function pointer to pointer cast                      |
+| [Function pointer]    | Integer               | Function pointer to address cast                      |
+| Closure [^no-capture] | Function pointer      | Closure to function pointer cast                      |
+
+[^meta-compat]: where `T` and `V` have compatible metadata:
```
BoxyUwU, line 475: This being a footnote makes it *really* difficult to actually find this information. Why is it not just present in the pointer-to-pointer cast section as proper text?

WaffleLapkin: Can you expand on this being difficult to find? You can just click the footnote link, right?

The reason it's not in pointer-to-pointer cast section is that it's a requirement, rather than effect.

I wish footnotes would be rendered where they are defined, rather than at the bottom. And if it's impossible I suppose we could refactor this so it doesn't use footnotes...

BoxyUwU: I don't know or expect to have to click the footnote to find *fundamental* information about how the cast behaves, especially when there's already a non-footnote section about the behaviour of the cast. Footnotes are for "unimportant" extra bits of important information, not main parts of the text.

BoxyUwU: The little footnote indicator is also just really easy to miss as part of a big table of information. (But even if I had noticed it I would not have clicked on it to try and find this information)

WaffleLapkin: If the notes would be right next to the table, would that work for you?

this page rendered on doc.rust-lang.org/reference with footnotes moved right next to the table

BoxyUwU: Honestly I could still imagine missing it, when I see that table my first instinct is to click on the "Pointer to pointer casts" header to see *all* information about pointer to pointer casts. Having the footnotes right below would definitely be an improvement though

PR #1687, opened 2024-11-29
Update function-pointer.md for stabilization of `extended_varargs_abi_support`

Soveu:
Related change in compiler: https://github.com/rust-lang/rust/pull/116161
cc @traviscross @ehuss

PR #1734, opened 2025-02-12
Revert "Update function-pointer.md for stabilization of `extended_varargs_abi_support`"

chorman0773:
Reverts rust-lang/reference#1687

In Re. https://github.com/rust-lang/rust/pull/116161#issuecomment-2652398659 and https://github.com/rust-lang/rust/issues/136896

Waiting on action on rust#136896, so I've marked this as S-waiting-on-team.

Tracking:

- https://github.com/rust-lang/rust/issues/100189


traviscross (approved):

cuviper:
> How important is that?

That's more a question for you and the rest of lang-docs -- will you be bothered if the 1.85 reference states varargs support that doesn't actually work?

PR #1735, opened 2025-02-13
Switch from AVX to SSE in the example

ehuss:
I realized a few seconds too late after merging https://github.com/rust-lang/reference/pull/1720 that AVX will never get tested in CI because that is not part of the default `x86-64` CPU. SSE/SSE2 however should be.


ehuss:
The CI should pass after nightly is published tonight.

traviscross (approved):

PR #1736, opened 2025-02-15
Remove attributes from struct field rest patterns

Skepfyr:
rustc never properly supported attributes in this position so this just removes them entirely.

See rustc change at: rust-lang/rust#136490

I could inline the `StructPatternEtCetera` but it felt slightly more consistent with the rest of the document to leave it separate.
I don't think there are any other changes required here as these attributes don't appear to be mentioned anywhere in the text.


ehuss (approved):
Thanks! Will merge after https://github.com/rust-lang/rust/pull/136490

Issue #1561, opened 2024-08-06
Tracking Issue: Adding mdbook-spec identifiers to reference chapters


chorman0773:
From #1550: 
> The go forward plan adopted by T-spec on 2024-06-13 includes, as one step, to reformat all of the chapters of the Reference to attach named identifiers to each claim, more or less. 

This issue tracks the status of each chapter of the reference being updated in this manner, in alphabetical order.

- [ ] abi.md (Resolve status of currently open PRs)
- [ ] attributes.md (Open #1560)
    - [ ] attributes/codegen.md (Open #1560)
    - [ ] attributes/debugger.md (Open #1560)
    - [ ] attributes/derive.md (Open #1560)
    - [ ] attributes/diagnostics.md (Open #1560)
    - [ ] attributes/limits.md (Open #1560)
    - [ ] attributes/testing.md (Open #1560)
    - [ ] attributes/type_system.md (Open #1560)
- [ ] behaviour-considered-undefined.md
- [ ] behaviour-not-considered-unsafe.md
- [ ] comments.md
- [ ] conditional-compliation.md
- [ ] const_eval.md
- [ ] crates-and-source-files.md
- [ ] destructors.md
- [ ] dynamically-sized-types.md
- [ ] expressions.md
    - [ ] expressions/array-expr.md
    - [ ] expressions/await-expr.md
    - [ ] expressions/block-expr.md
    - [ ] expressions/call-expr.md
    - [ ] expressions/closure-expr.md
    - [ ] expressions/field-expr.md
    - [ ] expressions/grouped-expr.md
    - [ ] expressions/if-expr.md
    - [ ] expressions/literal-expr.md
    - [ ] expressions/loop-expr.md
    - [ ] expressions/match-expr.md
    - [ ] expressions/method-call-expr.md
    - [ ] expressions/operator-expr.md
    - [ ] expressions/path-expr.md
    - [ ] expressions/range-expr.md
    - [ ] expressions/return-expr.md
    - [ ] expressions/struct-expr.md
    - [ ] expressions/tuple-expr.md
    - [ ] expressions/underscore-expr.md
- [ ] glossary.md
- [ ] identifiers.md
- [ ] inline-assembly.md (Open #1550)
- [ ] input-format.md
- [ ] interior-mutability.md
- [ ] items.md
- [ ] keywords.md
- [ ] lexical-structure.md
- [ ] lifetime-elision.md
- [ ] linkage.md
- [ ] macro-ambiguity.md
[...]

PR #1740, opened 2025-02-24
Document `let_chains` again

est31:
Mostly a reapply of #1179 which was reverted by #1251. The reapply is not trivial as in the meantime there has been #1561 and #1787, both of which caused merge conflicts. Also add some modifiers.

Tracking issue: https://github.com/rust-lang/rust/issues/53667
Stabilization report: https://github.com/rust-lang/rust/pull/132833

### DIFF
--- a/src/expressions/if-expr.md
+++ b/src/expressions/if-expr.md
@@ -1,34 +1,40 @@
 r[expr.if]
-# `if` and `if let` expressions
-
-## `if` expressions
+# `if` expressions
 
 r[expr.if.syntax]
 ```grammar,expressions
 IfExpression ->
-    `if` Expression _except [StructExpression]_ BlockExpression
-    (`else` ( BlockExpression | IfExpression | IfLetExpression ) )?
+    `if` IfConditions BlockExpression
+    (`else` ( BlockExpression | IfExpression ) )?
+
+IfConditions -> IfCondition ( `&&` IfCondition )*
+
+IfCondition ->
+      Expression _except [StructExpression]_
+    | `let` Pattern `=` Scrutinee _except [LazyBooleanExpression]_
 ```
 <!-- TODO: The exception above isn't accurate, see https://github.com/rust-lang/reference/issues/569 -->
 
 r[expr.if.intro]
-An `if` expression is a conditional branch in program control.
-The syntax of an `if` expression is a condition operand, followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
+The syntax of an `if` expression is a sequence of one or more condition operands separated by `&&`,
+followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
 
-r[expr.if.condition-bool]
-The condition operands must have the [boolean type].
+r[expr.if.condition]
+Condition operands must be either an [_Expression_] with a [boolean type] or a conditional `let` match.
 
 r[expr.if.condition-true]
-If a condition operand evaluates to `true`, the consequent block is executed and any subsequent `else if` or `else` block is skipped.
+If all of the condition operands evaluate to `true` and all of the `let` patterns successfully match their [scrutinee]s,
+the consequent block is executed and any subsequent `else if` or `else` block is skipped.
 
 r[expr.if.else-if]
-If a condition operand evaluates to `false`, the consequent block is skipped and any subsequent `else if` condition is evaluated.
+If any condition operand evaluates to `false` or any `let` pattern does not match its scrutinee,
+the consequent block is skipped and any subsequent `else if` condition is evaluated.
 
 r[expr.if.else]
 If all `if` and `else if` conditions evaluate to `false` then any `else` block is executed.
 
 r[expr.if.result]
-An if expression evaluates to the same value as the executed block, or `()` if no block is evaluated.
+An `if` expression evaluates to the same value as the executed block, or `()` if no block is evaluated.
 
 r[expr.if.type]
 An `if` expression must have the same type in all situations.
@@ -43,6 +49,7 @@ if x == 4 {
     println!("x is something else");
 }
 
+// `if` can be used as an expression.
 let y = if 12 * 15 > 150 {
     "Bigger"
 } else {
@@ -52,31 +59,17 @@ assert_eq!(y, "Bigger");
 ```
 
 r[expr.if.let]
-## `if let` expressions
-
-r[expr.if.let.syntax]
-```grammar,expressions
-IfLetExpression ->
-    `if` `let` Pattern `=` Scrutinee _except [LazyBooleanExpression]_ BlockExpression
-    (`else` ( BlockExpression | IfExpression | IfLetExpression ) )?
-```
+## `if let` patterns
 
 r[expr.if.let.intro]
-An `if let` expression is semantically similar to an `if` expression but in place of a condition operand it expects the keyword `let` followed by a pattern, an `=` and a [scrutinee] operand.
+`let` patterns in an `if` condition allow binding new variables into scope when the pattern matches successfully.
 
-r[expr.if.let.pattern]
-If the value of the scrutinee matches the pattern, the corresponding block will execute.
-
-r[expr.if.let.else]
-Otherwise, flow proceeds to the following `else` block if it exists.
-
-r[expr.if.let.result]
-Like `if` expressions, `if let` expressions have a value determined by the block that is evaluated.
+The following examples illustrate bindings using `let` patterns:
 
 ```rust
 let dish = ("Ham", "Eggs");
 
-// this body will be skipped because the pattern is refuted
+// This body will be skipped because the pattern is refuted.
 if let ("Bacon", b) = dish {
     println!("Bacon is served with {}", b);
 } else {
@@ -84,7 +77,7 @@ if let ("Bacon", b) = dish {
     println!("No bacon will be served");
 }
 
-// this body will execute
+// This body will execute.
 if let ("Ham", b) = dish {
     println!("Ham is served with {}", b);
 }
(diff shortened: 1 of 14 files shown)


ehuss:
Some notes about temporaries and drop order:

TC's example playground: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=0ce26b5398487e95672adbf389ad946b

Niko's writeup: https://github.com/rust-lang/rust/pull/132833#issuecomment-2701981915

src/expressions/if-expr.md
```diff
@@ -1,35 +1,41 @@
...
-> | _IfLetExpression_ ) )<sup>\?</sup>
+> &nbsp;&nbsp; `if` _IfConditions_ [_BlockExpression_]\
+> &nbsp;&nbsp; (`else` ( [_BlockExpression_] | _IfExpression_ ) )<sup>\?</sup>
+>
+> _IfConditions_ :\
+> &nbsp;&nbsp; _IfCondition_ ( && _IfCondition_ )*
+>
+> _IfCondition_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_]<sub>_except struct expression_</sub>\
+> &nbsp;&nbsp; | `let` [_Pattern_] `=` [_Scrutinee_]
```
ehuss, line 17: The expression cannot be a lazy-boolean, correct?

(_Scrutinee_ already says it doesn't allow struct. It's a little redundant to say that here, but maybe it helps with clarity?)

```suggestion
> &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_]<sub>_except struct expression or lazy boolean expression_</sub>\
> &nbsp;&nbsp; | `let` [_Pattern_] `=` [_Scrutinee_]<sub>_except struct expression or lazy boolean expression_</sub>
>
> [^if-condition-2024]: Editions before 2024 only allow a single _IfCondition_.
```

src/expressions/if-expr.md
```diff
@@ -1,35 +1,41 @@
...
 
 r[expr.if.intro]
-An `if` expression is a conditional branch in program control.
-The syntax of an `if` expression is a condition operand, followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
+The syntax of an `if` expression is a sequence of one or more condition operands separated by `&&`,
+followed by a consequent block, any number of `else if` conditions and blocks, and an optional trailing `else` block.
 
 r[expr.if.condition-bool]
```
ehuss, line 23: Suggesting to rename this, since the condition can now have `let` in addition to boolean expressions.

```suggestion
r[expr.if.condition]
```

src/expressions/if-expr.md
```diff
@@ -1,35 +1,41 @@
...
-> &nbsp;&nbsp; `if` [_Expression_]<sub>_except struct expression_</sub> [_BlockExpression_]\
-> &nbsp;&nbsp; (`else` (
->   [_BlockExpression_]
-> | _IfExpression_
-> | _IfLetExpression_ ) )<sup>\?</sup>
+> &nbsp;&nbsp; `if` _IfConditions_ [_BlockExpression_]\
+> &nbsp;&nbsp; (`else` ( [_BlockExpression_] | _IfExpression_ ) )<sup>\?</sup>
+>
+> _IfConditions_ :\
+> &nbsp;&nbsp; _IfCondition_ ( && _IfCondition_ )*
```
ehuss, line 13:
```suggestion
> &nbsp;&nbsp; _IfCondition_ ( && _IfCondition_ )*[^if-condition-2024]
```

ehuss (commented):
### Scopes
I think the rule in `scopes.md` labeled `names.scopes.pattern-bindings.if-let` needs to be updated. My rough understanding is that the scope of an `if let` binding extends into the *following* _IfCondition_s and the consequent block.

### Edition
At the bottom of `if-expr.md`, can you add something like:

```
r[expr.if.edition2024]
> **Edition differences**: Before the 2024 edition, let chains are not supported and only a single _IfCondition_ is allowed in an `if` expression.
```

I think it makes sense to be at the end of the "Chains of conditions" section? Also just want to double-check the accuracy here.

### Destructors and drop scopes

I'd like to see some updates to the destructors page to explain the drop scopes affected by this change. I'm uncertain exactly which changes to make, but I think roughly they are:

* `destructors.scope.desugaring` -- `if let` (now just `if`) can't be directly converted to a match expression, so I don't know how to phrase that.
* `destructors.scope.temporary.enclosing` -- `if let` should be rephrased somehow. Perhaps it could be phrased as just `if`, and somewhere say that multiple _IfCondition_s are translated to the equivalent nested `if` with single conditions?
    * Or, somehow word it in a similar way to how lazy boolean expressions are worded?
* `destructors.scope.temporary.edition2024` -- I'm uncertain if this needs rewording?

est31:
Thanks for the review. Applied your suggestions and addressed the other review points as well.

> if let (now just if) can't be directly converted to a match expression, so I don't know how to phrase that.

When I read that sentence, I thought it's fine, because there is probably a way to express everything with nested match plus break from loops. But it's a bit underspecified, so I made it a little bit more explicit that it's not just a straightforward translation to `match`.

I made the new sentence suggest the new meaning a bit more clearly, but it's also now stating a bit of a tautology. 

> * `destructors.scope.temporary.edition2024` -- I'm uncertain if this needs rewording?

I think it's fine. One could swap out the `if let` with some longer wording, but it already feels complicated enough for me, I think.

ehuss:
@est31 Just checking if you saw the question about `names.scopes.pattern-bindings.if-let`, where I think the name resolution rules need updating? Is it correct that the bindings are in scope for the following conditions (and consequent)?

Looks like there is also a broken link error in `scopes.md`.

est31:
Indeed the bindings are available for all following conditions in the chain (as well as the body). I will adjust the PR.

src/expressions/if-expr.md
```diff
@@ -1,34 +1,40 @@
...
-    `if` Expression _except [StructExprStruct]_ BlockExpression
-    (`else` ( BlockExpression | IfExpression | IfLetExpression ) )?
+    `if` IfConditions BlockExpression
+    (`else` ( BlockExpression | IfExpression ) )?
+
+IfConditions -> IfCondition ( `&&` IfCondition )*
+
+IfCondition ->
+      Expression _except [StructExprStruct]_
+    | `let` Pattern `=` Scrutinee _except [LazyBooleanExpression]_
```
traviscross, line 14: This can't be right, e.g. (considering only parsing):

```rust
_ = if let () = 0..0 {}; //~ ERROR
```

That is, the exception isn't just `LazyBooleanExpression`; it's any boolean expression with a lower precedence than `&&`.

We're also asking, I think, for something kind of magic to happen here.  When we have,

```rust
_ = if true == false && let () = () {}; //~ OK
```

we're hoping for `Expression` to be `true == false`.  But when we have,

```rust
_ = if true || false && let () = () {}; //~ ERROR
```

we're hoping for `Expression` to try to be `true || false && let () = ()`, notice that's impossible, and fail (i.e., rather than allowing `Expression` to be `true || false`, which would be wrong).  But this isn't really encoded anywhere.

Ideas?

est31: > That is, the exception isn't just `LazyBooleanExpression`; it's any boolean expression with a lower precedence than `&&`.

You are right that the exception isn't just that, and maybe it's misleading to list only `LazyBooleanExpression` because people might think that's an exhaustive list of exceptions.

From what I can see, the reference doesn't require the grammar to be "precedence fitting", i.e. only allow what's possible to express in text form following precedence rules. Instead, it puts the precedence into a table and assumes the grammar is parsed according to the precedence. I would say that table also applies here, as it's part of the larger "expression" family of things.

Because people might think it's exhaustive, we could maybe remove the exception altogether and replace it with a note instead?

> we're hoping for `Expression` to try to be `true || false && let () = ()`, notice that's impossible, and fail (i.e., rather than allowing `Expression` to be `true || false`, which would be wrong).

`||` has weaker precedence, so there isn't a top level && chain any more. On the other hand, `==` has stronger precedence.

The example with `||` is parsed to sth like:

```
If(
    BinOp(
        ||,
        true,
        BinOp(
            &&,
            false,
            Let((), ()),
        )
    ),
    Block([])
)
```

The example with `==` is parsed like:

```

If(
    BinOp(
        &&,
        true,
        BinOp(
            ==,
            false,
            Let((), ()),
        )
    ),
    Block([])
)
```

So in the `||` example, the BinOp with `&&` is actually not top level, its arms can't be `IfCondition`s. So its arms don't have a "permit to let". While in the `==` example, it has such a permission.

So I think this property is emergent from the grammar and the fact that it is parsed according to precedence.

src/expressions/if-expr.md
```diff
@@ -148,27 +103,71 @@ if let E::X(n) | E::Y(n) = v {
...
+# let condition2 = false;
+// Parentheses are required here.
+if let Some(x) = foo && (condition1 || condition2) { /*...*/ }
 ```
 
-[_eRFCIfLetChain_]: https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#rollout-plan-and-transitioning-to-rust-2018
-[`match` expression]: match-expr.md
+r[expr.if.edition2024]
+> [!EDITION-2024]
+> Before the 2024 edition, let chains are not supported and only a single _IfCondition_ is allowed in an `if` expression.
```
ehuss, line 164: IIUC, this will need to be duplicate in loop-expr.md for `while` loops, too, correct?
(That is, until that restriction is removed.)

est31: when I wrote https://github.com/rust-lang/rust/pull/140204#issuecomment-2824605396 I thought so too, but there is this paragraph in the `while let patterns` section:

> Multiple condition operands can be separated with `&&`.
> These have the same semantics and restrictions as [`if` condition chains].

I'd say that this one implies that while let chains are under the same edition specific restriction.

src/expressions/if-expr.md
traviscross: @est31: Is the following discrepancy intentional?  How would we describe the rule here?

```rust
#[cfg(any())]
fn f() {
    if let () = () && ..0 {} //~ OK
    if 0.. && let () = () {} //~ ERROR
}
```

est31: Dumping the AST, it makes sense that it stumbles across that.

First example is parsed like:
```
If(
    BinOp(
        &&,
        Let((), ()),
        Range(None, Some(0))
    ),
    Block([])
)
```
While the second example is parsed like:
```
If(
    Range(
        Some(0),
        Some(
            Let((), ())
        )
    ),
    Block([])
)
```

It's apparently how the parsing works for ranges, you can see similar effects with `_ = false && ..0;` vs `_ = 0.. && false`.

Not sure whether it's a bug or if it's derived from the reference, but I'd say it's existing (and mostly orthogonal) behaviour.

ehuss (approved):
Thanks!

Rather than block on any more discussion on the grammar, I pushed a few modifications to start to express some of the precedence restriction, but it is incomplete and I filed https://github.com/rust-lang/reference/issues/1811 to track follow up work.

PR #1743, opened 2025-02-25
Closures: fix example of desugaring

meithecatte:
Passing `left_top.x` into `right_bottom_x`. This invalidates the comment being made. In fact, the closure lowering doesn't have any special borrow checking rules that would allow it to have two overlapping mutable borrows.


ehuss (approved):
Thanks!

I have a vague memory that there was something this was intentionally trying to illustrate, but I don't remember what is is now.

PR #1484, opened 2024-03-22
patterns: include yet unstable exclusive range patterns

MultisampledNight:
Done as recommended by the [the stabilization guide](https://rustc-dev-guide.rust-lang.org/stabilization_guide.html).

When [exclusive range patterns](https://github.com/rust-lang/rust/issues/37854) are stabilized, this PR will update the reference appropriately to reflect this and remove the mention of there being no exclusive range patterns.

There's no stabilization PR yet, I'm simply going through the guide for now.

PR #1746, opened 2025-03-02
add `RangeToExclusivePattern` to the reference

m4rch3n1ng:
they were stabilized alongside the other exclusive patterns, but were missed by #1484, presumably because they were not mentioned on the [tracking issue](https://github.com/rust-lang/rust/issues/37854), presumably because they were also part of [another feature](https://github.com/rust-lang/rust/issues/67264) and were not added to the other tracking issue after that feature was stabilized.


ehuss:
Thanks for the PR! I started looking at fixing some other issues, and then as I started to tug on the threads it started to unravel, so I have moved my work over to https://github.com/rust-lang/reference/pull/1756.

PR #1747, opened 2025-03-03
Mention `explicit_builtin_cfgs_in_flags` lint cfg chapter

Urgau:


ehuss (approved):
Thanks! I decided to just delete the warning, since the danger of setting these is now essentially gone due to the lint.

PR #1748, opened 2025-03-04
fix: no equivalence between never type and zero-variant enum.

TmSalviano:
Zero-variant enums and the never type are not equivalent. 

None of them can be instantiated since they have no possible values, but:
  1. Only the !  type can be coerced into any other type. (this by itself already invalidate any claims of equivalence)
  2. The ! type has a very well defined use case and semantical meaning in the language itself: represent the result of computations that never complete. At best, Zero-variant enums can have user-defined semantical meaning inside a program when implemented by the developer which, if it's a sound implementation, will be closely related to it's defining property: "a type that has no possible values".

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -320,8 +320,8 @@ enum ZeroVariants {}
 ```
 
 r[items.enum.empty.uninhabited]
-Zero-variant enums are equivalent to the [never type], but they cannot be
-coerced into other types.
+Only the [never type] can be coerced into zero-variant enums. Conversely, 
+zero-variant enums cannot be coerced into any other type.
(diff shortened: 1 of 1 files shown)


traviscross:
Thanks for the suggestion.  Your points 1 and 2 are correct, of course.  And there's probably a better way that we could state this overall.  There are a lot of tricky behaviors here.

Still, in context, we think the existing language is actually OK here.  It says it's "equivalent, but...".  I.e., it's not exactly equivalent.  And that's fine.  That's how language is sometimes.

So we're going to close this, but we'll keep this in mind if we ever take a pass at changing this in other ways.

PR #1750, opened 2025-03-05
do not reference LLVM in our definition of UB

RalfJung:
We should define what is and is not UB ourselves. In particular for the aliasing model, it is far from clear that we will exactly copy what LLVM does. So let's avoid pointing to LLVM here.

- For the aliasing model, add a one-sentence summary.
- For "uninitialized memory", I don't think we define that anywhere yet -- but linking to LLVM's `undef` is not really helpful, either. For now, I think the term itself is sufficiently clear for a first start, and making it more precise requires spelling our the operational semantics in more detail -- that's not the job of this summary page.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -49,13 +49,10 @@ r[undefined.alias]
   `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules.
   The exact liveness duration is not specified, but some bounds exist:
   * For references, the liveness duration is upper-bounded by the syntactic
-    lifetime assigned by the borrow checker; it cannot be live any *longer* than
-    that lifetime.
-  * Each time a reference or box is passed to or returned from a function, it is
-    considered live.
-  * When a reference (but not a `Box`!) is passed to a function, it is live at
-    least as long as that function call, again except if the `&T` contains an
-    [`UnsafeCell<U>`].
+    lifetime assigned by the borrow checker; it cannot be live any *longer* than that lifetime.
+  * Each time a reference or box is dereferenced or reborrowed, it is considered live.
+  * Each time a reference or box is passed to or returned from a function, it is considered live.
+  * When a reference (but not a `Box`!) is passed to a function, it is live at least as long as that function call, again except if the `&T` contains an [`UnsafeCell<U>`].
 
   All this also applies when values of these types are passed in a (nested) field of a compound type, but not behind pointer indirections.
 
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -43,7 +43,7 @@ r[undefined.place-projection]
   [array/slice index expression][project-slice].
 
 r[undefined.alias]
-* Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is a rough sketch of what the requirements look like:
+* Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is an outline of the general principles:
   `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]),
   and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live.
   `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules.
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -43,10 +43,11 @@ r[undefined.place-projection]
...
-* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
-  LLVM’s scoped [noalias] model, except if the `&T` contains an
-  [`UnsafeCell<U>`]. References and boxes must not be [dangling] while they are
-  live. The exact liveness duration is not specified, but some bounds exist:
+* Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is a rough sketch of what the requirements look like:
+  `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]),
+  and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live.
+  `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules.
+  The exact liveness duration is not specified, but some bounds exist:
   * For references, the liveness duration is upper-bounded by the syntactic
```
nikomatsakis, line 51: We have a lower-bound too, right? e.g., if you have a use of some reference `x` at time T1 and then a subsequent use of `x` (or some reference derived from `x`) at time T2, then it is live from `T1 ..= T2`

traviscross (approved):
Looks OK broadly; I know @ehuss is planning to revise editorially.

src/behavior-considered-undefined.md
```diff
@@ -43,22 +43,18 @@ r[undefined.place-projection]
   [array/slice index expression][project-slice].
 
 r[undefined.alias]
-* Breaking the [pointer aliasing rules]. `Box<T>`, `&mut T` and `&T` follow
-  LLVM’s scoped [noalias] model, except if the `&T` contains an
-  [`UnsafeCell<U>`]. References and boxes must not be [dangling] while they are
-  live. The exact liveness duration is not specified, but some bounds exist:
+* Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is a rough sketch of what the requirements look like:
```
ehuss, line 46: We were somewhat uncomfortable with the "rough sketch" nature of this statement. Are the statements here accurate on their own? "rough sketch" implies to me that this isn't just incomplete, but possibly wrong in some way, so I wanted to clarify this point.

If this is just incomplete, can we delete this whole sentence? Like this:

```suggestion
* Breaking the pointer aliasing rules:
```

There is a giant warning just above this that this is not complete, so I was wondering if that covers your concern about the status here?

Or if you wanted to really point to this, maybe a footnote like `[^alias-incomplete]: The exact aliasing rules are not determined yet.`

Or maybe another way to phrase that if the rules written here are accurate, but just incomplete: `The exact aliasing rules are incomplete.`

RalfJung: The statements given here are not on their own sufficient to confidently write sound code. That is already the case before this PR, so nothing changes, we're just being more honest. The aliasing models are the least stable part of our unsafe code semantics so I feel they deserve a dedicated warning.

I don't know what you mean by "accurate and incomplete". I assume you do not mean "complete" in the formal sense of a logic (soundness/completeness), but I also see no way in which an "incomplete" set of rules could be described as "accurate".

RalfJung: > But perhaps you are saying there are too many unknowns to even say that?

Yeah, I am not very comfortable making any hard promises of the form "if you follow these rules, you're definitely good" -- not for the case where references and raw pointers are interleaved in any way. We could conceivably carve out a fragment that must work, where you only ever use raw pointers or references but never both "at the same time", but it'd be a bunch of work to spell this out in sufficient detail, and doing so was not my intent with this PR.

> How about doing something like pulling the sentence out into a dedicated warning block with something like:

That still basically say that what follows are "rules", that might just be altered later. I wouldn't call them rules. They are a sketch or a guiding principle that rules will be developed from, but they are nowhere near precise enough to be described as a set of "rules".

RalfJung: I have replaced "rough sketch" by "outline", is that better?

ehuss (approved):
Thanks!

PR #1752, opened 2025-03-06
Guarantee soundness of pointer-to-int transmutes

joshlf:
Resolves https://github.com/rust-lang/unsafe-code-guidelines/issues/286

### DIFF
--- a/src/types/pointer.md
+++ b/src/types/pointer.md
@@ -79,9 +79,12 @@ r[type.pointer.validity]
 ## Bit validity
 
 r[type.pointer.validity.pointer-fragment]
-Despite pointers and references being similar to `usize`s in the machine code emitted on most platforms,
-the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
-Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
+A pointer or reference type, `P`, is guaranteed to have all of its bytes initialized. Thus, it is always
+sound to transmute `p0: P` to `bytes: [u8; size_of::<P>()]`. However, this operation may not preserve
+provenance, and so transmuting `bytes` back to `p1: P` may result in a pointer or reference without
+valid provenance. If `P` is a raw pointer type, then it may be the case that dereferencing `p1` is undefined
+behavior. If `P` is a reference type, then it may be the case that the act of transmuting to `p1` is
+undefined behavior even if `p1` is never used.
 
 r[type.pointer.validity.raw]
 For thin raw pointers (i.e., for `P = *const T` or `P = *mut T` for `T: Sized`),


src/types/pointer.md
```diff
@@ -79,9 +79,12 @@ r[type.pointer.validity]
 ## Bit validity
 
 r[type.pointer.validity.pointer-fragment]
-Despite pointers and references being similar to `usize`s in the machine code emitted on most platforms,
-the semantics of transmuting a reference or pointer type to a non-pointer type is currently undecided.
-Thus, it may not be valid to transmute a pointer or reference type, `P`, to a `[u8; size_of::<P>()]`.
+A pointer or reference type, `P`, is guaranteed to have all of its bytes initialized. Thus, it is always
```
RalfJung, line 82: The phrasing is odd here -- it should say that a ptr or reference *requires* all its bytes to be initialized, shouldn't it? Bit validity primarily defines "which sequences of bytes can I transmute to this type T".

Furthermore, it is the representation relation of *integers* that decides whether bytes with provenance are permitted there, so it is odd to discuss that here.

joshlf: > The phrasing is odd here -- it should say that a ptr or reference *requires* all its bytes to be initialized, shouldn't it? Bit validity primarily defines "which sequences of bytes can I transmute to this type T" (and which values do they then represent). This here is like saying "a `bool` value is guaranteed to be represented as `0x00` or `0x01`", which is true but not what we usually say when we discuss validity of `bool`.

My goal here is to document what guarantees I can rely on if I currently have a pointer and want to look at its bytes. This prose amounts to a guarantee that we won't *loosen* the validity of pointers to permit uninit bytes. I would be concerned that the opposite wording would read like a sufficient-but-not-necessary condition that we could theoretically loosen in the future.

This is not just hypothetical: there's an argument that the current state of affairs is that int-to-ptr transmutes are well-defined, but that is not currently taken to imply that ptr-to-int transmutes are well-defined.

> Furthermore, it is the representation relation of *integers* that decides whether bytes with provenance are permitted there, so it is odd to discuss that here.

I just wanted to give the looser warning that ptr-to-int-to-ptr might not result in an identical ptr from the AM's perspective. Do you have a suggestion of how to word this better?

RalfJung: > I just wanted to give the looser warning that ptr-to-int-to-ptr might not result in an identical ptr from the AM's perspective. Do you have a suggestion of how to word this better?

I am objecting mostly to the placement, not the wording. Conceptually, what we eventually need is for every type a description of which byte sequences are valid for this type, and which value is represented by each valid byte sequence. This is, however, complicated by the fact that we're trying to give partial guarantees while some details are still being worked out... I don;t have a coherent plan for how to deal with this transitional phase.

chorman0773: FTR, by the stability promise, you can actually assume a given value is valid (assuming it is ever documented as such or otherwise permitted, such as by allowing its creation in safe code). What you cannot do is assume a given value is *not* valid. 

Also, you can rely on the validity invariant by name (or any other invariant) if your reliance is for the fact of the matter alone. IE. if your only requirement to the function parameter is "Call is valid under language rules/language invariant", then you can name the validity invariant by name. Of course, there are few useful functions with such an invariant. (`mem::forget()` however, is a trivial example of a function where the definition can be such that the parameter only need be valid, and not necessarily safe - note however that there is no library guarantee `mem::forget` is, in fact, such a function).

ia0: > Also, you can rely on the validity invariant by name (or any other invariant) if your reliance is for the fact of the matter alone

Indeed, if your function is polymorphic over the invariant it supports, then you can overspecify by naming the validity invariant. But I believe this is almost always useless and inaccurate (you better say you are polymorphic over the invariant).

ehuss:
@joshlf Just checking if you are still interested in working on this?

In general it would be helpful for these kinds of PRs to have a short summary about what it is changing instead of just linking to an issue. It would also be helpful to indicate what consensus has already been gathered with the opsem or lang team, or doing the legwork to gather that consensus, and making it clear what work needs to happen if the PR isn't ready. These changes will eventually need to be brought up with the lang team for a final review, and it is very helpful to have a summary of the context so that they can process it more easily.

joshlf:
> @joshlf Just checking if you are still interested in working on this?

Yes, I'm still interested. It may take a while to drive this towards consensus as it's a fairly abstract discussion. I can't promise that I'll have consistent cycles to push the discussion forward (my work goes through phases regarding what I can focus on), but I will push it forward periodically.

> In general it would be helpful for these kinds of PRs to have a short summary about what it is changing instead of just linking to an issue. It would also be helpful to indicate what consensus has already been gathered with the opsem or lang team, or doing the legwork to gather that consensus, and making it clear what work needs to happen if the PR isn't ready. These changes will eventually need to be brought up with the lang team for a final review, and it is very helpful to have a summary of the context so that they can process it more easily.

I've updated the PR description; let me know if the updated text looks good, or if you'd like to see other changes.

ia0:
By the way, this can be considered an incremental part of the larger effort started in #1664. Commenting to create the link between both PRs since there was none yet.

RalfJung:
FWIW there is ongoing discussion on the LLVM side about the [byte type](https://discourse.llvm.org/t/rfc-add-a-new-byte-type-to-llvm-ir/89522), which also involves the semantics of type-punning loads between integers and pointers, making them relevant here. The current proposal does say that ptr2int type punning strips provenance, but it also proposes adding a `bytecast exact` operation that is fundamentally incompatible with having ptr2int type punning strip provenance. It might be worth waiting for that dust to settle before committing to anything on the Rust side.

PR #1755, opened 2025-03-11
Add: orphan rule rationale.

TmSalviano:
1. I thought it would be helpful to introduce the reader to the problem the orphan rule solves before being exposed to the rule itself. Understanding the orphan rule for the first time can be challenging, and knowing why it exists beforehand can make it easier to grasp.
2. Also, including an explanation of the rationale behind the orphan rule in the orphan rule section of the page just makes sense.

### DIFF
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -175,7 +175,7 @@ r[items.impl.trait.orphan-rule]
 #### Orphan rules
 
 r[items.impl.trait.orphan-rule.rationale]
-The orphan rules ensure that other peoples code can't break your code, and vice versa.
+The orphan rule helps ensure that other people's code can't break your code, and vice versa.
 If an external crate implements an external trait for an external type, and your crate also
 implements the same trait for the same type, the compiler wouldn't know which implementation
 to use.\
--- a/src/items/implementations.md
+++ b/src/items/implementations.md
@@ -174,14 +174,12 @@ be instantiable with the same set of types for the input type parameters. -->
 r[items.impl.trait.orphan-rule]
 #### Orphan rules
 
-r[items.impl.trait.orphan-rule.rationale]
-The orphan rule helps ensure that other people's code can't break your code, and vice versa.
-If an external crate implements an external trait for an external type, and your crate also
-implements the same trait for the same type, the compiler wouldn't know which implementation
-to use.\
-The orphan rule prevents this by requiring that either the trait or some type in the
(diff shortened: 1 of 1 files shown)


src/items/implementations.md
```diff
@@ -174,6 +174,15 @@ be instantiable with the same set of types for the input type parameters. -->
...
 #### Orphan rules
 
+r[items.impl.trait.orphan-rule.rationale]
+The orphan rule helps ensure that other people's code can't break your code, and vice versa.
+If an external crate implements an external trait for an external type, and your crate also
+implements the same trait for the same type, the compiler wouldn't know which implementation
+to use.\
+The orphan rule prevents this by requiring that either the trait or some type in the
+implementation is local to your crate, ensuring only one crate defines the implementation and
+thereby maintaining coherence.
```
bluebear94, line 184: I would personally put a “that” here:

```suggestion
implementation is local to your crate, ensuring that only one crate defines the implementation
and thereby maintaining coherence.
```

traviscross: Probably we'll want to move this into a note section, but it's generally going in the right direction.  The one caveat I'd add is that the problem being prevented here isn't really about the compiler not knowing what to pick.  The compiler knows what to do about that.  It'd throw an ambiguity error.  But what that means is that adding an impl for your own types or of your own traits would be breaking.  So it's mostly about saving space, and that's the real rationale.  Of course, whenever you talk about removing the rule, you do find out all the ways that we are leaning on the rule more technically, and those reasons are a bit more subtle.  E.g.:

- https://github.com/rust-lang/rust/issues/136979

traviscross, line 184: Probably we'll want to move this into a note section, but it's generally going in the right direction.  The one caveat I'd add is that the problem being prevented here isn't really about the compiler not knowing what to pick.  The compiler knows what to do about that.  It'd throw an ambiguity error.  But what that means is that adding an impl for your own types or of your own traits would be breaking.  So it's mostly about saving space, and that's the real rationale.  Of course, whenever you talk about removing the rule, you do find out all the ways that we are leaning on the rule more technically, and those reasons are a bit more subtle.  E.g.:

- https://github.com/rust-lang/rust/issues/136979

ehuss (approved):
Thanks!

We had some conversations about reframing this a little, and so I pushed up some edits to try to capture some of the nuances that we discussed.

PR #1761, opened 2025-03-17
Mention that “every address” ≠ “every pointer”

kpreid:
This adds a note in case people are tempted to interpret “`usize` ... can represent every memory address” as saying that pointer→usize→pointer roundtrips are definitely okay. Since the reference does not yet discuss provenance and that’s a much larger matter, I have merely linked to relevant existing documentation inside and outside the reference.

### DIFF
--- a/src/types/numeric.md
+++ b/src/types/numeric.md
@@ -63,4 +63,4 @@ r[type.numeric.validity]
 For every numeric type, `T`, the bit validity of `T` is equivalent to the bit
 validity of `[u8; size_of::<T>()]`. An uninitialized byte is not a valid `u8`.
 
-[type cast expressions]: ../expressions/operator-expr.html#type-cast-expressions
+[type cast expressions]: ../expressions/operator-expr.md#type-cast-expressions
--- a/src/types/numeric.md
+++ b/src/types/numeric.md
@@ -40,6 +40,10 @@ r[type.numeric.int.size.usize]
 The `usize` type is an unsigned integer type with the same number of bits as the
 platform's pointer type. It can represent every memory address in the process.
 
+> [!NOTE]
+> While a `usize` can represent every *address*, converting a *pointer* to a `usize` is not necessarily a reversible operation.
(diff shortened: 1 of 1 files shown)


traviscross:
cc @rust-lang/opsem

src/types/numeric.md
```diff
@@ -40,6 +40,10 @@ r[type.numeric.int.size.usize]
 The `usize` type is an unsigned integer type with the same number of bits as the
 platform's pointer type. It can represent every memory address in the process.
 
+> [!NOTE]
+> While a `usize` can represent every *address*, converting a *pointer* to a `usize` is not necessarily a reversible operation.
+> For more information, see the documentation for [type cast expressions] and [`std::ptr`].
```
joshtriplett, line 45:
```suggestion
> For more information, see the documentation for [type cast expressions], [`std::ptr`], and [provenance](https://doc.rust-lang.org/std/ptr/index.html#provenance) in particular.
```

traviscross (approved):

src/types/numeric.md
kpreid: I was steering clear of writing “provenance” in the text, but perhaps that is too cautious. Added.

src/types/numeric.md
```diff
@@ -58,3 +62,5 @@ r[type.numeric.validity]
 
 For every numeric type, `T`, the bit validity of `T` is equivalent to the bit
 validity of `[u8; size_of::<T>()]`. An uninitialized byte is not a valid `u8`.
+
+[type cast expressions]: ../expressions/operator-expr.html#type-cast-expressions
```
RalfJung, line 66:
```suggestion
[type cast expressions]: ../expressions/operator-expr.md#type-cast-expressions
```

RalfJung (commented):
LGTM, but the link should go to the md file so it works on GH as well.

ehuss (approved):
Thanks!

PR #1762, opened 2025-03-17
    Add `cfg(true)` and `cfg(false)` to conditional compilation (RFC 3695)

joshtriplett:
Reference update for RFC 3695 (tracking issue https://github.com/rust-lang/rust/issues/131204).

PR #1763, opened 2025-03-17
Use `cfg(false)` instead of `cfg(FALSE)`

joshtriplett:
Split out of https://github.com/rust-lang/reference/pull/1762 so that that PR
doesn't depend on a version of Rust that has `cfg(false)` implemented.


ehuss (approved):
Thanks!

PR #1766, opened 2025-03-24
Add docs for field default values

bluebear94:
Tracking issue: rust-lang/rust#132162

### DIFF
--- a/src/expressions/struct-expr.md
+++ b/src/expressions/struct-expr.md
@@ -22,7 +22,7 @@ r[expr.struct.syntax]
 > &nbsp;&nbsp; )
 >
 > _StructBase_ :\
-> &nbsp;&nbsp; `..` [_Expression_]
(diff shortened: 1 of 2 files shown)


traviscross:
cc @estebank

ehuss:
@estebank In particular, I'm curious what the status of this feature is in terms of how close it is to stabilization? Is this something you can help with review? I'm generally shy from doing reviews if there is still changes forthcoming, and I have no context on this feature.

compiler-errors (commented):
From a compiler perspective, it's probably still worth baking the feature for some time, since it's still pretty new implementation-wise. But I don't think it's going to change much, so it's probably fine to keep this open and tagged as waiting for stabilization and revisit it for tweaks when the stabilization happens.

theemathas:
This PR collides with existing undocumented syntax for rest-patterns in assignee expressions. See https://github.com/rust-lang/reference/issues/2208

PR #1767, opened 2025-03-24
Fix diagnostic attribute typo

mejrs:


ehuss (approved):
Thanks, 🤦 looks like we did this twice.

PR #1769, opened 2025-03-25
Tweak reference for precise capturing in traits

compiler-errors:
* Remove restriction for `use<..>` in trait items
* Mention that all generics including `Self` must be captured

cc https://github.com/rust-lang/rust/pull/138128#issuecomment-2748169854


traviscross (approved):

ehuss (approved):
Thanks!

PR #1771, opened 2025-03-27
Rework trait parameter patterns

ehuss:
This updates the description of trait parameter patterns, primarily to shift it to the style where the main text describes the behavior in the current edition, and changes in the edition are moved to separate edition blocks.

The specific changes:

- Rework the edition presentation (as per our style guide).
- Try to use more consistent wording.
- Add examples.
- Rename some of the rules to switch to the updated edition style, and try to be a little clearer about their intent.
- Change the way it refers to "methods". The reference used to define methods as a separate kind of associated item, but we changed it so that functions allow an optional _SelfParam_, and methods aren't as special as they used to be. This text was still written assuming the previous presentation.

### DIFF
--- a/src/items/traits.md
+++ b/src/items/traits.md
@@ -287,43 +287,70 @@ The [trait implementation] must also begin with the `unsafe` keyword.
 r[items.traits.params]
 ## Parameter patterns
 
-r[items.traits.params.allowed-patterns]
-Function or method declarations without a body only allow [IDENTIFIER] or
-`_` [wild card][WildcardPattern] patterns. `mut` [IDENTIFIER] is currently
-allowed, but it is deprecated and will become a hard error in the future.
+r[items.traits.params.patterns-no-body]
+Parameters in associated functions without a body only allow [IDENTIFIER] or `_` [wild card][WildcardPattern] patterns, as well as the form allowed by [_SelfParam_]. `mut` [IDENTIFIER] is currently allowed, but it is deprecated and will become a hard error in the future.
 <!-- https://github.com/rust-lang/rust/issues/35203 -->
-
-r[items.traits.params.edition2015]
-In the 2015 edition, the pattern for a trait function or method parameter is
-optional:
-
-```rust,edition2015
-// 2015 Edition
-trait T {
-    fn f(i32);  // Parameter identifiers are not required.
-}
-```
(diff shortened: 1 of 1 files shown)


src/items/traits.md
```diff
@@ -287,44 +287,71 @@ The [trait implementation] must also begin with the `unsafe` keyword.
...
+> * `&&` [IDENTIFIER]
+>
+> ```rust,edition2015,compile_fail,E0642
+> // 2015 Edition
+> trait T {
+>     fn f1((a, b): (i32, i32)) {} // ERROR: pattern not allowed
+> }
+> ```
+>
+> Beginning in 2018, all irrefutable patterns are allowed as described in [items.traits.params.patterns-with-body].
```
traviscross, line 353: It's a thought for another PR, but I wonder whether it'd be useful for us to link from these edition differences blocks to the relevant chapter of the edition guide.

src/items/traits.md
```diff
@@ -287,44 +287,71 @@ The [trait implementation] must also begin with the `unsafe` keyword.
...
-The kinds of patterns for parameters is limited to one of the following:
-
-* [IDENTIFIER]
-* `mut` [IDENTIFIER]
-* [`_`][WildcardPattern]
-* `&` [IDENTIFIER]
-* `&&` [IDENTIFIER]
+```rust,compile_fail,E0642
+trait T {
+    fn f2(&x: &i32); // ERROR: Patterns aren't allowed in functions without bodies
```
traviscross, line 303: This should be lowercased to match the others.  I'll fix.

src/items/traits.md
```diff
@@ -287,44 +287,71 @@ The [trait implementation] must also begin with the `unsafe` keyword.
...
+```
+
+r[items.traits.params.pattern-required.edition2018]
+> [!EDITION-2018]
+> Prior to the 2018 edition, the pattern for an associated function parameter is optional:
+>
+> ```rust,edition2015
+> // 2015 Edition
+> trait T {
+>     fn f(i32);  // Parameter identifiers are not required.
```
traviscross, line 330: Other things here are going with the `// OK: ..` pattern, so I'll edit to match that.

traviscross (approved):

PR #1772, opened 2025-03-30
Define byte

### LINKED ISSUE #1673: Disambiguate "byte" size
[The Glossary](https://github.com/rust-lang/reference/blob/5110ba486b7ad8750f8f9c26000189416a1aca75/src/glossary.md?plain=1#L207-L220) and ["Type Layout"](https://github.com/rust-lang/reference/blob/5110ba486b7ad8750f8f9c26000189416a1aca75/src/type-layout.md) don't specify the size of a "byte".

- Was that intentional?
- Is it a real byte? (platform-dependent smallest unit of addressable memory?)
- Is it guaranteed to be an octet?

I think it's worth being explicit about that.

BTW, the [`bool` reference](https://github.com/rust-lang/reference/blob/5110ba486b7ad8750f8f9c26000189416a1aca75/src/types/boolean.md?plain=1#L20-L26) seems somewhat contradictory:
- `size_of::<bool>() == 1` implies platform-dependent byte
- The hexadecimal bit-patterns imply `u8` (guaranteed octet)

> [!note]
> Previous discussions:
> - rust-lang/rust#46176
> - rust-lang/rust#46156

IMO, bytes are more honest and correct than octets (for `bool` and some `enum`s, at least). For primitives such as `char` and `u8`, and exact bit-count is better than bytes

ehuss:
This is split off from https://github.com/rust-lang/reference/pull/1664 which defines what a "byte" is.

Closes https://github.com/rust-lang/reference/issues/1673

cc @rust-lang/opsem, would you be willing to review?

### DIFF
--- a/src/memory-model.md
+++ b/src/memory-model.md
@@ -11,7 +11,7 @@ r[memory.bytes.intro]
 The most basic unit of memory in Rust is a byte. All values are computed from zero or more bytes read from an allocation.
 
 > [!NOTE]
-> While bytes are typically lowered to hardware bytes, Rust uses an "abstract" notion of bytes that can make distinctions which are absent in hardware, such as being uninitialized, or storing part of a pointer.
+> While bytes are typically lowered to hardware bytes, Rust uses an "abstract" notion of bytes that can make distinctions which are absent in hardware, such as being uninitialized, or storing part of a pointer. Those distinctions can affect whether your program has undefined behavior, so they still have tangible impact on how compiled Rust programs behave.
 
 r[memory.bytes.contents]
 Each byte may have one of the following values:
--- a/src/memory-model.md
+++ b/src/memory-model.md
@@ -8,7 +8,7 @@ r[memory.bytes]
 ## Bytes
 
 r[memory.bytes.intro]
-The most basic unit of memory in Rust is a byte. All values are computed from zero or more bytes read from an allocation.
+The most basic unit of memory in Rust is a byte.
 
 > [!NOTE]
 > While bytes are typically lowered to hardware bytes, Rust uses an "abstract" notion of bytes that can make distinctions which are absent in hardware, such as being uninitialized, or storing part of a pointer. Those distinctions can affect whether your program has undefined behavior, so they still have tangible impact on how compiled Rust programs behave.
--- a/src/memory-model.md
+++ b/src/memory-model.md
@@ -1,5 +1,26 @@
+r[memory]
 # Memory model
 
-Rust does not yet have a defined memory model. Various academics and industry professionals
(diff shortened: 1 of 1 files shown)


src/memory-model.md
```diff
@@ -1,5 +1,26 @@
...
+> The memory model of Rust is incomplete and not fully decided.
+
+r[memory.bytes]
+## Bytes
+
+r[memory.bytes.intro]
+The most basic unit of memory in Rust is a byte. All values are computed from zero or more bytes read from an allocation.
+
+> [!NOTE]
+> While bytes are typically lowered to hardware bytes, Rust uses an "abstract" notion of bytes that can make distinctions which are absent in hardware, such as being uninitialized, or storing part of a pointer.
```
RalfJung, line 14:
```suggestion
> While bytes are typically lowered to hardware bytes, Rust uses an "abstract" notion of bytes that can make distinctions which are absent in hardware, such as being uninitialized, or storing part of a pointer. Those distinctions can affect whether your program has undefined behavior, so they still have tangible impact on how compiled Rust programs behave!
```

RalfJung:
This seems fairly non-controversial, given in particular that we do not say what exactly a "provenance" looks like.

However, it does seem to me like we should involve @rust-lang/lang as well, given that this will make stable promises. Though I cannot entirely articulate what we are actually committing to -- I guess it is the fact that uninitialized memory is "real" as far as Rust is concerned, and distinct from all 256 regular initialized states of a byte. But that's long-established consensus, isn't it?

src/memory-model.md
```diff
@@ -1,5 +1,26 @@
...
+
+r[memory.bytes.contents]
+Each byte may have one of the following values:
+
+r[memory.bytes.init]
+* An initialized byte containing a `u8` value and optional [provenance][std::ptr#provenance],
+
+r[memory.bytes.uninit]
+* An uninitialized byte.
```
RalfJung, line 23: I find this enumeration interrupted by `r[...]` quite hard to read. But maybe that can be fixed with better stylesheets? There was some discussion about moving the `r` headers to the left of the text if there is enough horizontal space, is that going to happen?

traviscross: It already did.  You should be seeing it.

RalfJung: Ah, I looked at https://doc.rust-lang.org/reference/behavior-considered-undefined.html. The [nightly](https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html) version looks a lot better indeed. :)

traviscross (approved):

traviscross:
Yes, I think it is.

PR #1775, opened 2025-03-30
Add the ability for rules to be specified in link definitions

ehuss:
This adds the ability to link to rules from a link definition. For example:

```markdown
See [this rule].

[this rule]: expr.array
```

This will convert `[this rule]` to point to the link of the file for that rule.

This is somewhat hacky, as parsing markdown with regex is unwise (and this is intentionally incomplete), and could maybe be done more efficiently.

I don't have an immediate need for this, but I have wanted this a few times in the past.

### DIFF
--- a/mdbook-spec/src/lib.rs
+++ b/mdbook-spec/src/lib.rs
@@ -24,6 +24,10 @@ static ADMONITION_RE: Lazy<Regex> = Lazy::new(|| {
     Regex::new(r"(?m)^ *> \[!(?<admon>[^]]+)\]\n(?<blockquote>(?: *>.*\n)+)").unwrap()
 });
 
+/// A primitive regex to find link reference definitions.
+static MD_LINK_REFERENCE_DEFINITION: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"(?m)^\[(?<label>[^]]+)]: (?<dest>.*)").unwrap());
+
 pub fn handle_preprocessing() -> Result<(), Error> {
     let pre = Spec::new(None)?;
     let (ctx, book) = CmdPreprocessor::parse_input(io::stdin())?;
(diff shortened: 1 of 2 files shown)


mdbook-spec/src/lib.rs
```diff
@@ -114,6 +118,34 @@ impl Spec {
...
+                let dest = &caps["dest"];
+                if let Some((_source_path, path)) = rules.def_paths.get(dest) {
+                    let label = &caps["label"];
+                    let relative = pathdiff::diff_paths(path, current_path).unwrap();
+                    // Adjust paths for Windows.
+                    let relative = relative.display().to_string().replace('\\', "/");
+                    return format!("[{label}]: {relative}#r-{dest}");
+                } else {
+                    return caps.get(0).unwrap().as_str().to_string();
+                }
```
traviscross, line 144: Since all branches diverge, it makes sense to move the `return` outward.  Will update.

mdbook-spec/src/lib.rs
```diff
@@ -114,6 +118,34 @@ impl Spec {
...
+    /// See [this rule].
+    ///
+    /// [this rule]: expr.array
+    /// ```
+    ///
+    /// This will convert the `[this rule]` definition to point to the actual link.
+    fn rule_link_references(&self, chapter: &Chapter, rules: &Rules) -> String {
+        let current_path = chapter.path.as_ref().unwrap().parent().unwrap();
+        MD_LINK_REFERENCE_DEFINITION
+            .replace_all(&chapter.content, |caps: &Captures<'_>| {
```
traviscross, line 134: This is OK of course, but as you mention, it'd be better to do it otherwise.  Remind me to ask you sometime about what the challenges are here.

mdbook-spec/src/lib.rs
traviscross: Actually, of course, here we don't need the `return` at all.

traviscross (approved):

PR #1780, opened 2025-04-04
Update book.toml fix the authors field

szabgab:
See https://rust-lang.github.io/mdBook/format/configuration/general.html#general-metadata


ehuss (approved):
Thanks!

PR #1781, opened 2025-04-06
Add `target_env = "macabi"` and `target_env = "sim"`

madsmtm:
Replaces `target_abi` variants. The old variants are still available, but may be deprecated in the future, so I've removed them as examples from the reference.

Depends on https://github.com/rust-lang/rust/pull/139451.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -196,17 +196,19 @@ Example values:
 * `"msvc"`
 * `"musl"`
 * `"sgx"`
+* `"sim"`
+* `"macabi"`
 
 r[cfg.target_abi]
 ### `target_abi`
 
 r[cfg.target_abi.general]
-Key-value option set to further disambiguate the `target_env` with information
-about the target ABI.
+Key-value option set to further disambiguate the `target_env` or `target_arch`
+with information about the target ABI.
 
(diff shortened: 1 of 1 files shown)


madsmtm:
The `rustc` PR has been approved (and the change is insta-stable), so this is ready for review now

src/conditional-compilation.md
```diff
@@ -196,17 +196,19 @@ Example values:
...
+* `"sim"`
+* `"macabi"`
 
 r[cfg.target_abi]
 ### `target_abi`
 
 r[cfg.target_abi.general]
-Key-value option set to further disambiguate the `target_env` with information
-about the target ABI.
+Key-value option set to further disambiguate the `target_env` or `target_arch`
```
ehuss: I was wondering if we could make this more general about what it is disambiguating? (Instead of being specific about `env` or `arch`.) Maybe something like this?


```suggestion
Key-value option set to further disambiguate the target
```

This mirrors the wording in `target_env`. My reasoning is that we don't have any specific rules that the `target_abi` has to be related to any particular field.

madsmtm: I'm fine with that, have done so in the latest commit.

ehuss (approved):
Thanks!

PR #1788, opened 2025-04-12
Enable `[canonicalize-issue-links]` and `[no-mentions]` in triagebot

Urgau:
This PR enables two new triagebot feature `[canonicalize-issue-links]` and `[no-mentions]`.

Documentation at https://forge.rust-lang.org/triagebot/canonicalize-issue-links.html and https://forge.rust-lang.org/triagebot/no-mentions.html.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -7,6 +7,11 @@ allow-unauthenticated = [
 
 [shortcut]
 
+[no-mentions]
+
+[issue-links]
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks! We didn't feel like we wanted the commit checking side of the issue links for now, since it is somewhat common to include those in commit messages.

PR #1790, opened 2025-04-15
Clarify interaction of asm-goto with IBT

nbdd0121:
Document the restriction that the asm-goto labels should not be jumped to indirectly if there's an architectural distinction (e.g. x64 with IBT).

This restriction is in place for rustc_codegen_gcc, and for LLVM, it's currently allowed, but https://discourse.llvm.org/t/rfc-asm-goto-vs-branch-target-enforcement/85845 may disallow this in the future.

This is to clarify https://github.com/rust-lang/rust/issues/119364#issuecomment-2803036285. cc @programmerjake

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -371,6 +371,7 @@ assert_eq!(y, [3, 2, 0, 1]);
 r[asm.operand-type.supported-operands.label]
 * `label <block>`
   - The address of the block is substituted into the asm template string. The assembly code may jump to the substituted address.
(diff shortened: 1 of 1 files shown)


programmerjake (approved):

joshtriplett:
This looks reasonable. Hopefully in the future there will be a `label indirect` mechanism or similar.

@rfcbot merge

nikomatsakis:
@rfcbot reviewed

tmandry:
@rfcbot reviewed

traviscross (approved):

PR #1794, opened 2025-04-17
Remove broken footnote links from grammar summary

traviscross:
Our grammar syntax supports linking to footnotes which we render with the Markdown version.  This works when the rendering is done on the same page that the footnote appears.

On the grammar summary page, however, where we also render the grammar, these footnotes are not present, and so the links were all broken.

Let's fix this for now by not rendering these footnote links on the grammar summary page.


ehuss (approved):
Makes sense. I thought I had this on my todo list, but I think I had planned to remove footnotes altogether, and then lost track of this.

PR #1796, opened 2025-04-17
Refactor rendering with `RenderCtx`

traviscross:
We thread certain contextual information through to most of our rendering functions.  Currently this information includes whether the rendering is for the summary page and the two link maps.  Adding new information to thread through, or changing something that was only using a subset to use more of the context, requires touching a lot of places.

Let's improve this by refactoring these details into a `RenderCtx` that we'll thread through everywhere instead.  In this way, new details can simply be added to this struct, and callees can use whatever details are needed.

### DIFF
--- a/mdbook-spec/src/grammar.rs
+++ b/mdbook-spec/src/grammar.rs
@@ -88,6 +88,13 @@ enum Characters {
     Range(char, char),
 }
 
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Seems good to me.

PR #1797, opened 2025-04-18
Rename "unsized" coercion as "unsizing"

WaffleLapkin:
Built on top of #1731

### DIFF
--- a/src/expressions/method-call-expr.md
+++ b/src/expressions/method-call-expr.md
@@ -25,7 +25,7 @@ The following procedure is used:
 
 r[expr.method.candidate-receivers]
 The first step is to build a list of candidate receiver types.
-Obtain these by repeatedly [dereferencing][dereference] the receiver expression's type, adding each type encountered to the list, then finally attempting an [unsized coercion] at the end, and adding the result type if that is successful.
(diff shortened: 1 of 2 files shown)


traviscross:
For whatever reason, "unsized coercion" feels more linguistically correct to me than "unsizing coercion".  Here's one linguistic analog that perhaps underlies that feeling.  We say "condo conversions" rather than "condoing conversions".  The conversion makes the thing into a condo, so that's the kind of conversion that it is.  Similarly, here, the coercion makes the thing unsized, so it's an unsized coercion.

tshepang:
heh, *the coercion makes the thing unsized" explanation tells me *unsizing coercion* sound closer to what is meant

*unsized coercion* sounds like a coercion that happens to an unsized type (or to a type that may or may not be sized)

traviscross:
Yes, I understand that as the motivation.  I just can't find a lot of linguistic precedent for that reading in the context of converting things.  E.g., "metric conversion" is to convert things to metric; it's a synonym for metrication.  Similarly, if X is a religion, then an "X conversion" is a conversion to that religion.  If I said we were doing an "octal conversion", I'd mean we were converting to octal; a "Markdown conversion" would be a conversion to Markdown.  If we do a "prime factorization", it's the output rather than the input that is prime.  The "red shift" (not "reddening shift") makes things appear more red.  Etc.

traviscross:
We talked about this on the lang call today without consensus on the renaming question, but with consensus to leave this matter in the hands of lang-docs. @tmandry is interested in being part of the discussion we have on this on the lang-docs side.

PR #1798, opened 2025-04-19
Relabel grammarRailroad-Button

lukaslueg:
I suggest to relabel the button, and use the term "syntax diagram" instead of "railroad". The term "railroad" has no meaning to most readers.

---

On a different topic: Some of the diagrams can be simplified quite a bit. See the example below, which somewhat makes the written grammar more compact, but is to the benefit of the visual representation. Would you like a PR?

```diff
diff --git a/src/visibility-and-privacy.md b/src/visibility-and-privacy.md
index 847d469..e5469be 100644
--- a/src/visibility-and-privacy.md
+++ b/src/visibility-and-privacy.md
@@ -4,11 +4,7 @@ r[vis]
 r[vis.syntax]
 ```grammar,items
 Visibility ->
-      `pub`
-    | `pub` `(` `crate` `)`
-    | `pub` `(` `self` `)`
-    | `pub` `(` `super` `)`
-    | `pub` `(` `in` SimplePath `)`
+      `pub` ( `(` (`crate` | `self` | `super` | (`in` SimplePath))  `)` )?
```

a
b

### DIFF
--- a/src/notation.md
+++ b/src/notation.md
@@ -40,11 +40,11 @@ When such a string in `monospace` font occurs inside the grammar,
 it is an implicit reference to a single member of such a string table
 production. See [tokens] for more information.
 
-## Railroad visualizations
+## Grammar visualizations
 
-Below each grammar block is a button to toggle the display of a [railroad diagram]. A square element is a non-terminal rule, and a rounded rectangle is a terminal.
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks, seems reasonable! I also pushed some changes to change the documentation, and to adjust the styling so the text doesn't wrap.

PR #1799, opened 2025-04-20
Improve documentation of struct expressions

WaffleLapkin:
This compiles:
```rust
trait Tr { type T; }
impl<T> Tr for T { type T = T; }

struct Enum { Unit, Tuple() }

<Enum as Tr>::T::Unit; // OK
<Enum as Tr>::T::Tuple(); // OK
```
So qualified paths can be used when constructing tuple/unit variants. This PR documents that. (note that tuple/unit struct expressions aren't _really_ a thing -- they are just calls/paths, so this doesn't necessarily change much, calls/paths already have correct docs)

Additionally I added more mentions of the fact that these expressions are the same for enums.

### DIFF
--- a/src/expressions/struct-expr.md
+++ b/src/expressions/struct-expr.md
@@ -23,12 +23,9 @@ StructExprField ->
 
 StructBase -> `..` Expression
 
-StructExprTuple ->
-    PathInExpression `(`
-      ( Expression (`,` Expression)* `,`? )?
-    `)`
+StructExprTuple -> CallExpression
 
-StructExprUnit -> PathInExpression
+StructExprUnit -> PathExpression
 ```
 
 r[expr.struct.intro]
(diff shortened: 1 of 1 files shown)


src/expressions/struct-expr.md
```diff
@@ -24,11 +24,11 @@ StructExprField ->
 StructBase -> `..` Expression
 
 StructExprTuple ->
-    PathInExpression `(`
+    PathExpression `(`
       ( Expression (`,` Expression)* `,`? )?
     `)`
```
ehuss, line 29: Maybe to simplify this further it can be `StructExprTuple -> CallExpression`

ehuss (commented):
Thanks!

I opened https://github.com/rust-lang/reference/issues/1802 to capture my thoughts on this in general, and to consider just removing all of this. (Not a blocker for this PR.) I've never been particularly comfortable with how this is presented.

src/expressions/struct-expr.md
WaffleLapkin: Done ✅

ehuss (approved):
Thanks!

PR #1813, opened 2025-05-07
Add new temporary lifetime extension rule

m-ou-se:
See https://github.com/rust-lang/rust/pull/140593

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -383,11 +383,11 @@ expression which is one of the following:
 * The operand(s) of an extending [array][array expression], [cast][cast
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
+* The arguments to an extending [tuple struct] or [tuple variant] constructor expression.
 * The final expression of any extending [block expression].
 
-So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some { 0: &mut 3 }`
-are all extending expressions. The borrows in `&0 + &1` and `Some(&mut 0)` are
-not: the latter is syntactically a function call expression.
+So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some(&mut 3)`
+are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
 The operand of any extending borrow expression has its temporary scope
 extended.
@@ -405,7 +405,7 @@ Here are some examples where expressions have extended temporary scopes:
 let x = &temp();
 let x = &temp() as &dyn Send;
 let x = (&*&temp(),);
-let x = { [Some { 0: &temp(), }] };
+let x = { [Some(&temp()) ] };
 let ref x = temp();
 let ref x = *&temp();
 # x;
@@ -420,7 +420,7 @@ Here are some examples where expressions don't have extended temporary scopes:
 // The temporary that stores the result of `temp()` only lives until the
 // end of the let statement in these cases.
 
-let x = Some(&temp());         // ERROR
+let x = std::convert::identity(&temp()); // ERROR
 let x = (&temp()).use_temp();  // ERROR
 # x;
 ```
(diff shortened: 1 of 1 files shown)


m-ou-se:
@WaffleLapkin I vaguely remember you were working on rewriting this section at some point? Maybe I misremember.

WaffleLapkin:
I indeed found this message:
> [14:11] waffle: thank uuu!! with this knowledge I can document this stuff in the compiler and reference >:3 

I did forget about it and can't find any changes I did back then, so I'd assume I haven't even started rewriting it. I'll write this down so I don't forget again ^^'

src/destructors.md
```diff
@@ -374,6 +374,7 @@ expression which is one of the following:
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
 * The final expression of any extending [block expression].
+* The arguments to [tuple struct][tuple struct] and [tuple variant][tuple variant] constructors.
 
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some { 0: &mut 3 }`
 are all extending expressions. The borrows in `&0 + &1` and `Some(&mut 0)` are
```
WaffleLapkin, line 380: This is no longer true. And similarly in the examples down below.

joshtriplett:
This looks correct to me.

src/destructors.md
```diff
@@ -467,6 +467,8 @@ There is one additional case to be aware of: when a panic reaches a [non-unwindi
 [struct pattern]: patterns.md#struct-patterns
 [tuple pattern]: patterns.md#tuple-patterns
 [tuple struct pattern]: patterns.md#tuple-struct-patterns
+[tuple struct]: types/struct.md#r-type.struct.tuple
+[tuple variant]: types/enum.md#r-type.enum.declaration
```
traviscross, line 471: Let's switch these out for the `r` identifiers.

src/destructors.md
```diff
@@ -374,10 +374,10 @@ expression which is one of the following:
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
 * The final expression of any extending [block expression].
+* The arguments to [tuple struct][tuple struct] and [tuple variant][tuple variant] constructors.
```
ehuss, line 377: We were uncertain about two things here:

* What is an "extending ... expression"? Does that mean it is a subexpression of something else that is extending?
* Related to that, do these need the adjective "extending"? That is, "extending tuple struct" and "extending tuple variant"? If so, would it make sense to merge this with the third bullet point, or at least to move this bullet point to be next to it?

ehuss (commented):
It looks like the examples below still need updating (just below in the "Examples" section).

src/destructors.md
ehuss: I think to answer my own question, I do believe the "extending" part is relevant. My understanding is that:
* `    Some{0: &temp()}` — ok
* `   (Some{0: &temp()},)` — ok
* `foo(Some{0: &temp()},)` — ERROR

because in argument position it is not an "extending expression". The same seems to be true for tuple constructors in https://github.com/rust-lang/rust/pull/140593:

* `    Some(&temp())` — ok
* `   (Some(&temp()),)` — ok
* `foo(Some(&temp()),)` — ERROR

src/destructors.md
m-ou-se: ~~What do you mean? These are already the `r-` anchors?~~ Edit: I see.

src/destructors.md
m-ou-se: Yes. Updated.

m-ou-se:
This now fails the code tests because https://github.com/rust-lang/rust/pull/140593 (which this PR is trying to document) isn't merged yet.

ehuss:
Thanks @m-ou-se! I think this should be ready to merge after https://github.com/rust-lang/rust/pull/140593 merges.

traviscross (approved):

PR #1815, opened 2025-05-07
Add a note that `!` can only be coerced to sized types

WaffleLapkin:
This commit is dated 2025-02-02, I accidentally found it while looking for something else lol.

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -187,7 +187,7 @@ r[coerce.types.closure]
 * Non capturing closures to `fn` pointers
 
 r[coerce.types.never]
-* `!` to any `T`
+* `!` to any `T` which is `Sized`
 
 ### Unsized Coercions
 


src/type-coercions.md
```diff
@@ -187,7 +187,7 @@ r[coerce.types.closure]
 * Non capturing closures to `fn` pointers
 
 r[coerce.types.never]
-* `!` to any `T`
+* `!` to any `T` which is `Sized`
```
traviscross, line 190: Do you happen to have an example of something that would work but for this?

WaffleLapkin:
Checking again, it doesn't seem like this restriction is present in rustc (anymore?). Indeed this program (using incomplete features though!) compiles:
```rust
#![feature(unsized_locals)]

fn main() {
    let x: [()] = panic!();
    f(&x);
}

fn f(_: &[()]) {}
```

In practice though you can't make code that would do this coercion, simply because all coercion sites require sized types.

PR #1816, opened 2025-05-08
Minor fixes to `$crate` behavior

Daniel-Aaron-Bloom:
The description of the behavior of passing `$crate` to a proc-macro is missing, this PR adds one.

Similarly the behavior of `ident` matching `$crate` isn't captured. Every other fragment specifier which matches `$crate` goes through `PathIdentSegment` which does contain `$crate`, but neither `IDENTIFIER_OR_KEYWORD` nor `RAW_IDENTIFIER` can (or should) match against `$crate`. Introducing a new definition just to capture this behavior seems like overkill, especially since it would only be used once, so adding an inline or-clause seems like the cleanest solution.

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -134,7 +134,7 @@ Valid fragment specifiers are:
   * `block`: a [BlockExpression]
   * `expr`: an [Expression]
   * `expr_2021`: an [Expression] except [UnderscoreExpression] and [ConstBlockExpression] (see [macro.decl.meta.edition2024])
-  * `ident`: an [IDENTIFIER_OR_KEYWORD], [RAW_IDENTIFIER], or `$crate`
+  * `ident`: an [IDENTIFIER_OR_KEYWORD], [RAW_IDENTIFIER], or [`$crate`]
   * `item`: an [Item]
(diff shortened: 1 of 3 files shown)


joshtriplett:
This *looks* correct to me.

ehuss (approved):
Thanks!

I noticed there were quite a few changes to `$crate` behavior in https://github.com/rust-lang/rust/pull/37213 in 1.14.

I'm slightly uneasy with this because I don't know what is going to happen in the future with [RFC 3243](https://rust-lang.github.io/rfcs/3243-packages-as-optional-namespaces.html) where a crate may end up being a path (`foo::bar`). I suppose since it will be opaque it's probably not a big deal, but still feels a little weird.

I'm not super happy with how `$crate` is defined, because it is a bit scattered (in paths.md, `macro.decl.meta.dollar-crate`, and `macro.decl.hygiene.crate`). For example, a naive reading could make one think that they could pass `$crate` into an ident from *outside* a macro definition, but that is not the case. Only the text in paths.md makes that really clear that it can only be passed in from a transcriber. Anyway, maybe for another day.

PR #1818, opened 2025-05-08
Compactify syntax diagrams

### LINKED ISSUE #1805: Simplify syntax diagrams
Coming back to the off-topic point in #1798: Some of the syntax diagrams can be simplified quite a bit by removing superfluous elements. There are basically two ways to solve this:

* Implement an optimizer that takes the grammar's AST and performs the simplification automatically. Such as [here](https://github.com/lukaslueg/macro_railroad/blob/191c516fd2548b4a7cee21f0b99ad2ad043b11a9/src/lowering.rs#L708).
* Manually take care not to introduce superfluous elements in the grammar rules themselves. There is a conflict of goals here, as the rules become more compact/nested. That is, the visual representation becomes easier to follow at the cost of the textual representation becoming harder to understand.

See #1798 on what this does to the syntax. Do you fancy a PR for the either of those options? The second one is obviously much more simple to do.

Image

Image

Image

Image

Image

Image

lukaslueg:
Fixes #1805 

This makes the grammar rules more compact by removing superfluous elements. It leads to much simpler syntax diagrams at the expense of more difficult to follow (nested) text representation. See #1805 for some examples.

I visually compared every before-and-after to make sure that the grammar rules represent the same syntax. Please give it firm look anyway.

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -116,9 +116,7 @@ attributes]. It has the following grammar:
 r[attributes.meta.syntax]
 ```grammar,attributes
 MetaItem ->
-      SimplePath
-    | SimplePath `=` Expression
-    | SimplePath `(` MetaSeq? `)`
+      SimplePath ((`=` Expression) | (`(` MetaSeq? `)`))?
 
 MetaSeq ->
     MetaItemInner ( `,` MetaItemInner )* `,`?
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -116,7 +116,10 @@ attributes]. It has the following grammar:
 r[attributes.meta.syntax]
 ```grammar,attributes
 MetaItem ->
-      SimplePath ((`=` Expression) | (`(` MetaSeq? `)`))?
+    SimplePath (
+        `=` Expression
+      | `(` MetaSeq? `)`
+    )?
 
 MetaSeq ->
     MetaItemInner ( `,` MetaItemInner )* `,`?
--- a/src/comments.md
+++ b/src/comments.md
@@ -4,8 +4,7 @@ r[comments]
 r[comments.syntax]
 ```grammar,lexer
 @root LINE_COMMENT ->
-      `//` (~[`/` `!` LF] | `//`) ~LF*
-    | `//`
+    `//` ( ( ~[`/` `!` LF] | `//` ) ~LF* )?
 
 BLOCK_COMMENT ->
       `/*`
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -8,7 +8,10 @@ r[expr.array.syntax]
 ArrayExpression -> `[` ArrayElements? `]`
 
 ArrayElements ->
-      Expression (( `,` Expression )* `,`? | (`;` Expression ))
+    Expression (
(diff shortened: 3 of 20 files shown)


ehuss:
When I get a chance I can take a closer look, but I want to exercise caution with this. Some of these productions were intentionally written in a "not maximally minimized" form, as certain alternations were intended to have semantically meaningful distinctions (but for brevity didn't bother giving them distinct names, which might be worthwhile).

Some of these are definitely improvements, though, so will need to go through them carefully.

traviscross:
Yes, I debate too on a number of these.  It's something to discuss.

traviscross:
> ...as certain alternations were intended to have semantically meaningful distinctions (but for brevity didn't bother giving them distinct names, which might be worthwhile).

That is perhaps the main thing to consider here.  This PR now approximates a factoring that prefers choosing the longest shared prefix.  We could adopt that as our guiding rule, and then also say that anywhere that makes us unhappy, we should just be naming the production separately, and that might work out OK.

traviscross:
In the follow on commits, I had updated the text for these productions to make them as readable as I could, given this system of trying to consolidate shared prefixes and suffixes where possible.  In discussion today, however, @ehuss and I decided that these still weren't good enough.  In particular, the statement of the productions that is there now, while it repeats a lot of terminals and nonterminals, in framed in such a way that each alternation represents a different "kind" of item, and the grammar looks more like a readable syntactic Rust pattern.

I might take another pass to see about naming more of the nonterminals, but unless we come up with a better idea here, I think most of this is going to get reverted in this PR.  So in that context, over in https://github.com/rust-lang/reference/issues/1805#issuecomment-2863400456, @joshtriplett, you had said:

> All of the proposed diagram simplifications look great to me. If the resulting _text_ is less readable, I think we have some options to try to improve the readability.

I wonder whether you might have some ideas, in looking at this PR, about what we might be able to do here.

traviscross:
I'm working on  revision that leans heavily into how the compiler thinks about it.  I'm going through the `rustc` AST and naming more productions to model the state that it keeps.  This is, I suspect, another principled way to go about it, in a bit of the opposite direction.

I'm finding, in doing this, that I do tend to prefer more names to fewer.  One of the goals of the Reference is to provide common language for talking about things, and this does seem maybe helpful in that respect.  I'm going with names similar to those in the compiler, when those are reasonable, as it is always convenient when there is shared nomenclature between lang and compiler.

lukaslueg:
There'll always be tension between the graphical and textual representation. While I agree with https://github.com/rust-lang/reference/issues/1805#issuecomment-2863400456 in spirit, I'd find it perfectly reasonable to reject this PR (or at least 7d720ecf3ac46c641c3043dce633b71f8644fa87), as it makes the textual representation much more difficult to understand than it eases the graphical one. In addition, I myself would find PRs such as this very hard to review.
Having said all that, I'd encourage to reconsider if the diagrams should have their own visual-transformation-pass.

joshtriplett:
@traviscross I think the principle of giving things distinct names if they can't be unified is a good idea.

traviscross:
Let's just close out this PR as we're not going with the approach here.

PR #1821, opened 2025-05-09
document that the `rustc` tool namespace is reserved

jyn514:
this section in the reference is not quite right; or at least is an incomplete summary. here is a more detailed explanation of what's currently implemented:

|Tool|Lints|Attributes|
|-|-|-|
|`clippy`|✅|✅|
|`rustfmt`|❌|✅|
|`miri`|❌|✅|
|`rust_analyzer`|❌|✅|
|`rustdoc`|✅|❌|
|`rustc`|✅ (with `-Z unstable-options`)|❌|
|`diagnostic`|❌|✅|

i'm not sure how to write that up, and it doesn't belong in the attributes section anyway, so i've just added this short snippet.

thanks to @BD103 for putting together the pretty markdown table.
cc https://github.com/rust-lang/rfcs/pull/3808

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -229,6 +229,7 @@ pub fn f() {}
 
 > [!NOTE]
 > `rustc` currently recognizes the tools "clippy", "rustfmt", "diagnostic", "miri" and "rust_analyzer".
+> Additionally, the `rustc` tool, and all tools starting with the literal string "rustc", are reserved for future extensions.
 
 r[attributes.builtin]
 ## Built-in attributes index


traviscross:
Maybe we misunderstand the intent here -- and if so, please feel free to clarify -- but we don't generally document unstable features in the Reference, and there's no way on stable Rust to use tools other than those that we've allowed, so it doesn't really mean anything to reserve a namespace in that context.  So we'll close this out.

jyn514:
hmm, ok. i will make this part of the pr that documents `#![register_tool]`, once that's stabilized.

PR #1822, opened 2025-05-09
Explain why nested receivers are dyn-incompatible

tbodt:
It's not true that they're not supported *yet* - fundamentally, it's not possible to downcast the inner pointer (i.e. removing its vtable) while it is stuck inside the outer pointer


ehuss (approved):
Thanks!

PR #1823, opened 2025-05-10
`if let` guards documentation

Kivooeo:
It’s been a while since the original PR for `if let` guards was opened, so I decided to take the initiative and push it a bit closer to stabilization

This PR adds some documentation — I did my best to keep it consistent with the existing docs on `if` guards
There might still be some rough edges or small inaccuracies, so feel free to point them out

It’s also my first time contributing to the docs, so any feedback is appreciated!

Tracking issue rust-lang/rust#51114
Stabilization: https://github.com/rust-lang/rust/pull/141295

[rendered](https://github.com/rust-lang/reference/blob/19de488bde270f1e336eb7fca07336ef7b5353a5/src/expressions/match-expr.md#if-let-guards)

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -126,6 +126,15 @@ r[destructors.scope.nesting.match-arm]
 * The parent of the expression after the `=>` in a `match` expression is the
   scope of the arm that it's in.
 
+r[destructors.scope.nesting.match-guard-pattern]
+* The parent of an `if let` guard pattern is the scope of the guard expression.
+
+r[destructors.scope.nesting.match-guard-bindings]
+* Variables bound by `if let` guard patterns have their drop scope determined by
+  guard success:
+  - On guard failure: dropped immediately after guard evaluation
+  - On guard success: scope extends to the end of the match arm body
+
 r[destructors.scope.nesting.match]
 * The parent of the arm scope is the scope of the `match` expression that it
   belongs to.
@@ -204,8 +213,8 @@ smallest scope that contains the expression and is one of the following:
 * A statement.
 * The body of an [`if`], [`while`] or [`loop`] expression.
 * The `else` block of an `if` expression.
-* The non-pattern matching condition expression of an `if` or `while` expression,
-  or a `match` guard.
+* The condition expression of an `if` or `while` expression.
+* A `match` guard expression, including `if let` guard patterns.
 * The body expression for a match arm.
 * Each operand of a [lazy boolean expression].
 * The pattern-matching condition(s) and consequent body of [`if`] ([destructors.scope.temporary.edition2024]).
@@ -413,6 +422,58 @@ Here are some examples where expressions don't have extended temporary scopes:
 let x = Some(&temp());         // ERROR
 let x = (&temp()).use_temp();  // ERROR
 # x;
+```
+
+r[destructors.scope.match-guards]
+### Match Guards and Pattern Binding
+
+r[destructors.scope.match-guards.basic]
+Match guard expressions create their own temporary scope. Variables bound within
+guard patterns have conditional drop scopes based on guard evaluation results.
+
+r[destructors.scope.match-guards.if-let]
+For `if let` guards specifically:
+
+1. **Guard pattern evaluation**: The `if let` pattern is evaluated within the
+   guard's temporary scope.
+2. **Conditional binding scope**:
+   - If the pattern matches, bound variables extend their scope to the match arm body
+   - If the pattern fails, bound variables are dropped immediately
+3. **Temporary cleanup**: Temporaries created during guard evaluation are dropped
+   when the guard scope ends, regardless of pattern match success.
+
+r[destructors.scope.match-guards.multiple]
+For multiple guards connected by `&&`:
+- Each guard maintains its own temporary scope
+- Failed guards drop their bindings before subsequent guard evaluation
+- Only successful guard bindings are available in the arm body
+- Guards are evaluated left-to-right with short-circuit semantics
+
+```rust
+# struct PrintOnDrop(&'static str);
+# impl Drop for PrintOnDrop {
+#     fn drop(&mut self) {
+#         println!("drop({})", self.0);
+#     }
+# }
+# fn expensive_operation(x: i32) -> Option<PrintOnDrop> {
+#     Some(PrintOnDrop("expensive result"))
+# }
+
+match Some(42) {
+    // Guard creates temporary scope for pattern evaluation
+    Some(x) if let Some(y) = expensive_operation(x) => {
+        // Both x (from main pattern) and y (from guard) are live
+        // y will be dropped at end of this arm
+        println!("Success case");
+    } // y dropped here
+    Some(x) => {
+        // If guard failed, y was already dropped during guard evaluation
+        // expensive_operation result was also dropped
+        println!("Guard failed case");
+    }
+    None => {}
+}
 ```
 
 r[destructors.forget]
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -23,11 +23,25 @@ MatchArmGuard ->
     `if` MatchConditions
 
 MatchConditions ->
-    MatchCondition ( `&&` MatchCondition )*
+     Expression
+   | MatchConditionChain
 
-MatchCondition ->
-      OuterAttribute* `let` Pattern `=` Scrutinee
-    | Expression
+MatchConditionChain ->
+    MatchChainCondition ( `&&` MatchChainCondition )*
+
+MatchChainCondition ->
+     Expression _except [ExcludedMatchConditions]_
+   | OuterAttribute* `let` Pattern `=` MatchGuardScrutinee
+
+MatchGuardScrutinee -> Expression _except [ExcludedMatchConditions]_
+
+@root ExcludedMatchConditions ->
+      LazyBooleanExpression
+    | RangeExpr
+    | RangeFromExpr
+    | RangeInclusiveExpr
+    | AssignmentExpression
+    | CompoundAssignmentExpression
 ```
 <!-- TODO: The exception above isn't accurate, see https://github.com/rust-lang/reference/issues/569 -->
 
@@ -161,71 +175,42 @@ r[expr.match.guard.no-mutation]
 Moreover, by holding a shared reference while evaluating the guard, mutation inside guards is also prevented.
 
 r[expr.match.if.let.guard]
-## If Let Guards
-Match arms can include `if let` guards to allow conditional pattern matching within the guard clause.
+## If-let Guards
+Match arms can have additional conditions using *if-let guards*. These allow using `if let` expressions within match guard clauses, enabling conditional pattern matching directly in the guard.
 
 r[expr.match.if.let.guard.syntax]
-```rust,ignore
-match expression {
-    pattern if let subpattern = guard_expr => arm_body,
-    ...
+```rust
+enum Command {
+    Run(String),
+    Stop,
+}
+
+match cmd {
+    Command::Run(name) if let Some(first_char) = name.chars().next() => {
+        // Both `name` and `first_char` are available here
+        println!("Running: {} (starts with '{}')", name, first_char);
+    }
+    Command::Run(name) => {
+        println!("Cannot run command: {}", name);
+    }
+    _ => {}
 }
 ```
-Here, `guard_expr` is evaluated and matched against `subpattern`. If the `if let` expression in the guard matches successfully, the arm's body is executed. Otherwise, pattern matching continues to the next arm.
+Here, the guard condition `let Some(first_char) = name.chars().next()` is evaluated. If the `if let` expression successfully matches (i.e., the string has at least one character), the arm's body is executed with both `name` and `first_char` available. Otherwise, pattern matching continues to the next arm.
+
+The key point is that the `if let` guard creates a new binding (`first_char`) that's only available if the guard succeeds, and this binding can be used alongside the original pattern bindings (`name`) in the arm's body.
 
 r[expr.match.if.let.guard.behavior]
 When the pattern matches successfully, the `if let` expression in the guard is evaluated:
   * The guard proceeds if the inner pattern (`subpattern`) matches the result of `guard_expr`.
   * Otherwise, the next arm is tested.
 
-```rust,ignore
-let value = Some(10);
-
-let msg = match value {
-    Some(x) if let Some(y) = Some(x - 1) => format!("Matched inner value: {}", y),
-    _ => "No match".to_string(),
-};
-```
-
 r[expr.match.if.let.guard.scope]
-* The `if let` guard may refer to variables bound by the outer match pattern.
-* New variables bound inside the `if let` guard (e.g., `y` in the example above) are available within the body of the match arm where the guard evaluates to `true`, but are not accessible in other arms or outside the match expression.
+
+For detailed information about variable scope and drop behavior, see the [scope and drop section].
 
 ```rust,ignore
-let opt = Some(42);
-
-match opt {
-    Some(x) if let Some(y) = Some(x + 1) => {
-        // Both `x` and `y` are available in this arm,
-        // since the pattern matched and the guard evaluated to true.
-        println!("x = {}, y = {}", x, y);
-    }
-    _ => {
-        // `y` is not available here --- it was only bound inside the guard above.
-        // Uncommenting the line below will cause a compile-time error:
-        // println!("{}", y); // error: cannot find value `y` in this scope
-    }
-}
-
-// Outside the match expression, neither `x` nor `y` are in scope.
-```
-
-* The outer pattern variables (`x`) follow the same borrowing behavior as in standard match guards (see below).
-
-r[expr.match.if.let.guard.drop]
-
-* Variables bound inside `if let` guards are dropped before evaluating subsequent match arms.
-
-* Temporaries created during guard evaluation follow standard drop semantics and are cleaned up appropriately.
-
-r[expr.match.if.let.guard.borrowing]
-Variables bound by the outer pattern follow the same borrowing rules as standard match guards:
-* A shared reference is taken to pattern variables before guard evaluation
-* Values are moved or copied only when the guard succeeds
-* Moving from outer pattern variables within the guard is restricted
-
-```rust,ignore
-fn take<T>(value: T) -> Option<T> { Some(value) }
+# fn take<T>(value: T) -> Option<T> { Some(value) }
 
 let val = Some(vec![1, 2, 3]);
 match val {
@@ -236,7 +221,7 @@ match val {
 ```
 > [!NOTE]
 > Multiple matches using the `|` operator can cause the pattern guard and the side effects it has to execute multiple times. For example:
-> ```rust,ignore
+> ```rust
 > use std::cell::Cell;
 >
 > let i: Cell<i32> = Cell::new(0);
@@ -268,3 +253,4 @@ r[expr.match.attributes.inner]
 [Range Pattern]: ../patterns.md#range-patterns
 [scrutinee]: ../glossary.md#scrutinee
 [value expression]: ../expressions.md#place-expressions-and-value-expressions
+[scope and drop section]: ../destructors.md#match-guards-and-pattern-binding
--- a/src/expressions/match-expr.md
+++ b/src/expressions/match-expr.md
@@ -9,27 +9,23 @@ MatchExpression ->
         MatchArms?
     `}`
 
-Scrutinee ->
-    Expression _except_ [StructExpression]
+Scrutinee -> Expression _except_ [StructExpression]
 
 MatchArms ->
     ( MatchArm `=>` ( ExpressionWithoutBlock `,` | ExpressionWithBlock `,`? ) )*
     MatchArm `=>` Expression `,`?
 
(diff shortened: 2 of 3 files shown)


Kivooeo:
Test fails because it requires `#![feature(if_let_guard)]` but code style says we can't use this attribute in docs?

src/expressions/match-expr.md
```diff
@@ -18,6 +18,7 @@ MatchArms ->
 MatchArm -> OuterAttribute* Pattern MatchArmGuard?
 
 MatchArmGuard -> `if` Expression
+               | `if` `let` Pattern Expression
```
WaffleLapkin, line 21: This is not full picture, you need to document to if let chains here too. E.g. ([play](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=ae6b60a43fc358927d1a3e169057280f)):
```rust
#![feature(if_let_guard)]
fn main() {
    
    match () {
        () if let a = 1 && a == 1 => (),
        () if false && let Some(()) = None => (),
        _ => (),
    }
}
```

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+Match arms can include `if let` guards to allow conditional pattern matching within the guard clause. This feature is currently unstable and requires the attribute. It is tracked in issue [#51114](https://github.com/rust-lang/rust/issues/51114).
+
+r[expr.match.if.let.guard.syntax]
+```rust,ignore
+match expression {
+    pattern if let subpattern = guard_expr => arm_body,
+    ...
+}
+```
+Here, `guard_expr` is evaluated and matched against `subpattern`. If the match succeeds, the guard evaluates to `true` and the arm is selected. Otherwise, pattern matching continues to the next arm.
```
WaffleLapkin, line 165: > the guard evaluates to `true`

that isn't really a thing

WaffleLapkin: > the arm is selected

"is run" or "is executed" or "is evaluated" maybe/

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+    _ => "No match".to_string(),
+};
+
+```
+
+r[expr.match.if.let.guard.scope]
+* The `if let` guard may refer to variables bound by the outer match pattern.
+* New variables bound inside the `if let` guard (e.g., `y` in the example above) are available within the body of the match arm where the guard evaluates to `true`, but are not accessible in other arms or outside the match expression.
+
+```rust,ignore
```
WaffleLapkin, line 186:
```suggestion
```rust
```

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+      * You cannot move from the scrutinee inside the guard.
+      * New bindings created inside the guard (e.g., via `if let Some(y) = expr`) are local to the guard and do not persist into the match arm body.
+      ```rust,ignore
+      let val = Some(vec![1, 2, 3]);
+
+      let result = match val {
+          Some(v) if let Some(_) = take(v) => "ok", // ERROR: cannot move out of `v`
+          _ => "nope",
+      };
+
```
WaffleLapkin, line 228:
```suggestion
```

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+      ```
+      In the above example, `v` is already bound in the outer pattern, and the guard attempts to move it --- this is not allowed. You can fix it by cloning or borrowing:
+      ```rust,ignore
+      Some(v) if let Some(_) = take(v.clone()) => "ok",
+      ```
+> [!NOTE]
+> Unlike regular if guards, `if let` guards execute only once per match arm, even if the pattern uses the `|` operator to match multiple patterns. This avoids repeated evaluation and potential side effects.
```
WaffleLapkin, line 235: This is not true: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=04262046a736908a3c9d4ce9bfade175

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+## If Let Guards
+Match arms can include `if let` guards to allow conditional pattern matching within the guard clause. This feature is currently unstable and requires the attribute. It is tracked in issue [#51114](https://github.com/rust-lang/rust/issues/51114).
+
+r[expr.match.if.let.guard.syntax]
+```rust,ignore
+match expression {
+    pattern if let subpattern = guard_expr => arm_body,
+    ...
+}
+```
```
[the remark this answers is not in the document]
Kivooeo: Im not really sure, because it's not stable feature and contains behind `#![feature...]` gate that's aren't allowed in documentation and CI will fail with this error
```
Running target/debug/style-check ../src
error in ../src/expressions/match-expr.md: #![feature] attributes are not allowed
```

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
 r[expr.match.guard.no-mutation]
 Moreover, by holding a shared reference while evaluating the guard, mutation inside guards is also prevented.
 
+r[expr.match.if.let.guard]
+## If Let Guards
+Match arms can include `if let` guards to allow conditional pattern matching within the guard clause. This feature is currently unstable and requires the attribute. It is tracked in issue [#51114](https://github.com/rust-lang/rust/issues/51114).
```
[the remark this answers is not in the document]
Kivooeo: So, should I write about this like this feature is already stabilized? And documentaion will only be updated after stabilize this feature?

src/expressions/match-expr.md
Kivooeo: It's interesting because I'm remembrer it's pretty good that this was work locally, I tested that and I fixed this from "execute twice" to "not execute twice", but ok I will fix that, maybe i forgot to run the programm or kinda

src/expressions/match-expr.md
Kivooeo:
```md
MatchArmGuard -> `if` Expression
               | `if` `let` Pattern Expression (`&&` MatchArmGuardCondition)*
```
is this fine syntax? it shows that we can add there 0 or more expressions like this that followed by `&&`. becauase i really dont know how to accurate show user's ability to chain this

src/expressions/match-expr.md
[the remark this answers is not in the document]
WaffleLapkin: Yes

src/expressions/match-expr.md
WaffleLapkin: See [`expr.if`](https://doc.rust-lang.org/nightly/reference/expressions/if-expr.html#r-expr.if)

src/expressions/match-expr.md
```diff
@@ -150,6 +151,99 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+      ```rust,ignore
+      match Some(String::from("hello")) {
+          Some(s) if /* guard */ => { /* s is moved here */ }
+          _ => {}
+      }
+      ```
+  2. Guard evaluation happens after that, and:
+      * It runs using a shared borrow of the scrutinee
+      * You cannot move from the scrutinee inside the guard.
+      * New bindings created inside the guard (e.g., via `if let Some(y) = expr`) are local to the guard and do not persist into the match arm body.
```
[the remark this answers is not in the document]
Kivooeo: I've update it and it now should make more sense

Kivooeo:
@WaffleLapkin not sure if you got notifications when i update something, but i update almost all your reviews expect ones that asks to make code runnable, because im not sure if this is possible, we can remove `ignore` after it will stabilize, is that fine?

src/expressions/match-expr.md
```diff
@@ -9,15 +9,19 @@ MatchExpression ->
...
     MatchArm `=>` Expression `,`?
 
-MatchArm -> OuterAttribute* Pattern MatchArmGuard?
+MatchArm ->
+    OuterAttribute* Pattern MatchArmGuard?
 
-MatchArmGuard -> `if` Expression
+MatchArmGuard ->
+      `if` Expression
+    | `if` Expression `&&` LetChain
```
ehuss, line 24: My understanding is that let guards do not have the same syntactic restrictions the LetChain has, which means we will likely need a different grammar here. 

For example, the following allows struct literals:

```rust
    struct Foo {}
    match () {
        _ if let _ = Foo {} => {}
        () => todo!(),
    }
```

ehuss: Also, this grammar doesn't look correct in terms of chaining, since the thing immediately following `if` can be a `let`, but this grammar requires it to be an Expression.

Kivooeo: should be better now

traviscross:
@bors r-

I appreciate and share the enthusiasm, but this shouldn't merge until the Reference PR is approved, which it is not yet.

src/expressions/match-expr.md
```diff
@@ -150,6 +160,87 @@ This allows shared borrows to be used inside guards without moving out of the sc
...
+
+let msg = match value {
+    Some(x) if let Some(y) = Some(x - 1) => format!("Matched inner value: {}", y),
+    _ => "No match".to_string(),
+};
+```
+
+r[expr.match.if.let.guard.scope]
+* The `if let` guard may refer to variables bound by the outer match pattern.
+* New variables bound inside the `if let` guard (e.g., `y` in the example above) are available within the body of the match arm where the guard evaluates to `true`, but are not accessible in other arms or outside the match expression.
```
jackh726: May be worth being a bit more precise w.r.t drop behavior (i.e. that anything bound in the `if let` guards are dropped before evaluating other match arms)?

Kivooeo: Sure, we can add something like this

```
r[expr.match.if.let.guard.drop]

* Variables bound inside `if let` guards are dropped before evaluating subsequent match arms.

* Temporaries created during guard evaluation follow standard drop semantics and are cleaned up appropriately.
```

What's your opinion on this?

src/expressions/match-expr.md
```diff
@@ -9,15 +9,25 @@ MatchExpression ->
...
-MatchArmGuard -> `if` Expression
+MatchArmGuard ->
+    `if` MatchConditions
+
+MatchConditions ->
+    MatchCondition ( `&&` MatchCondition )*
+
+MatchCondition ->
+      OuterAttribute* `let` Pattern `=` Scrutinee
+    | Expression
```
ehuss: This seems to be missing some things.

Similar to `if` expressions, this will need two separate rules. Something like `Expression | MatchConditions` because without a `let` it is allowed to have LazyBooleanExpression. Then, `MatchConditions` will need to exclude LazyBooleanExpression.

I don't think this should use Scrutinee, because that rejects StructExpression, but that is allowed here.

This will also need similar precedence restrictions as `if let` has with restricting `LazyBooleanExpression | RangeExpr | RangeFromExpr | RangeInclusiveExpr | AssignmentExpression | CompoundAssignmentExpression`.

src/expressions/match-expr.md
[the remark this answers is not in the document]
ehuss: You can just add a full example that will fail CI for now.

src/expressions/match-expr.md
ehuss: Drop behavior should be documented in the destructors chapter. I would expect it to be somewhat similar to the changes that were made for `if` (such as in destructors.scope.temporary.enclosing).

The stabilization report doesn't explain what the drop order is or how temporaries are handled, so it is difficult for me to evaluate what this should say.

Kivooeo: Something like this? I'm just not sure how full should be explanation about drop order

```
r[destructors.scope.temporary.if-let-guard]
### If Let Guards in Match Expressions

When an `if let` guard is evaluated in a match expression:

1. **Guard evaluation scope**: Variables bound by the `if let` pattern and any temporaries created during guard evaluation have a scope that extends through the match arm body if the guard succeeds.

2. **Failed guard cleanup**: If the `if let` guard fails to match:
   - Any variables bound by the guard pattern are immediately dropped
   - Temporaries created during guard evaluation are dropped
   - Pattern matching continues to the next arm

3. **Successful guard**: If the guard succeeds:
   - Guard-bound variables remain live for the duration of the match arm body
   - They are dropped at the end of the match arm execution

4. **Multiple guards**: In arms with multiple guards connected by `&&`:
   - Each failed guard drops its own bindings before evaluating the next guard
   - Only successful guard bindings are available in the arm body

#### Example
...
```

src/expressions/match-expr.md
Kivooeo: I guess, something simple like this should work right? I just not sure if I can reuse LetChain like this

```
MatchArmGuard ->
    `if` MatchConditions

MatchConditions ->
      Expression _except [StructExpression]_
    | LetChain
```

ehuss: No, because LetChain excludes StructExpression, but we *don't* want to exclude that here.
We also don't want MatchConditions to exclude StructExpression.

Kivooeo: Ok, it's little bit hard to write a working syntax to cover all cases, let's try something like this as a start

```
MatchArmGuard ->
    `if` MatchConditions

MatchConditions ->
    MatchCondition ( `&&` MatchCondition )*

MatchCondition ->
      OuterAttribute* `let` Pattern `=` Expression _except [ExcludedConditions]_
    | Expression _except [ExcludedConditions]_

@root ExcludedConditions ->
      LazyBooleanExpression
    | RangeExpr
    | RangeFromExpr
    | RangeInclusiveExpr
    | AssignmentExpression
    | CompoundAssignmentExpression
```

ehuss: I think there needs to be a separate case for a chain vs not-chain.

For example:

```rust
match () {
    _ if x || y => {} // OK
    _ if x || y && let _ = () => {} // ERROR
    _ => {}
}
```

so I think what it needs is something like:

```
MatchConditions ->
     Expression
   | MatchConditionChain

MatchConditionChain ->
    MatchChainCondition ( `&&` MatchChainCondition )*

MatchChainCondition ->
     Expression _except [ExcludedMatchConditions]_
   | OuterAttribute* `let` Pattern `=` MatchGuardScrutinee

MatchGuardScrutinee -> Expression _except [ExcludedMatchConditions]_

@root ExcludedMatchConditions ->
      LazyBooleanExpression
    | RangeExpr
    | RangeFromExpr
    | RangeInclusiveExpr
    | AssignmentExpression
    | CompoundAssignmentExpression
```

right?

Kivooeo: well, yeah, separate `MatchGuardScrutinee` is a good idea

Kivooeo:
@ehuss could you please check last commit if this any better than before

src/destructors.md
```diff
@@ -126,6 +126,15 @@ r[destructors.scope.nesting.match-arm]
 * The parent of the expression after the `=>` in a `match` expression is the
   scope of the arm that it's in.
 
+r[destructors.scope.nesting.match-guard-pattern]
+* The parent of an `if let` guard pattern is the scope of the guard expression.
```
ehuss, line 130: I'm struggling with this one. Isn't the parent of both the scrutinee and the bindings of an `if let` guard the entire arm?  I'm not able to find a way that it is different from destructors.scope.nesting.match-guard. 

For example, in:
```rust
match x {
    Some(a) if let [one] = some_expression => {}
    _ => {}
}
```

I would expect the parent of both `one` and `some_expression` to be the match arm (and drop *after* the arm body).

Can you clarify this for me?

src/destructors.md
```diff
@@ -126,6 +126,15 @@ r[destructors.scope.nesting.match-arm]
...
+
+r[destructors.scope.nesting.match-guard-bindings]
+* Variables bound by `if let` guard patterns have their drop scope determined by
+  guard success:
+  - On guard failure: dropped immediately after guard evaluation
+  - On guard success: scope extends to the end of the match arm body
```
ehuss, line 136: I'm not sure I understand the need for this rule. Since the bindings are scoped to the match arm, I would expect destructors.scope.nesting.match-guard to cover this. That is, as we leave the arm (success or failure), the bindings will get dropped. Can you clarify how this would be different from the other rule?

src/destructors.md
```diff
@@ -204,8 +213,8 @@ smallest scope that contains the expression and is one of the following:
 * A statement.
 * The body of an [`if`], [`while`] or [`loop`] expression.
 * The `else` block of an `if` expression.
-* The non-pattern matching condition expression of an `if` or `while` expression,
-  or a `match` guard.
+* The condition expression of an `if` or `while` expression.
```
ehuss, line 216: Why is this changing `if` and `while` to drop the "non-pattern matching" qualifier?

src/destructors.md
```diff
@@ -204,8 +213,8 @@ smallest scope that contains the expression and is one of the following:
...
+* A `match` guard expression, including `if let` guard patterns.
```
ehuss, line 217: I'm a little confused by this. Don't `match` guards have different temporary behavior of non-pattern matching conditions versus `let` conditions? Looking at [`drop-order.rs`](https://github.com/rust-lang/rust/blob/5796073c134eaac30475f9a19462c4e716c9119c/tests/ui/rfcs/rfc-2294-if-let-guard/drop-order.rs) I see that if_guard and if_let_guard have different drop orders.

src/destructors.md
```diff
@@ -415,6 +424,58 @@ let x = (&temp()).use_temp();  // ERROR
 # x;
 ```
 
+r[destructors.scope.match-guards]
+### Match Guards and Pattern Binding
```
ehuss, line 428: Overall I'm uncertain what this section is saying that is different from the temporary scope rules in destructors.scope.temporary.enclosing. I'm also unclear about why this is differentiating if the pattern matches or not. Either case is handled when execution leaves the arm (success or failure), right?

src/destructors.md
Kivooeo: Hm, yes, we can delete this because it duplucate of destructors.scope.nesting.match-guard, or at least alligns in a way it shouldnt

src/destructors.md
Kivooeo: i forgot to return it actually, but after this change i realise that we should keep them separatly actually, something like this would be better?

```
* The non-pattern matching condition expression of an `if` or `while` expression.
* A `match` guard expression.
* The scrutinee expression of an `if let` guard pattern.
```

src/destructors.md
Kivooeo: This one also can be removed by the same reason as above

src/destructors.md
Kivooeo: i guess im over-engineering documentation, i will try to look on it from different angle

src/destructors.md
Kivooeo: But I'm unsure where should I write about drop on failure case, because they are different on success and failure, I think we can keep it, but I'm not sure where exactly

dianne: I can't speak to the spec, but implementation-wise, there's no conceptual difference between success and failure. `if let` guards declare their bindings in the scope of the match arm. Likewise, the scrutinee of an `if let` guard has the match arm scope as its temporary scope. Maybe that's the missing detail? Here, wildcards are used to demonstrate the temporary scope of the scrutinee [(playground link)](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=eef4650279dfc99791e9215a13f9968c). 
```rust
fn main() {
    assert_drop_order(1..=3, |o| match () {
        () if let (_, _) = (o.log(2), o.log(3))
            && true =>
        {
            o.log(1);
        }
        _ => unreachable!(),
    });
}
```
On both success and failure, the reason `if let` guards' bindings and temporaries are dropped is because of breaking out of the arm's scope. This is even the case when the guard has to be evaluated multiple times for the same arm because of or-pattern expansion.

The difference between the guard's bindings and the match arm's pattern's bindings is all in sequencing. By-value bindings for the match arm's pattern are only created if the guard succeeds [(expr.match.guard.value)](https://doc.rust-lang.org/reference/expressions/match-expr.html#r-expr.match.guard.value). When a guard fails, the guard's scrutinee and bindings/temporaries from earlier in the `let` chain will be dropped, since we leave the arm's scope. But the values bound by the arm's pattern won't, because they haven't been moved yet; the match's scrutinee is still in the match's scope, which we don't leave when a guard fails.

Edit: Updated the example to actually demonstrate the scrutinee's temporary scope. I guess the drop order for the tuple is because tuples drop the left field first, then the right? Whereas if you create bindings for the fields, the right one is dropped first, then left.

src/destructors.md
dianne: I believe the temporary scope of an `if let` match guard is the whole match arm, yes [(see my other comment for an example)](https://github.com/rust-lang/reference/pull/1823/files#r2144393397). Temporaries for normal boolean guards are dropped by the end of the guard, but the scrutinee of an `if let` guard lives for the arm.

theemathas:
There probably needs to be a mention somewhere that `if let` guards also have temporary lifetime extension, like `let` statements do.

Kivooeo:
Closing in favour of https://github.com/rust-lang/reference/pull/1957

PR #1826, opened 2025-05-15
Parse optionals and repeats without regexes

traviscross:
Rather than parsing optionals and repeats fully in the recursive descent style, we were using regular expressions to do part of the matching and parsing.  That's fine for what it is, but as we think about extending the grammar language surrounding repeats further, it might be more straightforward for this to be parsed in the more usual way.  So let's do that.  Doing this also results in better and more targeted errors when parsing malformed syntax.

We had been supporting a space between an expression and the optional and repeat sigils `?`, `*`, and `+` (but not between an expression and the `{a..b}` ranged repeat syntax).  In making this change, we drop this support and adjust the affected productions.  We were only using this in a handful of places, and the clarity of the productions seem the same or better by removing these spaces.

We verified that, setting aside the removal of these spaces, the rendered output of the Reference is byte identical before and after this change.

cc @ehuss

### DIFF
--- a/mdbook-spec/src/grammar/parser.rs
+++ b/mdbook-spec/src/grammar/parser.rs
@@ -214,7 +214,7 @@ impl Parser<'_> {
             return Ok(None);
         };
 
-        let mut kind = if self.take_str("U+") {
+        let kind = if self.take_str("U+") {
             self.parse_unicode()?
         } else if self.input[self.index..]
             .chars()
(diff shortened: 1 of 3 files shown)


traviscross:
The motivation for this refactoring is to next add repeats with separators.

ehuss (approved):
Thanks!

PR #1827, opened 2025-05-16
Parse grammar without regexes

traviscross:
We'd been parsing the grammar with a combination of recursive descent and regular expression matchers.  This combination has its merits, and it's done tastefully here, but it seems maybe more straightforward to do the parsing entirely with recursive descent.  Among other things, doing it this way allows us to provide more precise error reporting on malformed inputs.

The cost, in terms of lines of code, of doing this is rather modest, and the result seems at least as clear -- there's some mental cost to code switching between the two worlds.  So let's make the switch and parse the grammar without regular expressions.

We verified that the rendered output of the Reference is byte identical before and after this change.

cc @ehuss

### DIFF
--- a/mdbook-spec/src/grammar/parser.rs
+++ b/mdbook-spec/src/grammar/parser.rs
@@ -1,11 +1,9 @@
 //! A parser of the ENBF-like grammar.
 
 use super::{Characters, Expression, ExpressionKind, Grammar, Production};
-use regex::{Captures, Regex};
 use std::fmt;
 use std::fmt::Display;
 use std::path::Path;
-use std::sync::LazyLock;
 
 struct Parser<'a> {
     input: &'a str,
@@ -74,18 +72,6 @@ impl Parser<'_> {
         }
         self.index += upper;
         &self.input[i..i + upper]
-    }
-
-    /// If the input matches the given regex, it is returned and the head is moved forward.
-    ///
-    /// Note that regexes must start with `^`.
-    fn take_re(&mut self, re: &Regex) -> Option<Captures<'_>> {
-        if let Some(cap) = re.captures(&self.input[self.index..]) {
-            self.index += cap[0].len();
-            Some(cap)
-        } else {
-            None
-        }
     }
 
     /// Returns whether or not the given string is next, and advances the head if it is.
@@ -168,13 +154,12 @@ impl Parser<'_> {
     }
 
     fn parse_expression(&mut self) -> Result<Option<Expression>> {
-        static ALT_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^ *\| *").unwrap());
-
         let mut es = Vec::new();
         loop {
             let Some(e) = self.parse_seq()? else { break };
             es.push(e);
-            if self.take_re(&ALT_RE).is_none() {
+            _ = self.space0();
+            if !self.take_str("|") {
                 break;
             }
         }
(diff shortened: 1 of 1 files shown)


mdbook-spec/src/grammar/parser.rs
```diff
@@ -268,21 +254,28 @@ impl Parser<'_> {
...
+        Ok(term)
     }
 
     fn parse_charset(&mut self) -> Result<ExpressionKind> {
         self.expect("[", "expected opening [")?;
         let mut characters = Vec::new();
         loop {
             self.space0();
-            let Some(ch) = self.parse_characters() else {
+            let Some(ch) = self.parse_characters()? else {
```
traviscross, line 277: We now notice and report errors when parsing the elements of a character class.

mdbook-spec/src/grammar/parser.rs
```diff
@@ -295,27 +288,49 @@ impl Parser<'_> {
...
+    /// `` `a`-`b` `` | `` `term` `` | `` NonTerminal ``.
+    fn parse_characters(&mut self) -> Result<Option<Characters>> {
+        if let Some(b'`') = self.peek() {
+            let recov = self.index;
+            let a = self.parse_terminal_str()?;
+            if let Some(b'-') = self.peek() {
+                //~^ Parse `` `a`-`b` `` character range.
+                if a.len() > 1 {
+                    self.index = recov + 1;
+                    bail!(self, "invalid start terminal in range");
```
traviscross, line 300: We recover the parser index before reporting errors so as to put the arrow in the correct place.

mdbook-spec/src/grammar/parser.rs
```diff
@@ -344,13 +359,19 @@ impl Parser<'_> {
...
     fn parse_unicode(&mut self) -> Result<ExpressionKind> {
-        static UNICODE_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[A-Z0-9]{4}").unwrap());
-
-        match self.take_re(&UNICODE_RE) {
-            Some(s) => Ok(ExpressionKind::Unicode(s[0].to_string())),
-            None => bail!(self, "expected 4 hexadecimal uppercase digits after U+"),
+        let mut xs = Vec::with_capacity(4);
+        for _ in 0..4 {
+            match self.peek() {
+                Some(x @ (b'0'..=b'9' | b'A'..=b'F')) => {
```
traviscross, line 365: The old code was accepting `A-Z` in a hex digit.  We narrow this down to `A-F`.

mdbook-spec/src/grammar/parser.rs
```diff
@@ -428,16 +447,17 @@ impl Parser<'_> {
...
     fn parse_footnote(&mut self) -> Result<Option<String>> {
-        static FOOTNOTE_RE: LazyLock<Regex> =
-            LazyLock::new(|| Regex::new(r"^([^\]\n]+)]").unwrap());
         if !self.take_str("[^") {
             return Ok(None);
         }
-        match self.take_re(&FOOTNOTE_RE) {
-            Some(cap) => Ok(Some(cap[1].to_string())),
-            None => bail!(self, "unterminated footnote, expected closing `]`"),
+        let id = self.take_while(&|x| !['\n', ']'].contains(&x)).to_string();
```
traviscross, line 455: Really I'd prefer this `to_string` call to happen at the bottom, but alas, this is one of those borrow checker limitations.

mdbook-spec/src/grammar/parser.rs
```diff
@@ -295,27 +287,48 @@ impl Parser<'_> {
...
+            Ok(None)
         }
     }
 
+    /// Parse e.g. `<prose text>`.
     fn parse_prose(&mut self) -> Result<ExpressionKind> {
-        static PROSE_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^<([^>\n]+)>").unwrap());
-        match self.take_re(&PROSE_RE) {
-            Some(cap) => Ok(ExpressionKind::Prose(cap[1].to_string())),
-            None => bail!(self, "unterminated prose, expected closing `>`"),
```
traviscross, line 317: Here and elsewhere, we were incorrectly reporting things being unterminated when they were actually empty.  This is now fixed.

ehuss (approved):
Thanks!

Yea, the intent was to reduce LOC at the expense of worse error messages, but this seems fine to me.

PR #1828, opened 2025-05-17
Document #[cfg(version(...))]

est31:
Tracking issue: https://github.com/rust-lang/rust/issues/64796
RFC: https://github.com/rust-lang/rfcs/pull/2523
Stabilization PR: https://github.com/rust-lang/rust/pull/141137

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -55,6 +55,11 @@ r[cfg.predicate.not]
 r[cfg.predicate.literal]
 * `true` or `false` literals, which are always true or false respectively.
 
+r[cfg.predicate.version]
+* `version()` with a version number inside. It is true if the language version
+  the compiler targets is higher or equal to the contained version number.
+  It is false otherwise.
+
 r[cfg.option-spec]
 _Configuration options_ are either names or key-value pairs, and are either set or unset.
 
@@ -299,6 +304,25 @@ r[cfg.proc_macro]
 Set when the crate being compiled is being compiled with the `proc_macro`
 [crate type].
 
+r[cfg.version]
+## The `version()` predicate
+
+r[cfg.version.behaviour]
+The `version()` predicate evaluates to true if both:
+
+* The version number contained inside follows the format and
+* The version number contained inside is less than or equal to the version
+  of the language the compiler targets. Usually the compiler version and
+  language version match. So compiler version `1.50.0` targets language
+  `1.50.0`.
+
+r[cfg.version.format]
+In order for it to be considered of valid format, the version number has to
+follow either the `"a.b.c"` scheme or the `"a.b"` scheme, where `a,b,c` are
+decimal integers between `0` and `65535`, inclusively. Semantically, assume `c`
+to be 0 if not present. Order wise, version numbers behave as if they were
+Rust tuples `(a,b,c)` with `a,b,c` being `u16` integers.
+
 r[cfg.panic]
 ### `panic`
 
@@ -368,6 +392,12 @@ fn on_32bit_unix() {
 // This function is only included when foo is not defined
 #[cfg(not(foo))]
 fn needs_not_foo() {
+  // ...
+}
+
+// This function is only included if the language version is newer than 1.50.0
+#[cfg(version("1.50.0"))]
+fn needs_new_compiler() {
   // ...
 }
 


src/conditional-compilation.md
```diff
@@ -55,6 +55,11 @@ r[cfg.predicate.not]
 r[cfg.predicate.literal]
 * `true` or `false` literals, which are always true or false respectively.
 
+r[cfg.predicate.version]
+* `version()` with a version number inside. It is true if the language version
+  the compiler targets is higher or equal to the contained version number.
+  It is false otherwise.
```
traviscross, line 61: This needs grammar, above, as for the other predicates.

src/conditional-compilation.md
```diff
@@ -299,6 +304,25 @@ r[cfg.proc_macro]
 Set when the crate being compiled is being compiled with the `proc_macro`
 [crate type].
 
+r[cfg.version]
+## The `version()` predicate
```
ehuss, line 308: It seems a little confusing to have both this and `cfg.predicate.version`. I'm wondering if it would make sense to extend the "option" grammar to include `version(...)`, and just remove `cfg.predicate.version`? Then `version()` would just be another option shown in the list here.

And the section header should probably be third level, and match the style of the other options.

```suggestion
r[cfg.version]
### `version()`
```

src/conditional-compilation.md
```diff
@@ -299,6 +304,25 @@ r[cfg.proc_macro]
...
+
+r[cfg.version.behaviour]
+The `version()` predicate evaluates to true if both:
+
+* The version number contained inside follows the format and
+* The version number contained inside is less than or equal to the version
+  of the language the compiler targets. Usually the compiler version and
+  language version match. So compiler version `1.50.0` targets language
+  `1.50.0`.
```
ehuss, line 317: Usually the reference focuses on the language and not specifics about the compiler or specific implementations. The text here seems a little confusing in that regard. I would probably just strike this content.

```suggestion
  of the language the compiler targets.
```

src/conditional-compilation.md
```diff
@@ -299,6 +304,25 @@ r[cfg.proc_macro]
...
+
+r[cfg.version.format]
+In order for it to be considered of valid format, the version number has to
+follow either the `"a.b.c"` scheme or the `"a.b"` scheme, where `a,b,c` are
```
ehuss, line 321:
```suggestion
follow either the `"a.b.c"` scheme or the `"a.b"` scheme, where `a`, `b`, and `c` are
```

ehuss (commented):
Can you make sure to not wrap lines when adding new content?

src/conditional-compilation.md
```diff
@@ -299,6 +304,25 @@ r[cfg.proc_macro]
...
+decimal integers between `0` and `65535`, inclusively. Semantically, assume `c`
```
madsmtm, line 322: Question: Does this detail matter in the reference? Isn't it enough to say "these should be integers", and leave the range to the compiler implementation?

src/conditional-compilation.md
madsmtm: It might be beneficial to though to discuss a bit how this would work in alternative implementations like `mrustc` / `gccrs`? Though maybe "the language the compiler targets" is enough for that.

src/conditional-compilation.md
```diff
@@ -55,6 +65,11 @@ r[cfg.predicate.not]
 r[cfg.predicate.literal]
 * `true` or `false` literals, which are always true or false respectively.
 
+r[cfg.predicate.version]
+* `version()` with a version number inside. It is true if the language version
```
ehuss, line 63: Can we merge `cfg.version` into this section so that it isn't described in two places?

src/conditional-compilation.md
ehuss: @traviscross and I discussed this a little bit and felt like this probably *should* be specified? For example, if you say `cfg(version("1.81.65535"))`, that will be true, but `cfg(version("1.81.65536"))` will be false. Since the magnitude of what gets parsed seems to matter, it might be good to be clear about that.

This is a bit extreme, though. Even if an implementation used u8, it's still 19 years before 1.256 (assuming we stick with 6 week releases), and seems silly for someone to put an unnecessarily large number there.

est31: hmm yeah good point. For future compatibility we will warn on unknown version literal formats, which includes `1.81.65536`. We should probably word it in a way that the unknown formats are not guaranteed to evaluate to false.

PR #1829, opened 2025-05-17
Add doc for keylocker target features

sayantn:
Adds some docs for the target features `kl` and `widekl`, in lieu of the stabilization PR at rust-lang/rust#140766 and rust-lang/stdarch#1795.

Also, please lmk if anyone finds a better link for them, the wikipedia page is not _too_ good.

@rustbot label S-waiting-on-stabilization


traviscross (approved):

ehuss (approved):
Thanks!

PR #1830, opened 2025-05-17
Add doc for `sha512`, `sm3` and `sm4` target features

sayantn:
Add some docs for the stabilization of these in rust-lang/rust#140767 and rust-lang/stdarch#1796.

Also, please lmk if anyone finds a better source for them (especially `sm4`), the wikipedia page is not too helpful

@rustbot label S-waiting-on-stabilization


traviscross (approved):

ehuss (approved):
Thanks!

PR #1833, opened 2025-05-24
triagebot: enable issue transfer

lolbinarycat:
I believe it was already possible to transfer issues into the reference repo without this, but this will allow transferring issues both ways.

Issue transfer is still only available to members of teams and certain working groups.

https://forge.rust-lang.org/triagebot/transfer.html

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -5,6 +5,8 @@ allow-unauthenticated = [
 
 [assign]
 
+[transfer]
+
 [shortcut]
 
 [no-mentions]


traviscross:
@ehuss and I are talking about this on the lang-docs call, and while we're not opposed to adding this, we're struggling to think of a case where transferring an issue out of the Reference would be the right thing to do.

Even if, e.g., someone filed something against the Reference and we decided that it was actually a bug in the compiler, it would seem better to just close and refile.  The issue filed against the Reference is likely to be written in such a way that it wouldn't make a good issue against `r-l/r` on its own.

We're curious to know if there are other cases one might have in mind here.

ehuss:
I'm going to close for now to clear up the queue. We can always revisit this later. In practice, I don't recall ever really needing to transfer issues, and since I and many other people have write access here, they can use the UI to do that.

PR #1835, opened 2025-05-26
Document inferred const args (`feature(generic_arg_infer)`)

BoxyUwU:
Tracking: https://github.com/rust-lang/rust/issues/85077
Stabilization: https://github.com/rust-lang/rust/pull/141610
Stabilization report: https://github.com/rust-lang/rust/pull/141610#issuecomment-2910505867

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -35,7 +35,18 @@ r[expr.array.length-operand]
 The expression after the `;` is called the *length operand*.
 
 r[expr.array.length-restriction]
-It must have type `usize` and be a [constant expression], such as a [literal] or a [constant item].
+The length operand must either be an [inferred const] or be a [constant expression] of type `usize` (e.g. a [literal] or a [constant item]).
+
+```rust
+const C: usize = 1;
+let _: [u8; C] = [0; 1]; // Literal.
+let _: [u8; C] = [0; C]; // Constant item.
+let _: [u8; C] = [0; _]; // Inferred const.
+let _: [u8; C] = [0; (((_)))]; // Inferred const.
+```
+
+> [!NOTE]
+> In an array expression, an [inferred const] is parsed as an [expression][Expression] but then semantically treated as a separate kind of [const generic argument].
 
 r[expr.array.repeat-behavior]
 An array expression of this form creates an array with the length of the value of the length operand with each element being a copy of the repeat operand.
@@ -111,8 +122,10 @@ The array index expression can be implemented for types other than arrays and sl
 [IndexMut]: std::ops::IndexMut
 [Index]: std::ops::Index
 [array]: ../types/array.md
+[const generic argument]: items.generics.const.argument
 [constant expression]: ../const_eval.md#constant-expressions
 [constant item]: ../items/constant-items.md
+[inferred const]: items.generics.const.inferred
 [literal]: ../tokens.md#literals
 [memory location]: ../expressions.md#place-expressions-and-value-expressions
 [panic]: ../panic.md
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -146,43 +146,56 @@ r[items.generics.const.argument]
 A const argument in a [path] specifies the const value to use for that item.
 
 r[items.generics.const.argument.const-expr]
-The argument must be a [const expression] of the type ascribed to the const
-parameter. The const expression must be a [block expression][block]
-(surrounded with braces) unless it is a single path segment (an [IDENTIFIER])
-or a [literal] (with a possibly leading `-` token).
+The argument must either be an [inferred const] or be a [const expression] of the type ascribed to the const parameter. The const expression must be a [block expression][block] (surrounded with braces) unless it is a single path segment (an [IDENTIFIER]) or a [literal] (with a possibly leading `-` token).
 
 > [!NOTE]
 > This syntactic restriction is necessary to avoid requiring infinite lookahead when parsing an expression inside of a type.
 
 ```rust
-fn double<const N: i32>() {
-    println!("doubled: {}", N * 2);
-}
+struct S<const N: i64>;
+const C: i64 = 1;
+fn f<const N: i64>() -> S<N> { S }
 
-const SOME_CONST: i32 = 12;
-
-fn example() {
-    // Example usage of a const argument.
-    double::<9>();
-    double::<-123>();
-    double::<{7 + 8}>();
-    double::<SOME_CONST>();
-    double::<{ SOME_CONST + 5 }>();
-}
+let _ = f::<1>(); // Literal.
+let _ = f::<-1>(); // Negative literal.
+let _ = f::<{ 1 + 2 }>(); // Constant expression.
+let _ = f::<C>(); // Single segment path.
+let _ = f::<{ C + 1 }>(); // Constant expression.
+let _: S<1> = f::<_>(); // Inferred const.
+let _: S<1> = f::<(((_)))>(); // Inferred const.
 ```
+
+> [!NOTE]
+> In a generic argument list, an [inferred const] is parsed as an [inferred type][InferredType] but then semantically treated as a separate kind of [const generic argument].
 
 r[items.generics.const.inferred]
-Where a const argument is expected, an `_` (optionally surrounding by any number of matching parentheses), called the "inferred const", can be used instead. This asks the compiler to infer the const argument if possible based on surrounding information.
+Where a const argument is expected, an `_` (optionally surrounding by any number of matching parentheses), called the *inferred const* ([path rules][paths.expr.complex-const-params], [array expression rules][expr.array.length-restriction]), can be used instead. This asks the compiler to infer the const argument if possible based on surrounding information.
 
 ```rust
-fn make_buf() -> [u8; 1024] {
-    [0x1; _]
-    //    ^ Infers `1024`.
+fn make_buf<const N: usize>() -> [u8; N] {
+    [0; _]
+    //  ^ Infers `N`.
 }
+let _: [u8; 1024] = make_buf::<_>();
+//                             ^ Infers `1024`.
 ```
 
+> [!NOTE]
+> An [inferred const] is not semantically an [expression][Expression] and so is not accepted within braces.
+>
+> ```rust,compile_fail
+> fn f<const N: usize>() -> [u8; N] { [0; _] }
+> let _: [_; 1] = f::<{ _ }>();
+> //                    ^ ERROR `_` not allowed here
+> ```
+
 r[items.generics.const.inferred.constraint]
-It cannot be used in item signatures.
+The inferred const cannot be used in item signatures.
+
+```rust,compile_fail
+fn f<const N: usize>(x: [u8; N]) -> [u8; _] { x }
+//                                       ^ ERROR not allowed
+```
 
 r[items.generics.const.type-ambiguity]
 When there is ambiguity if a generic argument could be resolved as either a
@@ -306,6 +319,7 @@ struct Foo<#[my_flexible_clone(unbounded)] H> {
 [block]: ../expressions/block-expr.md
 [const contexts]: ../const_eval.md#const-context
 [const expression]: ../const_eval.md#constant-expressions
+[const generic argument]: items.generics.const.argument
 [const item]: constant-items.md
 [enumerations]: enumerations.md
 [functions]: functions.md
@@ -314,6 +328,7 @@ struct Foo<#[my_flexible_clone(unbounded)] H> {
 [generic parameter scopes]: ../names/scopes.md#generic-parameter-scopes
 [higher-ranked lifetimes]: ../trait-bounds.md#higher-ranked-trait-bounds
 [implementations]: implementations.md
+[inferred const]: items.generics.const.inferred
 [item declarations]: ../statements.md#item-declarations
 [item]: ../items.md
 [literal]: ../expressions/literal-expr.md
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -227,6 +227,27 @@ fn generic<const B: bool>() {
 }
 ```
 
+r[items.generics.const.inferred]
+
+r[items.generics.const.inferred.syntax]
+```grammar,items
+InferredConst ->
+      `_`
+    | `(` InferredConst `)`
+```
+
+r[items.generics.const.inferred.intro]
+The inferred const asks the compiler to infer the const argument if possible based on the surrounding information available.
+
+```rust
+fn make_buf() -> [u8; 1024] {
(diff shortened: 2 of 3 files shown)


src/const-generics.md
```diff
@@ -0,0 +1,174 @@
...
+
+It is often used in repeat expressions:
+```rust
+fn make_array() -> [u32; 2] {
+    [Default::default(); _]
+}
+```
+
+r[const-generics.concrete-expr]
+## Concrete expressions
```
BoxyUwU, line 107: Something I left out here is that the "concrete expressions" are not required to be quite so concrete for repeat expr counts as a backwards compatibility FCW. Does the reference intend to document the *actual state* of the compiler, or can I take liberties in documenting things as if FCWs are already hard errors?

BoxyUwU:
Sure, done

BoxyUwU:
@ehuss @traviscross any chance y'all will be able to get around to reviewing this soon? The stabilization PR is now ready to be merged and it'd be unfortunate to block a relatively minor feature that's already been waiting years, on this PR.

BoxyUwU:
Alternatively I could file a separate smaller PR that just adds a small extra bit of wording to the existing section noting that `_` can be used as an argument to const parameters. Then leave this more involved clean up/rework of const generics handling in the reference off of the "blocking" path for feature stabilization.

ehuss:
If possible, I think it would be best to drop the rearrangement commit. This probably isn't the organization we'll land on, and I would prefer to keep these as separate concerns so they don't block one another.

BoxyUwU:
Ok this should be fine now

src/items/generics.md
```diff
@@ -227,6 +227,22 @@ fn generic<const B: bool>() {
 }
 ```
 
+r[items.generics.const.inferred]
+
+r[items.generics.const.inferred.syntax]
+```grammar,types
+InferredConst -> 
+      `_`
+    | (InferredConst)
```
traviscross:
```suggestion
    | `(` InferredConst `)`
```

traviscross: Also, this rule will need to be used somewhere.

BoxyUwU: thanks :3

BoxyUwU: okay @ root didn't fix anything. I'm not sure how i'm supposed to use this anywhere

BoxyUwU:
@traviscross @ehuss friendly poke about this :)

src/items/generics.md
```diff
@@ -227,6 +227,22 @@ fn generic<const B: bool>() {
...
 ```
 
+r[items.generics.const.inferred]
+
+r[items.generics.const.inferred.syntax]
+```grammar,const
+InferredConst ->
+      `_`
+    | `(` InferredConst `)`
+```
```
ehuss: I'm not sure I fully understand this grammar since it isn't referred to anywhere.

Am I understanding correctly that this is added for GenericArgsConst? Or is it somewhere else? That is, should it be:

```
GenericArgsConst ->
      BlockExpression
    | LiteralExpression
    | `-` LiteralExpression
    | SimplePathSegment
    | InferredConst

InferredConst ->
      `_`
    | `(` InferredConst `)`
```

src/items/generics.md
```diff
@@ -227,6 +227,22 @@ fn generic<const B: bool>() {
...
+
+r[items.generics.const.inferred.intro]
+The inferred const asks the compiler to infer the const argument if possible
+based on the surrounding information available.
+
+r[items.generics.const.inferred.constraint]
+It cannot be used in item signatures.
```
ehuss: I'm not quite following this. If we are defining this as a new form allowed in generic *args*, why would there be a rule that it is not allowed in item signatures?

src/items/generics.md
```diff
@@ -227,6 +227,22 @@ fn generic<const B: bool>() {
...
+r[items.generics.const.inferred.syntax]
+```grammar,const
+InferredConst ->
+      `_`
+    | `(` InferredConst `)`
+```
+
+r[items.generics.const.inferred.intro]
+The inferred const asks the compiler to infer the const argument if possible
+based on the surrounding information available.
```
ehuss: Can you add a semi-realistic looking example here?

ehuss (commented):
I'm a little lost, as it seems like there are two distinct things being supported here:

1. `_` in generic args: `foo::<_>()`
2. `_` in array expression repeat lengths: `[123; _]`

Am I understanding that correctly? If so, shouldn't there be some update to the array expression chapter? And the underscore-expr.md may need a significant update to mention that it is allowed for array repeats.

src/items/generics.md
BoxyUwU: Yep exactly :+1: It's a little bit funny because it means that InferredType and InferredConst share the same grammar but I'm not sure that matters too much for this?

src/items/generics.md
BoxyUwU: We don't allow type inference in signatures in general, not just as a restriction on const arguments. For example this is an error on stable:
```rust
struct Foo<T>(T);
fn bar(param: Foo<_>) {}
```

src/items/generics.md
BoxyUwU: ah yeah sure

BoxyUwU:
> I'm a little lost, as it seems like there are two distinct things being supported here

There is one thing being added, we support `_` (optionally wrapped in arbitrary many parens) as const arguments. The repeat count of an array expr is a const argument so falls under the same thing as in `let a: [u8; _] = ...`. This is equivalent in functionality to how we support `_` (optionally wrapped in arbitrary many parens) as type arguments (e.g. `let a: [_; 2]`).

I didn't think the array-expr chapter would need updating as I've updated the definition of a const argument but it seems like the reference attempts to separately specify repeat expr counts from const arguments. This also means that the chapter does not currently talk about the fact that `[0x1; N + 1]` is illegal. I'm not sure what to do about that, can we just leave it as-is or would you prefer me to update the array-expr chapter to specify repeat counts as being const arguments.

> And the underscore-expr.md may need a significant update

I don't think we need to update underscore-expr, this feature is (confusingly) entirely orthogonal from `_` as an expression. `[u8; _]` isnt giving the `_` expression new meaning, its adding a new *kind* of const argument that is not an expression.

src/items/generics.md
ehuss: Isn't this already stated in type.inferred.constraint (inferred.md)?

BoxyUwU: Yeah we state that for `_`  type arguments but I felt like like we should note that the restriction applies for `_` const arguments too when discussing them as it might be confusing otherwise; people might get the idea that you *can* use `_` const arguments in signatures.

ehuss:
> The repeat count of an array expr is a const argument

If I'm understanding correctly, we define the array repeat length as a [const expression](https://doc.rust-lang.org/nightly/reference/const_eval.html#r-const-eval.const-expr.const-context) with the limitations defined in [const context](https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context) among other places.

If we are adding `_` as an allowed expression for a const expression in this context, then it seems like something that needs to be added to const chapter and the array chapter since there is nothing that says it is allowed in that position. For example, the const expr section doesn't say it is allowed. Similarly, as I understand it, the operand for the array expression length expression is any Expression with the restriction that it has to play by the const-expression limitations (and various other limitations like [items.generics.const.standalone](https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.const.standalone).

> falls under the same thing as in let a: [u8; _] = ....

I'm not quite following this. That looks like an `_` in type position which we define as the [inferred type](https://doc.rust-lang.org/nightly/reference/types/inferred.html#grammar-InferredType) which seems to be something different from [`_` in expressiosn position](https://doc.rust-lang.org/nightly/reference/expressions/underscore-expr.html).

I understand that there may be some implementation similarities, but as we define type vs expression as two different concepts, I would think that introducing `_` in expression contexts for array lengths is something novel here.

> This also means that the chapter does not currently talk about the fact that `[0x1; N + 1]` is illegal.

I believe these limitations are defined in https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context and https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.const.standalone.

BoxyUwU:
Sorry I feel like I'm repeating myself a bit here and am not sure how to explain this better. This feature is orthogonal to underscores as an expression. It's not that `_` is now a valid expression by itself, it's that there is a new kind of *const generic argument* (which happens to have the same syntax as the _ expression).

Maybe the core confusion here is about this distinction between a const *expression* and a const *argument*?

> If I'm understanding correctly, we define the array repeat length as a [const expression](https://doc.rust-lang.org/nightly/reference/const_eval.html#r-const-eval.const-expr.const-context)

I think the entire way that const generics are handled in the reference right now is just not structured right for this feature to be honest ^^.

It doesn't make sense to talk about const arguments as being "expressions with more restrictions"; `_` arguments are *not* expressions and are instead a new *kind* of const argument. But as far as I can tell there is no distinct concept of a "const argument" in the reference (like how there are different kinds of for type arguments). Instead everywhere simply specifies things as being "const expressions" with some additional restrictions.

I originally tried to refactor the reference to properly talk about const arguments as a real concept in the type system section with different sections for each kind of const argument. But that was quite a big change and this seemed smaller/easier to review.

Do you think you'd prefer going back to the old approach that has more involved explanations of this topic? It sounds like the new version of this PR is actually quite confusing in practice which is unfortunate.

Admittedly, I'm finding this PR to be quite stressful. I'd really like to get `generic_arg_infer` stabilized and having it be blocked on this when there's such a disconnect between the language itself and how the reference specifies things is making me worried that it's going to wind up blocked for a really long time.

I'd be more than happy to put a bunch of time into documenting things in the reference and talking things over with you, but if that's where things need to go in order to get something you're happy with would it be possible to just land the stabilization PR *first* and continue work on documenting this afterwards?

> I'm not quite following this. That looks like an _ in type position

I guess with this terminology it would be correct to say that the repeat count of array exprs is a type position ^^. Though being nitpicky it doesn't make sense to say that a `_` there is an inferred type because it is not a type, its a const generic value. Hence why I've added the concept of an "inferred const" to the reference which roughly mirrors the structure of the inferred type documentation.

> I believe these limitations are defined in https://doc.rust-lang.org/nightly/reference/const_eval.html#const-context and https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.const.standalone.

Ah yes you're right it is documented :+1: Then I guess array-expr really doesn't need updating because it is specified as being a const generic argument :D

src/paths.md
```diff
@@ -65,6 +65,7 @@ GenericArgsConst ->
     | LiteralExpression
     | `-` LiteralExpression
     | SimplePathSegment
+    | InferredConst
```
traviscross: This ends up making the grammar ambiguous, because `GenericArgsConst` is in `GenericArg`, and that also contains `Type`, which contains `ParenthesizedType` and `InferredType`.

Not sure how we want to handle that.  What are your thoughts, @ehuss?

BoxyUwU: The "proper" way to represent this would be to maintain a distinction between ambig vs unambig positions. So we have one grammar for consts/types in ambiguous positions and one for unambiguous positions. Then you  have an `InferredArg` thing on `GenericArg` and the grammar for consts/types in ambig positions doesn't contain InferredType/InferredConst.

BoxyUwU: That seems quite involved though and given the initial reaction to this PR was that it was way too big I was hesitant to actually refactor this too.

BoxyUwU: Possibly also a `ParenthesisedTyOrConst` thing on `GenericArgs` to solve the parens issue too

BoxyUwU: Basically you'd split `GenericArgsConst/Type` into `GenericArgsConst/TypeUnambig/Ambig` and then lift the shared syntax out of the `Ambig` versions and onto `GenericArg` which would also be changed to use the `Ambig` forms

BoxyUwU: ...and then update everywhere in the grammar elsewhere to say `Unambig` in cases where we know for sure we're parsing a type, e.g. function parameters' type

traviscross: In other discussion, @BoxyUwU pointed out this is a problem elsewhere for generic arguments, as one could similarly resolve down either the type productions or the const generic productions.  I buy that.

traviscross:
~~In separate discussion, based on Boxy's confirmation that the `ArrayElements` production (an array expression) and the `ArrayType` production can correctly be changed to use `GenericArgsConst` for the length, rather than `Expression`, I've suggested making that targeted change (and making targeted matching adjustments to the corresponding verbiage), as it would resolve the matter about the fact that `_` isn't a const generic expression but is a new kind of const generic argument.~~

Edit: It turns out that's not correct.

src/expressions/array-expr.md
```diff
@@ -26,16 +26,16 @@ r[expr.array.array-behavior]
 This produces an array containing each of these values in the order they are written.
 
 r[expr.array.repeat]
-The syntax for the second form is two expressions separated by a semicolon (`;`) enclosed in square brackets.
+The syntax for the second form is an expression and a const argument separated by a semicolon (`;`) enclosed in square brackets.
```
traviscross: What's the theory behind "expression and a const argument"?

BoxyUwU: as in, it consists of *both* an expression and a const argument. what would ur preferred form for this sentence be?

BoxyUwU: [expr; constargument]

traviscross: Got it, thanks.

src/items/generics.md
```diff
@@ -227,6 +227,27 @@ fn generic<const B: bool>() {
 }
 ```
 
+r[items.generics.const.inferred]
+
+r[items.generics.const.inferred.syntax]
+```grammar,const
```
traviscross:
```suggestion
```grammar,items
```

(I'll fix.)

src/expressions/array-expr.md
```diff
@@ -9,7 +9,7 @@ ArrayExpression -> `[` ArrayElements? `]`
 
 ArrayElements ->
       Expression ( `,` Expression )* `,`?
-    | Expression `;` Expression
+    | Expression `;` GenericArgsConst
```
traviscross: Unfortunately this is not correct, as we accept:

```rust
_ = [0u8; 1 + 2];
```

Thoughts on the right approach?

traviscross: After discussion, we all agreed this should just be `Expression` after all.

traviscross (approved):
@ehuss and I talked through this on a call today, resulting in the updates here that we're both happy with.  We may do some proofreading of the language here, but this is basically good to go, so we can approve this.

PR #1837, opened 2025-05-30
Document how closure capturing interacts with discriminant reads

meithecatte:
This is the behavior after the bugfixes in rust-lang/rust#138961. I have successfully ran `mdbook test` with `RUSTUP_TOOLCHAIN` pointed at a stage1 built on top of the aforementioned PR – until it's merged, the CI here will fail.

### DIFF
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -288,7 +288,11 @@ let c = || {
 r[type.closure.capture.precision.discriminants]
 ### Capturing for discriminant reads
 
-If pattern matching requires inspecting a discriminant, the relevant place will get captured by `ImmBorrow`.
+r[type.closure.capture.precision.discriminants.reads]
+If pattern matching reads a discriminant, the place containing that discriminant is captured by `ImmBorrow`.
+
+r[type.closure.capture.precision.discriminants.multiple-variant]
+Matching against a variant of an enum that has more than one variant reads the discriminant, capturing the place by `ImmBorrow`.
 
 ```rust
 enum Example {
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -354,14 +354,14 @@ c();
 ```
 
 r[type.closure.capture.precision.discriminants.range-patterns]
-Matching against a [range pattern][patterns.range] performs a read of the place being matched, causing the closure to borrow it by `ImmBorrow`. This is the case even if the range matches all possible values.
+Matching against a [range pattern][patterns.range] reads the place being matched, even if the range includes all possible values of the type, and captures the place by `ImmBorrow`.
 
-```rust,compile_fail,E0506
-let mut x = 7_u8;
+```rust,compile_fail,E0502
+let mut x = 0u8;
 let c = || {
-    let 0..=u8::MAX = x; // captures `x` by ImmBorrow
+    let 0..=u8::MAX = x; // Captures `x` by `ImmBorrow`.
 };
-x += 1; // ERROR: cannot assign to `x` because it is borrowed
+let _ = &mut x; // ERROR: Cannot borrow `x` as mutable.
 c();
 ```
 
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -300,7 +300,7 @@ let mut x = Example::A(42);
 let c = || {
     let Example::A(_) = x; // does not capture `x`
 };
-x = Example::A(57); // x can be modified while the closure is live
+x = Example::A(57); // `x` can be modified while the closure is live
 c();
 ```
 
@@ -327,6 +327,55 @@ let c = || {
     let Example::A(_) = x; // captures `x` by ImmBorrow
 };
 x = Example::A(57); // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.range-patterns]
+Matching against a [range pattern][patterns.range] constitutes a discriminant read, even if
+the range matches all possible values.
+
+```rust,compile_fail,E0506
+let mut x = 7_u8;
+let c = || {
+    let 0..=u8::MAX = x; // captures `x` by ImmBorrow
+};
+x += 1; // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.slice-patterns]
+Matching against a [slice pattern][patterns.slice] constitutes a discriminant read if
+the slice pattern needs to inspect the length of the scrutinee.
+
+```rust,compile_fail,E0506
+let mut x: &mut [i32] = &mut [1, 2, 3];
+let c = || match x { // captures `*x` by ImmBorrow
+    [_, _, _] => println!("three elements"),
+    _ => println!("something else"),
+};
+x[0] += 1; // ERROR: cannot assign to `x[_]` because it is borrowed
+c();
+```
+
+Thus, matching against an array doesn't constitute a discriminant read, as the length is fixed.
+
+```rust
+let mut x: [i32; 3] = [1, 2, 3];
+let c = || match x { // does not capture `x`
+    [_, _, _] => println!("three elements, obviously"),
+};
+x[0] += 1; // `x` can be modified while the closure is live
+c();
+```
+
+Likewise, a slice pattern that matches slices of all possible lengths does not constitute a discriminant read.
+
+```rust
+let mut x: &mut [i32] = &mut [1, 2, 3];
+let c = || match x { // does not capture `x`
+    [..] => println!("always matches"),
+};
+x[0] += 1; // `x` can be modified while the closure is live
 c();
 ```
 
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -338,9 +338,7 @@ c();
 ```
 
 r[type.closure.capture.precision.discriminants.non_exhaustive]
-If [the `#[non_exhaustive]` attribute][non_exhaustive] is applied to an enum defined in an external crate, it is considered to have multiple variants, even if only one variant is actually present.
-
-[non_exhaustive]: attributes.type-system.non_exhaustive
+If [`#[non_exhaustive]`][attributes.type-system.non_exhaustive] is applied to an enum defined in an external crate, the enum is treated as having multiple variants for the purpose of deciding whether a read occurs, even if it actually has only one variant.
 
 r[type.closure.capture.precision.discriminants.uninhabited-variant]
 Even if all other variants are uninhabited, the discriminant read still occurs.
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -186,9 +186,8 @@ If this were to capture `m`, then the closure would no longer outlive `'static`,
 r[type.closure.capture.precision.wildcard]
 ### Wildcard pattern bindings
 
-Closures only capture data that needs to be read.
-Binding a value with a [wildcard pattern] does not count as a read, and thus won't be captured.
-For example, the following closures will not capture `x`:
+r[type.closure.capture.precision.wildcard.reads]
+Closures only capture data that needs to be read. Binding a value with a [wildcard pattern] does not read the value, so the place is not captured.
 
 ```rust
 let x = String::from("hello");
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -340,22 +340,16 @@ c();
 r[type.closure.capture.precision.discriminants.non_exhaustive]
 If [`#[non_exhaustive]`][attributes.type-system.non_exhaustive] is applied to an enum defined in an external crate, the enum is treated as having multiple variants for the purpose of deciding whether a read occurs, even if it actually has only one variant.
 
-r[type.closure.capture.precision.discriminants.uninhabited-variant]
-Even if all other variants are uninhabited, the discriminant read still occurs.
+r[type.closure.capture.precision.discriminants.uninhabited-variants]
+Even if all variants but the one being matched against are uninhabited, making the pattern [irrefutable][patterns.refutable], the discriminant is still read if it otherwise would be.
 
-```rust,compile_fail,E0506
-enum Void {}
-
-enum Example {
-    A(i32),
-    B(Void),
-}
-
-let mut x = Example::A(42);
+```rust,compile_fail,E0502
+enum Empty {}
+let mut x = Ok::<_, Empty>(42);
 let c = || {
-    let Example::A(_) = x; // captures `x` by ImmBorrow
+    let Ok(_) = x; // Captures `x` by `ImmBorrow`.
 };
-x = Example::A(57); // ERROR: cannot assign to `x` because it is borrowed
+let _ = &mut x; // ERROR: Cannot borrow `x` as mutable.
 c();
 ```
 
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -205,6 +205,9 @@ c();
 r[type.closure.capture.precision.wildcard.destructuring]
 Destructuring tuples, structs, and single-variant enums does not, by itself, cause a read or the place to be captured.
 
+> [!NOTE]
+> Enums marked with [`#[non_exhaustive]`][attributes.type-system.non_exhaustive] from other crates are always treated as having multiple variants. See *[type.closure.capture.precision.discriminants.non_exhaustive]*.
+
 ```rust,no_run
 struct S; // A non-`Copy` type.
 
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -202,6 +202,7 @@ let c = || match x {  // x is not captured
 c();
 ```
 
+r[type.closure.capture.precision.wildcard.destructuring]
 Destructuring tuples, structs, and single-variant enums does not, by itself, cause a read or the place to be captured.
 
 ```rust,no_run
--- a/src/types/closure.md
+++ b/src/types/closure.md
@@ -342,7 +342,7 @@ x = Example::A(57); // `x` can be modified while the closure is live
 c();
 ```
 
-r[type.closure.capture.precision.discriminants.non-exhaustive]
+r[type.closure.capture.precision.discriminants.non_exhaustive]
 If [the `#[non_exhaustive]` attribute][non_exhaustive] is applied to an enum defined in an external crate, it is considered to have multiple variants, even if only one variant is actually present.
 
 [non_exhaustive]: attributes.type-system.non_exhaustive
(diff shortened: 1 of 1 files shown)


src/types/closure.md
```diff
@@ -264,6 +267,69 @@ let c = || {
 
 [wildcard pattern]: ../patterns.md#wildcard-pattern
 
+r[type.closure.capture.precision.discriminants]
+### Capturing for discriminant reads
+
+If pattern matching requires inspecting a discriminant, the relevant place will get captured by `ImmBorrow`.
```
Nadrieril, line 273: Unless we mutably access the contents, right? E.g.
```rust
let c = || match x {
    (Example::A(ref mut _y), _) => println!("variant A"),
    (Example::B(_), _) => println!("variant B"),
};
```

Nadrieril: ok yep, that's the "Shared prefix" thing

Nadrieril (approved):
The changes look accurate to me. I don't know about things related to how the reference is organized so I'll leave to someone who knows to approve this.

src/types/closure.md
meithecatte: Yup. Do you think this deserves a clarification? (if so, how would you word it)

Nadrieril: no, I think it's clear in the context

src/types/closure.md
```diff
@@ -264,6 +267,118 @@ let c = || {
...
+```rust
+let mut x: [i32; 3] = [1, 2, 3];
+let c = || match x { // does not capture `x`
+    [_, _, _] => println!("three elements, obviously"),
+};
+x[0] += 1; // `x` can be modified while the closure is live
+c();
+```
+
+Likewise, a slice pattern that matches slices of all possible lengths does not constitute a discriminant read.
```
joshtriplett: We're talking about changing this behavior. cc @Nadrieril.

That said, I think we should merge it in this form, and then update this when we've changed this behavior.

meithecatte: Yes, the idea to change this came from the PR that is being documented by this change.

src/types/closure.md
```diff
@@ -264,6 +267,118 @@ let c = || {
...
+
+```rust
+let mut x: &mut [i32] = &mut [1, 2, 3];
```
ehuss:
```suggestion
let x: &mut [i32] = &mut [1, 2, 3];
```

src/types/closure.md
```diff
@@ -264,6 +267,118 @@ let c = || {
...
+x += 1; // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.slice-patterns]
+Matching against a [slice pattern][patterns.slice] constitutes a discriminant read if
+the slice pattern needs to inspect the length of the scrutinee.
+
+```rust,compile_fail,E0506
+let mut x: &mut [i32] = &mut [1, 2, 3];
```
ehuss:
```suggestion
let x: &mut [i32] = &mut [1, 2, 3];
```

src/types/closure.md
```diff
@@ -264,6 +267,118 @@ let c = || {
...
+let c = || {
+    let Example::A(_) = x; // captures `x` by ImmBorrow
+};
+x = Example::A(57); // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.range-patterns]
+Matching against a [range pattern][patterns.range] constitutes a discriminant read, even if
+the range matches all possible values.
```
ehuss: I'm not quite following this terminology, particularly around the user of the term "discriminant". My understanding is that a discriminant is specifically a tag associated with an enum. I wouldn't consider an integer to have a discriminant. Can you clarify what this means?

src/types/closure.md
```diff
@@ -264,6 +267,118 @@ let c = || {
...
+let c = || {
+    let 0..=u8::MAX = x; // captures `x` by ImmBorrow
+};
+x += 1; // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.slice-patterns]
+Matching against a [slice pattern][patterns.slice] constitutes a discriminant read if
+the slice pattern needs to inspect the length of the scrutinee.
```
ehuss: Same question about the use of the term "discriminant".

ehuss (commented):
Thanks, can you also unwrap any new text being added?

src/types/closure.md
meithecatte: Right, this happened because the relevant rustc has a distinction between (a) binding sub-values to identifiers with various binding modes, and (b) discriminant reads—with other "shallow/immediate" reads, such as the length field of a slice, being shoe-horned into the category of discriminant reads.

From the perspective of reference text, this obviously doesn't make sense. I'll rephrase that as just "constitutes a read", I think, unless you have any better ideas on how to word this?

ehuss: That sounds good to me.

meithecatte: Ah, I gave this another look and it's a "discriminant read" in the sense of `type.closure.capture.precision.discriminants`. I'll try to find a way to reword this all.

meithecatte:
I've reworded the slice and range sections to not use the "discriminant read" term, do you think this is clearer?

src/types/closure.md
```diff
@@ -344,11 +340,10 @@ c();
 ```
 
 r[type.closure.capture.precision.discriminants.slice-patterns]
-Matching against a [slice pattern][patterns.slice] constitutes a discriminant read if
-the slice pattern needs to inspect the length of the scrutinee.
+Matching against a [slice pattern][patterns.slice] performs a read if the slice pattern needs to inspect the length of the scrutinee. The read will cause the closure to borrow the relevant place by `ImmBorrow`.
```
Nadrieril: Maybe say "performs a read of the pointer value" to avoid confusion about what is being read? Should we even say that this reads only the pointer metadata (not sure that's what we're doing today even)?

meithecatte: Okay, how about this new revision?

Nadrieril (approved):

Nadrieril:
This looks good for me, could someone with authority approve this plz?

traviscross:
Thanks.  Reviewing.

src/types/closure.md
```diff
@@ -264,6 +266,114 @@ let c = || {
...
+x += 1; // ERROR: cannot assign to `x` because it is borrowed
+c();
+```
+
+r[type.closure.capture.precision.discriminants.slice-patterns]
+Matching against a [slice pattern][patterns.slice] that needs to inspect the length of the scrutinee performs a read of the pointer value in order to fetch the length. The read will cause the closure to borrow the relevant place by `ImmBorrow`.
+
+```rust,compile_fail,E0506
+let x: &mut [i32] = &mut [1, 2, 3];
+let c = || match x { // captures `*x` by ImmBorrow
```
traviscross: Does it capture `*x` or does it capture `x`?  It would seem to make more sense for it to be capturing the wide pointer itself rather than the pointee, and the MIR makes it look like that's what it's doing (i.e., within the closure, `x: &&mut [i32]`).

src/types/closure.md
```diff
@@ -98,7 +98,9 @@ Async closures always capture all input arguments, regardless of whether or not
 ## Capture Precision
 
 r[type.closure.capture.precision.capture-path]
-A *capture path* is a sequence starting with a variable from the environment followed by zero or more place projections that were applied to that variable.
+A *capture path* is a sequence starting with a variable from the environment followed by zero or more place projections that were applied to that variable, as well as any [further projections performed by matching against patterns][pattern-wildcards].
+
+[pattern-wildcards]: type.closure.capture.precision.wildcard
```
traviscross: This is a bit of an odd link.  The new text is talking about "further projections performed by matching against patterns", but then it links to a section that's about things that we don't capture.

traviscross: It also seems that this new language might better fit in the rule below.

traviscross (approved):

traviscross:
This PR was in good shape.  Thanks to @meithecatte for that and to @Nadrieril for the review of it.  I've pushed some mostly editorial adjustments.  Please have a look, of course, for any errors.  I'll ask @ehuss to have a look at well on our lang-docs call.  Overall, though, this is OK to go forward, so we can mark this as approved and unblock the underlying stabilization.

src/types/closure.md
meithecatte: I don't have the bandwidth to test this myself right now, but I assume you're hitting this optimization: https://github.com/rust-lang/rust/blob/016baacaf6ed5698625036e3afe3900535015680/compiler/rustc_hir_typeck/src/upvar.rs#L2543-L2572

My understanding, based on knowledge that has faded a bit already, is that the language semantics says we capture `*x`, but the actual compiled output optimizes that to capturing `x` instead, on account of that not being observable. I assume that if you stare 5–10 minutes at the doc comment I linked, you'll be able to come up with an example where the optimization doesn't apply and the capture of `*x` can be observed in the MIR.

traviscross: Thanks for those details.  I've revised it to state explicitly that we capture the pointee in this case and added an admonition referencing the doc comment you mention.

src/types/closure.md
traviscross: Addressed this in the revisions.

src/types/closure.md
traviscross: In reviewing this with @ehuss, we found, intriguingly, that prior to your PR https://github.com/rust-lang/rust/pull/138961, it captures the pointer, and after your PR, it captures the pointee.  E.g., with this code:

```rust
fn main() {
    let x: &mut [u8] = &mut [];
    let c = || match x {
        [] => (),
        _ => (),
    };
    c();
}
```

Without the PR, we get this MIR for the closure:

```rust
fn main::{closure#0}(_1: &{closure@src/main.rs:3:13: 3:15}) -> () {
    debug x => (*((*_1).0: &&mut [u8]));
    let mut _0: ();

    bb0: {
        return;
    }
}
```

With the PR, we get this MIR instead:

```rust
fn main::{closure#0}(_1: &{closure@src/main.rs:3:13: 3:15}) -> () {
    debug x => (*((*_1).0: &[u8]));
    let mut _0: ();

    bb0: {
        return;
    }
}
```

src/types/closure.md
```diff
@@ -185,85 +189,259 @@ If this were to capture `m`, then the closure would no longer outlive `'static`,
...
 };
 ```
 
 [wildcard pattern]: ../patterns.md#wildcard-pattern
 
+r[type.closure.capture.precision.discriminants]
+### Capturing for discriminant reads
+
+r[type.closure.capture.precision.discriminants.reads]
+If pattern matching reads a discriminant, the place containing that discriminant is captured by `ImmBorrow`.
```
Nadrieril, line 295: Question: shouldn't this bit be enough for the section, and shouldn't the "which discriminants are read" be moved to a more appropriate place, like wherever we describe pattern-matching?

Nadrieril: Even if we do that I'd keep the examples, as they're a really good way to observe what exactly is being read.

src/types/closure.md
```diff
@@ -98,10 +98,13 @@ Async closures always capture all input arguments, regardless of whether or not
 ## Capture precision
 
 r[type.closure.capture.precision.capture-path]
-A *capture path* is a sequence starting with a variable from the environment followed by zero or more place projections that were applied to that variable.
+A *capture path* is a sequence starting with a variable from the environment followed by any number of place projections from that variable.
```
Nadrieril: nit: I find this change reduces clarity because "zero projections" is not an immediately obvious option. I like "zero or more", or maybe this:
```suggestion
A *capture path* is a sequence starting with a variable from the environment optionally followed by any number of place projections from that variable.
```

src/types/closure.md
Nadrieril: Having read things in more detail there's actually an open question here: we might want the static semantics and runtime semantics to differ. The case I have in mind is `#[non_exhaustive]` enums: for closure capture we should definitely pretend they have multiple variants, but I'm not sure we should emit the discriminant read in MIR (this [recently](https://github.com/rust-lang/rust/issues/147722) tripped someone up).

Nadrieril:
EDIT: nevermind, moved this to another issue

src/types/closure.md
Nadrieril: More discussion is happening in https://github.com/rust-lang/rust/issues/147722.

traviscross: Yes, I'd had the same thoughts.  I agree that it makes more sense, high level, for the dynamic semantics of patterns with respect to reads to be in the patterns chapter.  At the same time, I think there'd be some risk of the rules then that affect the static semantics with respect to captures being somewhat spread out.

Given how tied in these reads are to the question of the static semantics described here, and given that for better or worse the Reference is more organized by static semantics currently, I'm OK with leaving this reorganization to future work.

Nadrieril: Sounds good. I recall that the dynamic semantics of patterns needs quite some work anyway, that'll be its own endeavor I imagine.

src/types/closure.md
```diff
@@ -185,85 +189,259 @@ If this were to capture `m`, then the closure would no longer outlive `'static`,
...
+let x = E::V(());
+let c = || {
+    let E::V(_) = x; // Does not capture `x`.
+};
+x; // OK: `x` can be moved here.
+c();
+```
+
+r[type.closure.capture.precision.discriminants.non_exhaustive]
+If [`#[non_exhaustive]`][attributes.type-system.non_exhaustive] is applied to an enum defined in an external crate, the enum is treated as having multiple variants for the purpose of deciding whether a read occurs, even if it actually has only one variant.
```
theemathas, line 344: Can adding/removing `#[non_exhaustive]` affect run time behavior in closure captures? See also https://github.com/rust-lang/rust/issues/147722

Can removing `#[non_exhaustive]` cause code in another crate to stop compiling?

meithecatte: > Can removing `#[non_exhaustive]` cause code in another crate to stop compiling?

Unfortunately, yes. If we have the following upstream library `crate1`:

```rust
#[non_exhaustive]
pub enum Meow {
    A(u32, String),
}
```

and a downstream crate with the following code

```rust
use crate1::Meow;

pub fn f(x: Meow) -> impl FnOnce() {
    || {
        match x {
            Meow::A(a, b) => {
                drop((a, b));
            }
            _ => unreachable!(),
        }
    }
}
```

then with rust-lang/rust#138961, removing `#[non_exhaustive]` breaks the downstream crate. This would be another argument for the change proposed in rust-lang/rust#147722, i.e. that regardless of `#[non_exhaustive]`, there should be a read of the discriminant.

meithecatte: > Can adding/removing `#[non_exhaustive]` affect run time behavior in closure captures?

I gave this part of your question some thought and I guess the closest you'd get to that is the potential for a difference in the size of some closures due to a different set of precise captures being computed...

SimonSapin: I had trouble seeing what the error would be so I reproduced it (in Nightly)

```rust
error[E0373]: closure may outlive the current function, but it borrows `x.0`, which is owned by the current function
  --> src/main.rs:8:5
   |
 8 |     || {
   |     ^^ may outlive borrowed value `x.0`
 9 |         match x {
   |               - `x.0` is borrowed here
   |
note: closure is returned here
  --> src/main.rs:8:5
   |
 8 | /     || {
 9 | |         match x {
10 | |             Meow::A(a, b) => {
11 | |                 drop((a, b));
...  |
15 | |     }
   | |_____^
help: to force the closure to take ownership of `x.0` (and any other referenced variables), use the `move` keyword
   |
 8 |     move || {
   |     ++++

For more information about this error, try `rustc --explain E0373`.
```

(I filed a diagnostic papercut https://github.com/rust-lang/rust/issues/150144)

ehuss (approved):
Thanks! Merging now that https://github.com/rust-lang/rust/pull/138961 has merged.

src/types/closure.md
theemathas: > Can adding/removing `#[non_exhaustive]` affect run time behavior in closure captures?

Turns out the answer is **yes**. Consider the following code:

dep/src/lib.rs:
```rust
#[non_exhaustive]
pub enum Thing {
    One(i32, LoudDrop),
}

pub struct LoudDrop(pub &'static str);
impl Drop for LoudDrop {
    fn drop(&mut self) {
        println!("dropping {}", self.0);
    }
}
```

src/main.rs:
```rust
use dep::{LoudDrop, Thing};

#[allow(unused)]
fn main() {
    let mut thing = Thing::One(0, LoudDrop("a"));
    let closure = move || match thing {
        Thing::One(x, _) => {}
        _ => unreachable!(),
    };
    println!("before assign");
    thing = Thing::One(1, LoudDrop("b"));
    println!("after assign");
}
```

Using nightly rust, this code prints:
```
before assign
after assign
dropping a
dropping b
```

However, if we delete the `#[non_exhaustive]`, this code prints:
```
before assign
dropping a
after assign
dropping b
```

It seems that the `LoudDrop` is moved into the closure only if the enum is `#[non_exhaustive]`

src/types/closure.md
```diff
@@ -185,85 +189,259 @@ If this were to capture `m`, then the closure would no longer outlive `'static`,
...
 };
+let c = || match x {
+    _ => (), // Does not capture `x`.
+};
+x; // OK: `x` can be moved here.
 c();
+```
+
+r[type.closure.capture.precision.wildcard.destructuring]
+Destructuring tuples, structs, and single-variant enums does not, by itself, cause a read or the place to be captured.
```
RalfJung, line 209: Is the intention that these are the final semantics? I find it really strange to special-case single-variant enums.

RalfJung: Ah, that's basically the same as the [other thread](https://github.com/rust-lang/reference/pull/1837#discussion_r2532838829) above, which in turn is the same as https://github.com/rust-lang/rust/issues/147722.

theemathas: The other thread is not the same as https://github.com/rust-lang/rust/issues/147722. That thread concerns what is captured in a closure. Additionally, it involves a case where *removing* `#[non_exhaustive]` from an enum is a breaking change.

RalfJung: It is "the same" in the sense that it's all caused by special-casing single-variant enums except when they are foreign-crate non-exhaustive enums, isn't it?

theemathas: Fair enough. They're "the same" in that sense.

src/types/closure.md
meithecatte: Ah, I was considering something like this but couldn't get it to a full example because I forgot about drop being observable :sweat_smile:

PR #1838, opened 2025-06-03
remove the `safe` keyword from a c-variadic foreign function.

folkertdev:
Calling a c-variadic function is never safe: passing an unexpected number of arguments, or arguments of an unexpected type, is UB.

As requested in https://github.com/rust-lang/rust/pull/141733#issuecomment-2935156116, cc @ehuss

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -176,7 +176,7 @@ identifier.
 
 ```rust
 unsafe extern "C" {
-    safe fn foo(...);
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

traviscross (approved):

ehuss (approved):
Thanks! I went ahead and changed it to also include the `unsafe` keyword for consistency, and there is a general intention of making this more of a style it.

PR #1839, opened 2025-06-03
add a warning about using `safe` on extern c-variadic functions

folkertdev:
edit: the lang team found an edge case. 

An extern c-variadic function can be marked as safe when it does not look at its arguments at all. That is not very useful, but technically allowed. 

So, I've now added a warning with this information to the section on extern c-variadic functions.


traviscross (approved):

traviscross:
Thanks for the PR.  Pushed an update to revise a bit, to add a `SAFETY` comment, to frame this in terms of the guarantees that the function must make, and to use wording that makes it more clear that this would be library-level UB.

PR #1844, opened 2025-06-06
intro note: make text more simple

tshepang:
Grammar did not look correct, so preserve meaning, while making it more simple.


ehuss (approved):
Thanks!

Issue #1845, opened 2025-06-06
Confusing use of "wildcard" in destructuring pattern documentation


archer-321:
The reference contains the following sentence as [`patterns.destructure.placeholder`](https://doc.rust-lang.org/stable/reference/patterns.html#r-patterns.destructure.placeholder):

> In a pattern whose scrutinee expression has a struct, enum or tuple type, a placeholder (_) stands in for a single data field, whereas a wildcard .. stands in for all the remaining fields of a particular variant.

The same page defines the `_` pattern as a wildcard pattern and `..` as a rest pattern. Is this difference intended?

The line was introduced by [0003310e](https://github.com/rust-lang/reference/commit/0003310e8eb8fd588ea0868e16a17364d068e72c#diff-cc629f15712821139bc706c63b3845ab59a008e2a998e08ffad42e3aebcbcbe2R71-R77).

If this is indeed an inconsistency, would the following be an acceptable patch?

```diff
- In a pattern whose [scrutinee] expression has a `struct`, `enum` or `tuple` type, a placeholder (`_`) stands in for a *single* data field, whereas a wildcard `..`  stands in for *all* the remaining fields of a particular variant.
+ In a pattern whose [scrutinee] expression has a `struct`, `enum` or `tuple` type, a wildcard pattern (`_`) stands in for a *single* data field, whereas a rest pattern (`..`)  stands in for *all* the remaining fields of a particular variant.
```

iampi31415:
I'm newish and might be wrong.

My analysis is:

* The _underscore expression_ (`_`) [is different](https://doc.rust-lang.org/reference/expressions/underscore-expr.html) from the wildcard pattern (also `_`).
* I would not use "`..` wildcard". Because it is never defined.

So I think some adjustments are clearly needed.

---------

EDIT: removed some parts here to make easier to read the thread. They can be found in the reply below if needed, or history.

@archer-321

archer-321:
> * The _underscore expression_ (`_`) [is different](https://doc.rust-lang.org/reference/expressions/underscore-expr.html) from the wildcard pattern (also `_`).

Good catch! I hadn't noticed this before, as I was only focusing on patterns until now.

> For `..` there is the rest pattern, but there isn't an equivalent expression? They used "wildcard"; the rest pattern is not for Structs (excepting Tuple Structs):
> 
[...]

PR #1850, opened 2025-06-09
Fix(typo): 'though' should be 'through'

iampi31415:


ehuss (approved):
Thanks!

iampi31415:
You are welcome, could you look at #1845 if you get a chance @ehuss ? I think it's correct (not raised by myself)

PR #1854, opened 2025-06-12
Use extern "system" instead of "stdcall" in example

ChrisDenton:
`stdcall` is only correct for 32-bit Windows and may be a hard error on other platforms in the future. Whereas `system` is correct for all Windows targets.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -103,9 +103,8 @@ standard C ABI on the specific platform. Other ABIs may be specified using an
 `abi` string, as shown here:
 
 ```rust
-# #[cfg(any(windows, target_arch = "x86"))]
 // Interface to the Windows API
-unsafe extern "stdcall" { }
+unsafe extern "system" { }
 ```
 
 r[items.extern.abi.standard]


src/items/external-blocks.md
```diff
@@ -103,9 +103,8 @@ standard C ABI on the specific platform. Other ABIs may be specified using an
 `abi` string, as shown here:
 
 ```rust
-# #[cfg(any(windows, target_arch = "x86"))]
 // Interface to the Windows API
-unsafe extern "stdcall" { }
+unsafe extern "system" { }
```
ehuss, line 107: This now doesn't seem to quite match with the comment above about being for a Windows API (the example was intended to show an example of specifically doing something on Windows, with the prior expectation that stdcall was a windows thing).  Maybe we should just drop the comment?

ChrisDenton: Hm, "system" is what you should always use for the Windows API. Using stdcall would be wrong because it'll break on non-x86 platforms (well, not currently but that's the plan for the future).

I definitely don't want us to be recommending a future compatibility hazard in official docs. I'd be ok with also removing the comment but I do think it's still correct.

ehuss: Oh, ok, seems like I misunderstood. Thanks!

ehuss (approved):

PR #1855, opened 2025-06-12
Mention that `thiscall` is a 32-bit calling convention

ChrisDenton:
[`thiscall`](https://learn.microsoft.com/en-us/cpp/cpp/thiscall?view=msvc-170) is a Windows 32-bit calling convention only and will error on other targets.


ehuss (approved):
Thanks!

IIUC, this validation is done [here](https://github.com/rust-lang/rust/blob/015c7770ec0ffdba9ff03f1861144a827497f8ca/compiler/rustc_target/src/spec/abi_map.rs#L113-L114)

PR #1862, opened 2025-06-17
doc(modules): Remove module style recommendation

ranger-ross:
I was having a conversation about Rust modules with one of my coworkers who is relatively new to Rust.
They mentioned that using `foo/mod.rs` is not recommended in the Rust reference which surprised me.  

Even in modern Rust, I still find the `foo/mod.rs` style much more common than using the "newer" style of adding `foo.rs` and adding a `foo/` directory.

I think its worth removing this recommendation  from the reference to make it unopinionated and avoid confusing newcomers.


ehuss:
We recently revisited this [here](https://github.com/rust-lang/reference/pull/1703#issuecomment-2740833824) and decided to keep the existing text.

ranger-ross:
Ah I see, given this has already been discussed, closing.

PR #1863, opened 2025-06-17
Use commonly used (yet precise) words: parameter and argument.

iampi31415:
actual parameter -> argument
formal parameter -> parameter

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -36,8 +36,8 @@ r[coerce.site.value]
 r[coerce.site.argument]
 * Arguments for function calls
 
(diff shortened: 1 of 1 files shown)


iampi31415:
Notice that [the section says](https://doc.rust-lang.org/nightly/reference/type-coercions.html#r-coerce.site.argument): "Arguments for function calls" and then calls it "actual parameter". 

In my view, one would rather use the same and commonly used term "argument" in both cases.

iampi31415:
Maybe the words were chosen because others found "parameter" and "argument" ambiguous. They aren't for me, but this may not be representative of most readers.

Changed my mind, this may not be that useful.

PR #1865, opened 2025-06-18
const-eval.const-expr.borrows: mention indirect places

RalfJung:
Follow-up to https://github.com/rust-lang/reference/pull/1858.

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -83,9 +83,10 @@ r[const-eval.const-expr.builtin-arith-logic]
 r[const-eval.const-expr.borrows]
 * All forms of [borrow]s, including raw borrows, with one limitation:
   mutable borrows and shared borrows to values with interior mutability
-  are only allowed to refer to *transient* places or to *static* places. A place is *transient*
-  if its lifetime is strictly contained inside the current [const context].
-  A place is *static* if it is a `static` item or a [promoted expression].
+  are only allowed to refer to *transient* places, to *indirect* places, or to *static* places.
+  A place is *transient* if it is based on a local variable whose lifetime is strictly contained inside the current [const context].
+  A place is *indirect* if it is based on a [dereference expression][dereference operator].
+  A place is *static* if it is based on a `static` item or a [promoted expression].
 
 r[const-eval.const-expr.deref]
 * The [dereference operator] except for raw pointers.
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -81,14 +81,126 @@ r[const-eval.const-expr.builtin-arith-logic]
   operators used on integer and floating point types, `bool`, and `char`.
 
 r[const-eval.const-expr.borrows]
-* All forms of [borrow]s, including raw borrows, with one limitation:
-  mutable borrows and shared borrows to values with interior mutability
-  are not allowed to refer to [lifetime-extended temporaries in the top-level scope of a `const` or `static` initializer expression][lifetime-extension-const].
+* All forms of [borrow]s, including raw borrows, except borrows of expressions whose temporary scopes would be extended (see [temporary lifetime extension]) to the end of the program and which are either:
+  * Mutable borrows.
+  * Shared borrows of expressions that result in values with [interior mutability].
 
-  In other words, they are only allowed to refer to *transient* places, to *indirect* places, or to *static* places.
-  A place is *transient* if it is based on a local variable whose lifetime is strictly contained inside the current [const context].
-  A place is *indirect* if it is based on a [dereference expression][dereference operator].
-  A place is *static* if it is based on a `static` item or a [promoted expression].
+  ```rust,compile_fail,E0764
+  // Due to being in tail position, this borrow extends the scope of the
+  // temporary to the end of the program. Since the borrow is mutable,
+  // this is not allowed in a const expression.
+  const C: &u8 = &mut 0; // ERROR not allowed
+  ```
+
+  ```rust,compile_fail,E0764
+  // Const blocks are similar to initializers of `const` items.
+  let _: &u8 = const { &mut 0 }; // ERROR not allowed
+  ```
+
+  ```rust,compile_fail,E0492
+  # use core::sync::atomic::AtomicU8;
+  // This is not allowed as 1) the temporary scope is extended to the
+  // end of the program and 2) the temporary has interior mutability.
+  const C: &AtomicU8 = &AtomicU8::new(0); // ERROR not allowed
+  ```
+
+  ```rust,compile_fail,E0492
+  # use core::sync::atomic::AtomicU8;
+  // As above.
+  let _: &_ = const { &AtomicU8::new(0) }; // ERROR not allowed
+  ```
+
+  ```rust
+  # #![allow(static_mut_refs)]
+  // Even though this borrow is mutable, it's not of a temporary, so
+  // this is allowed.
+  const C: &u8 = unsafe { static mut S: u8 = 0; &mut S }; // OK
+  ```
+
+  ```rust
+  # use core::sync::atomic::AtomicU8;
+  // Even though this borrow is of a value with interior mutability,
+  // it's not of a temporary, so this is allowed.
+  const C: &AtomicU8 = {
+      static S: AtomicU8 = AtomicU8::new(0); &S // OK
+  };
+  ```
+
+  ```rust
+  # use core::sync::atomic::AtomicU8;
+  // This shared borrow of an interior mutable temporary is allowed
+  // because its scope is not extended.
+  const C: () = { _ = &AtomicU8::new(0); }; // OK
+  ```
+
+  ```rust
+  // Even though the borrow is mutable and the temporary lives to the
+  // end of the program due to promotion, this is allowed because the
+  // borrow is not in tail position and so the scope of the temporary
+  // is not extended via temporary lifetime extension.
+  const C: () = { let _: &'static mut [u8] = &mut []; }; // OK
+  //                                              ~~
+  //                                     Promoted temporary.
+  ```
+
+  > [!NOTE]
+  > In other words -- to focus on what's allowed rather than what's not allowed -- shared borrows of interior mutable data and mutable borrows are only allowed in a [const context] when the borrowed [place expression] is *transient*, *indirect*, or *static*.
+  >
+  > A place expression is *transient* if it is a variable local to the current const context or an expression whose temporary scope is contained inside the current const context.
+  >
+  > ```rust
+  > // The borrow is of a variable local to the initializer, therefore
+  > // this place expresssion is transient.
+  > const C: () = { let mut x = 0; _ = &mut x; };
+  > ```
+  >
+  > ```rust
+  > // The borrow is of a temporary whose scope has not been extended,
+  > // therefore this place expression is transient.
+  > const C: () = { _ = &mut 0u8; };
+  > ```
+  >
+  > ```rust
+  > // When a temporary is promoted but not lifetime extended, its
+  > // place expression is still treated as transient.
+  > const C: () = { let _: &'static mut [u8] = &mut []; };
+  > ```
+  >
+  > A place expression is *indirect* if it is a [dereference expression].
+  >
+  > ```rust
+  > const C: () = { _ = &mut *(&mut 0); };
+  > ```
+  >
+  > A place expression is *static* if it is a `static` item.
+  >
+  > ```rust
+  > # #![allow(static_mut_refs)]
+  > const C: &u8 = unsafe { static mut S: u8 = 0; &mut S };
+  > ```
+
+  > [!NOTE]
+  > One surprising consequence of these rules is that we allow this,
+  >
+  > ```rust
+  > const C: &[u8] = { let x: &mut [u8] = &mut []; x }; // OK
+  > //                                    ~~~~~~~
+  > // Empty arrays are promoted even behind mutable borrows.
+  > ```
+  >
+  > but we disallow this similar code:
+  >
+  > ```rust,compile_fail,E0764
+  > const C: &[u8] = &mut []; // ERROR
+  > //               ~~~~~~~
+  > //           Tail expression.
+  > ```
+  >
+  > The difference between these is that, in the first, the empty array is [promoted] but its scope does not undergo [temporary lifetime extension], so we consider the [place expression] to be transient (even though after promotion the place indeed lives to the end of the program). In the second, the scope of the empty array temporary does undergo lifetime extension, and so it is rejected due to being a mutable borrow of a lifetime-extended temporary (and therefore borrowing a non-transient place expression).
+  >
+  > The effect is surprising because temporary lifetime extension, in this case, causes less code to compile than would without it.
+  >
+  > See [issue #143129](https://github.com/rust-lang/rust/issues/143129) for more details.
 
 r[const-eval.const-expr.deref]
 * The [dereference operator] except for raw pointers.
@@ -178,6 +290,7 @@ of whether you are building on a `64` bit or a `32` bit system.
 [const generic parameters]: items/generics.md#const-generics
 [constants]:            items/constant-items.md
 [Const parameters]:     items/generics.md
+[dereference expression]: expressions/operator-expr.md#the-dereference-operator
 [dereference operator]: expressions/operator-expr.md#the-dereference-operator
 [destructors]:          destructors.md
 [enum discriminants]:   items/enumerations.md#discriminants
@@ -190,7 +303,6 @@ of whether you are building on a `64` bit or a `32` bit system.
 [interior mutability]:  interior-mutability.md
 [if]:                   expressions/if-expr.md#if-expressions
 [lazy boolean]:         expressions/operator-expr.md#lazy-boolean-operators
-[lifetime-extension-const]: destructors.md#r-destructors.scope.lifetime-extension.static
 [let statements]:       statements.md#let-statements
 [literals]:             expressions/literal-expr.md
 [logical]:              expressions/operator-expr.md#arithmetic-and-logical-binary-operators
@@ -200,10 +312,13 @@ of whether you are building on a `64` bit or a `32` bit system.
 [overflow]:             expressions/operator-expr.md#overflow
 [paths]:                expressions/path-expr.md
 [patterns]:             patterns.md
+[place expression]:     expr.place-value.place-memory-location
 [promoted expression]:  destructors.md#constant-promotion
+[promoted]:             destructors.md#constant-promotion
 [range expressions]:    expressions/range-expr.md
 [slice]:                types/slice.md
 [statics]:              items/static-items.md
 [struct]:               expressions/struct-expr.md
+[temporary lifetime extension]: destructors.scope.lifetime-extension
 [tuple expressions]:    expressions/tuple-expr.md
 [while]:                expressions/loop-expr.md#predicate-loops
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -83,7 +83,9 @@ r[const-eval.const-expr.builtin-arith-logic]
 r[const-eval.const-expr.borrows]
 * All forms of [borrow]s, including raw borrows, with one limitation:
   mutable borrows and shared borrows to values with interior mutability
-  are only allowed to refer to *transient* places, to *indirect* places, or to *static* places.
+  are not allowed to refer to [lifetime-extended temporaries in the top-level scope of a `const` or `static` initializer expression][lifetime-extension-const].
+
+  In other words, they are only allowed to refer to *transient* places, to *indirect* places, or to *static* places.
   A place is *transient* if it is based on a local variable whose lifetime is strictly contained inside the current [const context].
   A place is *indirect* if it is based on a [dereference expression][dereference operator].
   A place is *static* if it is based on a `static` item or a [promoted expression].
@@ -188,6 +190,7 @@ of whether you are building on a `64` bit or a `32` bit system.
 [interior mutability]:  interior-mutability.md
 [if]:                   expressions/if-expr.md#if-expressions
 [lazy boolean]:         expressions/operator-expr.md#lazy-boolean-operators
+[lifetime-extension-const]: destructors.md#r-destructors.scope.lifetime-extension.static
 [let statements]:       statements.md#let-statements
 [literals]:             expressions/literal-expr.md
 [logical]:              expressions/operator-expr.md#arithmetic-and-logical-binary-operators


RalfJung:
Based on the last commit I added in https://github.com/rust-lang/rust/pull/140942, instead of a list of things we allow, we could consider just saying what is *not* allowed -- mutable / interior mutable borrows of lifetime-extended places. Maybe that would be better?

traviscross:
Thanks @RalfJung; @ehuss and talked about this new language.  On first pass, it looks pretty good to us.  To your question about stating this in terms of what's not allowed, that also had some appeal to us, and we actually think it'd be best in this case to do both, i.e. to state what's allowed, as is done here, and to then also approach the same truth from the other side by stating what's not allowed.  We do that in a number of places, and this seems like a good candidate for it, in terms of being as clear as possible.

RalfJung:
I've given that a shot, please let me know what you think.

I also realized `destructors.scope.lifetime-extension.static` fails to mention that it applies specifically to the *top-level scope* of const/static initializers.

traviscross:
Thanks.  Probably we'll want to add some examples here for each.

Regarding the "top-level scope" bit, I wonder if there's a more clear way we could say that.  We allow, of course, e.g.:

```rust
struct S;
const fn temp() -> S { S }
const C1: &S = { let y = { let x = &temp(); x }; y };
const C2: &S = 'top: { let y = { let x = &temp(); break 'top x; }; };
```

That is, there's a distinction between the temporary being created in the top-level scope and being referenced by the final value.

RalfJung:
All of those are immutable references so yes of course we allow them. That doesn't really exercise the clauses this PR is about, though.

There are no examples for any of the surrounding text here so I'll not try to figure out the best way to integrate examples into this part of the docs.

traviscross:
> All of those are immutable references so yes of course we allow them. That doesn't really exercise the clauses this PR is about, though.

The question is about this change specifically, which is independent of mutability:

```diff
 r[destructors.scope.lifetime-extension.static]
-Lifetime extension also applies to `static` and `const` items, where it makes temporaries live until the end of the program. 
+Lifetime extension also applies to the top-level scope of `static` and `const` items, where it makes temporaries live until the end of the program.
```

It seems that the rule for this should be essentially the same as the rule that precedes it, except that we're extending to `'static` rather than to the scope of the containing block:

> r[destructors.scope.lifetime-extension.let]
> The temporary scopes for expressions in `let` statements are sometimes *extended* to the scope of the block containing the `let` statement. This is done when the usual temporary scope would be too small, based on certain syntactic rules.

(Admittedly, that rule is rather ambiguous itself.)

So if this "top-level scope" qualification makes sense for one, then it would seem to make sense for the other.

But I think it gets into some ambiguity about "are we talking about the temporary being created in the top-level scope or about the temporary being referenced by the value returned from the top-level scope?".  Clearly we don't mean the former.

RalfJung:
We mean "the temporary is extended to the scope *outside* the const initializer expression". Not sure what is the best way to put that...

traviscross:
> There are no examples for any of the surrounding text here so I'll not try to figure out the best way to integrate examples into this part of the docs.

We'll take care of integrating these, but @ehuss and I would be interested if you could provide here, just as a reply, differentiating examples (i.e. examples that fit in exactly one category) for each of the things we're defining here, i.e. transient places, indirect places, places based on static items, and places based on promoted expressions.

traviscross:
> We mean "the temporary is extended to the scope _outside_ the const initializer expression". Not sure what is the best way to put that...

Thanks.  I pushed a clarification to define this in terms of the other rule that touches on this.

src/destructors.md
```diff
@@ -336,9 +336,7 @@ println!("{}", x);
 ```
 
 r[destructors.scope.lifetime-extension.static]
-Lifetime extension also applies to the top-level scope of `static` and `const` items,
-where it makes temporaries live until the end of the program.
-For example:
+The temporary scopes for expressions in `const` and `static` item initializers are sometimes extended until the end of the program following the same rule as in [destructors.scope.lifetime-extension.let]. For example:
```
RalfJung: "sometimes" isn't very helpful, is it? Can't we say something more precise, such as what you say in the commit message?

> In `const` and `static` item initializers, if the scope of a temporary would be extended to a scope outside that of the initializer expression, then the temporary lives until the end of the program. 

And then maybe we want to add something like this somewhere:

> This effectively makes such a temporary a new global variable. This global variable is always immutable; the limitations in [const-eval.const-expr.borrows] rule out constants where a mutable temporary would have to live until the end of the program.

RalfJung:
> differentiating examples (i.e. examples that fit in exactly one category) for each of the things we're defining here, i.e. transient places, indirect places, places based on static items, and places based on promoted expressions.
```rust
static mut S: i32= 0;

#[allow(static_mut_refs)]
const C: () = {
  let mut x = 0;
  let mref = &mut x; // reference to transient place
  let mref2 = &mut *mref; // reference to indirect place
  let sref = unsafe { &mut S }; // reference to static place
  let pref: &mut [i32] = &mut []; // reference to promoted
};
```

src/destructors.md
traviscross: In reading the rest of the section more carefully, I actually feel like we should just defer this to another PR.  The existing language has to be read in the context of everything that follows it -- most of which is talking about `let` statements even though it also applies to `const` and `static` items -- and in that context, all this is trying to say, when it says,

> Lifetime extension also applies to `static` and `const` items, where it makes temporaries live until the end of the program.

is that the rules that precede and follow this rule, even though they're referring to `let` statements, "also" apply to `static` and `const` items, modulo that in this case, rather than extending to the scope of the enclosing block, "it makes temporaries live until the end of the program."

Probably there's a way to improve this section overall, but I'm now convinced doing that coherently requires touching many of the rules here.

(I pulled the change to this rule out of the commit and dropped my commit.)

RalfJung: https://github.com/rust-lang/rust/pull/140942 now already introduced the "top-level" wording into compiler errors. That should then probably be adjusted if that is not the term we use in the reference. Any suggestion for what the error should say?

traviscross:
Thanks @RalfJung for the examples.  I've pushed commits to integrate those and some others and to hopefully clarify some language.  Let me know if that all looks right.

traviscross:
In testing, I notice this:

```rust
---- const_eval.md - Constant_evaluation::Constant_expressions (line 212) stdout ----
error[E0764]: mutable borrows of lifetime-extended temporaries in the top-level scope of a constant are not allowed
 --> const_eval.md:213:18
  |
3 | const C: &[u8] = &mut [];
  |                  ^^^^^^^

error: aborting due to 1 previous error
```

This borrow isn't of an interior mutable temporary.  Does this imply a second or a broader exception, you think, or is this covered by something else?

RalfJung:
It's mutable, but not interior mutable.

Though what I find odd is that `&mut []` should be promoted, and then I'd expect this to actually compile...

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
   operators used on integer and floating point types, `bool`, and `char`.
 
 r[const-eval.const-expr.borrows]
-* All forms of [borrow]s, including raw borrows, with one limitation:
-  mutable borrows and shared borrows to values with interior mutability
-  are only allowed to refer to *transient* places or to *static* places. A place is *transient*
-  if its lifetime is strictly contained inside the current [const context].
-  A place is *static* if it is a `static` item or a [promoted expression].
+* All forms of [borrow]s, including raw borrows, with one limitation: mutable borrows and shared borrows of expressions whose temporary scope would be extended (see [temporary lifetime extension]) to the end of the program are not allowed when those borrows refer to values with interior mutability.
```
RalfJung: This makes it sound like the "when those borrows refer to values with interior mutability" part applies to everything, which isn't correct, it only applies to "shared borrows". Basically, borrows with any form of mutability are disallowed to refer to such temporaries, and we have two kinds of mutable borrows: actually mutable borrows, and shared borrows with interior mutability.

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+
+  ```rust,compile_fail,E0764
+  # use core::sync::atomic::AtomicU8;
+  const C: &AtomicU8 = &mut AtomicU8::new(0); // ERROR not allowed
+  ```
+
+  ```rust,compile_fail,E0492
+  # use core::sync::atomic::AtomicU8;
+  let _: &'static _ = const { &AtomicU8::new(0) }; // ERROR not allowed
+  ```
```
RalfJung, line 105: I can't figure out which of these examples is meant to demonstrate what.

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+      let _ = unsafe { &mut S }; // Borrow of static item.
+  };
+  ```
+
+  ```rust
+  const C: &[u8] = &mut []; // Borrow of promoted expression.
+  ```
+
+  ```rust
+  const C: &[u8] = &[]; // Borrow of promoted expression.
```
RalfJung: This isn't even in the class of "borrows with some form of mutability", so I don't understand how it makes sense as an example here.

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+* All forms of [borrow]s, including raw borrows, with one limitation: mutable borrows and shared borrows of expressions whose temporary scope would be extended (see [temporary lifetime extension]) to the end of the program are not allowed when those borrows refer to values with interior mutability.
+
+  ```rust,compile_fail,E0492
+  # use core::sync::atomic::AtomicU8;
+  const C: &AtomicU8 = &AtomicU8::new(0); // ERROR not allowed
+  ```
+
+  ```rust,compile_fail,E0764
+  # use core::sync::atomic::AtomicU8;
+  const C: &AtomicU8 = &mut AtomicU8::new(0); // ERROR not allowed
```
RalfJung: Mixing `&mut` and interior mutability just causes confusion about which source of mutability actually matters here, I think we should avoid it in these examples.

RalfJung:
So, uh, looks like we are doing the static const checks *before* we do promotion? @oli-obk is that expected?

That means we should remove promotion from the text and examples in this PR.

oli-obk:
static const checks happen before promotion intentionally. I think they should even be a part of typeck, and I think we should be able to move more of it into typeck

RalfJung:
Okay, looks like I just entirely forgot about that then.

That's good, discussing promoteds here got a bit awkward IMO...

src/const_eval.md
traviscross: Yes, it's there because `const C: &[u8] = &mut []`, which I had expected to work, failed.  The language we're demonstrating is:

> The following expressions are constant expressions... All forms of [borrow]s, including raw borrows, with one limitation: mutable borrows and shared borrows of expressions whose temporary scope would be extended to the end of the program... In other words, they are only allowed to refer to ... or to *static* places ... A place is *static* if it is based on a `static` item or a [promoted expression].

Most of the constraints we're navigating here, especially in the framing of what is allowed (rather than rejected), do not depend on mutability.

So what this (and the original) example was meant to demonstrate is that 1) you can borrow a promoted expression and 2) that borrow may appear in tail position and undergo lifetime extension as long as 3) the temporary does not have interior mutability.

src/const_eval.md
traviscross: Perhaps, though it's directly modeling the language in the text:

> with one limitation: mutable borrows and shared borrows of expressions whose temporary scope would be extended to the end of the program are not allowed when those borrows refer to values with interior mutability.

src/const_eval.md
traviscross: Fair enough.  I'll add some comments.

src/const_eval.md
traviscross: Ah, I see in the comment over in https://github.com/rust-lang/reference/pull/1865#discussion_r2172028165 this does touch on an ambiguity in the text.

src/const_eval.md
traviscross: Interesting.  Prior to my commit, the language was:

> All forms of [borrow]s, including raw borrows, with one limitation: mutable borrows and shared borrows to values with interior mutability are not allowed to refer to lifetime-extended temporaries in the top-level scope of a `const` or `static` initializer expression.

I had read this same ambiguity into this verbiage.  Thinking about it now, of course, this doesn't make any sense, as mutable references to `UnsafeCell` aren't special.  I'll clarify the verbiage.

src/const_eval.md
traviscross: This needs to be reworked due to https://github.com/rust-lang/reference/pull/1865#discussion_r2172740786 as well.

src/const_eval.md
RalfJung: "mutable borrows and shared borrows to values with interior mutability" was intended to parse as
- mutable borrows, and
- shared borrows to values with interior mutability

English makes it hard to express this unambiguously...

src/destructors.md
traviscross: I'm thinking about this.   My first inclination is to say something to the effect of:

> mutable borrow of a temporary whose lifetime would be extended to `'static` is not allowed in a `const` initializer

and:

> shared borrow of an interior mutable temporary whose lifetime would be extended to `'static` is not allowed in a `const` initializer

(This is maybe a bit of a pun, because strictly speaking references have their lifetimes extended while temporaries have their scopes extended, and as phrased, it's not clear whether we're talking about the reference (due to the borrow) or the temporary, but it doesn't matter here, so it seems OK to me, and it seems probably more familiar here to talk about lifetime extension to `'static` rather than about scope extension to the end of the program.)

RalfJung: That's very close to what I did in https://github.com/rust-lang/rust/pull/143092, except I do refer to scopes there.
(Sorry, I should have mentioned here that I made a PR.)

traviscross:
It seems a rather subtle point to document, given that:

```rust
const C1: &[u8] = { let x: &'static mut [u8] = &mut []; x }; //~ OK
const C2: &[u8] = { &mut [] }; //~ ERROR
```

src/destructors.md
traviscross: Ah, yes, that looks good too.

traviscross: > ...except I do refer to scopes there.

(Looking at the diff, it actually went with "lifetime extended until the end of the program".)

RalfJung:
Yeah, that is kind of odd -- the fact that lifetime extension kicks in here means that code that would otherwise have compiled (if there was no lifetime extension) now does not compile.

src/destructors.md
RalfJung: Ah, right. So a mix of "scope extended" and "lifetime becomes `'static`", I guess.

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+  ```
+
+  In other words, they are only allowed to refer to *transient* places, to *indirect* places, or to *static* places. A place is *transient* if it is based on a local variable whose lifetime is strictly contained inside the current [const context].
+
+  ```rust
+  const C: () = {
+      let mut x = 0;
+      let _ = &mut x; // Reference to a transient place.
+  };
+  ```
```
traviscross: When we say, @RalfJung, "local variable" above, are we meaning to exclude non-promoted non-lifetime-extended temporaries from the definition of transient?  Intuitively, those seem transient.

src/const_eval.md
```diff
@@ -81,11 +81,81 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+
+  A place is *indirect* if it is based on a [dereference expression][dereference operator].
+
+  ```rust
+  const C: () = {
+      let x = &mut 0;
+      let _ = &mut *x; // Reference to an indirect place.
+  };
+  ```
```
traviscross: If we do, @RalfJung, include non-promoted non-lifetime extended temporaries in the definition of transient, then is there any remaining case that's covered by "indirect" that's not already covered by the place being either transient or static?

Describing the place as indirect here seems maybe a bit odd, as that seems more a property of the borrow than of the place (unlike the other two categories).

Curious what you think there.  I'll handle updating the text.

RalfJung: Everywhere that the text says "place", it should really be "place expression". This is describing a static analysis -- we don't actually know where `x` points to in this example, we just syntactically analyze the `*x` place expression that is the operand of this borrow. And since it contains a `*`, it is indirect.

src/const_eval.md
RalfJung: Hm... there's some friction here since the analysis runs on MIR, where there's no distinction between explicitly `let`-bound variables and temporaries. In MIR, literally what we do is check that the place expression is of the form `$local.$projections` where the projections do not contain a `*`, and then we check that `local` is a local variable that has a `StorageDead` which indicates that it has not been lifetime-extended. (If the projections do contain a `*`, then this is an easy case, we can just allow it as this is an indirect place.)

In terms of the surface language, this means indeed non-lifetime-extended temporaries are transient. Promotion is irrelevant, as we recently learned.

traviscross: Thanks.  That makes sense.  On this bit:

> Promotion is irrelevant, as we recently learned.

For this example:

```rust
const C: () = {
    let _: &'static mut [u8] = &mut []; // OK
};
```

How does the current impl think about this, as transient or as static?  What's the story with the `StorageDead` check here?

traviscross: Looking at https://github.com/rust-lang/rust/issues/143129, I gather this must indeed be transient, as you say for `const C2: &[u8] = &mut []` that :

> If we didn't do lifetime extension here, then the code would work, since then const-checks would consider the borrow to be transient, and finally promotion would make the rest work.

Still, I'm curious how that interacts with that `StorageDead` check you mention, as presumably that wouldn't be there due to the promotion.

RalfJung: > How does the current impl think about this, as transient or as static?

It thinks of this as transient. It doesn't see the lifetimes, and the MIR at this point looks like
```rust
const C: () = {
    let mut arr = [];
    let _: &mut [u8] = &mut arr;
};
```
> Still, I'm curious how that interacts with that StorageDead check you mention, as presumably that wouldn't be there due to the promotion.

It is there. Promotion later removes it.

traviscross: Ah, removed later.  Got it, thanks.

RalfJung: For some context: MIR building knows nothing about promotion. It compiles that code like it would compile `let x: &mut [u8] = &mut [13];`. Promotion is a MIR pass that runs later (between const check and borrow checking, apparently) that does some heavy patching of the MIR to remove a local and its initializing expression and replace it by references to a new anonymous const.

traviscross:
I may yet make some further editorial revisions, but @RalfJung, how does this look?

src/const_eval.md
traviscross: Makes sense.  I've changed it to say "place expression" throughout.

src/const_eval.md
```diff
@@ -81,11 +81,127 @@ r[const-eval.const-expr.builtin-arith-logic]
...
 r[const-eval.const-expr.borrows]
-* All forms of [borrow]s, including raw borrows, with one limitation:
-  mutable borrows and shared borrows to values with interior mutability
-  are only allowed to refer to *transient* places or to *static* places. A place is *transient*
-  if its lifetime is strictly contained inside the current [const context].
-  A place is *static* if it is a `static` item or a [promoted expression].
+* All forms of [borrow]s, including raw borrows, except:
+
+  * Mutable borrows of expressions whose temporary scopes would be extended (see [temporary lifetime extension]) to the end of the program.
+  * Shared borrows of expressions whose temporary scopes would be extended to the end of the program when those expressions result in values with interior mutability.
```
RalfJung: Personally I'd prefer to avoid the redundancy here and spell this out a bit differently. For instance, we could say something like

* All forms of [borrow]s, including raw borrows, except for those that satisfy all of the following:
  * The borrows is either a mutable borrows, or it is a shared borrow and the borrowed expression results in a value with interior mutability.
  * The borrowed expression is a temporary that has its scope extended to the end of the program (see [destructors.scope.lifetime-extension.static]).

But this is just editorial so :shrug:

src/const_eval.md
```diff
@@ -81,11 +81,127 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+
+  ```rust,compile_fail,E0764
+  // Due to being in tail position, this borrow extends the scope of the
+  // temporary to the end of the program. Since the borrow is mutable,
+  // this is not allowed in a const expression.
+  const C: &[u8] = &mut []; // ERROR not allowed
```
RalfJung: I would prefer to start with an example here that does not *also* rely on the weird promotion quirk we discovered, e.g.
```suggestion
  const C: &[u8] = &mut [0, 1, 2]; // ERROR not allowed
```

src/const_eval.md
```diff
@@ -81,11 +81,127 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+  ```
+
+  ```rust,compile_fail,E0764
+  // Const blocks are similar to initializers of `const` items.
+  let _: &[u8] = const { &mut [] }; // ERROR not allowed
```
RalfJung:
```suggestion
  let _: &[u8] = const { &mut [0, 1, 2] }; // ERROR not allowed
```

src/const_eval.md
```diff
@@ -81,11 +81,127 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+  const C: &u8 = unsafe { static mut S: u8 = 0; &mut S }; // OK
+  ```
+
+  ```rust
+  # use core::sync::atomic::AtomicU8;
+  // Even though this borrow is of a value with interior mutability,
+  // it's not of a temporary, so this is allowed.
+  const C: &AtomicU8 = {
+      static S: AtomicU8 = AtomicU8::new(0); &S // OK
+  };
```
RalfJung, line 126: I would put these two above the "promoted temporary" example -- easier examples should come before more complicated ones.

src/const_eval.md
```diff
@@ -81,11 +81,127 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+  ```
+
+  > [!NOTE]
+  > In other words, shared borrows of interior mutable data and mutable borrows are only allowed in a const context when the borrowed place expression is *transient*, *indirect*, or *static*.
+  >
+  > A place expression is *transient* if it is a variable local to the current [const context] or an expression whose temporary scope is contained inside the current const context.
```
RalfJung: Is this note really worth it?

src/const_eval.md
traviscross: Pros and cons.  Will think on this.

src/const_eval.md
traviscross: Makes sense.  Done.

src/const_eval.md
traviscross: Makes sense.  Done.

src/const_eval.md
traviscross: I don't know. I'll try it on someone who hasn't read this yet.

traviscross: On the one hand, approaching it from this direction felt more necessary before adding the examples in the main section.  With that explication, this doesn't carry so much weight.

On the other hand, one purpose of the Reference is to document common terms for things, and writing down and naming the three kinds of place expressions does seem useful.  It could be that this is in the wrong place in the document and it'd be better elsewhere.

RalfJung: > one purpose of the Reference is to document **common** terms for things,

(emphasis mine)

The thing I am wondering is if these are truly common terms. If they were, they'd be defined elsewhere and then just references here. Right now they look more like ad-hoc artifacts of const-check than common terms.

traviscross: Interesting... stepped on a homonym there.  By "common", I meant "shared" rather than "frequently observed" (i.e. in the sense of "tragedy of the commons").  In the sense I meant it, putting it in the Reference (and people's use of the Reference for this purpose) makes it common.

src/const_eval.md
traviscross: Came up with something here that achieves this with good readability.

src/const_eval.md
traviscross: We're going to keep this for now.  The other note references the terminology introduced in this note, and I think it's a bit easier to explain why that odd case does what it does using this framing rather than the other framing.  But it's a bit of a coin flip.

traviscross (approved):

traviscross:
Thanks @RalfJung for the follow-up here.

src/const_eval.md
```diff
@@ -81,11 +81,126 @@ r[const-eval.const-expr.builtin-arith-logic]
...
+* All forms of [borrow]s, including raw borrows, except borrows of expressions whose temporary scopes would be extended (see [temporary lifetime extension]) to the end of the program and which are either:
+  * Mutable borrows.
+  * Shared borrows of expressions that result in values with [interior mutability].
+
+  ```rust,compile_fail,E0764
+  // Due to being in tail position, this borrow extends the scope of the
+  // temporary to the end of the program. Since the borrow is mutable,
+  // this is not allowed in a const expression.
+  const C: &u8 = &mut 0; // ERROR not allowed
+  ```
```
RalfJung, line 93: Should all these here be marked as "NOTE" or "EXAMPLE" or so?

traviscross: No, when we put an example under a rule that's directly demonstrating what that rule means we just put it inline.  It's a bit nuanced.  It's a distinction we've been been working out recently.

PR #1871, opened 2025-06-26
fix: swap places for 2 words in associated-items.md sentence.

iampi31415:
The sentence is still a mouthful. But this fixes a grammatical mistake.

Here is the nightly version of the doc https://doc.rust-lang.org/nightly/reference/items/associated-items.html#r-items.associated.related


ehuss (approved):
Thanks!

PR #1873, opened 2025-06-28
introduction.md: replace hard-to-read example

tshepang:


ehuss:
This was intentionally written that way to emphasize that the block should document the *previous* edition's behavior. It certainly could be reworded to make that clearer, but I would prefer to keep the intent.

tshepang:
how about
> This syntax is disallowed as of the 2018 edition.

traviscross (approved):

PR #1875, opened 2025-06-30
fix (expr): add never-type note to expressions.

iampi31415:
Not all expressions evaluate to a value. Examples are `return`, `loop`, `break`. This is described for infinite `loop`s but not as a statement applicable to a subset of the expressions.

After some consideration and reading the `authoring.md` a "> [!Note]" was added instead of a standalone paragraph. It becomes more of a detail for careful readers.

The first two paragraphs are nearly identical to the original document. The only word removed from the initial paragraph is _always_ because that's not true, and would trip some users up.

It links to [the reference `!`](https://doc.rust-lang.org/reference/types/never.html) but another useful link could be [std !](https://doc.rust-lang.org/std/primitive.never.html). That would make both the text longer, and link to a page that starts with "this is unstable". So the latter was left out.

CC: @ehuss @RalfJung just in case you can review or simply dismiss this.

### DIFF
--- a/src/expressions.md
+++ b/src/expressions.md
@@ -46,10 +46,13 @@ ExpressionWithBlock ->
 ```
 
 r[expr.intro]
-An expression may have two roles: it always produces a *value*, and it may have *effects* (otherwise known as "side effects").
+An expression may have two roles: it produces a *value*, and it may have *effects* (otherwise known as "side effects").
 
 r[expr.evaluation]
 An expression *evaluates to* a value, and has effects during *evaluation*.
+
+> [!NOTE]
+> Some expressions _do not_ produce a value; however, they are exceptions rather than the rule. Examples that do not produce a value are: [`return`][expr.return], [`continue`][expr.loop.continue], [`break`][expr.loop.break] and [`loop`][expr.loop.infinite]. They do have a type, the [never type][Never Type] denoted `!`.
 
 r[expr.operands]
 Many expressions contain sub-expressions, called the *operands* of the expression.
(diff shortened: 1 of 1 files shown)


iampi31415:
looks like

iampi31415:
I don't think improvements are desired, so I will simply close this. If anyone wants just use the PR, I won't contribute anymore.

RalfJung:
Not sure what you are concluding that from, but you definitely cannot expect a review in 48h on any open source project.

I don't have a strong opinion on this PR either way.

iampi31415:
I've done many others. In the book, in the reference, in rust. They all take weeks. This one was closed because of that trend, not because I expect a review in 48h.

RalfJung:
The book has a note in the README setting expectations:

> Because the book is [printed](https://nostarch.com/rust-programming-language-2nd-edition), and because we want to keep the online version of the book close to the print version when possible, it may take longer than you're used to for us to address your issue or pull request.
> 
> So far, we've been doing a larger revision to coincide with [Rust Editions](https://doc.rust-lang.org/edition-guide/). Between those larger revisions, we will only be correcting errors. If your issue or pull request isn't strictly fixing an error, it might sit until the next time that we're working on a large revision: expect on the order of months or years. Thank you for your patience!

I acknowledge this is unusual, but it's written out explicitly to minimize surprises.

In terms of rustc PRs, https://github.com/rust-lang/rust/pull/143098 didn't even take 5 days until you closed it. You got a message setting expectations when opening that PR:
>  They will have a look at your PR within the next two weeks and either review your PR or reassign to another reviewer.

If you expect reviews within a few days, then yeah Rust is not the right project for you to contribute to. Thank you for trying anyway! Unfortunately, we have too many contributors and too few reviewers for such fast turn-around times. Sorry to hear that our volunteers did not do their free labor as fast as you expected them to. I hope you'll find other projects that can live up to your expectations.

iampi31415:
I do think it's not, indeed. I'll look for other projects to exploit volunteers, thanks. Note, that it will save my free and unpaid time as well, as little as it matters.

PR #1886, opened 2025-06-30
Update `macro_use` to use the attribute template

ehuss:
New rules:
- ❗ `macro.decl.scope.macro_use.intro`
- ❗ `macro.decl.scope.macro_use.syntax`
- ❗ `macro.decl.scope.macro_use.allowed-positions`
- ❗ `macro.decl.scope.macro_use.duplicates`

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -343,6 +343,9 @@ The `macro_use` attribute may be applied to modules or `extern crate`.
 > [!NOTE]
 > `rustc` currently warns in other positions, but this may be rejected in the future.
 
+r[macro.decl.scope.macro_use.extern-crate-self]
+The `macro_use` attribute may not be used on [`extern crate self`].
+
 r[macro.decl.scope.macro_use.duplicates]
 Duplicate instances of `macro_use` that are in the [MetaWord] syntax have no effect if there is already a `macro_use` with the [MetaWord] syntax.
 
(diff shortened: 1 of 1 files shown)


traviscross (approved):

src/macros-by-example.md
```diff
@@ -326,47 +326,86 @@ fn foo() {
...
-conflict, the last macro imported wins. Optionally, a list of macros to import
-can be specified using the [MetaListIdents] syntax; this is not supported
-when `#[macro_use]` is applied to a module.
+Specifying `macro_use` on an `extern crate` declaration in the crate root imports exported macros from that crate.
 
-<!-- ignore: requires external crates -->
-```rust,ignore
-#[macro_use(lazy_static)] // Or #[macro_use] to import all macros.
-extern crate lazy_static;
+Macros imported this way are imported into the [`macro_use` prelude], not textually, which means that they can be shadowed by any other name. While macros imported by `macro_use` can be used before the import statement, in case of a conflict, the last macro imported wins.
```
petrochenkov, line 393: It's unfortunate that we document that the last macro wins.

I always wanted to rework this into the standard order-independent scheme with "glob imports" (`#[macro_use]`), "single imports" shadowing glob imports (`#[macro_use(single)]`), and glob-vs-glob conflicts reported on use.
We should be able to do in practice, I think, so it would be good to leave unspecified what happens in case of conflicts.

traviscross: Is there a good issue to cite for this?

traviscross: See:

- https://github.com/rust-lang/reference/pull/2057

PR #1890, opened 2025-06-30
Update diagnostics to use the attribute template

ehuss:
New rules:
- ❗ `attributes.diagnostics.lint.intro`
- ❗ `attributes.diagnostics.lint.syntax`
- ❗ `attributes.diagnostics.lint.allowed-positions`
- ❗ `attributes.diagnostics.lint.duplicates`

Renamed rules:
- `attributes.diagnostics.expect.intro` is now `attributes.diagnostics.expect.behavior`


ehuss:
@traviscross I think when we were talking you said that you had some edits here but I don't think they got pushed up. Were they complete?

PR #1908, opened 2025-06-30
Update `naked` to use the attribute template

ehuss:
New rules:
- ❗ `attributes.codegen.naked.syntax`
- ❗ `attributes.codegen.naked.allowed-positions`
- ❗ `attributes.codegen.naked.duplicates`
- ❗ `attributes.codegen.naked.target_feature`
- ❗ `attributes.codegen.naked.abi`

Renamed rules:
- `attributes.codegen.naked.unsafe-attribute` is now `attributes.codegen.naked.unsafe`

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -154,7 +154,7 @@ The `naked` attribute may only be applied to:
 - [Free functions][items.fn]
 - [Inherent associated functions][items.associated.fn]
 - [Trait impl functions][items.impl.trait]
-- [Trait definition functions][items.traits] with a body
+- [Trait definition functions][items.traits.associated-item-decls] with a body
 
 r[attributes.codegen.naked.duplicates]
 The `naked` attribute may be used any number of times on a form.
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -119,6 +119,7 @@ Only the first use of `cold` on a function has effect.
 r[attributes.codegen.cold.trait]
 When `cold` is applied to a function in a [trait], it applies only to the code of the [default definition].
 
+<!-- template:attributes -->
 r[attributes.codegen.naked]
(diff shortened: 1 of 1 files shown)


src/attributes/codegen.md
```diff
@@ -52,15 +52,51 @@ r[attributes.codegen.naked]
...
+r[attributes.codegen.naked.syntax]
+The `naked` attribute uses the [MetaWord] syntax and thus does not take any inputs.
+
+r[attributes.codegen.naked.allowed-positions]
+The `naked` attribute may only be applied to:
+
+- [Free functions][items.fn]
+- [Inherent associated functions][items.associated.fn]
+- [Trait impl functions][items.impl.trait]
+- [Trait definition functions][items.traits] with a body
```
tshepang: does this work
```suggestion
- [Trait][items.traits] [associated functions][items.associated.fn] with a body
```

tshepang:
just 1 comment, otherwise this is good

src/attributes/codegen.md
ehuss: This wording can be ambiguous, since it is unclear if it is an associated function in a trait definition or a trait impl. The original wording was specifically chosen to avoid that ambiguity and contrast with the previous item.

tshepang: initially wanted to link this to `items.traits.associated-item-decls`, due to being more specific, because `items.traits` was too general

> - [Trait definition functions][items.traits.associated-item-decls] with a body

is that alright

ehuss:
TODO: The error "patterns not allowed in naked function parameters" does not seem to be documented.

traviscross:
Thanks @ehuss.

cc https://github.com/rust-lang/rust/pull/157201

traviscross (approved):

PR #1915, opened 2025-06-30
Update `path` to use the attribute template

ehuss:
New rules:
- ❗ `items.mod.outlined.path.syntax`
- ❗ `items.mod.outlined.path.allowed-positions`
- ❗ `items.mod.outlined.path.duplicates`


DanielEScherzer (approved):
look good to me but I don't have write access so this isn't a real approval

PR #1921, opened 2025-06-30
Update `non_exhaustive` to use the attribute template

ehuss:
New rules:
- ❗ `attributes.type-system.non_exhaustive.duplicates`
- ❗ `attributes.type-system.non_exhaustive.enum-exhaustiveness`
- ❗ `attributes.type-system.non_exhaustive.enum-cast`
- ❗ `attributes.type-system.non_exhaustive.inhabited`

Renamed rules:
- `attributes.type-system.non_exhaustive.external-crate` moved to `.intro` because it is generally explaining the purpose.

### DIFF
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -7,7 +7,7 @@ r[attributes.type-system.non_exhaustive]
 ## The `non_exhaustive` attribute
 
 r[attributes.type-system.non_exhaustive.intro]
-The *`non_exhaustive` attribute* indicates that a type or variant may have more fields or variants added in the future. Outside of the defining crate, types annotated with `non_exhaustive` have limitations that preserve backwards compatibility when new fields or variants are added.
+The *`non_exhaustive` [attribute][attributes]* indicates that a type or variant may have more fields or variants added in the future. Outside of the defining crate, types annotated with `non_exhaustive` have limitations that preserve backwards compatibility when new fields or variants are added.
 
 > [!EXAMPLE]
 > The following non-exhaustive definitions will be used in the examples that follow.
--- a/src/attributes/type_system.md
+++ b/src/attributes/type_system.md
@@ -47,6 +47,12 @@ The `non_exhaustive` attribute may only be applied to [`struct`s][struct], [`enu
 > [!NOTE]
 > `rustc` currently warns in some other positions, but this may be rejected in the future.
 
+r[attributes.type-system.non_exhaustive.duplicates]
+Duplicate instances of the `non_exhaustive` attribute have no effect.
+
+> [!NOTE]
+> `rustc` warns on subsequent duplicate `non_exhaustive` attributes.
+
 r[attributes.type-system.non_exhaustive.same-crate]
 Within the defining crate, `non_exhaustive` has no effect.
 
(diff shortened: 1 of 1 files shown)


src/attributes/type_system.md
```diff
@@ -3,207 +3,232 @@ r[attributes.type-system]
...
-```
-
-r[attributes.type-system.non_exhaustive.external-crate]
-Outside of the defining crate, types annotated with `non_exhaustive` have limitations that
-preserve backwards compatibility when new fields or variants are added.
+> [!EXAMPLE]
+> Using the definitions from [above][attributes.type-system.non_exhaustive.intro], the following examples when used within the same crate are allowed without restrictions.
+>
+> ```rust,ignore
+> // Non-exhaustive structs can be constructed as normal within the defining crate.
```
DanielEScherzer, line 87: can we include the definitions from above as hidden lines so that the example can be fully run?

src/attributes/type_system.md
```diff
@@ -3,207 +3,232 @@ r[attributes.type-system]
...
-
-// Cannot construct an instance of `Message::Quit`; if this were converted to
-// a tuple enum variant `upstream`, this would fail to compile.
-let message = Message::Quit;
-```
+`non_exhaustive` applied to a [`struct`][struct] or [`enum` variant][enum] prevents constructing that struct or enum variant outside of the defining crate.
+
+Non-exhaustive variants ([`struct`][struct] or [`enum` variant][enum]) cannot be constructed with a [StructExpression] \(including with [functional update syntax]).
+
+The implicitly defined same-named constant of a [unit-like struct][struct], or the same-named constructor function of a [tuple struct][struct], has a [visibility] no greater than `pub(crate)`. That is, if the struct’s visibility is `pub`, then the constant or constructor’s visibility is `pub(crate)`, and otherwise the visibility of the two items is the same (as is the case without `#[non_exhaustive]`).
```
DanielEScherzer: If the visibility rules aren't changed with the `non_exhaustive`, then why is this paragraph even needed?

If the rules did change, then the "as is the case without" part is confusing - what is the difference, and what is the part that is unchanged?

src/attributes/type_system.md
ehuss: Sure!

src/attributes/type_system.md
ehuss: Yea, that is a little awkward. I have updated the wording. It's essentially saying that the things in the value namespace change their visibility. The latter half is just emphasizing that the type (in the type namespace) is unaffected.

DanielEScherzer: Ah, okay

> The implicitly defined same-named constant of a [unit-like struct][struct], or the same-named constructor function of a [tuple struct][struct], has a [visibility] no greater than `pub(crate)`. For example, if the struct’s visibility is `pub`, then the constant or constructor’s visibility is `pub(crate)`. The visibility of the type itself is unaffected.

makes sense now, thanks

DanielEScherzer (approved):
looks good to me now, though obviously I don't have actual approval rights

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/1921/1d674cba8866cf7ab768d9b14530c3ba7f226620..670daf688675a44ed63daa8481e5777c80d4b747)*

PR #1925, opened 2025-07-03
Document (tuple) struct pattern namespace behavior

traviscross:
We had not documented what namespaces were used when matching struct and tuple struct patterns.  As it turns out, this is a bit nuanced. Let's document the current behavior, describe the nature of this nuance with an example, and note the possible uncertainly about the desirability of this behavior long-term.

cc @ehuss


ehuss (approved):
Thanks!

PR #1926, opened 2025-07-04
Replace set of en dashes with set of em dashes

traviscross:
We had used a pair of en dash somewhere we had meant to use a pair of em dashes.  Let's fix that.

cc @ehuss


ehuss (approved):
Thanks!

PR #1929, opened 2025-07-11
mention an important use for the naked attribute

tshepang:
Am thinking it's important since it was mentioned in `attributes.codegen.naked.intro`


ehuss (approved):
Thanks!

PR #1930, opened 2025-07-14
Change "allocated object" to "allocation".

theemathas:
This matches the change in <https://github.com/rust-lang/rust/pull/141224>


ehuss (approved):
Thanks!

Issue #1574, opened 2024-08-15
`Foo<T>` can implement `CoerceUnsized<Foo<U>>` when `T` implements `CoerceUnsized<U>` rather than `CoerceUnsized<Foo<U>>`


xmh0511:
https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions

> Additionally, a type `Foo<T>` can implement `CoerceUnsized<Foo<U>>` when `T` implements `Unsize<U>` or `CoerceUnsized<Foo<U>>`. This allows it to provide a unsized coercion to `Foo<U>`.

However, the [document](https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html) of `CoerceUnsized` regarding the corresponding case says:
> you can directly implement `CoerceUnsized<Wrap<U>>` for `Wrap<T>` where T: `CoerceUnsized<U>`

PR #1935, opened 2025-07-17
fix coerce unsized impl restriction

lcnr:
fixes https://github.com/rust-lang/reference/issues/1574

Unsure whether we want to document them even though `CoerceUnsized` is unstable. Source

https://github.com/rust-lang/rust/blob/f8f6997469237299c1d60814c7b9828602a1f8e4/compiler/rustc_hir_analysis/src/coherence/builtin.rs#L430-L547


ehuss:
Is there a test in rust-lang/rust for this particular rule?

lcnr:
not that I can see xd

ehuss:
Note: I believe @traviscross indicated he was working on this, but needed to figure out how to generalize how it is expressed.

PR #1938, opened 2025-07-18
glossary.md: tweak description of "dispatch"

tshepang:
- trait objects are not a *mechanism*
- *prefers* is not explained, so rather leave it out


ehuss (approved):
Thanks!

PR #1944, opened 2025-07-25
Fix and clarify CR LF normalization and CR in string literals

LukasKalbertodt:
This was slightly incorrect before. Relevant commits changing this:
- fa56fdba0e9dba35eb29d11c95c7a009ed67cb35
- 27e1ec97a75267d3c9efb9c91c7509eff98d11db

The normalization is not applied repeatedly, so CR LF pairs can still exist. Further, given that the normalization happens before lexing, the part "other than as part of such a string continuation escape" is not useful. Either it was CR LF in the raw input, but has already been transformed already (so the lexical grammar does not see CR). Or there is a surviving CR LF pair after the normalization, which is disallowed tho.

Here are two test programs showing this behavior:

    printf 'fn main() { "a\r\r\n\nb"; }' > code.rs | rustc -

Results in:

```
error: bare CR not allowed in string, use `\r` instead
 --> <anon>:1:15
  |
1 | fn main() { "a␍
  |               ^
  |
help: escape the character
  |
1 | fn main() { "a\r
  |               ++
```

And

    printf 'fn main() { "a\\\r\r\n\nb"; }' > code.rs | rustc -

Results in

```
error: unknown character escape: `\r`
 --> <anon>:1:16
  |
1 | fn main() { "a\␍
  |                ^ unknown character escape
  |
  = help: this is an isolated carriage return; consider checking your editor and version control settings
```

### DIFF
--- a/src/input-format.md
+++ b/src/input-format.md
@@ -24,6 +24,7 @@ r[input.crlf]
 ## CRLF normalization
 
 Each pair of characters `U+000D` (CR) immediately followed by `U+000A` (LF) is replaced by a single `U+000A` (LF).
+This happens once, not repeatedly, so after the normalization, there can still exist `U+000D` (CR) immediately followed by `U+000A` (LF) in the input (e.g. if the raw input contained "CR CR LF LF").
 
 Other occurrences of the character `U+000D` (CR) are left in place (they are treated as [whitespace]).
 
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

Issue #1946, opened 2025-07-27
Description of "aapcs" ABI is wrong


RalfJung:
The reference [says](https://doc.rust-lang.org/nightly/reference/items/external-blocks.html#r-items.extern.abi.aapcs) about the "appcs" ABI that it is "the default for ARM", but I think this may not be right. I am far from sure since I know next to nothing about ARM ABIs, but https://github.com/rust-lang/rust/issues/37810 indicates there might be a difference on 32bit ARM.

Cc @folkertdev @workingjubilee @Amanieu

Amanieu:
`"aapcs"` specifically refers to the soft-float ABI where floating-point values are passed in integer registers. So it is not actually the default ABI for hard-float targets.

PR #1948, opened 2025-07-27
fix aapcs ABI docs, and state where which ABIs are available

RalfJung:
Fixes https://github.com/rust-lang/reference/issues/1946.

Regarding ABI availability, this documents a post-https://github.com/rust-lang/rust/issues/137018 world. Some of the combinations documented as "not available" here actually do get accepted but trigger a future-compat warning.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -130,30 +130,33 @@ r[items.extern.abi.platform]
 There are also some platform-specific ABI strings:
 
 r[items.extern.abi.cdecl]
-* `unsafe extern "cdecl"` -- The default for x86\_32 C code.
+* `unsafe extern "cdecl"` -- The default for x86\_32 C code. Only available on x86\_32 targets.
 
 r[items.extern.abi.stdcall]
-* `unsafe extern "stdcall"` -- The default for the Win32 API on x86\_32.
+* `unsafe extern "stdcall"` -- The default for the Win32 API on x86\_32. Only available on x86\_32 targets.
 
 r[items.extern.abi.win64]
-* `unsafe extern "win64"` -- The default for C code on x86\_64 Windows.
+* `unsafe extern "win64"` -- The default for C code on x86\_64 Windows. Only available on x86\_64 targets.
 
 r[items.extern.abi.sysv64]
-* `unsafe extern "sysv64"` -- The default for C code on non-Windows x86\_64.
+* `unsafe extern "sysv64"` -- The default for C code on non-Windows x86\_64. Only available on x86\_64 targets.
 
 r[items.extern.abi.aapcs]
-* `unsafe extern "aapcs"` -- The default for ARM.
+* `unsafe extern "aapcs"` -- Softfloat ABI for ARM. Only available on ARM32 targets.
 
 r[items.extern.abi.fastcall]
 * `unsafe extern "fastcall"` -- The `fastcall` ABI -- corresponds to MSVC's
-  `__fastcall` and GCC and clang's `__attribute__((fastcall))`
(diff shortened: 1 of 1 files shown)


ehuss:
Is there somewhere in the compiler that we can easily see these defaults?

My understanding is that if the ABI is not specified, the default is [always `"C"`](https://github.com/rust-lang/rust/blob/ec7c02612527d185c379900b613311bc1dcbf7dc/compiler/rustc_abi/src/extern_abi.rs#L265). Is there something that says "C" is equivalent to these various ABIs for different targets? Or am I misunderstanding what it means by default?

Is aapcs the default for softfloat arm32?

RalfJung:
> My understanding is that if the ABI is not specified, the default is [always "C"](https://github.com/rust-lang/rust/blob/ec7c02612527d185c379900b613311bc1dcbf7dc/compiler/rustc_abi/src/extern_abi.rs#L265).

Correct. And "C" can be identical to other ABIs depending on the target. This PR documents my understanding of these equivalences. Cc @workingjubilee @Amanieu for people that know more about the nitty-gritty details.

Basically, `extern "C"` is always the default for C code on the *current* target, but some of these ABI strings (such as sysv64) let one use that ABI on other targets as well. On non-Windows x86_64, "C" and "sysv64" are equivalent, but on win64 they are not.

> Is aapcs the default for softfloat arm32?

I would think so but I don't know.

Amanieu:
> Is aapcs the default for softfloat arm32?

Yes, "C" == "aapcs" on soft-float arm32.

ehuss:
So if I am understanding correctly, the only way to determine what the behavior of the "C" calling convention is to dig through LLVM source, and each target seems to implement it differently, is that correct? For example aarch32/aarch64 use a method called `CCAssignFnForCall`. And x86 doesn't have any single place, but has methods like `isCallingConvWin64`. 

I'm not entirely clear for stdcall, it says "default for the Win32 API". Why does it explicitly say "API"? Is this saying, this is the default that *Microsoft* uses for the [Win32 API functions](https://learn.microsoft.com/en-us/cpp/cpp/stdcall?view=msvc-170) but is *not* the default of "C". That is, on i686-pc-windows-msvc, "C" is different from "stdcall"? Is that correct?

"thiscall" is also confusing, as it uses the term "default" for "C++ member functions". So it's not the default for the "C" calling convention?

(I'm finding the wording of "default" here confusing, so I'm trying to get clarity here.)

RalfJung:
> I'm not entirely clear for stdcall, it says "default for the Win32 API". Why does it explicitly say "API"? Is this saying, this is the default that Microsoft uses for the [Win32 API functions](https://learn.microsoft.com/en-us/cpp/cpp/stdcall?view=msvc-170) but is not the default of "C". That is, on i686-pc-windows-msvc, "C" is different from "stdcall"? Is that correct?

Yes, that is correct. The default for win32 in general ("C" on that target) is "cdecl".

> (I'm finding the wording of "default" here confusing, so I'm trying to get clarity here.)

Yeah that's fair. I went for a most minimal edit here to correct some wrong information. This section could probably use a larger rewrite, but I'm not sure if/when I will have the time for that.

RalfJung:
Looks like https://github.com/rust-lang/reference/pull/2000 now landed almost everything I wrote here, except that the reference now still says that "aapcs" is the "default for ARM" which is not correct.

ehuss:
Yea, we split this up into multiple PRs to try to make it easier to merge (#2000, #2001, #2003).

> "aapcs" is the "default for ARM" which is not correct.

Can you take a look at #2003 and see if that wording looks correct to you?

RalfJung:
Closing in favor of https://github.com/rust-lang/reference/pull/2003.

PR #1953, opened 2025-07-30
specify relative drop order of pattern bindings

### LINKED ISSUE #59: Document the "bindings drop in the reverse order of declaration" rule
This is documented in the book's first edition, but it doesn't seem to be in the reference at all. Just a mention in the Drop trait section, maybe copy-pasting the example from the book too, would suffice.

dianne:
Reference PR for rust-lang/rust#143764.

This replaces the unspecified `destructors.scope.bindings.match-pattern-order` with two new rules describing the run-time drop order of pattern bindings as of rust-lang/rust#143764. The included examples behave correctly on stable Rust as well.

Closes https://github.com/rust-lang/reference/issues/59

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -185,9 +185,48 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
 let declared_last = PrintOnDrop("Dropped first in outer scope");
 ```
 
-r[destructors.scope.bindings.match-pattern-order]
-If multiple patterns are used in the same arm for a `match` expression, then an
-unspecified pattern will be used to determine the drop order.
+r[destructors.scope.bindings.pattern-drop-order]
+If a pattern binds multiple variables, they are dropped in reverse order of declaration.
+
+```rust
+# struct PrintOnDrop(&'static str);
+# impl Drop for PrintOnDrop {
+#     fn drop(&mut self) {
+#         println!("drop({})", self.0);
+#     }
+# }
+let (declared_first, declared_last) = (
+    PrintOnDrop("Dropped last"),
+    PrintOnDrop("Dropped first"),
+);
+```
+
+r[destructors.scope.bindings.or-pattern-declaration-order]
+For the purpose of drop order, [or-patterns] declare their bindings in the order given by their first sub-pattern.
+
+```rust
+# struct PrintOnDrop(&'static str);
+# impl Drop for PrintOnDrop {
+#     fn drop(&mut self) {
+#         println!("drop({})", self.0);
+#     }
+# }
+// Drops `declared_last`, then `declared_first`.
+fn fixed_variable_drop_order<T>(
+    (Ok([declared_first, declared_last])
+    | Err([declared_last, declared_first])): Result<[T; 2], [T; 2]>
+) {}
+
+fixed_variable_drop_order(Ok([
+    PrintOnDrop("Dropped last"),
+    PrintOnDrop("Dropped first"),
+]));
+
+fixed_variable_drop_order(Err([
+    PrintOnDrop("Dropped first"),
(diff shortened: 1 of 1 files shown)


src/destructors.md
```diff
@@ -185,9 +185,48 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
 let declared_last = PrintOnDrop("Dropped first in outer scope");
 ```
 
-r[destructors.scope.bindings.match-pattern-order]
-If multiple patterns are used in the same arm for a `match` expression, then an
-unspecified pattern will be used to determine the drop order.
+r[destructors.scope.bindings.pattern-drop-order]
+If a pattern binds multiple variables, they are dropped in reverse order of delcaration.
```
dianne: The order of "declaration" is left mostly unspecified, other than by `destructors.scope.bindings.or-pattern-declaration-order` below. I like specifying the order as the reverse of another order to better match the other rules around drop order, and "reverse declaration order" is used for this purpose in the FLS[^1]. Other possible terms could be "reverse of source order", "reverse order of appearance", or "reverse order of first appearance" to combine the two rules. I don't think any of them read quite as well though.

[^1]: https://rust-lang.github.io/fls/ownership-and-deconstruction.html#drop-order

src/destructors.md
```diff
@@ -185,9 +185,48 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
 let declared_last = PrintOnDrop("Dropped first in outer scope");
 ```
 
-r[destructors.scope.bindings.match-pattern-order]
-If multiple patterns are used in the same arm for a `match` expression, then an
-unspecified pattern will be used to determine the drop order.
+r[destructors.scope.bindings.pattern-drop-order]
```
dianne: This also only specifies the dynamic semantics of drops, not their static semantics; those seem to be handled by the Nomicon rather than the Reference. See e.g. rust-lang/rust#142057 for an instance where the static semantics would need additional specification. Or for (probably) a non-bug, consider an `x @ Struct(y)` pattern: `x` is declared before `y`, but `y` is bound before `x` and (incidentally, as an implementation detail) is considered potentially live until after `x` is dropped. This is to allow copying into `y` before moving into `x`. This can't affect the order in which destructors are run, since if they both had significant drops, neither would be `Copy`, and the match scrutinee can't be moved into both `y` and `x`. As far as I can tell, the only effect `x` being dropped first has is slightly more permissive drop-checking in edge cases where e.g. `x` is mutated to contain a reference to `y`.

Since the Reference only specifies when destructors are run, I've chosen to keep things simple and not handle any of that.

src/destructors.md
```diff
@@ -185,9 +185,48 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
...
+
+fixed_variable_drop_order(Ok([
+    PrintOnDrop("Dropped last"),
+    PrintOnDrop("Dropped first"),
+]));
+
+fixed_variable_drop_order(Err([
+    PrintOnDrop("Dropped first"),
+    PrintOnDrop("Dropped last"),
+]));
```
dianne: I feel like these examples would benefit from comments, but I haven't been able to come up with something concise enough yet.

traviscross (approved):

src/destructors.md
traviscross: Agreed.

src/destructors.md
traviscross: Agreed that makes sense for this PR.  If you're interested in adding the content about the static semantics to the Reference separately, that's material we'd like to have in the Reference and a PR we'd certainly appreciate.

src/destructors.md
traviscross: Added comments here.

PR #1954, opened 2025-08-02
Document assignment expression as coercion site

paldepind:
I was recently confused as for why this is allowed in Rust
```rust
let mut a = &12;
let b = &&34;
a = b; 
```
as I couldn't find that documented anywhere.

I was told that assignment expressions are in fact coercion sites which explains the example above.

This PR documents assignment expressions as coercion sites in the reference.

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -78,6 +78,16 @@ r[coerce.site.return]
   }
   ```
 
+r[coerce.site.assignment]
+* Assigned value operands in assignment expressions
+
+  For example, `y` is coerced to have type `&i8` in the following:
+  ```rust
+  let mut x = &0i8;
+  let y = &mut 42i8;
+  x = y;
+  ```
+
 r[coerce.site.subexpr]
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! This seems correct to me.

I'm also feeling like the coercion site listing is also missing all of the bin operators. 

Can someone at @rust-lang/types maybe give this a quick review? And let us know if it would make sense to also include bin ops, too, and if there are any complications are caveats around those. (The code for that seems a bit more tricky to follow.)

jackh726:
This looks correct. I'm not sure about bin ops in general.

paldepind:
Thanks for taking a look! I'm happy to see this gaining some traction as it's such a simple gap to get filled out.

I don't think binary operators needs to be mentioned as they can be understood by their desugaring. E.g., `a + b` is equivalent to `std::ops::Add::add(a, b)` and then `a` and `b` being coercion sites falls out the fact that arguments are coercion sites.

ehuss (approved):
Thanks!

PR #1955, opened 2025-08-07
Remove note on accepted invalid `should_panic` syntax

traviscross:
We had left a note about the fact that we accepted syntax for the `should_panic` attribute that was invalid.  We no longer do, so let's remove that note.

See:

- https://github.com/rust-lang/rust/pull/143808

cc @ehuss @Zalathar


ehuss (approved):
Thanks!

PR #1957, opened 2025-08-12
specify `if let` guards with updated scoping rules

dianne:
This is based on rust-lang/reference#1823 by @Kivooeo and @ehuss, rebased to resolve conflicts. Per https://github.com/rust-lang/rust/pull/141295#issuecomment-3172434801, I'm opening this as a separate PR to contribute an updated specification of `if let` guards' drop-scoping rules. The new spec is based on the compiler's implementation[^1], accounting for the changes in rust-lang/rust#143376 and additional corner-cases in examples. I've also done some minor edits to the sections on lexical scope and match expressions, but for the most part they're the same as in rust-lang/reference#1823.

[^1]: As with the rules for other syntactic constructs (e.g. `let` statements and `if` expressions), this is slightly simplified. Due to implementation details, the compiler's notion of drop scopes is more fine-grained than is necessary for the language spec.

Tracking issue: rust-lang/rust#51114
Stabilization: https://github.com/rust-lang/rust/pull/141295

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -107,14 +107,6 @@ r[destructors.scope.nesting.match-guard]
 r[destructors.scope.nesting.match-arm]
 * The parent of the expression after the `=>` in a `match` expression is the scope of the arm that it's in.
 
-r[destructors.scope.nesting.match-guard-pattern]
-* The parent of an `if let` guard pattern is the scope of the guard expression.
-
-r[destructors.scope.nesting.match-guard-bindings]
-* Variables bound by `if let` guard patterns have their drop scope determined by guard success:
-  - On guard failure: dropped immediately after guard evaluation
-  - On guard success: scope extends to the end of the match arm body
-
 r[destructors.scope.nesting.match]
 * The parent of the arm scope is the scope of the `match` expression that it belongs to.
 
@@ -149,8 +141,8 @@ patterns_in_parameters(
 r[destructors.scope.bindings]
 ### Scopes of local variables
 
-r[destructors.scope.bindings.intro]
-Local variables declared in a `let` statement are associated to the scope of the block that contains the `let` statement. Local variables declared in a `match` expression are associated to the arm scope of the `match` arm that they are declared in.
+r[destructors.scope.bindings.let]
+Local variables declared in a `let` statement are associated to the scope of the block that contains the `let` statement.
 
 ```rust
 # struct PrintOnDrop(&'static str);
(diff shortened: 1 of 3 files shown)


src/destructors.md
```diff
@@ -255,9 +289,8 @@ smallest scope that contains the expression and is one of the following:
 * A statement.
 * The body of an [`if`], [`while`] or [`loop`] expression.
 * The `else` block of an `if` expression.
-* The non-pattern matching condition expression of an `if` or `while` expression,
-  or a `match` guard.
-* The body expression for a match arm.
+* The non-pattern matching condition expression of an `if` or `while` expression or a non-pattern-matching `match` guard condition operand.
+* The pattern-matching guard, if present, and body expression for a `match` arm.
```
dianne, line 272: To address https://github.com/rust-lang/reference/pull/1823#issuecomment-3173098920 (cc @theemathas), the temporary lifetime of an `if let` guard scrutinee is always extended, like for `if let` expressions. There aren't lifetime extension rules like there are for `let` statements. This rule should hopefully cover that (similar to how the analogous rule for "The pattern-matching condition(s) and consequent body of `if`" does below). There's an example further down too showing that the scrutinee lives until the end of the arm. Do you think this is sufficient, or would this benefit from additional clarification?

theemathas: This code currently compiles. I assumed that it does so due to temporary lifetime extension. Without lifetime extension, the `Thing` temporary would have been dropped before `x` is used. Could you explain why exactly does it compile?

```rust
#![feature(if_let_guard)]

fn main() {
    match () {
        _ if let x = &make_thing() => {
            let _a = x;
        },
        _ => {}
    }
}

struct Thing;

fn make_thing() -> Thing {
    Thing
}
```

theemathas: Oh, are you saying that both the guard and the body expression *together* count as a single temporary scope (and not two)? I find that unintuitive, but I suppose that works.

dianne: There's a single temporary scope for the arm, which encompasses the guard and body expression, so that `if let` scrutinees' temporaries live through the body, yes; `if let` scrutinees aren't temporary destruction scopes by default. Maybe it would be clearer to phrase the temporary scope as being as for the whole arm, with a clarifying note that it includes the guard. Something like "* The arm of a `match` expression. This includes the arm's guard, if present."

Personally, I'd like stricter scoping rules for `if let`, but this is consistent with `if let`/`while let` expressions. Making their scrutinees be temporary drop scopes by default with lifetime extension rules (like `let` statements have) would be a larger change requiring an edition break.

src/expressions/match-expr.md
```diff
@@ -9,15 +9,35 @@ MatchExpression ->
         MatchArms?
     `}`
 
-Scrutinee -> Expression _except [StructExpression]_
+Scrutinee -> Expression _except_ [StructExpression]
```
ehuss: This is a pre-existing mistake.

```suggestion
Scrutinee -> Expression _except [StructExpression]_
```

src/destructors.md
```diff
@@ -185,6 +182,43 @@ let declared_first = PrintOnDrop("Dropped last in outer scope");
 let declared_last = PrintOnDrop("Dropped first in outer scope");
 ```
 
+r[destructors.scope.bindings.match-arm]
+Local variables declared in a `match` expression or pattern-matching `match` guard are associated to the arm scope of the `match` arm that they are declared in.
+
+```rust
+# #![allow(irrefutable_let_patterns)]
```
ehuss, line 165: This is missing the struct.

```suggestion
# #![allow(irrefutable_let_patterns)]
# struct PrintOnDrop(&'static str);
# impl Drop for PrintOnDrop {
#     fn drop(&mut self) {
#         println!("drop({})", self.0);
#     }
# }
```

src/expressions/match-expr.md
dianne: Fixed: [(diff)](https://github.com/rust-lang/reference/compare/6f1597adebc92c4978a102753434212c39fdc34f..f1b193ce9e1f582125dbdddc9b42a2b569384f70#diff-c2e020de393737ddbcc629dee190deb3ceddea0b5328452a4d255fff4c202c19)

src/destructors.md
dianne: Fixed: [(diff)](https://github.com/rust-lang/reference/compare/6f1597adebc92c4978a102753434212c39fdc34f..f1b193ce9e1f582125dbdddc9b42a2b569384f70#diff-76f8b59537d5af10271dfb21f24d9288b354500192420a6438ec9efbd119f986)

theemathas:
Is this waiting on anything in particular?

dianne:
It looks like my removal of the "Match Guards and Pattern Binding" section was undone? I've re-removed it. I wanted to leave commits before my own mostly untouched, but I can clean up the commit history to remove anything that I'm not preserving in this PR if that'd help keep what's actually meant to be here clearer.

ehuss:
Sorry about that @dianne! I had double-checked the reflog when I was fixing that, and I'm not sure how it got missed.

traviscross (approved):
Looks good to me.

src/expressions/match-expr.md
```diff
@@ -137,11 +157,74 @@ r[expr.match.guard.shared-ref]
...
+If any guard condition operand is a `let` pattern, then none of the condition operands can be a `||` [lazy boolean operator expression][expr.bool-logic] due to ambiguity and precedence with the `let` scrutinee.
+
+> [!EXAMPLE]
+> If a `||` expression is needed, then parentheses can be used. For example:
+>
+> ```rust
+> # let foo = Some(123);
+> match foo {
+>     // Parentheses are required here.
+>     Some(x) if (x < -100 || x > 20) => {}
```
est31: As this doesn't have a `let`, no parens are required here. I suppose you meant to add a `let`.

dianne: Nice catch; I'd missed that. I've made it look a bit more like the example above it: [(diff)](https://github.com/rust-lang/reference/compare/20e729c93b46c91325aba4e0e3d3ef779161ded7..053d9fef3bfab5b7de63f59320afd7b6fb826184)

ehuss (approved):
Thanks! Merging now that https://github.com/rust-lang/rust/pull/141295 has merged.

PR #1961, opened 2025-08-15
Fix indentation of static path restriction text

traviscross:
In the const eval chapter, we have a list of kinds of expressions that are allowed.  Paths to statics are one kind that is allowed, but have certain restrictions we describe in bullet points.  We then have text that clarifies when these bullet points are checked.

This text was outdented beyond the relevant item, which broke up the outer bullet point list and created (momentary) ambiguity about the scope of the rules the language might be affecting.

cc @ehuss


ehuss (approved):
Thanks!

PR #1962, opened 2025-08-15
Add missing rule identifier for const outer generics

traviscross:
In the section on const contexts, we have a rule about when outer generics cannot be used.  This was missing a rule identifier, so the text appeared to be part of the preceding rule.  Let's fix that.

cc @ehuss


ehuss (approved):
Thanks!

PR #1966, opened 2025-08-19
Remove tuple index carve out

jieyouxu:
> [!CAUTION]
>
> This should only be merged after FCP on https://github.com/rust-lang/rust/pull/145463 passes and the PR is itself merged.

This is being proposed to be removed in <https://github.com/rust-lang/rust/pull/145463> as we never intended for this to be valid syntax. The concrete example forms are in the nomination report in https://github.com/rust-lang/rust/pull/145463#issue-3325732856.

<details>
<summary>Outdated discussion on the grammar</summary>

### Note on `TUPLE_INDEX`

This production looks wrong to me

```grammar,lexer
TUPLE_INDEX -> INTEGER_LITERAL
```

`INTEGER_LITERAL` covers *all* the decimal, binary, octal and hex literal forms, but AFAICT `TUPLE_INDEX` as implemented does not even accept the full range of decimal syntax. I.e.

```grammar,lexer
DEC_LITERAL -> DEC_DIGIT (DEC_DIGIT|`_`)*
```

but e.g.

```rs
fn main() {
    let x = (1,);
    println!("{}", x.0_0); //~ ERROR no field `0_0` on type `({integer},)`
}
```

I believe the accepted `TUPLE_INDEX` grammar here is *only*

```grammar,lexer
DEC_LITERAL -> DEC_DIGIT+
```

Let me know if that sounds right.

</details>

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -624,7 +624,7 @@ r[lex.token.literal.int.tuple-field]
 
 r[lex.token.literal.int.tuple-field.syntax]
 ```grammar,lexer
-TUPLE_INDEX -> INTEGER_LITERAL
+TUPLE_INDEX -> DEC_LITERAL | BIN_LITERAL | OCT_LITERAL | HEX_LITERAL
 ```
 
 r[lex.token.literal.int.tuple-field.intro]
@@ -637,6 +637,8 @@ start with `0` and each successive index increments the value by `1` as a
 decimal value. Thus, only decimal values will match, and the value must not
 have any extra `0` prefix characters.
 
+Tuple indices may not include any suffixes (such as `usize`).
+
 ```rust,compile_fail
 let example = ("dog", "cat", "horse");
 let dog = example.0;
@@ -644,10 +646,9 @@ let cat = example.1;
 // The following examples are invalid.
 let cat = example.01;  // ERROR no field named `01`
 let horse = example.0b10;  // ERROR no field named `0b10`
+let unicorn = example.0usize; // ERROR suffixes on a tuple index are invalid
+let underscore = example.0_0; // ERROR no field `0_0` on type `(&str, &str, &str)`
 ```
-
-> [!NOTE]
-> Tuple indices may include certain suffixes, but this is not intended to be valid, and may be removed in a future version. See <https://github.com/rust-lang/rust/issues/60210> for more information.
 
 r[lex.token.literal.float]
 #### Floating-point literals


ehuss:
The grammar is defined by the AST. It is essentially the pre-expansion accepted text, or to think of it another way, it is the valid input to a macro metavariable.

I haven't looked at your PR to determine exactly what it does, but from what I can tell it rejects the suffix at parsing time. So I think the new grammar is something like:

```
TUPLE_INDEX -> DEC_LITERAL | BIN_LITERAL | OCT_LITERAL | HEX_LITERAL
```

Personally I think that is weird, and would be simpler if it could just be DEC_LITERAL. However, since it is an exact textual match, it still has the oddity that `04` is not the same as `4`, so 🤷 .

jieyouxu:
It really does feel weird, because e.g.

```rs
let _x = (4,).0x0;
              ^^^
```

clearly does not work... before or after my PR.

ehuss:
To be clear, that does parse successfully. For example:

```rust
// Parse, but don't process:
#[cfg(false)]
fn f()  {
    let _x = (4,).0x0;
}

// Or via a macro:
macro_rules! m {
    ($e:expr) => {};
}

fn main() {
    m!((4,).0x0);
}
```

`0x0` gives the same error message as `00`. Neither is a textual match for `0`.

jieyouxu:
Right... I'll update it to

```
TUPLE_INDEX -> DEC_LITERAL | BIN_LITERAL | OCT_LITERAL | HEX_LITERAL
```

like you suggested.

src/tokens.md
```diff
@@ -637,18 +637,18 @@ start with `0` and each successive index increments the value by `1` as a
...
+Tuple indices may not include any suffixes (such as `usize`).
+
 ```rust,compile_fail
 let example = ("dog", "cat", "horse");
 let dog = example.0;
 let cat = example.1;
 // The following examples are invalid.
 let cat = example.01;  // ERROR no field named `01`
 let horse = example.0b10;  // ERROR no field named `0b10`
+let unicorn = example.0usize; // ERROR suffixes on a tuple index are invalid
```
traviscross, line 649: Probably worth giving an example to show that we don't accept underscores either.

traviscross (approved):

src/tokens.md
jieyouxu: Included example

```rs
let underscore = example.0_0; // ERROR no field `0_0` on type `(&str, &str, &str)`
```

jieyouxu:
Included

```rs
let underscore = example.0_0; // ERROR no field `0_0` on type `(&str, &str, &str)`
```

example.

ehuss (approved):
Merging now that https://github.com/rust-lang/rust/pull/145463 has merged. Thanks!

PR #1974, opened 2025-08-21
docs(ref): Specify frontmatter

epage:
Tracking issue: rust-lang/rust#136889

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -6,6 +6,7 @@
 
 - [Lexical structure](lexical-structure.md)
     - [Input format](input-format.md)
+    - [Frontmatter](frontmatter.md)
     - [Keywords](keywords.md)
     - [Identifiers](identifiers.md)
     - [Comments](comments.md)
--- a/src/frontmatter.md
+++ b/src/frontmatter.md
@@ -1,0 +1,58 @@
+r[frontmatter]
+# Frontmatter
+
+r[frontmatter.syntax]
+```grammar,lexer
+@root FRONTMATTER ->
+      FRONTMATTER_FENCE HORIZONTAL_WHITESPACE* INFOSTRING? HORIZONTAL_WHITESPACE* LF
+      (FRONTMATTER_LINE LF )*
+      FRONTMATTER_FENCE[^matched-fence] HORIZONTAL_WHITESPACE* LF
+
+FRONTMATTER_FENCE -> `-`{3..255}
+
+INFOSTRING -> (XID_Start | `_`) ( XID_Continue | `-` | `.` )*
+
+FRONTMATTER_LINE -> (~INVALID_FRONTMATTER_LINE_START (~INVALID_FRONTMATTER_LINE_CONTINUE)*)?
+
+INVALID_FRONTMATTER_LINE_START -> (FRONTMATTER_FENCE[^escaped-fence] | CR | LF)
+
+INVALID_FRONTMATTER_LINE_CONTINUE -> CR | LF
+
+HORIZONTAL_WHITESPACE ->
+      U+0009  // horizontal tab, `'\t'`
+    | U+0020  // space, `' '`
+```
+
+[^matched-fence]: The closing fence must have the same number of `-` as the opening fence
+[^escaped-fence]: A `FRONTMATTER_FENCE` at the beginning of a `FRONTMATTER_LINE` is only invalid if it has the same or more `-` as the `FRONTMATTER_FENCE`
+
+r[frontmatter.intro]
+Frontmatter is an optional section for content intended for external tools without requiring these tools to have full knowledge of the Rust grammar.
+
+```rust
+#!/usr/bin/env cargo
+---
+[dependencies]
+fastrand = "2"
+---
+
+fn main() {
+    let num = fastrand::i32(..);
+    println!("{num}");
+}
+```
+
+r[frontmatter.document]
+Frontmatter may only be preceded by a [shebang] and [whitespace].
+
+r[frontmatter.fence]
+The delimiters are referred to as a *fence*. The opening and closing fences must be at the start of a line. They must be a matching pair of hyphens (`-`), from 3 to 255. A fence may be followed by horizontal whitespace.
+
+r[frontmatter.infostring]
+Following the opening fence may be an infostring for identifying the intention of the contained content. An infostring may be followed by horizontal whitespace.
+
+r[frontmatter.body]
+The body of the frontmatter may contain any content except for a line starting with as many or more hyphens (`-`) than in the fences or carriage returns.
+
+[shebang]: input-format.md#shebang-removal
+[whitespace]: whitespace.md
--- a/src/input-format.md
+++ b/src/input-format.md
@@ -54,6 +54,11 @@ fn main() {
 r[input.shebang.inner-attribute]
 As an exception, if the `#!` characters are followed (ignoring intervening [comments] or [whitespace]) by a `[` token, nothing is removed. This prevents an [inner attribute] at the start of a source file being removed.
 
+r[input.frontmatter]
+## Frontmatter removal
+
+After some [whitespace], [frontmatter] may next appear in the input.
+
 r[input.tokenization]
 ## Tokenization
 
@@ -64,7 +69,7 @@ The resulting sequence of characters is then converted into tokens as described 
 >
 > - Byte order mark removal.
 > - CRLF normalization.
-> - Shebang removal when invoked in an item context (as opposed to expression or statement contexts).
+> - Shebang and frontmatter removal when invoked in an item context (as opposed to expression or statement contexts).
 >
 > The [`include_str!`] and [`include_bytes!`] macros do not apply these transformations.
 
@@ -73,4 +78,5 @@ The resulting sequence of characters is then converted into tokens as described 
 [comments]: comments.md
 [Crates and source files]: crates-and-source-files.md
 [_shebang_]: https://en.wikipedia.org/wiki/Shebang_(Unix)
+[frontmatter]: frontmatter.md
 [whitespace]: whitespace.md
--- a/src/items/modules.md
+++ b/src/items/modules.md
@@ -123,7 +123,7 @@ r[items.mod.attributes]
 ## Attributes on modules
 
 r[items.mod.attributes.intro]
-Modules, like all items, accept outer attributes. They also accept inner attributes: either after `{` for a module with a body, or at the beginning of the source file, after the optional BOM and shebang.
+Modules, like all items, accept outer attributes. They also accept inner attributes: either after `{` for a module with a body, or at the beginning of the source file, after the optional BOM, shebang, and frontmatter.
 
 r[items.mod.attributes.supported]
 The built-in attributes that have meaning on a module are [`cfg`], [`deprecated`], [`doc`], [the lint check attributes], [`path`], and [`no_implicit_prelude`]. Modules also accept macro attributes.


src/SUMMARY.md
```diff
@@ -8,6 +8,7 @@
     - [Input format](input-format.md)
     - [Keywords](keywords.md)
     - [Identifiers](identifiers.md)
+    - [Frontmatter](frontmatter.md)
```
epage: Not stable yet: https://github.com/rust-lang/rust/issues/136889

epage: I put this under "Lexical structure" because this shebang is there

src/crates-and-source-files.md
```diff
@@ -4,6 +4,7 @@ r[crate]
 r[crate.syntax]
 ```grammar,items
 @root Crate ->
+    FRONTMATTER?
     InnerAttribute*
     Item*
```
epage: I used comments as my guide which were mostly SCREAMING_CASE.  Unsure when things should be SCREAMING_CASE vs UpperCamelCase.

epage: Despite shebang's not having being here, I assumed I should put FRONTMATTER here

src/frontmatter.md
```diff
@@ -0,0 +1,42 @@
...
+The delimiters are referred to as a "fence."
+The opening and closing fences must be at the start of a line.
+They must be a matching pair of three or more dashes (`-`).
+
+r[frontmatter.infostring]
+Following the opening fence may be an infostring for identifying the intention of the contained content.
+An infostring may be preceded by non-newline whitespace.
+
+r[frontmatter.body]
+The body of the frontmatter may contain any content except for a line starting with as many or more dashes (`-`) than in the fences.
```
epage: This follows the RFC rather than rustc's current implementation, see rust-lang/rust#141367

src/crates-and-source-files.md
```diff
@@ -4,6 +4,7 @@ r[crate]
 r[crate.syntax]
 ```grammar,items
```
epage, line 5: Aside: it took me a while before I found the appropriate documentation for this.

I quickly went to [CONTRIBUTING.md](https://github.com/rust-lang/reference/blob/master/CONTRIBUTING.md) but skipped over the link to the authoring page because it was at the end of the intro.  I skimmed the sections until I found "Adding Documentation" which seemed to describe my situation but I found no link.

Eventually I found [authoring.md](https://github.com/rust-lang/reference/blob/master/docs/authoring.md) and thankfully I read thoroughly enough to notice the "Grammar" section at the bottom.  At that point, I was mostly able to interpret the results to figure out what I needed (e.g. the limitations of `~`, what to search for to understand how to do footnotes).

As noted at https://github.com/rust-lang/reference/pull/1974/files#r2292171963, it doesn't really give a style guidance on casing.

src/frontmatter.md
```diff
@@ -0,0 +1,42 @@
...
+
+FRONTMATTER_FENCE -> `---` `-`*
+
+INFOSTRING -> XID_Start ( XID_Continue | `.` )*
+
+FRONTMATTER_LINE -> (~INVALID_FRONTMATTER_LINE_START (~INVALID_FRONTMATTER_LINE_CONTINUE)*)?
+
+INVALID_FRONTMATTER_LINE_START -> (FRONTMATTER_FENCE[^escaped-fence] | LF)
+
+INVALID_FRONTMATTER_LINE_CONTINUE -> LF
```
epage: Was unsure whether to create a new rule for non-newline whitespace and specify that in every location or not.

src/crates-and-source-files.md
mattheww: The reason shebang isn't there is that shebangs are ignored at the start of any file, not just in the file which defines a crate.

If that's true for frontmatter too, `input-format.md` might be a better place to say how `frontmatter.md` fits in.

src/frontmatter.md
```diff
@@ -0,0 +1,42 @@
...
+r[frontmatter.syntax]
+```grammar,lexer
+FRONTMATTER ->
+      FRONTMATTER_FENCE INFOSTRING? LF
+      (FRONTMATTER_LINE LF )*
+      FRONTMATTER_FENCE[^matched-fence] LF
+
+FRONTMATTER_FENCE -> `---` `-`*
+
+INFOSTRING -> XID_Start ( XID_Continue | `-` | `.` )*
```
mattheww: `XID_Start` doesn't include `_`, but the current implementation does allow infostrings to start with `_`.

src/crates-and-source-files.md
epage: This is a bit confusing.

`input-format.md` is specifically under the section for lexical analysis.  

`crates-and-source-files.md` includes what looks like its the AST but its specifically scoped to the crate root. `r[crate-items]` talks generally about any rust source file.

So, like shebang, I'll leave this out but this feels like something that could be cleaned up.

src/frontmatter.md
epage: Thanks for catching that!

src/frontmatter.md
```diff
@@ -0,0 +1,42 @@
...
+
+FRONTMATTER_FENCE -> `---` `-`*
+
+INFOSTRING -> (XID_Start | `_`) ( XID_Continue | `-` | `.` )*
+
+FRONTMATTER_LINE -> (~INVALID_FRONTMATTER_LINE_START (~INVALID_FRONTMATTER_LINE_CONTINUE)*)?
+
+INVALID_FRONTMATTER_LINE_START -> (FRONTMATTER_FENCE[^escaped-fence] | LF)
+
+INVALID_FRONTMATTER_LINE_CONTINUE -> LF
```
epage: Thinking more on this, I feel like I should be explicit where frontmatter whitespace is allowed.

I assume I should then pull out a whitespace grammar rule to build on that.

epage: Something that i don't make explicit is `CR`.  In most places that can be chalked up to whitespace except for `INVALID_FRONTMATTER_LINE_START` and `INVALID_FRONTMATTER_LINE_CONTINUE`.

src/whitespace.md
```diff
@@ -4,23 +4,32 @@ r[lex.whitespace]
...
+
+HORIZONTAL_WHITESPACE ->
+      TAB
+    | U+0020  // space ' '
+
+TAB -> U+0009  // horizontal tab ('\t')
+
+LF -> U+000A  // line feed ('\n')
+
+CR -> U+000D  // carriage return ('\r')
```
traviscross: In these sort of grammar productions that are essentially lists, we generally find it more clear to not mix terminals and non-terminals even if that leads to some duplication.  As applied here, that would suggest inlining `LF` and `CR` in `END_OF_LINE` and `TAB` in `HORIZONTAL_WHITESPACE`.  Does that make sense, or are there non-obvious problems with that here?

traviscross: There is a particular style we prefer for notating the character literal for each code point, e.g.:

```rust
// Carriage return, `'\r'`
```

This PR takes a different approach and doesn't use the backtick literal syntax.  Is there a reason for that, or can it align with the other style while still achieving the purpose?

traviscross: Comments in the Reference aim to follow a particular style.  One part of that is using capitalization consistently and using sentence case.  This PR diverges from that.  Is there a good reason for this, or can we try to align it with our conventions?

traviscross: This PR notates some characters with "Unicode" but not others.  But, of course, Unicode has code points for all of these characters.  Is there a rationale for why only some are so marked that's important for this PR, or could we normalize this?

src/whitespace.md
```diff
@@ -4,23 +4,32 @@ r[lex.whitespace]
...
+    | HORIZONTAL_WHITESPACE
+
+END_OF_LINE ->
+      LF
+    | U+000B // vertical tabulation
+    | U+000C // form feed
+    | CR
+    | U+0085 // Unicode next line
+    | U+2028 // Unicode LINE SEPARATOR
+    | U+2029 // Unicode PARAGRAPH SEPARATOR
```
traviscross: In my copy of the [Unicode Character Database](https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt), I see:

```
0085;<control>;Cc;0;B;;;;;N;NEXT LINE (NEL);;;;
```

```
2028;LINE SEPARATOR;Zl;0;WS;;;;;N;;;;;
2029;PARAGRAPH SEPARATOR;Zp;0;B;;;;;N;;;;;
```

In a somewhat more pretty form:

- https://util.unicode.org/UnicodeJsps/character.jsp?a=0085
- https://util.unicode.org/UnicodeJsps/character.jsp?a=2028
- https://util.unicode.org/UnicodeJsps/character.jsp?a=2029

That is, in the database, U+0085 has a null name and a name alias of "NEXT LINE|NEL", U+2028 has a name of "LINE SEPARATOR" and a null name alias, and U+2029 has a name of "PARAGRAPH SEPARATOR" and a null name alias.

Given that, is there a reason that we need to capitalize these differently, or could we perhaps align the capitalization with Reference norms without losing something important here?

I do see, in [*The Unicode Standard, Version 17.0.0, §5.8.1 "Definitions", Table 5-1*](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-5/#G41643) that all of "next line", "line separator", and "paragraph separator" are presented in lowercase, but that doesn't seem to explain why we might want to treat these differently.

src/frontmatter.md
```diff
@@ -0,0 +1,43 @@
...
+Frontmatter is an optional section for content intended for external tools without requiring these tools to have full knowledge of the Rust grammar.
+
+r[frontmatter.document]
+Frontmatter may only be preceded by a [shebang] and whitespace.
+
+r[frontmatter.fence]
+The delimiters are referred to as a "fence."
+The opening and closing fences must be at the start of a line.
+They must be a matching pair of three or more hyphens (`-`).
+A fence may be followed by horizontal whitespace.
```
traviscross: In the Reference, for new content, we don't hard wrap lines.  Please unwrap these, here and elsewhere.

src/whitespace.md
```diff
@@ -4,23 +4,32 @@ r[lex.whitespace]
...
+
+IGNORABLE_CODE_POINT ->
+      U+200E // Unicode LEFT-TO-RIGHT MARK
+    | U+200F // Unicode RIGHT-TO-LEFT MARK
+
+HORIZONTAL_WHITESPACE ->
+      TAB
+    | U+0020  // space ' '
```
traviscross: Looking at the grammar productions above for frontmatter, it seems that only `HORIZONTAL_WHITESPACE` is used (and `LF`, of course).  If we were willing to accept some duplication, i.e. by defining `HORIZONTAL_WHITESPACE` directly in terms of terminals (that also would appear under `WHITESPACE`), does that work for defining frontmatter without these other changes, or is there a reason these other changes to the layout, organization, and styling of the `WHITESPACE` grammar are important for defining frontmatter?

traviscross (commented):
Thanks for the PR.  Before digging into the rest, let me ask some questions about changes here related to layout, organization, and style.

src/whitespace.md
epage: I can see it both ways.

Extra abstractions can make things less clear.

Lack of use of abstractions highlights a difference which can also be confusing or make searching more difficult.

Which gets me thinking, with how thin this abstraction is, should we even have a production for `TAB`, `LF`, or `CR`?

epage: I don't remember what led to the backticks being removed.  I can add them back in.

epage: For end-of-line markers and ignorable code points, the casing and text matches https://www.unicode.org/reports/tr31/#R3a-1

In that, space is `SPACE` and horizontal tab is `horizontal tabulation, TAB`.

epage: I don't remember how that got introduced, removing them

src/whitespace.md
epage: This was to match how they are specified where it talks about the different categories of whitespace, see https://www.unicode.org/reports/tr31/#R3a-1

src/whitespace.md
traviscross: Thanks.

traviscross: Thanks.

traviscross: Let's take the remainder of this thread in https://github.com/rust-lang/reference/pull/1974#discussion_r2476064079.

traviscross: It's fairly common in grammars to have separate productions for `TAB`, `LF`, and `CR`, even if duplicated.  I'd suggest keeping these three but inlining the terminals into the other productions.

src/whitespace.md
traviscross: It's not clear to me that the capitalization and other matters of presentation there are intended to be more normative than the capitalization in the Unicode Character Database (which is normatively part of the Unicode standard).

To my eyes, it looks like they made the editorial choice in tr31 to notate code point names as presented in the database, and then where the name is null, to present the primary code point alias in parentheses and in lowercase (and to elide secondary aliases).  This presentation was, I'm guessing, intended to notate this difference between names and aliases.

Is there a reason that it's important to follow the presentation in tr31 exactly?  If so, I'd expect that we'd want to follow the convention with the parentheses as well.

However, it's not clear to me that it's important, for our purposes, to make this distinction between code point names and code point aliases.  If that's right, and there's not, then I'd prefer we render these according to our normal conventions for comments, i.e., in sentence case (especially as the authors of Unicode standards documents also seem to treat the capitalizations as malleable).

But, is that right, or is there an important reason we need to distinguish, for our purposes, code point names and aliases (where the name is null)?

src/frontmatter.md
epage: Huh, I thought I had seen a comment about semantic line feeds (which were introduced to me by ehuss).

I'm surprised then by the motivation

> Do not wrap long lines. This helps with reviewing diffs of the source.

For myself, I find it makes reviewing harder.

src/whitespace.md
epage: I think duplicating an entry like `HORIZONTAL_WHITESPACE` is anti-user.

src/whitespace.md
epage: > Is there a reason that it's important to follow the presentation in tr31 exactly? If so, I'd expect that we'd want to follow the convention with the parentheses as well.

That is the document we are referring back to on this page.

traviscross: Yes, understood.  How would you then apply that to the points and questions raised above?

src/frontmatter.md
traviscross: Understand.  However, it's our current policy for the document.  If you want us to reconsider this policy, I'd suggest filing a separate issue or PR to make the case.  For this PR, it would seem to make the most sense to follow the existing policy.  Is that reasonable?  Does that work for you?

src/whitespace.md
traviscross: > I think duplicating an entry like `HORIZONTAL_WHITESPACE` is anti-user.

Obviously we strive to be pro-user.  However, I'm not clear how this stylistic choice would be anti-user.  If you could please elaborate, I'd appreciate it.

src/frontmatter.md
epage: Sorry, I had meant to already follow it but I got mixed up and did it on the wrong file (since so much commentary is on `whitespace.md`)

traviscross: Thanks.  No worries.

src/whitespace.md
epage: First off, I am not a DRY absolutist.  Things having the appearance of duplication isn't a reason to de-duplicate but we should work towards understanding the underlying principles and requirements and de-duplicate those where it works well and reducing the burden of duplication where it doesn't work well (e.g. having two definitions next to each other).  In this case, we are working off of an item (whitespace) that is defined in three categories that is then defined as literals that natural leads to a way to divide things up to not duplicate knowledge.

I see this is similar to but worse than https://github.com/rust-lang/reference/pull/1974#discussion_r2476014266

At least with `TAB`, the abstraction is thin (too thin as I suggested in that thread).  Here, we'd be duplicating a lot of the definition.  This puts a burden on the user to match up the entries to 
1. Understand that one is a superset of the other
2. Understand what the role of the items mean

When seeing duplication like this, it is natural to expect there to a reason for it and that reason to be there are differences.  For me, this makes me doubt myself and pore over something more, wasting my time and making me not confident that I understand what I'm working off of, and more annoyed in the process.

I suspect this will also be a honey trap for contributors who will see this and then submit PRs to do what I did.

src/input-format.md
```diff
@@ -59,6 +59,11 @@ This prevents an [inner attribute] at the start of a source file being removed.
 > [!NOTE]
 > The standard library [`include!`] macro applies byte order mark removal, CRLF normalization, and shebang removal to the file it reads. The [`include_str!`] and [`include_bytes!`] macros do not.
```
ehuss: What is the impact on these macros?

src/input-format.md
```diff
@@ -59,6 +59,11 @@ This prevents an [inner attribute] at the start of a source file being removed.
...
 
+r[input.frontmatter]
+## Frontmatter removal
+
+After some whitespace, [frontmatter] may next appear in the input.
```
ehuss: Can this be worded more explicitly how it relates to the items around it (particularly shebang)? This rule isn't quite standing on its own and it isn't quite clear how it fits.

For example, something like: "after the optional [shebang] and then optional [whitespace], [frontmatter] may appear next in the input".

src/whitespace.md
```diff
@@ -4,23 +4,32 @@ r[lex.whitespace]
...
-
-LF -> U+000A  // Line feed, `'\n'`
-
-CR -> U+000D  // Carriage return, `'\r'`
+      END_OF_LINE
+    | IGNORABLE_CODE_POINT
+    | HORIZONTAL_WHITESPACE
+
+END_OF_LINE ->
+      U+000A // line feed, `'\n'`
```
ehuss: Can you please keep the original comments and same comment style in these rules?

src/frontmatter.md
```diff
@@ -0,0 +1,39 @@
...
+# Frontmatter
+
+r[frontmatter.syntax]
+```grammar,lexer
+@root FRONTMATTER ->
+      FRONTMATTER_FENCE HORIZONTAL_WHITESPACE* INFOSTRING? HORIZONTAL_WHITESPACE* LF
+      (FRONTMATTER_LINE LF )*
+      FRONTMATTER_FENCE[^matched-fence] HORIZONTAL_WHITESPACE* LF
+
+FRONTMATTER_FENCE -> `---` `-`*
```
ehuss:
```suggestion
FRONTMATTER_FENCE -> `-`{3..}
```

src/SUMMARY.md
ehuss: Maybe this could be sequenced just after "Input format" since it is tightly related to that? I realize the rest of these aren't in any particularly logical order, but I'm not quite seeing why it is placed in the middle here.

src/frontmatter.md
```diff
@@ -0,0 +1,39 @@
+r[frontmatter]
+# Frontmatter
+
+r[frontmatter.syntax]
+```grammar,lexer
+@root FRONTMATTER ->
+      FRONTMATTER_FENCE HORIZONTAL_WHITESPACE* INFOSTRING? HORIZONTAL_WHITESPACE* LF
+      (FRONTMATTER_LINE LF )*
+      FRONTMATTER_FENCE[^matched-fence] HORIZONTAL_WHITESPACE* LF
```
ehuss, line 9: It's a bit unusual to have load-bearing footnotes in the grammar. Would it be possible to define this recursively so that isn't necessary? Or is there maybe some other way around it?

src/frontmatter.md
```diff
@@ -0,0 +1,39 @@
...
+[^matched-fence]: The closing fence must have the same number of `-` as the opening fence
+[^escaped-fence]: A `FRONTMATTER_FENCE` at the beginning of a `FRONTMATTER_LINE` is only invalid if it has the same or more `-` as the `FRONTMATTER_FENCE`
+
+Frontmatter is an optional section for content intended for external tools without requiring these tools to have full knowledge of the Rust grammar.
+
+r[frontmatter.document]
+Frontmatter may only be preceded by a [shebang] and whitespace.
+
+r[frontmatter.fence]
+The delimiters are referred to as a "fence." The opening and closing fences must be at the start of a line. They must be a matching pair of three or more hyphens (`-`). A fence may be followed by horizontal whitespace.
```
ehuss: We typically use emphasis for introducing terms.

```suggestion
The delimiters are referred to as a *fence*. The opening and closing fences must be at the start of a line. They must be a matching pair of three or more hyphens (`-`). A fence may be followed by horizontal whitespace.
```

src/frontmatter.md
```diff
@@ -0,0 +1,39 @@
...
+
+INVALID_FRONTMATTER_LINE_START -> (FRONTMATTER_FENCE[^escaped-fence] | LF)
+
+INVALID_FRONTMATTER_LINE_CONTINUE -> LF
+```
+
+[^matched-fence]: The closing fence must have the same number of `-` as the opening fence
+[^escaped-fence]: A `FRONTMATTER_FENCE` at the beginning of a `FRONTMATTER_LINE` is only invalid if it has the same or more `-` as the `FRONTMATTER_FENCE`
+
+Frontmatter is an optional section for content intended for external tools without requiring these tools to have full knowledge of the Rust grammar.
```
ehuss, line 30: We always use "intro" for the introduction of something.

Also, perhaps this intro could be moved to the top of the file?

```suggestion
r[frontmatter.intro]
Frontmatter is an optional section for content intended for external tools without requiring these tools to have full knowledge of the Rust grammar.
```

Can you also include an example here?

src/frontmatter.md
ehuss: Just FYI, TC and I have been trying to come up with a grammar here, but we have not yet been able to land on something. I'm still working on it.

src/frontmatter.md
epage: `CR` doesn't need to be explicit because of [CRLF Normalization](https://doc.rust-lang.org/reference/input-format.html#crlf-normalization)

src/input-format.md
epage: >  This rule isn't quite standing on its own and it isn't quite clear how it fits.

At least shebang is written in an additive manner:

> If the remaining sequence begins with the characters `#!`,

Seems like it would be good to be consistent.  I could make the additive nature more explicit.

src/whitespace.md
epage: There is some existing discussion on the comment style at https://github.com/rust-lang/reference/pull/1974#discussion_r2476064079

The intent of the comment style is to match the linked to standard for easier comparison.

src/frontmatter.md
epage: Personally, I think the existing content is easier for an end user to look at but whatever.

epage: Reverted for a sec as I re-think how I want to handle some of this feedback

src/SUMMARY.md
epage: I don't remember why I chose that spot.  I've moved it.

src/frontmatter.md
epage: > Also, perhaps this intro could be moved to the top of the file?

All other Lexical structure pages with a grammar open with the grammar.  Most (but not all) other pages I randomly sampled did the same.

epage: The into reference and an example were added

epage: CI identified two problems with the example
- A feature flag must be present to build
  - How do we handle the feature flag in the example when feature flags are not within the scope of the reference?
  - Would we be removing it from the example when this is stabilized and released?  Unfortunate that we have to wait until then
- It doesn't understand frontmatter and fails to build
  - I think it is important for examples to be practical so I made a small, runnable cargo script

src/input-format.md
epage: Waiting on rust-lang/rust#146377 to update this

src/frontmatter.md
epage, line 1: There are two categories of feedback I am concerned about for this PR
- ones I disagree with as currently stated (e.g. https://github.com/rust-lang/reference/pull/1974#discussion_r2476064079, https://github.com/rust-lang/reference/pull/1974#discussion_r2476085909)
- ones that require additional structural change which I have not had the best experience contributing to (e.g. https://github.com/rust-lang/reference/pull/1974#discussion_r2508123877)

Please keep in mind that I am not contributing to this for my own interest in the Reference (I was interested in that at one point but no longer) but to fulfill a requirement of another team.

How about an alternative route for this: this PR gets re-framed as a perma-draft that is a content dump without any editorial changes (changing the content to be as surgical as possible).  Once we're agreed to the technical content, a T-reference person can then copy the commits, make any editorial changes they wish, and post their own PR.

src/frontmatter.md
epage: I worry that being too clever in encoding this could have a negative impact on users understanding this.

Another route is to extend the grammar for inline context notes like how `*?` gets a `(non-greedy)` note and `!` gets a warning icon with `with the exception of`.

src/input-format.md
```diff
@@ -59,6 +59,11 @@ This prevents an [inner attribute] at the start of a source file being removed.
 > [!NOTE]
```
epage: From @traviscross at https://github.com/rust-lang/rust/pull/148051#issuecomment-3567383545 related to `r[input.shebang.inner-attribute]` (just above this)

> Let's please document the behavior of frontmatter removal, including how it interacts with the documented shebang removal exception

There is no change to the shebang vs attribute detection.  I've been trying to think this through and so far I've not come up with a way to include frontmatter that doesn't feel artificial (saying what doesn't affect this is an open set).

src/frontmatter.md
epage: I've gone ahead and taken the approach of treating this PR as a way to share raw content that others can take and editorialize as they wish, making all of my changes surgical and making the PR un-editable.

epage:
This is updated to reflect
- current state of `include!`
- disallowed CRs assuming https://github.com/rust-lang/rust/pull/149823 moves forward

traviscross:
By way of an update on this, I have a branch in progress following the suggestion in https://github.com/rust-lang/reference/pull/1974#discussion_r2615644035.  We've been working toward being able to express the grammar correctly and cleanly for this with the work on and merging of https://github.com/rust-lang/reference/pull/2104, https://github.com/rust-lang/reference/pull/2172, and https://github.com/rust-lang/reference/pull/2181.  The merging of https://github.com/rust-lang/reference/pull/2192, which will also help here, is what created the conflict flagged above.  I'll handle that in my branch.

PR #1975, opened 2025-08-24
Or-patterns are extending

theemathas:
This is specified in the compiler at https://github.com/rust-lang/rust/blob/f6d23413c399fb530be362ebcf25a4e788e16137/compiler/rustc_hir_analysis/src/check/region.rs#L538


traviscross (approved):

traviscross:
Thanks.

PR #1979, opened 2025-08-26
clean up and properly test temporary lifetime extension in doctests

dianne:
- Removes an unused trait and impl from the successful tests.
- Adds a use following each assignment to test lifetime extension. Previously all but the last assignment were shadowed before use, so the temporary lifetimes were untested.


traviscross (approved):

traviscross:
Thanks.

PR #1980, opened 2025-08-26
Specify lifetime extension of `pin!` and `format_args!` arguments

dianne:
Reference PR for rust-lang/rust#145838, given the `format_args!` change in rust-lang/rust#145882. cc @m-ou-se

Based on rust-lang/reference#1979; the first commit is the commit from that PR.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -474,6 +474,7 @@ let &ref x = &*&temp(); // OK
 r[destructors.scope.lifetime-extension.exprs]
 #### Extending based on expressions
 
+r[destructors.scope.lifetime-extension.exprs.extending]
 For a let statement with an initializer, an *extending expression* is an
 expression which is one of the following:
 
@@ -486,12 +487,18 @@ expression which is one of the following:
 * The final expression of an extending [block expression] except for an [async block expression].
 * The final expression of an extending [`if`] expression's consequent, `else if`, or `else` block.
 * An arm expression of an extending [`match`] expression.
+* The argument(s) to an extending [`pin!`] or [`format_args!`] [macro invocation] expression.
 
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some(&mut 3)`
 are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
+r[destructors.scope.lifetime-extension.exprs.borrow]
 The operand of any extending borrow expression has its temporary scope
 extended.
+
+r[destructors.scope.lifetime-extension.exprs.macros]
+The built-in macros [`pin!`] and [`format_args!`] create temporaries.
+Any extending [`pin!`] or [`format_args!`] [macro invocation] expression has an extended temporary scope.
 
 > [!NOTE]
 > `rustc` does not treat [array repeat operands] of extending [array] expressions as extending expressions. Whether it should is an open question.
@@ -504,6 +511,7 @@ Here are some examples where expressions have extended temporary scopes:
 
 ```rust,edition2024
 # use core::sync::atomic::{AtomicU64, Ordering::Relaxed};
+# use std::pin::pin;
 # static X: AtomicU64 = AtomicU64::new(0);
 # struct S;
 # impl Drop for S { fn drop(&mut self) { X.fetch_add(1, Relaxed); } }
@@ -527,6 +535,8 @@ let x = if true { &temp() } else { &temp() };
 //           Final exprs of `if`/`else` blocks.
 # x;
 let x = match () { _ => &temp() }; // `match` arm expression.
+# x;
+let x = pin!(&temp()); // Argument to `pin!`.
 # x;
 //
 // All of the temporaries above are still live here.
@@ -618,6 +628,7 @@ There is one additional case to be aware of: when a panic reaches a [non-unwindi
 [initialized]: glossary.md#initialized
 [interior mutability]: interior-mutability.md
 [lazy boolean expression]: expressions/operator-expr.md#lazy-boolean-operators
+[macro invocation]: macros.md#macro-invocation
 [non-unwinding ABI boundary]: items/functions.md#unwinding
 [panic]: panic.md
 [place context]: expressions.md#place-expressions-and-value-expressions
@@ -664,3 +675,6 @@ There is one additional case to be aware of: when a panic reaches a [non-unwindi
 [`match`]: expressions/match-expr.md
 [`while let`]: expressions/loop-expr.md#while-let-patterns
 [`while`]: expressions/loop-expr.md#predicate-loops
+
+[`pin!`]: std::pin::pin
+[`format_args!`]: core::format_args
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -398,11 +398,7 @@ println!("{:?}", C);
 ```
 
 r[destructors.scope.lifetime-extension.sub-expressions]
-If a [borrow][borrow expression], [dereference][dereference expression],
-[field][field expression], or [tuple indexing expression] has an extended
-temporary scope then so does its operand. If an [indexing expression] has an
-extended temporary scope then the indexed expression also has an extended
-temporary scope.
+If a [borrow], [dereference][dereference expression], [field][field expression], or [tuple indexing expression] has an extended temporary scope then so does its operand. If an [indexing expression] has an extended temporary scope then the indexed expression also has an extended temporary scope.
 
 r[destructors.scope.lifetime-extension.patterns]
 #### Extending based on patterns
@@ -479,7 +475,8 @@ For a let statement with an initializer, an *extending expression* is an
 expression which is one of the following:
 
 * The initializer expression.
-* The operand of an extending [borrow expression].
+* The operand of an extending [borrow] expression.
+* The [super operands] of an extending [super macro call] expression.
 * The operand(s) of an extending [array][array expression], [cast][cast
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
@@ -487,18 +484,15 @@ expression which is one of the following:
 * The final expression of an extending [block expression] except for an [async block expression].
 * The final expression of an extending [`if`] expression's consequent, `else if`, or `else` block.
 * An arm expression of an extending [`match`] expression.
-* The argument(s) to an extending [`pin!`] or [`format_args!`] [macro invocation] expression.
 
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some(&mut 3)`
 are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
-r[destructors.scope.lifetime-extension.exprs.borrow]
-The operand of any extending borrow expression has its temporary scope
-extended.
+r[destructors.scope.lifetime-extension.exprs.borrows]
+The operand of an extending [borrow] expression has its [temporary scope] [extended].
 
-r[destructors.scope.lifetime-extension.exprs.macros]
-The built-in macros [`pin!`] and [`format_args!`] create temporaries.
-Any extending [`pin!`] or [`format_args!`] [macro invocation] expression has an extended temporary scope.
+r[destructors.scope.lifetime-extension.exprs.super-macros]
+The [super temporaries] of an extending [super macro call] expression have their [scopes][temporary scopes] [extended].
 
 > [!NOTE]
 > `rustc` does not treat [array repeat operands] of extending [array] expressions as extending expressions. Whether it should is an open question.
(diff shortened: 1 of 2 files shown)


src/destructors.md
```diff
@@ -434,6 +434,7 @@ expression which is one of the following:
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
 * The arguments to an extending [tuple struct] or [tuple variant] constructor expression.
+* The argument(s) to an extending [`pin!`] or [`format_args!`] [macro invocation] expression.
```
dianne: Ideally this should be clearer about the format arguments being extended, not the format string (which doesn't need extension), but I've had trouble including that detail without breaking the flow/clarity of the section

src/destructors.md
```diff
@@ -434,29 +434,37 @@ expression which is one of the following:
...
 * The arguments to an extending [tuple struct] or [tuple variant] constructor expression.
+* The argument to an extending [`pin!`] or the value arguments to an extending [`format_args!`] [macro invocation] expression.
 * The final expression of any extending [block expression].
 
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some(&mut 3)`
 are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
 The operand of any extending borrow expression has its temporary scope
 extended.
+Likewise, the argument to any extending [`pin!`] macro invocation expression and the value arguments to any extending [`format_args!`] macro invocation expression have their temporary scopes extended.
```
dianne: Took a second pass at this to try and distinguish the operands being formatted from the format string for `format_args!`, but I can't find consistent terminology to refer to them by. I went with ["value argument"](https://doc.rust-lang.org/nightly/std/fmt/index.html#positional-parameters), which distinguishes them from the "format arguments" in the format string, but ["format argument"](https://doc.rust-lang.org/nightly/std/fmt/index.html#precision) may also be used for the values, and ["formatting argument"](https://doc.rust-lang.org/nightly/std/macro.format_args.html#lifetime-limitation) is used elsewhere, ambiguously. Calling them "value arguments" also feels a bit strange when they're borrowed as part of the macro expansion, but it at least feels less ambiguous in what it refers to than the alternatives I could find. "Operands" might also be clearer, but I'm not sure anything else refers to macro arguments in that way; the `fmt` docs certainly don't, and the Reference describes macros from the lang perspective (in terms of operating on token trees), rather than catering to their function-like uses in the standard library API, so there's no notion of comma-separated arguments at all.

My hope is that linking to the stdlib in these cases should make it clear that `pin!` and `format_args!` have function-like syntax, making the mention of "arguments" clear in context, but it feels a bit messy.

src/destructors.md
```diff
@@ -442,21 +442,33 @@ are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 The operand of any extending borrow expression has its temporary scope
 extended.
 
+> [!NOTE]
+> [Invocations][macro invocation] of the built-in macros [`pin!`] and [`format_args!`] expand to borrow their arguments as temporaries.
+> They are subject to lifetime extension as follows:
+> * The argument to an extending [`pin!`] macro invocation expression is extending and has an extended temporary scope.
+> * Other than the format string, the arguments to an extending [`format_args!`] macro invocation expression are extending and have extended temporary scopes.
+
```
dianne, line 501: Took a third pass at this, inspired by https://github.com/rust-lang/rust/pull/145838#issuecomment-3222740561 (cc @traviscross). The primary goal here is to separate the rules for macro expansions out from the core syntactic rules, while keeping them within `destructors.scope.lifetime-extension.exprs` to make it possible to link to the rules as a whole in library documentation (e.g. in rust-lang/rust#145943). As a secondary goal, I feel it's helpful context to mention that these macros expand to borrow their arguments; doing so in a note is less harmful to the flow of the section.

theemathas: If `format_args!` is given a place expression that isn't a temporary (for example, a variable name), then the macro will *not* create a new temporary. It will just borrow from that existing place. This is different from the behavior of `pin!()`, which always creates a new temporary.

dianne: Oh, oops! I had already fixed that but completely forgot to commit it before pushing. Thanks for the catch!

The new wording is still muddy, since they are different as you point out (with `format_args!` borrowing, then assigning to a temporary, whereas `pin!` moves into a temporary, then borrows that). I'll think about what I can do there. I'd rather not get too into the weeds on how they expand. Maybe it's sufficient to say they "expand to create temporaries", and note that they're extended when the macro invocation expression is extending?

dianne: Tried something like that here: [(diff)](https://github.com/rust-lang/reference/compare/6c07751473b937ebc21ac2e67982c5e08d2b81d6..1c9c73c3ef78c3d66e794de34c2ce3f24b867708). I think this is more correct, but I'm not happy with the formatting yet. It feels like maybe it should go back to being integrated with the other lifetime extension rules, rather than being a separate note. The detail that `pin!` and `format_args!` create temporaries does feel appropriate to a note though, and that's key to specifying that those temporaries have extended scopes.

src/destructors.md
```diff
@@ -434,29 +435,41 @@ expression which is one of the following:
...
 So the borrow expressions in `&mut 0`, `(&1, &mut 2)`, and `Some(&mut 3)`
 are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
+r[destructors.scope.lifetime-extension.exprs.borrow]
 The operand of any extending borrow expression has its temporary scope
 extended.
 
+r[destructors.scope.lifetime-extension.exprs.macros]
+The built-in macros [`pin!`] and [`format_args!`] create temporaries.
+Any extending [`pin!`] or [`format_args!`] [macro invocation] expression has an extended temporary scope.
```
dianne: Fourth pass: I've broken up `destructors.scope.lifetime-extension.exprs` to match `destructors.scope.lifetime-extension.patterns` and to put the rule for built-in macros' temporaries in a subsection. I've also moved the rule for arguments' extension back into the the newly-delimited rule `destructors.scope.lifetime-extension.exprs.extending`. I'm not satisfied with the wording yet, but structurally I think it's an improvement.

I'm doing a bit of conflation here. The "temporaries" here are both:
- `super let` bindings; since they have (extended) temporary scopes, I feel like referring to them as "temporaries" is most fitting for the moment.
- The borrowed temporaries created when a value expression is passed to `format_args!`.

Let me know if it needs further clarification. My hope is that it's a suitable level of detail for how these macros behave, to avoid specifying their exact expansion.

src/destructors.md
traviscross: I'll push a revision that will clarify this.

src/destructors.md
traviscross: My reading is that this rule may not be needed at all.  The "extending based on expressions" section states a recursive algorithm for determining, starting on the outside and working inward, which expressions are extending.  This definition then serves the following rule, which is the one that has language effect:

> The operand of any extending borrow expression has its temporary scope extended.

That's the whole game, right there, I think, is deciding whether a particular borrow expression is extending.

In that context, then, the only thing that matters with respect to these built-in macros is whether expressions in their argument positions are extending expressions.

(I'll push a revision to drop this rule.)

traviscross (approved):

src/destructors.md
dianne: > In that context, then, the only thing that matters with respect to these built-in macros is whether expressions in their argument positions are extending expressions.

Unless I'm missing something, I don't think that's sufficient to describe the behavior of `pin!` and `format_args!`. I was struggling to write out the missing rule though, since it can't quite be expressed precisely with stable terminology: the bindings of a `super let` statement in an extending block expression have their scopes extended[^1]. In that way, `super let` bindings are scoped like borrowed temporaries. The compiler implementation is spread across [here](https://github.com/rust-lang/rust/blob/12eb345e5593a10e61b4aef766076c8bd552fea6/compiler/rustc_hir_analysis/src/check/region.rs#L633-L644) and [here](https://github.com/rust-lang/rust/blob/12eb345e5593a10e61b4aef766076c8bd552fea6/compiler/rustc_hir_analysis/src/check/region.rs#L471-L483).

This can be observed through `pin!` and `format_args!`:

In `pin!($expr)`, the result of `$expr` is moved into a `super let` binding, giving it the same scope a borrowed temporary would have: if the `pin!` invocation is extending, its scope is extended, and otherwise, it's dropped in the enclosing temporary scope. Since `pin!` moves its argument, this can't simply be described as it borrowing it, but the `Pin` doesn't own it either. To enforced pinnedness, it has to treat its argument as a value, but it also needs to scope it like it's borrowed; this (as I understand it) is why `super let` is needed in Rust 2024[^2].

`format_args!` does borrow its arguments, but `super let`'s unique scoping can be observed there too: even if `format_args!` is borrowing from long-lived places, the `super let` bindings created to store the arguments have the scope a borrowed temporary would: they have extended scopes if the `format_args!` invocation is extending, and otherwise they're dropped in the enclosing temporary scope.

[^1]: Arguments to extending `pin!` and `format_args!` invocations being extending covers a separate `super let` property: the initializer of `super let` in an extending block is extending. This is what rust-lang/rust#145838 affects.
[^2]: Likewise the reason `Pin { __pinned: &mut { $expr } }` works is it forces `$expr` to evaluate to a temporary in the appropriate scope (on account of fields and block tails of extending struct and block expressions being extending). Effectively, the `super let`-based implementation captures that property without the block tail scope being a problem in Rust 2024.

traviscross: Right.  The mental model that I think is correct is that `pin!($expr)` is, in this regard, exactly like `&pin mut $expr`, which is to say that the argument/operand is a place expression context, that the operand is an extending expression when the borrow is, and that the operand of such an extending borrow has its temporary scope extended.

If that's right, the cleanest way I can think of to express this is to create a concept of a "borrow macro call expression", and then to reframe the rules for extending based on expressions to work with these.

Let me know if that looks right.

dianne: With regard to `pin!`, I think the operand is a value expression. In the [current implementation](https://github.com/rust-lang/rust/blob/beeb8e3af54295ba494c250e84ecda4c2c5d85ff/library/core/src/pin.rs#L1955-L1961), the `super let mut pinned = $value;` binds the operand by value, effectively evaluating it to a temporary[^2]. In the [old implementation](https://github.com/rust-lang/rust/blob/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/pin.rs#L2018) it uses a block tail to force a value expression context. Functionally, `pin!` has to move out of its operand, to ensure that its operand can't be moved after being pinned: https://doc.rust-lang.org/nightly/std/pin/macro.pin.html#remarks. If its operand was just borrowed, the place would be able to be moved from after the `Pin` was no longer in use, violating `Pin`'s invariant.

`format_args!`'s non-format-string arguments are place expressions and implicitly borrowed, so the inclusion as a borrow macro call expression works[^1]. I don't think it's sufficient to describe its behavior though, since the returned `fmt::Arguments` also borrows from temporaries created by `format_args!`, which may have shorter scopes than its operands [(playground)](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=aeab9c18f9e0077abfa3564ff4281fee).

In both cases, the best I was able to come up with was that `pin!` and `format_args!` themselves create temporaries[^3], the scopes of which can be extended[^4]. This is especially clear in the old implementation of `pin!`: in `Pin { __pointer: &mut { $value } }`, `{ $value }` evaluates to a temporary that may be extended because it's the operand of a borrow expression.

[^1]: A wording complexity though: `format_args!("{x}")` borrows `x` despite it not appearing "after" the format string.
[^2]: Technically, the initializer of a `super let` is a place expression, but then creating the `mut pinned` binding moves out of that place, effectively treating it like it like a value expression.
[^3]: Currently implemented as `super let` bindings.
[^4]: Because the bindings of a `super let` in an extending block have extended scopes.

traviscross: Right.  This is I think the key:

> I was struggling to write out the missing rule though, since it can't quite be expressed precisely with stable terminology...

It's just too hard to be precise here without introducing terms, so I've pushed a revision that does go ahead and introduce new terms.

Let me know what you think.

dianne: This looks right to me[^1]. I think introducing terms is the right call; that helped a lot with precision and clarity. Thanks for writing that out!

[^1]: Though I think one of the examples is left over from an old revision? I'll leave a comment on it.

src/destructors.md
```diff
@@ -528,6 +530,12 @@ let x = if true { &temp() } else { &temp() };
 # x;
 let x = match () { _ => &temp() }; // `match` arm expression.
 # x;
+let x = pin!(temp()); // Operand of borrow macro call expression.
+# x;
+# // TODO: Simplify after this PR lands:
+# //       <https://github.com/rust-lang/rust/pull/145882>.
+let x = format_args!("{:?}{:?}", (), temp()); // As above.
+# x;
```
dianne:
```suggestion
let x = pin!(&temp()); // Super operand of super macro call expression.
# x;
# // TODO: Simplify after this PR lands:
# //       <https://github.com/rust-lang/rust/pull/145882>.
let x = format_args!("{:?}{:?}", (), &temp()); // As above.
# x;
```

Besides the terminology update, my reasoning here is to highlight the super operands being extending, rather than the super temporaries' scopes being extended.

traviscross: Thanks, fixed.  I actually like the `pin!(temp())` because of the more direct analogy to `&pin mut temp()`, but I added another example that makes the other point as well.

traviscross:
I've updated this to account for

- https://github.com/rust-lang/rust/pull/145882

not having merged yet so that we can merge this.

PR #1981, opened 2025-08-26
Specify lifetime extension of `match` arms and `if` consequent/`else` block tail expressions

dianne:
This is present in the compiler [here](https://github.com/rust-lang/rust/blob/5ab69249f36678c0a770a08d3d1b28a8103349ff/compiler/rustc_hir_analysis/src/check/region.rs#L605-L606), but was not yet present in the Reference.

Based on https://github.com/rust-lang/reference/pull/1979; the first commit is the commit from that PR.


traviscross (approved):

traviscross:
Great. Thanks.

PR #1988, opened 2025-09-04
Enable folding of chapter listing in navigation sidebar

traviscross:
We have many chapters and subchapters.  Listing all of the subchapters in the navigation sidebar by default can make it difficult to see the high level structure of the Reference and can inhibit rapid navigation based on that hierarchical structure.

Fortunately, mdbook has an option to fold the chapter listing.  This causes only the top-level chapters to be shown by default while giving the user the ability to unfold individual chapters.  Let's turn on that option.

cc @ehuss

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -10,6 +10,10 @@ git-repository-url = "https://github.com/rust-lang/reference/"
 edit-url-template = "https://github.com/rust-lang/reference/edit/master/{path}"
 smart-punctuation = true
 
+[output.html.fold]
+enable = true
+level = 0
+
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Overall I think I like this, though one of the things we talked about is there is a slight regression in being able to search/jump to a specific chapter if you know the name of the chapter (and using the browser's "find" command). mdbook's search engine fails in a bunch of cases. We talked about there might be some low-hanging fruit to improve the search engine, but larger changes might be more challenging. Perhaps this will motivate someone to improve the search engine.

PR #1989, opened 2025-09-04
Link out to the notation from grammar summary

epage:
As a reference, users are likely to jump in at random points and it can then be hard to know where to look for information on grammars.

While it would likely be overkill to link to the notation wherever there is a grammar, the grammary summary page seems like a reasonable place to at least include a link.

### DIFF
--- a/src/input-format.md
+++ b/src/input-format.md
@@ -1,6 +1,19 @@
 r[input]
 # Input format
 
+r[input.syntax]
+```grammar,lexer
+@root CHAR -> <a Unicode scalar value>
+
+NUL -> U+0000
+
+TAB -> U+0009
+
+LF -> U+000A
+
+CR -> U+000D
+```
+
 r[input.intro]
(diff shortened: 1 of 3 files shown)


traviscross (approved):

traviscross:
Thanks Ed.

src/input-format.md
```diff
@@ -1,6 +1,19 @@
...
+@root CHAR -> <a Unicode scalar value>
+
+NUL -> U+0000
+
+TAB -> U+0009
+
+LF -> U+000A
+
+CR -> U+000D
+```
```
epage, line 15: This feels a bit odd to reference this as the input syntax (and put it here.  I would have expected `input.syntax` to be something like
```grammar
@root INPUT -> SHEBANG? (BLANK_LINES FRONTMATTER)? TOKEN*
```
`BLANK_LINES` might refer to part of this but not all of it.

PR #1991, opened 2025-09-04
Create Whitespace grammar productions

epage:
This does not create any new productions, instead preferring comments.
rust-lang/reference#1974 will involve pulling out the horizontal
whitespace into a separate production.

Comment wording (and casing) is modeled off of
https://www.unicode.org/reports/tr31/#R3a.
I left off a "unicode" prefix for ASCII items as they are likely common
enough in that context that specifying them as "unicode" could cause
more confusion.

### DIFF
--- a/src/whitespace.md
+++ b/src/whitespace.md
@@ -1,21 +1,34 @@
 r[lex.whitespace]
 # Whitespace
 
+r[whitespace.syntax]
+```grammar,lexer
+@root WHITESPACE ->
+    // end of line
+      LF
+    | U+000B // vertical tabulation
+    | U+000C // form feed
+    | CR
+    | U+0085 // Unicode next line
+    | U+2028 // Unicode LINE SEPARATOR
+    | U+2029 // Unicode PARAGRAPH SEPARATOR
+    // Ignorable Code Point
+    | U+200E // Unicode LEFT-TO-RIGHT MARK
+    | U+200F // Unicode RIGHT-TO-LEFT MARK
+    // horizontal whitespace
+    | TAB
+    | U+0020  // space ' '
+
+TAB -> U+0009  // horizontal tab ('\t')
+
+LF -> U+000A  // line feed ('\n')
+
+CR -> U+000D  // carriage return ('\r')
+```
+
 r[lex.whitespace.intro]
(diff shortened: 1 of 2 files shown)


src/whitespace.md
```diff
@@ -1,21 +1,62 @@
...
+END_OF_LINE ->
+      LINE_FEED
+    | VERTICAL_TAB
+    | FORM_FEED
+    | CARRIAGE_RETURN
+    | NEXT_LINE
+    | LINE_SEPARATOR
+    | PARAGRAPH_SEPARATOR
+
+LINE_FEED -> U+000A
```
traviscross: Rather than having duplicate productions like this (that we wouldn't want people to use elsewhere in the Reference), it'd perhaps work out better to simply comment the `LF` production with something like `// Unicode character "LINE FEED (LF)".`, so I've added support for comments in the grammar:

- https://github.com/rust-lang/reference/pull/1993

epage: Thanks!

> that we wouldn't want people to use elsewhere in the Reference

In that case, maybe I should remove all productions that are not used elsewhere.

I also went with comment wording (and casing) to align with the unicode spec as much as possible.

ehuss (approved):
Thanks!

traviscross (approved):

traviscross:
Thanks @epage.

epage:
Why was this reformatted? The change was to match a change I made to rustc to communicate intent, match the unicode standard we reference, and prep for pulling out horizontal whitespace.

traviscross:
Could you perhaps say more about what specifically is missing in the revision with respect to matching what's in rustc, communicating intent, referencing the standard, prepping for pulling out horizontal whitespace, etc.?

epage:
The list was grouped in the unicode standard groups, ordered like it, and with comments that matched.

epage:
Also, all of my 3 reference PRs, all of which have been merged in the last month, have had direct edits made instead of engaging with me. This is the second time the edits were questionable and would have been helped to have been discussed first, even if we still go with the edits in the end. For the other, see https://github.com/rust-lang/reference/pull/1989#discussion_r2325943918

PR #1992, opened 2025-09-04
Document temporary scoping for destructuring assignments

dianne:
Pursuant to https://github.com/rust-lang/rust/pull/145838#issuecomment-3239833028, I've added a note and some examples detailing how destructuring assignment expressions effectively introduce two scopes for its temporaries, whereas basic assignment expressions don't introduce any.

I've included it alongside the description of destructuring assignment desugaring, but I could see it being at home in the destructors chapter as well. I also considered adding the assigned value operand of a destructuring assignment operand to the list of temporary scopes, but it's a bit more complicated than that since it has two scopes, so I figured it would be simplest to talk directly in terms of the desugaring, where the two scopes are visible.

I've used borrowck errors to demonstrate scoping in the tests since the test scaffolding is much simpler in that case than for testing runtime drop order, but a runtime drop order comparison may help further illustrate it.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -266,6 +266,9 @@ smallest scope that contains the expression and is one of the following:
 > [!NOTE]
 > The [scrutinee] of a `match` expression is not a temporary scope, so temporaries in the scrutinee can be dropped after the `match` expression. For example, the temporary for `1` in `match 1 { ref mut z => z };` lives until the end of the statement.
 
+> [!NOTE]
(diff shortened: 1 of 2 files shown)


traviscross:
Revised.  @dianne, does this look correct?

dianne:
Looks good to me. Thanks!

traviscross (approved):

traviscross:
Thanks @dianne.

PR #1993, opened 2025-09-07
Add support to grammar for single line comments

traviscross:
Sometimes it would be useful in the grammar to be able to add comments inline.  Let's allow this.

We'll treat these comments as expressions and add special handling to allow these on lines ahead of the production name.

For now we'll render these only in the Markdown and not in the railroad diagrams.

---

This is best reviewed commit by commit.

cc @ehuss

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -27,7 +27,9 @@ BACKTICK -> U+0060
 
 LF -> U+000A
 
-Production -> `@root`? Name ` ->` Expression
+Production ->
+    ( Comment LF )*
+    `@root`? Name ` ->` Expression
 
 Name -> <Alphanumeric or `_`>+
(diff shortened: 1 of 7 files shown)


ehuss (approved):
Thanks!

A remark I have is that there is pretty broad inconsistency on how different groups and tools define comments. The ones I'm aware of are:

* IETF — `;`
* ISO — `(* ... *)`
* W3C — `;` and `/* .. */`
* ANTLR, Bison — `//` and `/* .. */`

Most grammars for languages that I've seen do not have comments.

Staying familiar with Rust syntax is probably the best choice here.

I'll also mention that it now becomes a little unclear if certain comments should be attached with a footnote or a comment. We'll see how that plays out.

PR #1996, opened 2025-09-14
Proposal of a new Concurrency chapter

borsakv:
cc @joshtriplett 

**Explanation:**
After analyzing the FLS, I found that there were various chapters within the FLS that are missing from the Rust Reference. Most of the information in the said chapters was present in the Reference, just scattered. However, the information from the [Concurrency Chapter in the FLS](https://rust-lang.github.io/fls/concurrency.html) seemed to be missing almost entirely, due to this I wrote the named Concurrency chapter with the help of @joshtriplett under the [Rust 2025H2 Reference Expansion Goal](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/reference-expansion.md).

I believe this chapter covers stable concepts that will not change often and is an important addition for Rust users to reference when they are working with concurrency.

**Changes:**
Added the initial proposal of the Concurrency Chapter with the following subheadings:
- Send and Sync
- Atomics
- Asynchronous Computation

Also added new entries to the Glossary:
- Atomic types
- Data race
- Marker trait

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -30,6 +30,10 @@ items are defined in [implementations] and declared in [traits]. Only
 functions, constants, and type aliases can be associated. Contrast to a [free
 item].
 
+### Atomic types
+
+Atomic types provide primitive shared-memory communication between threads, and are the building blocks of other concurrent types.
+
 ### Blanket implementation
 
 Any implementation where a type appears [uncovered](#uncovered-type). `impl<T> Foo
@@ -59,6 +63,10 @@ library crates, called external crates. A crate has a self-contained tree of
 may be made visible to other crates by marking them as public in the crate
 root, including through [paths] of public modules.
 [More][crate].
+
+### Data race
+
+A data race occurs when multiple threads attempt to access the same shared memory location concurrently.
 
 ### Dispatch
 
@@ -136,6 +144,12 @@ A `struct`, `enum`, or `union` which was defined in the current crate.
 This is not affected by applied type arguments. `struct Foo` is considered local, but
 `Vec<Foo>` is not. `LocalType<ForeignType>` is local. Type aliases do not
 affect locality.
+
+### Marker trait
+A trait that has no associated items (no methods, no associated types, and no constants)
+and is used only to mark that a type has a particular property.
+Implementing a marker trait does not change a type’s behavior at runtime;
+it simply communicates to the compiler and to other code that the type satisfies some condition.
 
 ### Module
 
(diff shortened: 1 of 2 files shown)


traviscross:
Thanks for your efforts on this chapter.

@ehuss and I reviewed and talked about this chapter today.  Tough one.  There's a lot of good here.  And it's good to cover places where the FLS has a chapter but the Reference doesn't.

At the same time, there are some questions and challenges here.  The section on "combining `async` and `unsafe`" probably just needs to be deleted -- this is covered elsewhere in the Reference -- and this may be true of some of the other content.

That said, I realize what this chapter may be trying to do is to "put things in one place".  There may be some merit in that, but in that same way, this chapter reads in places a bit like a tutorial rather than Reference material.

Similarly, one of the longstanding questions we've had on the Reference is how much of the standard library documentation we want to duplicate, in what cases we should do that, how to go about it, etc.  In its coverage of `Send`, Sync`, atomics, etc., this chapter dives right into that challenge.

Probably I question whether the async bits belong in the same chapter as the discussion of `Send`, `Sync`, atomics, etc., in the sense that parallelism and concurrency are rather different.

There's a lot more to untangle here, but the main point we want to get across is this meta one:

It'd probably be better if you were to have a conversation with us first before writing a chapter like this about how to approach it.  Maybe file an issue here with an outline of what you have in mind or open a thread on our Zulip channel and ping us.  That's likely to be more efficient than trying to work out, with a full chapter here in front of us, what can be accepted and what maybe-substantial changes will need to be made to the rest.

src/glossary.md
```diff
@@ -60,6 +64,10 @@ may be made visible to other crates by marking them as public in the crate
 root, including through [paths] of public modules.
 [More][crate].
 
+### Data race
+
+A data race occurs when multiple threads attempt to access the same shared memory location concurrently.
```
RalfJung, line 69: (I realize this is currently held up in different discussions, but just for posterity:)

This is *not* the typical definition of a data race in a language using the C++ memory model, such as Rust. The definition given here is close to the definition of a "race condition", but not quite the right one for that either:
- A race condition occurs when the same location is accessed concurrently by two threads, *and at least one of the accesses is a write*
- A data race is a race condition where at least one access is non-atomic

More broadly, this text is mixing up opsem concerns and library concerns, making it quite confusing to me. But I'll defer that discussion to until after the scope question is resolved.

chorman0773: Race condition isn't quite the right word here, because a race condition can have more general meeting than the narrow constraint of a data race. 
Also, as a nit, C++ defines a data race as when at least one of the accesses is a side effect (which is either a write, or any volatile access).

PR #1999, opened 2025-09-16
Clarify that omitting `nostack` is a promise from the compiler to the programmer

jyn514:
cc @Amanieu

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -1226,7 +1226,7 @@ let _: () = unsafe {
 
 r[asm.options.supported-options.nostack]
 - `nostack`: The assembly code does not push data to the stack, or write to the stack red-zone (if supported by the target).
-  If this option is *not* used then the stack pointer is guaranteed by the compiler to be suitably aligned (according to the target ABI) for a function call at the start of the assembly code.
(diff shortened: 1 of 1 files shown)


src/inline-assembly.md
```diff
@@ -1184,7 +1184,7 @@ let _: () = unsafe {
 
 r[asm.options.supported-options.nostack]
 - `nostack`: The assembly code does not push data to the stack, or write to the stack red-zone (if supported by the target).
-  If this option is *not* used then the stack pointer is guaranteed to be suitably aligned (according to the target ABI) for a function call.
+  If this option is *not* used then the stack pointer is guaranteed by the compiler to be suitably aligned (according to the target ABI) for a function call at the start of the assembly code.
```
jyn514: > at the start of the assembly code

this makes me feel weird — i suspect that inline asm isn't allowed to modify the stack pointer at all? but i'm not sure if that's true, i don't want to write it in the reference without more background info.

Amanieu: There's already a requirement that the stack pointer must be restored to its original value before returning from an asm block. This applies regardless of whether `nostack` is used.

Amanieu (approved):

traviscross (approved):

PR #2001, opened 2025-09-16
More clearly define the default abi of an extern block

ehuss:
This attempts to make it clearer that if the ABI string is not specified, it defaults to the "C" ABI. I felt like the existing text didn't emphasize that enough.

I also took the opportunity to reword the intro to more generally explain what the ABI string does.

A followup PR will further clarify some of the confusing use of the word "default" in the specific ABI definitions.

This is implemented in [`ExternAbi::FALLBACK`](https://github.com/rust-lang/rust/blob/a9d0a6f15533a364816c4d81e2192009ef601d33/compiler/rustc_abi/src/extern_abi.rs#L287)

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -98,14 +98,19 @@ r[items.extern.abi]
 ## ABI
 
 r[items.extern.abi.intro]
-By default external blocks assume that the library they are calling uses the
-standard C ABI on the specific platform. Other ABIs may be specified using an
-`abi` string, as shown here:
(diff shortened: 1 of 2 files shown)


src/items/external-blocks.md
```diff
@@ -98,14 +98,19 @@ r[items.extern.abi]
...
-// Interface to the Windows API
-unsafe extern "system" { }
-```
+> [!EXAMPLE]
+> ```rust
+> // Interface to the Windows API
+> unsafe extern "system" { }
+> ```
+
+r[iterms.extern.abi.default]
```
tshepang:
```suggestion
r[items.extern.abi.default]
```

traviscross (approved):

PR #1987, opened 2025-08-30
RISC-V Extensions update including 29 extensions to stabilize

a4lg:
This commit directly corresponds to rust-lang/rust#145948.
See that proposal for criteria of 29 new extensions.

## Notes (for the Rust reference)

*   Updated existing `a` and `c` to imply its subset(s).
    *   `a` equals `zalrsc` and `zaamo` combined (imply both from `a`).
    *   The C extension always implies the Zca extension and conditionally implies Zcf and/or Zcd extensions (on RISC-V) but
        `c` (Rust target feature) only implies `zca` as both `zcf` and `zcd` are not to be stabilized on this proposal.
*   For each extension, linked to the first document containing it with the ratified state.
    For instance, all extensions present in the RVA23 profiles link to (3) (despite that they are present in (2)) because they were not ratified at the time when (2) is published.

## References

1.  RISC-V Instruction Set Manual (version 20250508)
    <https://github.com/riscv/riscv-isa-manual/tree/20250508>
2.  RISC-V Profiles (version 1.0 - RVA23 profiles were not ratified at the time)
    <https://github.com/riscv/riscv-profiles/tree/v1.0>
3.  RISC-V Profiles (RVA23/RVB23-ratified version)
    <https://github.com/riscv/riscv-profiles/tree/rva23-rvb23-ratified>

PR #2002, opened 2025-09-17
RISC-V: Extension Updates (including document references)

a4lg:
This PR updates documentation links and fixes minor issues:

1.  Non-existent `zb` is removed (`b` exists but not stabilized yet).
2.  Wrong link for the `zbkc` target feature is fixed.
3.  Implication from the `zbc` target feature is updated (reflects the behavior of rust-lang/rust#140139).

Note that this PR itself does not attempt to stabilize any of new extensions (target features) but makes the documentation more solid.

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -503,7 +503,6 @@ Feature     | Implicitly Enables  | Description
 `a`         |                     | [A][rv-a] --- Atomic instructions
 `c`         |                     | [C][rv-c] --- Compressed instructions
 `m`         |                     | [M][rv-m] --- Integer Multiplication and Division instructions
-`zb`        | `zba`, `zbc`, `zbs` | [Zb][rv-zb] --- Bit Manipulation instructions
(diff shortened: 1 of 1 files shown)


a4lg:
Note that, this PR is separated from rust-lang/reference#1987 (which includes stabilization of new extensions) for early adoption of the non-stabilization part.
~~The next submission of~~ rust-lang/reference#1987 ~~will be~~ *is now* based on this PR.

ehuss:
@Amanieu can you take a look at this?
Also ping @taiki-e since you approved #2005. I'm not sure if you're interested in reviewing these.

We need an appropriate group of people to ping for RISC-V.

a4lg:
@ehuss This is a strict subset of #1987 (which contains all of this PR (4 commits) + one stabilization commit) and it seems @Amanieu approved that superset.

So IIUC, @Amanieu effectively approved all of the changes here.

Amanieu (approved):

taiki-e (approved):

ehuss (approved):
Thanks!

PR #2003, opened 2025-09-17
Rework the definitions of the extern ABIs

### LINKED ISSUE #1946: Description of "aapcs" ABI is wrong

ehuss:
This updates the definitions of the extern ABIs to try to make them a little clearer as to what they mean.

This also adds new statements about the equivalence of some ABIs to the "C" ABI.

Fixes https://github.com/rust-lang/reference/issues/1946.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -141,7 +141,10 @@ r[items.extern.abi.cdecl]
   * Corresponds to MSVC's `__cdecl` and GCC and clang's `__attribute__((cdecl))`.
 
   > [!NOTE]
-  > See <https://learn.microsoft.com/en-us/cpp/cpp/cdecl> and <https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl> for more information.
+  > For details, see:
+  >
+  > - <https://learn.microsoft.com/en-us/cpp/cpp/cdecl>
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl>
 
 r[items.extern.abi.stdcall]
 * `unsafe extern "stdcall"` --- The calling convention typically used by the [Win32 API] on x86_32.
@@ -149,7 +152,10 @@ r[items.extern.abi.stdcall]
   * Corresponds to MSVC's `__stdcall` and GCC and clang's `__attribute__((stdcall))`.
 
   > [!NOTE]
-  > See <https://learn.microsoft.com/en-us/cpp/cpp/stdcall> and <https://en.wikipedia.org/wiki/X86_calling_conventions#stdcall> for more information.
+  > For details, see:
+  >
+  > - <https://learn.microsoft.com/en-us/cpp/cpp/stdcall>
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#stdcall>
 
 r[items.extern.abi.win64]
 * `unsafe extern "win64"` --- The Windows x64 ABI.
@@ -158,7 +164,10 @@ r[items.extern.abi.win64]
   * Corresponds to GCC and clang's `__attribute__((ms_abi))`.
 
   > [!NOTE]
-  > See <https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions> and <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_x64_calling_convention> for more information.
+  > For details, see:
+  >
+  > - <https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions>
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_x64_calling_convention>
 
 r[items.extern.abi.sysv64]
 * `unsafe extern "sysv64"` --- The System V ABI.
@@ -167,7 +176,10 @@ r[items.extern.abi.sysv64]
   * Corresponds to GCC and clang's `__attribute__((sysv_abi))`.
 
   > [!NOTE]
-  > See <https://wiki.osdev.org/System_V_ABI> or <https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI> for more information.
+  > For details, see:
+  >
+  > - <https://wiki.osdev.org/System_V_ABI>
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI>
 
 r[items.extern.abi.aapcs]
 * `unsafe extern "aapcs"` --- The soft-float ABI for ARM.
@@ -176,7 +188,9 @@ r[items.extern.abi.aapcs]
   * Corresponds to clang's `__attribute__((pcs("aapcs")))`.
 
   > [!NOTE]
-  > See [Arm Procedure Call Standard](https://developer.arm.com/documentation/107656/0101/Getting-started-with-Armv8-M-based-systems/Procedure-Call-Standard-for-Arm-Architecture--AAPCS-) for more information.
+  > For details, see:
+  >
+  > - [Arm Procedure Call Standard](https://developer.arm.com/documentation/107656/0101/Getting-started-with-Armv8-M-based-systems/Procedure-Call-Standard-for-Arm-Architecture--AAPCS-)
 
 r[items.extern.abi.fastcall]
 * `unsafe extern "fastcall"` --- A "fast" variant of stdcall that passes some arguments in registers.
@@ -184,7 +198,10 @@ r[items.extern.abi.fastcall]
   * Corresponds to MSVC's `__fastcall` and GCC and clang's `__attribute__((fastcall))`.
 
   > [!NOTE]
-  > See <https://learn.microsoft.com/en-us/cpp/cpp/fastcall> and <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_fastcall> for more information.
+  > For details, see:
+  >
+  > - <https://learn.microsoft.com/en-us/cpp/cpp/fastcall>
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_fastcall>
 
 r[items.extern.abi.thiscall]
 * `unsafe extern "thiscall"` --- The calling convention typically used on C++ class member functions on x86_32 MSVC.
@@ -192,7 +209,10 @@ r[items.extern.abi.thiscall]
   * Corresponds to MSVC's `__thiscall` and GCC and clang's `__attribute__((thiscall))`.
 
   > [!NOTE]
-  > See <https://en.wikipedia.org/wiki/X86_calling_conventions#thiscall> and <https://learn.microsoft.com/en-us/cpp/cpp/thiscall> for more information.
+  > For details, see:
+  >
+  > - <https://en.wikipedia.org/wiki/X86_calling_conventions#thiscall>
+  > - <https://learn.microsoft.com/en-us/cpp/cpp/thiscall>
 
 r[items.extern.abi.efiapi]
 * `unsafe extern "efiapi"` --- The ABI used for [UEFI] functions.
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -138,6 +138,7 @@ There are also some platform-specific ABI strings:
 r[items.extern.abi.cdecl]
 * `unsafe extern "cdecl"` --- The calling convention typically used with x86_32 C code.
   * Only available on x86_32 targets.
+  * Corresponds to MSVC's `__cdecl` and GCC and clang's `__attribute__((cdecl))`.
 
   > [!NOTE]
   > See <https://learn.microsoft.com/en-us/cpp/cpp/cdecl> and <https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl> for more information.
@@ -145,6 +146,7 @@ r[items.extern.abi.cdecl]
 r[items.extern.abi.stdcall]
 * `unsafe extern "stdcall"` --- The calling convention typically used by the [Win32 API] on x86_32.
   * Only available on x86_32 targets.
+  * Corresponds to MSVC's `__stdcall` and GCC and clang's `__attribute__((stdcall))`.
 
   > [!NOTE]
   > See <https://learn.microsoft.com/en-us/cpp/cpp/stdcall> and <https://en.wikipedia.org/wiki/X86_calling_conventions#stdcall> for more information.
@@ -153,6 +155,7 @@ r[items.extern.abi.win64]
 * `unsafe extern "win64"` --- The Windows x64 ABI.
   * Only available on x86_64 targets.
   * "win64" is the same as the "C" ABI on Windows x86_64 targets.
+  * Corresponds to GCC and clang's `__attribute__((ms_abi))`.
 
   > [!NOTE]
   > See <https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions> and <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_x64_calling_convention> for more information.
@@ -161,6 +164,7 @@ r[items.extern.abi.sysv64]
 * `unsafe extern "sysv64"` --- The System V ABI.
   * Only available on x86_64 targets.
   * "sysv64" is the same as the "C" ABI on non-Windows x86_64 targets.
+  * Corresponds to GCC and clang's `__attribute__((sysv_abi))`.
 
   > [!NOTE]
   > See <https://wiki.osdev.org/System_V_ABI> or <https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI> for more information.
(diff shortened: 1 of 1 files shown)


ehuss:
@ChrisDenton We rephrased the description of the `"system"` ABI, and were wondering if you could take a look to see if the new text was accurate and conveyed the right meaning.

@Amanieu or @ChrisDenton  We made a few new claims in this edit about the `"C"` ABI. We were sure who to ask, but were wondering if you knew the answer. In particular it is specifying that:
* "win64" is the same as the "C" ABI on Windows x86_64 targets
* "sysv64" is the same as the "C" ABI on non-Windows x86_64 targets

Amanieu:
Yes, that seems correct to me.

ChrisDenton:
Should probably @workingjubilee and @RalfJung who are knowledgable in this area.

This looks good to me though https://github.com/rust-lang/rust/issues/136946 will need to adjust the the Windows 32-bit documentation to note the varargs peculiarity.

src/items/external-blocks.md
```diff
@@ -111,13 +111,16 @@ r[items.extern.abi.standard]
...
-* `unsafe extern "Rust"` --- The default ABI when you write a normal `fn foo()` in any Rust code.
+* `unsafe extern "Rust"` --- The native calling convention for Rust functions and closures. This is the default when the function's ABI is not specified. The Rust ABI offers no stability guarantees.
 
 r[items.extern.abi.c]
-* `unsafe extern "C"` --- This is the same as `extern fn foo()`; whatever the default your C compiler supports.
+* `unsafe extern "C"` --- The "C" ABI attempts to match the default ABI chosen by the dominant C compiler for the target.
 
 r[items.extern.abi.system]
-* `unsafe extern "system"` --- Usually the same as `extern "C"`, except on Win32, in which case it's `"stdcall"`, or what you should use to link to the Windows API itself.
+* `unsafe extern "system"` --- This is equivalent to `extern "C"` except on Win32. On Win32, it's equivalent to `"stdcall"`.
```
ChrisDenton: It's a pre-existing issue but maybe this should spell out "Windows x86_32" rather than abbreviating to "Win32"?

src/items/external-blocks.md
```diff
@@ -111,13 +111,16 @@ r[items.extern.abi.standard]
 The following ABI strings are supported on all platforms:
 
 r[items.extern.abi.rust]
-* `unsafe extern "Rust"` --- The default ABI when you write a normal `fn foo()` in any Rust code.
+* `unsafe extern "Rust"` --- The native calling convention for Rust functions and closures. This is the default when the function's ABI is not specified. The Rust ABI offers no stability guarantees.
 
 r[items.extern.abi.c]
-* `unsafe extern "C"` --- This is the same as `extern fn foo()`; whatever the default your C compiler supports.
+* `unsafe extern "C"` --- The "C" ABI attempts to match the default ABI chosen by the dominant C compiler for the target.
```
RalfJung: "attempts to match" is a bit... weak?^^

src/items/external-blocks.md
```diff
@@ -111,13 +111,16 @@ r[items.extern.abi.standard]
...
 
 r[items.extern.abi.system]
-* `unsafe extern "system"` --- Usually the same as `extern "C"`, except on Win32, in which case it's `"stdcall"`, or what you should use to link to the Windows API itself.
+* `unsafe extern "system"` --- This is equivalent to `extern "C"` except on Windows x86_32 where it is equivalent to `"stdcall"`.
+
+  > [!NOTE]
+  > As the correct underlying ABI on Windows is target-specific, it's best to use `extern "system"` when attemping to link to the Windows API.
```
RalfJung: This isn't true for *all* of the Windows API though, some of it uses "C"...

src/items/external-blocks.md
```diff
@@ -126,33 +129,57 @@ r[items.extern.abi.platform]
...
-* `unsafe extern "cdecl"` --- The default for x86_32 C code.
+* `unsafe extern "cdecl"` --- The calling convention typically used with x86_32 C code.
   * Only available on x86_32 targets.
 
+  > [!NOTE]
+  > See <https://learn.microsoft.com/en-us/cpp/cpp/cdecl> and <https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl> for more information.
+
 r[items.extern.abi.stdcall]
-* `unsafe extern "stdcall"` --- The default for the Win32 API on x86_32.
+* `unsafe extern "stdcall"` --- The calling convention typically used by the [Win32 API] on x86_32.
```
RalfJung, line 150: Does this correspond to `__stdcall` or `__attribute__((stdcall))` or anything like that on any other compiler?

src/items/external-blocks.md
```diff
@@ -126,33 +129,57 @@ r[items.extern.abi.platform]
 There are also some platform-specific ABI strings:
 
 r[items.extern.abi.cdecl]
-* `unsafe extern "cdecl"` --- The default for x86_32 C code.
+* `unsafe extern "cdecl"` --- The calling convention typically used with x86_32 C code.
```
RalfJung, line 139: Does this correspond to `__scdecl` or `__attribute__((cdecl))` or anything like that on any other compiler?

src/items/external-blocks.md
```diff
@@ -126,33 +129,57 @@ r[items.extern.abi.platform]
...
   * Only available on x86_64 targets.
+  * "win64" is the same as the "C" ABI on Windows x86_64 targets.
+
+  > [!NOTE]
+  > See <https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions> and <https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_x64_calling_convention> for more information.
 
 r[items.extern.abi.sysv64]
-* `unsafe extern "sysv64"` --- The default for C code on non-Windows x86_64.
+* `unsafe extern "sysv64"` --- The System V ABI.
   * Only available on x86_64 targets.
```
RalfJung, line 174: Do clang/gcc/MSVC have attributes that correspond to our win64/sysv64?

RalfJung (commented):
This sounds right but I don't actually know that much about all these target-specific ABIs.

Not sure how I feel about linking to the Microsoft docs for some of these (even when they are not Microsoft-defined ABIs, or are they?), but I don't know a better place to link to either.

src/items/external-blocks.md
ehuss: Updated, though I think you meant cdecl here?

src/items/external-blocks.md
ehuss: I believe the answer is "yes". Are you suggesting this should explicitly say that like the entry for fastcall does?

src/items/external-blocks.md
ehuss: I believe the answer here is also "yes" (assuming the `s` is a typo).

src/items/external-blocks.md
ehuss: I believe the answer is "kinda".

AFAIK, Microsoft's `__cdecl` is the same as our "win64" or "C" (and I believe clang/gcc also do the same).

I believe for "sysv64", the answer is "no" for MSVC, you cannot explicitly specify that on MSVC. It looks like clang and gcc use the attribute `sysv_abi`.

Are you suggesting that we should be more explicit about these equivalences?

src/items/external-blocks.md
RalfJung: No I meant "C". "cdecl" is x86-32-only.

src/items/external-blocks.md
RalfJung: Seems worthwhile, doesn't it?

src/items/external-blocks.md
RalfJung: > AFAIK, Microsoft's __cdecl is the same as our "win64" or "C"

That's just a consequence of "C" being "win64" for them, right?

Do clang/gcc support explicitly using the win64 ABI somehow? That would be the most direct equivalent.

> Are you suggesting that we should be more explicit about these equivalences?

We're very explicit about it in some cases but not others. That seemed odd.

ChrisDenton: It's truer to say that `__cdecl`, `__fastcall`, etc is ignored outside of x86_32. I don't believe MSVC has an explicit calling convention for arm64 or x86_64. It just ignores any calling convention other than `__vector_call` and thus uses the default. I guess you could claim that one of these ignored calling conventions is the "real" name for the calling convention on x64 or arm64 but I don't think it's particularly useful to do so.

ehuss:
OK, I pushed up a change that tries to be more consistent with showing the equivalent attributes.

My homework for investigating this:

Primary sources are:
* Clang: https://clang.llvm.org/docs/AttributeReference.html
* LLVM: https://llvm.org/doxygen/namespacellvm_1_1CallingConv.html
* GCC: https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html
* Descriptions in wikipedia: https://en.wikipedia.org/wiki/X86_calling_conventions
* Other links in the PR.

For the translation in rustc to LLVM, there are multiple places to look:
* [`AbiMapping`](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L76-L186) -- This is a first canonicalization of ABIs. Most ABIs have a clear translation ("Rust" is "Rust"). Some are erased at this stage (efiapi gets translated to various other ABIs based on the target).
* [`llvm::CallConv`](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L666-L705) -- Converts rustc's call convention to LLVM's.
* LLVM [`CallingConv.h`](https://github.com/llvm/llvm-project/blob/HEAD/llvm/include/llvm/IR/CallingConv.h) -- LLVM's definition of the different calling conventions.
* Clang [`CallingConv`](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L278-L315) -- How attributes to mapped to a calling convention.
* Clang [`ClangCallConvToLLVMCallConv`](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L52-L120) -- Converts clang's calling convention to LLVM.

Generally I tried to make sure that clang translation to LLVM is the same as rustc's.

* cdecl: 
    * [rustc -> CanonAbi::C](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L129-L130) -> [llvm::CCallConv](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L669)
    * [clang -> CC_C](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L279) -> [llvm::C](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L53-L55)

* stdcall:
    * [rustc -> X86Call::Stdcall](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L138-L142) -> [llvm::X86StdcallCallConv](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L669)
    * [clang -> CC_X86StdCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L280) -> [llvm::X86_StdCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L56-L57)

* win64:
    * [rustc -> X86Call::Win64](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L153-L154) -> [llvm::X86_64_Win64](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L701)
    * [clang -> CC_Win64](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L285) -> [llvm::Win64](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L65)

* sysv64:
    * [rustc -> X86Call::SysV64](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L152-L154) -> [llvm::X86_64_SysV](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L698)
    * [clang -> CC_X86_64SysV](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L286) -> [llvm::X86_64_SysV](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L66-L67)

* aapcs:
    * [rustc -> ArmCall::Aapcs](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L110-L111) -> [llvm::ArmAapcsCallConv](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L692)
    * [clang -> CC_AAPCS](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L288C31-L288C60) -> [llvm::ARM_AAPCS](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L68-L69)
    * Note: does not seem documented in GCC.

* fastcall:
    * [rustc -> X86Call::Fastcall](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L132-L136) -> [llvm::X86FastcallCallConv](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L696)
    * [clang -> CC_X86FastCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L281) -> [llvm::X86_FastCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L58-L59)

* thiscall:
    * [rustc -> X86Call::Thiscall](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L144-L145) -> [llvm::X86_ThisCall](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_codegen_llvm/src/abi.rs#L699)
    * [clang -> CC_X86ThisCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/include/clang/Basic/Specifiers.h#L282) -> [llvm::X86_ThisCall](https://github.com/llvm/llvm-project/blob/0fcce4fb7b85ed42feb2f2291405fe4a2292b2b4/clang/lib/CodeGen/CGCall.cpp#L62-L63)

* efiapi:
    * [rustc -> ArmCall::Aapcs | X86Call::Win64 | CanonAbi::C](https://github.com/rust-lang/rust/blob/dc2c3564d273cf8ccce32dc4f47eaa27063bceb9/compiler/rustc_target/src/spec/abi_map.rs#L104-L107) -> llvm (see above)
    * clang: N/A? Maybe uses the EFIAPI macro.

traviscross (approved):

PR #2005, opened 2025-09-18
RISC-V: Add vector state registers

a4lg:
`vxsat` (vector fixed-point saturation flag) and `vxrm` (vector fixed-point rounding mode) are subset, readable/writable mirrors of the `vcsr` register.  Still, adding them to the default preserved register list will improve clarity.

*   `vcsr`
    *   Bit 0–0 mirror: `vxsat`
    *   Bit 1–2 mirror: `vxrm`

Note that, the reason `vxrm` (rounding mode register which normally the developer writes) is in the list is, it is not preserved across calls on all ratified ABIs.

Quoting `asm.rules.preserves_flags`,
> As a general rule, the flags covered by `preserves_flags` are those which are not preserved when performing a function call.

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -1343,7 +1343,7 @@ r[asm.rules.preserved-registers]
     - Floating-point status (`FPSR` register).
   - RISC-V
     - Floating-point exception flags in `fcsr` (`fflags`).
-    - Vector extension state (`vtype`, `vl`, `vcsr`).
+    - Vector extension state (`vtype`, `vl`, `vxsat`, and `vxrm`).
   - LoongArch
     - Floating-point condition flags in `$fcc[0-7]`.
   - s390x


a4lg:
Related: rust-lang/rust#146530

cc: @nnethercote, @Amanieu, @taiki-e

Amanieu:
I would remove vcsr in this case since it is redundant.

a4lg:
Removed `vcsr` from the reference itself but preserved it in the commit message.

Amanieu (approved):

taiki-e (approved):

src/inline-assembly.md
```diff
@@ -1343,7 +1343,7 @@ r[asm.rules.preserved-registers]
     - Floating-point status (`FPSR` register).
   - RISC-V
     - Floating-point exception flags in `fcsr` (`fflags`).
-    - Vector extension state (`vtype`, `vl`, `vcsr`).
+    - Vector extension state (`vtype`, `vl`, `vxsat` and `vxrm`).
```
ehuss: Caveat: I don't know anything about RISC-V.

Is there a reason this wouldn't still include `vcsr` in the list?
Maybe worded something like "`vcsr` (and the subset mirrors `vxsat` and `vxrm`)"?
Or just list all three like `vcsr`, `vxsat`, and `vxrm`?

We weren't generally clear why it would be removed.

Amanieu: `vxsat` and `vxrm` are 2 bitfields in `vcsr`, but can also be accessed separately as individual registers. The new text is clearer since we're referring to the actual pieces of state rather than the specific register they are stored in.

Amanieu: To be clear, they are the *only* bit fields in `vcsr`, there is nothing else in `vcsr`.

ehuss: Makes sense, thank you!

ehuss (approved):
Thanks!

PR #2006, opened 2025-09-23
destructors.md: point to core:: instead of std::

tshepang:
std re-exports these from core, so rather show the source path (to avoid an indirection)


ehuss (approved):
Thanks!

PR #2007, opened 2025-09-23
destructors.md: improve readability by adding pauses

tshepang:
Had to read second sentence twice to more easily understand it, something the pause would have helped avoid.


ehuss (approved):
Thanks!

PR #2010, opened 2025-09-24
Clarify `use $crate` restriction

ehuss:
The previous wording for this restriction was pretty confusing to me. I don't remember what I was thinking when I wrote it, and I can't find any historical explanation either. `use` paths can use `$crate` as long as they have more than one segment (`use $crate::foo` is obviously OK).

I have rewritten this to make it clear it is specifically about `use $crate`. One could say that restriction is already covered by the previous point that says `use crate;` requires an `as`, but for some reason `use $crate as foo` doesn't work either. So I have left this as a separate rule for now.

cc https://github.com/rust-lang/rust/pull/146972#discussion_r2376951695 for context.

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -375,7 +375,18 @@ r[items.use.restrictions.duplicate-name]
 * As with any item definition, `use` imports cannot create duplicate bindings of the same name in the same namespace in a module or block.
 
 r[items.use.restrictions.macro-crate]
-* `use` paths with `$crate` are not allowed in a [`macro_rules`] expansion.
+* `use $crate` is not allowed in a [`macro_rules`] expansion.
+
+  ```rust,compile_fail
+  macro_rules! example {
+      () => {
+          use $crate;  // ERROR: `$crate` may not be imported
+          use $crate as foo; // ERROR: `$crate` may not be imported
+      };
+  }
+
+  example!{}
+  ```
 
 r[items.use.restrictions.variant]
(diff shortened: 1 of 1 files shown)


ehuss:
Opening as a draft because I'm wondering how intentional this is.

traviscross:
cc @yaahc

yaahc:
https://github.com/yaahc/rust/blob/adaa838976ff99a4f0661136322f64cb466b58a0/compiler/rustc_resolve/src/build_reduced_graph.rs#L630-L631 is the logic that controls this, it doesn't make a carveout for renaming imports but it doesn't mention them either so I'm not positive if that means it was intended to also disallow this case.

https://github.com/rust-lang/rust/pull/45846 is the PR where this logic was introduced. I did a quick search of the discussion on this PR and the associated PR, tracking issue, and RFC (discussion and content) and honestly couldn't find any mention of the $crate syntax at all outside of the PR diff itself. If I had to guess I'd assume this was probably an oversight. @petrochenkov's comment https://github.com/rust-lang/rust/issues/35612#issuecomment-578425321 seems to support this theory.

much better example: https://github.com/rust-lang/rust/pull/146972#discussion_r2382960171
in this one vadim explicitly mentions that renaming $crate should be allowed. Also looks like the contributor on this PR is quite actively working on improving the situation and consistency of this logic, I'll make sure any updates they make get properly integrated into the associated language in the reference.

ehuss:
Closing as this got rewritten in https://github.com/rust-lang/reference/pull/2136.

PR #2013, opened 2025-09-25
const functions: separate rule about users and rule about what is allowed in such functions

tshepang:


ehuss (approved):
Thanks! I found the existing text was a little awkward, so I did a little bit of reworking. There were two separate sections defining "const functions". Instead of trying to conjure some logic for how to split them, I decided to just merge them.

PR #2015, opened 2025-09-25
use "tuple enum variant" more consistently

tshepang:
spent too long looking for the meaning of "tuple variant", so this should hopefully help others avoid that


ehuss (approved):
Thanks!

PR #2017, opened 2025-09-25
Remove caveats related to `format_args!` expansion

traviscross:
The behavior of `format_args!`, with respect to `super let`, had some caveats that we can remove now that Rust [rust-lang/reference#145880] has landed, so let's remove our remarks on that and simplify the tests accordingly.

[rust-lang/reference#145880]: https://github.com/rust-lang/rust/issues/145880

cc @ehuss @dianne


ehuss (approved):
Thanks!

PR #2018, opened 2025-09-25
Fix field-less `repr(C)` enum docs

RalfJung:
The docs for field-less repr(C) enums are wrong in the sense that they say "the `C` representation has the size and alignment of the default `enum` size and alignment for the target platform's C ABI" which implies that there's a single size and alignment (determined by the target) that all repr(C) enums share -- which isn't true. The size of the enum depends on the discriminant values and is intended to mimic what the default C compiler for the target would do with an enum that has the same discriminant values.

Also, it seems worth mentioning the type that the discriminant expressions of an enum are type-checked at: that's `isize` for all enums expect those with primitive representation.

This PR presupposes that we are going ahead with https://github.com/rust-lang/rust/pull/147017 and documents things as-if the FCW added there was already a hard error. This is mostly because otherwise it's unclear what we should document as the logic before that bug doesn't always match the target's C compiler (see https://github.com/rust-lang/rust/pull/146504).

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -117,6 +117,9 @@ Each enum instance has a _discriminant_: an integer logically associated to it t
 r[items.enum.discriminant.type]
 Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.
 
+r[items.enum.discriminant.type-primitive]
+Enums with a [primitive representation] have discriminants of the type named by the representation. This also applies to enums that combine the `C` representation with a primitive representation (see [layout.repr.primitive-c]). For example, an enum with a `u8` representation can only have discriminant values between 0 and 255 inclusive.
+
 ### Assigning discriminant values
 
 r[items.enum.discriminant.explicit]
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -115,7 +115,7 @@ r[items.enum.discriminant.intro]
 Each enum instance has a _discriminant_: an integer logically associated to it that is used to determine which variant it holds.
 
 r[items.enum.discriminant.type]
-Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.
+Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type or another means of distinguishing variants in the actual memory layout.
 
 ```rust
 # use core::mem::size_of;
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -117,8 +117,38 @@ Each enum instance has a _discriminant_: an integer logically associated to it t
 r[items.enum.discriminant.type]
 Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.
 
+```rust
+# use core::mem::size_of;
+enum E {
+    V1 = 0isize, // OK: `isize` is the discriminant type.
+    V2,
+}
+
+assert!(size_of::<E>() <= size_of::<isize>());
+```
+
+```rust,compile_fail,E0308
+enum E {
+    V = 0u8, // ERROR: Expected `isize`, found `u8`.
+}
+```
+
 r[items.enum.discriminant.type-primitive]
-Enums with a [primitive representation] have discriminants of the type named by the representation. This also applies to enums that combine the `C` representation with a primitive representation (see [layout.repr.primitive-c]). For example, an enum with a `u8` representation can only have discriminant values between 0 and 255 inclusive.
+Enums with a [primitive representation] have discriminants of the type named by the representation. This also applies to enums that combine the `C` representation with a primitive representation (see [layout.repr.primitive-c]).
+
+```rust
+#[repr(u8)]
+enum E {
+    V = 0u8, // OK: `u8` is the discriminant type.
+}
+```
+
+```rust,compile_fail,E0308
+#[repr(u8)]
+enum E {
+    V = 0isize, // ERROR: Expected `u8`, found `isize`.
+}
+```
 
 ### Assigning discriminant values
 
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -114,8 +114,8 @@ r[items.enum.discriminant]
 r[items.enum.discriminant.intro]
 Each enum instance has a _discriminant_: an integer logically associated to it that is used to determine which variant it holds.
 
-r[items.enum.discriminant.repr-rust]
-Under the [`Rust` representation], the discriminant is interpreted as an `isize` value. However, the compiler is allowed to use a smaller type (or another means of distinguishing variants) in its actual memory layout.
+r[items.enum.discriminant.type]
+Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.
 
 ### Assigning discriminant values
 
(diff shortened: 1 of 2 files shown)


src/type-layout.md
```diff
@@ -344,11 +344,15 @@ assert_eq!(std::mem::align_of::<SizeRoundedUp>(), 4); // From a
...
-the default `enum` size and alignment for the target platform's C ABI.
+For [field-less enums], the `C` representation requires the discriminant values to
+be representable by the `int` type in the target platform's C ABI.
+Nevertheless, the type of the discriminant is `isize`.
+The size and alignment of the enum then match that of a C enum with the same
+discriminant values.
 
 > [!NOTE]
 > The enum representation in C is implementation defined, so this is really a "best guess". In particular, this may be incorrect when the C code of interest is compiled with certain flags.
+> For maximum portability, it is always preferred to set the size and alignment explicitly using a [primitive representation](#r-layout.repr.primitive.enum).
```
workingjubilee: This is the *Rust* reference but here we are explicitly steering people on how to do an FFI portability thing primarily designed to be compatible with C code. I think we might as well go ahead and note how this can be done in the C source code. Please feel free to disagree, I don't want to scope creep anything unduly.
```suggestion
> For maximum portability, it is always preferred to set the size and alignment explicitly using a [primitive representation](#r-layout.repr.primitive.enum).
> This can be mimicked in C23 or C++17 using the syntax for enums with fixed underlying types.
```

RalfJung:
@traviscross  is there anything this needs to proceed? :)

src/type-layout.md
```diff
@@ -282,10 +282,12 @@ assert_eq!(std::mem::align_of::<SizeRoundedUp>(), 4); // From a
 r[layout.repr.c.enum]
 #### `#[repr(C)]` Field-less Enums
 
-For [field-less enums], the `C` representation has the size and alignment of the default `enum` size and alignment for the target platform's C ABI.
+For [field-less enums], the `C` representation requires the discriminant values to either all be representable by the `int` type in the target platform's C ABI, or to all be representable by the `unsigned int` type. Nevertheless, the type of the discriminant is `isize`. The size and alignment of the enum then match that of a C enum with the same discriminant values (and without a fixed underlying type). Crucially, the equivalent C type is determined based on the discriminant values *after* they have been cast to `isize`.
```
ehuss: We were looking at this paragraph, and we were feeling like this may be packing too many things into a single rule. We were wondering about maybe splitting it up into a few separate things, perhaps along these lines:

Change the `items.enum.discriminant.repr-rust` rule to also cover `repr(C)`. It could maybe look something like this:

r[items.enum.discriminant.repr-discriminant]

Enums with the [`Rust` representation] or [`C` representation] have discriminants of type `isize`.

With the [`Rust` representation], the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.

----

And then in the layout section:

r[layout.repr.c.enum.discriminant]

For [field-less enums] with the `C` representation, all discriminant values (which are of [type `isize`][items.enum.discriminant.repr-discriminant]) must be representable by either the `int` or `unsigned int` type in the target platform's C ABI.

r[layout.repr.c.enum.size-align]

A [field-less enum] with the `C` representation will have the same size and alignment as a C enum that has the same discriminant values and does not explicitly specify the underlying integer type.

----

WDYT?

RalfJung: > With the [Rust representation], the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.

This is also true with the C representation -- the memory layout will usually be `i32`, not `isize`, and may be even smaller (e.g. on ARM). The only difference is that with the Rust representation we leave it unspecified which layout we pick (today we pick the smallest that can fit all values), while with the C representation we pick the same layout as a C compiler would.

The rest sounds good. However here...

> A [field-less enum] with the C representation will have the same size and alignment as a C enum that has the same discriminant values and does not explicitly specify the underlying integer type.

... it may be worth mentioning (possibly as a non-normative note) that this is about the values *after they have been cast to `isize`*. The order of operations is:
- Compute all discriminant values. This produces `isize` result, so things get wrapped if they don't fit.
- Check that they all fit into `int` or all fit into `unsigned int`.
- Compute the size the C enum with those (already wrapped!) values would have.

RalfJung: Oh, also, this...

> all discriminant values (which are of [type isize][items.enum.discriminant.repr-discriminant]) must be representable by either the int or unsigned int type in the target platform's C ABI.

...is subtly wrong. The correct requirement is that either all values fit in `int` or all values fit in `unsigned int`. This is not equivalent to what you wrote: if one value is `-1` and another one is 0xFFFFFFFF, then they both "fit into `int` or `unsigned int`", but the enum still gets rejected because they don't all fit into `int`, and they also don't all fit into `unsigned int`. I specifically worded this very carefully in my PR. :)

ehuss: Thanks, I see now.

Also, just to double check, I was uncertain if the "without a fixed underlying type" is referring specifically to the new C23 `enum name: type {...}` syntax, or something else (`__attribute__((__packed__))`? `-fshort-enums`?). I could have sworn this ability existed before C23, but my memory is clearly clouded.

And...Would you be OK if I push these changes on top of your PR here?

RalfJung: > if the "without a fixed underlying type" is referring specifically to the new C23 enum name: type {...} syntax

That is what I meant, yes. AFAIK that is the only tweak that the standard has here. GCC and other compilers have their extensions but if we start talking about all compiler extensions here we'll never be done. ;)

> And...Would you be OK if I push these changes on top of your PR here?

Yes, that's entirely fine. :)

RalfJung:
@traviscross @ehuss  what is the current status of this? I think I answered all questions above.

traviscross:
Thanks @RalfJung for the ping.  I pushed a set of commits addressing the items that were raised in discussion earlier.  Please let me know if this looks right to you.

src/items/enumerations.md
```diff
@@ -114,8 +114,37 @@ r[items.enum.discriminant]
...
 
-r[items.enum.discriminant.repr-rust]
-Under the [`Rust` representation], the discriminant is interpreted as an `isize` value. However, the compiler is allowed to use a smaller type (or another means of distinguishing variants) in its actual memory layout.
+r[items.enum.discriminant.type]
+Enums without a [primitive representation] have discriminants of type `isize`. However, the compiler may use a smaller type (or another means of distinguishing variants) in the actual memory layout.
+
+```rust
+enum E {
+    V = 0isize, // OK: `isize` is the discriminant type.
+}
```
RalfJung, line 125: Maybe add something like this to drive home the point from above?
```rust
// The discriminant is not actually stored as an `isize`.
assert!(size_of::<E>() < size_of::<isize>());
```

src/type-layout.md
```diff
@@ -367,10 +367,25 @@ assert_eq!(std::mem::offset_of!(SizeRoundedUp, b), 0);
...
+> `repr(C)` enums without a primitive representation have discriminant values of type `isize`. See [items.enum.discriminant.type].
+
+> [!NOTE]
+> `rustc` accepts enums whose discriminant values do not meet this requirement but lints against them. This will become an error in the future.
+
+r[layout.repr.c.enum.size-align]
+A [field-less enum] with the `C` representation has the same size and alignment as a C enum with the same discriminant values and no fixed underlying type.
+
+> [!NOTE]
+> The size and alignment are determined from the discriminant values *after* they have been cast to `isize`.
```
RalfJung: I think this should go together with the note above about the discriminant values having type `isize`. Currently it feels strangely isolated.

RalfJung (commented):
This is great, thanks. :) Two minor nits for the non-normative part.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2018/3b38834b39f732c64686f7c64aa29dcf3cd83ba5..12bf021ad22fa6e56ecbad9fcc412a9f542c11ce)*

src/items/enumerations.md
traviscross: Thanks.  I've added that and some other examples.

(As compared with your example, I used `<=` here, as we don't make a language guarantee that it will in fact be smaller than `isize`.)

src/type-layout.md
traviscross: Makes sense.  Moved it.

traviscross (approved):

traviscross:
Thanks @RalfJung.

PR #2019, opened 2025-09-26
Move inferred sentence to an example block

tshepang:


ehuss (approved):
Thanks! I reworked this a bit because the text wasn't an explicit rule, but more of a commentary on how the inferred type is sometimes used. I moved the text to an example block to better convey how it was intended.

Issue #2020, opened 2025-09-28
Further cases where lone underscore isn't covered


mattheww:
The `Token` production in `lex.token.syntax` is missing an entry for lone `_` (which isn't included in `IDENTIFIER_OR_KEYWORD`.

Similarly the lists in `macro.proc.token.macro_rules` and `macro.proc.token.tree` are missing lone underscore, because the Reference doesn't consider "identifier" to include lone uderscore.

ehuss:
The `Token` rule includes `PUNCTUATION` which includes `_`. Does that cover what you are looking for in `lex.token.syntax`?

As for the proc-macro sections, I have opened https://github.com/rust-lang/reference/pull/2021 to try to clarify that.

mattheww:
Ah yes, `Token` is correct as it stands.

Issue #1236, opened 2022-07-06
Grammar for "identifier or keyword" seems wrong


chichid:
Based on the grammar reference for [identifier](https://doc.rust-lang.org/stable/reference/identifiers.html) The following '_' identifier should be invalid, since:

_ is not  **XID_Start XID_Continue*** 
and it's not  **'_' XID_Continue+** either (it's not followed by at least one XID_CONTINUE)

I think the rule should be: 
```
IDENTIFIER_OR_KEYWORD :
      (XID_Start | '_') XID_Continue*

OR---

IDENTIFIER_OR_KEYWORD :
       XID_Start XID_Continue*
     | '_' XID_CONTINUE*
```

bjorn3:
`_` is indeed meant to not be a valid identifier. It is categorized as punctuation instead. As such

```
IDENTIFIER_OR_KEYWORD :
     XID_Start XID_Continue*
   | _ XID_Continue+
```

really is correct.

chichid:
hmmm, I did this playground, and it seems at least the TokenStream is parsed as an ident. So either the grammar is wrong, or the Proc-Macro API is.

![image](https://user-images.githubusercontent.com/6054451/177600665-d6f0cdb0-0e6f-45e9-b1dd-c6bd62b50ad0.png)

bjorn3:
Looks like the lexer indeed considers `_` to be an identifier and the parser considers it to be a keyword. (All keywords are identifiers as far as the lexer is concerned.)

```
IDENTIFIER_OR_KEYWORD :
      (XID_Start | '_') XID_Continue*
```

is closest to the actual logic the lexer implements.

ehuss:
I'm not sure how best to document this, but I think one option is have it be something specific to proc-macros. `_` is very carefully not an identifier in the rest of the reference because of how it interacts with everything else (like patterns, macro_rules, etc.). One thought is to keep proc-macro specific documentation in the [proc-macro](https://doc.rust-lang.org/proc_macro/index.html) API docs (for example, [Ident](https://doc.rust-lang.org/proc_macro/struct.Ident.html) could more clearly document that `_` is considered a keyword, as it already mentions its behavior around keywords). But then we would need to consider what happens if a proc-macro generates a token stream with that identifier, which I think might end up as a keyword? 

So maybe `_` could be changed to be a keyword? That would be a bit unusual since it more closely matches punctuation, though I suppose the distinction isn't all that important.

PR #2021, opened 2025-09-29
Clarify proc-macro token tree

ehuss:
This attempts to clarify some details about proc-macro tokens, in particular the differences in handling `_` from `macro_rules`. This includes several changes:

- Removed the wordy description of a `macro_rules` TokenTree. This is already defined in the grammar in more detail. This does remove the sentence discussing metavariable substitutions. I would prefer to defer this to a more detailed description in the `macro_rules` chapter on how metavariable substitutions work.
- Removed the English description of a proc-macro token tree, and instead write it as a grammar production.

Closes https://github.com/rust-lang/reference/issues/2020 
Closes https://github.com/rust-lang/reference/issues/1236

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -318,45 +318,75 @@ r[macro.proc.token]
 ## Declarative macro tokens and procedural macro tokens
 
 r[macro.proc.token.intro]
-Declarative `macro_rules` macros and procedural macros use similar, but
-different definitions for tokens (or rather [`TokenTree`s].)
-
-r[macro.proc.token.macro_rules]
-Token trees in `macro_rules` (corresponding to `tt` matchers) are defined as
-- Delimited groups (`(...)`, `{...}`, etc)
-- All operators supported by the language, both single-character and
-  multi-character ones (`+`, `+=`).
-    - Note that this set doesn't include the single quote `'`.
-- Literals (`"string"`, `1`, etc)
-    - Note that negation (e.g. `-1`) is never a part of such literal tokens,
-      but a separate operator token.
-- Identifiers, including keywords (`ident`, `r#ident`, `fn`)
-- Lifetimes (`'ident`)
-- Metavariable substitutions in `macro_rules` (e.g. `$my_expr` in
-  `macro_rules! mac { ($my_expr: expr) => { $my_expr } }` after the `mac`'s
-  expansion, which will be considered a single token tree regardless of the
-  passed expression)
+Declarative `macro_rules` macros and procedural macros use similar but different definitions for tokens. Token trees in `macro_rules` (corresponding to `tt` matchers) are defined as the [TokenTree] production. Token trees for procedural macros are described below.
 
 r[macro.proc.token.tree]
-Token trees in procedural macros are defined as
-- Delimited groups (`(...)`, `{...}`, etc)
-- All punctuation characters used in operators supported by the language (`+`,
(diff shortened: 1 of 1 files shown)


src/procedural-macros.md
```diff
@@ -318,45 +318,75 @@ r[macro.proc.token]
...
 r[macro.proc.token.intro]
-Declarative `macro_rules` macros and procedural macros use similar, but
-different definitions for tokens (or rather [`TokenTree`s].)
-
-r[macro.proc.token.macro_rules]
-Token trees in `macro_rules` (corresponding to `tt` matchers) are defined as
-- Delimited groups (`(...)`, `{...}`, etc)
-- All operators supported by the language, both single-character and
-  multi-character ones (`+`, `+=`).
-    - Note that this set doesn't include the single quote `'`.
```
petrochenkov, line 329: I wrote the English text to highlight the differences between macro_rules and proc macro tokens, like multi-character operators (which are not a thing in proc macros), lifetimes, negative literals.
From the formal grammar the difference is very hard to see.

src/procedural-macros.md
```diff
@@ -318,45 +318,75 @@ r[macro.proc.token]
...
+      `(` ProcMacroTokenTree* `)`
+    | `[` ProcMacroTokenTree* `]`
+    | `{` ProcMacroTokenTree* `}`
+
+ProcMacroToken ->
+      ProcMacroTokenIdent
+    | ProcMacroTokenPunct
+    | ProcMacroTokenLiteral
+
+ProcMacroTokenIdent -> IDENTIFIER_OR_KEYWORD | RAW_IDENTIFIER | `_`
```
petrochenkov, line 339: Doesn't reference consider `_` a keyword?
I think it should, it's lexically an identifier, and it's reserved, so it's a keyword (== reserved identifier).

src/procedural-macros.md
```diff
@@ -318,45 +318,75 @@ r[macro.proc.token]
...
 Note that the conversions below may happen lazily, so they might not happen if
 the tokens are not actually inspected.
 
 r[macro.proc.token.conversion.to-proc_macro]
-When passed to a proc-macro
-- All multi-character operators are broken into single characters.
+When tokens are passed to a proc-macro:
+
+- All multi-character punctuation is broken into single characters.
+- `_` is treated as an identifier.
```
petrochenkov, line 389: This is misleading, `_` *is* a (reserved) identifier.

src/procedural-macros.md
```diff
@@ -368,13 +398,15 @@ When passed to a proc-macro
...
-When emitted from a proc macro
+When tokens are emitted from a proc macro:
+
 - Punctuation characters are glued into multi-character operators
   when applicable.
 - Single quotes `'` joined with identifiers are glued into lifetimes.
 - Negative literals are converted into two tokens (the `-` and the literal)
   possibly wrapped into a delimited group ([`Group`]) with implicit delimiters
   ([`Delimiter::None`]) when it's necessary for preserving parsing priorities.
+- The `_` identifier is interpreted as the `_` punctuation.
```
petrochenkov, line 409: I don't think the reference should ever call `_` punctuation, it's a (reserved) identifier in all senses except one - `ident` matcher in `macro_rules` doesn't accept it for backward compatibility reasons (but it can likely be changed over an edition).

petrochenkov:
@rustbot author
(I don't have rights to change labels directly here.)

PR #2030, opened 2025-09-30
Consistently use sentence case

ehuss:
Part of our style guide is to use sentence case for headings. The Reference was very inconsistent with this. This updates all the headings to use sentence case.

### DIFF
--- a/src/abi.md
+++ b/src/abi.md
@@ -1,5 +1,5 @@
 r[abi]
-# Application Binary Interface (ABI)
+# Application binary interface (ABI)
 
 r[abi.intro]
 This section documents features that affect the ABI of the compiled output of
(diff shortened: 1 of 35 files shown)


src/abi.md
```diff
@@ -1,5 +1,5 @@
 r[abi]
-# Application Binary Interface (ABI)
+# Application binary interface (ABI)
```
tshepang, line 2: paused seeing this one, then "nah, it good" (then I checked how wikipedia does it, and it matches your change)

tshepang (approved):

joshtriplett (approved):
I read through the entire commit, and it looks reasonable. Good to have consistency here.

(Would love to see this automatically checked by CI, but that can come later.)

traviscross (approved):

PR #2032, opened 2025-10-03
fix(css): Correctly specify grammar-literal-bg

epage:
Some themes refer to `--gramar-literal-bg` rather than `--grammar-literal-bg`.

I found this when running [`typos`](https://github.com/crate-ci/typos) on the Reference out of curiosity after seeing rust-lang/rust has adopted it.  Since I was already creating a PR, I went ahead and included the other corrections.  If there is interest in adding a CI job to check this on every commit, like rust-lang/rust, I can set that up.

### DIFF
--- a/mdbook-spec/src/grammar/parser.rs
+++ b/mdbook-spec/src/grammar/parser.rs
@@ -382,7 +382,7 @@ impl Parser<'_> {
                     xs.push(x);
                     self.index += 1;
                 }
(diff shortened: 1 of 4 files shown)


traviscross (approved):

traviscross:
Thanks @epage.

Since the spelling fix to the CSS rule does affect the rendering, I verified it does not impair contrast.

Issue #1078, opened 2021-07-26
Clarify terminology around destructors and dropping


ehuss:
In #1037 a question came up about the exact wording around the introduction of the [destructors chapter](https://github.com/rust-lang/reference/blob/0fe357f176673b4bb9a093504a8ae1dcaf4c6f8f/src/destructors.md) which says

> When an initialized variable or temporary goes out of scope its *destructor* is run, or it is *dropped*.

The PR author interpreted the phrase *dropped* meaning a synonym for running the destructor. But I read it as two separate concepts:

* When any value goes out of scope, it is "dropped".
* If the value has a destructor, that destructor is run.

I don't think "dropping" and "running destructors" can be used interchangeably because not all values have a destructor.  However, I am uncertain if that is correct, or if that is the best way to document these concepts.  Either way, it would be good to clarify that a little.

crlf0710:
I think all values have destructors, but (in C++ term) not all destructors are non-trivial. On the other hand, `std::mem::needs_drop`'s naming is really a little confusing...

botahamec:
According to the documentation for the `Drop` trait, "[Copy types] cannot have destructors", so they can't possibly be synonyms.

PR #2034, opened 2025-10-04
Destructors: Reword introduction of terms "destructor" and "dropped"

joshtriplett:
Avoid implying a difference between the two. The existing wording could
be read as "either its destructor is run or it is dropped", implying a
difference.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -3,10 +3,11 @@ r[destructors]
 
 r[destructors.intro]
 When an [initialized]&#32;[variable] or [temporary] goes out of
-[scope](#drop-scopes), its *destructor* is run, or it is *dropped*. [Assignment]
-also runs the destructor of its left-hand operand, if it's initialized. If a
-variable has been partially initialized, only its initialized fields are
-dropped.
+[scope](#drop-scopes), it is *dropped*, which means its *destructor* is run.
+(The two terms are used interchangeably here.)
+[Assignment] also runs the destructor of its left-hand operand, if it's
+initialized. If a variable has been partially initialized, only its initialized
+fields are dropped.
 
 r[destructors.operation]
 The destructor of a type `T` consists of:


src/destructors.md
```diff
@@ -3,10 +3,11 @@ r[destructors]
 
 r[destructors.intro]
 When an [initialized]&#32;[variable] or [temporary] goes out of
-[scope](#drop-scopes), its *destructor* is run, or it is *dropped*. [Assignment]
-also runs the destructor of its left-hand operand, if it's initialized. If a
-variable has been partially initialized, only its initialized fields are
-dropped.
+[scope](#drop-scopes), it is *dropped*, which means its *destructor* is run.
+(The two terms are used interchangeably here.)
```
traviscross, line 7: *dropped* and *destructor* aren't the two terms used interchangeably.  (We use the emphasis to define terms, so what's under emphasis matters here.)  While extending the emphasis on the latter to include "is run" would be one option, that reads strangely to me still, so maybe there are other options here.

It'd seem better to me if we could find a factoring that avoided this parenthetical.  This is just an introduction.  Maybe something like reversing the terms back to the original order and saying something like "When a value or temporary goes out of scope and its destructor is run, we say that it has been *dropped*."

(I'm happy to make the revision here if something along these lines sounds OK to you.)

(This assumes that we want to treat these as equated, which as I mentioned in private communication, is probably not how this is meant currently.)

traviscross: (I'm happy to make the revision here if something along these lines sounds OK to you.)

joshtriplett: I'm fine with this PR having flagged the underlying issue; feel free to edit and merge, as I'm not attached to the exact wording here and I'm not currently building anything further atop this.

ehuss:
We have issue https://github.com/rust-lang/reference/issues/1078 for this. As what came up before, I still think that *dropping* and *running a destructor* are not the same thing.

The wording could certainly be clarified, but I think it would be good for us to first come to an agreement what terminology we want to use.

traviscross:
> We have issue #1078 for this. As what came up before, I still think that _dropping_ and _running a destructor_ are not the same thing.
> 
> The wording could certainly be clarified, but I think it would be good for us to first come to an agreement what terminology we want to use.

I'm talking with Josh separately, and this was my read as well that the existing text meant to distinguish these rather than to equate them.

We should at a minimum remove the comma there, since that makes it seem like the clause is non-restrictive when it's meant restrictively -- i.e. it can't be dropped without changing the meaning of the sentence.

traviscross:
See:

- https://github.com/rust-lang/reference/pull/2035

PR #2036, opened 2025-10-07
Fix grammar of `ConstParam`

yyny:
The grammar for `ConstParam` in https://doc.rust-lang.org/reference/items/generics.html#grammar-ConstParam was specified as:

```grammar
ConstParam →
    'const' IDENTIFIER : Type
    ( '=' BlockExpression | IDENTIFIER | '-'? LiteralExpression )?
```

This implies that `IDENTIFIER` and `'-'? LiteralExpression` may not be preceded by `=`, but this is obviously not the case.


ehuss (approved):
Thanks!

PR #2041, opened 2025-10-08
Further specify temporary scoping for `static`s and `const`s

dianne:
This PR applies various tweaks relating to how the destructors chapter handles `static` items, `const` items, and `const` blocks. In particular:
- I've tried to rework rules specific to `fn` items and closures so that they can work for `static`s and `const`s as well. In some cases, this was a bit awkward, so further reworks may be necessary. Most importantly, this specifies that `static` and `const` bodies are temporary scopes; this was a point of confusion in compiler comments, so I feel it should be documented here. See also rust-lang/rust#147471
- I've clarified that static lifetime extension prevents destructors from being run. See also rust-lang/rust#145868
- I reworked the example for `const` block lifetime extension to demonstrate that it extends borrowed temporaries to statics, preventing them from being dropped.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -77,10 +77,10 @@ Overloaded operators are not distinguished from built-in operators and [binding
 modes] are not considered.
 
 r[destructors.scope.list]
-Given a function, or closure, there are drop scopes for:
+Given a function, closure, [static][static item] or [constant item], or [const block][const block expression], there are drop scopes for:
 
-r[destructors.scope.function]
-* The entire function
+r[destructors.scope.body]
+* The entire function, the body of a static or constant item, or a const block
 
 r[destructors.scope.statement]
 * Each [statement]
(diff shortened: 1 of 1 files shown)


src/destructors.md
```diff
@@ -474,17 +480,17 @@ r[destructors.scope.lifetime-extension.exprs]
...
-* The initializer expression.
+* The initializer expression of a let statement, the body of a static or constant item, or the final expression of a const block.
 * The operand of an extending [borrow] expression.
 * The [super operands] of an extending [super macro call] expression.
 * The operand(s) of an extending [array][array expression], [cast][cast
   expression], [braced struct][struct expression], or [tuple][tuple expression]
   expression.
 * The arguments to an extending [tuple struct] or [tuple enum variant] constructor expression.
-* The final expression of an extending [block expression] except for an [async block expression].
+* The final expression of an extending [block expression] except for an [async block expression] or const block expression.
```
dianne, line 493: This in particular feels awkward to me, but I'm not sure how to make it read better without entirely reworking the format of the list and/or making "extending" a binary relation (mapping expressions to extended scopes). The first bullet point is a base case, which effectively duplicates `destructors.scope.lifetime-extension.let` (and now `destructors.scope.lifetime-extension.static`), except without specifying the scope of extended temporaries. This gets a bit messy for const blocks in particular: const blocks are one of the base cases (they extend temporaries to statics) but they also have to be an exception to the grammar for extending expressions (so that we're not also specifying that they extend temporaries to the end of a block when they appear in a `let` initializer).

dianne:
Closing since most of this has been subsumed by https://github.com/rust-lang/reference/pull/2051. Once rust-lang/rust#146098 goes through (or doesn't go through) I'll see if there's anything left that needs doing here and reopen/rework this if so.

PR #2047, opened 2025-10-09
generalise paragraph

tshepang:
"Avoid" was used because all entries before 1b57f6ca48f51a760d84df4d42b8f5927668f3c1 were about things to avoid


ehuss (approved):
Thanks!

Issue #588, opened 2019-04-21
Document `$crate` in macro matcher


ehuss:
`$crate` in a macro matcher can only match the literal tokens `$` `crate` (I think).  That is:

```rust
// This does NOT take a token tree. It takes 4 tokens: `$` `crate` `:` `tt`
macro_rules! m {
    ($crate:tt) => ()
}
```

ehuss:
See also https://github.com/rust-lang/reference/pull/2049#discussion_r2429634986 and https://github.com/rust-lang/reference/pull/1816#pullrequestreview-2884513401. The current docs for the `ident` matcher is confusing at best in terms of how it accepts `$crate`.

Issue #587, opened 2019-04-21
Split `$crate` into two tokens.


ehuss:
The grammar rules should be clear that `$` `crate` is two tokens. I think putting a space between them would be sufficient (such as in paths.md)

petrochenkov:
This is tricky.
`$ crate` is two tokens in `macro_rules` input, but it turns into a single identifier token in the macro output after the macro is expanded, see https://github.com/rust-lang/rust/issues/55640 for more details.

PR #2049, opened 2025-10-12
Change underscore to be a keyword

### LINKED ISSUE #1236: Grammar for "identifier or keyword" seems wrong

### LINKED ISSUE #2020: Further cases where lone underscore isn't covered

ehuss:
This changes underscore from being a punctuation character to a keyword. This is intended to help better align with proc-macros, which treat it as an [`Ident`](https://doc.rust-lang.org/proc_macro/struct.Ident.html).

Note one unusual rule is inline assembly `ParamName` which is `IDENTIFIER_OR_KEYWORD`. From what I can tell, it does accept `_`, but the fmt template does not. Templates are not specified in great detail in the std docs, and don't touch on this fact.

Closes https://github.com/rust-lang/reference/issues/1236 
Closes https://github.com/rust-lang/reference/issues/2020

cc @petrochenkov @mattheww

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -9,13 +9,13 @@ XID_Start -> <`XID_Start` defined by Unicode>
 
 XID_Continue -> <`XID_Continue` defined by Unicode>
 
-RAW_IDENTIFIER -> `r#` IDENTIFIER_OR_KEYWORD _except `crate`, `self`, `super`, `Self`, `_`_
+RAW_IDENTIFIER -> `r#` IDENTIFIER_OR_KEYWORD
 
 NON_KEYWORD_IDENTIFIER -> IDENTIFIER_OR_KEYWORD _except a [strict][lex.keywords.strict] or [reserved][lex.keywords.reserved] keyword_
 
 IDENTIFIER -> NON_KEYWORD_IDENTIFIER | RAW_IDENTIFIER
 
-RESERVED_RAW_IDENTIFIER -> `r#_`
+RESERVED_RAW_IDENTIFIER -> `r#` (`_` | `crate` | `self` | `Self` | `super`)
 ```
 
 <!-- When updating the version, update the UAX links, too. -->
@@ -72,7 +72,7 @@ Unlike a normal identifier, a raw identifier may be any strict or reserved
 keyword except the ones listed above for `RAW_IDENTIFIER`.
 
 r[ident.raw.reserved]
-It is an error to use the [RESERVED_RAW_IDENTIFIER] token `r#_` in order to avoid confusion with the [WildcardPattern].
+It is an error to use the [RESERVED_RAW_IDENTIFIER] token.
 
 [`extern crate`]: items/extern-crates.md
 [`no_mangle`]: abi.md#the-no_mangle-attribute
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -3,15 +3,13 @@ r[ident]
 
 r[ident.syntax]
 ```grammar,lexer
-IDENTIFIER_OR_KEYWORD ->
-      XID_Start XID_Continue*
-    | `_` XID_Continue+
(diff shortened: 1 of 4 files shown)


src/identifiers.md
```diff
@@ -3,15 +3,13 @@ r[ident]
...
+IDENTIFIER_OR_KEYWORD -> ( XID_Start | `_` ) XID_Continue*
 
 XID_Start -> <`XID_Start` defined by Unicode>
 
 XID_Continue -> <`XID_Continue` defined by Unicode>
 
-RAW_IDENTIFIER -> `r#` IDENTIFIER_OR_KEYWORD _except `crate`, `self`, `super`, `Self`_
+RAW_IDENTIFIER -> `r#` IDENTIFIER_OR_KEYWORD _except `crate`, `self`, `super`, `Self`, `_`_
 
 NON_KEYWORD_IDENTIFIER -> IDENTIFIER_OR_KEYWORD _except a [strict][lex.keywords.strict] or [reserved][lex.keywords.reserved] keyword_
```
petrochenkov, line 15: `RESERVED_RAW_IDENTIFIER` a few lines below may be redundant now.

src/macros-by-example.md
```diff
@@ -134,7 +134,7 @@ Valid fragment specifiers are:
   * `block`: a [BlockExpression]
   * `expr`: an [Expression]
   * `expr_2021`: an [Expression] except [UnderscoreExpression] and [ConstBlockExpression] (see [macro.decl.meta.edition2024])
-  * `ident`: an [IDENTIFIER_OR_KEYWORD], [RAW_IDENTIFIER], or [`$crate`]
+  * `ident`: an [IDENTIFIER_OR_KEYWORD] except `_`, [RAW_IDENTIFIER], or [`$crate`]
```
petrochenkov, line 137: Pre-existing, but this doesn't seem correct, `ident` accepts raw identifiers and expanded `$crate` (and unexpanded `$crate` is two tokens and not `IDENTIFIER_OR_KEYWORD`).

petrochenkov (approved):

src/macros-by-example.md
ehuss: Yea, we have a few issues related to this (like https://github.com/rust-lang/reference/issues/588 and https://github.com/rust-lang/reference/issues/587). I was also [uneasy](https://github.com/rust-lang/reference/pull/1816#pullrequestreview-2884513401) adding this in the first place.

I don't know the right way to document that the expanded `$crate` can be accepted by `ident`. Perhaps this needs to just be more explicit what it means?

src/identifiers.md
ehuss: Hm, I spent some time thinking about this and how to correctly express that these keywords are *rejected*. The current _except_ clause didn't express that in the way that I was intending. I pushed up a commit that instead of removing the reserved rule, it moves the _except_ part into the reserved rule.

Or, to put it in another way, `r#crate` is a token, it's just rejected as an error. The previous lexical grammar wasn't really conveying that.

traviscross (approved):

PR #2051, opened 2025-10-12
Specify temporary lifetime extension through expressions

dianne:
Reference PR for rust-lang/rust#146098. This includes a reworked definition of extending expressions with the aim of expressing the new semantics more uniformly.

### DIFF
--- a/src/scoping-model.lagda.md
+++ b/src/scoping-model.lagda.md
@@ -302,7 +302,7 @@ refScope-146098 (_ ▷ const…=●⨾) = scope ctxt-top
 -- Extending subexpressions preserve `&●`'s temporary scope.
 refScope-146098 (e-block ▷ ⦃…⨾●⦄) = refScope-146098 e-block
 refScope-146098 (e-array ▷ ⟦…,●,…⟧ _) = refScope-146098 e-array
-refScope-146098 (e-ref-outer ▷ &●) = refScope-146098 e-ref-outer
+refScope-146098 (e-ref ▷ &●) = refScope-146098 e-ref
 -- In other contexts, the temp scope of `&●` is the enclosing temp
 -- scope.
 {-# CATCHALL #-}


src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
 #### Extending based on expressions
 
 r[destructors.scope.lifetime-extension.exprs.extending]
-For a let statement with an initializer, an *extending expression* is an
-expression which is one of the following:
+An *extending expression* is an expression which is one of the following:
```
dianne: A major change here: an extending expression is now any expression that preserves lifetime extension, defined non-inductively. I found this helps with generalizing the definition beyond `let` statement initializers, but I also often found myself having to refer to an expression being "extending when its parent is extending"; that's are now just an extending expression.

src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
...
 
-* The initializer expression.
-* The operand of an extending [borrow] expression.
-* The [super operands] of an extending [super macro call] expression.
-* The operand(s) of an extending [array][array expression], [cast][cast
+* The initializer expression of a `let` statement or the body expression of a [static][static item] or [constant item].
```
dianne: Something slightly weird here: the last expression of a `const` block morally should be here, but it'd be a bit messy to have to exclude it from the rule for blocks lower down. Given that this definition of extending expressions doesn't care about where scopes are extended to, it shouldn't be a semantic issue, but it might warrant reformatting and/or an admonition.

src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
...
 r[destructors.scope.lifetime-extension.exprs.super-macros]
-The [super temporaries] of an extending [super macro call] expression have their [scopes][temporary scopes] [extended].
+The [scopes][temporary scopes] of the [super temporaries] of an extending [super macro call] expression are *extended through* the scope of the super macro call expression.
 
-> [!NOTE]
-> `rustc` does not treat [array repeat operands] of extending [array] expressions as extending expressions. Whether it should is an open question.
->
-> For details, see [Rust issue #146092](https://github.com/rust-lang/rust/issues/146092).
+r[destructors.scope.lifetime-extension.exprs.parent]
+If a temporary scope is extended through the scope of an extending expression, it is extended through that scope's [parent][destructors.scope.nesting].
```
dianne: I struggled a bit with how to express this; in its current form it's a bit of a hack. Ideally, I feel like it wouldn't need to be a separate rule or to refer to the definition of scope nesting, but it's working around something subtle: by ensuring that expressions' temporary scopes are only extended by their scope-ancestors, we can work around const blocks having parent expressions that (to my understanding) shouldn't be considered ancestor scopes of the const block's body; temporaries extended by const blocks are extended to the end of the program[^1]. Maybe there's a simpler way to express this, and regardless it could probably use an admonition.

I do think that some sort of "extended by" or "extended through" or "extending based on" relation is necessary though, regardless of how exactly we choose to define/present it. I feel there's too much ambiguity if we can't precisely associate expressions we're extending the temporary scopes of with the scopes they're being extended to.

[^1]: This PR doesn't make all the changes needed to iron that out, but see #2041.

src/destructors.md
```diff
@@ -552,6 +563,19 @@ let x = format_args!("{:?}", temp()); // As above.
...
+# fn use_temp(_: &()) {}
+// The final expression of a block is extending. Since the block below
+// is not itself extending, the temporary is extended to the block
+// expression's temporary scope, ending at the semicolon.
+use_temp({ &temp() });
+// As above, the final expressions of `if`/`else` blocks are
+// extending, which extends the temporaries to the `if` expression's
+// temporary scope.
+use_temp(if true { &temp() } else { &temp() });
+```
```
dianne: Some additional examples would probably be good. Maybe it would help to have one where temporaries are extended through a block but not to the end of a statement? That could also be used as a `compile_fail` example.

src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
...
->
-> For details, see [Rust issue #146092](https://github.com/rust-lang/rust/issues/146092).
+r[destructors.scope.lifetime-extension.exprs.parent]
+If a temporary scope is extended through the scope of an extending expression, it is extended through that scope's [parent][destructors.scope.nesting].
+
+r[destructors.scope.lifetime-extension.exprs.let]
+A temporary scope extended through a `let` statement scope is [extended] to the scope of the block containing the `let` statement ([destructors.scope.lifetime-extension.let]).
+
+r[destructors.scope.lifetime-extension.exprs.static]
+A temporary scope extended through a [static][static item] or [constant item] scope or a [const block][const block expression] scope is [extended] to the end of the program ([destructors.scope.lifetime-extension.static]).
```
dianne: I wonder if there's a way to cut down on the duplication here. I felt these rules were necessary to be precise about where temporaries' scopes are extended to, but having them in the introduction to the lifetime extension feels necessary too.

src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
...
+
+r[destructors.scope.lifetime-extension.exprs.other]
+A temporary scope extended through the scope of a non-extending expression is [extended] to that expression's [temporary scope].
```
dianne: Alternative to de-duplicating the above two rules, maybe there should be a more detailed section alongside `destructors.scope.lifetime-extension.let` and `destructors.scope.lifetime-extension.static` for this?

src/destructors.md
dianne: Maybe it could also use some additional text (or even an admonition?) to make clear the interaction with `if` block scopes and Rust 2024's tail expression scopes. I'm not sure exactly how much explaining is needed for that, though.

traviscross:
Thanks @dianne; will have a look.

traviscross:
@ehuss and I looked carefully through this today.  Some thoughts.

After staring at this awhile, I see why this makes it easier to express the core idea of https://github.com/rust-lang/rust/pull/146098 -- that the relative drop order within an expression should be consistent regardless of where that expression is.  Expressing the rules inductively for defining an extending expression ends up pushing against this.

So it does, I think, make a deep sort of sense to drop the inductive approach, as this PR does.

At the same time, the inductive approach was carrying a lot of load.  The reader could visualize walking down an expression, outside to inside, at each step checking whether it was still an extending expression.  Then there was essentially one key rule that leaned on that:

> The operand of an extending [borrow] expression has its [temporary scope] [extended].

What this PR does, in dropping the inductive framing, is to widen the definition of an extending expression significantly, and thereby put a lot more load on the rules that follow to narrow this back down.  That's where it ends up getting a bit subtle.  With this approach, both @ehuss and I found building intuition and checking the correctness of the rules to be more difficult.  So, in dropping the inductive approach, I think we have some work to do in order to recover the same level of clarity.

Adding some `!EXAMPLE` admonitions that walk through applying these rules would help and be a good start here.  It'll also be worth more carefully defining what it means to "extend to" or "extend through" something.  The bit where we define some things as "extending through" but then "clamp" them so as to only "extend to" with the final rule ("A temporary scope extended through the scope of a non-extending expression is [extended] to that expression's [temporary scope].") is another bit of subtlety that it might be worth finding another way to express.

I'll leave some other thoughts inline.

src/destructors.md
traviscross: Yes, this greatly widens the definition of an extending expression, putting more load on the other rules.

src/destructors.md
traviscross: Let's see the admonition for this (with one or more examples).

src/destructors.md
traviscross: Let's see the admonition that you have in mind here (with one or more examples).  Let's also add rules to define what it means exactly to "extend to", "extend through", etc.

src/destructors.md
traviscross: This is all rather subtle.  Erring on the side of more examples and explanation will be better here.

src/destructors.md
traviscross: Agreed they felt necessary here.

src/destructors.md
traviscross: Not sure what you mean here exactly.

traviscross: This rule is particular is one where I wonder how we might do better.  The idea, currently, is that the rules above define when a temporary scope would be "extended through" the scope of an expression, and then this rule says, well, if it's a non-extending expression, then we only "extend to" it.

src/destructors.md
```diff
@@ -474,36 +474,47 @@ r[destructors.scope.lifetime-extension.exprs]
...
+> `rustc` does not treat [array repeat operands] of [array] expressions as extending expressions. Whether it should is an open question.
+>
+> For details, see [Rust issue #146092](https://github.com/rust-lang/rust/issues/146092).
+
+So the borrow expressions in `{ &mut 0 }`, `(&1, &mut 2)`, and `Some(&mut 3)`
 are all extending expressions. The borrows in `&0 + &1` and `f(&mut 0)` are not.
 
 r[destructors.scope.lifetime-extension.exprs.borrows]
-The operand of an extending [borrow] expression has its [temporary scope] [extended].
+The [temporary scope] of the operand of a [borrow] expression is *extended through* the scope of the borrow expression.
```
traviscross: It's a bit subtle that, after having defined above what is an extending expression, that this rule then doesn't use that definition at all.  It makes sense -- all operands of any borrow expression are extending.  But it'll be worth adding an admonition under each of these rules that elaborates the rationale and gives one or more examples.

src/destructors.md
traviscross: In particular, what it means to "extend through" a parent scope is going to be worth careful explanation and one or more examples in an admonition.

dianne:
I'll try and address individual points soon, but first one higher-level note that I'll try to use as guidance when restructuring:

> At the same time, the inductive approach was carrying a lot of load. The reader could visualize walking down an expression, outside to inside, at each step checking whether it was still an extending expression.

This PR's approach still has a visual intuition to it, so it should be presented in a way that makes that clear: instead of walking down an extending expression to check whether an `&`'s operand is extended, you walk up from an `&` and check whether you're in an extending expression at each step to determine the temporary scope of the `&`'s operand. Possibly a separate inductive definition would be useful for making that clear? Along with examples, of course. I was worried defining too many things all at once would make it difficult to follow, but examples should help.

traviscross:
> This PR's approach still has a visual intuition to it, so it should be presented in a way that makes that clear: instead of walking down an extending expression to check whether an `&`'s operand is extended, you walk up from an `&` and check whether you're in an extending expression at each step to determine the temporary scope of the `&`'s operand. Possibly a separate inductive definition would be useful for making that clear? Along with examples, of course.

Yes, exactly.  This upward walk is what I'd been expecting but it has to be teased out a bit from the presentation here.  Hopefully making that more clear will make the rules here more clear.

dianne:
I did a pass to clean up definitions, reframe "extending based on expressions" as a property of `&` expressions (rather than of extending expressions or `let` statements), and add clarification and examples. Some parts are still awkward, unfortunately. I've left review comments on a couple things in particular I'm not happy with yet.

src/destructors.md
```diff
@@ -473,37 +476,102 @@ let &ref x = &*&temp(); // OK
 r[destructors.scope.lifetime-extension.exprs]
 #### Extending based on expressions
 
+r[destructors.scope.lifetime-extension.exprs.borrows]
+The [temporary scope] of the operand of a [borrow] expression is the *extended scope* of the operand expression, defined below.
+
+r[destructors.scope.lifetime-extension.exprs.super-macros]
+The [scope][temporary scope] of each [super temporary] of a [super macro call] expression is the extended scope of the super macro call expression.
```
dianne: Two things here:
- I think it's helpful to start `destructors.scope.lifetime-extension.exprs` with a straightforward explanation of what all the definitions are for. Otherwise, I found that either too many definitions are needed before any of them can be explained, or the explanations have to refer to terms that haven't been defined yet. For the sake of directness, I moved `destructors.scope.lifetime-extension.exprs.borrows` to the start, but that meant having to refer to terms that are defined later. A possible alternative would be writing a new `destructors.scope.lifetime-extension.exprs.intro` rule, but when I tried that it felt redundant.
- I don't love the term "extended scope", but I wasn't able to think of anything better, either in terms of naming or in terms of reframing to avoid defining it at all. I feel like it could probably also use some additional clarification but I'm still thinking of a good admonition to add for it.

src/destructors.md
```diff
@@ -473,37 +476,102 @@ let &ref x = &*&temp(); // OK
...
+
 r[destructors.scope.lifetime-extension.exprs.extending]
-For a let statement with an initializer, an *extending expression* is an
-expression which is one of the following:
+The extended scope of an expression is defined in terms of *extending expressions* and their *extending parents*. An extending expression is an expression which is one of the following:
```
dianne: Three things here:
- To simplify the definitions of extending expressions and extended scopes, I've made `let` initializers, `static`/`const` items, and `const` block tails no longer extending. This might hurt the intuition of extending expressions a bit though.
- I added the definition of "extending parent" when reworking the upwards walk. This keeps it rigorous without hacks (the parent of an expression isn't defined elsewhere, and going through the parent of a scope to save words is confusing). It also also makes it clear that extending expressions are all subexpressions and thus have parent expressions. But spending words on it feels a bit awkward too, either adding bloat or making things convoluted (or both).
- Extending expressions are necessary to define extended scopes, but extended scopes are necessary to justify the definition of extending expressions. As such, I felt a need to connect this to the surrounding sections, but I don't love how it turned out. Maybe there's a better way to handle that.

src/destructors.md
dianne: I think I meant adding a section directly under `destructors.scope.lifetime-extension` to cover that form of lifetime extension, since that's where extension through `let` and through `const`/`static` lived. This isn't a concern any more, I don't think, since I restructured to move those rules under`destructors.scope.lifetime-extension.exprs`.

src/destructors.md
dianne: I haven't added additional examples for this case of lifetime extension yet, but I fleshed out the first one quite a bit to walk through what's going on. I removed the second in the process, but I'll think on how best to readd it later (along with negative examples).

src/destructors.md
dianne: I've fully reworked the formulation of the upward walk from `&` expressions and added examples.

src/destructors.md
dianne: I've handled this by making const block tail expressions no longer be defined as "extending" (along with `let` statement initializers and `static`/`const` item bodies). That way, there's less room for confusion, hopefully. I've also added an example of extending through a const block tail expression to illustrate it (taken from https://github.com/rust-lang/reference/pull/2041)

src/destructors.md
dianne: I've reworked this: now it appears before the definition of extending expressions. It's difficult to provide admonitions and examples before defining all the relevant terms, but I've tried to connect the rules together and provided some detailed examples towards the end.

src/destructors.md
dianne: I've reworked this to avoid the hack of using scopes to express hierarchy entirely.

src/destructors.md
```diff
@@ -388,21 +385,27 @@ let x = &mut 0;
 println!("{}", x);
 ```
 
-r[destructors.scope.lifetime-extension.static]
-Lifetime extension also applies to `static` and `const` items, where it
-makes temporaries live until the end of the program. For example:
+r[destructors.scope.lifetime-extension.sub-expressions]
+If a [borrow], [dereference][dereference expression], [field][field expression], or [tuple indexing expression] has an extended temporary scope, then so does its operand. If an [indexing expression] has an extended temporary scope, then the indexed expression also has an extended temporary scope.
```
dianne: I think this may need a terminology update, a carve-out in `destructors.scope.lifetime-extension.exprs`, or at least clarification on what "extended temporary scope" means. In particular, if you have an expression like
```rust
{ &*&temp() }
```
it's important that this rule takes precedence, so that `temp()` lives past the block. `destructors.scope.lifetime-extension.exprs.borrows` currently is worded to be the definitive lifetime of borrow operators' operands, but if only that was considered, `temp()` would be dropped at the end of the tail expression.

traviscross:
We looked through this in the lang-docs office hours call today.  The recent changes here are great.  Big improvement in clarity.  Thanks to @dianne for those.

We talked about how this might benefit from an introduction (perhaps in an admonition) to the "extending based on expressions" section that tries to give some intuition -- a reading guide -- to a reader coming into this section.

I'm going to give this another careful read through, and @ehuss is planning to run some tests with the PR to make sure the interaction of this with the 2024 edition change is clear, but overall, this seems really close.

src/destructors.md
dianne: Following up on making "extending parents" make more sense and making it clear that "extending expression" isn't a classification of expressions but a classification of subexpressions, maybe it would be worth renaming "extending expression" to "extending subexpression"?  I think it makes more sense from a spec point of view, but it's more awkward to read and write (and it means having to adapt language elsewhere), so I haven't gone through with it.

traviscross: It has some appeal, but I similarly see the problems you mention.  If one says "extending subexpression", then one kind of wants the parent to be an "extending expression" (in the same way as "header", "subheader", etc.), but that's probably confusing.

src/destructors.md
dianne: This is partially addressed by the new example in https://github.com/rust-lang/reference/pull/2051/commits/6536b3be7957261bfe91567af8737454b6c34d40. I think it'll still also want some reformulation or clarification of `lifetime-extension.sub-expressions` to make really make sense, though.

traviscross: Doesn't it almost feel as though the things in this rule should be rolled in to "extending based on expressions"?  E.g.:

> The extended scope of an expression is defined in terms of *extending expressions* and their *extending parents*. An extending expression is an expression which is one of the following... the indexed expression of a tuple indexing expression, the extending parent of which is the tuple indexing expression.

Or is there some important distinction we're trying to encode here in these being handled separately?

dianne: There's a couple distinctions.

First, except for borrow expression operands, none of those subexpressions are extending. For instance,
```rust
let x = (&PrintOnDrop("temporary dropped")).0;
println!("next statement");
```
prints "temporary dropped", then "next statement". Intuitively, since the result of indexing is evaluated and then copied into `x`, the temporary doesn't need to be extended past its use, so we don't extend it. The purpose of `lifetime-extension.sub-expressions` is to handle reborrows, borrows of projections, etc.: if we'd instead written
```rust
let x = &(&PrintOnDrop("temporary dropped")).0;
println!("next statement");
```
we'd be reborrowing the temporary's field rather than evaluating it, so we would extend the temporary, so it'd print "next statement", then "temporary dropped". Extending subexpressions generally should only be those where if you substituted a borrow expression in, it's syntactically obvious that the borrow will need to be live to use the result of evaluating the extending parent expression. Maybe there's a way to express that intuition productively in an admonition somehow?

Second, `lifetime-extension.sub-expressions` also applies when extending based on patterns:
```rust
let ref x = PrintOnDrop("temporary dropped").0;
println!("next statement");
```
prints "next statement", then "temporary dropped". Because `x` is borrowing from a place based on the temporary, the entire temporary is extended. The example in `lifetime-extension.sub-expressions` illustrates this in a bit more detail.

traviscross: I don't fully understand how to apply this rule in the same way that `rustc` does.  I put together a series of examples and mechanically analyzed them according to the rules in this PR.  Can you explain the behavior of the last two examples in light of the behavior of the ones that precede them?

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5de79227c445ecab2814e3826620f832)

traviscross: Setting that aside, and setting aside the matter of the interaction with "extending based on patterns" (which as we discussed on Zulip, might end up needing to be restructured eventually given how it would interact with a binding version of `super let` or `super let in`), it *feels* to me as though these are properly a kind of extending expression but that, as we work outward, we're carrying a flag indicating whether we last stepped through a "full" extending expression or through a more "limited" extending expression, and the value of that flag affects our behavior, e.g. when we get out to the let initializer.

(To the degree that's not true, maybe it's worth considering whether it should be.)

dianne: My notes on each example, per the mental model I was aiming for with that change (I'll address the limited extending model after):
1. `let _ = (&e.log(1)).0; //~ Not extended.`
Our analyses agree here.
2. `let _ = &(&e.log(2)).0; //~ Extended.`
Our analyses agree here.
3. `let _ = &(&(e.log(2),).0).0; //~ Extended.`
`e.log(2)` here isn't extended; instead, it's immediately moved into the tuple temporary, which is extended. I believe its scope is that of the tuple expression, per [destructors.scope.operands](https://doc.rust-lang.org/reference/destructors.html#operands). I'd also add another step: the reason `(&(e.log(2),).0).0` uses its extended scope as its temporary scope is because it's the operand of a borrow expression. Otherwise, our analyses agree.
4. `let _ = &[(e.log(2),).0][0].0; //~ Extended.`
In this case, the temporary that gets extended is the array given by evaluating `[(e.log(2),).0]`. A place based on it, `array[0].0`, is borrowed, and since we want that borrow to be live for the block, we extend the array's lifetime.
5. `let _ = &[(e.log(2),).0]; //~ Extended.`
As with 4, the temporary that gets extended is the array given by evaluating `[(e.log(2),).0]`.
6. `let _ = &[(&e.log(1)).0]; //~ Not extended.`
As with 4 and 5, the temporary being extended is the array. The key here is that we're not borrowing from the `e.log(1)` temporary in the final value; we evaluate `(&e.log(1)).0` to something like `LogDrop(_e1, 1)` where `_e1` is a reborrow of `e` and then move that value into the array. As such, `e.log(1)` does not need to be extended; this is why tuple indexing expressions are not extending parents.
7. `let _ = &(&e.log(1),).0; //~ Not extended.`
This one we *should* extend according to the new model we're discussing. `e.log(1)` should have the tuple expression's temporary scope, which should be extended. Indeed, it's currently an error to use the result because it borrows from the dropped `e.log(1)` temporary. I don't think my compiler PR handles this yet either, but I'd like for it to work.
The way rustc does this is when it extends `(&e.log(1),).0` it applies lifetime-extension.sub-expressions to extend the place(s) it's based on as well (so `(&e.log(1),)`) and then stops.

---

For the limited extending model, I think there's a conceptual reason to keep them distinct. Extending parent expressions represent things like constructors and conditionals that build (or select) a value and are guaranteed to contain (or evaluate to) their extending subexpressions. Projections like `.0` are sort of a dual to that: if we want to extend the lifetime of a projected place, we have to extend the base place too, thus we extend the indexed operand. Actually, I think under this PR's model, we can clean it up and get `&` out of the "sub-expressions" list so it's just the base places of place expressions. I'll experiment with that on the compiler side of things and update this PR if it works out.

dianne: I wonder what the right thing to do is in cases like
```rust
let x = [&temp(), local][1];
let y = &[&temp(), local][1];
```
I see a few options:
1. Under rustc's current model (including under rust-lang/rust#146098 currently), neither `let` statement extends its `temp()`. This is safest in a way: this means we're not extending unused temporaries to the end of the block. Of course, it's still possible today to create unused extended temporaries, e.g. with `let _ = &temp();`.
2. Under this PR after the latest update, I think only the definition of `y` would extend its `temp()`, since the array would have an extended temporary scope and `temp()` would have the array's temporary scope as its temporary scope. In the case of `x`, the array wouldn't have an extended temporary scope because the result of indexing is used as a value.
3. Under a model where indexed expressions are limited-extending, I could possibly see both statements extending their `temp()`s, because in either case we can see that the `temp()` could end up referenced in the final value. I'd have to think through what the exact rules would be to end up with that result.

traviscross: Makes sense.  Regarding 3..=6... oops, yes, of course -- I must have been staring at too many variations for too long.  Interesting about 7.

Revised analysis of the interesting cases:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=%2F%2F+This+demonstrates+a+subtle+distinction+between+extending%0A%2F%2F+expressions+and+the+%60sub-expression%60+rule.%0A%2F%2F%0A%2F%2F+See%3A%0A%2F%2F%0A%2F%2F+-+https%3A%2F%2Fgithub.com%2Frust-lang%2Freference%2Fpull%2F2051%23discussion_r2526294771%0A%2F%2F%0A%2F%2F+Author%3A+TC%0A%2F%2F+Date%3A+2025-11-14%0A%0A%23%21%5Ballow%28path_statements%29%5D%0A%0Afn+main%28%29+%7B%0A++++assert_drop_order%281..%3D2%2C+%7Ce%7C+%7B+%2F%2F~+Not+extended.%0A++++++++let+x+%3D+%28%26e.log%281%29%29.0%3B%0A++++++++%2F%2F++++++-------------%0A++++++++x%3B%0A++++++++%2F%2F+-+%60e.log%281%29%60%2C+parent+%60%26e.log%281%29%60%0A++++++++%2F%2F+++-+extending+%3D+yes+%28per+%60extending%60%29%0A++++++++%2F%2F+++-+temporary+scope+%3D+extended+scope+%28per+%60borrows%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%26e.log%281%29%60%2C+parent+%60%28%26e.log%281%29%29.0%60%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+temporary+scope+%28per+%60other%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+The+%60sub-expressions%60+rule+says%2C+currently%2C+%22if+a+tuple%0A++++++++%2F%2F+indexing+expression+has+extended+temporary+scope%2C+then+so%0A++++++++%2F%2F+does+it+operand%22.++If+we+were+to+interpret+this+to+mean%0A++++++++%2F%2F+that+the+extended+scope+above+should+be+set+to+the+extended%0A++++++++%2F%2F+scope+of+the+parent%2C+then+we%27d+get+the+wrong+result.%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%28%26e.log%281%29%29.0%60%2C+let+initializer%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+block+%28per+%60exprs.let%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+Therefore%2C+the+temporary+is+dropped+at+the+end+of+the+%60let%60%0A++++++++%2F%2F+statement.%0A++++++++e.log%282%29%3B%0A++++%7D%29%3B%0A++++assert_drop_order%281..%3D2%2C+%7Ce%7C+%7B+%2F%2F~+Extended.%0A++++++++let+x+%3D+%26%28%26e.log%282%29%29.0%3B%0A++++++++%2F%2F++++++--------------%0A++++++++x%3B%0A++++++++%2F%2F+-+%60e.log%282%29%60%2C+parent+%60%26e.log%282%29%60%0A++++++++%2F%2F+++-+extending+%3D+yes%0A++++++++%2F%2F+++-+temporary+scope+%3D+extended+scope+%28per+%60borrows%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%26e.log%282%29%60%2C+parent+%60%28%26e.log%282%29%29.0%60%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+temporary+scope+%28per+%60other%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60sub-expressions%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+Both+the+%60other%60+and+%60sub-expressions%60+rules+could+apply%0A++++++++%2F%2F+above.++We+assume+that+%60sub-expressions%60+takes+precedence.%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%28%26e.log%282%29%29.0%60%2C+parent+%60%26%28%26e.log%282%29%29.0%60%0A++++++++%2F%2F+++-+extending+%3D+yes%0A++++++++%2F%2F+++-+temporary+scope+%3D+extended+scope+%28per+%60borrows%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%26%28%26e.log%282%29%29.0%60%2C+let+initializer%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+block+%28per+%60exprs.let%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+Therefore%2C+the+temporary+is+dropped+at+the+end+of+the%0A++++++++%2F%2F+block.%0A++++++++e.log%281%29%3B%0A++++%7D%29%3B%0A++++assert_drop_order%281..%3D2%2C+%7Ce%7C+%7B+%2F%2F~+Not+extended.%0A++++++++let+_x+%3D+%26%28%26e.log%281%29%2C%29.0%3B%0A++++++++%2F%2F++++++---------------%0A++++++++%2F%2F_x%3B+%2F%2F~+error%5BE0716%5D%3A+temporary+value+dropped+while+borrowed%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60e.log%281%29%60%2C+parent+%60%26e.log%281%29%60%0A++++++++%2F%2F+++-+extending+%3D+yes%0A++++++++%2F%2F+++-+temporary+scope+%3D+extended+scope+%28per+%60borrows%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%26e.log%281%29%60%2C+parent+%60%28%26e.log%281%29%2C%29%60%0A++++++++%2F%2F+++-+extending+%3D+yes%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%28%26e.log%281%29%2C%29%60%2C+parent+%60%28%26e.log%281%29%2C%29.0%60%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+temporary+scope+%28per+%60other%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60sub-expressions%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+Both+the+%60other%60+and+%60sub-expressions%60+rules+could+apply%0A++++++++%2F%2F+above.++We+assume+that+%60sub-expressions%60+takes+precedence.%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%28%26e.log%281%29%2C%29.0%60%2C+parent+%60%26%28%26e.log%281%29%2C%29.0%60%0A++++++++%2F%2F+++-+extending+%3D+yes%0A++++++++%2F%2F+++-+temporary+scope+%3D+extended+scope+%28per+%60borrows%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+extended+scope+of+parent+%28per+%60parent%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+-+%60%26%28%26e.log%281%29%2C%29.0%60%2C+let+initializer%0A++++++++%2F%2F+++-+extending+%3D+no%0A++++++++%2F%2F+++-+temporary+scope+%3D+statement+%28per+%60enclosing%60%29%0A++++++++%2F%2F+++-+extended+scope+%3D+block+%28per+%60exprs.let%60%29%0A++++++++%2F%2F%0A++++++++%2F%2F+Therefore%2C+the+temporary+should+be+dropped+at+the+end+of%0A++++++++%2F%2F+the+block%2C+though+it+is+not+by+%60rustc%60+currently.%0A++++++++e.log%282%29%3B%0A++++%7D%29%3B%0A%7D%0A%0A%2F%2F+%23+Test+scaffolding...%0A%0Ause+core%3A%3Acell%3A%3ARefCell%3B%0A%0Astruct+DropOrder%28RefCell%3CVec%3Cu64%3E%3E%29%3B%0Astruct+LogDrop%3C%27o%3E%28%26%27o+DropOrder%2C+u64%29%3B%0A%0Aimpl+DropOrder+%7B%0A++++fn+log%28%26self%2C+n%3A+u64%29+-%3E+LogDrop%3C%27_%3E+%7B%0A++++++++LogDrop%28self%2C+n%29%0A++++%7D%0A%7D%0A%0Aimpl%3C%27o%3E+Drop+for+LogDrop%3C%27o%3E+%7B%0A++++fn+drop%28%26mut+self%29+%7B%0A++++++++self.0.0.borrow_mut%28%29.push%28self.1%29%3B%0A++++%7D%0A%7D%0A%0A%23%5Btrack_caller%5D%0Afn+assert_drop_order%28%0A++++ex%3A+impl+IntoIterator%3CItem+%3D+u64%3E%2C%0A++++f%3A+impl+Fn%28%26DropOrder%29%2C%0A%29+%7B%0A++++let+order+%3D+DropOrder%28RefCell%3A%3Anew%28Vec%3A%3Anew%28%29%29%29%3B%0A++++f%28%26order%29%3B%0A++++let+order+%3D+order.0.into_inner%28%29%3B%0A++++let+expected%3A+Vec%3Cu64%3E+%3D+ex.into_iter%28%29.collect%28%29%3B%0A++++assert_eq%21%28order%2C+expected%29%3B%0A%7D%0A)

Regarding the `sub-expressions` rule:

> If a [borrow], [dereference][dereference expression], [field][field expression], or [tuple indexing expression] has an extended temporary scope, then so does its operand. If an [indexing expression] has an extended temporary scope, then the indexed expression also has an extended temporary scope.

In "extending based on expressions", we're defining expressions to have an extended scope even if that extended scope is never used.  E.g.:

> r[destructors.scope.lifetime-extension.exprs.let]
> The extended scope of the initializer expression of a `let` statement is the scope of the block containing the `let` statement.

For `sub-expressions`, then, it seems we have to account for that somehow.

src/destructors.md
```diff
@@ -473,37 +476,119 @@ let &ref x = &*&temp(); // OK
...
-  expression.
-* The arguments to an extending [tuple struct] or [tuple enum variant] constructor expression.
-* The final expression of an extending [block expression] except for an [async block expression].
-* The final expression of an extending [`if`] expression's consequent, `else if`, or `else` block.
-* An arm expression of an extending [`match`] expression.
+  expression, the extending parent of which is the array, cast, braced struct, or tuple expression.
+* The arguments to a [tuple struct] or [tuple enum variant] constructor expression, the extending parent of which is the constructor expression.
+* The final expression of a plain [block expression] or [`unsafe` block expression], the extending parent of which is the block expression.
+* The final expression of an [`if`] expression's consequent, `else if`, or `else` block, the extending parent of which is the `if` expression.
+* An arm expression of a [`match`] expression, the extending parent of which is the `match` expression.
```
traviscross: It's maybe not what we want to do right now, but it occurred to me that these and the other rules could probably be usefully presented in the manner of inference rules, e.g. (and making no assertions about correctness of this sketch):

```rust
enum Scope {
    Block,
    Stmt,
    // ...
}
struct S(scope: Scope, ext_scope: Scope);

/// Borrow expressions
S(x, x) ⊢  $expr
----------------
S(s, x) ⊢ &$expr

/// Array expressions
S(s, x) ⊢  $expr
----------------------
S(s, x) ⊢ [$expr, ...]

/// Function calls
S(s, s) ⊢       $expr
---------------------------
S(s, x) ⊢ $path($expr, ...)

/// Let initializers
S(Stmt, Block) ⊢           &$expr
---------------------------------
S(Stmt, Block) ⊢ let ... = &$expr

// Etc.
```

When @Nadrieril and I were working on match ergonomics, we ended up finding it rather helpful for analysis to present the rules in this manner.  See, e.g. [this document](https://hackmd.io/zUqs2ISNQ0Wrnxsa9nhD0Q) for a discussion of notation.

@Nadrieril ended up writing a cool [web tool](https://github.com/Nadrieril/typing-rust-patterns) for doing analysis and comparisons based on this kind of notation.

It's similarly been occurring to me here that it would be helpful to have a tool that encodes the formal rules in software and walks through the analysis of an expression.  (Of course, one option is always to do something like that in the compiler with an approach similar to `#[rustc_capture_analysis]`).

src/destructors.md
dianne: > In "extending based on expressions", we're defining expressions to have an extended scope even if that extended scope is never used.
 
> For `sub-expressions`, then, it seems we have to account for that somehow.

This is a terminology problem. I couldn't think of a better term than "extended scope" at the time, but there's meant to be a distinction. Every expression has an extended scope. Borrow expressions' operands use their extended scope as their temporary scope; as such, they're said to have an extended temporary scope (though it looks like I forgot to add that definition). The initializer expression of a `let` statement with an extending pattern also has an extended temporary scope.

I can think of a few things that might help:
- Actually defining "extended temporary scope". That term is a holdover from the stable version of the temporary lifetime extension section, where it was more obvious what it meant. It might still be obvious enough here if we can get rid of the other too-similar terms, and under the definition I proposed above it might be confusing that the "extended temporary scope" of an expression can sometimes be its normal temporary scope.
- Removing the term "extended temporary scope". `sub-expressions` is e.g. saying that the tuple operand expression of a tuple indexing expression has the same temporary scope as the tuple indexing expression. We don't need a notion of "extended temporary scope" to express that.
- Renaming the term "extended scope". The term I used in the compiler is "`extended_parent` scope" to match with the notions of "`parent` scope" and "`var_parent` scope" used there. That feels too close to "extending parent" though, maybe something like "extended ancestor scope" would be fine? I'm not sure.
- Removing the term "extended scope". e.g. a previous iteration of this PR defined the "extending ancestor" of an expression as the closest non-extending expression above that expression, then defined the scopes of borrow expressions' operands and super temporaries in terms of its temporary scope. It felt messier, so I ended making rules to determine its temporary scope directly (as the "extended scope" of an expression).

With this in mind, there isn't intended to be any ambiguity or a need to determine precedence between `exprs` and `sub-expressions`. The rules should be written in a way that makes everything clear. They just haven't quite gotten there yet.

src/destructors.md
dianne: I'm pretty familiar with inference rules; if it would help to discuss them in that way, I could write them out for different variants of lifetime extension (e.g. the stable version, the compiler PR version where `exprs.other` uses the `enclosing` scope of the extended parent, the version we've been talking about where `exprs.other` uses the temporary scope of the extended parent, etc.).

traviscross: That would definitely be helpful.  I could see including that directly in the Reference as well.

dianne: I'm having a bit of trouble coming up with good notation for this. When formalizing the relation "the temporary scope of `$expr` is `$scope`", `$expr` isn't really an expression, but the context in which the expression appears. Scopes, likewise, are contexts. This isn't a problem for the formalism, but it's more difficult for human-readability and understanding when the relation is between context stacks (or worse, zippers; or between a stack and an index into it; etc.). Any ideas on what would be most understandable for sake of communicating these rules?

If there's a good way to talk about contexts, maybe it could be helpful for clarifying the Reference text as well? The scoping rules are all about expressions' contexts, but in their text we tend not to distinguish between expressions and their contexts. E.g. when we say that the tail expression of a block is an extending expression, "the tail expression of a block" is a context, but we refer to it through the expression that goes there.

And regarding tools that walk through the analysis of an expression, it's not exactly purpose-built, but would it be helpful enough to have the formalism in an interactive proof assistant like Rocq or Lean? You could set a goal like "there exists a `scope` such that the temporary scope of `$ctxt` is `scope`" and use a tactic to apply the next applicable rule at each step. That's probably easier to make than a website (though less convenient to use for comparing the rulesets).

traviscross: Regarding contexts and how to formalize it, it's a good question and distinction.  Will think about it.  No immediate thoughts (other than the simplistic sketch above that approaches it syntactically and tracks the current category of the temporary scope and extended scope).  Probably it'd help to see sketches of some of the things you've considered or tried.

Regarding tooling, using Rocq or Lean could be interesting if we want to prove general properties about the rules, e.g. the assertion about the consistency of the relative drop order.  What might other interesting general properties to prove be?

For just applying the rules step-by-step to particular examples, perhaps a more `a-mir-formality`-style approach would also work.  I.e., perhaps it could be modeled in Rust, but the UI doesn't need to be fancy.

dianne: > Regarding contexts and how to formalize it, it's a good question and distinction. Will think about it. No immediate thoughts (other than the simplistic sketch above that approaches it syntactically and tracks the current category of the temporary scope and extended scope). Probably it'd help to see sketches of some of the things you've considered or tried.

The simplest formalization of a context I've thought of is effectively a stack of an expression's (or scope's) ancestors e.g. in
```rust
if cond() { &temp() } else { local }
```
the relevant syntactic context of `temp()` would be something like the concatenation of:
- The context of the `if` expression
- The consequent: `if … _ else …`
- The block tail: `{ …; _ }`
- The borrow expression: `&_`

without any of its ancestors' other subexpressions. This works great for the formalism, but it's a bit awkward to render for human consumption. Representing it as a list (e.g. `$parent, if … _ else …, { _ }, &_`) makes it difficult to visualize, but representing it as a partial expression with a hole (e.g. `$parent⟦if … { &_ } else …⟧`) makes the formalism less clear and requires inventing more notation. Maybe that's fine though?

For scopes, the context we want is slightly different, since e.g. there's a scope above both the condition and consequent which we'd want to be able to use as the scope of some temporaries, but we could get away with representing syntactic contexts and scopes the same way if we included all the information necessary for both. Then, since the drop scope of an expression would be a prefix of its context, we could represent it instead as an index into its context.

There's other ways to do contexts that include more information (e.g. through ["differentiation"](https://en.wikipedia.org/wiki/Zipper_(data_structure)#Contexts_and_differentiation)) but for drop scopes I don't think we need anything other than what's directly above an expression, so a stack/list should suffice.

> Regarding tooling, using Rocq or Lean could be interesting if we want to prove general properties about the rules, e.g. the assertion about the consistency of the relative drop order. What might other interesting general properties to prove be?

A really simple one: every expression context has a unique drop scope; i.e. scoping is decidable and deterministic. I'm not sure we need a tool for this, though. Those can both be eyeballed and we'd get determinism by construction writing it as a program rather than as inference rules.

As for more compelling properties, I'm honestly not sure. Relative drop order consistency is a nice property in theory, but practically, if you can compare runtime drop order then your program's compiling under stable temporary scoping rules anyway; iirc the differences in dynamic semantics are mostly observable when discarding a borrow or when dealing with raw pointers. The property I'm really looking for is more along the lines of "extending expressions preserve the temporary scopes of the operands of borrow operators" but I expect that, once formalized, that statement would look exactly like the rules for extending expressions, so it probably wouldn't need a formal proof either.

> For just applying the rules step-by-step to particular examples, perhaps a more `a-mir-formality`-style approach would also work. I.e., perhaps it could be modeled in Rust, but the UI doesn't need to be fancy.

That's true, though I do consider any UI that requires writing a parser to be at least slightly fancy. If that's what would be most helpful, I'll look into it. I don't mind putting in some extra effort if it ends up helping make this clearer ^^

traviscross: The key bit, if we could find it, is just having some more regular and concise way to reason about it.  That the notation is a tool for clear thinking for us is more important than having any tooling.

Even such notation, though, is a nice-to-have.  I don't want to distract us too much from working out good text here.  If we can find good notation that's helpful for understanding, communication, and working out clear text, then great.  If not, we'll iterate on the text and come up with something clear, and that'll be great too.

@RalfJung, do you perhaps have any ideas about how to express this more formally?  We're essentially talking about how to write, operationally, how the temporary scope is determined for the context of some expression.

@nikomatsakis, similar question, based on your work with `a-mir-formality`.

The relevant Reference sections are:

- https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope
- https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.temporary
- https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.lifetime-extension
- https://doc.rust-lang.org/nightly/reference/destructors.html#extending-based-on-expressions

RalfJung: I'd love that section to be written more formally, I currently find it very hard to understand.^^

In terms of how to express it, I am not sure I have much to add to what @dianne said. I would represent contexts as "expressions with a hole"; in Rocq this is a lot of work as you need to define an entirely new type, but somewhat informally we can just use the already existing expressions and put a hole somewhere. @dianne, you said this would require inventing more notation than an explicit stack -- why that? Other than picking a symbol for the hole, what notation do you need? I am not sure about using `_` for the hole since that is also often used for "does not matter", so I will use ● which is also the symbol we use in papers. My rather naive first impression is that the main thing we have to define is the grammar of permitted contexts -- that doesn't even need inference rules, it just needs to say which "expressions with a hole" we can put around an `&_` and have it be lifetime extended to the outside of that context:
```
lifetime-extension-ctx ::= ● | (..., lifetime-extension-ctx, ...) | { ...; lifetime-extension-ctx }
```
This grammar then makes it so that `(5, ●, g(16))` is a valid lifetime extension context, but `f(●, 33)` is not. I know there's more to lifetime extension than this context, in particular when considering `ref` patterns, but it seems to me that this would be the core of it. Or maybe I am just way too naive about how complicated this really is.

This is not operational at all, but I don't think of lifetime extension as something operational, so this is par for the course I would say. (The operational version would be... writing out the actual algorithm that computes when exactly which temporary gets dropped? I can't imagine that being nice or simple. Just because I think the runtime semantics of the program should be expressed operationally doesn't mean I think *everything* should be expressed operationally. :)

I'm not entirely sure what a "scope" would be formalized as, I just don't know lifetime extension well enough.

dianne: > @dianne, you said this would require inventing more notation than an explicit stack -- why that? Other than picking a symbol for the hole, what notation do you need?

The remaining bit of notation I was thinking of is for concatenation/substitution/composition of contexts, in order to talk about a local context (which we examine) within a larger context (which we assign a name). e.g., for determining the temporary scope of a borrow expression's operand, the object we're determining the temporary scope of is `&●`-within-some-larger-context. I assume this is a standard operation (especially as concerns going from "an expression with a hole" to "the expression with the hole filled in"), so it's likely there's some standard notation I'm not aware of. My first guesses would be some sort of brackets to evoke application, ∘ for composition, or ⧺ for concatenation.

> My rather naive first impression is that the main thing we have to define is the grammar of permitted contexts -- that doesn't even need inference rules, it just needs to say which "expressions with a hole" we can put around an `&_` and have it be lifetime extended to the outside of that context:

Defining a grammar of permitted contexts works, but the trick is that the scope of lifetime-extended temporaries depends on the context in which the lifetime extension context appears, so I'd like a way to capture that formally as well (hence reaching for composition). For stable Rust, we can be somewhat ad-hoc about this: if the extending context is the initializer of a `let` statement, we extend to the block the `let` statement is in; if the extending context is a `const`/`static` body, we extend to outside of the program's runtime. For the new lifetime extension semantics I'm proposing though, under this framing, I'd like for a lifetime extension context to be able to start in any non-lifetime-extension context, in which case it extends temporaries to the temporary scope enclosing the non-lifetime-extension context. This is so, e.g., the popular patterns
```rust
print!("{}", if cond() { &format!(...) } else { "..." });
```
and
```rust
print!("{}", if cond() { format_args!(...) } else { format_args!(...) });
```
work. Currently, since the `&format!(...)` and `format_args!(...)` are not in extending contexts, their temporaries are dropped within the `if` expression's block; they would however be extended if the `if` expression was directly made the initializer of a `let` statement. Under my proposed semantics, the `&format!(...)` and `format_args!(...)` would be in extending contexts that end at the argument to `print!`, so they'd be dropped in the temporary scope enclosing that context: the statement (a yet larger context).

Another way to view my proposal, which reflects the intuition I'd propose to capture in the Reference, is that temporary scope of the operand of a borrow operator is always determined by lifetime extension rules: you remove the longest extending suffix from its context before determining its temporary scope. Under this view, formally, I'm imagining lifetime extension would just be part of the definition of "the temporary scope of a context", though for pedagogical reasons they should maybe still be presented separately in the Reference.

> I'm not entirely sure what a "scope" would be formalized as, I just don't know lifetime extension well enough.

I'd say it could be viewed as a syntactic context as well, but with some semantics bolted on (when execution leaves the context, values associated with it are dropped). For the most part scopes do correspond to expressions' contexts, but you'd probably need more notation, e.g. to represent scopes that don't correspond directly to single AST objects, such as "The pattern-matching condition(s) and consequent body of `if`"; that one can't clearly be written as an expression with a hole, even though formally it's not that different.

RalfJung: > The remaining bit of notation I was thinking of is for concatenation/substitution/composition of contexts, in order to talk about a local context (which we examine) within a larger context (which we assign a name). e.g., for determining the temporary scope of a borrow expression's operand, the object we're determining the temporary scope of is &●-within-some-larger-context. I assume this is a standard operation (especially as concerns going from "an expression with a hole" to "the expression with the hole filled in"), so it's likely there's some standard notation I'm not aware of. My first guesses would be some sort of brackets to evoke application, ∘ for composition, or ⧺ for concatenation.

The standard notation is: if `K` is a context and `e` an expression, then `K[e]` is "fill `e` into the hole of `K`", and if `K1` and `K2` are both contexts then `K1[K2]` is what you call "concatenation", i.e. fill `K2` into the hole of `K1`. That's what I would use, anyway; `K1 ∘ K2` and `K1 ⧺ K2` would also be perfectly understandable.

That said, this may clash with Rust's use of `[` for its own syntax, so I like the `⟦` you used above.

> I'd say it could be viewed as a syntactic context as well, but with some semantics bolted on (when execution leaves the context, values associated with it are dropped). For the most part scopes do correspond to expressions' contexts, but you'd probably need more notation, e.g. to represent scopes that don't correspond directly to single AST objects, such as "The pattern-matching condition(s) and consequent body of if"; that one can't clearly be written as an expression with a hole, even though formally it's not that different.

Nice, I like that.

>  For stable Rust, we can be somewhat ad-hoc about this: if the extending context is the initializer of a let statement, we extend to the block the let statement is in; if the extending context is a const/static body, we extend to outside of the program's runtime. For the new lifetime extension semantics I'm proposing though, under this framing, I'd like for a lifetime extension context to be able to start in any non-lifetime-extension context, in which case it extends temporaries to the temporary scope enclosing the non-lifetime-extension context.

Oh I see. I can't immediately grasp the full set of consequences from this, but it does sound like a neat idea.

> Currently, since the &format!(...) and format_args!(...) are not in extending contexts, their temporaries are dropped within the if expression's block; they would however be extended if the if expression was directly made the initializer of a let statement. Under my proposed semantics, the &format!(...) and format_args!(...) would be in extending contexts that end at the argument to print!, so they'd be dropped in the temporary scope enclosing that context: the statement (a yet larger context).

Let me try to rephrase to make sure I understand: `print!("{}", &foo())` would work (with the result of `foo()` being dropped at the end of the surrounding statement), and hence you suggest that `print!("{}", if _ { &foo() } else { "bar" })` should also work, since the stuff that's "in between" the new and old location of the `&` already permits lifetime extension and therefore should not affect the lifetime of `&foo()`?

dianne: > Let me try to rephrase to make sure I understand: `print!("{}", &foo())` would work (with the result of `foo()` being dropped at the end of the surrounding statement), and hence you suggest that `print!("{}", if _ { &foo() } else { "bar" })` should also work, since the stuff that's "in between" the new and old location of the `&` already permits lifetime extension and therefore should not affect the lifetime of `&foo()`?

Exactly, yeah.

dianne: @traviscross I finally got around to writing up a comparison of temporary scoping rules: https://hackmd.io/@eN-bELC9R9WU10GfhlXaGw/Sy48pmDxbx

Since temporary scoping is total, deterministic, and has a single output, I felt defining it as a function would be more appropriate than defining it through inference rules. As such, I ended up writing it in literate Agda as a compromise between hopefully being readable[^1] and having nice formal properties. It's not intended to require an understanding of Agda to read, so hopefully it makes enough sense with the comments? This also meant I could prove the assertions I made, though I didn't include the proofs since they don't help readability. Let me know if there's anything that could use additional clarification, if it raises any questions, if there's something you'd like framed in the terms of that document, etc.

I'll try and update this PR soon to bring it back in line with rust-lang/rust#146098 and tidy up the presentation a bit (and think more on how to improve the terminology).

[^1]: HackMD doesn't seem to have Agda syntax highlighting so I set it to be highlighted like Haskell. Hopefully close enough?

src/destructors.md
dianne: Three updates [(diff)](https://github.com/rust-lang/reference/pull/2051#commits-pushed-a972545):
- I've brought the spec back in line with the behavior implemented by  rust-lang/rust#146098
- I've renamed "extended scope" to "borrow scope". I think this is fitting and less prone to ambiguity. Of course, there's some borrowed temporaries still that don't have temporary scopes defined in terms of their borrow scopes, so it's not a perfect name, but I think their temporary scopes and borrow scopes happen to coincide in that case anyway. It may deserve an admonition if that's the final name we go with. I'll keep thinking about potential better names too.
- I've added an extremely rough draft of a new take on `destructors.scope.lifetime-extension.sub-expressions` that hews closer to the Agda model. I'll have to think about how to integrate it with the rest of the text, but it reflects how I'd like to approach it.

src/destructors.md
traviscross: I've now had a close look at this.  It's really great.  I've added a proof of the conjecture that `∀ e → tempScope-stable e ≡ tempScope-stable′ e`, added an introduction and reading guide for Rust users, and massaged it into a literate Agda file and appendix chapter.  It's still in rough shape -- it's not getting syntax highlighting for the Agda; CI isn't checking it; we're speaking about stable behavior vs a PR when the Reference just needs to speak to the stable behavior (inclusive of the PR), etc.  But I pushed it up to the rebased branch for discussion, as it would be good to find a way to actually get this into the appendix.

traviscross: Probably the thing for us to do, actually, would be to add this as a Reference appendix chapter in a separate PR with only the currently-stable behavior.  Then this PR could update the behavior in the model accordingly.

(I'll talk the logistics of this over with @ehuss.)

traviscross: How would you state the conjecture exactly for proving that `tempScope-refPr′` comprises the separate rules?

dianne: If I correctly understand your question, the thing you want to prove is `∀ e → tempScope-refPr e ≡ tempScope-refPr′ e`; that states that the two formulations give you the same temporary scope for all expression contexts. The way I did that locally uses a lemma about `refScope-refPr` proving it's equal to the other formulation's way of expressing borrow scopes.

```agda
tempScope≡tempScope′ : ∀ e → tempScope-refPr e ≡ tempScope-refPr′ e
refScope≡refScope′ : ∀ e → refScope-refPr e ≡ tempScope-refPr′ (e ▷ &●)
```

Then, the proofs of both can be written as mutually recursive functions. I didn't organize or comment it, but it's just case analysis and induction (ideally it'd be automated):

```agda
tempScope≡tempScope′ (e ▷ |…|●) = refl
tempScope≡tempScope′ (e ▷ const…=●⨾) = refl
tempScope≡tempScope′ (e ▷ let…=●⨾ pat-ext) = refl
tempScope≡tempScope′ (e ▷ let…=●⨾ pat-non) = refl
tempScope≡tempScope′ (e ▷ ●⨾) = refl
tempScope≡tempScope′ (e ▷ ⦃…⨾●⦄) = refl
tempScope≡tempScope′ (e ▷ …⦅…,●,…⦆ n) = refl
tempScope≡tempScope′ (e ▷ ⟦…,●,…⟧ n) = refl
tempScope≡tempScope′ (e ▷ ●⟦…⟧) = tempScope≡tempScope′ e
tempScope≡tempScope′ (e ▷ |…|● ▷ &●) = refl
tempScope≡tempScope′ (e ▷ const…=●⨾ ▷ &●) = refl
tempScope≡tempScope′ (e ▷ let…=●⨾ pat ▷ &●) = refl
tempScope≡tempScope′ (e ▷ ●⨾ ▷ &●) = refl
tempScope≡tempScope′ (e ▷ ⦃…⨾●⦄ ▷ &●) = refScope≡refScope′ e
tempScope≡tempScope′ (e ▷ …⦅…,●,…⦆ n ▷ &●) = refl
tempScope≡tempScope′ (e ▷ ⟦…,●,…⟧ n ▷ &●) = refScope≡refScope′ e
tempScope≡tempScope′ (e ▷ ●⟦…⟧ ▷ &●) = tempScope≡tempScope′ e
tempScope≡tempScope′ (e ▷ &● ▷ &●) = refScope≡refScope′ e

refScope≡refScope′ (e ▷ |…|●) = refl
refScope≡refScope′ (e ▷ const…=●⨾) = refl
refScope≡refScope′ (e ▷ let…=●⨾ pat) = refl
refScope≡refScope′ (e ▷ ●⨾) = refl
refScope≡refScope′ (e ▷ ⦃…⨾●⦄) = refScope≡refScope′ e
refScope≡refScope′ (e ▷ …⦅…,●,…⦆ n) = refl
refScope≡refScope′ (e ▷ ⟦…,●,…⟧ n) = refScope≡refScope′ e
refScope≡refScope′ (e ▷ ●⟦…⟧) = tempScope≡tempScope′ e
refScope≡refScope′ (e ▷ &●) = refScope≡refScope′ e
```

theemathas:
Does the Agda model need to also cover temporaries created inside place expressions on the left hand side of assignment expressions?

dianne:
> Does the Agda model need to also cover temporaries created inside place expressions on the left hand side of assignment expressions?

I don't think so? To my knowledge, assignment expressions don't have any special-casing regarding expressions' temporary scopes[^1], but I could have overlooked something. I even left out things that do have sort of special behavior that's not sufficiently distinct; e.g. the operand of a `||` operator is both a temporary scope and non-extending, but those are modeled separately already so it's not particularly illustrative to include it.

If we want to model temporary scopes for all of Rust rather than just a representative subset that shows important cases, I think the approach should be different. As-is, anything extra being added to the model of Rust's syntax would mean lots of extra cases, hindering readability, especially if proofs are included. There's ways to deal with that; e.g. by adding abstractions, using macros/reflection, or using a different language that deals with tedious definitions/proofs better. But those all add some amount of complexity as well.

Another important omission is that the model doesn't actually say what expressions produce temporaries, what the temporaries are, or what their lifetimes will be if so; it just handles expressions' temporary scopes, which are used as the lifetimes of some temporaries. A lot more work would be needed to fully model that stuff.

[^1]: Destructuring assignments have weird temporary scoping, but we can assume those are desugared before temporary scopes are determined.

PR #2055, opened 2025-10-20
Add section on expansion-time (early) name resolution

yaahc:
currently mostly a skeleton of a draft so we can collaboratively massage it into shape more easily before filling in with proper reference verbiage.

hoping to take a significant chunk out of https://github.com/rust-lang/reference/issues/568

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -116,6 +116,7 @@ They may create bindings for:
 * [Built-in types]
 * [Attributes]
 * [Derive macros]
+* [`macro_rules`]
 
 r[items.use.path.disallowed]
 They cannot import [associated items], [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
@@ -302,6 +303,10 @@ mod clashing {
 }
 ```
 
+> [!NOTE]
+>
+> For areas where shadowing is not allowed, see [name resolution ambiguities].
+
 r[items.use.glob.last-segment-only]
 `*` cannot be used as the first or intermediate segments.
 
@@ -389,71 +394,19 @@ r[items.use.restrictions.variant]
   use TypeAlias::MyVariant; //~ ERROR
   ```
 
-r[items.use.ambiguities]
-## Ambiguities
-
-> [!NOTE]
-> This section is incomplete.
-
-r[items.use.ambiguities.intro]
-Some situations are an error when there is an ambiguity as to which name a `use` declaration refers. This happens when there are two name candidates that do not resolve to the same entity.
-
-r[items.use.ambiguities.glob]
-Glob imports are allowed to import conflicting names in the same namespace as long as the name is not used.
-For example:
-
-```rust
-mod foo {
-    pub struct Qux;
-}
-
-mod bar {
-    pub struct Qux;
-}
-
-use foo::*;
-use bar::*; //~ OK, no name conflict.
-
-fn main() {
-    // This would be an error, due to the ambiguity.
-    //let x = Qux;
-}
-```
-
-Multiple glob imports are allowed to import the same name, and that name is allowed to be used, if the imports are of the same item (following re-exports). The visibility of the name is the maximum visibility of the imports. For example:
-
-```rust
-mod foo {
-    pub struct Qux;
-}
-
-mod bar {
-    pub use super::foo::Qux;
-}
-
-// These both import the same `Qux`. The visibility of `Qux`
-// is `pub` because that is the maximum visibility between
-// these two `use` declarations.
-pub use bar::*;
-use foo::*;
-
-fn main() {
-    let _: Qux = Qux;
-}
-```
-
-[`extern crate`]: extern-crates.md
-[`macro_rules`]: ../macros-by-example.md
-[`self`]: ../paths.md#self
-[associated items]: associated-items.md
 [Attributes]: ../attributes.md
 [Built-in types]: ../types.md
 [Derive macros]: macro.proc.derive
 [Enum variants]: enumerations.md
+[`extern crate`]: extern-crates.md
+[`macro_rules`]: ../macros-by-example.md
+[`self`]: ../paths.md#self
+[associated items]: associated-items.md
 [extern prelude]: ../names/preludes.md#extern-prelude
 [generic parameters]: generics.md
 [items]: ../items.md
 [local variables]: ../variables.md
+[name resolution ambiguities]: names.resolution.expansion.imports.ambiguity
 [namespace]: ../names/namespaces.md
 [namespaces]: ../names/namespaces.md
 [paths]: ../paths.md
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -304,7 +304,6 @@ mod clashing {
 ```
 
 > [!NOTE]
->
 > For areas where shadowing is not allowed, see [name resolution ambiguities].
 
 r[items.use.glob.last-segment-only]
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -349,7 +349,7 @@ macro_rules! m {
     };
 }
 
-// Introduce m2 macro as path-based candidate.
+// Introduce `m2` macro as path-based candidate.
 //
 // This item is in scope for this entire example, not just below the
 // use declaration.
@@ -361,7 +361,6 @@ m!(); // prints "m\n"
 ```
 
 > [!NOTE]
->
 > For areas where shadowing is not allowed, see [name resolution ambiguities].
 
 r[macro.decl.scope.path-based]
@@ -370,22 +369,22 @@ r[macro.decl.scope.path-based]
 r[macro.decl.scope.path-based.intro]
 By default, a macro has no path-based scope. Macros can gain path-based scope in two ways:
 
-* [Use declaration re-export]
-* [`macro_export`]
+- [Use declaration re-export]
+- [`macro_export`]
 
 r[macro.decl.scope.path.reexport]
 Macros can be re-exported to give them path-based scope from a module other than the crate root.
 
 ```rust
-mac::m!(); // OK: Path-based lookup finds m in the mac module.
+mac::m!(); // OK: Path-based lookup finds `m` in the mac module.
 
 mod mac {
-    // Introduce macro m with textual scope.
+    // Introduce macro `m` with textual scope.
     macro_rules! m {
         () => {};
     }
 
-    // Re-export with path-based scope from within m's textual scope.
+    // Reexport with path-based scope from within `m`'s textual scope.
     pub(crate) use m;
 }
 ```
@@ -815,10 +814,8 @@ expansions, taking separators into account. This means:
 
 For more detail, see the [formal specification].
 
-[Hygiene]: #hygiene
 [Metavariables]: #metavariables
 [Repetitions]: #repetitions
-[Use declaration re-export]: items/use-declarations.md#use-visibility
 [`macro_export`]: #the-macro_export-attribute
 [`$crate`]: macro.decl.hygiene.crate
 [`extern crate self`]: items.extern-crate.self
@@ -829,3 +826,4 @@ For more detail, see the [formal specification].
 [loop labels]: expressions/loop-expr.md#loop-labels
 [name resolution ambiguities]: names/name-resolution.md#r-names.resolution.expansion.imports.ambiguity
 [token]: tokens.md
+[use declaration re-export]: items/use-declarations.md#use-visibility
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -326,6 +326,108 @@ fn foo() {
 // m!(); // Error: m is not in scope.
 ```
 
+r[macro.decl.scope.textual.shadow.path-based]
+Textual scope name bindings for macros shadow path-based scope bindings to macros.
+
+```rust
+macro_rules! m2 {
+    () => {
+        println!("m2");
+    };
+}
+
+// Resolves to path-based candidate from use declaration below.
+m!(); // prints "m2\n"
+
+// Introduce second candidate for `m` with textual scope.
+//
+// This shadows path-based candidate from below for the rest of this
+// example.
+macro_rules! m {
+    () => {
+        println!("m");
+    };
+}
+
+// Introduce m2 macro as path-based candidate.
+//
+// This item is in scope for this entire example, not just below the
+// use declaration.
+use m2 as m;
+
+// Resolves to the textual macro candidate from above the use
+// declaration.
+m!(); // prints "m\n"
+```
+
+> [!NOTE]
+>
+> For areas where shadowing is not allowed, see [name resolution ambiguities].
+
+r[macro.decl.scope.path-based]
+### Path-based scope
+
+r[macro.decl.scope.path-based.intro]
+By default, a macro has no path-based scope. Macros can gain path-based scope in two ways:
+
+* [Use declaration re-export]
+* [`macro_export`]
+
+r[macro.decl.scope.path.reexport]
+Macros can be re-exported to give them path-based scope from a module other than the crate root.
+
+```rust
+mac::m!(); // OK: Path-based lookup finds m in the mac module.
+
+mod mac {
+    // Introduce macro m with textual scope.
+    macro_rules! m {
+        () => {};
+    }
+
+    // Re-export with path-based scope from within m's textual scope.
+    pub(crate) use m;
+}
+```
+
+r[macro.decl.scope.path-based.visibility]
+Macros have an implicit visibility of `pub(crate)`. `#[macro_export]` changes the implicit visibility to `pub`.
+
+```rust
+// Implicit visibility is `pub(crate)`.
+macro_rules! private_m {
+    () => {};
+}
+
+// Implicit visibility is `pub`.
+#[macro_export]
+macro_rules! pub_m {
+    () => {};
+}
+
+pub(crate) use private_m as private_macro; // OK.
+pub use pub_m as pub_macro; // OK.
+```
+
+```rust,compile_fail,E0364
+# // Implicit visibility is `pub(crate)`.
+# macro_rules! private_m {
+#     () => {};
+# }
+#
+# // Implicit visibility is `pub`.
+# #[macro_export]
+# macro_rules! pub_m {
+#     () => {};
+# }
+#
+# pub(crate) use private_m as private_macro; // OK.
+# pub use pub_m as pub_macro; // OK.
+#
+pub use private_m; // ERROR: `private_m` is only public within
+                   // the crate and cannot be re-exported outside.
+```
+
 <!-- template:attributes -->
 r[macro.decl.scope.macro_use]
 ### The `macro_use` attribute
@@ -713,14 +815,17 @@ expansions, taking separators into account. This means:
 
 For more detail, see the [formal specification].
 
+[Hygiene]: #hygiene
+[Metavariables]: #metavariables
+[Repetitions]: #repetitions
+[Use declaration re-export]: items/use-declarations.md#use-visibility
+[`macro_export`]: #the-macro_export-attribute
+[`$crate`]: macro.decl.hygiene.crate
 [`extern crate self`]: items.extern-crate.self
 [`macro_use` prelude]: names/preludes.md#macro_use-prelude
 [block labels]: expr.loop.block-labels
 [delimiters]: tokens.md#delimiters
 [formal specification]: macro-ambiguity.md
-[Hygiene]: #hygiene
 [loop labels]: expressions/loop-expr.md#loop-labels
-[Metavariables]: #metavariables
-[Repetitions]: #repetitions
+[name resolution ambiguities]: names/name-resolution.md#r-names.resolution.expansion.imports.ambiguity
 [token]: tokens.md
-[`$crate`]: macro.decl.hygiene.crate
--- a/src/macros.md
+++ b/src/macros.md
@@ -106,15 +106,25 @@ macro_rules! example {
 example!();
 ```
 
+r[macro.invocation.name-resolution]
+
+Macros invocations can be resolved via two kinds of scopes:
+
+* Textual Scope
+    * [Textual scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.textual)
+* Path-based scope
+    * [Path-based scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.path-based)
+    * [Procedural macros]
+
+[External blocks]: items/external-blocks.md
 [Macros by Example]: macros-by-example.md
 [Procedural Macros]: procedural-macros.md
+[`macro_rules`]: macros-by-example.md
 [associated items]: items/associated-items.md
 [delimiters]: tokens.md#delimiters
 [expressions]: expressions.md
 [items]: items.md
-[`macro_rules`]: macros-by-example.md
 [patterns]: patterns.md
 [statements]: statements.md
 [types]: types.md
 [visibility qualifiers]: visibility-and-privacy.md
-[External blocks]: items/external-blocks.md
--- a/src/macros.md
+++ b/src/macros.md
@@ -110,11 +110,11 @@ r[macro.invocation.name-resolution]
 
 Macros invocations can be resolved via two kinds of scopes:
 
-* Textual Scope
-    * [Textual scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.textual)
-* Path-based scope
-    * [Path-based scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.path-based)
-    * [Procedural macros]
+- Textual Scope
+  - [Textual scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.textual)
+- Path-based scope
+  - [Path-based scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.path-based)
+  - [Procedural macros]
 
 [External blocks]: items/external-blocks.md
 [Macros by Example]: macros-by-example.md
--- a/src/names/name-resolution.md
+++ b/src/names/name-resolution.md
@@ -350,9 +350,6 @@ The available scope kinds are visited in the following order. Each of these scop
 > [!EDITION-2018]
 > Starting in edition 2018 the `#[macro_use]` prelude is not visited when [`#[no_implicit_prelude]`][names.preludes.no_implicit_prelude] is present.
 
-r[names.resolution.expansion.macros.derive-helpers]
-Derive helper scopes are not visited when resolving derive macros in the parent scope (starting scope).
-
 r[names.resolution.expansion.macros.reserved-names]
 The names `cfg` and `cfg_attr` are reserved in the macro attribute [sub-namespace].
 
--- a/src/names/name-resolution.md
+++ b/src/names/name-resolution.md
@@ -1,4 +1,553 @@
+r[names.resolution]
 # Name resolution
 
+r[names.resolution.intro]
+_Name resolution_ is the process of tying paths and other identifiers to the declarations of those entities. Names are segregated into different [namespaces], allowing entities in different namespaces to share the same name without conflict. Each name is valid within a [scope], or a region of source text where that name may be referenced. Access to a name may be restricted based on its [visibility].
+
+Name resolution is split into three stages throughout the compilation process. The first stage, *expansion-time resolution*, resolves all [`use` declarations] and [macro invocations]. The second stage, *primary resolution*, resolves all names that have not yet been resolved and that do not depend on type information to resolve. The last stage, *type-relative resolution*, resolves the remaining names once type information is available.
+
 > [!NOTE]
-> This is a placeholder for future expansion.
+> Expansion-time resolution is also known as *early resolution*. Primary resolution is also known as *late resolution*.
+
+r[names.resolution.general]
+## General
+
+r[names.resolution.general.intro]
+The rules within this section apply to all stages of name resolution.
+
+r[names.resolution.general.scopes]
+### Scopes
+
+r[names.resolution.general.scopes.intro]
+> [!NOTE]
+> This is a placeholder for future expansion about resolution of names within various scopes.
+
+r[names.resolution.expansion]
+## Expansion-time name resolution
+
+r[names.resolution.expansion.intro]
+Expansion-time name resolution is the stage of name resolution necessary to complete macro expansion and fully generate a crate's [AST]. This stage requires the resolution of macro invocations and `use` declarations. Resolving `use` declarations is required for macro invocations that resolve via [path-based scope]. Resolving macro invocations is required in order to expand them.
+
+r[names.resolution.expansion.unresolved-invocations]
+After expansion-time name resolution, the AST must not contain any unexpanded macro invocations. Every macro invocation resolves to a valid definition that exists in the final AST or in an external crate.
+
+```rust,compile_fail
+m!(); // ERROR: Cannot find macro `m` in this scope.
+```
+
+r[names.resolution.expansion.expansion-order-stability]
+The resolution of names must be stable. After expansion, names in the fully expanded AST must resolve to the same definition regardless of the order in which macros are expanded and imports are resolved.
+
+r[names.resolution.expansion.speculation]
+All name resolution candidates selected during macro expansion are considered speculative. Once the crate has been fully expanded, all speculative import resolutions are validated to ensure that macro expansion did not introduce any new ambiguities.
+
+> [!NOTE]
+> Due to the iterative nature of macro expansion, this causes so-called time traveling ambiguities, such as when a macro or glob import introduces an item that is ambiguous with its own base path.
+>
+> ```rust,compile_fail,E0659
+> # fn main() {}
+> macro_rules! f {
+>     () => {
+>         mod m {
+>             pub(crate) use f;
+>         }
+>     }
+> }
+> f!();
+>
+> const _: () = {
+>     // Initially, we speculatively resolve `m` to the module in
+>     // the crate root.
+>     //
+>     // Expansion of `f` introduces a second `m` module inside this
+>     // body.
+>     //
+>     // Expansion-time resolution finalizes resolutions by re-
+>     // resolving all imports and macro invocations, sees the
+>     // introduced ambiguity and reports it as an error.
+>     m::f!(); // ERROR: `m` is ambiguous.
+> };
+> ```
+
+r[names.resolution.expansion.imports]
+### Imports
+r[names.resolution.expansion.imports.intro]
+All `use` declarations are fully resolved during this stage of resolution. [Type-relative paths] cannot be resolved at this stage and will produce an error.
+
+```rust
+mod m {
+    pub const C: () = ();
+    pub enum E { V }
+    pub type A = E;
+    impl E {
+        pub const C: () = ();
+    }
+}
+
+// Valid imports resolved at expansion-time:
+use m::C; // OK.
+use m::E; // OK.
+use m::A; // OK.
+use m::E::V; // OK.
+
+// Valid expressions resolved during type-relative resolution:
+let _ = m::A::V; // OK.
+let _ = m::E::C; // OK.
+```
+
+```rust,compile_fail,E0432
+# mod m {
+#     pub const C: () = ();
+#     pub enum E { V }
+#     pub type A = E;
+#     impl E {
+#         pub const C: () = ();
+#     }
+# }
+// Invalid type-relative imports that can't resolve at expansion-time:
+use m::E::C; // ERROR: Unresolved import `m::E::C`.
+use m::A::V; // ERROR: Unresolved import `m::A::V`.
+```
+
+r[names.resolution.expansion.imports.shadowing]
+Names introduced via `use` declarations in an [outer scope] are shadowed by candidates in the same namespace with the same name from an inner scope except where otherwise restricted by [name resolution ambiguities].
+
+```rust,no_run
+pub mod m1 {
+    pub mod ambig {
+        pub const C: u8 = 1;
+    }
+}
+
+pub mod m2 {
+    pub mod ambig {
+        pub const C: u8 = 2;
+    }
+}
+
+// This introduces the name `ambig` in the outer scope.
+use m1::ambig;
+fn f() {
+    // This shadows `ambig` in the inner scope.
+    use m2::ambig;
+    // The inner candidate is selected here
+    // as the resolution of `ambig`.
+    use ambig::C;
+    const { assert!(C == 2) };
+}
+```
+
+r[names.resolution.expansion.imports.shadowing.shared-scope]
+Shadowing of names introduced via `use` declarations within a single scope is permitted in the following situations:
+
+- [`use` glob shadowing]
+- [Macro textual scope shadowing]
+
+r[names.resolution.expansion.imports.ambiguity]
+#### Ambiguities
+
+r[names.resolution.expansion.imports.ambiguity.intro]
+There are certain situations during expansion-time resolution where there are multiple macro definitions, `use` declarations, or modules an import or macro invocation's name could refer to where the compiler cannot consistently determine which candidate should shadow the other. Shadowing cannot be permitted in these situations and the compiler instead emits ambiguity errors.
+
+r[names.resolution.expansion.imports.ambiguity.glob-vs-glob]
+Names may not be resolved through ambiguous glob imports. Glob imports are allowed to import conflicting names in the same namespace as long as the name is not used. Names with conflicting candidates from ambiguous glob imports may still be shadowed by non-glob imports and used without producing an error. The errors occur at time of use, not time of import.
+
+For example:
+
+```rust,compile_fail,E0659
+mod m1 {
+    pub struct Ambig;
+}
+
+mod m2 {
+    pub struct Ambig;
+}
+
+// OK: This brings conficting names into scope and namespace
+// but they have not been used yet.
+use m1::*;
+use m2::*;
+
+fn ambiguous_use() {
+    let x = Ambig; // ERROR: `Ambig` is ambiguous.
+}
+```
+
+```rust
+# mod m1 {
+#     pub struct Ambig;
+# }
+#
+# mod m2 {
+#     pub struct Ambig;
+# }
+#
+# use m1::*;
+# use m2::*; // OK: No name conflict.
+fn ambiguous_shadow() {
+    // This is permitted, since resolution is not through the ambiguous globs.
+    struct Ambig;
+    let x = Ambig;
+}
+```
+
+Multiple glob imports are allowed to import the same name, and that name is allowed to be used if the imports are of the same item (following re-exports). The visibility of the name is the maximum visibility of the imports.
+
+```rust
+mod m1 {
+    pub struct Ambig;
+}
+
+mod m2 {
+    // This re-exports the same `Ambig` item through a second module.
+    pub use super::m1::Ambig;
+}
+
+// These both import the same `Ambig`.
+//
+// The visibility of `Ambig` is `pub` because that is the
+// maximum visibility between these two `use` declarations.
+mod m3 {
+    pub use super::m1::*;
+    use super::m2::*;
+}
+
+fn main() {
+    // `Ambig` can be used through the m3
+    // globs and still has `pub` visibility.
+    let _ = m3::Ambig;
+}
+```
+
+r[names.resolution.expansion.imports.ambiguity.glob-vs-outer]
+Names in imports and macro invocations may not be resolved through glob imports when there is another candidate available in an [outer scope].
+
+```rust,compile_fail,E0659
+mod glob {
+    pub mod ambig {
+        pub struct Name;
+    }
+}
+
+// Outer `ambig` candidate.
+pub mod ambig {
+    pub struct Name;
+}
+
+const _: () = {
+    // Cannot resolve `ambig` through this glob
+    // because of the outer `ambig` candidate above.
+    use glob::*;
+    use ambig::Name; // ERROR: `ambig` is ambiguous.
+};
+```
+
+```rust,compile_fail,E0659
+use m::f as ambig;
+pub mod m {
+    macro_rules! f {
+        () => {};
+    }
+    pub(crate) use f;
+}
+pub mod glob {
+    macro_rules! f {
+        () => {};
+    }
+    pub(crate) use f as ambig;
+}
+
+const _: () = {
+    use glob::*;
+    ambig!(); // ERROR: `ambig` is ambiguous.
+};
+```
+
+> [!NOTE]
+> These ambiguity errors are specific to expansion-time resolution. Having multiple candidates available for a given name during later stages of resolution is not considered an error. So long as none of the imports themselves are ambiguous, there will always be a single unambiguous closest resolution.
+>
+> ```rust
+> mod glob {
+>     pub const AMBIG: bool = true;
+> }
+>
+> mod outer {
+>     pub const AMBIG: bool = false;
+> }
+>
+> use outer::AMBIG;
+>
+> pub fn f() {
+>     use glob::*;
+>     assert!(AMBIG);
+>     //      ^--- This `AMBIG` is resolved during primary resolution.
+> }
+> ```
+
+r[names.resolution.expansion.imports.ambiguity.path-vs-textual-macro]
+Names may not be resolved through ambiguous macro re-exports. Macro re-exports are ambiguous when they would shadow a textual macro candidate for the same name in an [outer scope].
+
+```rust,compile_fail,E0659
+// Textual macro candidate.
+macro_rules! ambig {
+    () => {}
+}
+
+// Path-based macro candidate.
+macro_rules! path_based {
+    () => {}
+}
+
+pub fn f() {
+    // This re-export of the `path_based` macro definition
+    // as `ambig` may not shadow the `ambig` macro definition
+    // which is resolved via textual macro scope.
+    use path_based as ambig;
+    ambig!(); // ERROR: `ambig` is ambiguous.
+}
+```
+
+> [!NOTE]
+> This restriction is needed due to implementation details in the compiler,
+> specifically the current scope visitation logic and the complexity of supporting
+> this behavior. This ambiguity error may be removed in the future.
+
+r[names.resolution.expansion.macros]
+### Macros
+
+r[names.resolution.expansion.macros.intro]
+Macros are resolved by iterating through the available scopes to find the available candidates. Macros are split into two sub-namespaces, one for function-like macros, and the other for attributes and derives. Resolution candidates from the incorrect sub-namespace are ignored.
+
+r[names.resolution.expansion.macros.visitation-order]
+The available scope kinds are visited in the following order. Each of these scope kinds represent one or more scopes.
+
+* [Derive helpers]
+* [Textual scope macros]
+* [Path-based scope macros]
+* [`macro_use` prelude]
+* [Standard library prelude]
+* [Builtin attributes]
+
+> [!NOTE]
+>
+> The compiler will attempt to resolve derive helpers that are used before their associated macro introduces them into scope. This scope is visited after the scope for resolving derive helper candidates that are correctly in scope. This behavior is slated for removal.
+>
+> For more info see [derive helper scope].
+
+> [!NOTE]
+>
+> This visitation order may change in the future, such as interleaving the
+> visitation of textual and path-based scope candidates based on their lexical
+> scopes.
+
+> [!EDITION-2018]
+>
+> Starting in edition 2018 the `#[macro_use]` prelude is not visited when `#[no_implicit_prelude]` is present.
+
+r[names.resolution.expansion.macros.derivehelpers]
+Derive helper scopes are not visited when resolving derive macros in the parent scope (starting scope).
+
+r[names.resolution.expansion.macros.reserved-names]
+The names `cfg` and `cfg_attr` are reserved in the macro attribute [sub-namespace].
+
+r[names.resolution.expansion.macros.ambiguity]
+#### Ambiguities
+
+r[names.resolution.expansion.macros.ambiguity.more-expanded-vs-outer]
+Names may not be resolved through ambiguous candidates inside of macro expansions. Candidates inside of macro expansions are ambiguous when they would shadow a candidate for the same name from outside of the first candidate's macro expansion and the invocation of the name being resolved is also from outside of the first candidate's macro expansion.
+
+```rust,compile_fail,E0659
+macro_rules! define_ambig {
+    () => {
+        macro_rules! ambig {
+            () => {}
+        }
+    }
+}
+
+// Introduce outer candidate definition for `ambig` macro invocation.
+macro_rules! ambig {
+    () => {}
+}
+
+// Introduce a second candidate definition for `ambig` inside of a macro
+// expansion.
+define_ambig!();
+
+// The definition of `ambig` from the second invocation
+// of `define_ambig` is the innermost canadidate.
+//
+// The definition of `ambig` from the first invocation of `define_ambig`
+// is the second candidate.
+//
+// The compiler checks that the first candidate is inside of a macro
+// expansion, that the second candidate is not from within the same
+// macro expansion, and that the name being resolved is not from
+// within the same macro expansion.
+ambig!(); // ERROR: `ambig` is ambiguous.
+```
+
+The reverse is not considered ambiguous.
+
+```rust
+# macro_rules! define_ambig {
+#     () => {
+#         macro_rules! ambig {
+#             () => {}
+#         }
+#     }
+# }
+// Swap order of definitions.
+define_ambig!();
+macro_rules! ambig {
+    () => {}
+}
+// The innermost candidate is now less expanded so it may shadow more
+// the macro expanded definition above it.
+ambig!();
+```
+
+Nor is it ambiguous if the invocation being resolved is within the innermost candidate's expansion.
+
+```rust
+macro_rules! ambig {
+    () => {}
+}
+
+macro_rules! define_and_invoke_ambig {
+    () => {
+        // Define innermost candidate.
+        macro_rules! ambig {
+            () => {}
+        }
+
+        // Invocation of `ambig` is in the same expansion as the
+        // innermost candidate.
+        ambig!(); // OK
+    }
+}
+
+define_and_invoke_ambig!();
+```
+
+It doesn't matter if both definitions come from invocations of the same macro, the outermost candidate is still considered "less expanded" because it is not within the expansion containing the innermost candidate's definition.
+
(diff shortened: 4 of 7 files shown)


src/names/name-resolution.md
```diff
@@ -1,4 +1,114 @@
+r[names.resolution]
```
yaahc, line 1: duplicate of a section in names.md, not sure which to change or how

src/names/name-resolution.md
```diff
@@ -1,4 +1,114 @@
...
+
+* shadowing and ambiguity may or may not represent the same section or one may be a subsection of the other
+
+r[names.resolution.early.imports.errors.ambiguity.builtin-attr]
+* it is an error to have a user defined attribute or derive macro with the same name as a builtin attribute (e.g. inline)
+r[names.resolution.early.imports.errors.ambiguity.derivehelper]
+* derive helpers used before their associated derive may not shadow other attributes or other derive helpers that are otherwise in scope after their derive
+r[names.resolution.early.imports.errors.ambiguity.textualvspathbasedscope]
+* path-based scope bindings for macros may not shadow textual scope bindings to macros
+* https://doc.rust-lang.org/nightly/reference/names/namespaces.html#r-names.namespaces.sub-namespaces.use-shadow
```
yaahc: The linked to section definitely needs to be moved and updated in some way. I think its subtly wrong as is, and also its named like one of the shadowing sections but this is actually a disallowed case of shadowing aka ambiguity error that is being documented.

src/names/name-resolution.md
```diff
@@ -1,4 +1,114 @@
...
+
+_Name resolution_ is the process of tying paths and other identifiers to the
+declarations of those entities. Names are segregated into different
+[namespaces], allowing entities in different namespaces to share the same name
+without conflict. Each name is valid within a [scope], or a region of source
+text where that name may be referenced. Access to certain names may be
+restricted based on their [visibility].
+
+* Names are resolved at three different stages of compilation.
+* [Macros] and [use declarations] are resolved during macro expansion.
```
yaahc: I feel like we're probably going to want to craft new terms of art for the various stages of name resolution. I think early resolution should be named in a way that indicates its required for macro expansion, type dependent name resolution is something I haven't really seen documented anywhere else but follows the same pattern, these are the names that depend on type information to be fully resolved.

Everything else is called "late resolution" but I hate that and would prefer to drop the late if possible.

src/names/name-resolution.md
```diff
@@ -1,4 +1,114 @@
...
+
+r[names.resolution.early]
+## Early name resolution
+
+r[names.resolution.early.intro]
+
+* early name resolution is the part of name resolution that happens during macro expansion
+* early name resolution includes the resolution of imports and macros
+* early name resolution is the minimum amount of resolution required to resolve macro invocations so they can be expanded.
+* resolving imports is necessary to resolve macro invocations (as of 2018 edition i think, pretty sure this applies specifically to path-based scope for macros)
```
yaahc: I'm not sure on this but my guess based on what I know so far is that function like proc macros have path based scope in all editions so resolving imports is probably still necessary in all editions. Most of my attention has been focused on MBE macros.

traviscross:
Thanks for posting.  Lot of good here.  The main thing I'd suggest, by way of helping to sharpen this up, would be to try to write a concise example after each claim, in as many cases as that might make sense, that demonstrates that the claim is true (and would not pass if the claim were false).  Aside from the intrinsic benefit of having such examples, I think this might help to focus the text on the language-level effects.  (It's not surprising, given the good research you've been doing, that some bits of this currently have some "description of the implementation" flavor.)

yaahc (commented):
a fair number of comments are also contained inline in the documents

src/items/use-declarations.md
```diff
@@ -398,10 +398,18 @@ r[items.use.ambiguities]
 r[items.use.ambiguities.intro]
 Some situations are an error when there is an ambiguity as to which name a `use` declaration refers. This happens when there are two name candidates that do not resolve to the same entity.
 
+* except where shadowing is allowed
+r[names.resolution.early.imports.errors.ambiguity.globvsglob]
+* it is an error to name an item through ambiguous use declarations
+    * two globs imports which both have an item matching that name where the items are different
```
yaahc: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=11d42f996c2a1b12c922973aed10229d

src/items/use-declarations.md
```diff
@@ -398,10 +398,18 @@ r[items.use.ambiguities]
...
+        * this is still an error even if there is a third non glob binding resolution to an item with the same name
```
yaahc: citation needed

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d185f6aae481aef85d316894adf3d3c6

I expected this to be dependent upon ordering. Much of the ambiguity checking logic saves the first resolution result and checks that against subsequent results, which can end up suppressing errors that would have occured had the 2nd and 3rd results been compared against each other. I'm honestly surprised this isn't one of those cases, gonna check into why, but iirc globvsglob might be one of the few ambiguity errors that is checked outside of `resolve_ident_in_scope_set`

edit: yeah this is handled by `try_define_local` in rustc_resolve/src/imports.rs, entirely different piece of logic that I haven't looked at in a while

I feel like this should still be possible to trigger but I can't figure out how and afaict this claim came from my read of the logic in `try_define_local` not from some ui test, comment, or similar source

src/items/use-declarations.md
```diff
@@ -398,10 +398,18 @@ r[items.use.ambiguities]
 r[items.use.ambiguities.intro]
 Some situations are an error when there is an ambiguity as to which name a `use` declaration refers. This happens when there are two name candidates that do not resolve to the same entity.
 
+* except where shadowing is allowed
+r[names.resolution.early.imports.errors.ambiguity.globvsglob]
+* it is an error to name an item through ambiguous use declarations
```
yaahc: documented already

src/items/use-declarations.md
```diff
@@ -398,10 +398,18 @@ r[items.use.ambiguities]
...
+    * two globs imports which both have an item matching that name where the items are different
+        * this is still an error even if there is a third non glob binding resolution to an item with the same name
+* it is not an error to have two glob imports which include items which would be ambiguous so long as you do not name one of those items through the ambiguous glob imports
```
yaahc: documented

src/names/namespaces.md
```diff
@@ -119,6 +119,7 @@ For example, the [`cfg` attribute] and the [`cfg` macro] are two different entit
 
 r[names.namespaces.sub-namespaces.use-shadow]
 It is still an error for a [`use` import] to shadow another macro, regardless of their sub-namespaces.
```
yaahc: I feel conflicted about this piece of info being in this section. On the one hand sub-namespaces are quite similar to shadowing, though they work via an independent mechanism. On the other, I'm fairly certain this is exclusively referring to the `TextualVsPathBased` ambiguity error with the added bit of information of pointing out that this is unaffected by sub-namespaces. I need to verify this but I suspect this is because of how imports bring in names from all namespaces so it probably always counts as a sub-namespace match preventing the subnamespace mismatch early exit from ever bailing before we report the ambiguity error.

Assuming this explanation is correct, we can maybe find a place within the use-declarations or mbe chapters to mention how imports are unaffected by subnamespaces. I'm currently wondering if it would make sense to move the subnamespace stuff to live in one of the macros chapters and link to it from here.

src/names/name-resolution.md
petrochenkov: It is needed for 2015 edition too.
For proc macros, as you said, but also for `#[macro_export] macro_rules!` which puts an additional modularized "path-based" declaration into the crate root module (which can be used both internally and from other crates).

src/items/use-declarations.md
petrochenkov: >this is still an error even if there is a third non glob binding resolution to an item with the same name

This doesn't seem right.
If there's an explicit item (import or not) with the same name, then it will shadow the ambiguous glob and will be used as a resolution instead.

petrochenkov: >I expected this to be dependent upon ordering.

Resolution results are supposed to never depend on the item ordering (besides `macro_rules` with their "textual scope"), otherwise we won't be able to have things like [parallel import resolution](https://github.com/rust-lang/rust/pull/145108) or import reordering in rustfmt.
In practice there may be dependencies in corner cases, but every such case is a bug.

petrochenkov: >Resolution results are supposed to never depend on the item ordering

I guess this is worth putting into the reference as an axiom / expressed intent.

src/items/use-declarations.md
```diff
@@ -442,6 +449,111 @@ fn main() {
 }
 ```
 
+r[names.resolution.early.imports.errors.ambiguity.builtin-attr]
+* it is an error to have a user defined attribute or derive macro with the same name as a builtin attribute (e.g. inline)
```
petrochenkov: This doesn't look right, it's ok to define something named `inline`, but it's not ok to actually use it.
```rust
use derive as inline; // OK

#[inline] // Not OK, ambiguity at use time
fn main() {}
```

src/items/use-declarations.md
```diff
@@ -442,6 +449,111 @@ fn main() {
...
+    * I think we may special case this one and allow certain kinds of
+      ambiguities where the builtin-attr is shadowed by a user attribute (not
+      sure if this actually exists or is just proposed, TODO investigate)
+    * TODO example? This ones harder to do concisely afaik
+r[names.resolution.early.imports.errors.ambiguity.derivehelper]
+* derive helpers used before their associated derive may not shadow other attributes or other derive helpers that are otherwise in scope after their derive
```
petrochenkov: This is something that exists under a deprecation lint in the final stage of its lifecycle (https://github.com/rust-lang/rust/issues/79202), and hopefully will be removed next year.
Not sure if things like this are supposed to be documented in the reference.

src/items/use-declarations.md
```diff
@@ -442,6 +449,111 @@ fn main() {
...
+* path-based scope bindings for macros may not shadow textual scope bindings to macros
+    * This is sort of an intersection between macros and imports, because at
+      least in stable rust you can only get path-based macro resolutions from
+      imports of mbe macros (and presumably from proc macro crates), but you
+      can only get textual scope of macros from macro declarations
+    * https://doc.rust-lang.org/nightly/reference/names/namespaces.html#r-names.namespaces.sub-namespaces.use-shadow
+    * [macro.decl.scope.path.ambiguity]
+r[names.resolution.early.imports.errors.ambiguity.globvsouter]
+* it is an error to shadow an outer name binding with a glob import
+    * This seems to only apply to early resolution (duh, I documented this as part of an early resolution codepath)
```
petrochenkov: Most of the new text here need to be moved to name resolution, IMO.

I'd use src/items/use-declarations.md for things that happen at import definition time.
But most of ambiguity errors here do not happen at import definition time, but rather when someone tries to *resolve* a name from some unlucky point of view.

Only some situations create inherently ambiguous import definitions, like glob-vs-glob conflicts, and even then they are not reported at definition time.

src/names/name-resolution.md
```diff
@@ -1,4 +1,106 @@
...
+declarations of those entities. Names are segregated into different
+[namespaces], allowing entities in different namespaces to share the same name
+without conflict. Each name is valid within a [scope], or a region of source
+text where that name may be referenced. Access to certain names may be
+restricted based on their [visibility].
+
+* Names are resolved at three different stages of compilation.
+* [Macros] and [use declarations] are resolved during macro expansion.
+    * This stage of resolution is known as "Early Resolution".
+* Associated consts and functions, methods, and enum variants are resolved during type checking.
```
petrochenkov: `Trait::assoc_item`, `<Type as Trait>::assoc_item` and `Enum::Variant` are resolved during late resolution, or even during early resolution if used in imports (traits and enums are "modules" in name resolution sense).

`Type::assoc_item`, `<Type>::assoc_item`, `<Enum>::Variant` and `EnumTyAlias::Variant` on the other hand are resolved during type-relative resolution.

src/names/name-resolution.md
```diff
@@ -1,4 +1,106 @@
...
+r[names.resolution.early.imports]
+
+* All imports are fully resolved at this point.
+    * imports of names that cannot be fully resolved during macro expansion, such as those depending on type information, are not supported and will produce an error.
+
+TODO example of unsupported type dependent import
+
+r[names.resolution.early.imports.shadowing]
+
+The following is a list of situations where shadowing of use declarations is permitted:
```
petrochenkov:
```suggestion
The following is a list of situations where shadowing of use declarations is not permitted:
```
?

src/names/name-resolution.md
```diff
@@ -1,4 +1,106 @@
...
+* Glob vs Glob
+* ~~Glob vs Expanded~~ pretty certain we don't want to mention this one
+* More Expanded vs Outer
+
+r[names.resolution.early.macros]
+
+* .visitation-order
+    * derive helpers
+        * not visited when resolving derive macros in the parent scope (starting scope)
+    * derive helpers compat
```
petrochenkov: Same concern as above about this being deprecated and removed next year.

src/items/use-declarations.md
traviscross: If the current behavior in `rustc` is considered to be a bug, then we document the intended behavior in the rules and add an admonition that describes what `rustc` does instead (including that there's a lint).  In the admonition, we link to the relevant issue.

src/items/use-declarations.md
yaahc: > Resolution results are supposed to never depend on the item ordering (besides `macro_rules` with their "textual scope"), otherwise we won't be able to have things like [parallel import resolution](https://github.com/rust-lang/rust/pull/145108) or import reordering in rustfmt.

you're of course right, I went back and revisited the example that I was thinking of when I made this comment, it has to do with how `resolve_ident_in_scope_set` only tracks the innermost resolution when tracking ambiguity errors, so in some cases you can add code and remove ambiguity despite still having the code that is otherwise illegal. the first m decl and the m import never get compared for ambiguity to each other and neither is ambiguous with the second m decl (innermost) so no error is emitted.

```rust
macro_rules! m {
    () => {}
}

fn foo() {
    use bar::m2 as m; // remove `m` decl below and this becomes an ambiguity error
    macro_rules! m {
        () => {}
    }
    m!();
}

mod bar {
    macro_rules! m2 {
        () => {}
    }
    
    pub(crate) use m2;
}
```

This isn't necessarily a bug or anything but its where I got the ordering dependent thought. The order in which these are visited does impact whether errors get emitted, but that's the idea, and changing the order involves changing the code semantically, not reordering things lexically.

yaahc: Alright so I dug more into the bit of code which I think I have an incomplete understanding of that the initial comment was based off of. Here's the codepath I am trying to trigger and describe

- upon calling try_define_local
	- the new resolution is a glob import
	- there is already a resolution
	- that resolution contains a shadowed glob res and a shadowing non glob res
		- because the best binding is a non-glob and the new binding is a glob, we go into the second codepath
	- the bindings through the shadowed glob and the new glob are not equal
		- because there was a shadowed glob and it didn't match the res of the new glob binding, we produce a globvsglob ambiguity error

I think my problem is better understanding how try_define_local fits into the rest of import resolution, and when it will get called and when the different members get filled, gonna keep investigating

Edit: best I can tell it doesn't matter if that codepath gets executed. I only vaguely remember this but I recall us talking about how these ambiguity errors are attached to resolutions and only get emitted when we actually use an import. The fact that there exists a shadowed glob import is irrelevant because it is shadowed, so the non glob import will always be selected when resolving the associated name.

yaahc: The thing I don't understand is, when will the ambiguity errors from this codepath ever see the light of day?

As far as I can tell there are two scenarios where it gets triggered

* old binding is a glob import
    * the new binding cannot be a glob import
    * the old bindings glob import would have to not equal the glob import between the old and new best candidates, but those are both referring to old binding so they're always the same, impossible to go further in and attach the ambiguity error
* old binding is not a glob import
    * This is the case described above, the old binding has to have a shadowed glob which ends up being where any ambiguity error is attached, seems like it will always then be ignored.

src/names/name-resolution.md
yaahc: I think the original text is correct.

Use glob shadowing: https://doc.rust-lang.org/nightly/reference/items/use-declarations.html#r-items.use.glob.shadowing
macro textual scope shadowing: https://doc.rust-lang.org/nightly/reference/macros-by-example.html#r-macro.decl.scope.textual.shadow

In my mind the "not permitted" set is the list of ambiguity errors below

src/items/use-declarations.md
yaahc: I went ahead and added a bomb field to that associated ambiguity error to make it panic if it ever observed an error from that codepath and found the culprits

    [ui] tests/ui/resolve/issue-105069.rs
    [ui] tests/ui/resolve/issue-109153.rs

traviscross:
@ehuss and I are looking at this together on the lang-docs office hours call, and we just wanted to express our appreciation to @petrochenkov for having been so responsive with @yaahc on working out the details here.  This is a chapter that we've long wanted to exist, and we're thrilled and appreciative that @yaahc is digging in to shape this up.

src/items/use-declarations.md
yaahc: ive gone ahead and moved this to be in the proc macro helper attribute section and setup a first pass attempt at documenting the intended behavior as I understand it, the admonition about deprecation, and an example.

src/items/use-declarations.md
yaahc: I don't mind moving this but the reason I put them alongside their items was because that seemed most consistent with how we documented other parts of name resolution, specifically how the names that are introduced by each item are documented next to their items and then theres an aggregated list linking to all those sections in the namespaces chapter.

My plan was to put ambiguity and shadowing sections alongside their items and aggregate links to all those sections in name-resolution.md.

You can see where I discussed this plan with eric starting here: [#t-lang-docs/reference > name resolution @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/520709-t-lang-docs.2Freference/topic/name.20resolution/near/544623109)

src/names/name-resolution.md
yaahc: ACK

I went ahead and vaguely updated the associated section but I bet its still wrong. Don't worry about it too much until after I've had time to go thru and properly explore type-relative resolution and understand the difference between these cases.

src/names/name-resolution.md
yaahc: ACK, updated the text a bit, I still have a bit about how you can only re-export macros in 2018 edition which I'm pretty sure is true (edit: yup, tested it, works as expected only in 2018 and later). I also went and added a section about re-exporting to the path-based scope section in the mbe chapter.

src/names/name-resolution.md
yaahc: currently going with `type-relative resolution` for the last stage since that's the term vadim uses.

src/names/name-resolution.md
ehuss: Let's just delete the duplicates: https://github.com/rust-lang/reference/pull/2060

src/names/name-resolution.md
yaahc: So thinking about other things we could name the various stages of name resolution in the reference to more clearly indicate their role and relative timing.

For early resolution I'm leaning towards expansion-time resolution. I want the name to capture the core idea that this stage of name res is the minimum required for macro expansion, then document what that entails (macros and imports).

I think for late resolution I lean towards calling it "primary resolution" or something along those lines. I want to capture the fact that it is where most name resolution happens, and is defined as the empty space between the other forms of name resolution. I was briefly worried that users may be confused about when primary resolution happens relative to the other two stages of name resolution, but I think this can be easily navigated in the description of this stage. As long as we clarify that primary resolution is all names that haven't been resolved yet that don't depend on type information it becomes immediately clear that stage is not the last. This can be further reinforced by ensuring that the stages of name resolution are introduced in order of when they occur.

I'm happy with type-relative resolution as it is, I think it already fits the naming scheme I'm trying to move the other stages towards.

traviscross: If we decide to deviate from the names used in the compiler and used widely in discussions over the years, we'll want to add an admonition noting the other name.

src/items/use-declarations.md
yaahc: I've gone ahead and moved the ambiguity sections to the name-res chapter.

src/macros-by-example.md
yaahc, line 1: TODO: come back and add a section mentioning the macro-related ambiguity errors and linking to the name-res chapter

src/names/name-resolution.md
```diff
@@ -1,4 +1,423 @@
...
+pub fn bar() {}
+
+#[myinline]
+pub fn baz() {}
+
+#[inline] // ERROR `inline` is ambiguous
+pub fn qux() {}
+```
+
+r[names.resolution.early.imports.errors.ambiguity.derivehelper]
```
yaahc: TODO: go back and add a section mentioning the proc macro-related ambiguity errors to the proc macros chapter and link to the name-res chapter

src/macros-by-example.md
```diff
@@ -326,6 +326,28 @@ fn foo() {
 // m!(); // Error: m is not in scope.
 ```
 
+* textual scope name bindings for macros may shadow path-based scope bindings
```
yaahc: should this get a rule?

traviscross: Seems so, yes.

src/names/name-resolution.md
```diff
@@ -1,4 +1,425 @@
...
+[use declarations]: ../items/use-declarations.md
+[macro textual scope shadowing]: ../macros-by-example.md#r-macro.decl.scope.textual.shadow
+[`let` bindings]: ../statements.md#let-statements
+[item definitions]: ../items.md
+[namespaces]: ../names/namespaces.md
+[scope]: ../names/scopes.md
+[visibility]: ../visibility-and-privacy.md
+[permitted]: name-resolution.md#r-names.resolution.expansion.imports.shadowing
+[macro invocations]: ../macros.html#macro-invocation
+[path-based scope]: ../macros-by-example.html#r-macro.decl.scope.path-based
```
yaahc: This currently links to the macros-by-example chapter, but in actuality, this applies to all the various forms of proc macros as well. I'd like a better top level section to link to that covers both kinds of macros

src/names/name-resolution.md
```diff
@@ -1,4 +1,416 @@
...
+complete macro expansion and fully generate a crate's AST. This stage requires
+the resolution of macro invocations and use declarations. Resolving use
+declarations is required to resolve [path-based scope] macro invocations.
+Resolving macro invocations is required in order to expand them.
+
+The expansion process is iterative, alternately resolving imports, resolving
+and expanding macro invocations, then repeating until there are no further
+macros invocations to resolve. Once this process is completed all the imports
+are resolved again to ensure that the macro expansion process did not introduce
+any new ambiguious imports.
```
yaahc: do we want to talk about this? feels like an implementation detail but
also really helps to understand certain kinds of ambiguity errors that users
can run into.

src/names/name-resolution.md
```diff
@@ -1,4 +1,416 @@
...
+}
+pub fn foo() {
+    m!(); // ERROR `m` is ambiguous
+    use crate::m2 as m; // in scope for entire function body
+}
+```
+
+r[names.resolution.early.imports.errors.ambiguity.builtin-attr]
+It is an error to use a user defined attribute or derive macro with the same
+name as a builtin attribute (e.g. inline)
```
yaahc: I think we may special case this one and allow certain kinds of ambiguities where the builtin-attr is shadowed by a user attribute (not sure if this actually exists or is just proposed, TODO investigate)

src/names/name-resolution.md
```diff
@@ -1,4 +1,416 @@
...
+
+#[inline] // ERROR `inline` is ambiguous
+pub fn qux() {}
+```
+
+r[names.resolution.early.imports.errors.ambiguity.derivehelper]
+* derive helpers used before their associated derive may not shadow other attributes or other derive helpers that are otherwise in scope after their derive
+    * TODO example? This ones harder to do concisely afaik
+
+Helper attributes may not be used before the macro that introduces them.
```
yaahc: * What happens if two macros introduce the same helper, will the second one not
  be able to see the attribute of the first anymore, assuming their order is
  "firstmacro" "helper" "secondmacro"?

src/procedural-macros.md
```diff
@@ -440,3 +444,4 @@ their equivalent `#[doc = r"str"]` attributes when passed to macros.
 [type expressions]: types.md#type-expressions
 [type]: types.md
 [union]: items/unions.md
+[name resolution ambiguities]: ../names/name-resolution.html#r-names.resolution.expansion.imports.errors.ambiguity
```
yaahc: more precise link

src/names/name-resolution.md
ehuss: If you want to consider an alternate approach, it may be possible to not include this paragraph and don't focus on the "iterative" nature. The chapter could instead focus on what must be true or what the end result is. So it could have separate rules that say things like:

- After expansion-time name resolution, the AST must not contain any unexpanded macro invocations.
- Every (...) resolves to a valid definition that exists in the final AST (or an external crate).
- The resolution of imports must be *stable*. After expansion, imports in the fully expanded AST must resolve to the same definition...
- ?Maybe say something about order-dependency?
- etc.

src/names/name-resolution.md
```diff
@@ -1,4 +1,387 @@
...
+        * always visited
+    * macro rules bindings (textual scope macros)
+        * always visited
+    * modules (path-based scope macros)
+        * always visited
+    * macrouseprelude
+        * not visited in 2018 and later when `#[no_implicit_prelude]` is present
+    * stdlibprelude
+        * always visited for macro resolutions
+        * is it? what about no-std + no-core?
```
chorman0773: no_core isn't a user visible attribute, it's an implementation detail of `core`, and perma-unstable as far as I'm aware.

src/names/namespaces.md
```diff
@@ -119,6 +119,7 @@ For example, the [`cfg` attribute] and the [`cfg` macro] are two different entit
...
+* TODO revisit
```
yaahc: Revisiting this. I think I've decided against wanting to move the entire sub-namespaces section into the name resolution chapter. I was consider if it made sense to argue that they aren't a true namespace since the mechanism through which they're implemented is different and that they're actually a property of the name resolution algorithm that creates "apparent sub-namespaces" or something along those line. I ended up convincing myself that this is a distinction without a difference and its just simpler pedagogically to call them sub-namespaces.

I do think I want to give the use-shadowing subsection the same treatment as the other ambiguity errors, remove the section here and add an admonition linking to the relevant section in name-resolution.

src/names/name-resolution.md
```diff
@@ -1,4 +1,353 @@
...
+}
+
+fn foo() {
+    define_name!();
+    name!(); // ERROR `name` is ambiguous
+}
+```
+
+r[names.resolution.early.imports.ambiguity.pathvstextualmacro]
+Path-based scope bindings for macros may not shadow textual scope bindings to macros. For bindings from [use declarations], this applies regardless of their [sub-namespace].
```
yaahc: implication here that I need to investigate: that it is possible to have path-based bindings to macros that would shadow textual bindings if imported but not if otherwise defined in the same scope.

src/names/name-resolution.md
```diff
@@ -1,4 +1,353 @@
...
+> ```
+>
+> For more details, see [Rust issue #79202](https://github.com/rust-lang/rust/issues/79202).
+
+r[names.resolution.expansion.macros]
+
+Macros are resolved by iterating through the available scopes until a candidate
+is found. Macros are split into two sub-namespaces, one for bang macros, and
+the other for attributes and derives. Resolution candidates from the incorrect
+sub-namespace are ignored. The available scopes are visited in the following order.
```
yaahc: I'm not at all happy with this, in particular the first sentence seems like a gross oversimplification. I wrote this as is because I was trying to find some way to introduce the scope visitation order logic.

src/names/name-resolution.md
```diff
@@ -1,4 +1,351 @@
+r[names.resolution]
 # Name resolution
 
+r[names.resolution.intro]
+
+_Name resolution_ is the process of tying paths and other identifiers to the declarations of those entities. Names are segregated into different [namespaces], allowing entities in different namespaces to share the same name without conflict. Each name is valid within a [scope], or a region of source text where that name may be referenced. Access to certain names may be restricted based on their [visibility].
```
petrochenkov: What other identifiers?
Single-segment paths are also paths.

src/names/name-resolution.md
```diff
@@ -1,4 +1,351 @@
...
+The following is a list of situations where shadowing of use declarations is permitted:
+
+* [use glob shadowing]
+* [macro textual scope shadowing]
+
+r[names.resolution.expansion.imports.ambiguity]
+## Ambiguities
+
+r[names.resolution.expansion.imports.ambiguity.intro]
+Some situations are an error when there is an ambiguity as to which name a `use` declaration refers. This happens when there are two name candidates that do not resolve to the same entity where neither import is [permitted] to shadow the other.
```
petrochenkov:
```suggestion
Some situations are an error when there is an ambiguity as to which name a `use` declaration refers. This happens when there are two name candidates that do not resolve to the same entity where neither candidate is [permitted] to shadow the other.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,351 @@
...
+
+r[names.resolution.expansion.imports.ambiguity.globvsglob]
+Names may not be resolved through ambiguous glob statements. Glob imports are allowed to import conflicting names in the same namespace as long as the name is not used. Conflicting names from ambiguous glob statements may still be shadowed and used without producing an error. The errors occur at time of use, not time of import.
```
petrochenkov:
```
Conflicting names from ambiguous glob statements may still be ... used
```
Weird wording, they both can and cannot be used.
I guess the intent was to say that the ambiguous globs can be shadowed, and the thing that shadows then can then be used?

src/names/name-resolution.md
petrochenkov:
```suggestion
Some situations are an error when there is an ambiguity as to which name a `use` declaration or a macro path refers. This happens when there are two name candidates that do not resolve to the same entity where neither import is [permitted] to shadow the other.
```
I think all the cases below are relevant for both.

src/names/name-resolution.md
petrochenkov: >not sure if this actually exists or is just proposed

It doesn't exist now

src/names/name-resolution.md
petrochenkov: >will the second one not
be able to see the attribute of the first anymore

It will, helpers are not removed.
Helpers don't have "identity" really, technically the compiler tracks it, but for the compiler it's important that it's *someone's* helper and that's enough, and proc macros detect helpers by string comparison, so they'll happily use someone else's helpers with the same name as their own.

petrochenkov: Would be nice to add a test for this to rust-lang/rust, if there's none currently.

src/names/name-resolution.md
yaahc: yeah, the name can still be used even if the name is ambiguious from glob imports so long as the name is actually resolved through some third unambiguous non glob that shadows both.

I can change the wording to not immediately call the name conflicting, and talk about how its the globs with candidates for that name that are conflicting but if they're not actually selected as the final candidate it doesn't matter.

src/names/name-resolution.md
yaahc: lifetimes / type parameters come to mind as an example which I wouldn't consider paths that need their names resolved.

src/names/name-resolution.md
yaahc: you are correct. I went ahead and added the extra cases and improved the examples for globvsouter to better demonstrate the variety of cases. I also reworded it to clarify that the ambiguities come from the definitions and are found when resolving names for macro invocations or use declarations. This let me also capture that modules can be a source of ambiguity.

src/names/name-resolution.md
yaahc: I was setting up a test for this and I think I found a mistake I made in interpreting the logic in `resolve_ident_in_scope_set`. Specifically I said 

```rust
                                } else if innermost_res == derive_helper_compat
                                    || res == derive_helper_compat && innermost_res != derive_helper
                                {
                                    Some(AmbiguityKind::DeriveHelper)
```

means

- if a derive helper which is used before it's derive has the same name as an outer item
- or if a derive helper has the same name as an outer derive helper which is used before it's derive

The first bullet is true afaict, being in this codepath implies theres a res and innermost_res, so having the innermost be a derive_helper_compat means that we have a second candidate we found later in the same sub-namespace.

The second bullet is incorrect. It seems like I interpreted the `!=` as an `==`. This is just saying that the second resolution is a compat resolution, and the first resolution is not a helper, which means it could only be another helper compat due to visitation order. Given the lack of identity on helpers, I'm not sure how to trigger this second half of the condition.

Here's the test I setup:

```rust
#[empty_helper]
#[derive(Empty)]
#[empty_helper]
#[derive(Empty2)]
struct S;
```

(edited next two paragraphs)

Both derives introduce the same `#[empty_helper]` attribute. 

Assuming the language I drafted in this PR which this comment is attached to is correct, the first `#[empty_helper]` should produce a warning and the second one should produce an error, since `Empty2`'s helper attribute is shadowing the correctly in scope attribute from `Empty`. Instead, the only error this produces is the future incompat warning on the first helper for using a helper before it's derive which is converted into an error by `deny(future_incompat)`.

I'm going to try digging into this a bit deeper to see if I can pinpoint exactly when this subbranch gets triggered or whether I'm correct in thinking this is inert.

https://github.com/rust-lang/rust/pull/79078/files#diff-c75db064f0f7310431b21d0ecb61c9438c6e48f9ab2810a25251f504d21c6d9dR865 introduced this logic

Edit: looking at `visit_scopes` it looks like we're never "iterating through derive_helper_compat" scopes so I think it will only ever call the closure from `resolve_ident_in_scope_set` once for the derivehelpercompat scope and only ever produce one derive_helper_compat candidate, so the only candidates that could be visited earlier are the proper derivehelper candidates.

I went ahead and added a panic in the second case and I'm checking if that causes our testsuite to fall over anywhere but initial indications (running it in `tests/ui/proc-macro`) are coming up clean, implying it is in fact inert.

src/names/name-resolution.md
petrochenkov: Lifetimes are not paths, true.
I actually haven't seen this mentioned in the added text, but that's because named lifetimes are resolved in later resolution, which is not documented yet.

Type parameter use is a single-segment path, you don't know that it's a type parameter until you resolve that path.

I recalled one more example though, names of trait impl items are resolved to trait items, and they are not paths.
```rust
trait Trait { fn f(); }
impl Trait for () { fn f() {} } // `f` is resolved to its trait origin (during late resolution)
```

src/names/name-resolution.md
yaahc: added in https://github.com/rust-lang/rust/pull/149200

petrochenkov (approved):

src/names/name-resolution.md
```diff
@@ -1,4 +1,356 @@
...
+}
+
+fn ambiguous_shadow() {
+    // This is permitted, since resolution is not through the ambiguous globs
+    struct Qux;
+    let x = Qux;
+}
+```
+
+Multiple glob imports are allowed to import the same name, and that name is allowed to be used, if the imports are of the same item (following re-exports). The visibility of the name is the maximum visibility of the imports. For example:
```
petrochenkov: >The visibility of the name is the maximum visibility of the imports.

The rules here will need to be restricted to address the so called "`RustEmbed` issue" (https://github.com/rust-lang/rust/pull/147196).
https://github.com/rust-lang/rust/pull/149145 starts working on the solution.

Multiple competing bindings with different visibilities are typically not a problem because visibilities are checked (or at least re-checked) after expansion (during macro/import validation pass in early resolution, or later).
The problem is that visibility *can* be consumed early without later validation, and then propagated uncontrollably, if the ambiguous binding is reexported further.
```rust
// Pseudocode

vis1 use m1::*; // imports `Item`
vis2 use m2::*; // imports same `Item`

vis3 use self::Item;
// How far will `Item` be reexported, to `vis1`, to `vis2` or to `vis3`?
// The answer is unpredictable and depends on the resolution/expansion order.
// This is not an issue if `min(vis1, vis3) == min(vis2, vis3)`.
```

yaahc: Is this something that we consider a bug and plan to fix / remove in the future or is this something we are going to keep forever? (Wondering if I should document this as an admonition or not)

petrochenkov: It's something I personally consider a bug and plan to fix, the breakage shouldn't be large.
The prototype fix is just not in a good enough state yet to present it as a proposal to the lang team.

src/items/use-declarations.md
```diff
@@ -460,3 +412,4 @@ fn main() {
 [tool attributes]: ../attributes.md#tool-attributes
 [type alias]: type-aliases.md
 [type namespace]: ../names/namespaces.md
+[name resolution ambiguities]: ../names/name-resolution.html#r-names.resolution.expansion.imports.ambiguity
```
ehuss: These kinds of links should just be the rule name.


```suggestion
[name resolution ambiguities]: names.resolution.expansion.imports.ambiguity
```

src/macros.md
```diff
@@ -106,6 +106,16 @@ macro_rules! example {
...
 
+r[macro.invocation.name-resolution]
+
+Macros invocations can be resolved via two kinds of scopes:
+
+* Textual Scope
+    * [textual scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.textual)
+* Path-based scope
+    * [path-based scope `macro_rules`](macros-by-example.md#r-macro.decl.scope.path-based)
+    * [Procedural Macros]
```
ehuss: There's a mix of different capitalization here. I think I would expect to start with a capital letter, and lowercase the rest.

src/macros-by-example.md
```diff
@@ -326,6 +326,77 @@ fn foo() {
...
+> For areas where shadowing is not allowed, see [name resolution ambiguities].
+
+r[macro.decl.scope.path-based]
+### Path-based scope
+
+r[macro.decl.scope.path-based.intro]
+By default, a macro has no path-based scope. Macros can gain path-based scope in two ways:
+
+* [Use declaration re-export]
+* [`#[macro_export]`](macros-by-example.html#the-macro_export-attribute)
```
ehuss: For links, there are a few preferences:

- Just link to the rule if it exists. For example, the destination here could be `macro.decl.scope.macro_export`
- If linking to a file, use `.md` extension instead of `.html`
- Link destinations are usually preferred to go as link definitions at the bottom of the file (like `[Use declaration re-export]` was done here.

src/macros-by-example.md
```diff
@@ -724,3 +795,5 @@ For more detail, see the [formal specification].
 [Repetitions]: #repetitions
 [token]: tokens.md
 [`$crate`]: macro.decl.hygiene.crate
+[Use declaration re-export]: items/use-declarations.md#use-visibility
+[name resolution ambiguities]: names/name-resolution.md#r-names.resolution.expansion.imports.ambiguity
```
ehuss: And if possible, we sort these definitions together.

src/macros-by-example.md
```diff
@@ -326,6 +326,77 @@ fn foo() {
 // m!(); // Error: m is not in scope.
 ```
 
+r[macro.decl.scope.textual.shadow.path-based]
+Textual scope name bindings for macros may shadow path-based scope bindings to macros.
```
ehuss: The use of the word "may" here feels like it is weakening the statement, or making it sound like it is optional. I'm not sure if that was intentional or if that word is needed here? That is, could it just be dropped?


```suggestion
Textual scope name bindings for macros shadow path-based scope bindings to macros.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
+r[names.resolution]
 # Name resolution
 
+r[names.resolution.intro]
+_Name resolution_ is the process of tying paths and other identifiers to the declarations of those entities. Names are segregated into different [namespaces], allowing entities in different namespaces to share the same name without conflict. Each name is valid within a [scope], or a region of source text where that name may be referenced. Access to certain names may be restricted based on their [visibility].
+
+Name resolution is split into three stages throughout the compilation process. The first stage, Expansion-time resolution, resolves all [use declarations] and [macro invocations]. The second stage, Primary resolution, resolves all names that have not yet been resolved that do not depend on type information to resolve. The last stage, Type-relative resolution, resolves the remaining names once type information is available.
```
ehuss: For introducing new terms, they are typically placed with emphasis. I would also avoid capitals since I wouldn't consider these proper nouns.


```suggestion
Name resolution is split into three stages throughout the compilation process. The first stage, *expansion-time resolution*, resolves all [use declarations] and [macro invocations]. The second stage, *primary resolution*, resolves all names that have not yet been resolved that do not depend on type information to resolve. The last stage, *type-relative resolution*, resolves the remaining names once type information is available.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
...
+
+> [!NOTE]
+>
+> * Expansion-time resolution is also known as "Early Resolution".
+> * Primary resolution is also known as "Late Resolution".
```
ehuss: Similar comment about not capitalizing.


```suggestion
> * Expansion-time resolution is also known as "early resolution".
> * Primary resolution is also known as "late resolution".
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
...
+
+#[inline] // ERROR: `inline` is ambiguous
+pub fn qux() {}
+```
+
+r[names.resolution.expansion.imports.ambiguity.derivehelper]
+Helper attributes may not be used before the macro that introduces them.
+
+> [!NOTE]
+> rustc currently allows derive helpers to be used before their attribute macro introduces them into scope so long as they do not shadow any other attributes or derive helpers that are otherwise correctly in scope. This behavior deprecated and slated for removal.
```
tshepang:
```suggestion
> rustc currently allows derive helpers to be used before their attribute macro introduces them into scope so long as they do not shadow any other attributes or derive helpers that are otherwise correctly in scope. This behavior is deprecated and slated for removal.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
...
+> * Expansion-time resolution is also known as "Early Resolution".
+> * Primary resolution is also known as "Late Resolution".
+
+r[names.resolution.expansion]
+## Expansion-time name resolution
+
+r[names.resolution.expansion.intro]
+Expansion-time name resolution is the stage of name resolution necessary to complete macro expansion and fully generate a crate's AST. This stage requires the resolution of macro invocations and use declarations. Resolving use declarations is required to resolve [path-based scope] macro invocations. Resolving macro invocations is required in order to expand them.
+
+After expansion-time name resolution, the AST must not contain any unexpanded macro invocations. Every macro invocation resolves to a valid definition that exists in the final AST (or an external crate). The resolution of imports must be *stable*. After expansion, imports in the fully expanded AST must resolve to the same definition, regardless of the order in which macros are expanded. Once the crate has been fully expanded all speculative import resolutions are validated to ensure that no new ambiguities were introduced by macro expansion.
```
ehuss: I would probably just drop the parentheses here.


```suggestion
After expansion-time name resolution, the AST must not contain any unexpanded macro invocations. Every macro invocation resolves to a valid definition that exists in the final AST or an external crate. The resolution of imports must be *stable*. After expansion, imports in the fully expanded AST must resolve to the same definition, regardless of the order in which macros are expanded. Once the crate has been fully expanded all speculative import resolutions are validated to ensure that no new ambiguities were introduced by macro expansion.
```

What is a "speculative" import resolution? 

I would also consider breaking this paragraph into separate rules. For example.

Rule 1: The first two sentences about unexpanded macro invocations/unresolved. An example of this could possibly be simply `foo!()`, which is an unresolved macro.
Rule 2: Next two sentences about stability. 
Rule 3: Validation of ambiguities.

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
...
+>     //   imports and macro invocations, sees the introduced ambiguity
+>     //   and reports it as an error.
+>     bar::m!(); // ERROR: `bar` is ambiguous
+> }
+> ```
+
+r[names.resolution.expansion.imports]
+### Imports
+
+
```
ehuss, line 92: Try to avoid multiple blank lines in a row.


```suggestion

```

src/names/name-resolution.md
```diff
@@ -1,4 +1,348 @@
...
+    let _ = my_mod::TypeAlias::MyVariant; // OK
+    let _ = my_mod::MyEnum::Const; // OK
+}
+```
+
+r[names.resolution.expansion.imports.shadowing]
+The following is a list of situations where shadowing of use declarations is permitted:
+
+* [use glob shadowing]
+* [macro textual scope shadowing]
```
ehuss: Lists should usually start with a capital letter. Also, I realize we are not very consistent with this, but I think in this case the `use` would be inline code to differentiate it from the English word "use".

```suggestion
* [`use` glob shadowing]
* [Macro textual scope shadowing]
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,342 @@
+r[names.resolution]
 # Name resolution
 
+r[names.resolution.intro]
+_Name resolution_ is the process of tying paths and other identifiers to the declarations of those entities. Names are segregated into different [namespaces], allowing entities in different namespaces to share the same name without conflict. Each name is valid within a [scope], or a region of source text where that name may be referenced. Access to certain names may be restricted based on their [visibility].
+
+Name resolution is split into three stages throughout the compilation process. The first stage, *expansion-time resolution*, resolves all [use declarations] and [macro invocations]. The second stage, *primary resolution*, resolves all names that have not yet been resolved that do not depend on type information to resolve. The last stage, *type-relative resolution*, resolves the remaining names once type information is available.
```
ehuss: Style-wise, I would say here and throughout the chapter that "use declaration" would have backticks around the `use`.

ehuss:
Something I'm missing is a discussion about blocks, and how they can introduce essentially an anonymous module (`ModuleKind::Block`). That's important when trying to understand the module tree and how some resolutions and ambiguities resolve differently. I was looking at this example:

```rust
mod bar {
    pub struct Name;
}

mod baz {
    pub struct Name;
}

use baz::Name;

pub fn foo() {
    use bar::*;
    Name; // resolves to bar::Name
}
```

I think the reason it resolves to `bar::Name` is because it is nested inside an anonymous module. If we were to rewrite this to:

```rust
mod bar {
    pub struct Name;
}

mod baz {
    pub struct Name;
}

use baz::Name;
use bar::*;

pub fn foo() {
    Name; // resolves to baz::Name
}
```

If we were to ignore blocks, and pretend they don't exist from a module-hierarchy perspective, I would expect both these examples to be the same. I'm not sure if I'm understanding this correctly, since resolution within a block still "sees" everything outside of the block (inside it's module).

src/names/name-resolution.md
```diff
@@ -1,4 +1,342 @@
...
+>
+> Starting in edition 2018 the `#[macro_use]` prelude is not visited when `#[no_implicit_prelude]` is present.
+
+TODO linkify
+
+r[names.resolution.expansion.macros.derivehelpers]
+not visited when resolving derive macros in the parent scope (starting scope)
+
+r[names.resolution.expansion.macros.reserved-names]
+The names cfg and cfg_attr are reserved in the macro attribute [sub-namespace].
```
ehuss:
```suggestion
The names `cfg` and `cfg_attr` are reserved in the macro attribute [sub-namespace].
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,359 @@
...
+
+r[names.resolution.primary]
```
ehuss, line 526: I would go ahead and put headers here to separate it out from the previous section.

```suggestion
r[names.resolution.primary]
## Primary name resolution
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,359 @@
...
+> [!NOTE]
+> This is a placeholder for future expansion about primary name resolution.
+
+r[names.resolution.type-dependent]
```
ehuss:
```suggestion
r[names.resolution.type-dependent]
# Type-dependent resolution
```

traviscross:
Thanks to @yaahc for joining us on the lang-docs office hours today to speedrun many revisions to this chapter, which is really shaping up.

src/procedural-macros.md
```diff
@@ -240,6 +240,26 @@ A helper attribute for a derive macro is declared by adding its identifier to th
 > }
 > ```
 
+r[macro.proc.derive.attributes.scope]
+Helper attributes are in scope after the macro that introduces them.
```
ehuss: Would it be possible to be more specific about what "after" means here? There's some start point and some end point, correct?

That is, there are different interpretations of what "after" means:
- Anything after that point to the end of the file.
- Only *inside* the attached item.
- Inside the attached item *and* inside the output of the derive.
- Inside the attached item *and* any outer attribute *after* the `derive`.

The last one is my understanding of what it covers. I'm not sure what the best way to phrase that is.

src/procedural-macros.md
```diff
@@ -240,6 +240,26 @@ A helper attribute for a derive macro is declared by adding its identifier to th
...
+
+> [!NOTE]
+> rustc currently allows derive helpers to be used before the macro that
```
ehuss: Unwrap the blockquote.

src/names/name-resolution.md
```diff
@@ -1,4 +1,384 @@
...
+* [derive helpers]
+* [textual scope macros]
+* [path-based scope macros]
+* [`macro_use` prelude]
+* [Standard library prelude]
+* [Builtin attributes]
+
+> [!NOTE]
+>
+> The compiler will attempt to resolve derive helpers that are used before
```
ehuss: Unwrap.

(I realize blockquotes don't look terribly great in source. I'm wondering if we should just choose a different syntax...)

src/names/name-resolution.md
```diff
@@ -1,4 +1,384 @@
...
+> removal.
+>
+> For more info see [derive helper scope].
+
+> [!EDITION-2018]
+>
+> Starting in edition 2018 the `#[macro_use]` prelude is not visited when `#[no_implicit_prelude]` is present.
+
+r[names.resolution.expansion.macros.derivehelpers]
+not visited when resolving derive macros in the parent scope (starting scope)
```
ehuss: Seems like an incomplete rule?

src/names/name-resolution.md
```diff
@@ -1,4 +1,384 @@
...
+* [`use` glob shadowing]
+* [Macro textual scope shadowing]
+
+r[names.resolution.expansion.imports.ambiguity]
+#### Ambiguities
+
+r[names.resolution.expansion.imports.ambiguity.intro]
+Some situations are an error when there is an ambiguity as to which macro definition, `use` declaration, or module an import or macro invocation's name refers to. This happens when there are two name candidates that do not resolve to the same entity where neither candidate is [permitted] to shadow the other.
+
+r[names.resolution.expansion.imports.ambiguity.globvsglob]
```
ehuss: The style is to use `-` between words.


```suggestion
r[names.resolution.expansion.imports.ambiguity.glob-vs-glob]
```

Same for derivehelpers globvsglob globvsouter moreexpandedvsouter pathvstextualmacro.

src/names/name-resolution.md
```diff
@@ -1,4 +1,384 @@
...
+}
+
+pub fn qux() {
+    use bar::*;
+    m!(); // ERROR: `m` is ambiguous
+}
+```
+
+> [!NOTE]
+> These ambiguity errors are specific to imports, even though they are only observed when those imports are used, having multiple candidates available for a given name during later stages of resolution is not considered an error, so long as none of the imports themselves are ambiguous, there will always be a single unambiguous closest resolution during later stages.
```
ehuss: Can this be split up into a few sentences?  (and the "during later stages" seems duplicated.)

Maybe something like:


```suggestion
> These ambiguity errors are specific to imports, even though they are only observed when those imports are used. Having multiple candidates available for a given name during later stages of resolution is not considered an error, so long as none of the imports themselves are ambiguous, there will always be a single unambiguous closest resolution.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,384 @@
...
+fn main() {
+    foo!(); // ERROR: cannot find macro `foo` in this scope
+}
+```
+
+r[names.resolution.expansion.expansion-order-stability]
+The resolution of names must be *stable*. After expansion, names in the fully expanded AST must resolve to the same definition, regardless of the order in which macros are expanded.
+
+r[names.resolution.expansion.speculation]
+All name resolution candidates selected during macro expansion are considering speculative. Once the crate has been fully expanded all speculative import resolutions are validated to ensure that no new ambiguities were introduced by macro expansion.
```
ehuss:
```suggestion
All name resolution candidates selected during macro expansion are considered speculative. Once the crate has been fully expanded, all speculative import resolutions are validated to ensure that macro expansion did not introduce any new ambiguities.
```

src/procedural-macros.md
traviscross:
```suggestion
A derive macro may introduce certain helper attributes. When a derive macro invocation is applied to an item, the helper attributes introduced by that derive macro become in scope 1) for attributes that are applied to that item and are applied lexically after the derive macro invocation and 2) for attributes that are applied to fields and variants inside of the item.
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,381 @@
...
+
+r[names.resolution.expansion.macros.intro]
+Macros are resolved by iterating through the available scopes to find the available candidates. Macros are split into two sub-namespaces, one for bang macros, and the other for attributes and derives. Resolution candidates from the incorrect sub-namespace are ignored.
+
+r[names.resolution.expansion.macros.visitation-order]
+The available scopes are visited in the following order.
+
+* [derive helpers]
+* [textual scope macros]
+* [path-based scope macros]
```
traviscross: These should be sentence-cased.

src/names/name-resolution.md
petrochenkov: New PR with the completed fix - https://github.com/rust-lang/rust/pull/149596.

src/names/name-resolution.md
```diff
@@ -1,4 +1,399 @@
...
+>
+> pub fn foo() {
+>     use bar::*;
+>     assert!(NAME);
+>     //      ^--- This `NAME` is resolved during primary resolution.
+> }
+> ```
+
+r[names.resolution.expansion.imports.ambiguity.path-vs-textual-macro]
+Path-based scope bindings for macros may not shadow textual scope bindings to macros. For bindings from [`use` declarations], this applies regardless of their [sub-namespace].
```
yaahc: I went to update this example because

- I realized that for the purposes of this ambiguity error, only path-based scope bindings from use declarations are relevant. The only other sources of path-based scope bindings are:
  - `#[macro_export]`, which unconditionally adds an entry to the crate root, which by definition is always going to be in the outermost module scope and couldn't otherwise lexically shadow any textual macro bindings in a crate.
  - proc macros, which cannot be used in the crate they're defined, and must be imported via a use declaration in which case it's using an import, or (and I'm not sure this works) a macro_use attr on an extern crate statement for the proc macro crate, in which case the candidate would live in the macro_use prelude and this this ambiguity error would never apply (nor need to)
- I wanted to demonstrate the "regardless of their sub-namespace" claim as well as the core claim.

unfortunately, my attempt compiled without issue:

```rust
macro_rules! m {
    () => {};
}

pub fn foo() {
    m!();

    use myinline::inline as m;
}
```

I think the issue here is that I attempted to combine this lint with [a rule in namespaces.md](https://doc.rust-lang.org/reference/names/namespaces.html#r-names.namespaces.sub-namespaces.use-shadow) based on a misunderstanding of what that rule was talking about.

I believe this rule is meant to clarify the scope of https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.restrictions.duplicate-name

src/names/name-resolution.md
```diff
@@ -1,4 +1,420 @@
...
+
+r[names.resolution.expansion.macros.visitation-order]
+The available scopes are visited in the following order.
+
+* [Derive helpers]
+* [Textual scope macros]
+* [Path-based scope macros]
+* [`macro_use` prelude]
+* [Standard library prelude]
+* [Builtin attributes]
```
yaahc: should this clarify that each of these bullets represent one or more scopes or is this intuitively obvious?

traviscross: Probably worth clarifying.

src/names/name-resolution.md
```diff
@@ -1,4 +1,420 @@
...
+> struct S;
+> ```
+
+r[names.resolution.primary]
+## Primary name resolution
+> [!NOTE]
+> This is a placeholder for future expansion about primary name resolution.
+
+r[names.resolution.type-relative]
+# Type-relative resolution
```
ehuss:
```suggestion
## Type-relative resolution
```

src/names/name-resolution.md
```diff
@@ -1,4 +1,553 @@
...
+#             () => {}
+#         }
+#     }
+# }
+define_ambig!();
+define_ambig!();
+ambig!(); // ERROR: `ambig` is ambiguous.
+```
+
+This also applies to imports, so long as the innermost candidate for the invocation of name is from within a macro expansion.
```
ehuss: We were puzzling over this sentence and were wondering if you could clarify something.

(The phrase "the invocation of name" seemed out of place here, and we are assuming this should be "the name", since there isn't an invocation here.)

This ambiguities section starts off talking about an ambiguity involving macro invocations and expanded macros. But then this sentence seems to say, "oh and the ambiguity applies to imports, too.". Am I understanding it correctly that the rule here applies to macro invocations **and** imports? Is there a reason the start of the rule leaves out the "and imports" part?

traviscross (approved):

traviscross:
Thanks @yaahc.  Great work here, and great working with you on it.  Thanks to @petrochenkov for the critical support.

src/names/name-resolution.md
yaahc: The "invocation" language came up because of [this function](https://github.com/rust-lang/rust/blob/main/compiler/rustc_resolve/src/lib.rs#L1024) and how it refers to the site the name being resolved came from as an invocation regardless of whether it was an import or a macro. I wasn't really sure if this was the more proper term or not, elsewhere in the reference I've used language like "name is used" to the same effect. 

I just reread the beginning of this ambiguity section and at least the way I read and meant it it's not talking about this as only applying to macros. Whether or not the invocation is a macro or a use decl there have to be macro expansions involved in order for this lint to trigger. I did my best to make the language generic in the beginning and talk about candidates without specifically saying imports or macro definitions as the sources of those candidate bindings.

PR #2056, opened 2025-10-22
Document ppc inline asm support

pmur:
This PR is a sub-part of rust-lang/rust#147996.

r? @Amanieu

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -16,6 +16,7 @@ Support for inline assembly is stable on the following architectures:
 - RISC-V
 - LoongArch
 - s390x
+- PowerPC and PowerPC64
 
 The compiler will emit an error if an assembly macro is used on an unsupported target.
 
@@ -627,11 +628,24 @@ Here is the list of currently supported register classes:
 | s390x | `freg` | `f[0-15]` | `f` |
 | s390x | `vreg` | `v[0-31]` | Only clobbers |
 | s390x | `areg` | `a[2-15]` | Only clobbers |
+| PowerPC | `reg` | `r0`, `r[3-12]`, `r[14-28]` | `r` |
+| PowerPC | `reg_nonzero` | `r[3-12]`, `r[14-28]` | `b` |
+| PowerPC | `spe_acc` | `spe_acc` | Only clobbers |
+| PowerPC64 | `reg` | `r0`, `r[3-12]`, `r[14-29]` | `r` |
+| PowerPC64 | `reg_nonzero` | `r[3-12]`, `r[14-29]` | `b` |
+| PowerPC/PowerPC64 | `freg` | `f[0-31]` | `f` |
+| PowerPC/PowerPC64 | `vreg` | `v[0-31]` | `v` |
+| PowerPC/PowerPC64 | `vsreg` | `vs[0-63]` | `wa` |
+| PowerPC/PowerPC64 | `cr` | `cr[0-7]`, `cr` | Only clobbers |
+| PowerPC/PowerPC64 | `ctr` | `ctr` | Only clobbers |
+| PowerPC/PowerPC64 | `lr` | `lr` | Only clobbers |
+| PowerPC/PowerPC64 | `xer` | `xer` | Only clobbers |
 
 > [!NOTE]
 > - On x86 we treat `reg_byte` differently from `reg` because the compiler can allocate `al` and `ah` separately whereas `reg` reserves the whole register.
 > - On x86-64 the high byte registers (e.g. `ah`) are not available in the `reg_byte` register class.
 > - Some register classes are marked as "Only clobbers" which means that registers in these classes cannot be used for inputs or outputs, only clobbers of the form `out(<explicit register>) _` or `lateout(<explicit register>) _`.
+> - The `spe_acc` register is only available on PowerPC SPE targets
 
 r[asm.register-operands.value-type-constraints]
 Each register class has constraints on which value types they can be used with. This is necessary because the way a value is loaded into a register depends on its type. For example, on big-endian systems, loading a `i32x4` and a `i8x16` into a SIMD register may result in different register contents even if the byte-wise memory representation of both values is identical. The availability of supported types for a particular register class may depend on what target features are currently enabled.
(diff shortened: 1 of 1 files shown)


src/inline-assembly.md
```diff
@@ -1346,6 +1379,13 @@ r[asm.rules.preserved-registers]
     - Vector extension state (`vtype`, `vl`, `vxsat`, and `vxrm`).
   - LoongArch
     - Floating-point condition flags in `$fcc[0-7]`.
+  - PowerPC/PowerPC64
+    - Floating-point status and sticky bits in the `fpscr` (any field other than DRN, VE, OE, UE, ZE, XE, NI, or RN).
+    - Vector status and sticky bits in the `vscr` (any field other than NJ).
+    - The `mq` register (this register only exists on POWER1 and POWER2).
+  - PowerPC SPE
+    - The sticky and status bits of the `spefscr` (any field other than FINXE, FINVE, FDBZE, FUNFE, FOVFE, or FRMC)
+    - The Accumulator register `acc` (LLVM names this `spe_acc`).
```
Amanieu: This should probably be defined as a clobber-only register. This should be fine to stabilize along with the other registers here.

src/inline-assembly.md
```diff
@@ -1316,6 +1352,10 @@ r[asm.rules.stack-below-sp]
   - You should adjust the stack pointer when allocating stack memory as required by the target ABI.
   - The stack pointer must be restored to its original value before leaving the assembly code.
 
+r[asm.rules.stack-above-sp]
+- Unless the `nostack` option is set, assembly code is allowed to modify the caller's stack frame in specific cases.
+  - The target ABI requires storing certain values in the caller's frame (e.g saving the `lr` on PowerPC64)
```
traviscross: Is this meaning to say:

> When the `nostack` option is not set, assembly code is allowed to modify the caller's stack frame if and only if the target ABI requires storing values in the caller's frame (e.g., saving the `lr` on PowerPC64).

?

pmur: I intended it to be a sublist. I am not sure there aren't other cases which may need mentioning.

traviscross: Right; I'm asking here about content rather than factoring -- I'm checking with you my reading is correct.

Are there any cases other than when the target ABI requires values to be stored in the caller's frame?

pmur: The parameter save space, if any, of the caller can be used to spill arguments on ppc64 in some cases. That's usually for C variadic functions.

Amanieu: As far as I know writing into the caller's stack frame is something that only happens on PowerPC.

src/inline-assembly.md
```diff
@@ -691,6 +663,17 @@ The availability of supported types for a particular register class may depend o
 | s390x | `freg` | None | `f32`, `f64` |
 | s390x | `vreg` | N/A | Only clobbers |
 | s390x | `areg` | N/A | Only clobbers |
+| PowerPC | `spe_acc` | None | Only clobbers |
+| PowerPC/PowerPC64 | `reg` | None | `i8`, `i16`, `i32`, `i64` (PowerPC64 only) |
+| PowerPC/PowerPC64 | `reg_nonzero` | None | `i8`, `i16`, `i32`, `i64` (PowerPC64 only) |
+| PowerPC/PowerPC64 | `freg` | None | `f32`, `f64` |
+| PowerPC/PowerPC64 | `vreg` | `altivec` | `i8x16`, `i16x8`, `i32x4`, `f32x4` |
+| PowerPC/PowerPC64 | `vreg` | `vsx` | `f32`, `f64`, `i64x2`, `f64x2` |
```
folkertdev, line 693: The source has the following `FIXME` linking to https://github.com/rust-lang/rust/pull/131551#discussion_r1862535963, maybe that can be resolved?

https://github.com/rust-lang/rust/blob/a3f2d5abe45a9acfaccbf09266b33e1fd7ab193e/compiler/rustc_target/src/asm/powerpc.rs#L58-L62

```rust
            // FIXME: vsx also supports integers?: https://github.com/rust-lang/rust/pull/131551#discussion_r1862535963
            Self::vreg => types! {
                altivec: VecI8(16), VecI16(8), VecI32(4), VecF32(4);
                vsx: F32, F64, VecI64(2), VecF64(2);
            },
```

cc @taiki-e

pmur: The integer support for vsx is quite limited. It's only 128b integers (power8) with very restricted functionality prior to power10 (notably, no divide support). Maybe that could be supported via `VecI128(1)`? I suspect the llvm issues remain. Is that a blocker for stabilization?


F128 is another type which should be allowed, but I think that support remains unstable?

folkertdev: > The integer support for vsx is quite limited. It's only 128b integers (power8) with very restricted functionality prior to power10 (notably, no divide support). Maybe that could be supported via VecI128(1)? I suspect the llvm issues remain. Is that a blocker for stabilization?

Not a blocker I think, this just seemed like a good moment to go over the implementation while the specifics are "in ram". If you believe the rust implementation is complete given what powerpc is able to provide we're all good here.

> F128 is another type which should be allowed, but I think that support remains unstable?

LLVM currently does not allow putting `f128` into `vreg` or `vsreg`.

https://github.com/llvm/llvm-project/blob/f59d12001fd877e44e25f260db888a352d5ab755/llvm/lib/Target/PowerPC/PPCISelLowering.cpp#L18201-L18207

Do you think that is an oversight in LLVM?

(the `f128` type is still unstable but we'd like to add support for it in the compiler where possible, see https://github.com/rust-lang/rust/issues/125398)

pmur: > Do you think that is an oversight in LLVM?

I think so. Gcc allows it, and I believe ELFv2 calling conventions require they be passed via vector registers. LLVM also crashed when I tried wrapping it into a vector.

As for I128. It behaves strangely using LLVM. If you don't wrap it in a vector, it gets treated like a register pair. If you wrap it in  a vector and fail to set the minimum cpu to pwr8 or newer (on ppc64), it quietly generates bad code to move between gpr and vr.

folkertdev: maybe our target maintainers can confirm the expected behavior here?

cc @daltenty @gilamn5tr @amy-kwan

I'd be happy to submit a patch to LLVM if we can agree on how this should work.

traviscross (approved):

PR #2058, opened 2025-10-22
Add caveats about mutable references in consts

traviscross:
As it turns out, the story that "the final value of a constant cannot contain any mutable references" has some nuance to it.  Let's describe these caveats.

Thanks to theemathas for noticing these and to RalfJ for filling in context.

cc @RalfJung @oli-obk @theemathas @tshepang @ehuss

### DIFF
--- a/src/items/constant-items.md
+++ b/src/items/constant-items.md
@@ -76,6 +76,24 @@ const _: &mut u8 = unsafe { &mut S }; // ERROR.
 // Not allowed as the mutable reference appears in the final value.
 ```
 
+> [!NOTE]
+> Constant initializers can be thought of, in most cases, as being inlined wherever the constant appears. If a constant whose value contains a mutable reference to a mutable static were to appear twice, and this were to be allowed, that would create two mutable references, each having `'static` lifetime, to the same place. This could produce undefined behavior.
+>
+> Constants that contain mutable references to temporaries whose scopes have been extended to the end of the program have that same problem and an additional one.
+>
+> ```rust,compile_fail,E0764
+> const _: &mut u8 = &mut 0; // ERROR.
+> //                 ^^^^^^
+> // Not allowed as the mutable reference appears in the final value and
+> // because the constant expression contains a mutable borrow of an
+> // expression whose temporary scope would be extended to the end of
+> // the program.
+> ```
+>
+> Here, the value `0` is a temporary whose scope is extended to the end of the program (see [destructors.scope.lifetime-extension.static]). Such temporaries cannot be mutably borrowed in constant expressions (see [const-eval.const-expr.borrows]).
+>
+> To allow this, we'd have to decide whether each use of the constant creates a new `u8` value or whether each use shares the same lifetime-extended temporary. The latter choice, though closer to how `rustc` thinks about this today, would break the conceptual model that, in most cases, the constant initializer can be thought of as being inlined wherever the constant is used. Since we haven't decided, and due to the other problem mentioned, this is not allowed.
+
 ```rust,compile_fail,E0080
 # #![allow(static_mut_refs)]
 static mut S: u8 = 0;
@@ -158,6 +176,8 @@ const _: &&mut u8 = unsafe { &S }; // OK.
 > ```
 >
 > Here, the `AtomicU8` is a temporary whose scope is extended to the end of the program (see [destructors.scope.lifetime-extension.static]). Such temporaries with interior mutability cannot be borrowed in constant expressions (see [const-eval.const-expr.borrows]).
+>
+> To allow this, we'd have to decide whether each use of the constant creates a new `AtomicU8` or whether each use shares the same lifetime-extended temporary. The latter choice, though closer to how `rustc` thinks about this today, would break the conceptual model that, in most cases, the constant initializer can be thought of as being inlined wherever the constant is used. Since we haven't decided, this is not allowed.
 
 r[items.const.expr-omission]
 The constant expression may only be omitted in a [trait definition].
--- a/src/items/constant-items.md
+++ b/src/items/constant-items.md
@@ -49,32 +49,104 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
 ```
 
 r[items.const.no-mut-refs]
-The final value of a `const` item cannot contain any mutable references.
+The final value of a `const` item, after the initializer is evaluated to a value that has the declared type of the constant, cannot contain any mutable references except as described below.
 
 ```rust
 # #![allow(static_mut_refs)]
 static mut S: u8 = 0;
-const C: &u8 = unsafe { &mut S }; // OK
+const _: &u8 = unsafe { &mut S }; // OK.
+//                      ^^^^^^
+// Allowed since this is coerced to `&S`.
 ```
 
 ```rust
 # use core::sync::atomic::AtomicU8;
 static S: AtomicU8 = AtomicU8::new(0);
-const C: &AtomicU8 = &S; // OK
+const _: &AtomicU8 = &S; // OK.
+//                   ^^
+// Allowed even though the shared reference is to an interior
+// mutable value.
 ```
 
 ```rust,compile_fail,E0080
 # #![allow(static_mut_refs)]
 static mut S: u8 = 0;
-const C: &mut u8 = unsafe { &mut S }; // ERROR not allowed
+const _: &mut u8 = unsafe { &mut S }; // ERROR.
+//                          ^^^^^^
+// Not allowed as the mutable reference appears in the final value.
 ```
+
+```rust,compile_fail,E0080
+# #![allow(static_mut_refs)]
+static mut S: u8 = 0;
+const _: &dyn Send = &unsafe { &mut S }; // ERROR.
+//                             ^^^^^^
+// Not allowed as the mutable reference appears in the final value,
+// even though type erasure occurs.
+```
+
+Mutable references where the referent is a value of a [zero-sized type] are allowed.
+
+```rust
+# #![allow(static_mut_refs)]
+static mut S: () = ();
+const _: &mut () = unsafe { &mut S }; // OK.
+//            ^^ This is a zero-sized type.
+```
+
+```rust
+# #![allow(static_mut_refs)]
+static mut S: [u8; 0] = [0; 0];
+const _: &mut [u8; 0] = unsafe { &mut S }; // OK.
+//            ^^^^^^^ This is a zero-sized type.
+```
+
+> [!NOTE]
+> This is allowed as, for a value of a zero-sized type, no bytes can actually be mutated. We must accept this as `&mut []` is [promoted].
+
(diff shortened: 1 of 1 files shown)


src/items/constant-items.md
```diff
@@ -49,32 +49,104 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
...
+
+```rust
+# #![allow(static_mut_refs)]
+static mut S: [u8; 0] = [0; 0];
+const _: &mut [u8; 0] = unsafe { &mut S }; // OK.
+//            ^^^^^^^ This is a zero-sized type.
+```
+
+> [!NOTE]
+> This is allowed as, for a value of a zero-sized type, no bytes can actually be mutated. We must accept this as `&mut []` is [promoted].
```
theemathas, line 123: I believe that this is the first time in the reference that we document that mutable references to zero-length arrays can be promoted.

Note that this only works with zero-length arrays. Also, this promotion does not seem to always happen, see https://github.com/rust-lang/rust/issues/140126

src/items/constant-items.md
```diff
@@ -49,32 +49,104 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
...
+const _: &&mut u8 = unsafe { &S }; // OK.
+//        ^^^^^^^
+// This mutable references comes from an extern static.
+```
+
+> [!NOTE]
+> This is allowed as it's separately not allowed to read from an external static during constant evaluation. See [const-eval.const-expr.path-static].
+
 > [!NOTE]
 > We also disallow, in the final value, shared references to mutable statics created in the initializer for a separate reason. Consider:
```
theemathas: This preexisting note seems incorrect. It talks about mutable statics, and then shows an example with a reference to an interior-mutable temporary. Such temporaries do not live for `'static`. And even if they were to, they would be a const-promoted thing, which is subtly different from statics.

src/items/constant-items.md
```diff
@@ -49,32 +49,104 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
...
 >
 > ```rust,compile_fail,E0492
 > # use core::sync::atomic::AtomicU8;
-> const C: &AtomicU8 = &AtomicU8::new(0); // ERROR
+> const _: &AtomicU8 = &AtomicU8::new(0); // ERROR.
 > ```
 >
 > Here, the `AtomicU8` is a temporary that is lifetime extended to `'static` (see [destructors.scope.lifetime-extension.static]), and references to lifetime-extended temporaries with interior mutability are not allowed in the final value of a constant expression (see [const-eval.const-expr.borrows]).
```
theemathas: I believe that interior-mutable values are not lifetime-extended to `'static` at all.

src/items/constant-items.md
traviscross: > Note that this only works with zero-length arrays. Also, this promotion does not seem to always happen, see [rust-lang/rust#140126](https://github.com/rust-lang/rust/issues/140126)

Look at the bottom note in <https://doc.rust-lang.org/nightly/reference/const_eval.html#constant-expressions> and https://github.com/rust-lang/rust/issues/143129.

src/items/constant-items.md
traviscross: In an initializer such as,

```rust
const _: &AtomicU8 = &AtomicU8::new(0);
```

the rules for ["extending based on expressions"](https://doc.rust-lang.org/nightly/reference/destructors.html#extending-based-on-expressions) apply as follows:

- The initializer is an extending expression.
- The borrow expression is therefore an extending expression, so the operand of the borrow is an extending expression.

The "operand of an extending [borrow](https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#r-expr.operator.borrow) expression has its [temporary scope](https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.temporary) [extended](https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.lifetime-extension)".  I.e., we extend the temporary scope of `AtomicU8::new(0)`.

How far do we extend it?  ["To the end of the program"](https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.lifetime-extension.static):

> Lifetime extension also applies to static and const items, where it makes temporaries live until the end of the program.

traviscross: We also specify that interior mutable values are not subject to constant promotion.

https://doc.rust-lang.org/nightly/reference/destructors.html#constant-promotion

> Promotion of a value expression to a `'static` slot occurs when the expression could be written in a constant and borrowed, and that borrow could be dereferenced where the expression was originally written, without changing the runtime behavior. That is, the promoted expression can be evaluated at compile-time and the resulting value does not contain [interior mutability](https://doc.rust-lang.org/nightly/reference/interior-mutability.html) or [destructors](https://doc.rust-lang.org/nightly/reference/destructors.html) (these properties are determined based on the value where possible, e.g. `&None` always has the type `&'static Option<_>`, as it contains nothing disallowed).

src/items/constant-items.md
RalfJung: Lifetime extension doesn't know whether things are interior mutable, so this does get lifetime extended. Later we then reject it when we notice that an interior mutable value has been lifetime extended.

src/items/constant-items.md
RalfJung: It is still confusing to call this a "mutable static", which one could reasonably interpret as referring to a `static mut` or at least some sort of `static` item with mutability.

traviscross: (Edit: GH didn't show me @RalfJung's comment; this was posted in parallel to it:)

All that said, I agree the wording was imperfect.  I've pushed a commit to clarify the point.

src/items/constant-items.md
traviscross: Also see discussion above.

src/items/constant-items.md
traviscross: > It is still confusing to call this a "mutable static", which one could reasonably interpret as referring to a `static mut` or at least some sort of `static` item with mutability.

Yes, agreed; this is avoided in the revision.

src/items/constant-items.md
```diff
@@ -49,35 +49,115 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
 ```
 
 r[items.const.no-mut-refs]
-The final value of a `const` item cannot contain any mutable references.
+The final value of a `const` item, after the initializer is evaluated to a value that has the declared type of the constant, cannot contain any mutable references except as follows.
```
tshepang: "except as follows" implies that the examples that follow succeed, but there are some that do not

maybe this can be structured like

```text
r[items.const.no-mut-refs]
The final value [..] as follows.

- Allowed due to a coercion...

  ```rust
  [..]

- Allowed due to...

  ```rust
  [..]

Following are examples that are not allowed.

- Not allowed as the mutable reference appears in the final value:

  ```rust
  [..]
```

oli-obk (commented):
Should we give the failing examples a rationale, too? The passing ones have rationales, but we don't explain why e.g. we don't want to allow creating references to interior mutable data from a const unless the reference is to a static item

src/items/constant-items.md
traviscross: I've reworded this slightly to make this less ambiguous.

traviscross:
> Should we give the failing examples a rationale, too? The passing ones have rationales, but we don't explain why e.g. we don't want to allow creating references to interior mutable data from a const unless the reference is to a static item

Sure.  I've added some rationales.  I'm curious, though, what would you say for this one?

```rust
# #![allow(static_mut_refs)]
static mut S: u8 = 0;
const _: &dyn Send = &unsafe { &mut S }; // ERROR.
//                             ^^^^^^
// Not allowed as the mutable reference appears in the final value,
// even though type erasure occurs.
```

(It's easier to say, for this sort of thing, why we allow something rather than why we don't, as we're conservatively only allowing the things that we have good reason to allow and where we can prove that it's OK to allow those things.)

oli-obk:
Yea that case is just a limitation of the few provenance checks happening in CTFE. We do allow

```rust
use std::sync::atomic::AtomicBool;

static S: AtomicBool = AtomicBool::new(false);
const _: &dyn Send = &unsafe { &S };
```

oli-obk (approved):

ehuss (approved):
Thanks!

src/items/constant-items.md
```diff
@@ -49,35 +49,135 @@ const BITS_N_STRINGS: BitsNStrings<'static> = BitsNStrings {
...
+```rust
+# #![allow(static_mut_refs)]
+static mut S: &mut u8 = unsafe { static mut I: u8 = 0; &mut I };
+const _: &&mut u8 = unsafe { &S }; // OK.
+//        ^^^^^^^
+// This mutable reference comes from a `static mut`.
+```
+
+> [!NOTE]
+> This is allowed as it's separately not allowed to read from a mutable static during constant evaluation. See [const-eval.const-expr.path-static].
```
RalfJung, line 147: This note is a bit confusing since it does not relate back to what we said above about why these references are disallowed.

It should probably also say that this is sound to allow because it take an unsafe block to create a mutable reference to such a static, and that unsafe block carries the obligation of ensuring that no unsound aliasing will occur, Furthermore, the problem regarding "is there a new instance of the pointee for every use of the const or not" does not come up since unambiguously, these references all point to `S`.

traviscross: > It should probably also say that this is sound to allow because it take an unsafe block to create a mutable reference to such a static, and that unsafe block carries the obligation of ensuring that no unsound aliasing will occur,

Not sure I follow here.  The example in question is:

```rust
#![allow(static_mut_refs)]
static mut S: &mut u8 = unsafe { static mut I: u8 = 0; &mut I };
const _: &&mut u8 = unsafe { &S }; // OK.
//        ^^^^^^^
// This mutable reference comes from a `static mut`.
```

When you say, "because it takes an unsafe block to create a mutable reference to such a static", I'm not sure how you're meaning to connect this to the definition or use of the const.  I.e., do you mean this with respect to the inner mutable static, and hence this mutable reference,

```rust
static mut S: &mut u8 = unsafe { static mut I: u8 = 0; &mut I };
//                                                     ^^^^^^
```

or that an unsafe block is needed to get the immutable reference to the outer static with the inner mutable reference,

```rust
const _: &&mut u8 = unsafe { &S };
//                  ^^^^^^^^^^^^^
```

or do you mean one further downstream?  Maybe you could give an example.

RalfJung: In this example, the `&mut` is actually harmless as it is below an `&`. So safe code can't actually cause UB with those two.

Is that a general principle?

My main point is that it is unclear why "it's separately not allowed to read from a mutable static during constant evaluation" implies that we can allow this `const _: &&mut u8`.

RalfJung: Let's continue this in https://github.com/rust-lang/reference/issues/2074. (Why does github not show a pingback for that?!?)

PR #2061, opened 2025-10-24
Move punctuation index to an appendix, and expand for other syntaxes

### LINKED ISSUE #211: Document rust's (hard-to-search) sigils, characters, and symbols in one appendix
This idea is inspired by [Clojure's "weird characters" guide](https://clojure.org/guides/weird_characters). They grouped all of Clojure's hard-to-search symbols into one page for easy reference.

Given that Rust has tons of sigils and symbols itself, and that such an easily searchable guide would be invaluable, is the reference the place for it? Does such a thing already exist?

ehuss:
This adds a new syntax index which shows some common, potentially hard-to-search syntax examples along with links to the definitions for those forms.

This is based on the syntax index that is in the book (https://doc.rust-lang.org/1.90.0/book/appendix-02-operators.html). There have been occasional requests for this, with links in the reference.

I decided to organize it by concepts instead of by syntax. For example `(…)` could be a grouped expression, type, or pattern. However, I think the amount of overlap between concepts is small, and I liked the idea of emphasizing the distinction of these concepts to really convey that they should not be conflated. But I would also be open to a more syntax-based organization.

This is not intended to be 100% exhaustive. There are many permutations and qualifiers, and I felt it would just make the list too long to try to cover everything.

Closes https://github.com/rust-lang/reference/issues/211

### DIFF
--- a/src/syntax-index.md
+++ b/src/syntax-index.md
@@ -4,6 +4,66 @@ This appendix provides an index of tokens and common forms with links to where t
 
 ## Keywords
 
+| Keyword       | Usage |
+|---------------|-------|
+| `_`           | [wildcard patterns], [inferred types], unnamed items in [constants], [extern crates], [use declarations], [destructuring assignment] |
+| `abstract`    | [reserved keyword] |
+| `as`          | [extern crate alias], [use alias], [type cast expressions], [qualified paths] |
+| `async`       | [async functions], [async blocks], [async closures] |
+| `await`       | [await expressions] |
+| `become`      | [reserved keyword] |
+| `box`         | [reserved keyword] |
+| `break`       | [break expressions] |
+| `const`       | [const functions], [const items], [const generics], [const blocks], [raw borrow operator], [raw pointer type], [const assembly operands] |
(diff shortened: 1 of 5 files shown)


traviscross (approved):

src/syntax-index.md
```diff
@@ -0,0 +1,459 @@
...
+| `extern "C" {…}`              | [external blocks] |
+| `fn ident<…>(…) …`<br>`struct ident<…> {…}`<br>`enum ident<…> {…}`<br>`impl<…> Type<…> {…}` | [generic definitions] |
+
+## Type expressions
+
+[Type expressions] are used to refer to types.
+
+| Type                                  | Use |
+|---------------------------------------|-----|
+| `bool`, `u8`, `f64`, `str`, `…`       | [primitive types] |
```
ehuss: The others don't use backticks.

```suggestion
| `bool`, `u8`, `f64`, `str`, …       | [primitive types] |
```

traviscross: Good catch.

PR #2062, opened 2025-10-24
document `min_global_align`

folkertdev:
related to https://github.com/rust-lang/rust/pull/142198
zullip thread: [#t-lang-docs/reference > respecting &#96;target.min_global_align&#96;](https://rust-lang.zulipchat.com/#narrow/channel/520709-t-lang-docs.2Freference/topic/respecting.20.60target.2Emin_global_align.60/with/547000981)

### DIFF
--- a/src/target-properties.md
+++ b/src/target-properties.md
@@ -1,0 +1,13 @@
+r[target-properties]
+# Target properties
+
+r[target-properties.pointer-width]
+The size of a pointer in bits is equal to the `pointer_width` specified by the target.
+
+r[target-properties.endianness]
+The endianness is equal to the `options.endian` specified by the target.
+
+r[target-properties.min_global_align]
+Every data [symbol] is aligned to at least the `options.min_global_align` specified by the target.
+
+[symbol]: glossary.md#symbol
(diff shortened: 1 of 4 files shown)


src/target-properties.md
```diff
@@ -0,0 +1,13 @@
...
+# Target properties
+
+r[target-properties.pointer-width]
+The size of a pointer in bits is equal to the `options.pointer_width` specified by the target.
+
+r[target-properties.endianness]
+The endianness is equal to the `endian` specified by the target.
+
+r[target-properties.min_global_align]
+Every data [symbol] is aligned to at least the `min_global_align` specified by the target.
```
traviscross: What can we point to as canonical documentation of these target properties?  Or is this the canonical documentation, e.g. with respect to the names?  E.g., why is it `options.pointer_width`?

folkertdev: I'm basing the names here off of 

- https://doc.rust-lang.org/beta/nightly-rustc/rustc_target/spec/struct.Target.html
- https://doc.rust-lang.org/beta/nightly-rustc/rustc_target/spec/struct.TargetOptions.html, 

Although looking at that again I flipped what attributes get the `options.` prefix.

The `target.json` format is technically unstable as far as I know. So, I'm not sure beyond referring to `Target` and `TargetOptions`.

PR #2063, opened 2025-10-24
document `cfg` conditions on inline assembly templates and operands

folkertdev:
Reference PR for:

- https://github.com/rust-lang/rust/pull/147736

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -107,6 +107,7 @@ Attributes may be applied to many forms in the language:
 * [Function][functions], [closure] and [function pointer]
   parameters accept outer attributes. This includes attributes on variadic parameters
   denoted with `...` in function pointers and [external blocks][variadic functions].
+* [Inline assembly] template strings and operands accept outer attributes.
 
 r[attributes.meta]
 ## Meta item attribute syntax
@@ -409,3 +410,4 @@ The following is an index of all built-in attributes.
 [variadic functions]: items/external-blocks.html#variadic-functions
 [`diagnostic::on_unimplemented`]: attributes/diagnostics.md#the-diagnosticon_unimplemented-attribute
 [`diagnostic::do_not_recommend`]: attributes/diagnostics.md#the-diagnosticdo_not_recommend-attribute
+[Inline assembly]: inline-assembly.md
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -107,7 +107,7 @@ Attributes may be applied to many forms in the language:
 * [Function][functions], [closure] and [function pointer]
   parameters accept outer attributes. This includes attributes on variadic parameters
   denoted with `...` in function pointers and [external blocks][variadic functions].
-* [Inline assembly] template strings and operands accept outer attributes.
+* [Inline assembly] template strings and operands accept outer attributes. Only certain attributes are accepted semantically; for details, see [asm.attributes.supported-attributes].
 
 r[attributes.meta]
 ## Meta item attribute syntax
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -276,7 +276,7 @@ r[asm.attributes]
 r[asm.attributes.supported-attributes]
 Only the [`cfg`] and [`cfg_attr`] attributes are accepted semantically on inline assembly template strings and operands. Other attributes are parsed, but rejected when the assembly macro is expanded.
 
-```rust,
+```rust
 # fn main() {}
 # #[cfg(target_arch = "x86_64")]
 core::arch::global_asm!(
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -274,7 +274,7 @@ r[asm.attributes]
 ## Attributes
 
 r[asm.attributes.supported-attributes]
-Only the [`cfg`] and [`cfg_attr`] attributes are accepted semantically on inline assembly template strings and operands. Other attributes are parsed, but rejected when the assembly macro is expanded.
+Only the [`cfg`] and [`cfg_attr`] attributes are accepted semantically on inline assembly template strings and operands. Other attributes are parsed but rejected when the assembly macro is expanded.
 
 ```rust
 # fn main() {}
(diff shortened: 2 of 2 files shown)


src/inline-assembly.md
```diff
@@ -263,6 +267,11 @@ These targets impose an additional restriction on the assembly code: any assembl
 Assembly code that does not conform to the GAS syntax will result in assembler-specific behavior.
 Further constraints on the directives used by inline assembly are indicated by [Directives Support](#directives-support).
 
+r[asm.ts-args.attributes]
+Only the [`cfg`] and [`cfg_attr`] attributes are accepted on inline assembly template strings and operands. Other attributes are parsed, but rejected when the assembly macro is expanded.
```
traviscross: When I said to put it here, I had momentarily forgotten that this applied to operands as well.

Given that, probably it shouldn't go here.  There's not really another obvious section for it, though.  Probably syntax is the closest, but maybe it'd be better to just create a new `asm.attributes` section, even just for this one purpose, as we're describing particularly what is semantically accepted, so it's not a great fit for `syntax`.

Probably too, we'll want to use the phrase "accepted semantically" in this sentence to be more clear upfront about the distinction we're making.

traviscross: Let's also please describe the rule imposed in https://github.com/rust-lang/rust/pull/147736#issue-3519295953:

> **operand before template**
> 
> The current implementation expects at least one template string before any operands. In the example below, if the `cfg` condition evaluates to true, the assembly block is ill-formed. But even when it evaluates to `false` this block is rejected, because the parser still expects just a template (a template is parsed as an expression and then validated to ensure that it is or expands to a string literal).
> 
> Changing how this works is difficult.
> 
> ```rust
> // This is rejected because `a = out(reg) x` does not parse as an expresion.
> asm!(
> 	#[cfg(false)]
> 	a = out(reg) x, //~ ERROR expected token: `,`
> 	"",
> );
> ```

Let's include the example in the text.

src/attributes.md
```diff
@@ -108,6 +108,7 @@ Attributes may be applied to many forms in the language:
 * [Function][functions], [closure] and [function pointer]
   parameters accept outer attributes. This includes attributes on variadic parameters
   denoted with `...` in function pointers and [external blocks][variadic functions].
+* [Inline assembly] template strings and operands accept outer attributes.
```
petrochenkov: Again, it may be a mostly theoretical concern, but attributes are not supported here.
`asm!` rather have an attribute-like syntax in its DSL - https://github.com/rust-lang/rust/pull/147736#issuecomment-3435781687.

The "only `cfg` and `cfg_attr`" restriction is unique to all attribute positions and is a direct consequence of the above.

folkertdev: This list is about allowed positions, and with `asm_cfg` attributes are parsed on template strings and operands, so saying that they are a valid position for attributes feels right.

But yeah they aren't really supported like in the other places listed here.

src/inline-assembly.md
```diff
@@ -266,6 +270,27 @@ Further constraints on the directives used by inline assembly are indicated by [
 [format-syntax]: std::fmt#syntax
 [rfc-2795]: https://github.com/rust-lang/rfcs/pull/2795
 
+r[asm.attributes]
+## Attributes
+
+r[asm.attributes.supported-attributes]
+Only the [`cfg`] and [`cfg_attr`] attributes are accepted semantically on inline assembly template strings and operands. Other attributes are parsed, but rejected when the assembly macro is expanded.
+
```
ehuss, line 278: Can you add an example here showing `cfg` being used?

src/attributes.md
ehuss: This is an interesting question because for example in things like https://github.com/rust-lang/reference/pull/1890 we document that diagnostics are allowed in all attribute positions. I'm not sure how to resolve that.

folkertdev: Are there any diagnostics that are meaningful on a (macro call that expands to a) template string? We could parse and then silently discard them, but actually making the diagnostic attribute do something would be tough I think.

folkertdev: `#[rustfmt::skip]` could be meaningful, but assembly is not formatted at all currently, so in practice it won't do anything today.

ehuss: I don't know. You could have something silly like `asm!(#[allow(named_asm_labels)] "warned: nop");`, but I don't think that is something worth doing.

I'm not suggesting that it is something that should be changed. I'm just trying to think of how we can cleanly fit it in to the reference text.

src/inline-assembly.md
folkertdev: I added 

```rust
core::arch::naked_asm!(
    #[cfg(not(panic = "abort"))]
    ".cfi_startproc",
    // ...
    "ret",
    #[cfg(not(panic = "abort"))]
    ".cfi_endproc",
);
```

Loosely based on https://github.com/nbdd0121/unwinding/blob/3bc28cec6c30eade0e1caa094b49db78101b19ef/src/unwinder/arch/aarch64.rs#L66C24-L66C40.

It's tough to come up with simple full examples, so hopefully this conveys the idea.

src/attributes.md
folkertdev: We could move the asm line to its own section here 

```md
r[attributes.asm]
[Inline assembly] template strings and operands accept outer attributes syntactically, but 
any attribute besides `cfg` and `cfg_attr` is rejected during expansion.
```

Then asm arguments are not semantically an allowed position for attributes, they just specifically accep `cfg` and `cfg_attr`.

the section name could be `allowed-asm` too, or `allowed.asm`, I'm not sure which is best.

traviscross: I added a note about the limitations and a forward reference in the attributes chapter.  In the inline assembly chapter, I added an admonition describing the situation.  See the commit messages for more details.

traviscross (approved):

traviscross:
I've pushed a series of revisions.  Some of this is editorial; some has potential normative effect on other parts of the document where we may refer to other attributes as being accepted wherever attributes are allowed.  For details, see the commit messages.

With these revisions, this looks good to me.

PR #2066, opened 2025-10-27
Use American spelling for "labeled"

traviscross:
We follow American spellings in the Reference, but we had used the Commonwealth spelling "labelled" rather than the American spelling "labeled".  Let's fix that and add the appropriate redirects.

cc @ehuss


ehuss (approved):
Thanks!

Issue #595, opened 2019-04-21
Make sure all type checking rules are covered.


ehuss:
librustc_typeck is a rather large crate that does a lot of things. Presumably there are a lot of rules in there that are not covered in the reference.

PR #2067, opened 2025-10-28
Add a chapter on divergence

### LINKED ISSUE #271: Divergence
With #206 merged, I need to write the divergence docs so it doesn't get lost. I plan on doing so Monday. Was going to yesterday, but I misread my schedule and went to work 4 hours early. >_>

### LINKED ISSUE #1033: Document when the return type of a block expression implicitly becomes `!`
The reference currently states

> The type of a block is the type of the final expression, or `()` if the final expression is omitted.

However, the type of this block appears to be `!`:
```rust
{
    loop {}
    5;
}
```
whereas the type of this block is `i32`
```rust
{
    loop {}
    5
}
```
So it looks like the presence of an expression of type `!` changes the default type of the block from `()` to `!`.

As a side note, the doucmentation could also make clearer that "omitting the final expression" essentially means terminating the block with a semicolon.

jackh726:
It was little tricky when trying to describe *diverging blocks*. The compiler's implementation maintains sort of a "global state" when checking an expression and sub-expressions, which it resets on conditional things. Semantically, I think the way I worded it is much clearer than trying to match the implementation.

Happy to hear any specific feedback, Lcnr made did an initial review pass in https://github.com/rust-lang/project-goal-reference-expansion/commit/407917146a090fac7ec3ae31132447bce8c98137, so the second commit tries to address that.

Closes https://github.com/rust-lang/reference/issues/271
Closes https://github.com/rust-lang/reference/issues/1033

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -98,6 +98,7 @@
     - [Subtyping and variance](subtyping.md)
     - [Trait and lifetime bounds](trait-bounds.md)
     - [Type coercions](type-coercions.md)
+    - [Divergence](divergence.md)
     - [Destructors](destructors.md)
     - [Lifetime elision](lifetime-elision.md)
 
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -19,6 +19,8 @@ See the following rules for specific expression divergence behavior:
 
 - [expr.block.diverging] --- Block expressions.
 - [expr.if.diverging] --- `if` expressions.
+- [expr.loop.block-labels.type] --- Labeled block expressions with `break`.
+- [expr.loop.break-value.diverging] --- `loop` expressions with `break`.
 - [expr.loop.break.diverging] --- `break` expressions.
 - [expr.loop.continue.diverging] --- `continue` expressions.
 - [expr.loop.infinite.diverging] --- Infinite `loop` expressions.
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -18,10 +18,13 @@ The following language constructs unconditonally produce a _diverging expression
 * [A `continue` expression](./expressions/loop-expr.md#r-expr.loop.continue.type)
 * [A `return` expression](./expressions/return-expr.md#r-expr.return.type)
 * [A `match` expression with no arms](./expressions/match-expr.md#r-expr.match.type.diverging.empty)
-* [A `block` expression that it itself is diverging.](../expressions/block-expr.md#r-expr.block.type.diverging)
+* [A `block` expression that it itself is diverging.](./expressions/block-expr.md#r-expr.block.type.diverging)
 
 r[divergence.diverging-expressions.conditional]
 In a control flow expression, if all arms diverge, then the entire expression also diverges.
+
+r[divergence.diverging-expressions.place-expressions]
+A place expression of the type [`!`](./types/never.md) is considering _diverging_ only if it is read from.
 
 r[divergence.fallback]
 ## Fallback
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -16,13 +16,13 @@ Any expression of type [`!`] is a _diverging expression_, but there are also div
 >
 > fn diverging() -> ! {
 >     // This has a type of `!`.
->     // So, the entire function is considered diverging
+>     // So, the entire function is considered diverging.
 >     make_never();
 >     // OK: The type of the body is `!` which matches the return type.
 > }
 > fn not_diverging() -> ! {
 >     // This type is uninhabited.
->     // However, the entire function is not considered diverging
+>     // However, the entire function is not considered diverging.
 >     make_empty();
 >     // ERROR: The type of the body is `()` but expected type `!`.
 > }
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -1,0 +1,52 @@
+r[divergence]
+# Divergence
+
+r[divergence.intro]
+Divergence is the state where a particular section of code could never be encountered at runtime. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block.
+
+Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
+
+r[divergence.diverging-expressions]
+## Producing diverging expressions
+
+r[divergence.diverging-expressions.unconditional]
+The following language constructs unconditonally produce a _diverging expression_ of the type [`!`](./types/never.md):
+
+* [A call to a function returning `!`.](./types/never.md#r-type.never.constraint)
+* [A `loop` expression with no corresponding break.](./expressions/loop-expr.md#r-expr.loop.infinite.diverging)
+* [A `break` expression](./expressions/loop-expr.md#r-expr.loop.break.type)
+* [A `continue` expression](./expressions/loop-expr.md#r-expr.loop.continue.type)
+* [A `return` expression](./expressions/return-expr.md#r-expr.return.type)
+* [A `match` expression with no arms](./expressions/match-expr.md#r-expr.match.type.diverging.empty)
+* [A `block` expression that it itself is diverging.](../expressions/block-expr.md#r-expr.block.type.diverging)
+
+r[divergence.diverging-expressions.conditional]
+In a control flow expression, if all arms diverge, then the entire expression also diverges.
+
+r[divergence.fallback]
+## Fallback
+If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be `!`.
+
+The following fails to compile because `!` does not implement `Debug`:
+```rust,compile_fail,E0277
+fn foo() -> i32 { 22 }
+match foo() {
+    4 => Default::default(),
+    _ => return,
+};
+```
+
+> [!EDITION-2024]
+> Before the 2024 edition, the type was inferred to instead be `()`.
+
+Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The following compiles:
+```rust
+fn foo() -> i32 { 22 }
+// This has the type `Option<!>`, not `!`
+match foo() {
+    4 => Default::default(),
+    _ => Some(return),
+};
+```
+
+<!-- TODO: This last point should likely should be moved to a more general "type inference" section discussing generalization + unification. -->
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -30,8 +30,6 @@ See the following rules for specific expression divergence behavior:
 > [!NOTE]
 > The [`panic!`] macro and related panic-generating macros like [`unreachable!`] also have the type [`!`] and are diverging.
 
-Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
-
 r[divergence.never]
 Any expression of type [`!`] is a diverging expression. However, diverging expressions are not limited to type [`!`]; expressions of other types may also diverge (e.g., `Some(loop {})` has type `Option<!>`).
 
@@ -56,6 +54,9 @@ Any expression of type [`!`] is a diverging expression. However, diverging expre
 >     // ERROR: The type of the body is `()` but expected type `!`.
 > }
 > ```
+
+> [!NOTE]
+> Divergence can propagate to the surrounding block. See [expr.block.diverging].
 
 r[divergence.fallback]
 ## Fallback
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -6,6 +6,25 @@ Divergence is the state where a particular section of code could never be encoun
 
 Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
 
+> [!NOTE]
+> Though `!` is considered an uninhabited type, a type being uninhabited is not sufficient for it to diverge.
+>
+> ```rust,compile_fail,E0308
+> # #![ feature(never_type) ]
+> # fn make<T>() -> T { loop {} }
+> enum Empty {}
+> fn diverging() -> ! {
+>     // This has a type of `!`.
+>     // So, the entire function is considered diverging
+>     make::<!>();
+> }
+> fn not_diverging() -> ! {
+>     // This type is uninhabited.
+>     // However, the entire function is not considered diverging
+>     make::<Empty>();
+> }
+> ```
+
 r[divergence.fallback]
 ## Fallback
 If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be `!`.
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -2,9 +2,9 @@ r[divergence]
 # Divergence
 
 r[divergence.intro]
-If an expression diverges, then nothing after that expression will execute. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
+If an expression diverges, then nothing after that expression will execute. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
 
-Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
+Any expression of type [`!`] is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
 
 > [!NOTE]
 > Though `!` is considered an uninhabited type, a type being uninhabited is not sufficient for it to diverge.
@@ -30,7 +30,8 @@ Any expression of type [`!`](./types/never.md) is a _diverging expression_, but 
 
 r[divergence.fallback]
 ## Fallback
-If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be `!`.
+
+If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be [`!`].
 
 > [!EXAMPLE]
 > ```rust,compile_fail,E0277
@@ -58,3 +59,5 @@ If a type to be inferred is only unified with diverging expressions, then that t
 > ```
 
 <!-- TODO: This last point should likely should be moved to a more general "type inference" section discussing generalization + unification. -->
+
+[`!`]: type.never
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -32,7 +32,8 @@ See the following rules for specific expression divergence behavior:
 
 Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
 
-Any expression of type [`!`] is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
+r[divergence.never]
+Any expression of type [`!`] is a diverging expression. However, diverging expressions are not limited to type [`!`]; expressions of other types may also diverge (e.g., `Some(loop {})` has type `Option<!>`).
 
 > [!NOTE]
 > Though `!` is considered an uninhabited type, a type being uninhabited is not sufficient for it to diverge.
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -10,18 +10,21 @@ Any expression of type [`!`](./types/never.md) is a _diverging expression_, but 
 > Though `!` is considered an uninhabited type, a type being uninhabited is not sufficient for it to diverge.
 >
 > ```rust,compile_fail,E0308
-> # #![ feature(never_type) ]
-> # fn make<T>() -> T { loop {} }
 > enum Empty {}
+> fn make_never() -> ! {loop{}}
+> fn make_empty() -> Empty {loop{}}
+>
 > fn diverging() -> ! {
 >     // This has a type of `!`.
 >     // So, the entire function is considered diverging
->     make::<!>();
+>     make_never();
+>     // OK: The type of the body is `!` which matches the return type.
 > }
 > fn not_diverging() -> ! {
 >     // This type is uninhabited.
 >     // However, the entire function is not considered diverging
->     make::<Empty>();
+>     make_empty();
+>     // ERROR: The type of the body is `()` but expected type `!`.
 > }
 > ```
 
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -15,6 +15,21 @@ fn example() {
 }
 ```
 
+See the following rules for specific expression divergence behavior:
+
+- [expr.block.diverging] --- Block expressions.
+- [expr.if.diverging] --- `if` expressions.
+- [expr.loop.break.diverging] --- `break` expressions.
+- [expr.loop.continue.diverging] --- `continue` expressions.
+- [expr.loop.infinite.diverging] --- Infinite `loop` expressions.
+- [expr.match.diverging] --- `match` expressions.
+- [expr.match.empty] --- Empty `match` expressions.
+- [expr.return.diverging] --- `return` expressions.
+- [type.never.constraint] --- Function calls returning `!`.
+
+> [!NOTE]
+> The [`panic!`] macro and related panic-generating macros like [`unreachable!`] also have the type [`!`] and are diverging.
+
 Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
 
 Any expression of type [`!`] is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -6,50 +6,32 @@ Divergence is the state where a particular section of code could never be encoun
 
 Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
 
-r[divergence.diverging-expressions]
-## Producing diverging expressions
-
-r[divergence.diverging-expressions.unconditional]
-The following language constructs unconditonally produce a _diverging expression_ of the type [`!`](./types/never.md):
-
-* [A call to a function returning `!`.](./types/never.md#r-type.never.constraint)
-* [A `loop` expression with no corresponding break.](./expressions/loop-expr.md#r-expr.loop.infinite.diverging)
-* [A `break` expression](./expressions/loop-expr.md#r-expr.loop.break.type)
-* [A `continue` expression](./expressions/loop-expr.md#r-expr.loop.continue.type)
-* [A `return` expression](./expressions/return-expr.md#r-expr.return.type)
-* [A `match` expression with no arms](./expressions/match-expr.md#r-expr.match.type.diverging.empty)
-* [A `block` expression that it itself is diverging.](./expressions/block-expr.md#r-expr.block.type.diverging)
-
-r[divergence.diverging-expressions.conditional]
-In a control flow expression, if all arms diverge, then the entire expression also diverges.
-
-r[divergence.diverging-expressions.place-expressions]
-A place expression of the type [`!`](./types/never.md) is considering _diverging_ only if it is read from.
-
 r[divergence.fallback]
 ## Fallback
 If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be `!`.
 
-The following fails to compile because `!` does not implement `Debug`:
-```rust,compile_fail,E0277
-fn foo() -> i32 { 22 }
-match foo() {
-    4 => Default::default(),
-    _ => return,
-};
-```
+> [!EXAMPLE]
+> ```rust,compile_fail,E0277
+> fn foo() -> i32 { 22 }
+> match foo() {
+>     // ERROR: The trait bound `!: Default` is not satisfied.
+>     4 => Default::default(),
+>     _ => return,
+> };
+> ```
 
 > [!EDITION-2024]
 > Before the 2024 edition, the type was inferred to instead be `()`.
 
-Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The following compiles:
-```rust
-fn foo() -> i32 { 22 }
-// This has the type `Option<!>`, not `!`
-match foo() {
-    4 => Default::default(),
-    _ => Some(return),
-};
-```
+> [!NOTE]
+> Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The > following compiles:
+> ```rust
+> fn foo() -> i32 { 22 }
+> // This has the type `Option<!>`, not `!`
+> match foo() {
+>     4 => Default::default(),
+>     _ => Some(return),
+> };
+> ```
 
 <!-- TODO: This last point should likely should be moved to a more general "type inference" section discussing generalization + unification. -->
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -43,7 +43,8 @@ If a type to be inferred is only unified with diverging expressions, then that t
 > Before the 2024 edition, the type was inferred to instead be `()`.
 
 > [!NOTE]
-> Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The > following compiles:
+> Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The following compiles:
+>
 > ```rust
 > fn foo() -> i32 { 22 }
 > // This has the type `Option<!>`, not `!`
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -4,7 +4,7 @@ r[divergence]
 r[divergence.intro]
 Divergence is the state where a particular section of code could never be encountered at runtime. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block.
 
-Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
+Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
 
 r[divergence.fallback]
 ## Fallback
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -2,7 +2,7 @@ r[divergence]
 # Divergence
 
 r[divergence.intro]
-Divergence is the state where a particular section of code could never be encountered at runtime. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block.
+If an expression diverges, then nothing after that expression will execute. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
 
 Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
 
--- a/src/divergence.md
+++ b/src/divergence.md
@@ -2,7 +2,20 @@ r[divergence]
 # Divergence
 
 r[divergence.intro]
-If an expression diverges, then nothing after that expression will execute. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
+A *diverging expression* is an expression that never completes normal execution.
+
+```rust
+fn diverges() -> ! {
+    panic!("This function never returns!");
+}
+
+fn example() {
+    let x: i32 = diverges(); // This line never completes.
+    println!("This is never printed: {x}");
+}
+```
+
+Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`], divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
 
 Any expression of type [`!`] is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(loop {})` produces a type of `Option<!>`).
 
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -100,11 +100,19 @@ fn diverging_place_read() -> ! {
     // A read of a place expression produces a diverging block
     let _x = foo.x;
 }
-fn diverging_place_not_read() -> () {
+```
+
+```rust,compile_fail,E0308
+# #![ feature(never_type) ]
+# fn make<T>() -> T { loop {} }
+# struct Foo {
+#     x: !,
+# }
+fn diverging_place_not_read() -> ! {
     let foo = Foo { x: make() };
     // Asssignment to `_` means the place is not read
     let _ = foo.x;
-}
+} // ERROR: Mismatched types.
 ```
 
 r[expr.block.value]
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -69,12 +69,12 @@ A block is itself considered to be [diverging](../divergence.md) if all reachabl
 # #![ feature(never_type) ]
 # fn make<T>() -> T { loop {} }
 fn no_control_flow() -> ! {
-    // There are no conditional statements, so this entire block is diverging.
+    // There are no conditional statements, so this entire function body is diverging.
     loop {}
 }
 
 fn control_flow_diverging() -> ! {
-    // All paths are diverging, so this entire block is diverging.
+    // All paths are diverging, so this entire function body is diverging.
     if true {
         loop {}
     } else {
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -97,7 +97,7 @@ struct Foo {
 
 fn diverging_place_read() -> ! {
     let foo = Foo { x: make() };
-    // A read of a place expression produces a diverging block
+    // A read of a place expression produces a diverging block.
     let _x = foo.x;
 }
 ```
@@ -110,7 +110,7 @@ fn diverging_place_read() -> ! {
 # }
 fn diverging_place_not_read() -> ! {
     let foo = Foo { x: make() };
-    // Asssignment to `_` means the place is not read
+    // Assignment to `_` means the place is not read.
     let _ = foo.x;
 } // ERROR: Mismatched types.
 ```
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -43,7 +43,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-Except in the case of divergence [(see below)](block-expr.md#r-expr.block.type.diverging), the type of a block is the type of the final operand, or `()` if the final operand is omitted.
+The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges][expr.block.type.diverging], in which case it is the [never type].
 
 ```rust
 # fn fn_call() {}
@@ -337,6 +337,7 @@ fn is_unix_platform() -> bool {
 [inner attributes]: ../attributes.md
 [method]: ../items/associated-items.md#methods
 [mutable reference]: ../types/pointer.md#mutables-references-
+[never type]: type.never
 [place expression]: expr.place-value.place-memory-location
 [scopes]: ../names/scopes.md
 [shared references]: ../types/pointer.md#shared-references-
@@ -345,6 +346,7 @@ fn is_unix_platform() -> bool {
 [struct]: struct-expr.md
 [the lint check attributes]: ../attributes/diagnostics.md#lint-check-attributes
 [tuple expressions]: tuple-expr.md
+[unit type]: type.tuple.unit
 [unsafe operations]: ../unsafety.md
 [value expressions]: ../expressions.md#place-expressions-and-value-expressions
 [Loops and other breakable expressions]: expr.loop.block-labels
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -66,6 +66,7 @@ r[expr.block.type.diverging]
 A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md), unless that expression is a [place expression](../expressions.md#r-expr.place-value.place-memory-location) that is not read from.
 
 ```rust,no_run
+# #![ feature(never_type) ]
 # fn make<T>() -> T { loop {} }
 fn no_control_flow() -> ! {
     // There are no conditional statements, so this entire block is diverging.
@@ -94,19 +95,15 @@ struct Foo {
     x: !,
 }
 
-fn diverging_place_read() -> () {
+fn diverging_place_read() -> ! {
     let foo = Foo { x: make() };
-    let _: ! = {
-        // A read of a place expression produces a diverging block
-        let _x = foo.x;
-    };
+    // A read of a place expression produces a diverging block
+    let _x = foo.x;
 }
 fn diverging_place_not_read() -> () {
     let foo = Foo { x: make() };
-    let _: () = {
-        // Asssignment to `_` means the place is not read
-        let _ = foo.x;
-    };
+    // Asssignment to `_` means the place is not read
+    let _ = foo.x;
 }
 ```
 
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -43,7 +43,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges][expr.block.type.diverging], in which case it is the [never type].
+The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges][expr.block.diverging], in which case it is the [never type].
 
 ```rust
 # fn fn_call() {}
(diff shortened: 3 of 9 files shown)


src/expressions/block-expr.md
```diff
@@ -60,8 +60,8 @@ let five: i32 = {
 assert_eq!(5, five);
 ```
 
-> [!NOTE]
-> As a control flow expression, if a block expression is the outer expression of an expression statement, the expected type is `()` unless it is followed immediately by a semicolon.
```
jackh726, line 63: This note I removed made absolutely no sense to me, and seems to be quite old with no justification of what it is trying to say.

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+## Producing diverging expressions
+
+r[divergence.diverging-expressions.unconditional]
+The following language constructs unconditonally produce a _diverging expression_ of the type [`!`](./types/never.md):
+
+* [A call to a function returning `!`.](./types/never.md#r-type.never.constraint)
+* [A `loop` expression with no corresponding break.](./expressions/loop-expr.md#r-expr.loop.infinite.diverging)
+* [A `break` expression](./expressions/loop-expr.md#r-expr.loop.break.type)
+* [A `continue` expression](./expressions/loop-expr.md#r-expr.loop.continue.type)
+* [A `return` expression](./expressions/return-expr.md#r-expr.return.type)
```
traviscross: Probably I'd expect this to appear in:

- https://doc.rust-lang.org/1.90.0/reference/expressions/return-expr.html#return-expressions

We do say:

> Evaluating a `return` expression moves its argument into the designated output location for the current function call, destroys the current function activation frame, and transfers control to the caller frame.

That's a description of a diverging behavior; probably we should use the word and describe the type of the expression.

src/expressions/block-expr.md
```diff
@@ -44,7 +44,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-The type of a block is the type of the final operand, or `()` if the final operand is omitted.
+Typically, the type of a block is the type of the final operand, or `()` if the final operand is omitted.
```
traviscross: Saying "typically" is a bit loose for the normative text.  Perhaps it'd be better to put the diverging rule first (and to mention that such a block is of the never type) and then, in this rule, to say, "Otherwise, the type of a block is...".  Or, alternatively, in this rule, to say "the type of a block is... except ...".

src/expressions/block-expr.md
```diff
@@ -60,8 +60,8 @@ let five: i32 = {
 assert_eq!(5, five);
 ```
 
-> [!NOTE]
-> As a control flow expression, if a block expression is the outer expression of an expression statement, the expected type is `()` unless it is followed immediately by a semicolon.
+r[expr.block.type.diverging]
+A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md#r-divergence.diverging-expressions).
```
traviscross: It'd be good to add some examples and counterexamples under this rule.

src/expressions/match-expr.md
```diff
@@ -96,6 +96,28 @@ Every binding in each `|` separated pattern must appear in all of the patterns i
...
+> [!NOTE]
+> If even the entire `match` expression diverges, its type may not be [`!`](../types/never.md).
+>
+>```rust,compile_fail,E0004
+>    let a = match true {
+>        true => Some(panic!()),
+>        false => None,
+>    };
+>    // Fails to compile because `a` has the type `Option<!>`
+>    // (or, `Option<()>` in edition 2021 and below)
```
traviscross:
```suggestion
>    // Fails to compile because `a` has the type `Option<!>`.
```

Outside of edition admonitions, we only document the behavior of the current edition.

src/expressions/match-expr.md
```diff
@@ -96,6 +96,28 @@ Every binding in each `|` separated pattern must appear in all of the patterns i
...
 Every binding of the same name must have the same type, and have the same binding mode.
 
+r[expr.match.type]
+The type of the overall `match` expression is the [least upper bound](../type-coercions.md#r-coerce.least-upper-bound) of the individual match arms.
+
+r[expr.match.type.diverging.empty]
+If there are no match arms, then the `match` expression is diverging and the type is [`!`](../types/never.md).
+
+r[expr.match.type.diverging.conditional]
+If either the scrutinee expression or all of the match arms diverge, then the entire `match` expression also diverges.
```
traviscross, line 113: Regarding the rule identifiers, we've been matching the hierarchical structural of these to the header hierarchy.  Here, I'd probably suggest `expr.match.type`, `expr.match.empty`, and `expr.match.diverging`.

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+If a type to be inferred is only unified with diverging expressions, then that type will be inferred to be `!`.
+
+The following fails to compile because `!` does not implement `Debug`:
+```rust,compile_fail,E0277
+fn foo() -> i32 { 22 }
+match foo() {
+    4 => Default::default(),
+    _ => return,
+};
+```
```
traviscross: You mean does not implement `Default`, I think.

To match style, either this entire bit should be moved into an `!EXAMPLE` admonition or the narrative about it should be moved into a comment in the code.  The error in the example should be notated with, e.g.:

```rust
// ERROR: The trait bound `!: Default` is not satisfied.
```

(It does not need to match the compiler output; use whatever description is most concise and clear in context to the reader.  If the reason is adequately described in other narrative comments, simply notating it with `// ERROR.` is OK.)

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+
+Importantly, type unification may happen *structurally*, so the fallback `!` may be part of a larger type. The following compiles:
+```rust
+fn foo() -> i32 { 22 }
+// This has the type `Option<!>`, not `!`
+match foo() {
+    4 => Default::default(),
+    _ => Some(return),
+};
+```
```
traviscross: Probably this makes sense in a `!NOTE` admonition.

src/expressions/block-expr.md
traviscross: I imagine it means this:

```rust
fn main() {
    ();
    { 0u8 } //~ error: mismatched types
    ();
}
```

```rust
fn main() {
    ();
    { 0u8 }; // OK.
    ();
}
```

```rust
fn main() {
    ();
    {} // OK.
    ();
}
```

The "as a control flow expression" verbiage is also used in a rule further above.

src/expressions/loop-expr.md
```diff
@@ -308,6 +308,9 @@ Example:
 r[expr.loop.break.value]
 A `break` expression is only permitted in the body of a loop, and has one of the forms `break`, `break 'label` or ([see below](#break-and-loop-values)) `break EXPR` or `break 'label EXPR`.
 
+r[expr.loop.break.type]
+A `break` expression itself has a type of [`!`](../types/never.md).
```
traviscross: Probably I'd locate this rule closer to the existing text:

> When `break` is encountered, execution of the associated loop body is immediately terminated...

I.e., that other rule is already describing divergence.

Have a look at this as well, if you would, for the other similar rules for other expressions.

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+r[divergence.diverging-expressions.unconditional]
+The following language constructs unconditonally produce a _diverging expression_ of the type [`!`](./types/never.md):
+
+* [A call to a function returning `!`.](./types/never.md#r-type.never.constraint)
+* [A `loop` expression with no corresponding break.](./expressions/loop-expr.md#r-expr.loop.infinite.diverging)
+* [A `break` expression](./expressions/loop-expr.md#r-expr.loop.break.type)
+* [A `continue` expression](./expressions/loop-expr.md#r-expr.loop.continue.type)
+* [A `return` expression](./expressions/return-expr.md#r-expr.return.type)
+* [A `match` expression with no arms](./expressions/match-expr.md#r-expr.match.type.diverging.empty)
+* [A `block` expression that it itself is diverging.](./expressions/block-expr.md#r-expr.block.type.diverging)
```
traviscross: We prefer to not duplicate normative rules, and this ends up adding a rule that duplicates the similar rules already present in the section for each expression.

Perhaps this should just be dropped.  If we were to decide to keep it, it'd probably go in an admonition, in an index, etc.

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+
+r[divergence.diverging-expressions.conditional]
+In a control flow expression, if all arms diverge, then the entire expression also diverges.
```
traviscross: This rule should probably be inlined into each control flow expression.

src/divergence.md
```diff
@@ -0,0 +1,55 @@
...
+
+r[divergence.diverging-expressions.place-expressions]
+A place expression of the type [`!`](./types/never.md) is considering _diverging_ only if it is read from.
```
traviscross: What's a good example for this?  It'd be good to include that.

src/SUMMARY.md
```diff
@@ -97,6 +97,7 @@
     - [Subtyping and variance](subtyping.md)
     - [Trait and lifetime bounds](trait-bounds.md)
     - [Type coercions](type-coercions.md)
+    - [Divergence](divergence.md)
```
traviscross, line 101: We try to keep the number of top-level chapters contained.  Looking at it, perhaps most of what's here that can't be inlined on the pages for each expression would make sense appearing in the [Expressions](https://doc.rust-lang.org/1.90.0/reference/expressions.html) chapter (e.g., we talk about place and value expressions there -- the verbiage about when a place expression is diverging might make sense near that) and in the [Never type](https://doc.rust-lang.org/1.90.0/reference/types/never.html) chapter.

src/expressions/match-expr.md
```diff
@@ -96,6 +96,28 @@ Every binding in each `|` separated pattern must appear in all of the patterns i
 r[expr.match.binding-restriction]
 Every binding of the same name must have the same type, and have the same binding mode.
 
+r[expr.match.type]
+The type of the overall `match` expression is the [least upper bound](../type-coercions.md#r-coerce.least-upper-bound) of the individual match arms.
```
traviscross: If we're inlining the rule about determining the type based on the LUB for match, from https://doc.rust-lang.org/1.90.0/reference/type-coercions.html#r-coerce.least-upper-bound.intro, probably we'd need to do that for the other rules there also (and then either remove the list from there or convert it to an admonition or index).

traviscross:
Thanks for the PR @jackh726; I can tell this was written carefully.  It will be good to get more of this documented.  In particular, it'll be good to have the fallback behavior documented.

I'll leave some notes inline.  Probably we'll want to move some things around.

Adding more examples -- even beyond what I'll note specifically inline -- would be particularly good for this material.  It's helpful when each rule has one or more concise and testable examples demonstrating exactly what the rule means to express.

jackh726:
Thanks for the review @traviscross. Good points here, I'll work on sorting through them today/tomorrow.

Happy to jump on a call at some point too, if you think any of these could use further discussion.

src/expressions/block-expr.md
```diff
@@ -63,6 +63,48 @@ assert_eq!(5, five);
 > [!NOTE]
 > As a control flow expression, if a block expression is the outer expression of an expression statement, the expected type is `()` unless it is followed immediately by a semicolon.
 
+r[expr.block.type.diverging]
+A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md#r-divergence.diverging-expressions), unless that expression is a place expression that is not read from.
+
+```rust
```
traviscross:
```suggestion
```rust,no_run
```

There are infinite loops, but that's OK as we just want to demonstrate the static semantics.

src/expressions/block-expr.md
```diff
@@ -63,6 +63,48 @@ assert_eq!(5, five);
...
+    x: !,
+}
+
+let foo = Foo { x: make() };
+let diverging_place_not_read: () = {
+    let _: () = {
+        // Asssignment to `_` means the place is not read
+        let _ = foo.x;
+    };
+};
```
traviscross: This compiles and produces an infinite loop.

[Playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&code=%23%21%5Bfeature%28never_type%29%5D%0A%0Afn+make%3CT%3E%28%29+-%3E+T+%7B+loop+%7B%7D+%7D%0A%0Astruct+Foo+%7B%0A++++x%3A+%21%2C%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+foo+%3D+Foo+%7B+x%3A+make%28%29+%7D%3B%0A++++let+diverging_place_not_read%3A+%28%29+%3D+%7B%0A++++++++let+_%3A+%28%29+%3D+%7B%0A++++++++++++%2F%2F+Asssignment+to+%60_%60+means+the+place+is+not+read%0A++++++++++++let+_+%3D+foo.x%3B%0A++++++++%7D%3B%0A++++%7D%3B%0A%7D%0A)

But then, so does:

```rust
    let foo = Foo { x: make() };
    let diverging_place_not_read: () = {
        let _: () = {
            // Asssignment to something other than `_` means?
            let _x = foo.x;
        };
    };
```

[Playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&code=%23%21%5Bfeature%28never_type%29%5D%0A%0Afn+make%3CT%3E%28%29+-%3E+T+%7B+loop+%7B%7D+%7D%0A%0Astruct+Foo+%7B%0A++++x%3A+%21%2C%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+foo+%3D+Foo+%7B+x%3A+make%28%29+%7D%3B%0A++++let+diverging_place_not_read%3A+%28%29+%3D+%7B%0A++++++++let+_%3A+%28%29+%3D+%7B%0A++++++++++++%2F%2F+Asssignment+to+something+other+than+%60_%60+means%3F%0A++++++++++++let+_x+%3D+foo.x%3B%0A++++++++%7D%3B%0A++++%7D%3B%0A%7D%0A)

Is there a way to demonstrate this such that `let _ = ..` produces distinct compile-time or runtime behavior from `let _x = ..`?

src/expressions/block-expr.md
```diff
@@ -63,6 +63,48 @@ assert_eq!(5, five);
...
+r[expr.block.type.diverging]
+A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md#r-divergence.diverging-expressions), unless that expression is a place expression that is not read from.
+
+```rust
+# #![ feature(never_type) ]
+# fn make<T>() -> T { loop {} }
+let no_control_flow: ! = {
+    // There are no conditional statements, so this entire block is diverging.
+    loop {}
+};
```
traviscross: You and I had talked about this.  We had both thought, "maybe it's worth using the nightly never type to express this more clearly."  I had mentioned I'd need to talk it over with @ehuss.  In that discussion, @ehuss made a good point: why not just use functions?  I.e., for the expression whose type we want to demonstrate, we can make that the trailing expression of a function that returns `!`.  E.g.:

```rust
fn no_control_flow() -> ! {
    loop {}
}
```

That does seem likely the best approach.  Sound right to you?

src/expressions/block-expr.md
traviscross: For this one, to remove use of the nightly never type, perhaps we could get there with generics.  E.g.:

```rust
fn make() -> ! { loop {} }

struct Foo<T> {
    x: T,
}

let foo = Foo { x: make() };
```

Etc.

src/divergence.md
```diff
@@ -0,0 +1,37 @@
...
+
+> [!EXAMPLE]
+> ```rust,compile_fail,E0277
+> fn foo() -> i32 { 22 }
+> match foo() {
+>     // ERROR: The trait bound `!: Default` is not satisfied.
+>     4 => Default::default(),
+>     _ => return,
+> };
+> ```
```
traviscross: It feels that we might need to say more here to guard against too simple a reading.  We say, 1) not all diverging expressions have type `!` (e.g. `Some(panic!())`), and 2) if a type to be inferred is only unified with diverging expressions, then the type will be inferred to be `!`.

However, of course, this does not compile:

```rust
trait Tr: Sized {
    fn m() -> Self { loop {} }
}
impl<T> Tr for T {}

fn f() -> u8 { 0 }
fn g() -> ! {
    match f() {
        0 => Tr::m(),
        //   ^^^^^^^ There's a type to be inferred here.
        _ => Some(panic!()),
        //   ^^^^^^^^^^^^^^ This is a diverging expression.
    } // ERROR: Mismatched types.
}
```

What might we be able to say to tease this apart?

src/expressions/block-expr.md
traviscross: To answer the question posed, this compiles:

```rust
trait RetId { type Ty; }
impl<T> RetId for fn() -> T { type Ty = T; }

struct S<T> {
    x: T,
}

fn f(x: S<<fn() -> ! as RetId>::Ty>) -> ! {
    let _x = x.x; // OK.
}
```

But this does not:

```rust
fn f(x: S<<fn() -> ! as RetId>::Ty>) -> ! {
    let _ = x.x; // ERROR: Mismatched types.
}
```

This is one, though, where it's not immediately coming to mind how to express this without either the never type or the never type hack.

traviscross: See instead https://github.com/rust-lang/reference/pull/2067/files#r2485028962.

traviscross: Here's one way.  (Of course, this is really just another instantiation of the never type hack.)

```rust
fn phantom_call<T>(_: impl FnOnce(T) -> T) {}

let _ = phantom_call(|x| -> ! {
    let _x = x; // OK.
});
let _ = phantom_call(|x| -> ! {
    let _ = x; // ERROR: Mismatched types.
});
```

traviscross: It's interesting how it really does need the `!` type to be ascribed for this to work.  I.e., it doesn't work with:

```rust
struct W<T>(T);

let x = W(loop {});
let _ = || -> ! {
    let _x = x.0; // ERROR.
};
```

Any thoughts about the reason for that?

src/expressions/match-expr.md
traviscross: As an aside, looking into this rule is what prompted me to file:

- https://github.com/rust-lang/rust/issues/148283

src/divergence.md
traviscross: This continues in https://github.com/rust-lang/reference/pull/2067#discussion_r2484963471.

src/expressions/block-expr.md
jackh726: > It's interesting how it really does need the `!` type to be ascribed for this to work. I.e., it doesn't work with:
> 
> ```rust
> struct W<T>(T);
> 
> let x = W(loop {});
> let _ = || -> ! {
>     let _x = x.0; // ERROR.
> };
> ```
> 
> Any thoughts about the reason for that?

This one nearly stumped me, because I had my mind set that this was a closure thing. But then I realized, this also fails:
```rust
let _: ! = {
    let _x = x.0;
};
```

The important bit to remember here is that `x.0` is an inference var which does *not* produce divergence (only place expressions that produce types of `!` do). The "expectation" value of the block (either as the block's type, or the closure's return value) does not constrain that inference variable.

traviscross: Makes sense, thanks.  (And yes, the only purpose of the closure in the examples is to be able to ascribe the type without nightly features.)

Coincidentally, prompted by lcnr apropos of lcnr's work, I was just looking at something else today with a similar flavor:

```rust
trait Tr: Sized {
    fn abs(self) -> i32 { 0 }
}
impl Tr for i32 {}

fn main() {
    let x = -42;
    let y: i32 = x.abs();
    assert!(y != x.abs()); // Surprising, but true.
}
```

src/expressions/block-expr.md
```diff
@@ -44,7 +44,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-The type of a block is the type of the final operand, or `()` if the final operand is omitted.
+Except in the case of divergence (see below), the type of a block is the type of the final operand, or `()` if the final operand is omitted.
```
yaahc: not positive this link is correct but I did my best
```suggestion
Except in the case of divergence [(see below)](block-expr.md#r-expr.block.type.diverging), the type of a block is the type of the final operand, or `()` if the final operand is omitted.
```

src/expressions/block-expr.md
```diff
@@ -63,6 +63,48 @@ assert_eq!(5, five);
 > [!NOTE]
 > As a control flow expression, if a block expression is the outer expression of an expression statement, the expected type is `()` unless it is followed immediately by a semicolon.
 
+r[expr.block.type.diverging]
+A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md#r-divergence.diverging-expressions), unless that expression is a place expression that is not read from.
```
yaahc: place expression could also use a link: expressions.html#r-expr.place-value.place-memory-location

src/divergence.md
```diff
@@ -0,0 +1,37 @@
+r[divergence]
+# Divergence
+
+r[divergence.intro]
+Divergence is the state where a particular section of code could never be encountered at runtime. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block.
+
+Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
```
yaahc: I got confused at first by the second half of this sentence. Can you expand on the `e.g.` by specifying the expressions type? I'm assuming it is `Option<!>`

```suggestion
Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())` with type `Option<!>`).
```

also: should this set of expressions be enumerated?

src/divergence.md
yaahc: is this addressed by the bit about type unification happening structurally below at https://github.com/rust-lang/reference/pull/2067/files#diff-11f5e85172f632195dd6a9a80daa852911fd3d66e788177a3bb39f6c5d7f21cfR27 ?

src/expressions/match-expr.md
```diff
@@ -96,6 +96,27 @@ Every binding in each `|` separated pattern must appear in all of the patterns i
...
+
+r[expr.match.empty]
+If there are no match arms, then the `match` expression is diverging and the type is [`!`](../types/never.md).
```
yaahc: does this require a diverging scrutinee expression to compile?

probably worth adding an example either way

src/divergence.md
jackh726: It's a good observation. I've tried to be a little nuanced in my wording - for example, in the list of expression that I removed from your review, I had the wording "the following language constructs unconditonally produce a _diverging expression_ of the type `!`". There also still is the wording "Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!`"

src/expressions/match-expr.md
jackh726: This is a good question! The answer is (subtly) no. Take the following:
```rust
fn make<T>() -> T { loop {} }
enum Empty {}

fn diverging_match_no_arms() -> ! {
    let e: Empty = make();
    match e {}
}
```

`e` is uninhabited, but is not considered diverging from a type-checking standpoint. To make it more confusing, this *also* fires the `unreachable_code` lint, like diverging expressions, but not in the same code path

Adding this as an example.

src/divergence.md
jackh726: > I got confused at first by the second half of this sentence. Can you expand on the `e.g.` by specifying the expressions type? I'm assuming it is `Option<!>`
> Suggested change
> Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
> Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())` with type `Option<!>`).

I can expand on this, but  I think this exact suggestion is not actually correct; `Some(panic!())` will only end up with a type of `Option<!>` if the inner variable is *unconstrained*.

> also: should this set of expressions be enumerated?

See https://github.com/rust-lang/reference/pull/2067#discussion_r2475872395. I had this before but TC suggested to remove it.

jackh726: > > I got confused at first by the second half of this sentence. Can you expand on the `e.g.` by specifying the expressions type? I'm assuming it is `Option<!>`
> > Suggested change
> > Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())`).
> > Any expression of type [`!`](./types/never.md) is a _diverging expression_, but there are also diverging expressions which are not of type `!` (e.g. `Some(panic!())` with type `Option<!>`).
> 
> I can expand on this, but I think this exact suggestion is not actually correct; `Some(panic!())` will only end up with a type of `Option<!>` if the inner variable is _unconstrained_.

Actually, I think this is right/fine (and I'm going to change it to `Some(loop {})`. It *does* produce a type of `Option<!>`, but there is also never-to-any coercions that are talked about elsewhere.

src/expressions/block-expr.md
```diff
@@ -63,6 +63,51 @@ assert_eq!(5, five);
...
 > As a control flow expression, if a block expression is the outer expression of an expression statement, the expected type is `()` unless it is followed immediately by a semicolon.
 
+r[expr.block.type.diverging]
+A block is itself considered to be [diverging](../divergence.md) if all reachable control flow paths contain a [diverging expression](../divergence.md), unless that expression is a [place expression](../expressions.md#r-expr.place-value.place-memory-location) that is not read from.
+
+```rust,no_run
+# #![ feature(never_type) ]
+# fn make<T>() -> T { loop {} }
+fn no_control_flow() -> ! {
+    // There are no conditional statements, so this entire block is diverging.
```
nikomatsakis: Does "this entire block" refer to the block of the fn body? I found it a touch ambiguous.

src/divergence.md
```diff
@@ -0,0 +1,37 @@
+r[divergence]
+# Divergence
+
+r[divergence.intro]
+Divergence is the state where a particular section of code could never be encountered at runtime. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block.
```
nikomatsakis: I would rather say:

"Divergence is the state where a particular section of code will never terminate" -- although it too is approximate.

"If an expression *diverges*, then nothing after that expression will execute"

something like that I guess.

nikomatsakis (commented):
Some drive-by comments.

src/expressions/match-expr.md
yaahc: lgtm, I was wondering if there would be a good place to link to the glossary item on uninhabited types as related but distinct but I couldn't find a good place. Wanted to mention the idea in case you can think of a way to fit it in and you think its valuable to do so.

jackh726: I think it's useful to mention this, thoughts on this: https://github.com/rust-lang/reference/pull/2067/commits/fa9825d290f8a58ff7850a339f1955ff7fa14db2

jackh726:
Okay, I have addressed the reviews to the best of my ability.

regarding moving `divergence.md` content elsewhere (https://github.com/rust-lang/reference/pull/2067#discussion_r2475967232):
I'm really not sure where else the remaining text can or should go. Importantly, this isn't a top-level chapter, it is under "Type system", which makes sense to me. It could be under some future "Type inference" sub-chapter, but that doesn't exist today and I'm not even sure.

regarding inlining other rules (https://github.com/rust-lang/reference/pull/2067#discussion_r2475977414):
I have no idea what is expected here.

regarding adding more examples:
I've added more examples over the course of addressing reviews, but if there are specific places that need more, I'm happy to respond specifically.

@rustbot review

src/expressions/match-expr.md
traviscross: Makes sense.

> Though `!` is considered an uninhabited type, a type being uninhabited is not sufficient for it to diverge.

As written, I'd probably move that into an admonition.  I.e., if what it's saying is already implied by the other rules, it doesn't need to be said normatively -- it's just a clarifying parenthetical remark for the reader.  If that's not the case, here, then probably it'd be better to restate the rule positively.  I.e., rather than saying what's not sufficient, enumerate exactly the sufficiency conditions.

src/SUMMARY.md
traviscross: > We try to keep the number of top-level chapters contained.

My mistake.  I misread the diff.  As you noted elsewhere, this is a subchapter of the type system chapter.

src/expressions/match-expr.md
jackh726: Changed this to be a `NOTE`.

src/expressions/block-expr.md
```diff
@@ -63,6 +63,51 @@ assert_eq!(5, five);
...
+    if true {
+        ()
+    } else {
+        loop {}
+    }
+}
+
+struct Foo {
+    x: !,
+}
```
ehuss, line 99: Can this be rewritten to use an empty enum to avoid the need for nightly features? For example:


```suggestion
enum Empty {}

struct Foo {
    x: Empty,
}
```

src/items/functions.md
```diff
@@ -56,7 +56,7 @@ r[items.fn.signature]
 Functions may declare a set of *input* [*variables*][variables] as parameters, through which the caller passes arguments into the function, and the *output* [*type*][type] of the value the function will return to its caller on completion.
 
 r[items.fn.implicit-return]
-If the output type is not explicitly stated, it is the [unit type].
+If the output type is not explicitly stated, it is the [unit type]. However, if the block expression is not [diverging](../divergence.md), then the output type is instead [`!`](../types/never.md).
```
ehuss: I don't understand this change. Can you explain what it is getting at?

We were initially thinking the `not` is wrong, but that also seems wrong since the implicit return type is still `()` for a function with a diverging block.

```rust
#![feature(never_type)]
pub fn f() {
    panic!("");
}

struct S {
    x: !
}

pub fn z() {
    S {x: f()}; // ERROR: expected `!`, found `()`
}
```

traviscross: In talking over this PR in the office hours, @ehuss and I puzzled over this rule.  Presumably the *not* is misplaced, but even with that removed, we weren't clear on how to assign meaning to this given, of course:

```rust
fn f() {
    loop {} // Diverging expression.
}

let _ = || -> ! { f() }; //~ ERROR mismatched types
let _: fn() -> ! = f; //~ ERROR mismatched types
```

traviscross:
@ehuss and I talked over this PR in the lang-docs office hours this week.  We're happy to be getting this documented; thanks to @jackh726 for that.

To set expectations here: @ehuss is planning to make a series of editorial revisions.  He'll push those onto this branch.  Our feeling was that, with these editorial revisions, this is likely to merge soon after.  Due to the upcoming holidays, this will happen in 2026.

src/items/functions.md
jackh726: Ah, this is a good catch. Indeed, the additional language is not correct for the function *signature*.

What I had in mind when writing this was that when type checking the function *body* the implicit return is `!` if it is diverging. This is *technically* covered by my changes to `expr.block.type.diverging`, so I'm not sure why I thought that these changes were also needed.

I think it was motivated by [this example](https://github.com/rust-lang/reference/pull/2067/changes#diff-77f5decfa4cf1deb58fc8f984f75b6702dfaa27986f8dce96f8fd1e5b3bb956aR80). I guess, maybe there is an equivalent statement missing about the implicit "final type" of a block (which is by default `()` and otherwise `!` when diverging).

theemathas:
```rust
#![allow(unreachable_code)]

fn one() -> impl Sized {
    panic!();
    ();
}

fn two() -> impl Sized {
    panic!();
    ()
}

fn main() {
    println!("{}", std::any::type_name_of_val(&(one as fn() -> _)));
    println!("{}", std::any::type_name_of_val(&(two as fn() -> _)));
}
```

The above code outputs:
```
fn() -> !
fn()
```

It seems that, if a block expression does not have a trailing semicolon (or equivalently, has a trailing expression), then the compiler ignores whether any statements inside diverge.

Is this intended? And is this addressed by this PR?

ehuss:
@theemathas We're discussing that above in https://github.com/rust-lang/reference/pull/2067#discussion_r2630177707.

EDIT: Or maybe I'll stand corrected, since that example depends on opaque types...

src/expressions/block-expr.md
```diff
@@ -44,7 +44,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-The type of a block is the type of the final operand, or `()` if the final operand is omitted.
+Except in the case of divergence [(see below)](block-expr.md#r-expr.block.type.diverging), the type of a block is the type of the final operand, or `()` if the final operand is omitted.
```
ehuss: This gives an exception for divergence, but doesn't explicitly say that the type of the block is the never type if the block diverges. I'm wondering if something like the following might clear that up?

```suggestion
The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges], in which case it is the [never type].
```

I'm guessing this logic comes from somewhere in [`check_expr_block`](https://github.com/rust-lang/rust/blob/b53da9908134f77da3bb36e7009ac81f02ee48dc/compiler/rustc_hir_typeck/src/fn_ctxt/checks.rs#L1014).

However, there's some things in there about `may_break` I don't quite follow.

src/expressions/block-expr.md
jackh726: No, [uninhabited types are not considered diverging from a type-system perspective](https://github.com/rust-lang/reference/pull/2067/changes#diff-11f5e85172f632195dd6a9a80daa852911fd3d66e788177a3bb39f6c5d7f21cfR10).

src/expressions/if-expr.md
```diff
@@ -73,6 +73,31 @@ let y = if 12 * 15 > 150 {
...
+```rust,no_run
+fn diverging_condition() -> ! {
+    // Diverges because the condition expression diverges
+    if loop {} {
+        ()
+    } else {
+        ()
+    };
+    // The semicolon above is important:
+    // The type of the `if` statement is `()`, despite being diverging.
```
ehuss: This is pretty subtle because you need to understand the rules for how an elided final expression works. I'm wondering if a little extra commentary could help here.


```suggestion
    // The semicolon above is important: The type of the `if` expression is
    // `()`, despite being diverging. When the final body expression is
    // elided, the type of the body is inferred to ! because the function body
    // diverges. Without the semicolon, the `if` would be the tail expression
    // with type `()`, which would fail to match the return type `!`.
```

src/items/functions.md
ehuss: Should this change be reverted, then?

src/divergence.md
```diff
@@ -0,0 +1,56 @@
+r[divergence]
+# Divergence
+
+r[divergence.intro]
+If an expression diverges, then nothing after that expression will execute. Importantly, while there are certain language constructs that immediately produce a _diverging expression_ of the type [`!`](./types/never.md), divergence can also propogate to the surrounding block --- where divergence indicates that the block itself will never finish executing.
```
ehuss: Is the following a correct definition of a diverging expression?

```
A *diverging expression* is an expression that never completes normal execution.
```

And then I would expect a rule below to explain which expressions are diverging. That can just be a list linking to the other relevant rules in their respective chapters. However, reading through this, I'm unable to come up a clear definition of what is diverging. Is it possible to come up with a list of what is diverging?

src/expressions/block-expr.md
jackh726: I like your wording, but yeah, this is a good point about `break`:

This fails because the type of the block is `usize`:
```
    fn block_type_diverging_break() -> ! {
        let x = 'block: {
            break 'block 0usize;
            loop {};
        };
        x
    }
```
The type of the block ends up being the LUB of all the break (or return for fns) types, and the final expression (which depends on whether the block is diverging, whether it is `()` or `!`).

src/expressions/if-expr.md
jackh726: This is accurate.

src/divergence.md
jackh726: That is imo correct, and close to what I originally had before I changed it based on [Niko's review](https://github.com/rust-lang/reference/pull/2067#discussion_r2534792433).

I did originally have a list, which TC asked me to remove: https://github.com/rust-lang/reference/pull/2067#discussion_r2475872395

jackh726: (though, that list is not entirely accurate, I've since realized, because a block expression can be diverging but not have a type of `!` - so, accurate as written, but not an "exhaustive" list of *diverging* expressions, just *diverging expressions producing `!`*)

src/items/functions.md
jackh726: Well, probably. But I wonder if there should be some additional text added somewhere that the type of the function *block* may not actually be the return type of the function, but rather must be *coercible into* the return type of the function.

jackh726:
> @theemathas We're discussing that above in [#2067 (comment)](https://github.com/rust-lang/reference/pull/2067#discussion_r2630177707).
> 
> EDIT: Or maybe I'll stand corrected, since that example depends on opaque types...

yeah, I mean, it's a different side of the same coin: this should be covered under the changes to `expr.block.type` and `expr.block.type.diverging`.

src/items/functions.md
ehuss: I believe that [`coerce.site.return`](https://doc.rust-lang.org/nightly/reference/type-coercions.html#r-coerce.site.return) specifies that the function body is coerced to the function's return type.

I can't find the subsequent part that the function body's type must match? the function's return type (after coercion). Type checking in general isn't documented (https://github.com/rust-lang/reference/issues/595), and I'm not sure which approach will make sense for that (for example, would there be a dedicated type checking chapter, or would the type checking rules be sprinkled throughout, or a mixture of both?).

I've removed this change in my local branch which I hope to push up soon. I've added this to my notes for future followup.

ehuss:
@jackh726 OK, sorry for the delay, I finally got around to doing a round of edits to clean things up and get it in the style of the Reference. Would you be able to give the changes a look over and make sure I didn't mess anything up? The commits should explain the reasoning for the changes.

The only thing on my todo list for later followup is to make sure the thing we discussed about labeled breaks in https://github.com/rust-lang/reference/pull/2067#discussion_r2637216415. There's already some rules about this, but I think there's some clarification that could be added. And I still don't 100% understand the `BreakableCtxt::may_break` stuff.

Besides the todo at the bottom of the `divergence.md` file, did you have a list of anything else that could be done as a followup? Is there more we should add to the divergence chapter, or does this pretty much cover it?

src/expressions/block-expr.md
```diff
@@ -44,7 +44,7 @@ r[expr.block.result]
 Then the final operand is executed, if given.
 
 r[expr.block.type]
-The type of a block is the type of the final operand, or `()` if the final operand is omitted.
+The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges][expr.block.diverging], in which case it is the [never type].
```
jackh726, line 46: This change isn't correct because of rust-lang#2067 (comment)

In that example, the final operand is loop {}, which has a type of !. But, the type of the entire block is usize. I think I maybe miscommunicated when I said "I like your wording" - because it seems nice, but has this flaw.

My comment at the end is essentially the rule that needs to be here:
The type of the block ends up being the LUB of all the break (or return for fns) types, and the final expression (which depends on whether the block is diverging, whether it is () or !).

jackh726:
Other than the one comment above, the changes look fine and correct to me. I don't think there is really anything related to divergence that's left out of this PR.

src/expressions/block-expr.md
ehuss: This was what I was getting at when I said it could be a followup, since I don't fully understand all the bits here. The problem is that blocks with break expressions are treated separately from normal blocks in the reference, and we need to fit this all together. I tried with the following:

~~~diff
diff --git a/src/expressions/loop-expr.md b/src/expressions/loop-expr.md
index 3d5b9420..bfe08687 100644
--- a/src/expressions/loop-expr.md
+++ b/src/expressions/loop-expr.md
@@ -311,6 +311,9 @@ Example:
 r[expr.loop.break.value]
 A `break` expression is only permitted in the body of a loop, and has one of the forms `break`, `break 'label` or ([see below](#break-and-loop-values)) `break EXPR` or `break 'label EXPR`.

+r[expr.loop.break-value.implicit-value]
+In the context of a [`loop` with break expressions][expr.loop.break-value] or a [labeled block expression], a `break` without an expression is considered identical to `break` with expression `()`.
+
 r[expr.loop.block-labels]
 ## Labeled block expressions

@@ -347,6 +350,24 @@ let result = 'block: {
 };
 ```

+r[expr.loop.block-labels.type]
+The type of a labeled block expression is the [least upper bound] of all of the break operands, and the final operand. If the final operand is omitted, the type of the final operand defaults to the [unit type], unless the block [diverges][expr.block.diverging], in which case it is the [never type].
+
+> [!EXAMPLE]
+> ```rust
+> fn example(condition: bool) {
+>     let s = String::from("owned");
+>
+>     let _: &str = 'block: {
+>         if condition {
+>             break 'block &s;  // &String coerced to &str via Deref
+>         }
+>         break 'block "literal";  // &'static str coerced to &str
+>     };
+> }
+> ```
+
+
 r[expr.loop.continue]
 ## `continue` expressions

@@ -394,15 +415,33 @@ let result = loop {
 assert_eq!(result, 13);
 ```

-r[expr.loop.break-value.loop]
-In the case a `loop` has an associated `break`, it is not considered diverging, and the `loop` must have a type compatible with each `break` expression.
-`break` without an expression is considered identical to `break` with expression `()`.
+r[expr.loop.break-value.type]
+A `loop` with an associated `break` does not [diverge], and its type is the [least upper bound] of all of the break operands.
+
+> [!EXAMPLE]
+> ```rust
+> fn example(condition: bool) {
+>     let s = String::from("owned");
+>
+>     let _: &str = loop {
+>         if condition {
+>             break &s; // &String coerced to &str via Deref
+>         }
+>         break "literal"; // &'static str coerced to &str
+>     };
+> }
+> ```

 [`!`]: type.never
 [`if` condition chains]: if-expr.md#chains-of-conditions
 [`if` expressions]: if-expr.md
 [`match` expression]: match-expr.md
 [boolean type]: ../types/boolean.md
+[diverge]: divergence
 [diverging]: divergence
+[labeled block expression]: expr.loop.block-labels
+[least upper bound]: coerce.least-upper-bound
+[never type]: type.never
 [scrutinee]: ../glossary.md#scrutinee
 [temporary values]: ../expressions.md#temporaries
+[unit type]: type.tuple.unit
diff --git a/src/type-coercions.md b/src/type-coercions.md
index 91b773ee..6bef99c3 100644
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -239,6 +239,8 @@ LUB coercion is used and only used in the following situations:
 + To find the common type for a series of if branches.
 + To find the common type for a series of match arms.
 + To find the common type for array elements.
++ To find the common type for a [labeled block expression] among the break operands and the final block operand.
++ To find the common type for an [`loop` expression with break expressions] among the break operands.
 + To find the type for the return type of a closure with multiple return statements.
 + To check the type for the return type of a function with multiple return statements.

@@ -325,5 +327,7 @@ precisely.
 [type cast operator]: expressions/operator-expr.md#type-cast-expressions
 [`Unsize`]: std::marker::Unsize
 [`CoerceUnsized`]: std::ops::CoerceUnsized
+[labeled block expression]: expr.loop.block-labels
+[`loop` expression with break expressions]: expr.loop.break-value
 [method-call expressions]: expressions/method-call-expr.md
 [supertraits]: items/traits.md#supertraits
~~~

Does that look correct?

One rule I was particularly uncertain about was `expr.loop.break-value.type`. The statement "`loop`…does not diverge"" seems too strong.  Does a `loop` with breaks that panics in all control flows, or break with diverging expressions, itself diverge? Does it require at least one control path to have a break without a diverging expression? If so, then it seems like "does not diverge" is too strongly worded. But I'm not sure how to word that. I'm wondering if it needs to be worded closer to how `expr.block.diverging` is done.

I'm also just guessing that the additions to the LUB list make sense.

jackh726: That diff looks good to me for the most part.

`expr.loop.break-value.type` is not correct though, for the reason you hinted at:

```rust
fn example() {
    let _: ! = loop {
        break loop {};
    };
}
```
This is the particular logic: https://github.com/rust-lang/rust/blob/ba2a7d33741a7ade4dc78e5e335d60d358cd1749/compiler/rustc_hir_typeck/src/expr.rs#L826

(`may_break` is only true if the break operand is not diverging)

jackh726: Wording here can be something like

>A `loop` does not diverge if it contains any `break` expressions with an operand that does not diverge, and its type is the [least upper bound] of all of the break operands.

ehuss:
@jackh726 while working on this, I realized it seems like this is missing a definition of what a "diverging expression" is for the rest of the expressions. It handles some of the more tricky expressions (`if`, `match`, `break`, etc.). But I think the rest of the expressions also have rules to determine how divergence propagates. I think we should be clear about that in all of the situations. The following summarizes my understanding:

Already handled in this PR:
- block
- labeled block
- loop (infinite and with breaks)
- `if`
- `match`
- `break`
- `continue`
- `return`

The following are generally straightforward where they diverge if any of their operands diverge:

- borrow
- derference
- `?`
- `!`
- arithmetic or logical expressions
- comparison expressions
- type cast expression
- assignment expression (either side)
- compound assignment expressions (either side)
- grouped expression
- array expression
- index expression (either operand)
- tuple expression
- tuple indexing expression
- struct expression
- call expression (function operand or any argument)
- method call expression (receiver or any argument)
- field access expression
- range expression
- await expression (the future operand)

But there are some that are more subtle:

- Lazy boolean expressions: Diverge only if left-hand-side expression diverges.
- let statement: only if initializer diverges.
- async block: never diverges
- literal expressions: never diverges
- closure expressions: never diverges
- path expression: If it refers to something with a type of `!`? I'm not sure if there are other scenarios?
- inline assembly:
    - `asm!` with the `noreturn` option: always diverges
    - `naked_asm!`: always diverges
    - `global_asm!`: always diverges. But is this observable? Does it mean anything?
- [`uninhabited-static`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#uninhabited-static) seems to imply there was a mistake for static extern. I assume this can be plugged at some point.

I also had a question about const-eval. My general understanding is that const eval is not allowed to diverge. Is that correct? I'm not sure how to express that or if my understanding is correct.

The following has me a little confused. `const` blocks can pass `cargo check`, but fail with `cargo build`.

```rust
fn f() -> ! {
    const { panic!() } // OK, seems const block has ! type
}
fn g() -> ! {
    const { panic!() }; 
    // ERROR, expected `!`, found `()`
}
```

`g` confuses me because if the const block is diverging, then it should pass `cargo check` (like a non-const block). Perhaps divergence for those doesn't really matter if divergence in const eval is never allowed?

Does all of that make sense? Do you have answers for the questions raised above?

traviscross:
In talking with @ehuss, we puzzled a bit over these contrasting examples:

This type checks:

```rust
fn f() -> ! {
    let _x = (const {
        panic!();
    },).0;
}
```

What we'e doing here, with the tuple, is reading from a place so as to invoke the rule that reading from a diverging place causes the block to diverge.  We're doing that to test whether the const block is considered to diverge.

Since that type checks, that would lead us to the conclusion that const block does diverge.

But then, this example does not type check:

```rust
fn f() -> ! { //~ ERROR: Mismatched types.
    let _x = (const {
        panic!()
    },).0;
}
```

Why might that be?

ehuss:
Also ping @lcnr in case you can help with the questions above.

jackh726:
The following summarizes my understanding:
> 
> Already handled in this PR:
> ...
> The following are generally straightforward where they diverge if any of their operands diverge:
> But there are some that are more subtle:

This is a good shout-out, I didn't quite think about these. I think I focused solely on expressions that both diverge *and* have a type of `!`. Of course, `if` and `match` are also special so it was obvious to include those. Many of the remaining expressions (I think) fall out from the `diverging.intro` statement. Of course, explicitly stating these may still be useful. (I can also go through and respond to each one-by-one another time.)

> I also had a question about const-eval. My general understanding is that const eval is not allowed to diverge. Is that correct? I'm not sure how to express that or if my understanding is correct.

I'm not entirely sure, but it's not immediately clear to me that this is the case. Importantly, there is *divergence as it pertains to type checking* and const eval itself.

> The following has me a little confused. `const` blocks can pass `cargo check`, but fail with `cargo build`.
> 
> ```rust
> fn f() -> ! {
>     const { panic!() } // OK, seems const block has ! type
> }
> fn g() -> ! {
>     const { panic!() }; 
>     // ERROR, expected `!`, found `()`
> }
> ```
> 
> `g` confuses me because if the const block is diverging, then it should pass `cargo check` (like a non-const block). Perhaps divergence for those doesn't really matter if divergence in const eval is never allowed?


This has a pretty simple explanation: the former results in an expected type of `!` for the const block so the type of the const block is `!` - the latter has no expectation because of the semicolon, so the type of the block `!` fallbacks to `()`.

jackh726:
> In talking with @ehuss, we puzzled a bit over these contrasting examples:
> 
> This type checks:
> 
> ```rust
> fn f() -> ! {
>     let _x = (const {
>         panic!();
>     },).0;
> }
> ```
> 
> What we'e doing here, with the tuple, is reading from a place so as to invoke the rule that reading from a diverging place causes the block to diverge. We're doing that to test whether the const block is considered to diverge.
> 
> Since that type checks, that would lead us to the conclusion that const block does diverge.
> 
> But then, this example does not type check:
> 
> ```rust
> fn f() -> ! { //~ ERROR: Mismatched types.
>     let _x = (const {
>         panic!()
>     },).0;
> }
> ```
> 
> Why might that be?

~both of these examples are the same and do not type check~
*I misread, I need to look more at this*

---

What I'm seeing is actually that `const` blocks (along with `async` blocks) is that they do not *propagate* their divergence because they essentially act as independent function bodies. Of course, divergence can exist inside:

```rust
fn f() {
    const {
        panic!();
        let _x = 0;
    }
}
```

jackh726:
> ~both of these examples are the same and do not type check~ _I misread, I need to look more at this_

Not exactly sure here - my best guess is that const eval is just masking typecheck here: we need to figure out the type of the `const` block, but can never do that because const eval doesn't complete.

ehuss:
@jackh726 I pushed a few more small edits. What do you think of merging this now as-is, and then making some follow-ups to try to address the questions in https://github.com/rust-lang/reference/pull/2067#issuecomment-3782115579?

I was experimenting with an approach in https://github.com/ehuss/reference/commit/93fd93eeac640ddec7ddb5b0d4c785ce7e5aa286. Maybe that is the wrong way to express the rules. It might be a starting point to discuss how to approach it.

jackh726:
The edits look good to me (I think, not entirely clear to me what's new and what's old but rebased).

I think landing as-is and making follow-ups is a good approach. I don't think anything in that branch contradicts the information in the current PR in any way.

Your branch looks overall like a decent start, I think would be good to get to opening a PR and I can comment on specific things that may be inaccurate or imprecise.

ehuss (approved):
Thanks, I really appreciate it! I'm going to go ahead and merge. I've opened a few issues to track followups.

PR #2068, opened 2025-10-29
const_eval.md: add missing word

tshepang:


traviscross (approved):

src/const_eval.md
```diff
@@ -14,11 +14,13 @@ Certain forms of expressions, called constant expressions, can be evaluated at
...
-In [const contexts](#const-context), these are the only allowed
-expressions, and are always evaluated at compile time.
+Expressions in a [const context] must be constant expressions.
+
+r[const-eval.const-expr.evaluation]
+Expressions in const contexts are always evaluated at compile time.
 
 r[const-eval.const-expr.runtime-context]
-In other places, such as [let statements], constant expressions *may* be, but are not guaranteed to be, evaluated at compile time.
+Outside of const contexts, constant expressions *may* be, but are not guaranteed to be, evaluated at compile time.
```
RalfJung, line 23: I find this entire paragraph very strange, since it describes something that cannot even be observed except by inspecting the generated assembly (which obviously we make no guarantees about). Making this a *normative* paragraph sounds like it would actually do something, but the entire paragraph is a NOP.

PR #2073, opened 2025-11-03
Note scoping differences between `*x` and `*x.deref()` , and `a[b]` and `*a.index(b)`

dianne:
The equivalences between the overloaded operator expressions and their associated function call forms don't hold in the presence of temporary lifetime extension; this adjusts the rules accordingly and adds examples.

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -84,7 +84,22 @@ r[expr.array.index.array]
 [Array] and [slice]-typed values can be indexed by writing a square-bracket-enclosed expression of type `usize` (the index) after them. When the array is mutable, the resulting [memory location] can be assigned to.
 
 r[expr.array.index.trait]
-For other types an index expression `a[b]` is equivalent to `*std::ops::Index::index(&a, b)`, or `*std::ops::IndexMut::index_mut(&mut a, b)` in a mutable place expression context. Just as with methods, Rust will also insert dereference operations on `a` repeatedly to find an implementation.
+For other types an index expression `a[b]` is equivalent to `*std::ops::Index::index(&a, b)`, or `*std::ops::IndexMut::index_mut(&mut a, b)` in a mutable place expression context, except that when the index expression undergoes [temporary lifetime extension], the indexed expression `a` also has its [temporary scope] extended. Just as with methods, Rust will also insert dereference operations on `a` repeatedly to find an implementation.
+
+```rust
+// The temporary holding the result of `vec![()]` is extended to
+// live to the end of the block, so `x` may be used in subsequent
+// statements.
+let x = &vec![()][0];
+# x;
+```
+
+```rust,compile_fail,E0716
+// The temporary holding the result of `vec![()]` is dropped at the
+// end of the statement, so it's an error to use `y` after.
+let y = &*std::ops::Index::index(&vec![()], 0); // ERROR
+# y;
+```
 
 r[expr.array.index.zero-index]
 Indices are zero-based for arrays and slices.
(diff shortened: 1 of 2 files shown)


dianne:
The sections on field access expressions and tuple indexing expressions may also benefit from similar examples or links to that example, in case it needs clarifying that the way they autoderef is scoped more like repeatedly applying `*` than repeatedly applying a deref method. The autoderef section on the field access expressions page is in line with that, but doesn't clarify. The tuple indexing expressions section doesn't account for autoderef at all, so it would need to be reworked first.

dianne:
@rustbot author

Looking at this again, it should probably be framed in terms of static semantics rather than runtime semantics, since one of the examples makes an unusable binding. I'll split that part off into a `compile_fail` test and remove the assertions.

While I'm at it, the wording of the `*` operator section should probably account for `Box`. It's covered in [lang.types.box.deref](https://doc.rust-lang.org/nightly/reference/special-types-and-traits.html#r-lang-types.box.deref); I'll add a reference to that.

dianne:
@rustbot ready

I'll open a separate PR to add text for `Box` to the `*` operator expression documentation.

src/expressions/array-expr.md
```diff
@@ -90,6 +90,30 @@ r[expr.array.index.trait]
 For other types an index expression `a[b]` is equivalent to `*std::ops::Index::index(&a, b)`, or `*std::ops::IndexMut::index_mut(&mut a, b)` in a mutable place expression context.
```
ehuss: When looking at this, we felt that this probably shouldn't be part of a note, but instead should be incorporated into the rule itself. The "is equivalent to" is not entirely true due to this exception. How about just adding this text here:

```suggestion
For other types an index expression `a[b]` is equivalent to `*std::ops::Index::index(&a, b)`, or `*std::ops::IndexMut::index_mut(&mut a, b)` in a mutable place expression context, except that when the index expression undergoes [temporary lifetime extension], the indexed expression `a` also has its temporary lifetime extended.
```

and then include the rust example as you have.

dianne: Seems good to me! I couldn't figure out a way to keep the link to [destructors.scope.lifetime-extension.sub-expressions](https://doc.rust-lang.org/reference/destructors.html#r-destructors.scope.lifetime-extension.sub-expressions) without it getting overly cluttered, but maybe it's fine to leave it out?

src/expressions/array-expr.md
```diff
@@ -84,7 +84,22 @@ r[expr.array.index.array]
...
+// statements.
+let x = &vec![()][0];
+# x;
+```
+
+```rust,compile_fail,E0716
+// The temporary holding the result of `vec![()]` is dropped at the
+// end of the statement, so it's an error to use `y` after.
+let y = &*std::ops::Index::index(&vec![()], 0); // ERROR
+# y;
```
ehuss, line 101: I'm curious about the thinking for hiding the "use" here? My thinking is that with it hidden it could be confusing since assigning to `y` isn't an error, it's not until the later use. For example, it could be:


```suggestion
let y = &*std::ops::Index::index(&vec![()], 0);
y;  // ERROR
```

And to contrast in the other example, it could be:

```rust
let x = &vec![()][0];
x; // OK
```

Is it maybe that a bare identifier statement is unusual?

dianne: I was just going for stylistic consistency with the examples in https://doc.rust-lang.org/nightly/reference/destructors.html#r-destructors.scope.lifetime-extension; they also hide usages and put the ok/error comments a line early, relying on the text to clarify. There's a lot more examples there, so there may be more of an argument to made for brevity. I wouldn't be opposed to making the usages explicit here (or even there too) for clarity.

ehuss: Ok, I'm not going to block on that. If you (or someone else) see an improvement that could be made to all those examples, we can tackle that in a separate PR.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2073/7446bf9697c95d155eef33c6a9d91fbd29a5e359..8e4d81ab52388fedb50151268ed85080ea0facda)*

PR #2075, opened 2025-11-04
Add special cases and links relating to dereferencing boxes

dianne:
Three things:
- I've added some text to `expr.deref` reenforcing that the `*` operator's behavior on boxes is built into the language.
- I've added a rule under `expr.deref` specifying that `*` may move out of boxes (linking to `lang-types.box` and `expr.move.movable-place` where this is also specified).
- To get the places where that rule is printed interconnected, I've added a link from `lang-types.box.deref` to `expr.move.movable-place` and redirected the `Box<T>` link on `expr.move.movable-place` to `lang-types.box` (which has a link to its old referent in the `alloc` docs). `lang-types.box.deref` and `expr.move.movable-place` already both linked to `expr.deref`.


traviscross (approved):

traviscross:
We talked about this in the lang-docs call.  Looks good.  Thanks @dianne.

PR #2077, opened 2025-11-08
Reflect explicit macro import in reference

Voultapher:
Companion PR for https://github.com/rust-lang/rust/pull/139493

### DIFF
--- a/src/names/preludes.md
+++ b/src/names/preludes.md
@@ -40,6 +40,11 @@ Edition | `no_std` not applied        | `no_std` applied
 >
 > [`core::prelude::rust_2015`] and [`core::prelude::rust_2018`] have the same contents as [`core::prelude::v1`].
 
+> [!NOTE]
+> When one of [`core::panic!`] or [`std::panic!`] is brought into scope due to the [standard library prelude], and a user-written [glob import] brings the other into scope, `rustc` currently allows use of `panic!`, even though it is ambiguous. The user-written glob import takes precedence to resolve this ambiguity.
+>
+> For details, see [names.resolution.expansion.imports.ambiguity.panic-hack].
+
 r[names.preludes.extern]
 ## Extern prelude
 
@@ -203,6 +208,7 @@ r[names.preludes.no_implicit_prelude.edition2018]
 [Built-in attributes]: ../attributes.md#built-in-attributes-index
 [extern prelude]: #extern-prelude
 [floating-point types]: ../types/numeric.md#floating-point-types
+[glob import]: items.use.glob
 [Integer types]: ../types/numeric.md#integer-types
 [Language prelude]: #language-prelude
 [Machine-dependent integer types]: ../types/numeric.md#machine-dependent-integer-types
--- a/src/names/preludes.md
+++ b/src/names/preludes.md
@@ -192,6 +192,41 @@ The `no_implicit_prelude` attribute may be used any number of times on a form.
 r[names.preludes.no_implicit_prelude.excluded-preludes]
 The `no_implicit_prelude` attribute prevents the [standard library prelude], [extern prelude], [`macro_use` prelude], and the [tool prelude] from being brought into scope for the module and its descendants.
 
+r[names.preludes.no_implicit_prelude.implicitly-imported-macros]
+> [!NOTE]
+> Despite `#![no_implicit_prelude]`, `rustc` currently brings certain macros implicitly into scope. Those macros are:
+>
+> - [`assert!`]
+> - [`cfg!`]
+> - [`cfg_select!`]
+> - [`column!`]
+> - [`compile_error!`]
+> - [`concat!`]
+> - [`concat_bytes!`]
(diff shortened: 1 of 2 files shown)


src/names/preludes.md
```diff
@@ -80,11 +80,7 @@ r[names.preludes.extern.no_std]
 ### The `no_std` attribute
 
 r[names.preludes.extern.no_std.intro]
-By default, the standard library is automatically included in the crate root
-module. The [`std`] crate is added to the root, along with an implicit
-[`macro_use` attribute] pulling in all macros exported from `std` into the
-[`macro_use` prelude]. Both [`core`] and [`std`] are added to the [extern
-prelude].
+By default, the standard library is automatically included in the crate root module. The [`std`] crate is added to the root. Both [`core`] and [`std`] are added to the [extern prelude]. Standard library macros are imported via [use declarations] and are *not* part of the [`macro_use` prelude].
```
traviscross: > The [`std`] crate is added to the root.

Is this sentence needed?  I.e., is there any sense in which "the `std` crate is added to the root" but the `core` crate is not, or that this is not covered by "Both [`core`] and [`std`] are added to the [extern prelude]"?

My reading of the old text is that the sentence was there to support the restrictive clause (that erroneously was preceded by a comma, making the clause seem non-restrictive, i.e. that the part before the comma could stand on its own without the part after it):

> The [`std`] crate is added to the root, along with an implicit [`macro_use` attribute] pulling in all macros exported from `std` into the [`macro_use` prelude].

traviscross: Or, maybe this language and the line below

> It does three things... Injects the [`core`] crate into the crate root instead of [`std`].

are leftovers from Rust 2015.

In Rust 2018 and later, this is a valid library:;

```rust
use ::core as c1;
use ::std as c2;
```

But it's not in Rust 2015.  It gives the error

```
error[E0432]: unresolved import `core`
 --> src/lib.rs:1:5
  |
1 | use ::core as c1;
  |     ^^^^^^^^^^^^ no `core` in the root
```

unless `#![no_std]` is used, in which case it accepts that line but not `use ::std`.  Interestingly, I don't see this documented specifically in the 2018 edition guide:

https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html

Perhaps this is simply an artifact of the change to make crates in the extern prelude accessible to `use` declarations.  Did that then make it fully unobservable whether `core` is separately added to the crate root or not, or is there still some way to tell?

cc @ehuss

traviscross: Looking, this is all covered in:

- https://github.com/rust-lang/reference/pull/1913

I'm going to go get that merged.

Voultapher: Ok, I'll wait for #1913 to be merged and will update this PR on that.

traviscross: Thanks.  Now merged.

Voultapher:
I've updated the PR, it's not clear to me what changes are still desired please re-review.

ehuss:
Can you rebase on latest master now that https://github.com/rust-lang/reference/pull/1913 has been merged?

I would suggest the following edits:

```diff
diff --git a/src/names/preludes.md b/src/names/preludes.md
index 3a2c3ff1..b25b4b8a 100644
--- a/src/names/preludes.md
+++ b/src/names/preludes.md
@@ -81,7 +81,7 @@ r[names.preludes.extern.no_std]
 ### The `no_std` attribute

 r[names.preludes.extern.no_std.intro]
-The *`no_std` [attribute][attributes]* causes the [`std`] crate to not be linked automatically, the [standard library prelude] to instead use the `core` prelude, and the [`macro_use` prelude] to instead use the macros exported from the `core` crate.
+The *`no_std` [attribute][attributes]* causes the [`std`] crate to not be linked automatically and the [standard library prelude] to instead use the `core` prelude.

 > [!EXAMPLE]
 > <!-- ignore: test infrastructure can't handle no_std -->
@@ -110,9 +110,6 @@ The `no_std` attribute may be used any number of times on a form.
 r[names.preludes.extern.no_std.module]
 The `no_std` attribute changes the [standard library prelude] to use the `core` prelude instead of the `std` prelude.

-r[names.preludes.extern.no_std.macro_use]
-By default, all macros exported from the `std` crate are added to the [`macro_use` prelude]. If the `no_std` attribute is specified, then all macros exported from the `core` crate are placed into the [`macro_use` prelude] instead.
-
 r[names.preludes.extern.no_std.edition2018]
 > [!EDITION-2018]
 > Before the 2018 edition, `std` is injected into the crate root by default. If `no_std` is specified, `core` is injected instead. Starting with the 2018 edition, regardless of `no_std` being specified, neither is injected into the crate root.
```

Voultapher:
My bad, I had forgotten to track upstream. First PR in this repo.

Voultapher:
I've applied your feedback and added a section to name-resolution.md

traviscross (approved):

PR #2080, opened 2025-11-09
get rid of const.no-mut-refs

### LINKED ISSUE #2074: Note in no-mutable-refs in consts is confusing
> This note is a bit confusing since it does not relate back to what we said above about why these references are disallowed.
> 
> 
> 
> It should probably also say that this is sound to allow because it take an unsafe block to create a mutable reference to such a static, and that unsafe block carries the obligation of ensuring that no unsound aliasing will occur, Furthermore, the problem regarding "is there a new instance of the pointee for every use of the const or not" does not come up since unambiguously, these references all point to `S`. 

 _Originally posted by @RalfJung in [#2058](https://github.com/rust-lang/reference/pull/2058/files/53f4392b0287ea2348f8f48b57ea2a8adf44f7eb#r2484940069)_

RalfJung:
The reference PR that goes with https://github.com/rust-lang/rust/pull/148746.

Fixes https://github.com/rust-lang/reference/issues/2074.


RalfJung:
FCP passed, https://github.com/rust-lang/rust/pull/148746 is in the bors queue now.

traviscross (approved):

traviscross:
I had a WIP commit to clean up the language we're about to delete.  If the commit looks OK to you, I'd like to put this in the history for posterity's stake.  I've stacked your commit onto it and also deleted the link reference definitions that are now unused.

RalfJung:
LGTM, thanks!

PR #2083, opened 2025-11-10
const_eval.md: be more clear where link leads to

tshepang:
Having just "Struct" as the link makes it look like the target is about structs and not just struct expressions


ehuss (approved):
Thanks!

PR #2084, opened 2025-11-10
use "lint" as placeholder instead of "C"

tshepang:
Seeing C make brain think of the language


traviscross:
Thanks for the PR.  We're going to close this in favor of https://github.com/rust-lang/reference/pull/1890 (with which this PR would conflict); we'll adjust that PR to not use a placeholder at all.

PR #2071, opened 2025-10-31
identifiers: bump Unicode from 16 to 17

Marcondiro:
This is a companion PR to rust-lang/rust#148321

Thanks!

PR #2086, opened 2025-11-12
Update Unicode version and links to 17.0

clubby789:
As specified by the change in https://github.com/rust-lang/rust/pull/148831.


ehuss:
Thanks! Though there is already a PR at https://github.com/rust-lang/reference/pull/2071 for this, and is waiting on FCP in https://github.com/rust-lang/rust/pull/148321.

clubby789:
Ah, got it. I'll back those changes out of the bump PR

PR #2091, opened 2025-11-16
UB: update the extra clause for provenance UB during const evaluation

RalfJung:
The old note didn't make it clear that the transmute is also illegal when it occurs nested inside a field. We already have the framework of "valid values" for this, so let's just use that also for this extra restriction.

Furthermore, there's another way to cause UB with provenance during const evaluation: by having a pointer whose bytes are mixed up.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -223,6 +223,22 @@ r[undefined.validity.const-provenance]
 
   This implies that transmuting or otherwise reinterpreting a pointer (reference, raw pointer, or function pointer) into a non-pointer type (such as integers) is undefined behavior if the pointer had provenance.
 
+  > [!EXAMPLE]
+  > All of the following are UB:
+  > ```rust,compile_fail
+  > const REINTERPRET_PTR_AS_INT: usize = {
+  >     let ptr = &0;
+  >     unsafe { (&raw const ptr as *const usize).read() }
+  > };
+  >
+  > const PTR_BYTES_IN_WRONG_ORDER: &i32 = {
+  >     let mut ptr = &0;
+  >     let ptr_bytes = &raw mut ptr as *mut std::mem::MaybeUninit::<u8>;
+  >     unsafe { std::ptr::swap(ptr_bytes.add(1), ptr_bytes.add(2)) };
+  >     ptr
+  > };
+  > ```
+
 r[undefined.validity.undef]
 **Note:** Uninitialized memory is also implicitly invalid for any type that has
 a restricted set of valid values. In other words, the only cases in which
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -221,6 +221,7 @@ r[undefined.validity.const-provenance]
 
   > [!EXAMPLE]
   > All of the following are UB:
+  >
   > ```rust,compile_fail
   > # use core::mem::MaybeUninit;
   > # use core::ptr;
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -76,13 +76,6 @@ r[undefined.asm]
 * Incorrect use of inline assembly. For more details, refer to the [rules] to
   follow when writing code that uses inline assembly.
 
-r[undefined.const-transmute-ptr2int]
-* **In [const context](const_eval.md#const-context)**: transmuting or otherwise
-  reinterpreting a pointer (reference, raw pointer, or function pointer) into
-  some allocation as a non-pointer type (such as integers).
-  'Reinterpreting' refers to loading the pointer value at integer type without a
-  cast, e.g. by doing raw pointer casts or using a union.
-
 r[undefined.runtime]
 * Violating assumptions of the Rust runtime. Most assumptions of the Rust runtime are currently not explicitly documented.
   * For assumptions specifically related to unwinding, see the [panic documentation][unwinding-ffi].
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -221,6 +214,15 @@ r[undefined.validity.valid-range]
...
   > `rustc` achieves this with the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
+r[undefined.validity.const-provenance]
+* **In [const context](const_eval.md#const-context)**: In addition to what is described above,
+  further provenance-related requirements apply during const evaluation.
+  Any value that holds pure integer data (the `i*`/`u*`/`f*` types as well as `bool` and `char`, enum discriminants, and slice metadata) must not carry any provenance.
+  Any value that holds pointer data (references, raw pointers, function pointers, and `dyn Trait` metadata) must either carry no provenance,
+  or all bytes must be fragments of the same original pointer value in the correct order.
+
+  This implies that transmuting or otherwise reinterpreting a pointer (reference, raw pointer, or function pointer) into a non-pointer type (such as integers) is Undefined Behavior of the pointer had provenance.
```
theemathas:
```suggestion
  This implies that transmuting or otherwise reinterpreting a pointer (reference, raw pointer, or function pointer) into a non-pointer type (such as integers) is Undefined Behavior if the pointer had provenance.
```

theemathas:
This text currently does not cover the validity of the following in consteval:
* `str` metadata (presumably the same as slice metadata)
* closures, async blocks, and async closures (I have no idea how this works. I'm currently testing this.)

RalfJung:
> str metadata (presumably the same as slice metadata)

This is already defined to be the same as `&[u8]`.

> closures, async blocks, and async closures (I have no idea how this works. I'm currently testing this.)

These are just structs, from an opsem perspective. We don't currently list them at all, even outside the const-eval clause.

theemathas:
My understanding is that async blocks behave more like unions? I'm not sure.

theemathas:
The current wording also doesn't prohibit mixing pointer fragments with different offsets, like in https://github.com/rust-lang/rust/issues/146291

RalfJung:
> The current wording also doesn't prohibit mixing pointer fragments with different offsets, like in [rust-lang/rust#146291](https://github.com/rust-lang/rust/issues/146291)

The wording is "all bytes must be fragments of the same original pointer value in the correct order". "same pointer value" means same address/offset and same provenance.

RalfJung:
> My understanding is that async blocks behave more like unions? I'm not sure.

More like a weird kind of enum. Anyway, they are container types that have fields, and we recurse into them like we recurse into structs and enums.

If you think we should be more explicit about closures and async lowering, please file an issue; that is entirely orthogonal to the const-eval clause.

theemathas:
Turns out that async blocks in const is [gated behind an unstable feature](https://github.com/rust-lang/rust/issues/85368), so that's a non-issue for the purposes of the reference.

src/behavior-considered-undefined.md
```diff
@@ -221,6 +214,15 @@ r[undefined.validity.valid-range]
...
   > `rustc` achieves this with the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
+r[undefined.validity.const-provenance]
+* **In [const context](const_eval.md#const-context)**: In addition to what is described above,
+  further provenance-related requirements apply during const evaluation.
+  Any value that holds pure integer data (the `i*`/`u*`/`f*` types as well as `bool` and `char`, enum discriminants, and slice metadata) must not carry any provenance.
+  Any value that holds pointer data (references, raw pointers, function pointers, and `dyn Trait` metadata) must either carry no provenance,
+  or all bytes must be fragments of the same original pointer value in the correct order.
+
+  This implies that transmuting or otherwise reinterpreting a pointer (reference, raw pointer, or function pointer) into a non-pointer type (such as integers) is Undefined Behavior if the pointer had provenance.
```
traviscross:
```suggestion
  This implies that transmuting or otherwise reinterpreting a pointer (reference, raw pointer, or function pointer) into a non-pointer type (such as integers) is undefined behavior if the pointer had provenance.
```

traviscross: It'd be good to add an example or two for this section (similar to the examples we looked at on the lang side that motivated the recent changes).  That would also help make clear what we mean by a pointer  fragment, apropos of the question @ehuss is about to ask.

src/behavior-considered-undefined.md
```diff
@@ -221,6 +214,15 @@ r[undefined.validity.valid-range]
   > [!NOTE]
   > `rustc` achieves this with the unstable `rustc_layout_scalar_valid_range_*` attributes.
 
+r[undefined.validity.const-provenance]
+* **In [const context](const_eval.md#const-context)**: In addition to what is described above,
+  further provenance-related requirements apply during const evaluation.
+  Any value that holds pure integer data (the `i*`/`u*`/`f*` types as well as `bool` and `char`, enum discriminants, and slice metadata) must not carry any provenance.
+  Any value that holds pointer data (references, raw pointers, function pointers, and `dyn Trait` metadata) must either carry no provenance,
+  or all bytes must be fragments of the same original pointer value in the correct order.
```
ehuss: It doesn't seem like "pointer fragments" is defined anywhere. Would it be possible to come up with a definition somewhere?

RalfJung: That's tricky to define in isolation.  Maybe it's better to avoid the term and just say all bytes must "come from" the same original pointer value?

The longer answer is that during const-eval, a byte doesn't look quite like what we have defined in the reference currently. It's more like
```rust
enum Byte {
  Uninit,
  /// Initialized byte without provenance.
  Init(u8),
  /// Pointer fragment. `idx` is in the range `0..ptr_size`.
  PointerFragment { ptr: Pointer, idx: u8 },
}
```

traviscross (approved):

PR #2093, opened 2025-11-21
Specify that range patterns must be nonempty.

theemathas:
This changes the reference to match the behavior of the compiler, based on my empirical testing.

See also https://github.com/rust-lang/rust/issues/149165

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -554,10 +554,12 @@ It is written as `..=` followed by the upper bound.
 
 For example, `..=10` will match any integer less than or equal to 10, such as 10, 1, 0, and for signed integer types, all negative values.
 
-r[patterns.range.constraint-less-than]
-The lower bound cannot be greater than the upper bound.
-That is, in `a..=b`, a &le; b must be the case.
-For example, it is an error to have a range pattern `10..=0`.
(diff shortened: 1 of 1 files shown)


src/patterns.md
```diff
@@ -554,10 +554,12 @@ It is written as `..=` followed by the upper bound.
 
 For example, `..=10` will match any integer less than or equal to 10, such as 10, 1, 0, and for signed integer types, all negative values.
 
-r[patterns.range.constraint-less-than]
-The lower bound cannot be greater than the upper bound.
-That is, in `a..=b`, a &le; b must be the case.
-For example, it is an error to have a range pattern `10..=0`.
+r[patterns.range.constraint-nonempty]
+A range pattern must match at least one possible value. In other words:
```
ehuss: The wording here seemed little strange to me. Does this make sense? Obviously range patterns that are outside the range of the scrutinee won't match the value at all.

```suggestion
A range pattern must be able to match at least one possible value. In other words:
```

theemathas: I wrote it this way to be consistent with the other lines in this section, for example:

> The exclusive range pattern matches all values from the lower bound up to, but not including the upper bound.

traviscross (approved):

Issue #2043, opened 2025-10-08
Guidelines for referring to tools and compilers


ehuss:
I think the authoring guide should have some guidelines for when, why, and how to refer to tools and implementations. In particular, things like `rustc`, `clippy`, `cargo`, `rustfmt`, `miri`, etc. The Reference occasionally refers to these things when it seems like that information would provide very helpful context, but there's a certain balance of doing that too little or too much. What are the guidelines for that balance?

There is a slightly different issue in https://github.com/rust-lang/reference/issues/2040 for implementation-specific guidelines.

Some examples from the Reference:
- `rustc` currently only allows the `clippy` and `rustfmt` tool attributes. It has not yet been decided if the tool attribute space should be extensible (and if so, how).
- The `feature` attribute links to the Unstable book.
- The description of user extensible `cfg` values mentions the way to do that is via `rustc --cfg`.
- The `test` cfg is set via `rustc --test`.
- Identifiers that start with an underscore are a convention to silence an unused warning in `rustc`.
- The dynamic and static C-runtime mentions `rustc -C target-feature=+crt-static`
- A note about the hazards of using `#[inline]` and the relationship to how `rustc` works.
- How target-features and target-cpu's are related via `rustc` options.
- Note about how you can find the available lints in the lint attribute docs.
- Note about what `#[deprecated]` does in `rustc`.
- Defaults for `recursion_limit` and `type_length_limit`.
- Mentions of certain hazards, like overflowing literals will generate a warning by `rustc`.
- Lots of documentation related to `#[link]` needs to cross-link with the `rustc` docs.
- The behavior of unwinding across nounwind abi uses an illegal instruction in `rustc`.
- An explanation of the requirement to use `extern alloc;` due to the way things work.
- How to write a proc-macro with Cargo.
- Note about security concerns of proc-macros.
- Source files discuss filesystems specifically, but in theory there could be other ways that source is loaded. How `rustc`-specific is that?

Issue #2042, opened 2025-10-08
Guidelines for documenting the *why* of something


ehuss:
I think the authoring guide should have some guidelines for mentioning *why* something is the way it is.

In the Reference, we have generally avoided that, but I think that is a detriment to some readers, since just specifying the behavior of something can be extremely hard to understand why it matters, or how it is relevant to a Rust programmer.
That kind of information can provide useful and interesting context. However, documenting the motivation for every aspect of the language could easily overwhelm the content and be distracting.

Graydon spoke highly of the [Ada rationale](http://www.ada-auth.org/standards/12rat/html/Rat12-TOC.html) (the 1979 version specifically), using it while working on Rust.

A very minor example: The `type_length_limit` attributes explains what it does, but not why it is there. I think it could be useful to have a note like: `type_length_limit` is used to prevent the compiler from hanging and to better deal with [polymorphic recursion](https://en.wikipedia.org/wiki/Polymorphic_recursion).

Regardless, I think it would be good to have a policy, as contributors sometimes include this information that we then need to tell them to remove. Guidelines could bring clarity on what we expect.

See also this comment: https://github.com/rust-lang/spec/issues/27#issuecomment-1974734900

PR #2100, opened 2025-11-29
Rework the introduction and add a scope chapter

### LINKED ISSUE #60: Should we document when features are stabilized?
The standard library documentation documents what version that various types, functions, and what not are stabilized. Since there's no "API docs" for features, other than this reference, should we be documenting when language features are stabilized into the reference?

### LINKED ISSUE #550: Clarify the audience and goals of the reference.
I think it would be useful to be explicit about who the reference is for and in what style it should be. There seems to be some uncertainty about being a [reference](https://en.wikipedia.org/wiki/Programming_language_reference) vs a [spec](https://en.wikipedia.org/wiki/Programming_language_specification) or if it should be both. I was confused when I first started, but I've gleaned the following along the way:

- [introduction.md](https://github.com/rust-lang-nursery/reference/blob/master/src/introduction.md) says it is "not normative", "not a spec", but a "spec is intended in the future".

- [CONTRIBUTING.md](https://github.com/rust-lang-nursery/reference/blob/master/CONTRIBUTING.md): "too teachy for people" and "too shallow for a reference".

- Over chat (some disagreement):
    - @Centril: "supposed to be near-spec-like"
    - @steveklabnik: "the reference is supposed to be spec-like, but not fully spec yet. the end goal is to have it be an actual spec. not there yet"
    - @Havvy: "The reference fills the niche for somebody who's already gone through learning material and needs a refresher or a point of confusion cleared up. Specifications fill the niche of describing the exact semantics of whatever they are specifying. The audiences for these are not the same."

- Rust All Hands 2019: https://paper.dropbox.com/doc/Topic-Spec--AZn5mRiLBp9zayEPAqu4RkjnAg-BB1tcOKwvpWUdzgQO4zgu
    - Was there any outcome from the Rust All Hands discussion?

I think it would be good to write down:
- Who is the audience, and what are the high-level goals?
- What is the intent? Should it be complete enough for someone to write a Rust compiler, parser, or other tooling? A guide for learning details of the language?
- Maybe even dive down into details like style and voice? Is there an established format to follow?
    - This is one place where I see the divide between a formal spec and a natural reference is difficult to bridge. For example, a spec could have every clause numbered, possibly written in formal semantics, perhaps with a test suite for every clause. The difference between natural language and a formal definition can be quite big. Some language specs include both formal and natural language.

I like the general style things have gone so far, in that it is mostly readable for a typical Rust programmer, with illustrative examples. My personal preference would be: continue along the current trajectory of having a non-normative, semi-formal (i.e. not a guide, not a formal spec)  reference which strives for complete coverage while remaining semi-neutral to the details of the reference implementation (i.e. doesn't usually delve into usage, bugs, implementation-specific things, etc.). A formal, standards-quality spec is a non-goal at this time, though it is desirable to have such a document in the future (possibly separate or combined?).

Some example specs and references:
- C++: [Spec](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf) [(source)](https://github.com/cplusplus/draft)
- [C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/) (same as [ECMA-334](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-334.pdf))
- [Go Language Specification](https://golang.org/ref/spec)
- [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se11/html/index.html)
- Javascript: [ECMA-262](https://www.ecma-international.org/ecma-262/9.0/index.html)
- [Typescript Language Specification](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md)
- [Scheme](http://www.r6rs.org/)
- [Standard ML](http://sml-family.org/sml97-defn.pdf)
- [Swift Language Reference](https://docs.swift.org/swift-book/ReferenceManual/AboutTheLanguageReference.html)
- [Ada Reference Manual](http://www.ada-auth.org/standards/12rm/html/RM-TOC.html)
- [Haskell language report](https://www.haskell.org/onlinereport/haskell2010/haskell.html#haskellpa1.html)
- [WASM Spec](https://webassembly.github.io/spec/core/index.html)
- [Python Reference](https://docs.python.org/3/reference/index.html)

Interesting little paper comparing different ways languages are specified: http://www.knosof.co.uk/vulnerabilities/langconform.pdf

### LINKED ISSUE #2024: Policy: Past behavior
I'd like the authoring guide to have guidance on whether or not to include information about past versions of Rust. If we do allow it, what is the guidance for how to do that?

I would prefer to generally avoid references to older versions, but I acknowledge that there are some rare cases where it is truly helpful. Is there some nuance or balance for that?

https://github.com/rust-lang/spec/issues/34#issuecomment-1973739688 mentions that [rfc3355] says:

> The specification is specific to the latest version of Rust. [...]

> While the specification might include notes about the Rust version that a feature was introduced in for informative purposes (similar to standard library documentation), it does not attempt to accurately specify older, unsupported versions of Rust.

The [Vision] blog post says:

> The descriptive definitions tell the reader how a construct will behave with
respect to a specific Rust version (e.g. Rust 1.76).

[rfc3355]: https://rust-lang.github.io/rfcs/3355-rust-spec.html#goals-and-non-goals

[Vision]:
https://blog.rust-lang.org/inside-rust/2023/11/15/spec-vision.html

### LINKED ISSUE #2025: Guidelines for parts that are incomplete or undecided
I think the authoring guide should have some guidelines on when something should be intentionally undecided, and how to state that. (Like certain evaluation orders.)

Some current examples:
- [`type.pointer.validity.pointer-fragment`](https://doc.rust-lang.org/1.90.0/reference/types/pointer.html#r-type.pointer.validity.pointer-fragment)
- [`macro.ambiguity.invariant.unseparated-complex-nt`](https://doc.rust-lang.org/1.90.0/reference/macro-ambiguity.html#r-macro.ambiguity.invariant.unseparated-complex-nt)

My notes:

* Maybe have something that talks about avoiding accidentally documenting something that is currently *undecided*. 
    * Guidelines on when and how to make notes about parts of the language that are still undecided.

### LINKED ISSUE #2026: Guidelines for parts that are intentionally unspecified
I think the authoring guide should have some guidelines on when something can be intentionally unspecified. Do we want to ever allow that? If so, do we want to formalize that (similar to how C spec does)?

### LINKED ISSUE #2027: Guidelines for referring to rustc lints
I think the authoring guide should have some guidelines on referring to rustc lints.

Generally we prefer to avoid referring to them, but in several cases it is a disservice to not mention them. It would be helpful to have guidance on how to balance those kinds of references.

Some examples of lints that are currently referred to:
* `unsafe_op_in_unsafe_fn`
* `ffi_unwind_calls`
* `unused_variables`
* `unused_must_use`
* `unfulfilled_lint_expectations`
* `overflowing_literals`

There's other cases where it currently doesn't refer to the specific lint, for example the `#[deprecated]` attribute, which arguably would benefit for a link to it.

As part of the attribute rework, we have a bunch of notes about `rustc` linting about certain things (like unused attributes, or future-incompatible warnings), but they don't refer to the lint by name.

### LINKED ISSUE #2037: Guidelines for documenting intended vs actual behavior
I think the authoring guide should have some guidance on whether to document the actual behavior of `rustc` vs the intended behavior, and how to handle known bugs.

If there is a known issue, should it be encouraged or discouraged to link to the GitHub issue?

Should it have specific callouts for future-incompatible warnings? How should these known problems be documented? Should it document how it is intended to work? What if the intention is not yet known?

### LINKED ISSUE #2038: Guidelines for deprecated language features and intended future changes
I think the author guide should have some guidance on referring to deprecated language features, and anything that is intended to change in the future.

For example, things that we know are deprecated, but will not be removed in an Edition, or whose replacement is not yet implemented or certain.

Similarly, if something is removed or changed in an Edition, should the Edition-specific docs mention that a particular thing is deprecated?
- ehuss's preference is to *not* mention that for Edition-specific changes. Non-edition changes are a bit more difficult to make a judgement on.

Some examples that are currently documented:
* `local_inner_macros`
* All the future-incompatible-warning stuff, which are (always?) in a note block.
* [`expr.method.edition2021`](https://doc.rust-lang.org/nightly/reference/expressions/method-call-expr.html?highlight=expr.method.edition2021#r-expr.method.edition2021) talks about possibly removing in the future.
* [`items.traits.params.patterns-no-body`](https://doc.rust-lang.org/nightly/reference/items/traits.html?highlight=items.traits.params.patterns-no-body#r-items.traits.params.patterns-no-body) deprecated `mut IDENTIFIER`

### LINKED ISSUE #2039: Guidelines for linking to other documentation
I think the authoring guide should have guidelines for linking to external documentation.

There are some resources that could be useful to typical Rust developers, like TRPL, or other books or guides. However, having lots of links can be noisy.

There are also non-Rust references, which might be relevant (like other standards). But where do you draw the line?

What is the policy for linking to historical information, such as RFCs, PRs, GitHub issues, blog posts, etc?  These sources are mostly static and not updated, and thus may provide outdated or incorrect information. However, as historical context they are useful.
- ehuss's preference: Don't ever link to RFCs, PRs, GitHub issues, blog posts, and similar things. However, we inevitably will hit sticky issues where we know something is wrong, and not linking to the relevant issue is withholding useful information.

### LINKED ISSUE #2040: Guidelines for implementation-specific rules
I think the authoring guide should have guidelines for how to handle implementation-specific rules. Should they ever be allowed? Where does the line get drawn?

Some current examples:
- Is the filesystem layout part of the language or part of the implementation?
- All the various notes we currently have regarding rustc and cargo.

In the past, there has been a sentiment from some in the project to never allow that. I largely agree with that. However, I think there will be gray areas that will be difficult to navigate with such a prohibition.

Unspecified is slightly different which is noted in https://github.com/rust-lang/reference/issues/2026. This is more related to the C-style "implementation-defined behavior — unspecified behavior where each implementation documents how the choice is made".

Note that there is another category of this that is "target-specific" behavior, though I think we have resolved that in https://github.com/rust-lang/spec/pull/58 which is now in the authoring guide.

ehuss:
This reworks the introduction and adds a scope chapter to hopefully make it clearer exactly what the Reference is intended to cover. In summary, the changes are:

- Reworks the introduction.
- Moves some of the conventions to a dedicated chapter.
- Adds a scope chapter.

This is based on my understanding of how the Reference currently works.

The goal is to provide guidance on all the topics covered here. However, since a lot of these guidelines use subjective words (such as "significant" or "exceptional"), it may not provide enough guidance. I expect that some of this may need more elaboration in the future to resolve uncertainties.

### Should we document when features are stabilized?

Closes rust-lang/reference#60

Answer: The scope now says version history is out of scope.

### Clarify the audience and goals of the reference

Closes rust-lang/reference#550

Answer: Introduction now has an audience section.

### Policy: Past behavior

Closes rust-lang/reference#2024

Answer: The scope now says past behavior is out of scope.

### Guidelines for parts that are incomplete or undecided

Closes rust-lang/reference#2025

Answer: The completness section largely covers this. There's a few points here that aren't really covered, but I couldn't think of a way to incorporate them:

- Things should be marked as undecided when the teams say so. That seems fairly self-evident.
- There is no standard notation for stating something is undecided, incomplete, unspecified, undefined, etc. I'll follow-up on that.
- There's no real way to avoid accidentally documenting undecided behavior except for involving the teams with reviews. We generally try to do that as a matter of policy when documenting things we aren't confident about.

### Guidelines for parts that are intentionally unspecified

Closes rust-lang/reference#2026

Answer: Covered in "unspecified behavior".

### Guidelines for referring to rustc lints

Closes rust-lang/reference#2027

Answer: Covered in scope chapter.

### Guidelines for documenting intended vs actual behavior

Closes rust-lang/reference#2037

Answer: This is largely covered in the intended behavior section. It uses subjective wording, which maybe isn't super helpful. It also doesn't get really explicit about when to link to an issue or how to handle future-incompatible-warnings. However, I think the spirit is there, and it's hard to be clearer.

### Guidelines for deprecated language features and intended future changes

Closes rust-lang/reference#2038

Answer: Covered in the scope chapter. Notes:

- Future-incompatible warnings are briefly mentioned in intended behavior.
- Recommendations are excluded in out-of-scope.
- Future plans are excluded in out-of-scope.
- Lints are covered in out-of-scope.
- Editions are plainly described as differences to the present behavior.

### Guidelines for linking to other documentation

Closes rust-lang/reference#2039

Answer: I think this is largely covered in out-of-scope, and other places.

### Guidelines for implementation-specific rules

Closes rust-lang/reference#2040

Answer: Essentially we have decided that we won't have a special "implementation-defined" category for now. The Reference will be explicit about the allowed behavior (even if that means there are different options), leave it up to unspecified-behavior, or leave it up to special uncovered topics like certain target-specific behavior.

This is not explicitly stated in the scope since it is covered by other things.

### Guidelines for documenting the *why* of something

Closes rust-lang/reference#2042

Answer: The new scope rationale section is brief, but I think largely covers this. There are subjective words for the guidelines, which maybe isn't super helpful. But I can't think of super clear guidance without going in too deep with examples and such.

### Guidelines for referring to tools and compilers

Closes rust-lang/reference#2043

Answer: I think this is generally covered, though it still uses subjective adjectives like "significant" or "exceptional".


src/scope.md
```diff
@@ -0,0 +1,132 @@
+# Scope
+
+## Scope of the Reference
+
+The scope of the Rust Reference is to fully describe the syntax and semantics of the Rust language. In this context, the language encompasses:
+
+- Valid syntax and forms.
+- Meaning and interpretation of the syntax.
+- Semantic properties of the runtime behavior.
```
RalfJung, line 9: What exactly is meant by "Semantic properties of the runtime behavior"? This sounds vaguely opsem-related and yet I have no idea what you mean. ;)

FWIW, in a typical PL textbook, we'd typically have 3 stages for defining a language; I am not sure if the first three items are meant to match that.
- Formal grammar (corresponds to "valid syntax and forms" I think)
- Static semantics -- this is the type system and all the other static checks the compiler might be doing (in Rust that would include borrow checking [unless you consider that to be part of the type system], const checking, things like that)
- Dynamic semantics -- the set of legal "observable behaviors" or "outcomes" or whatever you want to call it of any program

src/introduction.md
```diff
@@ -2,6 +2,16 @@
 
 This book is the primary reference for the Rust programming language. It serves as a description of the syntax, semantics, and interpretation of the Rust language.
 
+## Audience
+
+The audience for this book includes:
+
+- Rust users who want to learn about the specifics and details of the language.
```
tshepang, line 9: not sure what the difference is between *specifics* and *details*

src/conventions.md
```diff
@@ -0,0 +1,59 @@
...
+  fn main() {
+      println!("This is a code example");
+  }
+  ```
+
+  All examples are written for the latest edition unless otherwise stated.
+
+* The grammar and lexical productions are described in the [Notation] chapter.
+
+r[example.rule.label]
```
tshepang: shows that the notation accepts dashes
`[general.less-general.less-specific.specific]`

anyways, real reason I propose this is that seeing "example" here made me think of code example, perhaps because I just saw how a code example is done (in the previous bullet point)

src/introduction.md
ehuss: No real difference. I dropped the extra word.

src/scope.md
```diff
@@ -0,0 +1,132 @@
...
+- Version history or past behavior.
+    - Changes in Rust releases are recorded in the [Release Notes]. Prior releases of the Reference are available as described in [Rust releases].
+- Historical documentation.
+    - This includes documentation such as [RFCs], design documents, reports, blog posts, issues, pull requests, etc.
+- Rationale for most aspects of the language.
+    - Rationale may be included in a note in exceptional situations, particularly when a feature is not obvious or has especially useful relevance.
+- `rustc` lints.
+    - Lints are documented in the [rustc book][lints].
+    - Lints that have a specific relation to some part of the language may be referred to by name. For example, the `expect` attribute can issue the `unfulfilled_lint_expectations` lint. However, the exact behavior of when the lint is issued is usually not documented in the Reference.
+    - Notes may refer to lints when the lint has exceptional relevance to some aspect of the language.
```
ehuss, line 44: I'm wondering if we could provide further guidance here to mention [deny-by-default lints](https://doc.rust-lang.org/nightly/rustc/lints/listing/deny-by-default.html)? (The non-FCW ones at least.) Since they are relatively uncommon, and are deny because they are important and we want to steer people away from those hazards, maybe that is a good guideline to say that those are good examples of ones to mention in notes?

Unfortunately it gets a little difficult for some lints like `unconditional_panic` because that one lint covers a lot of different scenarios, and I don't think we want to document all of them (at least not here). (Because that lint is to some degree a best-effort thing, can change frequently over time, and I assume is pretty complicated.)

src/conventions.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Rules
+
+r[example.rule.label]
+Rule identifiers appear before each language rule, enclosed in square brackets. These identifiers provide a way to refer to and link to a specific rule in the language ([e.g.][example rule]). The rule identifier uses periods to separate sections from most general to most specific ([destructors.scope.nesting.function-body] for example). On narrow screens, the rule name will collapse to display `[*]`.
+
+The rule name can be clicked to link to that rule.
+
+> [!WARNING]
+> The organization of the rules is currently in flux. For the time being, these identifier names are not stable between releases, and links to these rules may fail if they are changed. We intend to stabilize these once the organization has settled so that links to the rule names will not break between releases.
```
DanielEScherzer, line 15: Maybe something like

> For now, consider linking to the release-specific versions of the specification to ensure that links continue to work.

Or something to give a workaround if you want to be able to link to what the current rule is?

src/conventions.md
```diff
@@ -0,0 +1,81 @@
...
+## Notes
+
+Notes that contain useful information about the state of the book or point out useful, but mostly out of scope, information are in note blocks.
+
+> [!NOTE]
+> This is an example note.
+
+## Warnings
+
+Warnings that show unsound behavior in the language or possibly confusing interactions of language features are in a special warning box.
```
DanielEScherzer:
```suggestion
Warnings that show unsound behavior in the language or possibly confusing interactions of language features are in special warning boxes.
```

Suggest aligning pluralization with the notes ("in note blocks") and edition alerts ("in edition blocks")

src/introduction.md
```diff
@@ -1,144 +1,51 @@
...
 
 There are two main ways to read this document.
 
-The first is to answer a specific question.
-If you know which chapter answers that question, you can jump to that chapter in the table of contents.
-Otherwise, you can press `s` or click the magnifying glass on the top bar to search for keywords related to your question.
-For example, say you wanted to know when a temporary value created in a let statement is dropped.
-If you didn't already know that the [lifetime of temporaries] is defined in the [expressions chapter], you could search "temporary let" and the first search result will take you to that section.
+The first is to answer a specific question. If you know which chapter answers that question, you can jump to that chapter in the table of contents. Otherwise, you can press `s` or click the magnifying glass on the top bar to search for keywords related to your question. For example, say you wanted to know when a temporary value created in a `let` statement is dropped. If you didn't already know that the [lifetime of temporaries] is defined in the [expressions chapter], you could search "temporary let" and the first search result will take you to that section.
```
DanielEScherzer, line 21: can I suggest doing the unrolling in a separate pull so that it is easier to identify what changes are made to the actual content?

src/introduction.md
```diff
@@ -1,144 +1,51 @@
...
 
-You can contribute to this book by opening an issue or sending a pull request to [the Rust Reference repository].
-If this book does not answer your question, and you think its answer is in scope of it, please do not hesitate to [file an issue] or ask about it in the `t-lang/doc` stream on [Zulip].
-Knowing what people use this book for the most helps direct our attention to making those sections the best that they can be.
-And of course, if you see anything that is wrong or is non-normative but not specifically called out as such, please also [file an issue].
+You can contribute to this book by opening an issue or sending a pull request to [the Rust Reference repository]. If this book does not answer your question and you think its answer is in scope, please do not hesitate to [file an issue] or ask about it in the `t-lang/doc` stream on [Zulip]. Knowing what people use this book for the most helps direct our attention to making those sections the best they can be. And of course, if you see anything that is wrong or is non-normative but not specifically called out as such, please also [file an issue].
+
+<!-- TODO: Link to contributing guide -->
 
-[book]: ../book/index.html
```
DanielEScherzer, line 130: I found the links to the other books (particularly the unstable book) to be helpful, can we keep those in some way easier to find than just looking in the separate scope page?

src/conventions.md
ehuss: Thanks, added in https://github.com/rust-lang/reference/pull/2100/changes/d9aa0d3238f6cb1c3358255deac11188c542db90.

PR #2103, opened 2025-12-08
`cfg_select!` macro

folkertdev:
Stabilization report: https://github.com/rust-lang/rust/pull/149783

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -441,19 +441,22 @@ r[cfg.cfg_select]
 
 r[cfg.cfg_select.syntax]
 ```grammar,configuration
-CfgSelect ->
-    cfg_select! `{` CfgSelectBranch* `}`
+@root CfgSelect -> CfgSelectArms?
+
+CfgSelectArms ->
+    CfgSelectConfigurationPredicate `=>`
+    (
+        `{` ^ TokenTree `}` CfgSelectArms?
+      | ExpressionWithBlockNoAttrs `,`? CfgSelectArms?
+      | ExpressionWithoutBlockNoAttrs ( `,` CfgSelectArms? )?
+    )
 
 CfgSelectConfigurationPredicate ->
     ConfigurationPredicate | `_`
-
-CfgSelectBranch ->
-    CfgSelectConfigurationPredicate `=>` `{` TokenTree `}`
-  | CfgSelectConfigurationPredicate `=>` TokenTree `,`
 ```
 
 r[cfg.cfg_select.general]
-The built-in `cfg_select` macro expands to the `TokenTree` on the right-hand side of the first configuration predicate that evaluates to `true`.
+The built-in `cfg_select` macro expands to the arm payload of the first configuration predicate that evaluates to `true`. If the payload is wrapped in curly braces, those are removed during expansion.
 
 For example:
 
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -498,8 +498,8 @@ Each right-hand side must be a syntactically valid expansion for the position in
 [`--test`]: ../rustc/command-line-arguments.html#--test-build-a-test-harness
 [`cfg`]: #the-cfg-attribute
 [`cfg!`]: #the-cfg-macro
-[`cfg_select!`]: #the-cfg_select-macro
 [`cfg_attr`]: #the-cfg_attr-attribute
+[`cfg_select!`]: #the-cfg_select-macro
 [`crate_name`]: crates-and-source-files.md#the-crate_name-attribute
 [`crate_type`]: linkage.md
 [`target_feature` attribute]: attributes/codegen.md#the-target_feature-attribute
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -478,7 +478,11 @@ CfgSelectConfigurationPredicate ->
     ConfigurationPredicate | `_`
 ```
 
+r[cfg.cfg_select.first-arm]
+`cfg_select` expands to the payload of the first arm whose configuration predicate evaluates to true.
 
+r[cfg.cfg_select.braces]
+If the entire payload is wrapped in curly braces, the braces are removed during expansion.
 
 r[cfg.cfg_select.wildcard]
 A `_` can be used to write a configuration predicate that always evaluates to `true`.
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -491,7 +491,7 @@ r[cfg.cfg_select.fallthrough]
 It is a compile error if none of the predicates evaluate to true.
 
 r[cfg.cfg_select.well-formed]
-Each right-hand side must syntactically be valid expansion for the position that the macro is invoked in.
+Each right-hand side must be a syntactically valid expansion for the position in which the macro is invoked.
 
 [Testing]: attributes/testing.md
 [`--cfg`]: ../rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -490,17 +490,6 @@ The configuration predicate `_` always evaluates to true.
 r[cfg.cfg_select.fallthrough]
 If none of the predicates evaluates to `true`, a compiler error is emitted.
 
-r[cfg.cfg_select.positions]
-The `cfg_select!` macro is accepted in the following macro expansion positions
-
-- items
-- statements
-- expression
-- impl items
-- trait impl items
-- trait items
-- foreign items
-
 r[cfg.cfg_select.well-formed]
 Each right-hand side must syntactically be valid expansion for the position that the macro is invoked in.
 
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -439,7 +439,21 @@ println!("I'm running on a {} machine!", machine_kind);
 r[cfg.cfg_select]
 ### The `cfg_select` macro
 
-The built-in `cfg_select` macro expands to the right-hand side of the first configuration predicate that evaluates to `true`.
+r[cfg.cfg_select.syntax]
+```grammar,configuration
+CfgSelect ->
+    cfg_select! `{` CfgSelectBranch* `}`
+
+CfgSelectConfigurationPredicate ->
+    ConfigurationPredicate | `_`
+
+CfgSelectBranch ->
+    CfgSelectConfigurationPredicate `=>` `{` TokenTree `}`
+  | CfgSelectConfigurationPredicate `=>` TokenTree `,`
+```
+
+r[cfg.cfg_select.general]
+The built-in `cfg_select` macro expands to the `TokenTree` on the right-hand side of the first configuration predicate that evaluates to `true`.
 
 For example:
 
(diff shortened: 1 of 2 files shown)


src/conditional-compilation.md
```diff
@@ -466,11 +466,43 @@ let machine_kind = if cfg!(unix) {
...
+The `cfg_select` macro can also be used in expression position:
+
+```rust
+let is_unix_str = cfg_select! {
+    unix => "unix",
+    _ => "not unix",
+};
+```
+
+A `_` can be used to write a configuration predicate that always evaluates to `true`.
```
traviscross: Thanks for the Reference PR.  It'd be worth filling in the details you provided in https://github.com/rust-lang/rust/pull/149783#issuecomment-3629464095, e.g. with the grammar, the removal of one level of braces, the compile error on fallthrough, etc.

traviscross: This can discuss as well how the bodies of each branch must be able to parse.

src/conditional-compilation.md
```diff
@@ -466,11 +466,75 @@ let machine_kind = if cfg!(unix) {
...
+CfgSelectConfigurationPredicate ->
+    ConfigurationPredicate | `_`
+
+CfgSelectBranch ->
+    CfgSelectConfigurationPredicate `=>` `{` TokenTree `}`
+  | CfgSelectConfigurationPredicate `=>` TokenTree `,`
+```
+
+r[cfg.cfg_select.general]
+The built-in `cfg_select` macro expands to the `TokenTree` on the right-hand side of the first configuration predicate that evaluates to `true`.
```
folkertdev: is it clear enough from the grammer and the language here that the `{ /* ... */ }` are dropped?

src/conditional-compilation.md
```diff
@@ -466,11 +466,75 @@ let machine_kind = if cfg!(unix) {
...
+r[cfg.cfg_select.positions]
+The `cfg_select!` macro is accepted in the following macro expansion positions
+
+- items
+- statements
+- expression
+- impl items
+- trait impl items
+- trait items
+- foreign items
```
folkertdev: I can't find the positions in which macro expansion is allowed in the reference, actually.

folkertdev (commented):
I think all the raw information is here now, but it probably needs some editing/clarification.

Also should the lint on branches that come after a wildcard be mentioned?

src/conditional-compilation.md
ehuss: The list of macro invocation sites is listed in https://doc.rust-lang.org/nightly/reference/macros.html#r-macro.invocation.intro. I don't think this needs to duplicate the list.

We haven't really gotten into documenting built-in attributes, so there isn't a well-trodden path here. I think it would be fine to say that it may specified in any position where macro invocations are allowed.

src/conditional-compilation.md
```diff
@@ -466,11 +466,75 @@ let machine_kind = if cfg!(unix) {
...
+CfgSelect ->
+    cfg_select! `{` CfgSelectBranch* `}`
+
+CfgSelectConfigurationPredicate ->
+    ConfigurationPredicate | `_`
+
+CfgSelectBranch ->
+    CfgSelectConfigurationPredicate `=>` `{` TokenTree `}`
+  | CfgSelectConfigurationPredicate `=>` TokenTree `,`
+```
```
ehuss, line 479: This grammar doesn't quite look correct to me.

Some comments:
- We don't have a precedent for documenting macro inputs. I don't think we can express it like `` `cfg_select!` `{` … `}` `` because the braces can also be `()` or `[]` (also `!` is a separate token), and the macro name can be renamed. I would suggest just documenting what the input to the macro is, and not the surrounding invocation syntax.
- This doesn't handle the difference in behavior of the last branch.
- This doesn't specify that the non-braced value needs to be an expression of a certain kind.
- This doesn't handle the optional `,` behavior.

I'm thinking the grammar would be something closer to:

```grammar,configuration
@root CfgSelect ->
    CfgSelectBranch*
    LastCfgSelectBranch?

CfgSelectBranch ->
    CfgSelectConfigurationPredicate `=>` `{` TokenTree `}`
  | CfgSelectConfigurationPredicate `=>` ExpressionWithBlockX `,`?
  | CfgSelectConfigurationPredicate `=>` ExpressionWithoutBlockX `,`

ExpressionWithBlockX ->
      BlockExpression
    | ConstBlockExpression
    | UnsafeBlockExpression
    | LoopExpression
    | IfExpression
    | MatchExpression

ExpressionWithoutBlockX ->
      LiteralExpression
    | PathExpression
    | OperatorExpression
    | GroupedExpression
    | ArrayExpression
    | AwaitExpression
    | IndexExpression
    | TupleExpression
    | TupleIndexingExpression
    | StructExpression
    | CallExpression
    | MethodCallExpression
    | FieldExpression
    | ClosureExpression
    | AsyncBlockExpression
    | ContinueExpression
    | BreakExpression
    | RangeExpression
    | ReturnExpression
    | UnderscoreExpression
    | MacroInvocation

LastCfgSelectBranch ->
    CfgSelectConfigurationPredicate => ExpressionX `,`?

ExpressionX -> ExpressionWithBlockX | ExpressionWithoutBlockX

CfgSelectConfigurationPredicate ->
    ConfigurationPredicate | `_`
```

Where the X expressions are from the Expression grammar, but without the outer attributes. If this is correct, we'll need to rework `Expression` to support that.

Does that make sense?

ehuss (commented):
> Also should the lint on branches that come after a wildcard be mentioned?

As for lints, we normally don't document those. However, in this situation, it seems like it would be fine to have a `> [!NOTE]` block that mentions the lint (in whichever way https://github.com/rust-lang/rust/pull/149960 ends up being).

src/conditional-compilation.md
ehuss: Though looking again, my suggestion above won't work because we are moving towards a grammar that does not have infinite lookahead for disambiguation. So the `LastCfgSelectBranch` won't work. I offhand can't think of a way to actually express that...

(Maybe lookahead is required?)

ehuss:
What does [this line](https://github.com/rust-lang/rust/blob/7704328ba5ae8d6ce0ac303c9d5a1a1605906766/compiler/rustc_parse/src/parser/cfg_select.rs#L25) do when checking for a closebrace?

I would expect that:
- If it is wrapped in braces, then it went through the initial braced-token-tree branch above.
- Otherwise, it's impossible for a branch to start with `}`.

That is, I would expect that `token::Eof` is what matters, and it is not possible for it to be `token::CloseBrace`.

src/conditional-compilation.md
traviscross: Pushed something that might be closer.

src/conditional-compilation.md
```diff
@@ -466,17 +466,85 @@ let machine_kind = if cfg!(unix) {
...
+
+r[cfg.cfg_select.syntax]
+```grammar,configuration
+@root CfgSelect -> CfgSelectArms?
+
+CfgSelectArms ->
+    CfgSelectConfigurationPredicate `=>`
+    (
+        ExpressionWithBlockNoAttrs `,`? CfgSelectArms?
+      | ExpressionWithoutBlockNoAttrs ( `,` CfgSelectArms? )?
```
ehuss: I think this needs to include `{` TokenTree `}`. Otherwise, this wouldn't match other non-expression things like patterns, types, etc.

src/conditional-compilation.md
ehuss: Also, did you intend to exclude patterns and types in this list?

folkertdev: I suspect that the list was accurate at the time, but it no longer is and should now include patterns and types.

src/conditional-compilation.md
traviscross: Yes... updated to:

```abnf
CfgSelectArms ->
    CfgSelectConfigurationPredicate `=>`
    (
        `{` ^ TokenTree `}` CfgSelectArms?
      | ExpressionWithBlockNoAttrs `,`? CfgSelectArms?
      | ExpressionWithoutBlockNoAttrs ( `,` CfgSelectArms? )?
    )
```

It needs the cut because if it takes the first branch but then there's a comma after the closing brace, we can't have it backtrack and succeed on the second branch.  (Negative lookahead in the second branch would also work for this.)

@folkertdev: Is there a reason for not allowing an optional comma after the closing brace?

traviscross: @ehuss: Do you think it's possible to be exhaustive here?  I.e., rather than `TokenTree`, could we have some more-semantic production that comprised patterns, types, etc., or does not work?

ehuss: I *think* it would be possible to say something like `{` Expression | Statement | Item | Pattern | Type `}`

I'm pretty sure that due to the way expansion works, an arbitrary TokenTree can't really be inserted into the token stream. It has to be a contiguous unit interpreted based on the context.

But there's a problem of how to interpret this grammar. Is the grammar saying, "this is the valid input" but no other interpretation? Or is it saying "something that matches a Statement means it expands to a Statement, and would be invalid in non-statement contexts"?

I think the former is the only interpretation we can make. For example, `ExpressionWithoutBlockNoAttrs` is allowed in pattern contexts. But a pattern is obviously not an expression. So we can't give meaning to the input beyond "these are the tokens that are accepted".

The former would also deal with the ambiguity. For example `Statement` overlaps with `Item`. So you can't interpret the grammar in a way that would fit with the syntax tree.  For example, given a token tree of a single identifier `foo`, just given that grammar above, you would say it is an Expression. But used in a pattern context, Expression is wrong (it is a Pattern element).

I don't have a strong opinion on using TokenTree versus something more specific.

src/conditional-compilation.md
```diff
@@ -466,11 +466,78 @@ let machine_kind = if cfg!(unix) {
...
+### The `cfg_select` macro
+
+r[cfg.cfg_select.syntax]
+```grammar,configuration
+@root CfgSelect -> CfgSelectArms?
+
+CfgSelectArms ->
+    CfgSelectConfigurationPredicate `=>`
+    (
+        `{` ^ TokenTree `}` CfgSelectArms?
```
ehuss:
```suggestion
        `{` ^ TokenTree `}` `,`? CfgSelectArms?
```

After https://github.com/rust-lang/rust/pull/152341.

src/conditional-compilation.md
```diff
@@ -466,11 +466,78 @@ let machine_kind = if cfg!(unix) {
...
+        `{` ^ TokenTree `}` `,`? CfgSelectArms?
+      | ExpressionWithBlockNoAttrs `,`? CfgSelectArms?
+      | ExpressionWithoutBlockNoAttrs ( `,` CfgSelectArms? )?
+    )
+
+CfgSelectConfigurationPredicate ->
+    ConfigurationPredicate | `_`
+```
+
+r[cfg.cfg_select.general]
```
ehuss:
```suggestion
r[cfg.cfg_select.intro]
```

These got renamed in https://github.com/rust-lang/reference/pull/2133.

Also, I would move this above the `.syntax` rule. I realize the rest of the current reference is a little inconsistent, but we are moving more towards the intro paragraphs being first.

Then, I might suggest reworking this so that the behavior rules are distinct from the intro. Generally we want the outline to be something like:

- intro
- syntax
- describe the behavior
- describe any restrictions, exceptions, etc.

I can certainly write a suggestion of what I'm thinking if that helps.

tshepang: @ehuss should this be in dev-guide

ehuss:
I pushed up some updates and I think this should be ready to go.

folkertdev:
I agree, this looks good to me.

ehuss (approved):
Thanks! Merging now that https://github.com/rust-lang/rust/pull/149783 is on nightly.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2103/70c99cb152e25d6d134d952be868a8c53c4c894f..5c1d0c5a9b616348d036f780d806ede5ecbd4f3c)*

PR #2097, opened 2025-11-25
Add a contributor guide

ehuss:
This adds a new contributor guide. It is built with mdbook with the intention that it is published on GitHub Pages. This replaces several of the loosely organized documents that we have accumulated over time. It also adds some new content aimed at contributors, such as an overview of the contribution process. 

Several of the chapters are essentially stubs or basic starting points with the intention that they get reworked in the future. There's also still a lot of work left to resolve the [Meta issues](https://github.com/rust-lang/reference/issues?q=is%3Aissue%20state%3Aopen%20label%3AMeta).

Note that I am also working on a separate but closely related initiative to add a scope chapter to the Reference itself, along with a rework of the introduction. This is intended to further clarify to contributors (and readers!) what is expected of the content.

PR #2104, opened 2025-12-14
Add cut operator (`^`) to grammar

traviscross:
The cut operator (`^`) is a backtracking fence.  Once the expression to its left succeeds, we become committed to the alternative; the remainder of the expression must parse successfully or parsing will fail.  See *Packrat Parsers Can Handle Practical Grammars in Mostly Constant Space*, Mizushima et al., <https://kmizu.github.io/papers/paste513-mizushima.pdf>.

This operator solves a problem for us with C string literals.  These literals cannot contain a null escape.  But if we simply fail to lex the literal (e.g. `c"\0"`), we may instead lex it successfully as two separate tokens (`c "\0"), and that would be incorrect.

As long as we only use cut to express constraints that can be expressed in a regular language and we keep our alternations disjoint, the grammar can still be mechanically converted to a CFG.

Let's add the cut operator to our grammar and use it for C string literals and some similar constructs.

In the railroad diagrams, we'll render the cut as a "no backtracking" box around the expression or sequence of expressions after the cut.  The idea is that once you enter the box the only way out is forward.

(H/t to @ehuss for suggesting the cut operator to solve this problem.)

cc @ehuss

---

This is stacked on https://github.com/rust-lang/reference/pull/2097 and should merge after it.

### DIFF
--- a/dev-guide/src/grammar.md
+++ b/dev-guide/src/grammar.md
@@ -35,7 +35,9 @@ Name -> <Alphanumeric or `_`>+
 
 Expression -> Sequence (` `* `|` ` `* Sequence)*
 
-Sequence -> (` `* AdornedExpr)+
+Sequence ->
+        (` `* AdornedExpr)* ` `* Cut
+      | (` `* AdornedExpr)+
 
 AdornedExpr -> ExprRepeat Suffix? Footnote?
 
@@ -92,6 +94,8 @@ Prose -> `<` ~[`>` LF]+ `>`
 Group -> `(` ` `* Expression ` `* `)`
 
 NegativeExpression -> `~` ( Charset | Terminal | NonTerminal )
+
+Cut -> `^` Sequence
 ```
 
 The general format is a series of productions separated by blank lines. The expressions are as follows:
@@ -110,6 +114,7 @@ The general format is a series of productions separated by blank lines. The expr
 | Prose | \<any ASCII character except CR\> | An English description of what should be matched, surrounded in angle brackets. |
 | Group | (\`,\` Parameter)+ | Groups an expression for the purpose of precedence, such as applying a repetition operator to a sequence of other expressions. |
 | NegativeExpression | ~\[\` \` LF\] | Matches anything except the given Charset, Terminal, or Nonterminal. |
+| Cut | Expr1 ^ Expr2 \| Expr3 | The hard cut operator. Once the expressions preceding `^` in the sequence match, the rest of the sequence must match or parsing fails unconditionally --- no enclosing expression can backtrack past the cut point. |
 | Sequence | \`fn\` Name Parameters | A sequence of expressions that must match in order. |
 | Alternation | Expr1 \| Expr2 | Matches only one of the given expressions, separated by the vertical pipe character. |
 | Suffix | \_except \[LazyBooleanExpression\]\_  | Adds a suffix to the previous expression to provide an additional English description, rendered in subscript. This can contain limited Markdown, but try to avoid anything except basics like links. |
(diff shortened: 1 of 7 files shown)


mattheww:
When you're defining a cut operator, it's important to specify the scope over
which it cancels re-attempts.

For the lexer's purposes it would be fine to make that scope unlimited, saying that if the right-hand side of a production containing a cut fails after reaching the cut then the entire lexing process fails.

But in this PR the nearest thing to a definition of the cut operator is the reference to https://kmizu.github.io/papers/paste513-mizushima.pdf .

That paper defines a cut operator with a narrower scope: it allows a cut only on the left-hand side of an ordered choice expression, and cancels only the re-attempt of the right-hand side of that expression.

That definition doesn't work for the positions in which this PR is placing cuts.

mattheww:
If you're still planning to use prioritised choice more widely in the Reference, then (given that the Reference already has the notion of reserved forms) perhaps the simplest way to define cut is to say that:

`FOO` ← a ^ b

is a shorthand for

`FOO` ← a b
`RESERVED_PREFIX_OF_FOO` ← a

with `RESERVED_PREFIX_OF_FOO` appearing immediately after `FOO` in the lexical grammar's top-level ordered choice for tokens.

That characterisation also illustrates why adding the notion of a cut doesn't buy very much.

In particular if you use prioritised choice for the token rules then (for the lexing dialect used in Rust 2021 and later) you can simplify the existing reserved token rules, getting rid of the "except b or c or r or br or cr" business, and end up with something like this:

`C_STRING_LITERAL` ← `c` `"` (more stuff)
`RESERVED_TOKEN_DOUBLE_QUOTE` ← `IDENTIFIER_OR_KEYWORD` `"`
`IDENTIFIER_OR_KEYWORD` ← (same as at present)

This way `c"\0"` is rejected as a reserved form, and there's no need to bother the Reference's readers with a discussion of cuts.

traviscross:
Yes, @ehuss and I had earlier discussed these same matters, point for point.  Thanks for elaborating them here; good to have these written out on the PR.

With regard to using `RESERVED_` and ordered choice, that was my first thought too for solving this problem.  But it's less theoretically satisfying than cut (with global escape) -- to me anyway -- since the input is still parsed or lexed successfully as far as the grammar is concerned.  The idea that `RESERVED_` rules matching are failures is something that has to be overlaid.

traviscross:
With regard to the locality (or lack thereof) of cut, it's interesting that [Python's grammar](https://docs.python.org/3/reference/grammar.html) has a rule with cut that would be a no-op under the Mizushima et al. interpretation:

```text
assignment_expression:
    | NAME ':=' ~ expression 
```

But their description of cut does not clearly suggest a global escape:

> `~` (“cut”): commit to the current alternative and fail the rule even if this fails to parse

traviscross:
One library that takes the local interpretation is [pegase](https://github.com/strblr/pegase).  As they [describe](https://strblr.github.io/pegase/advanced-concepts/Cut-operator/):

> Used outside an ordered choice expression, it's simply a no-op.

mattheww:
> With regard to the locality (or lack thereof) of cut, it's interesting that [Python's grammar](https://docs.python.org/3/reference/grammar.html) has a rule with cut that would be a no-op under the Mizushima et al. interpretation:

AIUI the cuts in the Python grammar are there for performance (and debuggability) reasons, not to change the accepted language.

The parser can discard some state as soon as it gets to the cut (as described in the paper you linked), so it's useful even when there's no choice operator following.

mattheww:
While you're looking at the quoted literals, you might consider helping the cut operator earn its keep by changing

`SUFFIX` → `IDENTIFIER_OR_KEYWORD`<sub>except _</sub>

to something equivalent to

`SUFFIX` → `XID_Start` `XID_Continue`<sup>*</sup> | `_` ^ `XID_Continue`<sup>+</sup>

That would fix another bug in this family, by preventing something like `"xxx"_` being analysed as two tokens.

traviscross:
@ehuss has a forthcoming PR, likely to supersede this one, that resolves a large number of grammar issues.  On that branch, looks like he went with:

```text
SUFFIX ->
      `_` ^ XID_Continue+
    | XID_Start XID_Continue*
```

ehuss:
Here's a commit that changes this to be an unary operator: https://github.com/ehuss/reference/commit/21ea969696f621ad39118d151a2b2de4f8f9df19

traviscross:
Thanks; cherry-picked.

ehuss (approved):
Thanks!

PR #2105, opened 2025-12-14
tokens: clarify in escape tables that digits are hex

DanielEScherzer:
Based on subsequent details later down the page, these are clarified to always be hex digits. Since "digit" generally refers to base-10 digits, note in the `lex.token.literal` escape tables that the digits are hex digits.


ehuss (approved):
Thanks!

PR #2107, opened 2025-12-14
macros-by-example: add space in `macro.decl.repetition.fragment` example

DanielEScherzer:
Other examples of repetitions being applied had a space both before and after the identifier, but `=> { $( $( $i)* )* }` was missing the space after `$i`.


ehuss (approved):
Thanks!

PR #2108, opened 2025-12-15
Fix restrictions of or-patterns

kirtchev-adacore:
A small discrepancy in the definition of `or-patterns` was discovered during the `rust-lang/fls` meeting on 2025-12-12 involving Eric Huss and TC - `or-patterns` must not be used in function and closure **parameters**, as opposed to function and closure arguments as stated in the text. Function and closure parameters are the constructs carrying `PatternNoTopAlt`s.


ehuss (approved):
Thanks!

PR #2109, opened 2025-12-16
conditional-compilation: add a space in `cfg.cfg_attr.attribute-list`

DanielEScherzer:
The other uses of `cfg` and `cfg_attr` have spaces on both sides of the `=`, but in the example of nested `cfg_attr` attributes, the space afterwards was missing; add it.


ehuss (approved):
Thanks!

PR #2110, opened 2025-12-16
items: clarify label for type-aliases documentation

DanielEScherzer:
"type definitions" could be about structs, enums, or unions too; the syntax diagram and the link target both refer to type aliases, so use that label for the link.


traviscross (approved):

traviscross:
Thanks @DanielEScherzer.

PR #2111, opened 2025-12-16
external-blocks: add missing "and" in list

DanielEScherzer:
When describing the type of functions declared in external blocks, include "and" before the last part of the description.


ehuss (approved):
Thanks!

PR #2116, opened 2025-12-18
associated-items: add missing periods

DanielEScherzer:
At the end of `items.associated.type.def` and `items.associated.type.alias`


ehuss (approved):
Thanks!

PR #2117, opened 2025-12-18
attributes: add missing punctuation to instruction_set

DanielEScherzer:
The short description for each built in attribute except `instruction_set` was terminated with a period; add the missing period. Add a missing apostrophe to "functions code" in that description.


ehuss (approved):
Thanks!

PR #2118, opened 2025-12-18
operator-expr: remove stray word in footnote

DanielEScherzer:
"only for closures that {x} can be {y}" repeats the limitation; it should either be "only closures that {x} can be {y}", or, if the capability ("{y}") is implicit, it can be "only for closures that {x}". Go with the first option that keeps the context, and remove the extra "for".


ehuss (approved):
Thanks, I pushed up some nearby fixes as well.

PR #2127, opened 2025-12-28
Update note about shebang removal in `include`d files

fmease:
Imports changes from https://github.com/rust-lang/rust/pull/146377.

I'm open to wording suggestions.

r? traviscross


ehuss (approved):
Thanks!

I applied a little editing for clarity.

PR #2128, opened 2025-12-30
Guarantee `repr(C)` union field offset

joshlf:
AFAICT, this behavior is widely assumed, and is the intended behavior. `repr(C)` unions are [designed to match the behavior of C unions](https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.union.intro), and [per the C standard](https://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16):

> The size of a union is sufficient to contain the
> largest of its members. The value of at most one of the
> members can be stored in a union object at any time. A
> pointer to a union object, suitably converted, points
> to each of its members (or if a member is a bit- field,
> then to the unit in which it resides), and vice versa.

Makes progress on https://github.com/rust-lang/unsafe-code-guidelines/issues/595

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -255,7 +255,7 @@ r[layout.repr.c.union.intro]
 A union declared with `#[repr(C)]` will have the same size and alignment as an equivalent C union declaration in the C language for the target platform.
 
 r[layout.repr.c.union.size-align]
-The union will have a size of the maximum size of all of its fields rounded to its alignment, and an alignment of the maximum alignment of all of its fields. These maximums may come from different fields.
+The union will have a size of the maximum size of all of its fields rounded to its alignment, and an alignment of the maximum alignment of all of its fields. These maximums may come from different fields. Each field lives at byte offset 0 from the beginning of the union.
 
 ```rust
 #[repr(C)]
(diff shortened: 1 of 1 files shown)


joshlf:
r? t-lang

(Is that valid r? syntax?)

RalfJung:
I'm surprised we don't already document this.^^

scottmcm (approved):
TBH I think this is so settled that we can just merge it without even an FCP, but cc @rust-lang/lang in case anyone has objections to that.

(It's hard to even argue that it's useful to allow differently even for `repr(Rust)`, so for `repr(C)` I see absolutely no reason we'd even consider something different.)

traviscross:
Agreed with @scottmcm it seems well settled, and I'm tempted to merge it too, but from a Reference maintainer point-of-view, it's a bit more straightforward to follow the program and lang FCP new (-ly-documented) language guarantees of this kind.

@rfcbot fcp merge lang

(I'll check @scottmcm's box.)

traviscross (approved):

tmandry:
If this weren't true we'd be in trouble :)

@rfcbot reviewed

traviscross:
cc @rust-lang/fls

ehuss (approved):
Thanks!

PR #2137, opened 2026-01-19
Denote examples better

nilslindemann:
By using `:` in the text and `> ` in the Markdown.

I replaced with a more distinct example in the first two `> ` blocks to denote better to the reader that these are examples.

[The way it is currently written in the reference](https://doc.rust-lang.org/stable/reference/#conventions) can create confusion. I looked awhile at it, trying to make sense why you would write examples for terms in italics too (= I parsed this sentence as meta language). Until I grasped that this sentence *is* an example (= the sentence is to be parsed as object language).

This is partly due to the fact that `> ` blocks are not rendered in a special way (at this position?) in the book. And further, the first example (the second paragraph of the first list item, which I have split into two examples) was missing the `> `s.

Please modify at will. Just remove the ambiguity between object and meta language.

### DIFF
--- a/src/introduction.md
+++ b/src/introduction.md
@@ -109,7 +109,7 @@ These conventions are documented here.
 
   All examples are written for the latest edition unless otherwise stated.
 
-* The grammar and lexical productions are described in the [Notation] chapter:
+* The grammar and lexical productions are described in the [Notation] chapter.
 
 r[example.rule.label]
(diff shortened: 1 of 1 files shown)


nilslindemann:
How this could be made even more explicit:

Wrap the specific examples [on that page](https://doc.rust-lang.org/stable/reference/#conventions) in a `<div class="how-it-looks"> ... </div>`.

Add some CSS rules at the end of the [reference.css](https://github.com/rust-lang/reference/blob/master/theme/reference.css):

```css
.how-it-looks {
    position:relative;
    border:1px solid var(--searchbar-border-color, silver);
    padding: 1em 2em;
}

.how-it-looks::before {
    content:"How it looks";
    color: var(--searchbar-border-color, grey);
    position:absolute;
    right:1.5em;
    top: -.8em;
    display:inline-block;
    background:white;
    padding: 0 0.5em;
    font-style: italic;
}

main > ul > li > .how-it-looks {
    margin-bottom: 1em;
}
```

Looks like this in my browser:

image

PR #2138, opened 2026-01-19
const-eval: explain the final-value-byte-provenance restriction

RalfJung:
This is the docs part for https://github.com/rust-lang/rust/pull/148967.

I had no idea where to put it. Putting it in "const-expr" is odd since that is generally a *compositional* notion (the subexpressions of a const expression must be const expressions), whereas the limitation I am describing here only applies *at the top level*, i.e. for the final value of a const/static initializer.

### Related

- https://github.com/rust-lang/rust/issues/148470
- https://github.com/rust-lang/rust/pull/148967
- https://github.com/rust-lang/rust/pull/144081

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -232,6 +232,12 @@ r[const-eval.const-expr.loop]
 r[const-eval.const-expr.if-match]
 * [if] and [match] expressions.
 
+r[const-eval.const-expr.final-value-provenance]
+The representation of the final value of a constant or static initializer must only contain provenance in whole-pointer groups: if a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
+
+If a byte in the representation of the final value is uninitialized, then it *may* end up having provenance, which can cause compilation to fail.
+As a quality-of-implementation concern, the compiler should only actually fail if the initializer copies or overwrites parts of a pointer and that memory ends up in the final value.
+
 r[const-eval.const-context]
 ## Const context
 [const context]: #const-context
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -264,6 +264,28 @@ const _: Pair = unsafe {
 > The bytes with provenance must form a complete pointer in the correct order. In the example above, the pointer is written at offset 20, but a pointer requires 8 bytes (assuming an 8-byte pointer). Four of those bytes fit in the `y` field; the rest extend into the padding at offset 24. When the fields are initialized, the `y` bytes get overwritten, leaving only a partial pointer (4 bytes) in the padding. These 4 bytes have provenance but don't form a complete pointer, causing compilation to fail.
 >
 > This restriction ensures that any bytes with provenance in the final value represent complete pointers. Binary formats such as ELF cannot represent pointer fragments, so the compiler cannot emit them in the final binary.
+>
+> Reversing the order of the pointer bytes also causes compilation to fail, even though all bytes are present:
+>
+> ```rust,compile_fail
+> # use core::mem::{self, MaybeUninit};
+> const PTR_WIDTH: usize = mem::size_of::<*const u8>();
+> const _: MaybeUninit<u64> = unsafe {
+> //       ^^^^^^^^^^^^^^^^
+> //       ERROR: Partial pointer in final value of constant.
+>     let mut m = MaybeUninit::<u64>::uninit();
+>     let ptr: *const u8 = &0;
+>     let ptr_bytes = &raw const ptr as *const MaybeUninit<u8>;
+>     // Write pointer bytes in reverse order.
+>     let dst: *mut MaybeUninit<u8> = m.as_mut_ptr().cast();
+>     let mut i = 0;
+>     while i < PTR_WIDTH {
+>         dst.add(i).write(ptr_bytes.add(PTR_WIDTH - 1 - i).read());
+>         i += 1;
+>     }
+>     m
+> };
+> ```
 
 > [!NOTE]
 > If a byte in the representation of the final value is uninitialized, then it *may* end up having provenance, which can cause compilation to fail. `rustc` tries (but does not guarantee) to only actually fail if the initializer copies or overwrites parts of a pointer and those bytes appear in the final value.
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -235,7 +235,7 @@ r[const-eval.const-expr.if-match]
 ## Const initializers
 
 r[const-eval.const-expr.final-value-provenance]
-The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
+The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of correctly-ordered bytes that form an entire pointer, compilation will fail.
 
 ```rust,compile_fail
 # use core::mem::MaybeUninit;
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -232,11 +232,13 @@ r[const-eval.const-expr.loop]
 r[const-eval.const-expr.if-match]
 * [if] and [match] expressions.
 
-r[const-eval.const-expr.final-value-provenance]
-The representation of the final value of a constant or static initializer must only contain provenance in whole-pointer groups: if a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
+## Const initializers
 
-If a byte in the representation of the final value is uninitialized, then it *may* end up having provenance, which can cause compilation to fail.
-As a quality-of-implementation concern, the compiler should only actually fail if the initializer copies or overwrites parts of a pointer and that memory ends up in the final value.
+r[const-eval.const-expr.final-value-provenance]
+The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
+
+> [!NOTE]
+> If a byte in the representation of the final value is uninitialized, then it *may* end up having provenance, which can cause compilation to fail. `rustc` tries (but does not guarantee) to only actually fail if the initializer copies or overwrites parts of a pointer and those bytes appear in the final value.
 
 r[const-eval.const-context]
 ## Const context
@@ -313,6 +315,7 @@ The types of a const function's parameters and return type are restricted to tho
 [const generic argument]: items/generics.md#const-generics
 [const generic parameters]: items/generics.md#const-generics
 [constant expressions]: #constant-expressions
+[constant initializer]: items.const
 [constants]:            items/constant-items.md
 [Const parameters]:     items/generics.md
 [dereference expression]: expr.deref
@@ -342,6 +345,7 @@ The types of a const function's parameters and return type are restricted to tho
 [promoted]:             destructors.md#constant-promotion
 [range expressions]:    expressions/range-expr.md
 [slice]:                types/slice.md
+[static initializer]: items.static.init
 [statics]:              items/static-items.md
 [Struct expressions]:   expressions/struct-expr.md
 [temporary lifetime extension]: destructors.scope.lifetime-extension
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -237,8 +237,61 @@ r[const-eval.const-expr.if-match]
 r[const-eval.const-expr.final-value-provenance]
 The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
 
+```rust,compile_fail
+# use core::mem::MaybeUninit;
+#
+#[repr(C, align(32))]
+struct Pair {
+    x: u128,             // Offset  0, 16 bytes.
+    y: MaybeUninit<u64>, // Offset 16,  8 bytes.
+                         // Offset 24,  8 bytes of padding.
+}
+
+const _: Pair = unsafe {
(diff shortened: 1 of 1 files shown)


src/const_eval.md
```diff
@@ -232,6 +232,13 @@ r[const-eval.const-expr.loop]
 r[const-eval.const-expr.if-match]
 * [if] and [match] expressions.
 
+r[const-eval.const-expr.final-value-provenance]
+The representation of the final value of a constant or static initializer must only contain provenance in whole-pointer groups:
+if a byte has provenance but is not part of an adjacent group of bytes that form an entire pointer, compilation will fail.
+
+If a byte in the representation of the final value is uninitialized, then it *may* end up having provenance, which can cause compilation to fail.
+As a quality-of-implementation concern, the compiler should only actually fail if the initializer copied or overwrite parts of a pointer and that memory ends up in the final value.
```
traviscross:
```suggestion
As a quality-of-implementation concern, the compiler should only actually fail if the initializer copies or overwrites parts of a pointer and that memory ends up in the final value.
```

(And we'll want to unwrap the lines.)

traviscross:
> I had no idea where to put it. Putting it in "const-expr" is odd since that is generally a _compositional_ notion (the subexpressions of a const expression must be const expressions), whereas the limitation I am describing here only applies _at the top level_, i.e. for the final value of a const/static initializer.

Good question.  We had put the restrictions on the final value of a const in the chapter on const items, but given that this applies to static items as well that'd be maybe a bit more annoying.

Maybe @ehuss will have a good idea on this.

src/const_eval.md
RalfJung: The lines aren't wrapped...?

traviscross: We'll remove the line break between lines 239 and 240 and between 236 and 237.

RalfJung: Oh, you don't want to break for colons? Okay, sure.

traviscross:
@RalfJung: I added some examples and admonitions with explanation and motivation.  These look right to you?

src/const_eval.md
```diff
@@ -232,6 +232,104 @@ r[const-eval.const-expr.loop]
...
+The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of correctly-ordered bytes that form an entire pointer, compilation will fail.
+
+```rust,compile_fail
+# use core::mem::MaybeUninit;
+#
+#[repr(C)]
+struct Pair {
+    x: u128,
+    y: MaybeUninit<u64>,
+    // 8 bytes of padding at offset 24.
```
RalfJung: That's only true if u128 has an alignment of 16 and u64 an alignment of 8.

src/const_eval.md
```diff
@@ -232,6 +232,104 @@ r[const-eval.const-expr.loop]
...
+    (*m.as_mut_ptr()).y = MaybeUninit::new(0);
+    // Now `m` contains a pointer fragment in the padding.
+    m.assume_init()
+};
+```
+
+> [!NOTE]
+> The bytes with provenance must form a complete pointer in the correct order.  In the example above, the pointer is written at offset 20, but it requires (on 64-bit platforms) 8 bytes.  Four of those bytes fit in the `y` field; the rest extend into the padding at offset 24.  When the fields are initialized, the `y` bytes get overwritten, leaving only a partial pointer (4 bytes) in the padding.  These 4 bytes have provenance but don't form a complete pointer, causing compilation to fail.
+>
+> This restriction ensures that any bytes with provenance in the final value represent complete, valid pointers.  The compiler cannot support pointer fragments because it would be unable to reason about them at compile time.
```
RalfJung: "valid" is a dangerous choice here. The pointer may well be out-of-bounds! That's fine as far as this particular restrictions is concerned.

RalfJung: > The compiler cannot support pointer fragments because it would be unable to reason about them at compile time.

This is incorrect. The compiler actually has no problem at all reasoning about them. However, typical binary formats like ELF do not support representing pointer fragments so we just can't produce a binary when they occur.

src/const_eval.md
```diff
@@ -232,6 +232,104 @@ r[const-eval.const-expr.loop]
...
+> # struct Pair {
+> #     x: u128,
+> #     y: MaybeUninit<u64>,
+> # }
+> const C: Pair = unsafe {
+> //    ^^^^^^^ ERROR: Partial pointer in final value of constant.
+>     let mut m = MaybeUninit::<Pair>::uninit();
+>     let ptr: *const u8 = &0;
+>     let ptr_bytes = &ptr as *const _ as *const MaybeUninit<u8>;
+>     // Write pointer bytes in reverse order into the padding.
```
RalfJung: I don't understand how padding causes the pointer bytes to be reversed...?

Generally this example seems unnecessarily complicated. You can just have a single field of type `MaybeUninit<u64>` (no padding) and store a byte-reversed pointer in there.

src/const_eval.md
```diff
@@ -232,6 +232,104 @@ r[const-eval.const-expr.loop]
...
+> E.g., `rustc` currently accepts this, even though the padding bytes are uninitialized:
+>
+> ```rust
+> # #[repr(C)]
+> # struct Pair { x: u128, y: u64 }
+> // The padding bytes are uninitialized.
+> const ALLOWED: Pair = Pair { x: 0, y: 0 };
+> ```
+>
+> Constant evaluation makes the details of typed copies observable: depending on whether a copy is performed field-by-field or as a memory-block copy, provenance in padding bytes might be discarded or preserved (both in the source and in the destination). Because the semantics of typed copies are not yet fully specified in Rust --- and to preserve the ability to change how they work in the future in constant evaluation (for example, to always set the padding bytes to uninitialized) --- the language allows the compiler to reject any initializer with an uninitialized padding byte. Since the compiler cannot currently guarantee that an uninitialized byte does not contain a pointer fragment without a full model of typed copies, this allowance is necessary to avoid relying on underspecified details of the language.
```
RalfJung: Even if we decide to go with the MiniRust semantics for typed copies, nothing about this changes. So it's not correct to say that this is "Because the semantics of typed copies are not yet fully specified in Rust".

src/const_eval.md
traviscross: Thanks, good point. Added `#[repr(C, align(32))]`.

src/const_eval.md
traviscross: Thanks, removed that.

traviscross: Yes; thanks, fixed.

src/const_eval.md
traviscross: Yes, thanks; I had meant to simplify this.  Done now.

src/const_eval.md
traviscross: Thanks.  Fixed.

traviscross (approved):
Looks good to me now, though I'm going to ask @RalfJung to have another look that everything is as it should be.

src/const_eval.md
```diff
@@ -232,6 +232,98 @@ r[const-eval.const-expr.loop]
 r[const-eval.const-expr.if-match]
 * [if] and [match] expressions.
 
+## Const initializers
+
+r[const-eval.const-expr.final-value-provenance]
+The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of correctly-ordered bytes that form an entire pointer, compilation will fail.
```
tshepang, line 238: Does _final value_ mean more than coercions? Should we use that term if so, or should we create a glossary entry maybe?

RalfJung: @tshepang  I don't understand the question. What does any of this have to do with coercions?

This is about the value computed by the const initializer expression, as opposed to some arbitrary value that occurs in the middle of computing said expression.

src/const_eval.md
```diff
@@ -232,6 +232,94 @@ r[const-eval.const-expr.loop]
...
+    // Initialize fields.
+    (*m.as_mut_ptr()).x = 0;
+    (*m.as_mut_ptr()).y = MaybeUninit::new(0);
+    // Now `m` contains a pointer fragment in the padding.
+    m.assume_init()
+};
+```
+
+> [!NOTE]
+> The bytes with provenance must form a complete pointer in the correct order. In the example above, the pointer is written at offset 20, but a pointer requires 8 bytes (assuming an 8-byte pointer). Four of those bytes fit in the `y` field; the rest extend into the padding at offset 24. When the fields are initialized, the `y` bytes get overwritten, leaving only a partial pointer (4 bytes) in the padding. These 4 bytes have provenance but don't form a complete pointer, causing compilation to fail.
```
RalfJung, line 264:
```suggestion
> The bytes with provenance must form a complete pointer in the correct order. In the example above, the pointer is written at offset 20, but a pointer requires 8 bytes (assuming a 64-bit target). Four of those bytes fit in the `y` field; the rest extend into the padding at offset 24. When the fields are initialized, the `y` bytes get overwritten, leaving only a partial pointer (4 bytes) in the padding. These 4 bytes have provenance but don't form a complete pointer, causing compilation to fail.
```

"a pointer requires 8 bytes (assuming an 8-byte pointer)" sounds too much like an obvious tautology.

src/const_eval.md
```diff
@@ -232,6 +232,94 @@ r[const-eval.const-expr.loop]
...
+
+const _: Pair = unsafe {
+//    ^^^^^^^ ERROR: Partial pointer in final value of constant.
+    let mut m = MaybeUninit::<Pair>::uninit();
+    // Store pointer that extends halfway into trailing padding.
+    m.as_mut_ptr().byte_add(20).cast::<&u8>().write_unaligned(&0);
+    // Initialize fields.
+    (*m.as_mut_ptr()).x = 0;
+    (*m.as_mut_ptr()).y = MaybeUninit::new(0);
+    // Now `m` contains a pointer fragment in the padding.
```
RalfJung, line 258: Do you think it would be useful to add some ASCII art to explain what happens?
```
      +-------------------------------------------------+-------------------------+-------------------------------+
Field | x (offsets 0..16)                               | y (offsets 16..24)      | <padding> (offsets 24..32)    |
      +-------------------------------------------------+-------------------------+-------------------------------+
Data  | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | 00 00 00 00 00 00 00 00 | <byte 4..8 of &0> UU UU UU UU |
      +-------------------------------------------------+-------------------------+-------------------------------+
```
Also, why is the `x` field even part of this example? It seems entirely unnecessary.

src/const_eval.md
```diff
@@ -232,6 +232,94 @@ r[const-eval.const-expr.loop]
...
+> const _: Pair = unsafe { // OK.
+>     let mut m = Pair { x: 0, y: 0 };
+>     # // Trigger failure if padding were not zeroed.
+>     # let ptr: *const u8 = &0;
+>     # let ptr_bytes = &raw const ptr as *const MaybeUninit<u8>;
+>     # let dst: *mut MaybeUninit<u8> = (&raw mut m).byte_add(31).cast();
+>     # dst.write(ptr_bytes.read());
+>     // Explicitly zero the padding.
+>     (&raw mut m).byte_add(24).cast::<u64>().write_unaligned(0);
+>     m
```
RalfJung, line 319: Technically speaking, the use of `m` here at the end incurs a typed copy which resets the padding back to "uninit".

With the hidden code, this example turns more into "how can I launder an arbitrary `Pair` value I got from somewhere to ensure it doesn't have nonsense it its padding", but that's not what the user sees in these docs.

Also, the more I think about it the more it makes me really uneasy that we'd recommend people write horrible code like this instead of the simple safe initializer above. There's no way I can imagine that initializer to actually cause a problem. You have to do "weird" things like reading/overwriting a part of a pointer value or doing a typed copy where the source memory has pointer bytes in places that the type doesn't store a pointer. It's non-trivial to fully precisely define this list, but even a vague and somewhat hand-wavy definition I think is better than people reading this and going "uh so should I use this `MaybeUninit` construction instead of normal safe Rust?".

src/const_eval.md
tshepang: Sorry for lacking clarity, was thinking *final value* is a technical term, guessing it could mean a value one gets after coercions, me lacking proper vocabulary/understanding.

Would it be good to have *final value* be a glossary entry, or should we use different terminology... because framing things like that maybe goes too much into the implementation details. Maybe that's ok, like these are details that can't be omitted. Thing is it was not clear to me why call it *final*, and I don't know what happens before *final*.

theemathas: Maybe this is about cases where a const item's path is used in an array repeat expression, and the const item needs to be coerced in some way before using it as the array, like in https://github.com/rust-lang/rust/issues/143671 and https://github.com/rust-lang/rust/issues/140123? I don't see how that could possibly matter though.

RalfJung: If anything the term here is "(final) value of a (const or static) initializer". There's no general concept of "final value".

src/const_eval.md
```diff
@@ -232,6 +232,94 @@ r[const-eval.const-expr.loop]
 r[const-eval.const-expr.if-match]
 * [if] and [match] expressions.
 
+## Const initializers
+
+r[const-eval.const-expr.final-value-provenance]
+The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of correctly-ordered bytes that form an entire pointer, compilation will fail.
```
tshepang, line 238: adding an explanation
```suggestion
The representation of the final value of a [constant][constant initializer] or [static initializer] must only contain bytes with provenance in whole-pointer groups. If a byte has provenance but is not part of an adjacent group of correctly-ordered bytes that form an entire pointer, compilation will fail.

> [!NOTE]
> The _final_ here indicates that the value is computed by the initializer expression, to make it clear that it's not some arbitrary value that occurs in the middle of computing said expression.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2138/879afe428308410239374948a6d66db9b9415a3f..57f47084d314b2edd05da11219a4ae77bbe14395)*

RalfJung:
This is indeed "waiting for author" but the author of most of the text here is @traviscross.

PR #2146, opened 2026-01-27
undefined behavior: add missing plural in `undefined.misaligned.ptr`

DanielEScherzer:
When discussing multiple projections, say "projections"


ehuss (approved):
Thanks!

PR #2147, opened 2026-01-28
Add section on method resolution regarding supertrait shadowing

Amanieu:
Stabilization PR: rust-lang/rust#148605
Tracking issue: rust-lang/rust#89151

### DIFF
--- a/src/expressions/method-call-expr.md
+++ b/src/expressions/method-call-expr.md
@@ -67,7 +67,7 @@ r[expr.method.ambiguous-target]
 If this results in multiple possible candidates, then it is an error, and the receiver must be [converted][disambiguate call] to an appropriate receiver type to make the method call.
 
 r[expr.method.supertrait-shadowing]
-As an exception to the above, if all candidates consist of trait methods and one candidate is a sub-trait of all other candidates then that candidate is selected instead of causing an ambiguity error.
+As an exception to the above, if all candidates consist of trait methods and one candidate is a subtrait of all other candidates then that candidate is selected instead of causing an ambiguity error.
 
 r[expr.method.receiver-constraints]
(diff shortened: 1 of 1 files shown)


ehuss:
@yaahc I am assuming the ambiguities regarding paths for this feature would fall under your work for type-dependent name resolution?

yaahc:
I believe you're correct, tho I'm not 100% sure. I've been viewing type-relative name res and method res as separate processes and haven't dug super deeply into `method/probe.rs` which is where [this logic](https://github.com/rust-lang/rust/pull/125782/files#diff-090eb0e88a58a0f9ee4138978029c8d73a05e77306555ab98f076bf24a6f35b9R1642-R1652) lives afaict. Earlier when I was digging through `resolve_fully_qualified_call` and saw that it invokes method probing in "Path Mode" which is a significantly abbreviated set of steps compared to method resolution I stopped digging deeper.

- https://github.com/rust-lang/rust/blob/main/compiler/rustc_hir_typeck/src/method/probe.rs#L256-L258
- https://github.com/rust-lang/rust/blob/main/compiler/rustc_hir_typeck/src/method/probe.rs#L407

Given that method resolution already has it's own section, my plan so far has been to document what type-relative resolution is and when it applies, and for the cases where it overlaps with method resolution I was going to document how it differs from method resolution rather than reiterate the applicable steps so we'll only have a single source of truth to maintain.

traviscross:
We talked about this in the lang-docs office hours with @yaahc.  She's going to touch base with @Amanieu about what we want to add here in terms of noting the rule on path-based resolution.  That will then later get picked up in her larger PR documenting everything here more comprehensively.

traviscross:
Thanks for the update.  As an editorial matter, we'll likely want to add some examples before approving and merging the PR.  What's here, though, is enough to resolve that lang concern, so I'll go do that.

PR #2148, opened 2026-01-30
Clarify that statics in generic context may not refer to the generics

meithecatte:


ehuss:
I believe this rule is already stated in https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.syntax.scope. Generics are not in scope for any item defined in a function body.

ehuss:
I'm going to close as this rule is already stated, and I don't think we want to duplicate it in each item.

PR #2157, opened 2026-02-03
const-eval.const-expr.field: make paragraph more clear

tshepang:
This also makes it more clear that this paragraph is about 3 distinct things

/me am not so sure about this, so this can act as request for clarification

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -62,7 +62,7 @@ r[const-eval.const-expr.field]
 * [Field] expressions.
 
 r[const-eval.const-expr.index]
-* Index expressions, [array indexing] or [slice] with a `usize`.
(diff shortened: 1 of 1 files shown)


ehuss:
Yea a little confusing, I think the original text is poorly worded.

I think what it is trying to say is that you can use an **index expression** as long as you are only indexing an **array** or **slice** type with a **usize**.

Should probably check if that is still correct, or if it has been relaxed at all since it was written in 2017 (https://github.com/rust-lang/reference/pull/13).

tshepang:
Am not sure how to check if the restriction is lifted, but when I try other types, I get this in the compile error...
> slice indices are of type `usize` or ranges of `usize`

And, in const context, ranges are not accepted as index type.

ehuss (approved):
Thanks!

PR #2158, opened 2026-02-04
expr.paren.evaluation: fix and make more simple

tshepang:
- I did not understand the first sentence
- The second sentence is now reduntant

/me not too sure about this one... not confident I understood what was said

### DIFF
--- a/src/expressions/grouped-expr.md
+++ b/src/expressions/grouped-expr.md
@@ -13,7 +13,7 @@ r[expr.paren.evaluation]
 Parenthesized expressions evaluate to the value of the enclosed operand.
 
 r[expr.paren.place-or-value]
-The enclosed operand can be either [a place expression or a value expression][place].
+A parenthesized expression is a [place expression][place] if the enclosed operand is a place expression, and is a value expression if the enclosed operand is a value expression.
 
 r[expr.paren.override-precedence]
 Parentheses can be used to explicitly modify the precedence order of subexpressions within an expression.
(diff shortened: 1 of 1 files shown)


src/expressions/grouped-expr.md
```diff
@@ -13,7 +13,7 @@ r[expr.paren.evaluation]
 Parenthesized expressions evaluate to the value of the enclosed operand.
 
 r[expr.paren.place-or-value]
-Unlike other expressions, parenthesized expressions are both [place expressions and value expressions][place]. When the enclosed operand is a place expression, it is a place expression and when the enclosed operand is a value expression, it is a value expression.
+The enclosed operand can be either [a place expression or a value expression][place].
```
ehuss: I think this isn't quite getting at the purpose of the rule. It is saying that the group expression itself is either a place or value expression based on whether its operand is a place or value expression. How about something like this:


```suggestion
A parenthesized expression is a [place expression][place] if the enclosed operand is a place expression, and is a value expression if the enclosed operand is a value expression.
```

tshepang: Your suggestion repeats the rule just before this one:
> Parenthesized expressions evaluate to the value of the enclosed operand.

Would you rather rules be this independent?

tshepang: hm, I think I am conflating `value` and expressions category`

ehuss (approved):
Thanks!

PR #2159, opened 2026-02-04
const-eval.const-context.outer-generics: make more clear/obvious

tshepang:
The wording took some shortcuts that I think are not worth it

### DIFF
--- a/src/const_eval.md
+++ b/src/const_eval.md
@@ -258,7 +258,7 @@ r[const-eval.const-context.block]
 * A [const block]
 
 r[const-eval.const-context.outer-generics]
-Const contexts that are used as parts of types (array type and repeat length expressions as well as const generic arguments) can only make restricted use of surrounding generic parameters: such an expression must either be a single bare const generic parameter, or an arbitrary expression not making use of any generics.
+Const contexts that are used as parts of types (array type expressions, array repeat length expressions, and const generic arguments) can only make restricted use of surrounding generic parameters: such an expression must either be a single bare const generic parameter, or an arbitrary expression not making use of any generics.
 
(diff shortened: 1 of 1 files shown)


src/const_eval.md
```diff
@@ -258,7 +258,7 @@ r[const-eval.const-context.block]
 * A [const block]
 
 r[const-eval.const-context.outer-generics]
-Const contexts that are used as parts of types (array type and repeat length expressions as well as const generic arguments) can only make restricted use of surrounding generic parameters: such an expression must either be a single bare const generic parameter, or an arbitrary expression not making use of any generics.
+Const contexts that are used as parts of types (array type expressions, array repeat length expressions, and const generic arguments) can only make restricted use of surrounding generic parameters: such an expression must either be a single bare const generic parameter, or an arbitrary expression not making use of any generics.
```
ehuss: I'm struggling a bit with how this is worded. The "used as parts of types" seems at odds with the parenthetical that says "array repeat length expressions", since that isn't a type. For example:

```rust
fn d<const N: usize>() {
    // Use of bare N in type and repeat expr is OK.
    let x: [i32; N] = [1; N];
    let y: _ = [1; N+1]; // ERROR: N may only be used as standalone arguments
}
```

How about we simplify it a little and turn it around:

```suggestion
Array type length expressions, array repeat length expressions, and const generic arguments are restricted in their use of outer generic parameters: such an expression must either be a single const generic parameter, or an expression that does not reference any generic parameters.
```

tshepang: not everything is clear to me, but this is a (very) clear improvement

ehuss (approved):
Thanks!

PR #2160, opened 2026-02-04
namespaces: Fix incorrectly listing enum VARIANTs as not living in a namespace.

bitcp:


ehuss:
I think the original is correct here. When it says "enum field", it means a field inside an enum variant. Something like:

```rust
enum E {
    V { x: i32 }   // field x
}
```

bitcp:
Sorry, I missed it. The page of "Enumeration" item heavily mentions "field".
Thanks for pointing out.

Issue #1857, opened 2025-06-13
[paths.expr.argument-order] is outdated


lcnr:
> The order of generic arguments is restricted to lifetime arguments, then type arguments, then const arguments, then equality constraints.

const and type arguments can now be in arbitrary order. This is already correct for parameters https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.syntax.decl-order. This restriction generally shouldn't exist. We parse generic arguments in arbitrary order and only error when checking whether the order matches the generic parameters of the item

PR #2164, opened 2026-02-10
Fix ordering of generic arguments

ehuss:
This was missed in https://github.com/rust-lang/reference/pull/1098 where the order of types and const generics was relaxed. See https://github.com/rust-lang/rust/pull/90207#issuecomment-958843546

Fixes https://github.com/rust-lang/reference/issues/1857


ehuss:
Will plan a different approach.

Issue #925, opened 2021-01-10
Remove general information from "Generic Functions"; put into Generics chapter.


Havvy:
A lot of the information in the Generic Functions section is true for most or all places generic parameters are used. As such, we should just have it in the generics chapter.

PR #2165, opened 2026-02-10
Rearrange the generics chapter

### LINKED ISSUE #1857: [paths.expr.argument-order] is outdated

ehuss:
This rearranges things so that there is a comprehensive "Generics" chapter under "Types" that covers the various aspects of generics. The general outline is:

- Generics moved out of "Items" and into "Types". Although generics are defined in items, it really belongs in the type system chapter, since generics themselves are not items.
- Moved where clauses to the "Trait and lifetime bounds" chapter because where clauses are more closely aligned with bounds.
- Created a "Generic arguments" section, and moved some of the generic argument rules there.
- Created dedicated sub-chapters for lifetimes, types, and const.

There's a little more reworking such as mentioned in https://github.com/rust-lang/reference/issues/925 about reworking the "Generic functions", but I didn't feel like doing more extensive rewriting at this time.

Fixes https://github.com/rust-lang/reference/issues/1857


ehuss:
I have pushed up some updates and I think this is now ready for review. I ultimately decided to go ahead and move each generic kind into a dedicated chapter, even though the current content is a bare-bones skeleton.

PR #2166, opened 2026-02-11
Fields must fit in the type, even for repr(Rust)

Darksonn:
This clarifies the `repr(Rust)` layout rules slightly by specifying that fields must fit in their struct. That is, the offset of a field plus its size must be less than the size of the struct.

**Accidental stabilization.** One question that needs to be addressed before we can merge this is the fact that we have guaranteed that fields of `repr(Rust)` do not overlap, even though one can imagine structs where this is not the case when the struct is uninhabited. For example:
```rs
// could this be size zero? it would violate sizeof<i32>() <= sizeof<Foo>()
struct Foo(i32, !);

// could the two integers both be at offset zero? it would violate the no-overlap guarantee
struct Foo(i32, i32, !);
```
So as part of this PR I am posing the following question to the lang team:

> Should we undo the guarantee about fields not overlapping if the struct has uninhabited fields?

In any case, I think we should definitely guarantee that fields fit in the struct for inhabited `repr(Rust)` structs.

Context: [#t-lang > ZST Uninhabited types with non-ZST fields @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/ZST.20Uninhabited.20types.20with.20non-ZST.20fields/near/572468369)

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -173,7 +173,7 @@ The only data layout guarantees made by this representation are those required f
 For enums, the above guarantees only apply to fields of inhabited variants.
 
 r[layout.repr.rust.layout.struct]
-For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+For [structs], it is further guaranteed that the fields do not overlap. That is, if we take the fields ordered by their offset, then for each field its offset plus its size is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
 
 Be aware that this guarantee does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields in the same struct.
 
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -168,6 +168,7 @@ The only data layout guarantees made by this representation are those required f
 
  1. The offset of a field is divisible by that field's alignment.
  2. The alignment of the type is at least the maximum alignment of its fields.
+ 3. For any field, its offset plus its size is at most the size of the type.
 
 r[layout.repr.rust.layout.struct]
 For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -166,20 +166,19 @@ The `Rust` representation is the default representation for nominal types withou
 r[layout.repr.rust.layout]
 The only data layout guarantees made by this representation are those required for soundness. These are:
 
- 1. The offset of a constructible field is divisible by that field's alignment.
- 2. The alignment of the type is at least the maximum alignment of its constructible fields.
- 3. For any constructible field, its offset plus its size is at most the size of the type.
+ 1. The offset of a field is divisible by that field's alignment.
+ 2. The alignment of the type is at least the maximum alignment of its fields.
+ 3. For any field, its offset plus its size is at most the size of the type.
 
-r[layout.repr.rust.layout.constructible]
-
-A field is considered constructible if it is possible to create a value of the type containing the field.
-
-For example, given `enum E { A { x: u32 }, B { y: u32, z: ! } }`, the field `x` is constructible because you can create the value `E::A { x: 0 }`, but the fields `y` and `z` are not constructible because the type of `z` is uninhabited, so it is impossible to create an `E::B` value.
+For enums, the above guarantees only apply to fields of inhabited variants.
 
 r[layout.repr.rust.layout.struct]
-For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
 
 Be aware that this guarantee does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields in the same struct.
+
+r[layout.repr.rust.layout.enum]
+For enums, a variant of the enum is said to be inhabited if all of its fields are [inhabited](glossary.html#inhabited). When an enum variant is inhabited, its fields are laid out according to the same guarantees as a struct. Otherwise, nothing is guaranteed about the fields. This allows the enum to be smaller than fields of uninhabited variants.
 
 r[layout.repr.rust.unspecified]
 There are no other guarantees of data layout made by this representation.
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -166,9 +166,15 @@ The `Rust` representation is the default representation for nominal types withou
 r[layout.repr.rust.layout]
 The only data layout guarantees made by this representation are those required for soundness. These are:
 
- 1. The offset of a field is divisible by that field's alignment.
- 2. The alignment of the type is at least the maximum alignment of its fields.
- 3. For any field, its offset plus its size is at most the size of the type.
+ 1. The offset of a constructible field is divisible by that field's alignment.
+ 2. The alignment of the type is at least the maximum alignment of its constructible fields.
+ 3. For any constructible field, its offset plus its size is at most the size of the type.
+
+r[layout.repr.rust.layout.constructible]
+
+A field is considered constructible if it is possible to create a value of the type containing the field.
+
+For example, given `enum E { A { x: u32 }, B { y: u32, z: ! } }`, the field `x` is constructible because you can create the value `E::A { x: 0 }`, but the fields `y` and `z` are not constructible because the type of `z` is uninhabited, so it is impossible to create an `E::B` value.
 
 r[layout.repr.rust.layout.struct]
 For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.


src/type-layout.md
```diff
@@ -167,14 +167,14 @@ r[layout.repr.rust.layout]
 The only data layout guarantees made by this representation are those required for soundness. They are:
 
  1. The fields are properly aligned.
- 2. The fields do not overlap.
+ 2. The fields do not overlap and fit in the struct.
```
RalfJung: The surrounding text is about " All user-defined composite types (structs, enums, and unions) " (`layout.repr.intro`). If some part of this text is meant to apply only to structs, that needs to be made clear upfront, not in the middle of a section.

Darksonn: Surely this section cannot apply to unions. It says the fields are not allowed to overlap.

RalfJung: It clearly shouldn't apply to unions, but I can't find anything in the wording that would prevent it from applying to unions.

RalfJung: Clearly if this was for structs only, the section identifier should make that explicit. But it is `layout.repr.rust.layout`, no "struct" anywhere.

teor2345: How about something like:
```suggestion
 2. The fields fit in the type:
     a. For structs, this means the fields do not overlap.
     b. For enums and unions, this means that fields from the same variant do not overlap (but fields from different variants can overlap).
```

I'm not sure if it's worth splitting enum variants and union fields into separate sentences.

RalfJung: unions do not have "variants".

Also we recently documented that union fields are all at offset 0: https://github.com/rust-lang/reference/pull/2128. EDIT: Ah, but just for `repr(C)`.

Darksonn: How about this?
```diff
 The only data layout guarantees made by this representation are those required for soundness. They are:
 
  1. The fields are properly aligned.
- 2. The fields do not overlap and fit in the struct.
+ 2. The fields fit in the type.
  3. The alignment of the type is at least the maximum alignment of its fields.
 
 r[layout.repr.rust.alignment]
 Formally, the first guarantee means that the offset of any field is divisible by that field's alignment.
 
 r[layout.repr.rust.field-storage]
-The second guarantee means that the fields can be ordered such that the offset plus the size of each field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+For structs, the second guarantee means that the fields can be ordered such that the offset plus the size of each field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+
+For enums, the same applies for each variant individually. That is, the fields of each variant of the enum may be ordered in the same way as the fields of a struct.
+
+For unions, fields are allowed to overlap, so all that is guaranteed is that for each field, its offset plus its size is less than or equal to the size of the union.
 
-Be aware that the second guarantee does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields in the same struct.
+Be aware that this does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields without overlapping with them.
 
 r[layout.repr.rust.unspecified]
 There are no other guarantees of data layout made by this representation.
```

RalfJung: > For structs, the second guarantee means that

This is very strange -- re-interpreting what "fits in" means depending on the kind of type? I would rather have two separate guarantees:
1. All fields fit in the type, i.e., the offset of each field plus its size does not exceed the size of the type.
2. All fields in a struct, and all fields within a single enum variant, do not overlap. (For zero-sized fields, this implies that the field is not contained within the range of another field.)

Either way this would make a guarantee about enums that has been quite contentious during the offset_of_enum discussion (and ultimately prevented stabilizing that feature) so I am not sure we're willing to commit to this.

Darksonn: > Either way this would make a guarantee about enums that has been quite contentious during the offset_of_enum discussion (and ultimately prevented stabilizing that feature) so I am not sure we're willing to commit to this.

Yes, but please keep this orthogonal.

To clarify, this PR is currently worded without an exception for uninhabited types, but as I wrote in the original PR description, we need to figure out whether there needs to be an exception for uninhabited types before we can merge anything in this PR, and that's why I lang-nominated this.

RalfJung:
> Should we undo the guarantee about fields not overlapping if the struct has uninhabited fields?

At least @scottmcm [seems to agree](https://github.com/rust-lang/lang-team/pull/346) that for structs we should guarantee that even uninhabited structs have space for all their fields (and presumably, the fields remain disjoint).

However, as currently worded the text is not just about structs but also about unions (where the no-overlap guarantee makes absolutely no sense) and enums. This is almost certainly also an accident. Not sure if it's best to discuss both of these at once -- I assume a PR that just rearranges the text to make it about structs only would not need much discussion.

src/type-layout.md
Darksonn: Alternate wording:
```diff
 The only data layout guarantees made by this representation are those required for soundness. They are:
 
  1. The fields are properly aligned.
- 2. The fields do not overlap and fit in the struct.
- 3. The alignment of the type is at least the maximum alignment of its fields.
+ 2. The alignment of the type is at least the maximum alignment of its fields.
+ 3. The fields fit in the type.
+ 4. Fields that can co-exist do not overlap.
 
 r[layout.repr.rust.alignment]
 Formally, the first guarantee means that the offset of any field is divisible by that field's alignment.
 
 r[layout.repr.rust.field-storage]
-The second guarantee means that the fields can be ordered such that the offset plus the size of each field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+The third guarantee means that for any field, the field's offset plus its size must be less than or equal to the size of the type.
+
+r[layout.repr.rust.field-overlap]
+Fields are said to overlap unless the size plus offset of one field is less than or equal to the offset of the other field. Fields are said to co-exist if the type can contain both fields at the same time, i.e. all fields of a struct co-exist, only fields of the same enum variant co-exist, and fields never co-exist in a union.
 
-Be aware that the second guarantee does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields in the same struct.
+Be aware that this does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields without overlapping with them.
 
 r[layout.repr.rust.unspecified]
 There are no other guarantees of data layout made by this representation.
```

Darksonn:
True, but I think we do want the other existing guarantees for union and enums:

1. The fields are properly aligned.
2. The alignment of the type is at least the maximum alignment of its fields.

Darksonn:
I filed a PR to fix the struct/union/enum issue:

https://github.com/rust-lang/reference/pull/2168

src/type-layout.md
RalfJung: > Fields are said to overlap unless the size plus offset of one field is less than or equal to the offset of the other field.

I'd prefer a positive definition instead of a negative one:
Two fields are said to overlap if the size plus offset of one field is strictly within the range of the other field.

Or use the opposite term:
Two fields are non-overlapping if the size plus offset of one field is less than or equal to the offset of the other field (`field1.offset + field1.size <= field2.offset || field2.offset + field2.size <= field1.offset`).

(It took me reading this multiple times until I realized you hid a disjunction here so I think the condition is worth spelling out formally, not just in English.)

Darksonn: Maybe the best is to keep the "you can order fields such that" wording ...

> Two fields are said to overlap if the size plus offset of one field is strictly within the range of the other field.

This definition does not work because it does not cover non-ZST fields of the same size and offset.

teor2345: Does using a niche for an enum discriminant change how we want to word the overlap guarantee?

I assume discriminants are not considered fields, but unless we say that somewhere, there could be ambiguity.

RalfJung: Yeah for the purpose of this we do not consider the discriminant to be a field.

Darksonn:
I'm unnominating this for now. I do not think it's useful for the lang team to spend time here before #2168 is merged and I can rebase this to make it less confusing.

@rustbot label -I-lang-nominated

Darksonn:
@rustbot ready
@rustbot label I-lang-nominated

The first commit contains the change I want to make. The second commit adds a caveat that may be necessary.

src/type-layout.md
```diff
@@ -166,8 +166,15 @@ The `Rust` representation is the default representation for nominal types withou
...
- 2. The alignment of the type is at least the maximum alignment of its fields.
+ 1. The offset of a constructible field is divisible by that field's alignment.
+ 2. The alignment of the type is at least the maximum alignment of its constructible fields.
+ 3. For any constructible field, its offset plus its size is at most the size of the type.
+
+r[layout.repr.rust.layout.constructible]
+
+A field is considered constructible if it is possible to create a value of the type containing the field.
+
+For example, given `enum E { A { x: u32 }, B { y: u32, z: ! } }`, the field `x` is constructible because you can create the value `E::A { x: 0 }`, but the fields `y` and `z` are not constructible because the type of `z` is uninhabited, so it is impossible to create an `E::B` value.
```
RalfJung: It seems easier to define this as
- all fields of a struct and union are constructible
- for an enum, fields of a variant are constructible if all fields of that variant are inhabited

Is there a definition of "inhabited" we can refer to?

RalfJung:
The PR title seems outdated now? The PR aims to make new guarantees not just for structs.

src/type-layout.md
Darksonn: Your proposed definition makes the guarantee apply to structs even if there are `!` fields. But if a struct has a field of type `!`, surely none of the fields are constructible? It seems like changing the guarantee is a better fix than redefining constructible to mean something less natural.

RalfJung: Oh, I didn't realize you meant to exclude structs with uninhabited fields. I think we do want to guarantee that fields are disjoint and fit in the type even for such structs -- that seems pretty much settled to me.

src/type-layout.md
```diff
@@ -166,8 +166,15 @@ The `Rust` representation is the default representation for nominal types withou
...
 
- 1. The offset of a field is divisible by that field's alignment.
- 2. The alignment of the type is at least the maximum alignment of its fields.
+ 1. The offset of a constructible field is divisible by that field's alignment.
+ 2. The alignment of the type is at least the maximum alignment of its constructible fields.
+ 3. For any constructible field, its offset plus its size is at most the size of the type.
+
+r[layout.repr.rust.layout.constructible]
+
+A field is considered constructible if it is possible to create a value of the type containing the field.
```
tmandry:
```suggestion
A field is considered constructible if it is possible to create a value containing the field.
```

Below the type is `E`; there is no type `E::B`. So the next line doesn't follow from the existing definition.

joshtriplett:
General vibes: the change to add the "constructible" concept seems great, and reserves enough space for things we *might* want to be able to do. We can then separately nail down when we do and don't want to mandate constructibility.

src/type-layout.md
Nadrieril: How about:
```suggestion
A field is considered constructible if it is part of an inhabited struct or enum variant, or part of a union.
```

scottmcm: I think this definition isn't quite right, because in the example `E` is the type and you can create a value of that type, which would mean that `y` is still "constructible" because it's talking about *type*.

Perhaps this should be talking about variants instead?  (At least in the compiler even structs have variants, just exactly one variant, so the reference might want to adopt that approach.)

Darksonn: You can create values of type `E`, but you can't create a value of type `E` *that contains the field `y`*.

Nadrieril: On second thought speaking in terms of individual fields feels weird, e.g. I could imagine:
```rust
enum Foo {
  A { x: u32 },
  B { x: u32, y: ! },
}
```
and I could imagine (I kinda want even) that we add a language feature whereby we can write `foo.x` because that field exists in all variants. Is `x` constructible then? Well no because `x` refers to two "real" fields.

So I'd rather say: a type has space for its inhabited variants, and an inhabited variant has space for its fields.

Darksonn: I think under this section of the reference, those xs *have* to be separate fields because otherwise a field might have multiple offsets, and this section talks about fields as if they have a single offset.

Darksonn:
I thought a bit more about it, and here is a different possible wording.

> The only data layout guarantees made by this representation are those required for soundness. These are:
> 
>  1. The offset of a field is divisible by that field's alignment.
>  2. The alignment of the type is at least the maximum alignment of its fields.
>  3. For any field, its offset plus its size is at most the size of the type.
> 
> For enums, the above guarantees only apply to fields of inhabited variants.
> 
> ### Structs
> 
> For structs, it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
> 
> Be aware that this guarantee does not imply that the fields have distinct addresses: zero-sized types may have the same address as other fields in the same struct.
> 
> ### Enums 
> 
> For enums, a variant of the enum is said to be inhabited if all of its fields are [inhabited](https://doc.rust-lang.org/reference/glossary.html#inhabited). When an enum variant is inhabited, its fields are laid out according to the same guarantees as a struct. Otherwise, nothing is guaranteed about the fields. This allows the enum to be smaller than fields of uninhabited variants.
 
This drops the 'constructible' wording, which I liked, but I think the above wording has several advantages:

1. Makes it much more clear that the uninhabited exception only applies to enums.
2. Defining 'inhabited' for enum variants avoids introducing new terminology.

scottmcm:
(Personally I'm a fan of not introducing new terminology, and leaning on the existing *inhabitedness* concept :+1:)

Darksonn:
I saw in the meeting notes from T-lang triage that there was discussion about why this PR has an exception for uninhabited enum variants, but not for uninhabited structs.

I was under the impression that T-lang had *already* decided that structs must have space for their fields even if the struct is uninhabited (hence https://github.com/rust-lang/lang-team/pull/346), and that is why I did not include an exception for structs.

I was also under the impression that T-lang did not want to promise the same for enum variants, and that this is the main obstacle for the set-discriminant proposals.

traviscross:
This PR guarantees that for all `repr(Rust)` structs, unions, and inhabited enum variants, each field must fit in the type.  It makes no guarantees about uninhabited enum variants (those with any uninhabited field).

The guarantee implies that `(u8, !)` can never be a ZST, fully closing that door.

This makes sense to me.  It does make me a bit sad, at some level, that we may end up with different guarantees for uninhabited structs and uninhabited enum variants.  That doesn't feel fundamentally necessary, but it may be practical.

Thanks @Darksonn for your work on this.

@rfcbot fcp merge lang

cc @rust-lang/opsem

zachs18:
> This PR guarantees that for all structs, unions, and inhabited enum variants, each field must fit in the type.

To clarify, this PR only guarantees this for structs/unions/inhabited-enum-variants with existing `repr`s, yes? (That is, this is already guaranteed for `repr(C)`, and this PR additionally guarantees it for `repr(Rust)`). It doesn't preclude a possible future `repr` that could *opt-in* to a layout optimization with uninhabited fields (e.g. something [like this](https://internals.rust-lang.org/t/pre-rfc-repr-collapse-uninhabited/23463))?

traviscross:
> To clarify, this PR only guarantees this for [`repr(Rust)`] structs/unions/inhabited-enum-variants...

Yes, thanks.  I've edited above, adding `repr(Rust)`, to clarify.

src/type-layout.md
```diff
@@ -168,12 +168,18 @@ The only data layout guarantees made by this representation are those required f
 
  1. The offset of a field is divisible by that field's alignment.
  2. The alignment of the type is at least the maximum alignment of its fields.
+ 3. For any field, its offset plus its size is at most the size of the type.
+
+For enums, the above guarantees only apply to fields of inhabited variants.
 
 r[layout.repr.rust.layout.struct]
-For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
+For [structs], it is further guaranteed that the fields do not overlap. That is, the fields can be ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
```
```suggestion
For [structs], it is further guaranteed that the fields do not overlap. That is, the fields are ordered such that the offset plus the size of any field is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
```
They actually are ordered, right? "can be ordered" is confusing to me.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2166/91f144460fcb93bed2b580887aa59c28d662a10b..b1f9d0596f54ca58d0e683a09cb6cd5a49335daa)*

Darksonn: Due to ZST the ordering may not be unique.

RalfJung: Yeah, that's true. I still think "are ordered" is more clear than "can be".

"can be" sounds like one *could* do this but the actual layout algorithm doesn't actually do this.

RalfJung:
```suggestion
For [structs], it is further guaranteed that the fields do not overlap. That is, if we take the fields ordered by their offset, then for each field its offset plus its size is less than or equal to the offset of the next field in the ordering, or for the last field, the size of the struct. The ordering does not have to be the same as the order in which the fields are specified in the declaration of the type.
```

scottmcm:
I like the new phrasing here (compared to last I remembered reading this) that re-uses *inhabitedness*.

@rfcbot reviewed

---

As I've written [elsewhere](https://github.com/rust-lang/lang-team/pull/346), the distinction between structs and enum variants here seems well-motivated to me because at runtime you'll have very few instances of partially-initialized uninhabited structs, so the cost of the extra space is tiny, whereas being able to "turn off" a variant of an enum could plausibly give broad wins since using *other* variants of the enum is presumably plausible.  And thus I wouldn't want to close the door *on that* for now.

joshtriplett:
This makes sense to me. I think in the future if we add `move` fields or similar mechanisms, we may allow those to not exist (since a `move` field doesn't have an offset or an address at all). But in the absence of mechanisms like that (which would also prevent you from writing the code that might break), let's go ahead and avoid footguns.

@rfcbot reviewed

tmandry:
@rfcbot concern consider interaction with enum variant types

As I wrote in https://github.com/rust-lang/lang-team/pull/346#issuecomment-4872290914:

> Widening the distinction between structs and enums might have directional implications for enum variant types (as discussed in e.g. https://github.com/rust-lang/lang-team/issues/122) and pattern types. We should accept this with an understanding of any doors we would be closing.

Darksonn:
I think it'd be reasonable to say that an enum variant type follows the layout rules of enum variants, rather than the layout rules of structs. Not sure I can put that in the reference now, though, as I don't believe we have enum variant types yet?

traviscross:
We talked about this in the lang meeting today without specific resolution.

RalfJung:
I don't think I understand the concern. The PR only makes guarantees that seem entirely uncontroversial to me. If that causes problems for enum variant types, then the enum variant types RFC will already have to deal with them. We wouldn't let that RFC say "oh actually structs with uninhabited fields are ZST now", would we?  And for enum variants where all fields are inhabited, we obviously *have* to give all those fields space.

So I am completely at loss about what there even is to discuss here, or which possible alternatives we might have to these guarantees.

PR #2167, opened 2026-02-12
dev-guide stabilization.md: add missing "not"

DanielEScherzer:
Presumably the intent was to say that information does not need to be duplicated, rather than saying that it needs to be duplicated.


ehuss (approved):
Thanks!

PR #2169, opened 2026-02-12
Document `UnsafeCell` special-casing in shared statics

madsmtm:
WIP on how the reference would change after https://github.com/rust-lang/rust/pull/152540 is stabilized.


src/items/static-items.md
```diff
@@ -126,7 +129,7 @@ fn bump_levels_safe() -> u32 {
 ```
 
 r[items.static.mut.sync]
```
madsmtm, line 131: Unsure about the name of this rule label now.

PR #2170, opened 2026-02-12
Use `clobber_abi`s corresponding to the called functions in `[asm.abi-clobbers.many]`'s example.

zachs18:
The example calls an `extern "sysv64" fn` and an `extern "win64" fn`, but only declares `clobber_abi("C")`, despite being in the section that says `clobber_abi` can be specified multiple times.

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -1009,7 +1009,8 @@ unsafe {
         sym foo,
         sym bar,
         out("rax") z,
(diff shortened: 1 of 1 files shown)


ehuss:
Makes sense to me. @Amanieu does this look good to you?

Amanieu (approved):
LGTM

ehuss (approved):
Thanks!

PR #2173, opened 2026-02-15
Switch to new range syntax

ehuss:
This changes the syntax for range repeat so that it handles inclusive and exclusive upper bounds with `..=` and `..`.

The old syntax was confusing. It used `..` for inclusive bound, but that's not how Rust syntax works. This changes it so that it uses `..=` for inclusive bounds to be consistent with Rust syntax.

There are some other options for range syntax that I considered:

- `{a,b}` which is the syntax used by most regex engines, and some parsers like Pest and Parsimonious.
- IETF ABNF and W3C EBNF uses `a*bexpr` where `a` and `b` are optional.
- Peg-rs uses `*<n,m>` where `n` and `m` are optional.
- Various languages use `:` (Python, Julia, Excel, etc.) or `..` (Rust, Kotlin, Swift, C#, F#, Zig, Perl, etc.) to represent ranges.

This will become more relevant when we switch the raw string literals to use a bounded range. We can't easily avoid the use of bounded repetition because of raw-string's bound of 255. Listing out 255 variants would be just too much, and it is convenient to avoid English-descriptive rules.

### DIFF
--- a/dev-guide/src/grammar.md
+++ b/dev-guide/src/grammar.md
@@ -66,7 +66,7 @@ RepeatPlusNonGreedy -> `+?`
 
 RepeatRange -> `{` Range? `..` Range? `}`
 
-RepeatRangeInclusive -> `{` Range `..=` Range `}`
+RepeatRangeInclusive -> `{` Range? `..=` Range `}`
 
 Range -> [0-9]+
 
@@ -140,7 +140,7 @@ The general format is a series of productions separated by blank lines. The expr
 | RepeatPlus | Expr+ | The preceding expression is repeated 1 or more times. |
 | RepeatPlusNonGreedy | Expr+? | The preceding expression is repeated 1 or more times without being greedy. |
 | RepeatRange | Expr{2..4} | The preceding expression is repeated between the range of times specified. Either bound can be excluded, which works just like Rust ranges. |
-| RepeatRangeInclusive | Expr{2..=4} | The preceding expression is repeated between the range of times specified (inclusive). |
+| RepeatRangeInclusive | Expr{2..=4} | The preceding expression is repeated between the inclusive range of times specified. The lower bound can be omitted. |
 
 ## Automatic linking
 
--- a/dev-guide/src/grammar.md
+++ b/dev-guide/src/grammar.md
@@ -81,7 +81,7 @@ Expr1 ->
     | Group
     | NegativeExpression
 
-Unicode -> `U+` [`A`-`Z` `0`-`9`]4..4
+Unicode -> `U+` [`A`-`Z` `0`-`9`]4..=4
 
 NonTerminal -> Name
 
--- a/dev-guide/src/grammar.md
+++ b/dev-guide/src/grammar.md
@@ -39,19 +39,34 @@ Sequence ->
         (` `* AdornedExpr)* ` `* Cut
       | (` `* AdornedExpr)+
 
-AdornedExpr -> ExprRepeat Suffix? Footnote?
+AdornedExpr -> Expr1 Quantifier? Suffix? Footnote?
 
 Suffix -> ` _` <not underscore, unless in backtick>* `_`
 
 Footnote -> `[^` ~[`]` LF]+ `]`
 
-ExprRepeat ->
-      Expr1 `?`
-    | Expr1 `*?`
-    | Expr1 `*`
-    | Expr1 `+?`
-    | Expr1 `+`
-    | Expr1 `{` Range? `..` Range? `}`
+Quantifier ->
+      Optional
+    | Repeat
+    | RepeatNonGreedy
+    | RepeatPlus
+    | RepeatPlusNonGreedy
+    | RepeatRange
+    | RepeatRangeInclusive
+
+Optional -> `?`
+
+Repeat -> `*`
+
+RepeatNonGreedy -> `*?`
+
+RepeatPlus -> `+`
+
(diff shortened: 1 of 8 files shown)


traviscross (commented):
Looks great. A few things:

1. In `grammar.md`, the `Unicode` production on line 69 still uses `4..4`, which under the new exclusive semantics would be the empty range rather than "exactly 4". Should that be `4..=4`?

2. The phrasing "a to b repetitions (exclusive) of x" and "a to b repetitions (inclusive) of x" reads a bit awkwardly — the parenthetical interrupts the noun phrase. Would something like "a to b (exclusive) repetitions of x" or moving the qualifier to the end work better?

3. The grammar rule for `RepeatRangeInclusive` requires both bounds (`Range` \`..=\` `Range`), but the parser appears to accept `{..=4}` (no lower bound). Is this intentional? If `{..=b}` should be valid, the grammar might want `Range?` for the lower bound.

4. Minor: in `render_railroad.rs`, the comment for the compound range decomposition arm (around diff line 381) seems to have a missing backtick and `{e..b-(a-1)}` where the min value should be 1.

One edge case I noticed in the railroad renderer: `e{a..a}` with `HalfOpen` (e.g., `e{2..2}`) — an exclusive range where min equals max — decomposes into copies of `e` rather than rendering as empty. The range `2..2` is empty, so this seems like it should produce nothing. This is probably academic (nobody would write such a range), but might be worth a parser validation or a special case in the renderer.

dev-guide/src/grammar.md
traviscross: We need to change this.

src/notation.md
```diff
@@ -16,7 +16,8 @@ The following notations are used by the *Lexer* and *Syntax* grammar snippets:
 | x<sup>?</sup>     | `pub`<sup>?</sup>             | An optional item                          |
 | x<sup>\*</sup>    | _OuterAttribute_<sup>\*</sup> | 0 or more of x                            |
 | x<sup>+</sup>     |  _MacroMatch_<sup>+</sup>     | 1 or more of x                            |
-| x<sup>a..b</sup>  | HEX_DIGIT<sup>1..6</sup>      | a to b repetitions of x                   |
+| x<sup>a..b</sup>  | HEX_DIGIT<sup>1..6</sup>      | a to b repetitions (exclusive) of x       |
+| x<sup>a..=b</sup> | HEX_DIGIT<sup>1..=6</sup>     | a to b repetitions (inclusive) of x       |
```
traviscross: Perhaps it'd be better to put the parenthetical earlier, "a to be (inclusive) repetitions".

tools/mdbook-spec/src/grammar/render_railroad.rs
```diff
@@ -168,38 +189,81 @@ fn render_expression(expr: &Expression, cx: &RenderCtx, stack: bool) -> Option<B
...
                     Box::new(r)
                 }
-                // Treat `e{a..}` as `e{a-1..a-1} e{1..}` and `e{a..b}` as
-                // `e{a-1..a-1} e{1..b-(a-1)}`, and treat `e{x..x}` for some
-                // `x` as a sequence of `e` nodes of length `x`.
-                ExpressionKind::RepeatRange(e, Some(a @ 2..), b) => {
+                // Treat:
+                // - `e{a..}` as `e{0..a-1} e{1..}`
+                // - `e{a..=b}` as `e{0..a-1} e{1..=b-(a-1)}`
+                // - `e{a..b} as `e{0..a-1} {e..b-(a-1)}`
```
traviscross:
```suggestion
                // - `e{a..b}` as `e{0..a-1} {e..b-(a-1)}`
```

Missing a closing backtick.

tools/grammar/src/parser.rs
```diff
@@ -428,24 +428,37 @@ impl Parser<'_> {
         })
     }
 
-    /// Parse `{a..}` | `{..b}` | `{a..b}` after expression.
+    /// Parse `{a..b}` | `{a..=b}` after expression.
```
traviscross, line 431: In testing, the renderer still seems to support `{..b}` and `{..=b}`.

dev-guide/src/grammar.md
```diff
@@ -39,19 +39,34 @@ Sequence ->
...
+
+RepeatNonGreedy -> `*?`
+
+RepeatPlus -> `+`
+
+RepeatPlusNonGreedy -> `+?`
+
+RepeatRange -> `{` Range? `..` Range? `}`
+
+RepeatRangeInclusive -> `{` Range `..=` Range `}`
```
traviscross: We seem to support `{..=b}` -- which is what I'd expect.  Just not `{a..=}` or `{..=}`.

tools/mdbook-spec/src/grammar/render_railroad.rs
```diff
@@ -168,38 +189,81 @@ fn render_expression(expr: &Expression, cx: &RenderCtx, stack: bool) -> Option<B
...
-                        e.clone(),
-                        Some(1),
-                        b.map(|x| x - (a - 1)),
-                    )));
+                    es.push(Expression::new_kind(ExpressionKind::RepeatRange {
+                        expr: e.clone(),
+                        min: Some(1),
+                        max: b.map(|x| x - (a - 1)),
+                        limit: *limit,
+                    }));
```
traviscross, line 267: We're rendering `x{a..a}` incorrectly; we're emitting `x` one time instead of zero times.

traviscross (approved):

PR #2174, opened 2026-02-15
block-expr: add new rule expr.block.result-value

DanielEScherzer:
Document that the value of a block expression is based on the final operand (if present), just like the type.

### DIFF
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -42,25 +42,42 @@ When evaluating a block expression, each statement, except for item declaration 
 r[expr.block.result]
 Then the final operand is executed, if given.
 
-r[expr.block.type]
-The type of a block is the type of its final operand; if that operand is omitted, the type is the [unit type], unless the block [diverges][expr.block.diverging], in which case it is the [never type].
-
-r[expr.block.result-value]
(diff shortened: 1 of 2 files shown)


DanielEScherzer:
Discovered while working on a conference talk about rust, I couldn't find any rule saying that a block expression had a value equal to the final operand, so I added such a rule
Not quite sure how to phrase the part about diverging blocks - on the playground the following compiles with just some warnings about unreachable code:
```rust
fn main () {
    let result: String = { panic!("test") };
    let result: bool = { panic!("test") };
    let result: Vec<String> = { panic!("test") };
}
```

also, rule name could be improved, I originally was going to use `expr.block.value` but that is an existing rule
> Blocks are always [value expressions](https://doc.rust-lang.org/nightly/reference/expressions.html#place-expressions-and-value-expressions) and evaluate the last operand in value expression context.

@rustbot label A-expressions

DanielEScherzer:
From the review policy

> If this would make a new guarantee about the language, this needs to go through the lang team to be accepted (unless the lang team has clearly accepted this guarantee elsewhere). Ask @traviscross if at all unsure about any of these.

Not sure if this is a new guarantee or just documenting a long-standing reality, @traviscross does this need to go through the lang team?

traviscross (approved):

traviscross:
We talked about this on the lang-docs call today.  Revised as we discussed on the call.  With that, look good to me.  Thanks @DanielEScherzer for the PR.

traviscross (approved):

PR #2176, opened 2026-02-17
lifetime-elision.md: add some missing periods

DanielEScherzer:
To the end of the `lifetime-elision.trait-object.containing-type-unique` and `lifetime-elision.trait-object.containing-type-explicit` rules, both of which have full sentences.


ehuss (approved):
Thanks!

PR #2177, opened 2026-02-17
add c-variadic function definitions

folkertdev:
I think this has all of the raw material, but needs polishing.

Here is a draft of the stabilization report, for additional context: https://hackmd.io/@Q66MPiW4T7yNTKOCaEb-Lw/S1iI3WIwZg

Tracking issue: https://github.com/rust-lang/rust/issues/44930

Stabilization:

- https://github.com/rust-lang/rust/pull/155697

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -395,6 +395,20 @@ A c-variadic functions cannot be `async`
 r[items.fn.c-variadic.const]
 A c-variadic functions cannot be `const`
 
+r[items.fn.c-variadic.platform-support]
+Some ABIs do not support c-variadic function definitions. The compiler errors in this case.
+
+```
+error: the `bpfel` target does not support c-variadic functions
+  --> $DIR/not-supported.rs:23:31
+   |
+LL | unsafe extern "C" fn variadic(_: ...) {}
+   |                               ^^^^^^
+```
+
+r[items.fn.c-variadic.dyn-compat]
+When a trait method is c-variadic, the trait is no longer dyn-compatible.
+
 r[items.fn.attributes]
 ## Attributes on functions
 
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -83,7 +83,7 @@ r[items.fn.params.self-restriction]
 Functions with a self parameter may only appear as an [associated function] in a [trait] or [implementation].
 
 r[items.fn.params.varargs]
-A parameter with the `...` token indicates a [c-variadic function]. The variadic parameter may have an optional identifier, such as `args: ...`.
+A parameter with the `...` token indicates a [c-variadic function], and may only be used as the last parameter. In an [`extern` block] the c-variadic parameter may have an optional identifier, such as `args: ...`, in a [c-variadic function definition] the identifier is mandatory.
 
 r[items.fn.body]
 ## Function body
@@ -340,27 +340,25 @@ A *c-variadic* function accepts a variable argument list `pat: ...` as its final
 
 ```rust
 unsafe extern "C" fn example(ap: ...) -> f64 {
-    unsafe { ap.arg::<f64>() }
+    unsafe { ap.next_arg::<f64>() }
 }
 ```
 
 This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
 
 > [!WARNING]
-> Passing an unexpected number of arguments or arguments of unexpected type to a variadic function may lead to [undefined behavior][undefined].
+> Passing an unexpected number of arguments or arguments of unexpected type to a c-variadic function may lead to [undefined behavior][undefined].
 
-r[items.fn.c-variadic.variadic-parameter-type]
+r[items.fn.c-variadic.c-variadic-parameter-type]
 The type of `pat` in the function body is [`VaList`].
 
 r[items.fn.c-variadic.desugar-brief]
 A c-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
 
-
-
 ```rust
 // Source
 unsafe extern "C" fn example(mut ap: ...) -> i32 {
-    unsafe { ap.arg::<i32>() }
+    unsafe { ap.next_arg::<i32>() }
 }
 ```
 
@@ -368,15 +366,11 @@ is roughly equivalent to:
 
 ```rust
 # use std::ffi::VaList;
-# use std::mem::MaybeUninit;
-# fn va_start() {}
 // Desugared
 unsafe extern "C" fn example() -> i32 {
-    let mut storage = MaybeUninit::<VaList<'_>>::uninit();
-    va_start(storage.as_mut_ptr()); // Initializes the VaList.
-    let mut ap: &mut VaList<'_> = ap.assume_init_mut();
+    let mut ap: VaList<'_> = /* compiler initializes the VaList */;
 
-    unsafe { ap.arg::<i32>() }
+    unsafe { ap.next_arg::<i32>() }
 
     va_end(ap)
 }
@@ -389,7 +383,7 @@ r[items.fn.c-variadic.ffi-compatibility]
 The rust [`VaList`] is ABI-compatible with the C `va_list` type.
 
 r[items.fn.c-variadic.abi]
-Only `extern "C"` and `extern "C-unwind"` functions can accept a variable argument list.
+Only `extern "C"` and `extern "C-unwind"` function defintions can accept a variable argument list.
 
 r[items.fn.c-variadic.safety]
 Only `unsafe` functions can accept a variable argument list.
@@ -403,7 +397,7 @@ A c-variadic functions cannot be `const`
 r[items.fn.c-variadic.platform-support]
 Some ABIs do not support c-variadic function definitions. The compiler errors in this case.
 
-```
+```text
 error: the `bpfel` target does not support c-variadic functions
   --> $DIR/not-supported.rs:23:31
    |
@@ -412,7 +406,7 @@ LL | unsafe extern "C" fn variadic(_: ...) {}
 ```
 
 r[items.fn.c-variadic.dyn-compat]
-When a trait method is c-variadic, the trait is no longer dyn-compatible.
+When a trait method is c-variadic, the trait is no longer [dyn-compatible].
 
 r[items.fn.attributes]
 ## Attributes on functions
@@ -510,3 +504,4 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [`extern` block]: external-blocks.md
 [zero-sized]: glossary.zst
 [`VaList`]: std::ffi::VaList
+[dyn-compatible]: traits.md#dyn-compatibility
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -349,6 +349,24 @@ This parameter stands in for an arbitrary number of arguments that may be passed
 > [!WARNING]
 > Passing an unexpected number of arguments or arguments of unexpected type to a c-variadic function may lead to [undefined behavior][undefined].
 
+[items.fn.c-variadic.stable-targets]
+Support for c-variadic function definitions is stable on the following architectures:
+
+- x86 and x86-64
+- ARM
+- AArch64 and Arm64EC
+- RISC-V (except when using the ilp32e ABI)
+- LoongArch
+- s390x
+- PowerPC and PowerPC64
+- AmdGpu and Nvptx64
+- wasm32 and wasm64
+- csky
+- xtensa
+- hexagon
+- sparc64
+- mips
+
 r[items.fn.c-variadic.c-variadic-parameter-type]
 The type of `pat` in the function body is [`VaList`].
 
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -350,7 +350,36 @@ r[items.fn.c-variadic.parameter-type]
 The type of `pat` in the function body is [`VaList<'_>`].
 
 r[items.fn.c-variadic.lifetime]
-The lifetime of a `VaList` is that of the function call that created it.
+A C-variadic function definition is implicitly generic over the lifetime of its variadic parameter, as if the parameter had type `VaList<'x>` for a fresh, unnameable lifetime `'x`. Because the function must be valid for any such lifetime, the `VaList` cannot be proved to outlive any caller-provided lifetime (and so cannot escape the call) and no caller-provided lifetime can be proved to outlive it.
+
+```rust,compile_fail
+# use core::ffi::VaList;
+fn b_outlives_a<'a, 'b: 'a>(_: &mut VaList<'a>, _: &mut &'b mut u8) {}
+unsafe extern "C" fn f(mut r: &mut u8, mut ap: ...) {
+    b_outlives_a(&mut ap, &mut r); // ERROR: Lifetime may not live long enough.
+}
+```
+
+```rust,compile_fail
+# use core::ffi::VaList;
+fn a_outlives_b<'a: 'b, 'b>(_: &mut VaList<'a>, _: &mut &'b mut u8) {}
+unsafe extern "C" fn f(mut r: &mut u8, mut ap: ...) {
+    a_outlives_b(&mut ap, &mut r); // ERROR: Lifetime may not live long enough.
+}
+```
+
+> [!NOTE]
+> This is different than if the data were a stack variable: any caller-provided lifetime can be proved to outlive a borrow of a callee stack variable.
+>
+> ```rust
+> struct MockVaList<'data>(&'data u8);
+> fn b_outlives_a<'a, 'b: 'a>(_: &mut MockVaList<'a>, _: &mut &'b mut u8) {}
+> unsafe extern "C" fn f(mut r: &mut u8) {
+>     let data = 0;
+>     let mut ap = MockVaList(&data);
+>     b_outlives_a(&mut ap, &mut r); // OK.
+> }
+> ```
 
 r[items.fn.c-variadic.desugar-brief]
 A C-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
@@ -383,6 +412,9 @@ unsafe extern "C" fn example() -> i32 {
     }
 }
 ```
+
+> [!NOTE]
+> In an actual C-variadic function definition, the lifetime in `VaList<'_>` is different from what this code would suggest. See [items.fn.c-variadic.lifetime].
 
 r[items.fn.c-variadic.next-arg-safety]
 Calling `VaList::next_arg` to read an argument of type `T` is only safe if all of the following conditions are satisfied:
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -332,6 +332,69 @@ Note that this behavior is a consequence of the desugaring to a function that re
 
 Unsafe is used on an async function in precisely the same way that it is used on other functions: it indicates that the function imposes some additional obligations on its caller to ensure soundness. As in any other unsafe function, these conditions may extend beyond the initial call itself -- in the snippet above, for example, the `unsafe_example` function took a pointer `x` as argument, and then (when awaited) dereferenced that pointer. This implies that `x` would have to be valid until the future is finished executing, and it is the caller's responsibility to ensure that.
 
+r[items.fn.c-variadic]
+## C-variadic functions
+
+r[items.fn.c-variadic.intro]
+A *c-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
+
+```rust
+unsafe extern "C" fn example(arg0: i32, ap: ...) { }
+```
+
+This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
+
+> [!WARNING]
+> Passing an unexpected number of arguments or arguments of unexpected type to a variadic function may lead to [undefined behavior][undefined].
+
+r[items.fn.async.desugar-brief]
+A c-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
+
+
+
+```rust
+// Source
+unsafe extern "C" fn example(mut ap: ...) -> i32 {
+    unsafe { ap.arg::<i32>() }
+}
+```
+
+is roughly equivalent to:
+
+```rust
+# use std::ffi::VaList;
+# use std::mem::MaybeUninit;
+# fn va_start() {}
+// Desugared
+unsafe extern "C" fn example() -> i32 {
+    let mut storage = MaybeUninit::<VaList<'_>>::uninit();
+    va_start(storage.as_mut_ptr()); // Initializes the VaList.
+    let mut ap: &mut VaList<'_> = ap.assume_init_mut();
+
+    unsafe { ap.arg::<i32>() }
+
+    va_end(ap)
+}
+```
+
+r[items.fn.c-variadic.lifetime]
+The lifetime of a `VaList` is that of the function that created it. Hence, the `VaList` value can never outlive the function that created it.
+
+r[items.fn.c-variadic.ffi-compatibility]
+The rust [`VaList`] is ABI-compatible with the C `va_list` type.
+
+r[items.fn.c-variadic.abi]
+Only `extern "C"` and `extern "C-unwind"` functions can accept a variable argument list.
+
+r[items.fn.c-variadic.safety]
+Only `unsafe` functions can accept a variable argument list.
+
+r[items.fn.c-variadic.async]
+A c-variadic functions cannot be `async`
+
+r[items.fn.c-variadic.const]
+A c-variadic functions cannot be `const`
+
 r[items.fn.attributes]
 ## Attributes on functions
 
@@ -427,3 +490,4 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [variadic function]: external-blocks.md#variadic-functions
 [`extern` block]: external-blocks.md
 [zero-sized]: glossary.zst
+[`VaList`]: std::ffi::VaList
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -83,7 +83,30 @@ r[items.fn.params.self-restriction]
 Functions with a self parameter may only appear as an [associated function] in a [trait] or [implementation].
 
 r[items.fn.params.varargs]
-A parameter with the `...` token indicates a [C-variadic function] and may only be used as the last parameter. In an [`extern` block], the C-variadic parameter may have a pattern, such as `args: ...`, and in a [C-variadic function definition], the pattern is mandatory.
+A parameter with the `...` token indicates a [C-variadic function] and may only be used as the last parameter. In a function declaration in an [`extern` block], the C-variadic parameter may have a pattern, such as `ap: ...`, and in a [C-variadic function] definition or C-variadic associated function declaration in a trait definition, the pattern is mandatory.
+
+```rust
+unsafe extern "C" {
+    unsafe fn f1(...);
+    unsafe fn f2(ap: ...);
+}
+
+unsafe extern "C" fn f3(ap: ...) {}
+
+trait Tr {
+    unsafe extern "C" fn f4(ap: ...);
+}
+```
+
+```rust,compile_fail
+unsafe extern "C" fn f(...) {} // ERROR: Missing pattern.
+```
+
+```rust,compile_fail
+trait Tr {
+    unsafe extern "C" fn f(...); // ERROR: Missing pattern.
+}
+```
 
 r[items.fn.body]
 ## Function body
@@ -637,7 +660,6 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [implementation]: implementations.md
 [value namespace]: ../names/namespaces.md
 [C-variadic function]: items.fn.c-variadic.intro
-[C-variadic function definition]: items.fn.c-variadic
 [`extern` block]: external-blocks.md
 [`VaList<'_>`]: lang-types.va-list
 [`VaList`]: lang-types.va-list
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -422,7 +422,14 @@ r[items.fn.c-variadic.abi]
 Only `extern "C"` and `extern "C-unwind"` function definitions can accept a variable argument list.
 
 r[items.fn.c-variadic.safety]
-Only `unsafe` functions can accept a variable argument list.
+When a variable argument list is used in the signature:
+
+- Function definitions must be `unsafe`.
+- Function declarations within trait definitions must be `unsafe`.
+- Function declarations in `extern` blocks may be `safe`.
+
+> [!NOTE]
+> For `safe` function declarations in an `extern` block, see the warning in [items.extern.variadic].
 
 r[items.fn.c-variadic.async]
 A c-variadic function cannot be `async`.
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -83,7 +83,7 @@ r[items.fn.params.self-restriction]
 Functions with a self parameter may only appear as an [associated function] in a [trait] or [implementation].
 
 r[items.fn.params.varargs]
-A parameter with the `...` token indicates a [variadic function], and may only be used as the last parameter of an [external block] function. The variadic parameter may have an optional identifier, such as `args: ...`.
+A parameter with the `...` token indicates a [c-variadic function]. The variadic parameter may have an optional identifier, such as `args: ...`.
 
 r[items.fn.body]
 ## Function body
@@ -339,7 +339,9 @@ r[items.fn.c-variadic.intro]
 A *c-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
 
 ```rust
-unsafe extern "C" fn example(arg0: i32, ap: ...) { }
+unsafe extern "C" fn example(ap: ...) -> f64 {
+    unsafe { ap.arg::<f64>() }
+}
 ```
 
 This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
@@ -347,7 +349,10 @@ This parameter stands in for an arbitrary number of arguments that may be passed
 > [!WARNING]
 > Passing an unexpected number of arguments or arguments of unexpected type to a variadic function may lead to [undefined behavior][undefined].
 
-r[items.fn.async.desugar-brief]
+r[items.fn.c-variadic.variadic-parameter-type]
+The type of `pat` in the function body is [`VaList`].
+
+r[items.fn.c-variadic.desugar-brief]
 A c-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
 
 
@@ -501,7 +506,7 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [associated function]: associated-items.md#associated-functions-and-methods
 [implementation]: implementations.md
 [value namespace]: ../names/namespaces.md
-[variadic function]: external-blocks.md#variadic-functions
+[c-variadic function]: external-blocks.md#variadic-functions
 [`extern` block]: external-blocks.md
 [zero-sized]: glossary.zst
 [`VaList`]: std::ffi::VaList
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -339,15 +339,26 @@ r[items.fn.c-variadic.intro]
 A *C-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
 
 ```rust
-unsafe extern "C" fn example(mut ap: ...) -> f64 {
+unsafe extern "C" fn f(mut ap: ...) -> f64 {
     unsafe { ap.next_arg::<f64>() }
 }
+```
+
+```rust,compile_fail
+unsafe extern "C" fn f(ap: ..., _: ()) {} // ERROR: `...` must be last.
 ```
 
 This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
 
 r[items.fn.c-variadic.parameter-type]
 The type of `pat` in the function body is [`VaList<'_>`].
+
+```rust
+# use core::ffi::VaList;
+unsafe extern "C" fn f(ap: ...) {
+    let _: VaList<'_> = ap;
+}
+```
 
 r[items.fn.c-variadic.lifetime]
 A C-variadic function definition is implicitly generic over the lifetime of its variadic parameter, as if the parameter had type `VaList<'x>` for a fresh, unnameable lifetime `'x`. Because the function must be valid for any such lifetime, the `VaList` cannot be proved to outlive any caller-provided lifetime (and so cannot escape the call) and no caller-provided lifetime can be proved to outlive it.
@@ -356,7 +367,7 @@ A C-variadic function definition is implicitly generic over the lifetime of its 
 # use core::ffi::VaList;
 fn b_outlives_a<'a, 'b: 'a>(_: &mut VaList<'a>, _: &mut &'b mut u8) {}
 unsafe extern "C" fn f(mut r: &mut u8, mut ap: ...) {
-    b_outlives_a(&mut ap, &mut r); // ERROR: Lifetime may not live long enough.
+    b_outlives_a(&mut ap, &mut r); // ERROR: May not live long enough.
 }
 ```
 
@@ -364,7 +375,7 @@ unsafe extern "C" fn f(mut r: &mut u8, mut ap: ...) {
 # use core::ffi::VaList;
 fn a_outlives_b<'a: 'b, 'b>(_: &mut VaList<'a>, _: &mut &'b mut u8) {}
 unsafe extern "C" fn f(mut r: &mut u8, mut ap: ...) {
-    a_outlives_b(&mut ap, &mut r); // ERROR: Lifetime may not live long enough.
+    a_outlives_b(&mut ap, &mut r); // ERROR: May not live long enough.
 }
 ```
 
@@ -385,7 +396,7 @@ r[items.fn.c-variadic.desugar-brief]
 A C-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
 
 ```rust
-unsafe extern "C" fn example(mut ap: ...) -> i32 {
+unsafe extern "C" fn f(mut ap: ...) -> i32 {
     unsafe { ap.next_arg::<i32>() }
 }
 ```
@@ -401,7 +412,7 @@ Roughly desugars to:
 use core::intrinsics::{va_arg, va_end};
 // `va_start` is magic and has no intrinsic.
 fn va_start(ap: *mut VaList<'_>) { /* magic */ }
-unsafe extern "C" fn example() -> i32 {
+unsafe extern "C" fn f() -> i32 {
     unsafe {
         let mut ap: MaybeUninit<VaList<'_>> = MaybeUninit::uninit();
         va_start(ap.as_mut_ptr());
@@ -445,8 +456,32 @@ Examples of incompatible types are:
 r[items.fn.c-variadic.abi-compatibility]
 [`VaList`] is ABI compatible with the C `va_list` type.
 
+```rust
+# use core::ffi::{c_char, c_int, VaList};
+unsafe extern "C" {
+    // The C `vprintf` function is:
+    //
+    //     int vprintf(const char *format, va_list ap);
+    //
+    unsafe fn vprintf(fmt: *const c_char, ap: VaList<'_>) -> c_int;
+}
+
+unsafe extern "C" fn print(fmt: *const c_char, ap: ...) -> c_int {
+    // The `VaList` is passed directly to the C function.
+    unsafe { vprintf(fmt, ap) }
+}
+```
+
 r[items.fn.c-variadic.abi]
 Only `extern "C"` and `extern "C-unwind"` function definitions can accept a variable argument list.
+
+```rust,compile_fail
+unsafe fn f(ap: ...) {} // ERROR: Not supported.
+```
+
+```rust,compile_fail
+unsafe extern "sysv64" fn f(ap: ...) {} // ERROR: Not supported.
+```
 
 r[items.fn.c-variadic.safety]
 When a variable argument list is used in the signature:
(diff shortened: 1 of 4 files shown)


src/items/functions.md
```diff
@@ -332,6 +332,69 @@ Note that this behavior is a consequence of the desugaring to a function that re
 
 Unsafe is used on an async function in precisely the same way that it is used on other functions: it indicates that the function imposes some additional obligations on its caller to ensure soundness. As in any other unsafe function, these conditions may extend beyond the initial call itself -- in the snippet above, for example, the `unsafe_example` function took a pointer `x` as argument, and then (when awaited) dereferenced that pointer. This implies that `x` would have to be valid until the future is finished executing, and it is the caller's responsibility to ensure that.
 
+r[items.fn.c-variadic]
+## C-variadic functions
+
+r[items.fn.c-variadic.intro]
+A *c-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
```
folkertdev: for c-variadic definitions only `pat: ...` is accepted semantically.

plain `...` is currently parsed, the `varargs_without_pattern` lint is meant to eventually disallow it. This syntax can only be used as an input to macros, when a bare `...` makes it past macro expansion, that will emit an error.

I didn't mention this here, but maybe we should: we follow C23 in that `pat: ...` may be the only argument. Earlier versions of C required at least one standard argument before the `...`.

src/items/functions.md
```diff
@@ -332,6 +332,69 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+```rust
+unsafe extern "C" fn example(arg0: i32, ap: ...) { }
+```
+
+This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
+
+> [!WARNING]
+> Passing an unexpected number of arguments or arguments of unexpected type to a variadic function may lead to [undefined behavior][undefined].
```
folkertdev: copied from 

https://doc.rust-lang.org/reference/items/external-blocks.html?highlight=variadic#r-items.extern.variadic

basically, the responsibility for passing valid arguments is on the caller.

src/items/functions.md
```diff
@@ -332,6 +332,69 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+// Desugared
+unsafe extern "C" fn example() -> i32 {
+    let mut storage = MaybeUninit::<VaList<'_>>::uninit();
+    va_start(storage.as_mut_ptr()); // Initializes the VaList.
+    let mut ap: &mut VaList<'_> = ap.assume_init_mut();
+
+    unsafe { ap.arg::<i32>() }
+
+    va_end(ap)
+}
```
folkertdev: `va_start` and `va_end` are C functions/concepts. We could instead handwave and say

```rust
let mut ap: VaList<'_> = /* ... */; // Initializes the VaList.
```

the `va_end` is called by the `VaList` `Drop` implementation.

src/items/functions.md
```diff
@@ -332,6 +332,83 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+```rust
+unsafe extern "C" fn example(arg0: i32, ap: ...) { }
+```
+
+This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
+
+> [!WARNING]
+> Passing an unexpected number of arguments or arguments of unexpected type to a variadic function may lead to [undefined behavior][undefined].
+
```
ehuss, line 378: Maybe it would be good to have a rule near the top here that explains the type of the variadic parameter is a `VaList`?

src/items/functions.md
```diff
@@ -332,6 +332,83 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+r[items.fn.async.desugar-brief]
+A c-variadic function definition is roughly equivalent to a function operating on a [`VaList`].
```
ehuss: What is "roughly" about it?

src/items/functions.md
```diff
@@ -332,6 +332,83 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
 Unsafe is used on an async function in precisely the same way that it is used on other functions: it indicates that the function imposes some additional obligations on its caller to ensure soundness. As in any other unsafe function, these conditions may extend beyond the initial call itself -- in the snippet above, for example, the `unsafe_example` function took a pointer `x` as argument, and then (when awaited) dereferenced that pointer. This implies that `x` would have to be valid until the future is finished executing, and it is the caller's responsibility to ensure that.
 
+r[items.fn.c-variadic]
+## C-variadic functions
+
+r[items.fn.c-variadic.intro]
+A *c-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
+
+```rust
+unsafe extern "C" fn example(arg0: i32, ap: ...) { }
```
ehuss: It might be helpful for this initial example to show extracting a value from `ap`.

ehuss (commented):
I think the rule `items.fn.params.varargs` will need to be updated, since it says it can only be used in an external block.



*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/3b30d76b02ae9878af6f3efc851f1a480a9a0b04..dc9cfa70404e9930911754d17e6c30c4a798d781)*

src/items/functions.md
ehuss: Is anything holding back changing `varargs_without_pattern` to `report_in_deps: true`?

src/items/functions.md
folkertdev: In how arguments are actually passed, these signatures are (very) different

```rust 
fn foo(ap: ...) { /* ... */ }
fn bar(ap: VaList) { /* ... */ }
```

Those two functions could have exactly the same body though.

src/items/functions.md
folkertdev: Not as far as I know. I can make that change an nominate for T-lang.

folkertdev: Based on https://github.com/rust-lang/rust/pull/143619#issuecomment-3133157725 there was only one actual impacted crate https://crates.io/crates/binrw. The issue was fixed in https://github.com/jam1garner/binrw/issues/342, however there has not been a release since.

So, if we made this warn in dependencies there would be nothing that users of that crate could do. I'll ping for a release with the fix.

joshtriplett: Looks like the releases happened: https://github.com/jam1garner/binrw/issues/342#issuecomment-3994700268

folkertdev: Right, PR is up at https://github.com/rust-lang/rust/pull/154599

folkertdev:
The `varargs_without_pattern` is now reported in dependencies (on nightly).

I've pushed some tweaks, I'm now planning to submit a stabilization PR in coming days, so if you could look at this again that would be helpful.

ehuss:
It looks like from https://github.com/rust-lang/rust/pull/155974 that there is an intent to make this unavailable on certain targets. I don't think we've ever done something like that, and I'm not sure how we're going to document that. I suppose there will be a rule. It will need to be careful to distinguish that it is a compile error during validation. Unfortunately we don't define that as a specific phase in the reference, so I'm not sure how we should approach that.

folkertdev:
I don't think this is very different from https://doc.rust-lang.org/beta/unstable-book/language-features/asm-experimental-arch.html. Consequently the page on inline assembly specifies https://doc.rust-lang.org/nightly/reference/inline-assembly.html?highlight=assemb#r-asm.stable-targets.

So we could have

> Support for c-variadic function definitions is stable on the following architectures:
>
> * x86 and x86-64
> * arm
> * aarch64 and arm64ec
> * ...
>
> The compiler will emit an error if `...` is used in a function definition on an unsupported target.

Also spirv and bpf just fundamentally do not support this feature, so there you'd always get an error.

src/items/functions.md
```diff
@@ -83,7 +83,7 @@ r[items.fn.params.self-restriction]
 Functions with a self parameter may only appear as an [associated function] in a [trait] or [implementation].
 
 r[items.fn.params.varargs]
-A parameter with the `...` token indicates a [variadic function], and may only be used as the last parameter of an [external block] function. The variadic parameter may have an optional identifier, such as `args: ...`.
+A parameter with the `...` token indicates a [c-variadic function], and may only be used as the last parameter. In an [`extern` block] the c-variadic parameter may have an optional identifier, such as `args: ...`, in a [c-variadic function definition] the identifier is mandatory.
```
tshepang: does this preserve what was meant
```suggestion
A parameter with the `...` token indicates a [c-variadic function], and may only be used as the last parameter. In an [`extern` block], the c-variadic parameter may have an optional identifier, such as `args: ...`, and in a [c-variadic function definition], the identifier is mandatory.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/3b30d76b02ae9878af6f3efc851f1a480a9a0b04..63b268ba59e853ed2affa6f59b776a3838114f81)*

src/items/functions.md
```diff
@@ -332,6 +332,100 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+```text
+error: the `bpfel` target does not support c-variadic functions
+  --> $DIR/not-supported.rs:23:31
+   |
+LL | unsafe extern "C" fn variadic(_: ...) {}
+   |                               ^^^^^^
+```
+
+r[items.fn.c-variadic.dyn-compat]
+When a trait method is c-variadic, the trait is no longer [dyn-compatible].
```
ehuss: Can we move this to the dyn-compatibility section? I assume it would just be added to the list in "Dispatchable functions must…".

I realize there's a bit of a challenge whether to document or link things in multiple places when they cover multiple concepts (dyn-compatibility and c-variadic in this case). In this particular case, I think it's fine to consolidate the rules in the dyn-compatibility list. For example, when describing Self or async or receivers, they don't mention their impact on dyn-compatibility.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..9880b05750bbcd9138815f8992cb67d15af66a25)*

src/items/functions.md
```diff
@@ -332,6 +332,100 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+    let mut ap: VaList<'_> = /* compiler initializes the VaList */;
+
+    unsafe { ap.next_arg::<i32>() }
+
+    va_end(ap)
+}
+```
+
+r[items.fn.c-variadic.lifetime]
+The lifetime of a `VaList` is that of the function that created it. Hence, the `VaList` value can never outlive the function that created it.
```
tmandry: Bringing the discussion from https://github.com/rust-lang/rust/pull/155697#issuecomment-4580889889 here because I think it's getting too in the weeds for the stabilization PR.

@RalfJung said:

> Miri's UB detection around `va_end` is about catching use-after-free, where you do something like `drop(transmute_copy(&my_va_list))` and then keep using `my_va_list`.
> [`tests/ui/consts/const-eval/c-variadic-fail.rs`](https://github.com/rust-lang/rust/blob/c560774cf35e52fc140048edacaa9a9bc10f624e/tests/ui/consts/const-eval/c-variadic-fail.rs) is a nice test to look at for various bits of UB that Miri/const-eval catch (and that should be all the UB there is).

What makes that case UB, but not the `forget` case? I don't see anything in the reference PR or stabilization report that defines it. Is it a more general rule about not using a value that isn't Copy after using `transmute_copy` or something like that? If it is a more general thing, why is there a Drop impl specific to `VaList`?

The question I want the reference to answer is: What semantic difference does it make to call the destructor of a VaList (and, by extension, what are its implications on the correctness of programs)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..9880b05750bbcd9138815f8992cb67d15af66a25)*

programmerjake: apparently there are targets (not currently supported by Rust) where `VaList::drop` needs to free memory allocated on the heap or similar, so that's why `VaList: Drop` and why checking that you didn't drop it twice or use it after dropping it is useful. Because it's just freeing memory (on the unusual target(s)) or doing nothing (on all existing Rust targets), it's safe to `forget` it, just like it's safe to `forget(my_box)`.

We mostly decided that we aren't supporting the requirement for the `va_end` call to be in the same function as `va_start`, since AFAICT that's just in C because some crazy platform(s) (cfront?) opened a code block in `va_start` and closed the block in `va_end`, so if you didn't balance them your program wouldn't compile since the `{}` aren't balanced.

RalfJung: The way we model varargs in Miri is that
- We have a new kind of "magic allocation" that represents a list of variadic arguments. The allocation is "magic" in that the list is represented as native AM state, there aren't actually any bytes here that the program can read or write. This is similar to how we already represent functions (the allocations that function pointers point to) and vtables.
- When calling a vararg function, a new such allocation gets created, and a pointer to it is stored in the VaList.
- va_next takes such a pointer, deallocates the allocation, takes the first element off the list, and then creates a new list which has a new pointer and updates the VaList to store that new pointer.
- va_copy duplicates such an allocation.
- va_end frees such an allocation.

This model is intended to operationally capture *some* of the UB that C ascribes to variadics.  But there are some things C makes UB that we say are allowed. Specifically C says it is UB to not call va_end, and it is UB to call va_end in a different function than where the VaList got created. We decided to not have all the UB (and we'll require backends to honor that of course) because we'd have to significantly redesign the API to make it sound under such a model. For LLVM this requirement is trivially upheld because va_end at the moment is a NOP on all targets we care about -- but the rules are designed in a way that we can later support targets where it is not a NOP. We just can't support targets where it's UB to actually leak a VaList. We doubt there'll ever be such a target, where would the UB even come from?

So, this is what I'd say the underlying spec for vararg UB is in Rust. The goal was to capture as much as possible of the UB in C. The cases that are UB in C but allowed in Rust satisfy all of the following conditions:
- No target LLVM currently supports actually makes them UB.
- We were unable to imagine a target that would make them UB (except for C silliness like needing the curly braces to match up).
- Making those cases UB would force us to make the safe API a lot less ergonomic and might even force some usecases to go unsafe that we otherwise would make safe. Specifically, we'd have to remove `Clone` from `VaList`, and replace it with something like `fn with_dup(self: &VaList, f: impl FnOnce(&mut VaList))` -- that's the only way we could enforce the VaList to actually be freed. Having such an unergonomic API just so that we could add entirely theoretical UB that we don't actually need seemed like a bad idea.

programmerjake: > * We were unable to imagine a target that would make them UB.

I wouldn't quite say that, but it definitely falls beyond what we expect even the craziest platforms to need:
Having a separate global stack (not the normal function call stack) that varargs are pushed on by `va_start` and popped off by `va_end`.

src/items/functions.md
folkertdev: Sure, done

src/items/functions.md
tmandry: Thanks, that all makes sense. I assume the magic allocation is tied to the call stack in some way, such that referencing it after the function where it is created returns is UB?

I think the reference text should explain this to some degree. FFI boundaries are one place where it helps to be a lot more precise. Right now the text talks about UB from using the wrong number of args, but not from calling drop or letting it fall out of scope.

RalfJung: > I assume the magic allocation is tied to the call stack in some way, such that referencing it after the function where it is created returns is UB?

No. The magic allocation is basically a list of pointers to the actual arguments which live in regular allocations, one per argument, that are created on stack push (just like regular arguments: every stack variable is its own little allocation). *Those* allocations get freed on stack pop, just like the allocations for regular arguments and other stack variables.

Furthermore, the one magic allocation that gets created on pop is freed by the destructor of the implicit `VaList`, which is guaranteed to be executed since the user only ever sees an `&mut VaList`.

So if you clone a `VaList` and then call it again after the stack frame it refers to got popped, then accessing the `VaList` works fine but you get UB because the memory it tries to load the argument from is gone.

> I think the reference text should explain this to some degree. FFI boundaries are one place where it helps to be a lot more precise. 

Note that everything I said is relevant even for pure Rust programs, which can both define and call variadic functions now.

folkertdev: > Right now the text talks about UB from using the wrong number of args, but not from calling drop or letting it fall out of scope.

This type isn't special in this regard. Using a `Vec` or `File` after they've been dropped is just as invalid. For those types (allocations, file descriptors) Miri already has infrastructure to track when they are "live", for c-variadic functions we needed new logic to provide the same kinds of tracking.

So can you be more precise with what you want to see here?

tmandry: > No. The magic allocation is basically a list of pointers to the actual arguments which live in regular allocations, one per argument, that are created on stack push (just like regular arguments: every stack variable is its own little allocation). _Those_ allocations get freed on stack pop, just like the allocations for regular arguments and other stack variables.

Ah, I see. That explains what was bothering me about the Drop impl.

Repeating back: You consider the VaList itself as kind of a builtin resource managed by the AM, and that's because a list of pointers never truly exists in memory (or if it does, it does at the platform level). That list is what gets dropped, not the argument values themselves, which explains the difference between dropping and forgetting the list. That difference is also mostly irrelevant because `next_arg` returns by value.

Thanks. My intuition is now consistent with everything I've read.

RalfJung: Yes that sounds right. Thanks for digging through this until we're sure we're talking about the same thing. :)

src/items/functions.md
```diff
@@ -332,6 +332,97 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+```rust
+unsafe extern "C" fn example(ap: ...) -> f64 {
+    unsafe { ap.next_arg::<f64>() }
+}
+```
+
+This parameter stands in for an arbitrary number of arguments that may be passed by the caller.
+
+> [!WARNING]
+> Passing an unexpected number of arguments or arguments of unexpected type to a c-variadic function may lead to [undefined behavior][undefined].
```
traviscross: This nonnormative note is standing in for more than it probably should.  The stabilization report in https://github.com/rust-lang/rust/pull/155697 discusses compatibility rules at some length.  Let's document these here.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..a42bf50e3cb8f7512cfb495de2b3923d0155739f)*

folkertdev: you're referring to this section? Note that this is already documented in the `std` docs https://doc.rust-lang.org/nightly/core/ffi/struct.VaList.html#method.next_arg.

In practice the callee can't actually know whether there is another argument (of a compatible type), so the safety requirements must be put on the caller instead.

---

Calling `VaList::next_arg` to read an argument of type `T` is only safe if all of the following conditions are satisfied:

- There is another c-variadic argument to read.
- The actual type of the argument `U` is compatible with `T` (as defined below).
- If `U` and `T` are both integer types, then the value passed by the caller must be
representable in both types.

Types `T` and `U` are compatible when:

- `T` and `U` are the same type (up to free lifetimes). 
- `T` and `U` are integer types of the same size.
- `T` and `U` are both pointers, and their target types are compatible.
- `T` is a pointer to `c_void` and `U` is a pointer to `i8` or `u8`, or vice versa.

---

traviscross: Yes, these are what I mean.  These amount to language guarantees, so it'd be good to find a way to also encode these in the Reference.  Probably `VaList` should also go in *Special Types and Traits*.

folkertdev: I've added both, and added some positive and negative examples to help clarify the rules.

src/items/functions.md
```diff
@@ -332,6 +332,120 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+- LoongArch
+- s390x
+- PowerPC and PowerPC64
+- AmdGpu and Nvptx64
+- wasm32 and wasm64
+- csky
+- xtensa
+- hexagon
+- sparc64
+- mips
```
tgross35: Nit: This mixes lowercase with the arches' name casing (C-SKY, Xtensa, Hexagon, SPARC64, MIPS, Wasm). Also some platforms are explicit about both 32- and 64-bit (x86, ppc) while others aren't (loongarch, mips)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..1faf0231046396c06eac9185cae02f4ffdadc45b)*

src/items/functions.md
tgross35, line 1: Is there somewhere it should be mentioned that variadics are allowed in traits?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..1faf0231046396c06eac9185cae02f4ffdadc45b)*

src/items/functions.md
```diff
@@ -332,6 +332,120 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+- `*const fn(&'static ())` and `*const for<'a> fn(&'a ())`, these types are not equal up to free lifetimes.
+
+r[items.fn.c-variadic.ffi-compatibility]
+The rust [`VaList`] is ABI-compatible with the C `va_list` type.
+
+r[items.fn.c-variadic.abi]
+Only `extern "C"` and `extern "C-unwind"` function defintions can accept a variable argument list.
+
+r[items.fn.c-variadic.safety]
+Only `unsafe` functions can accept a variable argument list.
```
tgross35: Should we explicitly note that `safe fn` in `extern` blocks is not allowed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..1faf0231046396c06eac9185cae02f4ffdadc45b)*

tgross35: Wait sorry, `safe fn` is not rejected https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=9f9f18b6df2d4265cf19ac7355546874 and I remember seeing that this is fine if the arguments are not consumed. This seems like a mismatch? Why can't you define a not-unsafe function that doesn't use the args?

folkertdev: Allowing `safe` was discussed in https://github.com/rust-lang/rust/pull/141733#issuecomment-2934372081. 

Also, `...` in extern blocks is already stable, and is not influenced by stabilizing c-variadic definitions.

Omitting the `unsafe` is left as a future option (see the "Defining safe C-variadic functions" section of the stabilization report). 

> At the moment we don't have a good reason to add this [that is, allowing a safe function when using `_: ...`] behavior. It is completely backwards compatible, so if a need arises in the future we can revisit this.

tgross35: Thanks for the info, makes sense

src/items/functions.md
folkertdev: I get the sense that this sort of thing is usually left implicit , but up to our reference maintainers.

src/items/functions.md
folkertdev: Looking at it again, I'm not sure this desugaring answers more questions than it raises? Because really it is not a desugaring, we're hooking into a bunch of custom compiler magic to make this work. We could provide an equivalent C function perhaps to illustrate what parts the rust compiler handles automatically in comparison.

src/items/functions.md
folkertdev: It is based on the list of stable architectures for `asm!`, I've tried to correct the names.

src/items/functions.md
folkertdev: i.e. it is equivalent to

```c
// Assuming C23, so `...` as the only argument is accepted.
int example(...) {
    va_list ap; 
    va_start(ap);

    int x = va_arg(ap, int);

    va_end(ap);
    return x;
}
```

folkertdev (commented):
The merge of the stabilization PR is now only blocked on the additions to the reference here (I'm not sure that is really needed, previously features have been merged with a reference PR that was "far enough along").

I've tried to address comments here as much possible now.

I'd be happy to look at this together in the reference office hours if they are at a somewhat european-friendly time.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..66cd6b058eeba1024f7c5d364261103b510379ed)*

src/items/functions.md
```diff
@@ -332,6 +332,120 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+}
+```
+
+is roughly equivalent to:
+
+```rust
+# use std::ffi::VaList;
+// Desugared
+unsafe extern "C" fn example() -> i32 {
+    let mut ap: VaList<'_> = /* compiler initializes the VaList */;
```
traviscross:
```suggestion
<!-- ignore: conceptual desugaring -->
```rust,ignore
# use std::ffi::VaList;
// Desugared
unsafe extern "C" fn example() -> i32 {
    let mut ap: VaList<'_> = /* compiler initializes the VaList */;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..66cd6b058eeba1024f7c5d364261103b510379ed)*

src/items/functions.md
```diff
@@ -332,6 +332,120 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+- LoongArch 32-bit and 64-bit
+- s390x
+- PowerPC and PowerPC64
+- AmdGpu and Nvptx64
+- Wasm32 and Wasm64
+- C-SKY
+- Xtensa
+- Hexagon
+- SPARC64
+- MIPS
```
traviscross: We talked about it in the meeting (with @folkertdev), probably this list makes more sense at the bottom.

As is done here, listing the ones supported (rather than the other way around) makes sense.  As @folkertdev pointed out, the bar to adding a new tier-3 architecture is not that high, and we want the list to remain correct.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..66cd6b058eeba1024f7c5d364261103b510379ed)*

src/items/functions.md
traviscross: We talked about this in the meeting.  We'll also add in here the `va_start` and the `MaybeUninit` as that will make more clear what's happening, match the `va_end`, and be what someone coming from C would expect to see for explaining this.

src/items/functions.md
```diff
@@ -332,6 +332,120 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+// Source
+unsafe extern "C" fn example(mut ap: ...) -> i32 {
+    unsafe { ap.next_arg::<i32>() }
+}
+```
+
+is roughly equivalent to:
+
+```rust
+# use std::ffi::VaList;
```
traviscross:
```suggestion
# use core::ffi::VaList;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..66cd6b058eeba1024f7c5d364261103b510379ed)*

src/items/functions.md
```diff
@@ -332,6 +332,123 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
 Unsafe is used on an async function in precisely the same way that it is used on other functions: it indicates that the function imposes some additional obligations on its caller to ensure soundness. As in any other unsafe function, these conditions may extend beyond the initial call itself -- in the snippet above, for example, the `unsafe_example` function took a pointer `x` as argument, and then (when awaited) dereferenced that pointer. This implies that `x` would have to be valid until the future is finished executing, and it is the caller's responsibility to ensure that.
 
+r[items.fn.c-variadic]
+## C-variadic functions
+
+r[items.fn.c-variadic.intro]
+A *c-variadic* function accepts a variable argument list `pat: ...` as its final parameter.
+
+```rust
+unsafe extern "C" fn example(ap: ...) -> f64 {
```
traviscross:
```suggestion
unsafe extern "C" fn example(mut ap: ...) -> f64 {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..256b0f242c7e400865b1f068b13194a8c80a09e4)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+The rust [`VaList`] is ABI-compatible with the C `va_list` type.
+
+r[items.fn.c-variadic.abi]
+Only `extern "C"` and `extern "C-unwind"` function defintions can accept a variable argument list.
+
+r[items.fn.c-variadic.safety]
+Only `unsafe` functions can accept a variable argument list.
+
+r[items.fn.c-variadic.async]
+A c-variadic functions cannot be `async`.
```
ehuss:
```suggestion
A c-variadic function cannot be `async`.
```

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+r[items.fn.c-variadic.const]
+A c-variadic functions cannot be `const`.
```
ehuss:
```suggestion
A c-variadic function cannot be `const`.
```

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+[items.fn.c-variadic.stable-targets]
```
traviscross:
```suggestion
r[items.fn.c-variadic.stable-targets]
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..256b0f242c7e400865b1f068b13194a8c80a09e4)*

src/special-types-and-traits.md
```diff
@@ -63,6 +63,12 @@ r[lang-types.phantom-data]
 is considered to own a `T` for the purposes of [variance], [drop check], and
 [auto traits](#auto-traits).
 
+r[lang-types.va-list]
+## `VaList<'_>`
+
+[`std::ffi::VaList<'_>`] is used for [c-variadic functions]. The rust type is
```
ehuss:
```suggestion
[`std::ffi::VaList<'_>`] is used for [c-variadic functions]. The Rust type is
```

ehuss (commented):
Some quick nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..256b0f242c7e400865b1f068b13194a8c80a09e4)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+- AmdGpu and Nvptx64
+- Wasm32 and Wasm64
+- C-SKY
+- Xtensa
+- Hexagon
+- SPARC64
+- MIPS
+
+r[items.fn.c-variadic.platform-support]
+Some ABIs do not support c-variadic function definitions. The compiler errors in this case.
```
traviscross: "ABI" or "target"?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..aab90b6489105e8464ddb672bf0b83da64a716c4)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+Examples of incompatible types are:
+
+- `usize` and `*const _`, pointers and integers are not compatible.
+- `*const fn(&'static ())` and `*const for<'a> fn(&'a ())`, these types are not equal up to free lifetimes.
+
+r[items.fn.c-variadic.ffi-compatibility]
+The rust [`VaList`] is ABI-compatible with the C `va_list` type.
+
+r[items.fn.c-variadic.abi]
+Only `extern "C"` and `extern "C-unwind"` function defintions can accept a variable argument list.
```
traviscross:
```suggestion
Only `extern "C"` and `extern "C-unwind"` function definitions can accept a variable argument list.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..aab90b6489105e8464ddb672bf0b83da64a716c4)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+- `T` and `U` are the same type (up to free lifetimes).
+- `T` and `U` are integer types of the same size.
+- `T` and `U` are both pointers, and their target types are compatible.
+- `T` is a pointer to `c_void` and `U` is a pointer to `i8` or `u8`, or vice versa.
+
+Examples of compatible types are:
+
+- `u32` and `i32`, but UB may still occur if the value is not representable in the target type.
+- `u64` and `usize`, on a 64-bit platform.
+- `*const &'a u32` and `*mut &'static u32`, these type are equal up to free lifetimes.
```
traviscross:
```suggestion
- `*const &'a u32` and `*mut &'static u32` --- these types are equal up to free lifetimes.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..e92d55c7d9dc7550f496c7d80ee656c2a160b6ad)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+Examples of incompatible types are:
+
+- `usize` and `*const _`, pointers and integers are not compatible.
+- `*const fn(&'static ())` and `*const for<'a> fn(&'a ())`, these types are not equal up to free lifetimes.
+
+r[items.fn.c-variadic.ffi-compatibility]
+The rust [`VaList`] is ABI-compatible with the C `va_list` type.
```
traviscross:
```suggestion
[`VaList`] is ABI compatible with the C `va_list` type.
```

(We tend to avoid mentioning "Rust" since everything in the Reference is about Rust by default.  We'd drop the hyphen on "ABI compatible" here because it's in predicate position rather than being used as a phrasal adjective.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..e92d55c7d9dc7550f496c7d80ee656c2a160b6ad)*

src/special-types-and-traits.md
```diff
@@ -63,6 +63,12 @@ r[lang-types.phantom-data]
...
+FFI-compatible with the C `va_list` type.
```
traviscross:
```suggestion
ABI compatible with the C `va_list` type.
```

This PR uses ABI compatible elsewhere.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..e92d55c7d9dc7550f496c7d80ee656c2a160b6ad)*

src/special-types-and-traits.md
traviscross:
```suggestion
[`std::ffi::VaList<'_>`] is used for [c-variadic functions] and is
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..e92d55c7d9dc7550f496c7d80ee656c2a160b6ad)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+Examples of compatible types are:
+
+- `u32` and `i32`, but UB may still occur if the value is not representable in the target type.
+- `u64` and `usize`, on a 64-bit platform.
+- `*const &'a u32` and `*mut &'static u32`, these type are equal up to free lifetimes.
+
+Examples of incompatible types are:
+
+- `usize` and `*const _`, pointers and integers are not compatible.
+- `*const fn(&'static ())` and `*const for<'a> fn(&'a ())`, these types are not equal up to free lifetimes.
```
traviscross:
```suggestion
- `usize` and `*const _` --- pointers and integers are not compatible.
- `*const fn(&'static ())` and `*const for<'a> fn(&'a ())` --- these types are not equal up to free lifetimes.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..f1227da54a305d1f79162733e06d79e93d6ba9ef)*

src/items/functions.md
```diff
@@ -332,6 +332,132 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+> [!NOTE]
+> Some target architectures do not support c-variadic function definitions. The compiler errors in this case.
+>
+> ```text
+> error: the `bpfel` target does not support c-variadic functions
+>   --> $DIR/not-supported.rs:23:31
+>    |
+> LL | unsafe extern "C" fn variadic(_: ...) {}
+>    |                               ^^^^^^
+> ```
```
traviscross: Probably worth dropping this; we don't generally include compiler output as it will drift.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/ac7b991181147dbb499f462b2d837c2e964dbdc7..302fba9ed10791322f51c849230f28f363cc80ec)*

src/items/functions.md
```diff
@@ -83,7 +83,7 @@ r[items.fn.params.self-restriction]
 Functions with a self parameter may only appear as an [associated function] in a [trait] or [implementation].
 
 r[items.fn.params.varargs]
-A parameter with the `...` token indicates a [variadic function], and may only be used as the last parameter of an [external block] function. The variadic parameter may have an optional identifier, such as `args: ...`.
+A parameter with the `...` token indicates a [c-variadic function], and may only be used as the last parameter. In an [`extern` block], the c-variadic parameter may have an optional identifier, such as `args: ...`, and in a [c-variadic function definition][items.fn.c-variadic], the identifier is mandatory.
```
traviscross: Some of this is preexisting, but it seems we support patterns here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..2f883bde50598bb0c02af9475e91bc89e0d60cec)*

traviscross (approved):

src/items/functions.md
```diff
@@ -83,7 +83,45 @@ r[items.fn.params.self-restriction]
...
+> #![allow(varargs_without_pattern)]
+> unsafe extern "C" fn f(...) {} // OK.
+> ```
+>
+> ```rust
+> #![allow(varargs_without_pattern)]
+> trait Tr {
+>     unsafe extern "C" fn f(...); // OK.
+> }
+> ```
```
traviscross: In adding these examples, I realized that the rule here is only enforced by a lint.  Is that intended?  Since C-variadic function definitions were not previously accepted on stable Rust, it would be more ordinary to give a hard error here rather than stabilizing something and immediately linting against it.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..a729f2d224f1ef2d9d9e7da0554203879d819c32)*

folkertdev: Hmm, we could maybe deny it later in the pipeline (so in hir validation)? The goal of the FCW is to eventually deny `...` in that position pre-expansion, which would be a breaking change.

We semantically did not accept c-variadic definitions so far, but syntactically we actually did, and macros could (and did) make use of that syntactic freedom.

src/items/functions.md
```diff
@@ -332,6 +370,218 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+unsafe extern "C" fn f(mut ap: ...) -> i32 {
+    unsafe { ap.next_arg::<i32>() }
+}
+```
+
+Roughly desugars to:
+
+<!-- no_run: conceptual desugaring -->
+```rust,no_run
+# #![ feature(core_intrinsics) ]
```
DanielEScherzer, line 431:
```suggestion
# #![feature(core_intrinsics)]
```

the other attributes don't have extra spaces here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..a729f2d224f1ef2d9d9e7da0554203879d819c32)*

src/items/functions.md
```diff
@@ -332,6 +370,218 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+r[items.fn.c-variadic.next-arg-safety]
+Calling `VaList::next_arg` to read an argument of type `T` is only safe if all of the following conditions are satisfied:
+
+- There is another C-variadic argument to read.
+- The actual type of the argument `U` is compatible with `T` (as defined below).
+- If `U` and `T` are both integer types, then the value passed by the caller must be
+representable in both types.
+
+Types `T` and `U` are compatible when:
```
DanielEScherzer: can we make it clearer that `T` and `U` are compatible when *any* of the following hold, rather than *all*? Maybe something like

```suggestion
Types `T` and `U` are compatible when one of the following is true:
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..a729f2d224f1ef2d9d9e7da0554203879d819c32)*

src/items/functions.md
```diff
@@ -332,6 +370,218 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+
+- `T` and `U` are the same type (up to free lifetimes).
+- `T` and `U` are integer types of the same size.
+- `T` and `U` are both pointers and their target types are compatible.
+- `T` is a pointer to `c_void` and `U` is a pointer to `i8` or `u8`, or vice versa.
+
+Examples of compatible types are:
+
+- `u32` and `i32` --- but UB may still occur if the value is not representable in the target type.
```
DanielEScherzer, line 470: can this be clarified? My understanding is that every bit pattern for a u32 is also a valid bit pattern for an i32, it might represent a different numerical value but that seems like a programming bug rather than UB

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..a729f2d224f1ef2d9d9e7da0554203879d819c32)*

DanielEScherzer:
@rustbot label -S-waiting-on-stabilization

Stabilization merged in rust-lang/rust#155697

src/items/functions.md
ehuss: This is intentional to circumvent the style checks which reject `feature` attributes. This allows us to temporarily include these attributes to see if CI passes.

EDIT: In this particular case, this is a perma-unstable feature, so it needs some permanent way to allow it through the checks.

My longer-term desire would be to have PR checks only generate a warning if `feature` attributes exist, and have the merge_group checks (merge queue checks) forbid them.

DanielEScherzer: then can we add a comment inline explaining that?

src/items/functions.md
traviscross: That is my inclination.  Filed as https://github.com/rust-lang/rust/issues/160109.

src/items/functions.md
folkertdev: > that seems like a programming bug rather than UB

No the C specification requires that the actual value is representable in both types, so `42` is valid but `-1` is not. Yes this is terrible, but those are the rules.

DanielEScherzer: then I would suggest adding a clarification about that, since this might confuse others too

src/items/functions.md
```diff
@@ -332,6 +370,218 @@ Note that this behavior is a consequence of the desugaring to a function that re
...
+r[items.fn.c-variadic.abi]
+Only `extern "C"` and `extern "C-unwind"` function definitions can accept a variable argument list.
+
+```rust,compile_fail
+unsafe fn f(ap: ...) {} // ERROR: Not supported.
+```
+
+```rust,compile_fail
+unsafe extern "sysv64" fn f(ap: ...) {} // ERROR: Not supported.
+```
```
folkertdev, line 507: Now that `c_variadic_naked_functions` (https://github.com/rust-lang/rust/pull/159746) is in FCP, this section needs to be extended. Do we just do that now, wait for FCP to be over, or do it in a separate PR?

I was thinking of something like this:

````
A variable argument list is only accepted on [naked functions] for the ABI strings that are listed in [items.extern.variadic.conventions].

```rust
#[unsafe(naked)]
unsafe extern "win64" fn variadic_win64(_: u32, _: ...) -> u32 {
    core::arch::naked_asm!(
        r#"
        push    rax
        mov     qword ptr [rsp + 40], r9
        mov     qword ptr [rsp + 24], rdx
        mov     qword ptr [rsp + 32], r8
        lea     rax, [rsp + 40]
        mov     qword ptr [rsp], rax
        lea     eax, [rdx + rcx]
        add     eax, r8d
        pop     rcx
        ret
    "#,
    )
}
```

[items.extern.variadic.conventions]: ../items/external-blocks.md#items.extern.variadic.conventions
[naked functions]: ../attributes.md#attributes.codegen.naked
````

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2177/afdc77bab886d4455c11247cdd32391bfab636ae..7d1f178f3d94bcdbae4257981c9b029c23e1dca0)*

src/items/functions.md
folkertdev: I added some examples

src/items/functions.md
folkertdev: Discussed in the lang meeting, this will be resolved by https://github.com/rust-lang/rust/pull/160165.

src/items/functions.md
traviscross: Thanks.  Let's do that in a separate PR (as this one needs to merge as the stabilization has already merged).

PR #2179, opened 2026-02-17
expr.block.value-(no-)trailing-expr: more clearly demonstrate intent

tshepang:
Showing type annotations clutters the examples somewhat, because we also use asserts to prove the rules.


traviscross:
We talked about this in the lang-docs office hours.  Thanks for PR.  But this doesn't seem an improvement to us, so we're going to close it.

PR #2182, opened 2026-02-19
coerce.site.return: make example more simple

tshepang:
- remove the import and use the type directly
- get the type from its original path, and not a re-export

### DIFF
--- a/src/type-coercions.md
+++ b/src/type-coercions.md
@@ -60,7 +60,7 @@ r[coerce.site.constructor]
 r[coerce.site.return]
 * Function results&mdash;either the final line of a block if it is not semicolon-terminated or any expression in a `return` statement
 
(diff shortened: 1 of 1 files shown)


src/type-coercions.md
```diff
@@ -63,8 +63,7 @@ r[coerce.site.return]
   For example, `x` is coerced to have type `&dyn Display` in the following:
```
DanielEScherzer: I think that without the `use`, the previous statement that `x` is coerced to be `&dyn Display` might be a bit confusing since `Display` is not in scope, only `core::fmt::Display` (or the std version)

tshepang: was not sure it mattered much, but now adjusted

traviscross:
We talked about this in the lang-docs office hours.  Thanks for PR.  But this doesn't seem an improvement to us, so we're going to close it.

PR #2185, opened 2026-02-20
Nightly test links: update rust branch name (use `main`)

DanielEScherzer:
Avoids the "Branch master was renamed to main." alert


ehuss (approved):
Thanks! I also updated the fallback branch names.

Issue #2153, opened 2026-02-01
What are the divergence rules around const eval?


ehuss:
https://github.com/rust-lang/reference/pull/2067 added a chapter on divergence, but we noted some uncertain behavior around const expressions. I had some specific questions that hopefully have clear answers:

- Can a const expression have type `!`?
- Can a const expression diverge?
- If it is possible to diverge:
    - Does the divergence of a const expression propagate to its outer block?
    - Are there any differences between different kinds of const expressions (like a static/const items, const contexts, `const {}` blocks)?

For this purpose, I think we can ignore [`uninhabited-static`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#uninhabited-static) which looks like it was not intended.

We struggled a little with examples because there is different behavior between `cargo check` and `cargo build`. We were initially assuming that if it passed `cargo check` it was also passing typechecking, but there might be something more complex going on there.

cc @jackh726 

----

Jack said:

> What I'm seeing is actually that const blocks (along with async blocks) is that they do not propagate their divergence because they essentially act as independent function bodies. Of course, divergence can exist inside:
>
> ```rust
> fn f() {
>     const {
>         panic!();
>         let _x = 0;
>     }
> }
> ```

and also:

> my best guess is that const eval is just masking typecheck here: we need to figure out the type of the const block, but can never do that because const eval doesn't complete.

----

## Example 1: diverging const block

Considering the following:

```rust
fn trailing_const_panic() -> ! {
    const { panic!() } // OK, seems const block has ! type
}
fn non_trailing_statement_const_panic() -> ! {
    const { panic!() };
    // ERROR, expected `!`, found `()`
}
```

I was confused on this (the semicolon). Jack says:

> This has a pretty simple explanation: the former results in an expected type of `!` for the const block so the type of the const block is `!` - the latter has no expectation because of the semicolon, so the type of the block `!` fallbacks to `()`.

but I don't understand that. My base assumption was that a `const {}` block would have the same rules and behavior as a normal block. And indeed, if you remove the `const` from `non_trailing_statement_const_panic`, it successfully compiles. Why would its type fall back to `()` if it is `const`?

## Example 2: Read from a place

@traviscross came up with a variation that includes a read from a place, and the semicolon is moved inside the `const {}` block:

```rust
fn read_from_place_with_semicolon() -> ! {
    let _x = (const {
        panic!();
    },).0;
}

[...]

PR #2186, opened 2026-02-20
Add more rules about when expressions diverge

### LINKED ISSUE #2151: What is considered a "read"?
https://github.com/rust-lang/reference/pull/2067 added a chapter on divergence with the rule `expr.block.diverging` that says:

> A block is considered to be diverging if all reachable control flow paths contain a diverging expression, unless that expression is a place expression that is not read from.

But I don't think the reference has a clear definition of what it means to "read" from a place expression. What are the rules for what is considered a "read"? Is this unique to divergence checking?

Closure capture has has a bunch of rules of what it considers a "read" (or more often, what is not considered a read). I believe there are slight differences for various reasons. For example, a reference does not seem to be read for the purpose of divergence, but for the purpose of closure capture it does seem to be considered a read and is captured. There are various things around fake reads that I have forgotten how they work. I don't know if the rules are fundamentally different, or if there is some commonality.

I'll take a wild guess as to what is *not* considered a read based on [`expr_guaranteed_to_constitute_read_for_never`](https://github.com/rust-lang/rust/blob/e823167aa6f7f03aea8e91208ce9bace8ad9ebf2/compiler/rustc_middle/src/hir/mod.rs#L228-L367):

- A borrow expression.
- The left-hand side of an assignment.
- The base of a field projection that doesn't itself constitute a read.
- The scrutinee of a match expression where at least one arm's pattern does not force a read.
- The initializer of a let statement where the pattern does not force a read.
- The scrutinee of a `let` condition where the pattern does not force a read.

(Of course non-place is always "read".)

I don't know if there are other considerations or how things like discriminants get involved.

It could also include details on which patterns force a read, which I think is [`is_guaranteed_to_constitute_read_for_never`](https://github.com/rust-lang/rust/blob/e823167aa6f7f03aea8e91208ce9bace8ad9ebf2/compiler/rustc_hir/src/hir.rs#L1833C1-L1880C6), which is more or less the following aren't reads:

- `_` wildcard
- `|` or patterns where at least one alternative doesn't force a read

See also some recent things:
- https://github.com/rust-lang/reference/pull/1837 --- Changed what a closure capture considers a "read", particularly around discriminants.
- https://github.com/rust-lang/rust/pull/150681 --- Change for enum discriminant read.

cc @jackh726

### LINKED ISSUE #2152: Detail divergence for the rest of the expressions
https://github.com/rust-lang/reference/pull/2067 added a chapter on divergence with some rules for the divergence of some of the more tricky expressions like `if` or `match`. However, we were thinking of defining the behavior of some or all of the expressions explicitly. See https://github.com/rust-lang/reference/pull/2067#issuecomment-3782115579 for the discussion.

I have a draft where I am exploring how this is presented.

cc @jackh726

ehuss:
https://github.com/rust-lang/reference/pull/2067 added a chapter on divergence along with rules for some of the more subtle expressions like `if` or `match`. This adds rules for almost all the rest of the expressions.

Most of these are pretty straightforward propagation rules which could be left implicit. It may seem a little excessive to have a separate rule for each one, but I think the consistency is worth it because there are various subtleties. This also may help us be clear about these things when adding new expressions to the language to ensure we think about the divergence rules.

This does not cover const expressions (const blocks, static, const, array repeat, etc.) because I still do not yet fully know how those should be documented (see https://github.com/rust-lang/reference/issues/2153).

I'm not entirely excited by having the long list of rules in the divergence chapter, mostly because of the length. However, I think it is helpful to cross-index these kinds of things.

There are some interesting subtleties that I noticed:

- `while` loops cannot diverge. It's a little surprising to me (I would expect the condition to allow it to diverge). I suspect this is due to the desugaring to a loop expression using `break`, which does not diverge.
- Updated details for let-chains.

References for the implementation:
- Place expression with never must be read: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/expr.rs#L318-L326
- `break` is never: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/expr.rs#L819-L820
- `continue` is never: https://github.com/rust-lang/rust/blob/0376d43d443cba463a0b6a6ec9140ea17d7b7130/compiler/rustc_hir_typeck/src/expr.rs#L861
- `return` is never: https://github.com/rust-lang/rust/blob/0376d43d443cba463a0b6a6ec9140ea17d7b7130/compiler/rustc_hir_typeck/src/expr.rs#L921
- `if` divergence: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/expr.rs#L1236-L1242
- `loop`: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/expr.rs#L1450-L1456
- `asm!`: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/expr.rs#L3684
- `match`: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/_match.rs#L19-L178
- block break: https://github.com/rust-lang/rust/blob/59fd4ef94daa991e6797b5aa6127e824f3067def/compiler/rustc_hir_typeck/src/fn_ctxt/checks.rs#L1167-L1171
- lazy bool: https://github.com/rust-lang/rust/blob/0376d43d443cba463a0b6a6ec9140ea17d7b7130/compiler/rustc_hir_typeck/src/op.rs#L111

Closes https://github.com/rust-lang/reference/issues/2152
Closes https://github.com/rust-lang/reference/issues/2151

### DIFF
--- a/src/expressions/array-expr.md
+++ b/src/expressions/array-expr.md
@@ -72,6 +72,9 @@ const EMPTY: Vec<i32> = Vec::new();
 [EMPTY; 2];
 ```
 
+r[expr.array.diverging]
+An array expression [diverges] if any of its operands diverges.
+
 r[expr.array.index]
 ## Array and slice indexing expressions
 
@@ -128,6 +131,35 @@ arr[10];                  // warning: index out of bounds
 r[expr.array.index.trait-impl]
 The array index expression can be implemented for types other than arrays and slices by implementing the [Index] and [IndexMut] traits.
 
+r[expr.array.index.diverging]
+An index expression [diverges] if either of its operands diverges, or if the type of the indexed element is the [never type] and the value is guaranteed to be read.
+
+> [!EXAMPLE]
+> ```rust
+> fn phantom_place<F: FnOnce() -> T, T>() -> [T; 1] { loop {} }
+>
+> fn diverging_place_read() -> ! {
+>     // Create an array with the never type.
+>     let x /* : [!; 1] */ = phantom_place::<fn() -> !, _>();
+>     // A read of a place expression produces a diverging block.
+>     x[0];
+> }
+> ```
+>
+> The following example contrasts with the above because it does not perform a read. This means it does not diverge, causing a compile error.
+>
+> ```rust,compile_fail,E0308
+> # fn phantom_place<F: FnOnce() -> T, T>() -> [T; 1] { loop {} }
+> #
+> fn diverging_place_no_read() -> ! {
+>     // Create an array with the never type.
+>     let x /* : [!; 1] */ = phantom_place::<fn() -> !, _>();
+>     // This does not constitute a read.
+>     let _ = x[0];
+>     // ERROR: Expected type !, found ()
+> }
+> ```
+
 [`Copy`]: ../special-types-and-traits.md#copy
 [IndexMut]: std::ops::IndexMut
 [Index]: std::ops::Index
@@ -136,9 +168,11 @@ The array index expression can be implemented for types other than arrays and sl
 [const block expression]: expr.block.const
 [constant expression]: ../const_eval.md#constant-expressions
 [constant item]: ../items/constant-items.md
+[diverges]: divergence
 [inferred const]: items.generics.const.inferred
 [literal]: ../tokens.md#literals
 [memory location]: ../expressions.md#place-expressions-and-value-expressions
+[never type]: type.never
 [panic]: ../panic.md
 [path]: path-expr.md
 [slice]: ../types/slice.md
--- a/src/expressions/block-expr.md
+++ b/src/expressions/block-expr.md
@@ -196,6 +196,9 @@ The actual data format for this type is unspecified.
 > [!NOTE]
 > The future type that rustc generates is roughly equivalent to an enum with one variant per `await` point, where each variant stores the data needed to resume from its corresponding point.
 
+r[expr.block.async.diverging]
+An async block expression does not itself [diverge], but evaluating the future (such as through `await`) diverges if the output type is the [never type].
+
 r[expr.block.async.edition2018]
 > [!EDITION-2018]
 > Async blocks are only available beginning with Rust 2018.
@@ -357,6 +360,7 @@ fn is_unix_platform() -> bool {
 [call expressions]: call-expr.md
 [capture modes]: ../types/closure.md#capture-modes
 [constant items]: ../items/constant-items.md
+[diverge]: divergence
 [diverges]: expr.block.diverging
 [final operand]: expr.block.inner-attributes
 [free item]: ../glossary.md#free-item
--- a/src/expressions/path-expr.md
+++ b/src/expressions/path-expr.md
@@ -34,9 +34,14 @@ let slice_reverse = <[i32]>::reverse;
 r[expr.path.const]
 Evaluation of associated constants is handled the same way as [`const` blocks].
 
+r[expr.path.diverging]
+A path expression [diverges] if the path resolves to a value with the [never type] and it is a place expression that is guaranteed to be read.
+
+[diverges]: divergence
+[never type]: type.never
+[path]: paths
 [place expressions]: ../expressions.md#place-expressions-and-value-expressions
 [value expressions]: ../expressions.md#place-expressions-and-value-expressions
-[path]: ../paths.md
 [`static mut`]: ../items/static-items.md#mutable-statics
 [`unsafe` block]: block-expr.md#unsafe-blocks
 [`const` blocks]: block-expr.md#const-blocks
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -1726,3 +1726,14 @@ On ARM, the following additional directives are guaranteed to be supported:
 - `.code`
 - `.thumb`
 - `.thumb_func`
+
+## Divergence
+
+r[asm.diverging.noreturn]
+The `asm!` macro [diverges] if it uses the [`noreturn`] option and no `label` block returns unit.
+
+r[asm.diverging.naked_asm]
+The `naked_asm!` macro always [diverges].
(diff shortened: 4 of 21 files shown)


ehuss:
@jackh726 Would you be able to review this?

src/expressions/array-expr.md
```diff
@@ -113,6 +116,35 @@ arr[10];                  // warning: index out of bounds
 r[expr.array.index.trait-impl]
 The array index expression can be implemented for types other than arrays and slices by implementing the [Index] and [IndexMut] traits.
 
+r[expr.array.index.diverging]
+An index expression [diverges] if either of its operands diverges, or if the type of the indexed element is the [never type] and the value is guaranteed to be read.
```
jackh726: >its operands diverges

worth an example for this?

src/expressions/block-expr.md
```diff
@@ -191,6 +191,9 @@ The actual data format for this type is unspecified.
 > [!NOTE]
 > The future type that rustc generates is roughly equivalent to an enum with one variant per `await` point, where each variant stores the data needed to resume from its corresponding point.
 
+r[expr.block.async.diverging]
+An async block expression does not itself [diverge], but evaluating the future (such as through `await`) diverges if the output type is the [never type].
```
jackh726, line 182: The latter half is covered by `expr.await.diverging`. Is the duplication worth it?

jackh726 (commented):
Seems good to me. Definitely the expression list in `divergence.md` is pretty long now.

src/expressions/array-expr.md
```diff
@@ -128,6 +131,35 @@ arr[10];                  // warning: index out of bounds
...
+>     // Create an array with the never type.
+>     let x /* : [!; 1] */ = phantom_place::<fn() -> !, _>();
+>     // A read of a place expression produces a diverging block.
+>     x[0];
+> }
+> ```
+>
+> The following example contrasts with the above because it does not perform a read. This means it does not diverge, causing a compile error.
+>
+> ```rust,compile_fail
```
traviscross:
```suggestion
> ```rust,compile_fail,E0308
```

(And below.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2186/6f2d1eac3b9b327536b7c481f7618f14ec83f314..74b0026e83f8e49cb24db9bb0b9c686309d10729)*

src/inline-assembly.md
```diff
@@ -1726,3 +1726,14 @@ On ARM, the following additional directives are guaranteed to be supported:
 - `.code`
 - `.thumb`
 - `.thumb_func`
+
+## Divergence
+
+r[asm.diverging.noreturn]
+The `asm!` macro [diverges] if it uses the [`noreturn`] option.
```
traviscross:
```suggestion
The `asm!` macro [diverges] if it uses the [`noreturn`] option and no `label` block returns unit.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2186/6f2d1eac3b9b327536b7c481f7618f14ec83f314..74b0026e83f8e49cb24db9bb0b9c686309d10729)*

traviscross: Our rule says:

> r[asm.options.supported-options.noreturn]
> - `noreturn`: The assembly code does not fall through; behavior is undefined if it does. It may still jump to `label` blocks. If any `label` blocks return unit, the `asm!` block will return unit. Otherwise it will return `!` (never). As with a call to a function that does not return, local variables in scope are not dropped before execution of the assembly code.

Interestingly, it doesn't actually say whether or not it *diverges*.  But, logically, it should not, and testing bears this out.

The `noreturn` option is misnamed now that we've stabilized labeled blocks.  It should really be `nofallthrough` or similar — it's asserting only that we don't fall through the assembly code.  If we jump from that code to a labeled block, that labeled block can return normally without diverging.

src/expressions/path-expr.md
```diff
@@ -34,9 +34,19 @@ let slice_reverse = <[i32]>::reverse;
...
 Evaluation of associated constants is handled the same way as [`const` blocks].
 
+r[expr.path.diverging]
+A path expression does not [diverge].
+
+<!--
+TODO: When stabilizing never type, this changes to:
+
+A path expression [diverges] if the path resolves to a value with the [never type] and it is a place expression that is guaranteed to constitute a read.
+-->
```
traviscross: It's actually possible to observe this on stable Rust.  E.g.:

```rust
fn phantom_call<T>(_: impl FnOnce(T) -> T) {}
fn main() {
    let _ = phantom_call(|x| -> ! {
        let _x = x; // OK.  
    });
    let _ = phantom_call(|x| -> ! {
        let _ = x; // ERROR: Mismatched types.
    });
}
```

Let's also adjust this language to use the "value is guaranteed to be read" phrasing used elsewhere.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2186/6f2d1eac3b9b327536b7c481f7618f14ec83f314..74b0026e83f8e49cb24db9bb0b9c686309d10729)*

ehuss:
(For the record, I am finding more mistakes here and am working on them.)

traviscross:
@ehuss and I talked about this briefly today.  Nominating for us to discuss in a future lang-docs call.  There's a question here about how to approach this, in particular about defining *reads*, which is something that also comes up for closure capturing and for static initializers.

PR #2187, opened 2026-02-20
Add section on type-relative name-resolution

yaahc:
Currently doing content reviews w/ other subject matter exports. Not yet ready for editorial review.

### DIFF
--- a/src/names/name-resolution.md
+++ b/src/names/name-resolution.md
@@ -701,7 +701,7 @@ fn main() {
 }
 ```
 
-consider the following:
+Candidate selection depends on inference, the traits which are considered "applicable" are determined via the Self type the name being resolved is relative to. consider the following:
 
 ```rust
 struct Foo;


src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+information to the declarations of those entities once that type information is
+available in the compiler. This stage of name resolution exclusively applies to
+a subset of [Qualified Paths].
+
+Specifically, Qualified Paths from:
+
+- Types
+- Expressions
+- Patterns
+- Structs & Tuple Structs
```
yaahc, line 573: I think this might be mcga related except for where it overlaps w/ enums

src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+
+`Type::assoc_item`, `<Type>::assoc_item`, `<Enum>::Variant` and
+`EnumTyAlias::Variant` are resolved during type checking. These cannot be
+resolved earlier because the type checker needs to resolve the trait an
+associated item comes from or the definition an alias points to.
+
+r[names.resolution.type-relative.stages]
+These type relative paths are resolved separately in item definitions and in function bodies.
+
+r[names.resolution.type-relative.stages.items]
```
yaahc, line 595: there's a lot more complexity in the compiler than what I've represented here, but afaict everything else is tied to unstable syntax.

This section covers the stable logic in `lower_type_relative_ty_path` in `rustc_hir_analysis`.

src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+
+Type::AssocTy
+
+```rust
+// ty path
+fn foo<T: Trait>() -> T::AssocTy { .. } // AssocTy can come from a trait and it could be ambiguous with an enum variant
+                                        // or an inherent associated type (on unstable)
+```
+
+r[names.resolution.type-relative.stages.bodies]
```
yaahc, line 606: This section represents the logic in `resolve_fully_qualified_call`. Still seems odd to me that variants are being resolved in this function given it's name but that seems to be whats happening in the code so /shrug.

src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+    - inherent associated functions
+    - trait associated functions
+    - methods via fqcs
+        - No autoderefs are performed
+        - static methods are included
+        - lookup is done based on the type each implementation is for
+            - what does this actually mean?
+
+r[names.resolution.type-relative.stages.bodies.call-expressions.differences-from-method-resolution]
+
```
yaahc, line 670: add link to existing docs on method resolution and disambiguating function calls

src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+fn main() {
+    //<_>::inherent_func(); // ERROR: unable to get the assoc item
+    //<_>::trait_method(); // ERROR: chose trait method, unable to infer the self type
+    let _: Foo = <_>::trait_method(); // OK
+}
+```
+
+- we only consider trait items for <_> but also consider - and prefer - inherent impls if the self type is not an infer var
+    - There would be no way to refer to inherent associated items otherwise, we lack a disambiguation syntax in their favor
```
yaahc: move this example before previous one

src/names/name-resolution.md
```diff
@@ -558,8 +558,201 @@ r[names.resolution.primary]
...
+fn main() {
+    let x: &dyn Trait = &1;
+    Trait::assoc(x); // trait
+    <Trait>::assoc(x); // ambiguous
+    <Trait as Trait>::assoc(x); // trait
+    x.assoc(); // ambiguous
+}
+```
+
+struct literals?
```
yaahc, line 751: double check why I thought this may be a type relative case. My guess is it probably has to do with caller's of `lower_qpath` and I think this might reduce down to just enum literals for which stable syntax it actually covers. If so, merge this back into the enum section as a subsection.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+
+trait Trait: Sized {
+    fn trait_method() -> Self {
+        todo!();
+    }
+}
+impl Trait for Foo {}
+
+fn main() {
+    //<_>::inherent_func(); // ERROR: unable to get the assoc item
```
yaahc, line 720: this is never considered because Self is an inference var and this method is an inherent method

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+    //<_>::trait_method(); // ERROR: chose trait method, unable to infer the self type
```
yaahc, line 721: This is an error because candidate selection depends upon inference resolving Self. In this case Self is unconstrained so we have no candidates at all. The fact that we only have one trait_method candidate that would be considered here cannot flow backwards and help constrain the Self inference var.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+    let _: Foo = <_>::trait_method(); // OK
```
yaahc, line 722: Self is constrained by return type annotation, we consider trait methods and find a candidate.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+}
+impl Trait for Foo {}
+
+fn main() {
+    <_>::overlap(Foo); // trait
+    <Foo>::overlap(Foo); // inherent
+}
+```
+
+consider the following:
```
yaahc:
```suggestion
Candidate selection depends on inference, the traits which are considered "applicable" are determined via the Self type the name being resolved is relative to. consider the following:
```

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
 r[names.resolution.type-relative]
 ## Type-relative resolution
-> [!NOTE]
-> This is a placeholder for future expansion about type-dependent resolution.
+
+r[names.resolution.type-relative.intro]
+_Type-relative resolution_ is the process of tying paths that depend on type
+information to the declarations of those entities once that type information is
+available in the compiler. This stage of name resolution exclusively applies to
+a subset of [Qualified Paths].
```
petrochenkov, line 566: No, even examples below include non-qualified paths like `Type::assoc_item`.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+
+Specifically, Qualified Paths from:
+
+- Types
+- Expressions
+- Patterns
+- Structs & Tuple Structs
```
petrochenkov, line 573: Structs are types, and tuple structs are expressions.
(Except tuple structs in patterns which are a special production.)

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+candidates purely based on information within the enum/trait's original
+definition.
+
+`Type::assoc_item`, `<Type>::assoc_item`, `<Enum>::Variant` and
+`EnumTyAlias::Variant` are resolved during type checking. These cannot be
+resolved earlier because the type checker needs to resolve the trait an
+associated item comes from or the definition an alias points to.
+
+r[names.resolution.type-relative.stages]
+These type relative paths are resolved separately in item definitions and in function bodies.
```
petrochenkov, line 593: You mean in signatures and in bodies?

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+enum variant via alias
+
+```rust
+enum Enum {
+    Variant,
+}
+type Alias = Enum;
+fn main() {
+    match Enum::Variant {
+        Alias::Variant => {}
```
petrochenkov, line 622: Path pattern is used as an example for a path expression.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+    }
+}
+```
+
+Fully qualified paths for enums are also resolved as type relative names. This
+is mainly because the implementation is simpler this way not because this
+information would not otherwise be available during primary resolution.
```
petrochenkov, line 629: If `Enum` has both a variant and an associated item called `Variant`, then the information about the associated item is not available during primary resolution.
But the statement still seems correct, because the disambiguation is always in favor of the variant.

src/names/name-resolution.md
```diff
@@ -558,8 +558,205 @@ r[names.resolution.primary]
...
+```
+
+When enum variants share the same name as associated types or consts, the enum variant candidate is given preference during resolution
+
+
+r[names.resolution.type-relative.stages.bodies.call-expressions]
+fully qualified call expressions:
+
+```rust
+Type::assoc_func();
```
petrochenkov, line 657: Hm? Why is it fully qualified? It's just a call expression.

src/names/name-resolution.md
petrochenkov: This is probably `TyAlias::AssocType { field1, field2, .. }`, or `<Enum>::Variant { field1, field2, .. }` or anything similar.
The thing before braces in a struct literal is an arbitrary type path (even including generic arguments).

src/names/name-resolution.md
yaahc: I've been working with the assumption that these are all normalized to a qualified path representation since all of the inputs to the type-relative name resolution code that I'm aware of are of the form `QPath::TypeRelative`.

This does beg the question though of whether that's purely an internal compiler detail, something we want to represent as part of our model of the language, or just something I got wrong.

petrochenkov: Ah, okay.
The reference just has a definition for a qualified path, which is syntax-based.

PR #2189, opened 2026-02-24
Fix an EN grammar error & add an item to place expr context list

bitcp:
The reference calls tuple components **field**s at some places, but the syntax distinguishes between **FieldExprssion**, IndexExpression and **TupleIndexingExpression**.

Not sure whether this adding is a redundancy.


ehuss (approved):
Thanks! Yea, it was a little ambiguous here, and seems reasonable to be more explicit (even if internally the compiler combines the different field kinds).

I added a small commit to add links to the expressions.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2189/442cbef9105662887d5eae2882ca551f3726bf28..928d6065b00304a2c71864905cc1624829250d39)*

PR #2190, opened 2026-02-24
Rewrite Charset so that it uses Expression

ehuss:
This removes the `Characters` type, and instead uses `Expression`. There was a lot of overlap with `Expression`, and this simplifies things a little bit and removes some duplication.

This also helps with some future changes I am working on that do coverage analysis, and this makes it easier to track coverage data for the different character values.

The downside is that without the enum it isn't clear from the code what limitations there are with `ExpressionKind::Charset`, or that `ExpressionKind::CharacterRange` should only be in a `Charset`. These are implicit based on how the parser works. I'm willing to make that tradeoff.

The `Characters` map as:

- `Characters::Named` to `ExpressionKind::Nt`
- `Characters::Terminal` to `ExpressionKind::Terminal`
- `Characters::Range` to the NEW `ExpressionKind::CharacterRange`

This incidentally fixes a small oversight where `Characters::Named` was missing span with the `grammar-text` class.

### DIFF
--- a/tools/grammar/src/lib.rs
+++ b/tools/grammar/src/lib.rs
@@ -80,7 +80,14 @@ pub enum ExpressionKind {
     /// `// Single line comment.`
     Comment(String),
     /// ``[`A`-`Z` `_` LF]``
-    Charset(Vec<Characters>),
+    ///
+    /// This should only contain expressions that are valid inside brackets
+    /// (`Terminal`, `Nt`, and `CharacterRange`).
+    Charset(Vec<Expression>),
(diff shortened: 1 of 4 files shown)


ehuss:
Opening as a draft as I'm not 100% confident that this is the right tradeoff to make. It would be nice if the type system could enforce the constraints, but I don't see an alternative way to do that which is any better than this.

ehuss:
Closing as this was essentially included in https://github.com/rust-lang/reference/pull/2325.

PR #2193, opened 2026-02-25
Remove RESERVED_NUMBER

ehuss:
This removes RESERVED_NUMBER and instead embeds the correctness requirements for a number in the literal productions using the cut operator.

I like the idea of inlining these requirements into the grammar so that the restrictions are clear and nearer to the rules they are related to.

I'm on the fence about keeping the rules describing the restrictions. In general we don't want to describe the grammar in English due to the potential ambiguity. If we do this here, why wouldn't we do this everywhere there is a cut operator? However, I think it's still helpful to include in this case.

SUFFIX_NO_E is no longer needed because all of the other rules handle rejecting the E.

### DIFF
--- a/src/tokens.md
+++ b/src/tokens.md
@@ -109,7 +109,7 @@ r[lex.token.literal.suffix]
 #### Suffixes
 
 r[lex.token.literal.literal.suffix.intro]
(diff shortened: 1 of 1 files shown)


traviscross (approved):

traviscross:
Looks good.  Thanks.

PR #2194, opened 2026-03-01
Align attribute template with applied conventions

traviscross:
During review and revision of the PRs applying the attribute template to attributes in the Reference, we've established several standard phrasings.  We had earlier updated the attribute template for most of these.  But for some, we hadn't yet.  Let's align the template with four of these actual conventions.

cc @ehuss

### DIFF
--- a/dev-guide/src/attributes.md
+++ b/dev-guide/src/attributes.md
@@ -50,7 +50,7 @@ The `example` attribute uses the [MetaNameValueStr] syntax. The value in the str
 
 ----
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2194/50a1075e879be75aeec436252c84eef0fad489f4..0e1358bf65d089890ba1af6eb3967c6eb5bbf11c)*

PR #2196, opened 2026-03-04
Support non-ASCII Unicode in grammar rule names

traviscross:
The grammar currently supports only ASCII rule names.  We want to support non-ASCII Unicode symbols such as `⊥` (bottom) since we plan to add that rule.

In this commit, we add `is_name_start` and `is_name_continue` predicates that centralize the decision of what can appear in a rule name.  `is_name_start` accepts alphabetic characters, underscores, and non-ASCII characters; `is_name_continue` accepts alphanumeric characters, underscores, and non-ASCII characters.

We use `is_name_start` in the `parse_expr1` condition that routes to `parse_nonterminal`.  The previous condition (`is_alphanumeric`) was slightly misaligned with what `parse_name` actually accepts -- it included digits (which `parse_name` rejects) and excluded underscores (which `parse_name` accepts).  Using `is_name_start` makes the dispatch condition match `parse_name` exactly.

The `NAMES_RE` regex in `mdbook-spec` encodes the same name-matching logic as a regex pattern, so let's add a comment tying it to the predicates.

cc @ehuss

### DIFF
--- a/tools/grammar/src/parser.rs
+++ b/tools/grammar/src/parser.rs
@@ -38,6 +38,24 @@ macro_rules! bail {
 
 type Result<T> = std::result::Result<T, Error>;
 
+/// Whether a character can start a grammar rule name.
+///
+/// This includes ASCII alphabetic characters, underscores, and
+/// non-ASCII Unicode symbols such as `⊥` (bottom) and `⊤` (top).
+/// ASCII symbols are excluded because characters such as `+`, `|`,
+/// `~`, and `^` are grammar syntax.
+fn is_name_start(ch: char) -> bool {
+    ch.is_alphabetic() || ch == '_' || !ch.is_ascii()
+}
+
+/// Whether a character can continue a grammar rule name.
+///
+/// Accepts alphanumeric characters, underscores, and non-ASCII
+/// characters.
+fn is_name_continue(ch: char) -> bool {
+    ch.is_alphanumeric() || ch == '_' || !ch.is_ascii()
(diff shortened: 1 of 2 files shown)


tools/grammar/src/parser.rs
```diff
@@ -152,18 +170,11 @@ impl Parser<'_> {
...
+        if !is_name_start(first) {
             return None;
         }
-        let name = self.take_while(&|c: char| c.is_alphanumeric() || c == '_');
-        if name.is_empty() {
-            None
-        } else {
-            Some(name.to_string())
-        }
+        Some(self.take_while(&|c| is_name_continue(c)).to_string())
```
ehuss, line 177: This is written in such a way that it assumes is_name_continue is a superset of is_name_start. My intention was instead to have something like this:

```rust
        let mut name = String::from(start);
        self.index += name.len();
        let rest = self.take_while(&|c: char| is_name_continue(c));
        name.push_str(rest);
        Some(name)
```

ehuss: Nevermind. AFAIK xid_start is always a subset of xid_continue.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2196/2dfed35b22f3bad2ddba15a0d1b4703d3915a7ca..93a2336435050180699bc7d2a25e9d7efc5c5565)*

PR #2197, opened 2026-03-04
Add frontmatter

traviscross:
This PR adds Reference text for frontmatter.

Stabilization:

- https://github.com/rust-lang/rust/pull/148051

Tracking:

- https://github.com/rust-lang/rust/issues/136889

Original PR:

- https://github.com/rust-lang/reference/pull/1974

cc @ehuss @epage

### DIFF
--- a/src/frontmatter.md
+++ b/src/frontmatter.md
@@ -1,6 +1,20 @@
 r[frontmatter]
 # Frontmatter
 
+r[frontmatter.intro]
+Frontmatter is an optional section of metadata whose syntax allows external tools to read it without parsing Rust.
+
+> [!EXAMPLE]
+> <!-- ignore: test runner doesn't support frontmatter -->
+> ```rust,ignore
+> #!/bin/env cargo
+> --- cargo
+> package.edition = "2024"
+> ---
+>
+> fn main() {}
+> ```
+
 r[frontmatter.syntax]
 ```grammar,lexer
 @root FRONTMATTER ->
@@ -33,20 +47,6 @@ MAYBE_INFOSTRING_OR_WS ->
 
 INFOSTRING -> (XID_Start | `_`) ( XID_Continue | `-` | `.` )*
 ```
-
-r[frontmatter.intro]
-Frontmatter is an optional section of metadata whose syntax allows external tools to read it without parsing Rust.
-
-> [!EXAMPLE]
-> <!-- ignore: test runner doesn't support frontmatter -->
-> ```rust,ignore
-> #!/bin/env cargo
-> --- cargo
-> package.edition = "2024"
-> ---
->
-> fn main() {}
-> ```
 
 r[frontmatter.position]
 Frontmatter may appear at the start of the file (after the optional [byte order mark]) or after a [shebang]. In either case, it may be preceded by [whitespace].


src/frontmatter.md
```diff
@@ -0,0 +1,65 @@
...
+r[frontmatter.fence]
+Frontmatter must start and end with a *fence*. Each fence must start at the beginning of a line. The opening fence must consist of at least 3 and no more than 255 hyphens (`-`). The closing fence must have exactly the same number of hyphens as the opening fence. The hyphens may be followed by [horizontal whitespace].
+
+r[frontmatter.infostring]
+The opening fence may be followed by an infostring that identifies the format or purpose of the body. An infostring may be followed by [horizontal whitespace].
+
+r[frontmatter.body]
+No line in the body may start with a sequence of hyphens (`-`) equal to or longer than the opening fence. The body may not contain carriage returns.
+
+[horizontal whitespace]: whitespace.md#grammar-HORIZONTAL_WHITESPACE
```
ehuss: If you want, I think this can use the automatic link:


```suggestion
[horizontal whitespace]: grammar-HORIZONTAL_WHITESPACE
```

traviscross: PR to make that work:

- https://github.com/rust-lang/reference/pull/2198

src/notation.md
```diff
@@ -45,6 +45,20 @@ Mizushima et al. introduced [cut operators][cut operator paper] to parsing expre
...
+In logic, ⊥ (*bottom*) represents absurdity --- a proposition that is always false. In type theory, it is the *empty type*: a type with no inhabitants. The grammar borrows both senses: the rule ⊥ matches nothing --- not any character, not even the end of input.
+
+```grammar,notation
+// The bottom rule does not match anything.
+⊥ -> !(CHAR | EOF)
+```
+
+Placed after a [hard cut operator], `^ ⊥` makes a rule fail unconditionally once the parser has committed past the cut. This gives the grammar a way to express *recognition without acceptance*: the parser identifies the input, commits so that no other alternative can be tried, and then rejects it. In the frontmatter grammar, for example, [`FRONTMATTER_INVALID`] uses `^ ⊥` to recognize an opening fence preceded by whitespace on the same line --- input that is close enough to frontmatter to rule out other interpretations, but that is not valid.
+
+[`FRONTMATTER_INVALID`]: frontmatter.md#grammar-FRONTMATTER_INVALID
```
ehuss: This is not necessary if we remove the backticks.

```suggestion
```

src/notation.md
```diff
@@ -45,6 +45,20 @@ Mizushima et al. introduced [cut operators][cut operator paper] to parsing expre
...
+### The bottom rule
+
+In logic, ⊥ (*bottom*) represents absurdity --- a proposition that is always false. In type theory, it is the *empty type*: a type with no inhabitants. The grammar borrows both senses: the rule ⊥ matches nothing --- not any character, not even the end of input.
+
+```grammar,notation
+// The bottom rule does not match anything.
+⊥ -> !(CHAR | EOF)
+```
+
+Placed after a [hard cut operator], `^ ⊥` makes a rule fail unconditionally once the parser has committed past the cut. This gives the grammar a way to express *recognition without acceptance*: the parser identifies the input, commits so that no other alternative can be tried, and then rejects it. In the frontmatter grammar, for example, [`FRONTMATTER_INVALID`] uses `^ ⊥` to recognize an opening fence preceded by whitespace on the same line --- input that is close enough to frontmatter to rule out other interpretations, but that is not valid.
```
ehuss:
```suggestion
Placed after a [hard cut operator], `^ ⊥` makes a rule fail unconditionally once the parser has committed past the cut. This gives the grammar a way to express *recognition without acceptance*: the parser identifies the input, commits so that no other alternative can be tried, and then rejects it. In the frontmatter grammar, for example, [FRONTMATTER_INVALID] uses `^ ⊥` to recognize an opening fence preceded by whitespace on the same line --- input that is close enough to frontmatter to rule out other interpretations, but that is not valid.
```

traviscross: Thanks.  Fixed.

src/notation.md
traviscross: Thanks.  Fixed.

src/frontmatter.md
traviscross: Now merged.  Rebased onto that.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2197/1d674cba8866cf7ab768d9b14530c3ba7f226620..4d9d363b12940f13614cb4ac9cb9ae4e308f3ebb)*

src/frontmatter.md
```diff
@@ -0,0 +1,66 @@
...
+> package.edition = 2024
+> ---
+>
+> fn main() {}
+> ```
+
+r[frontmatter.syntax]
+```grammar,lexer
+@root FRONTMATTER ->
+    WHITESPACE_ONLY_LINE*
```
DanielEScherzer, line 21: why is the whitespace considered part of the frontmatter for the lexer (here), but in `frontmatter.position` it describes whitespace as **preceding** (and thus not part of) the frontmatter?

I'm also a bit confused about the combination of this and the `FRONTMATTER_INVALID` grammar - what is the benefit of including the discussion of invalid frontmatter in the grammar specification itself, rather than just in the prose?

src/frontmatter.md
```diff
@@ -0,0 +1,66 @@
...
+Frontmatter may appear at the start of the file (after the optional [byte order mark]) or after a [shebang]. In either case, it may be preceded by [whitespace].
+
+r[frontmatter.fence]
+Frontmatter must start and end with a *fence*. Each fence must start at the beginning of a line. The opening fence must consist of at least 3 and no more than 255 hyphens (`-`). The closing fence must have exactly the same number of hyphens as the opening fence. The hyphens of either fence may be followed by [horizontal whitespace].
+
+r[frontmatter.infostring]
+The opening fence, after optional [horizontal whitespace], may be followed by an infostring that identifies the format or purpose of the body. An infostring may be followed by horizontal whitespace.
+
+r[frontmatter.body]
+No line in the body may start with a sequence of hyphens (`-`) equal to or longer than the opening fence. The body may not contain carriage returns.
```
DanielEScherzer: I got confused about the "body must not contain carriage returns" at first, since I use windows - then I remembered `input.crlf`, so it shouldn't be too much of an issue. But, can I suggest adding a note that that normalization is applied before the presence of carriage returns is validated? Other multi-line contexts like doc blocks do allow post-`input.crlf` normalization CR to remain

DanielEScherzer (commented):
should there also be documentation that the content of frontmatter (after validation) is entirely ignored by rustc?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2197/1d674cba8866cf7ab768d9b14530c3ba7f226620..4d9d363b12940f13614cb4ac9cb9ae4e308f3ebb)*

src/frontmatter.md
traviscross: Yes, I do want to extract `WHITESPACE_ONLY_LINE*` from the `FRONTMATTER` rule and make it part of an outer rule that defines the Rust file format overall.  That went a bit beyond what makes sense to do this in PR; hence accepting this minor tension for the moment.

Regarding the other, we're working on tooling that uses our formal grammar to interpret Rust syntax for correctness checking.  That tooling can read formal notation but not prose.

src/frontmatter.md
traviscross: Added a parenthetical.  This is most similar, grammatically, to fenced raw string literals which also have this restriction.

PR #2198, opened 2026-03-04
Resolve grammar rules in link reference definitions

traviscross:
Currently, in our Markdown, we support `[text][RULE_NAME]` and `[text][grammar-RULE_NAME]` for linking to grammar rules, but we don't support this syntax within link reference definitions, i.e., `[text]: grammar-RULE_NAME`, even though we do support linking to (non-grammar) rule identifiers within link reference definitions.  That's an inconsistency that continually surprises us.  Let's fix that.

In this commit, we add `grammar_link_references`, which scans link reference definitions for destinations that match a grammar rule name -- either with a `grammar-` prefix or not.  When a match is found, the destination is replaced with the resolved path and anchor, just as `rule_link_references` does for rules.  Unrecognized destinations pass through unchanged, falling through to `std_links` for rustdoc resolution -- the same behavior as unresolved `[text][NAME]` reference links.

We also update the dev-guide to document the new feature in both `links.md` and `grammar.md`.

cc @ehuss

### DIFF
--- a/dev-guide/src/grammar.md
+++ b/dev-guide/src/grammar.md
@@ -154,5 +154,19 @@ The [`mdbook-spec`] plugin automatically adds Markdown link definitions for all 
 
 In some cases, there might be name collisions with the automatic linking of rule names. In that case, disambiguate with the `grammar-` prefix, such as `[Type][grammar-Type]`. The prefix can also be used when explicitness would aid clarity.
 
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2198/caa4205267dbd50be1dae40128e231479e38a726..f7e2b38301917dc56e49a90f271b47b908aebe7f)*

PR #2192, opened 2026-02-25
Update shebang

ehuss:
This updates the shebang description, primarily to add a grammar for it. It also reworks the rules so the introduction explains what this section is talking about, and placing specific behaviors in individual rules.

PR #2199, opened 2026-03-04
Move shebang into its own subchapter

traviscross:
In rust-lang/reference#2192, we extended our documentation about shebangs but kept it in Input Format under a section on shebang removal.  But really, this section is mostly now about documenting what a shebang is, for our purposes.

Let's separate this out into a new subchapter.

We'll keep the rule about shebang removal in Input Format; it can link to Shebang for a description of what a shebang is, its grammar, etc.

This follows the pattern that we'll similarly be using for frontmatter.

cc @ehuss

### DIFF
--- a/src/shebang.md
+++ b/src/shebang.md
@@ -1,0 +1,34 @@
+r[shebang]
+# Shebang
+
+r[shebang.intro]
+A *[shebang]* is an optional line that is typically used in Unix-like systems to specify an interpreter for executing the file.
+
+> [!EXAMPLE]
+> <!-- ignore: tests don't like shebang -->
+> ```rust,ignore
+> #!/usr/bin/env rustx
+>
+> fn main() {
+>     println!("Hello!");
+> }
+> ```
+
+r[shebang.syntax]
+```grammar,lexer
+@root SHEBANG ->
+    `#!` !((WHITESPACE | LINE_COMMENT | BLOCK_COMMENT)* `[`)
+    ~LF* (LF | EOF)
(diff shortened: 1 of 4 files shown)


src/shebang.md
```diff
@@ -0,0 +1,34 @@
...
+# Shebang
+
+r[shebang.syntax]
+```grammar,lexer
+@root SHEBANG ->
+    `#!` !((WHITESPACE | LINE_COMMENT | BLOCK_COMMENT)* `[`)
+    ~LF* (LF | EOF)
+```
+
+r[shebang.intro]
```
ehuss, line 4: Is there a reason the intro was moved to be below the syntax?

Generally I'd like to move to an outline where sections are generally organized so they start with an introduction, then have the syntax, then key behaviors and rules, and then more minutia like restrictions.

I realize there are big sections of the reference that still have the syntax first, but I would prefer to move to this style.

traviscross: No reason other than, as you say, matching similar nearby chapters.  I had similarly considered it both ways.  I'll move it.

traviscross: OK.  Moved now.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2199/3199c6a83440bee2197fe72e356d9ba6244f04e4..65f5ce35db00c7bf170217fc62f3489c6661141b)*

PR #2200, opened 2026-03-04
[type layout] usize and isize have the same size and alignment

joshlf:
Everybody assumes this to be true, and I assume it was intended to be guaranteed to be true.

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -42,8 +42,10 @@ The size of most primitives is given in this table.
 | `f64`             | 8                  |
 | `char`            | 4                  |
 
-r[layout.primitive.size-int]
+r[layout.primitive.size-minimum]
 `usize` and `isize` have a size big enough to contain every address on the target platform. For example, on a 32 bit target, this is 4 bytes, and on a 64 bit target, this is 8 bytes.
+
+r[layout.primitive.size-align]
 `usize` and `isize` have the same size and alignment.
 
 r[layout.primitive.align]
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -44,6 +44,7 @@ The size of most primitives is given in this table.
 
 r[layout.primitive.size-int]
 `usize` and `isize` have a size big enough to contain every address on the target platform. For example, on a 32 bit target, this is 4 bytes, and on a 64 bit target, this is 8 bytes.
+`usize` and `isize` have the same size and alignment.
 
 r[layout.primitive.align]
 The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.


traviscross:
Makes sense.  It is a new language guarantee, though, so...

@rfcbot fcp merge lang

traviscross:
@rfcbot fcp merge lang

scottmcm:
Is there an equivalent statement about alignment that `iN` and `uN` have the same alignment in the reference?

joshlf:
> Is there an equivalent statement about alignment that `iN` and `uN` have the same alignment in the reference?

No. I'd be happy to relax this to *just* saying that sizes are equal, which is really what I (personally) care about.

scottmcm:
TBH I think that both are obvious things to provide.  Was just seeing that we have the size equality guarantees for `uN` & `iN` (via the table specifying their exact sizes) so this is blatantly obvious for size.  Was just thinking that if the alignment equality was worth guaranteeing for `usize` & `isize` then it's probably worth doing for the others too.

(I have no strong feelings about the best way to do that.  If you wanted to just just make this be size that's fine; if you wanted to just open another PR to also give the alignment guarantee for the specific-sized integer types that'd also be fine.)

joshlf:
Oh *I'm* totally happy guaranteeing both. From my perspective, more guarantees is always better. But I didn't know if there was some reason that the language might want to keep this option open – maybe theres' some weird opsem thing I don't know about (cc @RalfJung) or some weird architecture on which these two have different alignments. For example, there's that MSVC thing where its allocator allocates 8-byte-aligned values at 4 byte alignment – maybe there's some architecture where LLVM lowers signed values differently than unsigned ones or something weird like that.

joshlf:
If we end up deciding we're okay with alignment, I'd also be happy to update this PR to guarantee alignment equality b/w *all* `uN` and `iN`, not just `usize` and `isize`.

RalfJung:
No weird opsem thing I am aware of (LLVM doesn't even know the sign of our integers most of the time). But the potential for weird architectures is of course unbounded and that's not my expertise. That said, AFAIK the LLVM data layout string cannot even represent sign-dependent alignment so it seems unlikely that LLVM could work on a target where integers with the same size and different sign have different alignment. Cc @nikic

joshtriplett:
> Is there an equivalent statement about alignment that `iN` and `uN` have the same alignment in the reference?

I think there should be.

nikic:
> No weird opsem thing I am aware of (LLVM doesn't even know the sign of our integers most of the time). But the potential for weird architectures is of course unbounded and that's not my expertise. That said, AFAIK the LLVM data layout string cannot even represent sign-dependent alignment so it seems unlikely that LLVM could work on a target where integers with the same size and different sign have different alignment. Cc @nikic

Yes, LLVM data layout is sign independent. Technically, Clang could use different alignment based on signedness, in much the same way `__int128` and `_BitInt(128)` have different alignment despite being the same `i128` type on the LLVM IR level. I'm not aware of any targets where this would be the case though, and it seems pretty unlikely to me. Platform ABIs are crazy, but not *that* crazy.

nikomatsakis:
@rfcbot reviewed

god I hope this is not controversial

tmandry:
@rfcbot reviewed

scottmcm:
From the lang meeting: please also make another PR to add the `∀N, alignof(iN) == alignof(uN)` rule.

(Rather than needing to change this one that's already in FCP.)

joshlf:
Put up https://github.com/rust-lang/reference/pull/2205 to guarantee `∀N, alignof(iN) == alignof(uN)`.

chorman0773:
FWIW, I believe both C and C++ require all standard integer types with the same width to have the same size and alignment. C++ certainly does since 20. So there will be very few "Weird" platforms that have a problem with this.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2200/c49e89cc8c7c2c43ca625a8d5b7ad9a53a9ce978..dc2906c61a4b206fb4f44e0ccd20b4dad8f874ee)*

PR #2202, opened 2026-03-06
Values and `Sized` types are not larger than `isize::MAX`

joshlf:
See [#t-opsem > Constraints on abstract values? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Constraints.20on.20abstract.20values.3F/near/577798059) for context.

For `Sized` types, we *nearly* already guarantee this by implication. If we assume that [`Layout::new<T>() -> Layout`](https://doc.rust-lang.org/std/alloc/struct.Layout.html#method.new) can't fail to compile or panic at runtime, then it implies that, for any `T: Sized`, it is possible to acquire a `Layout` describing that `T`. `Layout`, in turn, guarantees that:

> The size, when rounded up to the nearest multiple of `align`, does not overflow `isize` (i.e., the rounded value will always be less than or equal to `isize::MAX`).

This is also already guaranteed for any [Rust allocation](https://doc.rust-lang.org/stable/std/ptr/index.html#allocation), although in theory it's possible that:
- A value could live somewhere *other than* an allocation
- A type could be compiled that is never instantiated, and thus the allocation constraint doesn't constrain the size of the type itself


joshlf:
r? @ehuss

cc @RalfJung

src/type-layout.md
```diff
@@ -21,7 +21,7 @@ r[layout.properties.size]
...
 
 r[layout.properties.sized]
-Types where all values have the same size and alignment, and both are known at compile time, implement the [`Sized`] trait and can be checked with the [`size_of`] and [`align_of`] functions. Types that are not [`Sized`] are known as [dynamically sized types]. Since all values of a `Sized` type share the same size and alignment, we refer to those shared values as the size of the type and the alignment of the type respectively.
+Types where all values have the same size and alignment, and both are known at compile time, implement the [`Sized`] trait and can be checked with the [`size_of`] and [`align_of`] functions. Types that are not [`Sized`] are known as [dynamically sized types]. Since all values of a `Sized` type share the same size and alignment, we refer to those shared values as the size of the type and the alignment of the type respectively. The size of a `Sized` type is guaranteed not to exceed `isize::MAX`.
```
programmerjake: maybe also add the same guarantee for dynamically-sized types

joshlf: How do you think that guarantee should be phrased?

I suppose we could just *directly* say that "all values are guaranteed to have a size which doesn't exceed `isize::MAX`" and then the text that I've added in this PR is redundant *and* DSTs are covered as well.

programmerjake: > I suppose we could just _directly_ say that "all values are guaranteed to have a size which doesn't exceed `isize::MAX`" and then the text that I've added in this PR is redundant _and_ DSTs are covered as well.

technically there are extern types which don't have a known size, but their actual size should still be `<= isize::MAX`. the wording seems good enough.

joshlf: Actually I fell into the trap I mentioned above regarding `Layout::new` diverging! We still need to guarantee that any sized type has size `<= isize::MAX` because the *value* language only implies that constraint *if the type is instantiated*, which it might not be in a given program.

joshlf: Okay take a look at the updated text, which guarantees this for both values *and* `Sized` types.

programmerjake (approved):
lgtm!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2202/c49e89cc8c7c2c43ca625a8d5b7ad9a53a9ce978..3a8400e6be5db6ee37b339f117d018e5028e9d8e)*

src/type-layout.md
```diff
@@ -18,10 +18,10 @@ r[layout.properties.align]
...
 
 r[layout.properties.sized]
-Types where all values have the same size and alignment, and both are known at compile time, implement the [`Sized`] trait and can be checked with the [`size_of`] and [`align_of`] functions. Types that are not [`Sized`] are known as [dynamically sized types]. Since all values of a `Sized` type share the same size and alignment, we refer to those shared values as the size of the type and the alignment of the type respectively.
+Types where all values have the same size and alignment, and both are known at compile time, implement the [`Sized`] trait and can be checked with the [`size_of`] and [`align_of`] functions. Types that are not [`Sized`] are known as [dynamically sized types]. Since all values of a `Sized` type share the same size and alignment, we refer to those shared values as the size of the type and the alignment of the type respectively. The size of a `Sized` type is guaranteed not to exceed `isize::MAX`. (Note that this guarantee is distinct from the guarantee that the size of a *value* does not exceed `isize::MAX` -- this guarantee applies *even if a `Sized` type is never instantiated*.)
```
programmerjake:
```suggestion
Types where all values have the same size and alignment, and both are known at compile time, implement the [`Sized`] trait and can be checked with the [`size_of`] and [`align_of`] functions. Types that are not [`Sized`] are known as [dynamically sized types]. Since all values of a `Sized` type share the same size and alignment, we refer to those shared values as the size of the type and the alignment of the type respectively. The size of a `Sized` type is guaranteed not to exceed `isize::MAX`. (Note that this guarantee is distinct from the guarantee that the size of a *value* does not exceed `isize::MAX` -- this guarantee applies *even if a `Sized` type is never instantiated*.)
```

PR #2203, opened 2026-03-10
Glossary: add new entry documenting zero-sized types

DanielEScherzer:


src/glossary.md
```diff
@@ -211,9 +211,30 @@ r[glossary.uninhabited]
 
 A type is uninhabited if it has no constructors and therefore can never be instantiated. An uninhabited type is "empty" in the sense that there are no values of the type. The canonical example of an uninhabited type is the [never type] `!`, or an enum with no variants `enum Never { }`. Opposite of [Inhabited](#inhabited).
 
+r[glossary.zst]
+### Zero-sized type (ZST)
+
+A type is zero-sized (a ZST) if its size is 0. Such types have only one possible value. Examples include:
```
traviscross: > Such types have only one possible value.

We may declare the never type (i.e., the empty type) to be a ZST.  So it may be more correct to say that a ZST has at most one possible value.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/a65a85110a50d312de798be5da6c5d1624a49a40..a46e8fa0bcc0519da377189928f61904de0c2c8e)*

DanielEScherzer: Should that be added even before the stabilization of the never type?

traviscross: The never type is already stable.  It just can't always be named easily.

DanielEScherzer: Ah, okay, I misunderstood - switched to "at most one possible value" but I didn't add the never type to the examples since the never type page doesn't currently guarantee that it is a ZST (I'll file a follow-up to add that guarantee at the same time as the follow-up for unit-like enums)

src/glossary.md
```diff
@@ -211,9 +211,30 @@ r[glossary.uninhabited]
 
 A type is uninhabited if it has no constructors and therefore can never be instantiated. An uninhabited type is "empty" in the sense that there are no values of the type. The canonical example of an uninhabited type is the [never type] `!`, or an enum with no variants `enum Never { }`. Opposite of [Inhabited](#inhabited).
 
+r[glossary.zst]
+### Zero-sized type (ZST)
+
+A type is zero-sized (a ZST) if its size is 0. Such types have at most one possible value. Examples include:
```
traviscross:
```suggestion
A type is zero sized (a ZST) if its size is 0. Such types have at most one possible value. Examples include:
```

In predicate position, we wouldn't hyphenate.  (Below, it is correct to hyphenate since there it's used as a phrasal adjective.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/18d42468b25d69a223aaee237616265aa6ec5be7..2fb38fc9c5f8538f4490f2931624ae22eefb92cd)*

src/glossary.md
```diff
@@ -211,9 +211,30 @@ r[glossary.uninhabited]
...
+
+- The [unit type] (see [layout.tuple.unit]).
+- [Arrays] of zero-sized types (see [layout.array]).
```
traviscross, line 225: Zero-length arrays are also ZSTs.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/18d42468b25d69a223aaee237616265aa6ec5be7..2fb38fc9c5f8538f4490f2931624ae22eefb92cd)*

src/glossary.md
```diff
@@ -211,9 +211,68 @@ r[glossary.uninhabited]
...
+- `#[repr(transparent)]` structs with no fields ([unit-like structs]) or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
+- [Arrays] of zero-sized types (see [layout.array]).
+- [Arrays] of length zero (see [layout.array]).
+- [Unions] of zero-sized types (see [items.union.common-storage]).
+
+```rust
+# use core::mem::{size_of, size_of_val};
+fn f() {}
+struct S(u8);
+enum E{ V(u8) }
```
traviscross: Please add the missing space here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/glossary.md
```diff
@@ -211,9 +211,68 @@ r[glossary.uninhabited]
...
 
+r[glossary.zst]
+### Zero-sized type (ZST)
+
+A type is zero sized (a ZST) if its size is 0. Such types have at most one possible value. Examples include:
+
+- The [unit type] (see [layout.tuple.unit]).
+- [Function items] (see [type.fn-item.intro]).
+- The constructors of tuple-like structs (see [type.fn-item.intro]).
+- The constructors of tuple-like enum variants (see [type.fn-item.intro]).
```
traviscross: Please link "tuple-like structs" and "tuple-like enum variants".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/glossary.md
```diff
@@ -211,9 +211,68 @@ r[glossary.uninhabited]
...
+- `#[repr(C)]` structs with no fields ([unit-like structs]) or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
+- `#[repr(transparent)]` structs with no fields ([unit-like structs]) or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
```
traviscross: In prose, we'd just say "`repr(C)` struct" rather than giving the attribute syntax.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/glossary.md
```diff
@@ -211,9 +211,68 @@ r[glossary.uninhabited]
...
+union U {
+    f1: (),
+    f2: [(); 10],
+    f3: [u8; 0],
+}
+assert_eq!(0, size_of::<()>());
+assert_eq!(0, size_of_val(&f));
+// Note that here we are checking the size of the constructors, *not* the
+// underlying type, for `S` and `E::V`. The constructors just have the same
+// names as the types.
```
traviscross: This could use some rewording.  This is value position, so it's syntactically clear the name doesn't refer here to a type.  Also, `E::V` isn't a type, so it's not correct to say that the constructors have the same name as the types.

Since the corresponding values couldn't be constructed in this way (due to needing arguments), it might be OK to just drop this comment.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/special-types-and-traits.md
```diff
@@ -246,6 +246,7 @@ These implicit `Sized` bounds may be relaxed by using the special `?Sized` bound
 [test functions]: attributes/testing.md#the-test-attribute
 [the standard library]: std
 [trait object]: types/trait-object.md
+[zero-sized]: glossary.zst
 [Tuples]: types/tuple.md
 [Type parameters]: types/parameters.md
 [variance]: subtyping.md#variance
```
traviscross: The Reference isn't perfecly consistent about this, but in this case, the sorting is insensitive to case.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/type-layout.md
```diff
@@ -561,6 +561,10 @@ Because this representation delegates type layout to another type, it cannot be
 [enumerations]: items/enumerations.md
 [zero-variant enums]: items/enumerations.md#zero-variant-enums
 [undefined behavior]: behavior-considered-undefined.md
+[zero-sized]: glossary.zst
+[zero-sized type]: glossary.zst
+[zero-sized types]: glossary.zst
+[zero-sized structs]: glossary.zst
```
traviscross: "zero-sized structs" would sort before "zero-sized type".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2203/55184c7049c68a2231d3a8c640543ccd5511185e..dc4cf6551d0db53f997b544da781af63b138f3d0)*

src/glossary.md
DanielEScherzer: Done, also for `repr(transparent)`

src/glossary.md
DanielEScherzer: removed

traviscross (approved):

traviscross:
Thanks @DanielEScherzer, and thanks for joining our calls and iterating on this with us.

PR #2204, opened 2026-03-11
Actually move out of Box in moving-out-of-Box test

dianne:
I looked at https://github.com/rust-lang/reference/pull/2075/ again after it merged and noticed a test I wrote there is wrong; oops! `let _ = $place;` doesn't actually perform a move; a place-to-value coercion is needed.

### DIFF
--- a/src/expressions/operator-expr.md
+++ b/src/expressions/operator-expr.md
@@ -180,13 +180,13 @@ On non-pointer types `*x` is equivalent to `*std::ops::Deref::deref(&x)` in an [
 
 ```rust
 # struct NoCopy;
-let x = &7;
-assert_eq!(*x, 7);
-let y = &mut 9;
-*y = 11;
-assert_eq!(*y, 11);
-let z = Box::new(NoCopy);
-let _: NoCopy = *z;
+let a = &7;
+assert_eq!(*a, 7);
+let b = &mut 9;
+*b = 11;
+assert_eq!(*b, 11);
+let c = Box::new(NoCopy);
+let d: NoCopy = *c;
 ```
 
 r[expr.try]


src/expressions/operator-expr.md
```diff
@@ -186,7 +186,7 @@ let y = &mut 9;
 *y = 11;
 assert_eq!(*y, 11);
 let z = Box::new(NoCopy);
-let _: NoCopy = *z;
+let _z: NoCopy = *z;
```
dianne: `drop::<NoCopy>(*z)` or `{ *z }` would also get the job done here. Maybe the `drop` would be the most explicit? Using a pattern binding is a bit more subtle; I expect it may not be obvious if you're not deeply in the weeds already that `_z` differs from `_`.

ehuss: I think to maintain consistency with the rest of the example above, it's fine to just assign to a local.

Since we have kind of run out of letters, and to avoid any confusion around `_` prefixes, how about just using different variable names here? Such as `a` and `b` or whatever.

dianne: Using `a`, `b`, etc. sounds fine to me; I've done that (after slightly misinterpreting and first renaming `_z` to `a` ^^). I broke the pattern intentionally since that assignment is part of the example for `Box` (I figured each reference/pointer would get one letter), but I agree about avoiding `_` prefixes. In any case, I don't have a strong preference.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2204/91f144460fcb93bed2b580887aa59c28d662a10b..95aaeee979fc7339d76ca75f9f38386dc1c86537)*

PR #2205, opened 2026-03-11
Guarantee size and alignment of more integer primitives

joshlf:
Follows up on https://github.com/rust-lang/reference/pull/2200#issuecomment-4040294894

As an optional extension, we could consider guaranteeing equality between the alignments of:
- `u32` and `char`
- `u32` and `f32`
- `u64` and `f64`

Those equalities strike me as less important, and I'm not sure if they're even valid (ie, whether they hold on all platforms).

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -49,7 +49,7 @@ r[layout.primitive.size-align]
 `usize` and `isize` have the same size and alignment.
 
 r[layout.primitive.align]
-The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8. Alignment is guaranteed to be the same for signed and unsigned variants – that is, for a given `N`, `align_of::<uN>() == align_of::<iN>()`.
+The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8. Alignment is guaranteed to be the same for signed and unsigned variants -- that is, for a given `N`, `align_of::<uN>() == align_of::<iN>()`.
 
 r[layout.pointer]
 ## Pointers and references layout
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -49,7 +49,7 @@ r[layout.primitive.size-align]
 `usize` and `isize` have the same size and alignment.
 
 r[layout.primitive.align]
-The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.
+The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8. Alignment is guaranteed to be the same for signed and unsigned variants – that is, for a given `N`, `align_of::<uN>() == align_of::<iN>()`.
 
 r[layout.pointer]
 ## Pointers and references layout
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -48,8 +48,11 @@ r[layout.primitive.size-minimum]
 r[layout.primitive.size-align]
 `usize` and `isize` have the same size and alignment.
 
-r[layout.primitive.align]
-The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8. Alignment is guaranteed to be the same for signed and unsigned variants -- that is, for a given `N`, `align_of::<uN>() == align_of::<iN>()`.
+r[layout.primitive.platform-specific-alignment]
+The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.
+
+r[layout.primitive.integer-alignment]
+Alignment is guaranteed to be the same for signed and unsigned variants -- that is, for a given `N`, `align_of::<uN>() == align_of::<iN>()`.
 
 r[layout.pointer]
 ## Pointers and references layout
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -52,7 +52,7 @@ r[layout.primitive.platform-specific-alignment]
 The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.
 
 r[layout.primitive.integer-alignment]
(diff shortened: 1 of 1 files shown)


traviscross:
Thanks.  I pushed a clarification and separated out the rules.

Clearly we want to guarantee this for fixed-width integers of the same size.

@rfcbot fcp merge lang

src/type-layout.md
```diff
@@ -45,9 +45,12 @@ The size of most primitives is given in this table.
 r[layout.primitive.size-int]
 `usize` and `isize` have a size big enough to contain every address on the target platform. For example, on a 32 bit target, this is 4 bytes, and on a 64 bit target, this is 8 bytes.
 
-r[layout.primitive.align]
+r[layout.primitive.platform-specific-alignment]
 The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.
 
+r[layout.primitive.integer-alignment]
+Alignment is guaranteed to be the same for signed and unsigned integer variants of the same size --- that is, for a given size `N`, `align_of::<uN>() == align_of::<iN>()`.
```
Darksonn: I would reword "of the same size" to something else. otherwise it sounds like `u64` and `usize` must have the same alignment on 64-bit platforms.

traviscross: > Alignment is guaranteed to be the same for fixed-width signed and unsigned integer variants of the same indicated size --- that is, for a given size `N`, `align_of::<uN>() == align_of::<iN>()`.

How about that?

Darksonn: Did you want to cover usize/isize too?

Darksonn: What about this?

> The following pairs of types have the same alignment:
>
> * u8 and i8
> * u16 and i16
> * u32 and i32
> * u64 and i64
> * u128 and i128
> * usize and isize

Darksonn: What about this?

> The alignment of primitives is platform-specific. In most cases, their alignment is equal to their size, but it may be less. In particular, `i128` and `u128` are often aligned to 4 or 8 bytes even though their size is 16, and on many 32-bit platforms, `i64`, `u64`, and `f64` are only aligned to 4 bytes, not 8.
>
> Furthermore, it is guaranteed that:
> 
> * u8 and i8 have the same alignment.
> * u16 and i16 have the same alignment.
> * u32 and i32 have the same alignment.
> * u64 and i64 have the same alignment.
> * u128 and i128 have the same alignment.
> * usize and isize have the same alignment.

traviscross: Probably at that point, then, I wonder whether we should be revising the table above to include this somehow rather than reiterating all of the the primitive types again.

joshlf: > Did you want to cover usize/isize too?

Those are covered in https://github.com/rust-lang/reference/pull/2200.

scottmcm:
Good to see this being specifically about for the stated `N`.  I *hope* we don't need `usize` ever aligned differently from the `uN` of the same size, but that's less "well obviously", so sticking with just "the `i` vs `u` doesn't impact alignment" as the guarantee sounds good.

@rfcbot reviewed

joshlf:
Now that the `usize`/`isize` language has merged (https://github.com/rust-lang/reference/pull/2200), the diff in this PR is slightly changed. PTAL and confirm that the updated language and organization still looks good.

programmerjake:
I created a minimal `#![no_core]` rust program that compiles when alignments match: it checks `u32`=`char`, `u32`=`f32`, and `u64`=`f64`
<details>

<summary>Source code</summary>

```rust
#![feature(intrinsics)]
#![feature(rustc_attrs)]
#![allow(internal_features)]
#![feature(lang_items)]
#![feature(no_core)]
#![no_core]

#[lang = "pointee_sized"]
pub trait PointeeSized {}

#[lang = "meta_sized"]
pub trait MetaSized: PointeeSized {}

#[lang = "sized"]
pub trait Sized {}

#[lang = "clone"]
pub trait Clone: Sized {
    fn clone(&self) -> Self;
}

#[lang = "copy"]
pub trait Copy: Clone {}

#[lang = "legacy_receiver"]
pub trait LegacyReceiver {}

impl<T: ?Sized> LegacyReceiver for &'_ T {}

#[rustc_nounwind]
#[rustc_intrinsic]
pub const fn align_of<T>() -> usize;

const _: [(); align_of::<u32>()] = [(); align_of::<char>()];
const _: [(); align_of::<u32>()] = [(); align_of::<f32>()];
const _: [(); align_of::<u64>()] = [(); align_of::<f64>()];
```

</details>

I checked with rustc 1.94.0 with:
```bash
for target in `rustc --print=target-list`; do RUSTC_BOOTSTRAP=1 cargo check --target="$target" || echo "failed on $target"; done
```

that gives failures on `m68k-unknown-linux-gnu`, `m68k-unknown-none-elf`, and `powerpc64-ibm-aix`

<details>

<summary>Compiler errors</summary>

```text
    Checking align-test v0.0.0 (/tmp/align-test)
error[E0308]: mismatched types
  --> src/lib.rs:35:36
   |
35 | const _: [(); align_of::<u32>()] = [(); align_of::<f32>()];
   |               -----------------    ^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a size of 2, found one with a size of 4
   |               |
   |               help: consider specifying the actual array length: `4`

error[E0308]: mismatched types
  --> src/lib.rs:36:36
   |
36 | const _: [(); align_of::<u64>()] = [(); align_of::<f64>()];
   |               -----------------    ^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a size of 4, found one with a size of 8
   |               |
   |               help: consider specifying the actual array length: `8`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `align-test` (lib) due to 2 previous errors
failed on m68k-unknown-linux-gnu
    Checking align-test v0.0.0 (/tmp/align-test)
error[E0308]: mismatched types
  --> src/lib.rs:35:36
   |
35 | const _: [(); align_of::<u32>()] = [(); align_of::<f32>()];
   |               -----------------    ^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a size of 2, found one with a size of 4
   |               |
   |               help: consider specifying the actual array length: `4`

error[E0308]: mismatched types
  --> src/lib.rs:36:36
   |
36 | const _: [(); align_of::<u64>()] = [(); align_of::<f64>()];
   |               -----------------    ^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a size of 4, found one with a size of 8
   |               |
   |               help: consider specifying the actual array length: `8`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `align-test` (lib) due to 2 previous errors
failed on m68k-unknown-none-elf
    Checking align-test v0.0.0 (/tmp/align-test)
error[E0308]: mismatched types
  --> src/lib.rs:36:36
   |
36 | const _: [(); align_of::<u64>()] = [(); align_of::<f64>()];
   |               -----------------    ^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a size of 8, found one with a size of 4
   |               |
   |               help: consider specifying the actual array length: `4`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `align-test` (lib) due to 1 previous error
failed on powerpc64-ibm-aix
```

programmerjake:
so, I think we can guarantee that `u32` and `char` have the same alignment, but not integers and the same-sized floating-point types.

RalfJung:
We already [guarantee](https://doc.rust-lang.org/nightly/std/primitive.fn.html#abi-compatibility) that u32 and char are ABI-compatible, that requires their alignment to be the same. We [test](https://github.com/rust-lang/rust/blob/d8ed639e3033d038220db73b087b61dcc4774dc0/tests/ui/abi/compatibility.rs#L195) for that as well.

And yeah float types  are definitely different from integer types in various ways, we should not guarantee anything there.

RalfJung:
FWIW we also guarantee that usize is ABI-compatible with the equal-sized unsigned int, and similar for isize (and we test for that). So unless we want to take that back, the ship has sailed for targets where that wouldn't be true.

joshlf:
Okay, updated to guarantee size/alignment equality for:
- `usize`, `isize`, `uN`, and `iN`
- `uN` and `iN`
- `char` and `u32`

traviscross (approved):

traviscross:
We're talking about this one on the lang-docs call.  @joshlf, you had added some commits expanding the guarantee based on feedback after the lang team had reviewed and approved this.  If you don't mind, let's handle those in a separate PR so that we can cleanly ensure lang-team signoff.  I've rolled this back to the commit the lang team reviewed and rebased (and backed up your other work; let me know if you need that).  Given the completed FCP, let's merge that.

PR #2206, opened 2026-03-11
Add a rule that enum discriminants may not use generic parameters

ehuss:
This is implemented in the compiler with the diagnostic [`ParamInEnumDiscriminant`](https://github.com/rust-lang/rust/blob/3b1b0ef4d80d3117924d91352c8b6ca528708b3c/compiler/rustc_resolve/src/errors.rs#L651-L661) and [`NoConstantGenericsReason::IsEnumDiscriminant`](https://github.com/rust-lang/rust/blob/3b1b0ef4d80d3117924d91352c8b6ca528708b3c/compiler/rustc_resolve/src/late.rs#L160-L168).

### DIFF
--- a/src/items/enumerations.md
+++ b/src/items/enumerations.md
@@ -199,21 +199,20 @@ enum OverflowingDiscriminantError2 {
 r[items.enum.discriminant.restrictions.generics]
 Explicit enum discriminants may not use generic parameters from the enclosing enum.
 
-> [!EXAMPLE]
-> ```rust,compile_fail
-> #[repr(u32)]
-> enum E<'a, T, const N: u32> {
->     Lifetime(&'a T) = {
->         let a: &'a (); // ERROR: generic parameters may not be used in enum discriminant
->         1
->     },
->     Type(T) = {
->         let x: T; // ERROR: generic parameters may not be used in enum discriminant
->         2
->     },
->     Const = N, // ERROR: generic parameters may not be used in enum discriminant
-> }
-> ```
+```rust,compile_fail
+#[repr(u32)]
+enum E<'a, T, const N: u32> {
+    Lifetime(&'a T) = {
+        let a: &'a (); // ERROR.
+        1
+    },
+    Type(T) = {
+        let x: T; // ERROR.
+        2
+    },
+    Const = N, // ERROR.
+}
+```
 
 ### Accessing discriminant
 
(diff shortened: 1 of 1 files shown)


src/items/enumerations.md
```diff
@@ -196,6 +196,25 @@ enum OverflowingDiscriminantError2 {
 }
 ```
 
+r[items.enum.discriminant.restrictions.generics]
+Explicit enum discriminants may not use generic parameters from the enclosing enum.
```
DanielEScherzer: > from the enclosing enum.

What about from anywhere else? Playground says that you also can't use from an enclosing item either

```rust
fn demo<const N: u32>() {
    enum Foo {
        Bar = N,
    }
}
```

though it does also have an unhelpful suggestion to
```
help: try introducing a local generic parameter here
  |
2 |     enum Foo<N> {
  |             +++
```

which then leads to

* same "can't use generic parameters from outer item" error
* with a new suggestion
```
help: try introducing a local generic parameter here
  |
2 |     enum Foo<N, N> {
  |              ++
```
* and also an error ``type parameter `N` is never used``

---
I suggest

```suggestion
Explicit enum discriminants may not use generic parameters.
```

DanielEScherzer (changes requested):

src/items/enumerations.md
ehuss: The restriction that generic parameters are not in scope within items in a function body is covered by [`items.generics.syntax.scope`](https://doc.rust-lang.org/nightly/reference/items/generics.html#r-items.generics.syntax.scope).

The reason it uses the "enclosing enum" terminology is because generic parameters defined within the enum discriminant initializer may be used, as in:

```rust
enum E {
    V1 = {
        const fn f<T>(x: T) -> T { x }
        f::<isize>(123)
    }
}
```

I didn't want to risk potential confusion around that.

DanielEScherzer: But the actual discriminant of `f::<isize>(123)` is not based on any generic parameter, just the invocation of a constant function that happens to be generic, same as
```rust
const fn f<T>(x: T) -> T { x }
enum E {
    V1 = {
        f::<isize>(123)
    }
}
```

what about

> Explicit enum discriminants may not use generic parameters. Note that they may still invoke generic functions.

or something like that?

traviscross (approved):

traviscross:
Looks good to me.  Thanks.

PR #2214, opened 2026-03-21
Improve description of `ident` fragment matcher, and `$crate` metavariable

### LINKED ISSUE #588: Document `$crate` in macro matcher

steffahn:
I am initially coming to work on this due to the line
> an IDENTIFIER_OR_KEYWORD except `_`, RAW_IDENTIFIER, or `$crate`

being unfortunately very ambiguous, since this can be read as “`_`, RAW_IDENTIFIER, or `$crate`” all belonging
to the “except” listing.

My best fix for that is to add parentheses around the “except `_`”.

> an IDENTIFIER_OR_KEYWORD (except `_`), a RAW_IDENTIFIER, or `$crate`

This seems to read more nicely than alternatives I have tried, for instance

> <mark>either</mark> an IDENTIFIER_OR_KEYWORD except `_`, <mark>or</mark> a RAW_IDENTIFIER, or "`$crate`"

*(which still seems ambiguous)*

And the exception for “`_`” seems minor enough (previously it had been missing completely for a while) that parentheses also seem sensible semantically, IMHO.

The meaning of accepting `$crate` was confusing to me though; and apparently a known issue, so this closes rust-lang/reference#588. (I think. Unless that issue wants *even more* information; or information in some particular other place?)

---

To achieve this, I have written a new section to the reference description of `$crate`, and worded the concept as `$crate` being “initially replaced”.

- The verb “replace” is for consistency with the wording I could find around the handling of other metavariables, which are being “replaced” during expansion;
- the qualifier “initially” is added to emphasize the important difference of this process from *macro expansion*. And also where I’m linking to it, the phrasing “replaced `$crate`” can easily be misunderstood for people to think that “`$crate`” has already been replaced by the name of the crate it’s referring to, even though that’s not at all how this process works!

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -109,7 +109,7 @@ Valid fragment specifiers are:
   * `block`: a [BlockExpression]
   * `expr`: an [Expression]
   * `expr_2021`: an [Expression] except [UnderscoreExpression] and [ConstBlockExpression] (see [macro.decl.meta.edition2024])
-  * `ident`: an [IDENTIFIER_OR_KEYWORD] except `_`, [RAW_IDENTIFIER], or [`$crate`]
+  * `ident`: an [IDENTIFIER_OR_KEYWORD] (except `_`), a [RAW_IDENTIFIER], or an [initially replaced `$crate`]
   * `item`: an [Item]
   * `lifetime`: a [LIFETIME_TOKEN]
   * `literal`: matches `-`<sup>?</sup>[LiteralExpression]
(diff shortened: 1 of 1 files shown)


src/macros-by-example.md
```diff
@@ -664,6 +664,48 @@ pub mod inner {
 }
 ```
 
+r[macro.decl.hygiene.crate.replaced]
+The raw syntax of `$crate` consists of two tokens (`$` followed by `crate`). When used as a metavariable within a macro definition, this syntax is initially replaced with a single token (also called&nbsp;"`$crate`"), which can be used an identifier.
```
ehuss, line 668: Can you say why this has an `&nbsp;`?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2214/7446bf9697c95d155eef33c6a9d91fbd29a5e359..b7838ba017b52affbbf3d9105468697176fd310b)*

src/macros-by-example.md
```diff
@@ -676,6 +718,8 @@ macro_rules! call_foo {
 fn foo() {}
 ```
 
+However, the crate being referred to does *not* itself need to be visible from the invocation site as a directly declared dependency. A main purpose of `$crate` is to offer a way of reliably naming crates (and their public items) in macro-generated code, even if the only exists as a transitive dependency (i.e. "dependency of a dependency") from the invocation site.
```
ehuss, line 721: I think this is what was intended?

```suggestion
However, the crate being referred to does *not* itself need to be visible from the invocation site as a directly declared dependency. A main purpose of `$crate` is to offer a way of reliably naming crates (and their public items) in macro-generated code, even if it only exists as a transitive dependency (i.e. "dependency of a dependency") from the invocation site.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2214/7446bf9697c95d155eef33c6a9d91fbd29a5e359..b7838ba017b52affbbf3d9105468697176fd310b)*

ehuss: Or maybe "the crate"?

ehuss:
Just for clarity, I also wanted to link to https://github.com/rust-lang/rust/pull/155121 as there are active changes happening in this area. I don't think that affects this PR. I'm not sure if you're already aware of those changes.

PR #2215, opened 2026-03-22
stabilize s390x vector registers

folkertdev:
cc https://github.com/rust-lang/rust/pull/154184

### DIFF
--- a/src/inline-assembly.md
+++ b/src/inline-assembly.md
@@ -318,6 +318,9 @@ unsafe { core::arch::asm!("/* {} */", in(reg) 5); }
 # }
 ```
 
+> [!NOTE]
+> If the value's type is smaller than the register, the value of the upper bits is platform-specific. Some targets zero out the upper bits, while others leave them untouched.
+
 r[asm.operand-type.supported-operands.out]
 * `out(<reg>) <expr>`
   - `<reg>` can refer to a register class or an explicit register. The allocated register name is substituted into the asm template string.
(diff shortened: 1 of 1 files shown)


src/inline-assembly.md
```diff
@@ -626,7 +629,7 @@ Here is the list of currently supported register classes:
 | s390x | `reg` | `r[0-10]`, `r[12-14]` | `r` |
 | s390x | `reg_addr` | `r[1-10]`, `r[12-14]` | `a` |
 | s390x | `freg` | `f[0-15]` | `f` |
-| s390x | `vreg` | `v[0-31]` | Only clobbers |
+| s390x | `vreg` | `v[0-31]` | `v` |
```
folkertdev, line 632: e.g. in https://godbolt.org/z/vxeWTe5KG

src/inline-assembly.md
```diff
@@ -318,6 +318,9 @@ unsafe { core::arch::asm!("/* {} */", in(reg) 5); }
 # }
 ```
 
+> [!NOTE]
+> If the value's type is smaller than the register, the value of the upper bits is platform-specific. Some targets zero out the upper bits, while others leave them untouched.
```
folkertdev, line 322: This came up in the stabilization PR. Amanieu mentions that aarch64 zeros the upper bits, while s390x does not touch them.

This might need some re-phrasing, but I think it's useful to include to prevent future confusion.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2215/7446bf9697c95d155eef33c6a9d91fbd29a5e359..41a2fc816ad74ac14094ea925f9cddfa45f1bc87)*

PR #2213, opened 2026-03-20
Update reference for attribute order changes

JonathanBrouwer:

PR #2216, opened 2026-03-24
Update link_name duplicates note

ehuss:
The behavior of the `link_name` attribute when there are duplicates has changed via https://github.com/rust-lang/rust/pull/153041.


traviscross (approved):

traviscross:
See also:

- https://github.com/rust-lang/reference/pull/2213

PR #2220, opened 2026-04-03
Add section about inner tool attributes

nik-rev:
For the stabilization PR: https://github.com/rust-lang/rust/pull/154770


ehuss:
It looks like the restriction on inner attributes isn't documented. So, I think the end result is that the reference doesn't need to change here. In other words, the reference is currently incorrect, and changing the compiler to allow tool attributes would make it correct.

Proc-macro attributes do say they can only be outer attributes. But I don't think that sufficiently covers the entirety of "custom inner attributes", which seems to also include tool attributes.

PR #2221, opened 2026-04-04
Clarify that `{self}` imports require a module parent

traviscross:
The `self` keyword in a use brace (e.g., `use m::{self}`) creates a binding for the parent entity.  The note in `items.use.self.intro` says that `self` "means the current module of the parent segment", but there's no rule restricting what the parent can be.

The parent path must resolve to a module, enumeration, or trait -- i.e., entities that act as modules for name resolution.  Using `self` with other entities such as structs or unions is rejected by the compiler (as of rust-lang/rust#152996).

Let's add a rule to make this restriction explicit.

See:

- https://github.com/rust-lang/rust/pull/152996

cc @ehuss @mu001999 @petrochenkov

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -203,6 +203,26 @@ mod example {
 > [!NOTE]
 > `self` may also be used as the first segment of a path. The usage of `self` as the first segment and inside a `use` brace is logically the same; it means the current module of the parent segment, or the current module if there is no parent segment. See [`self`] in the paths chapter for more information on the meaning of a leading `self`.
 
+r[items.use.self.module]
+When `self` is used within [brace syntax], the path preceding the brace group must resolve to a [module], [enumeration], or [trait].
+
+```rust
+mod m {
+    pub enum E { V1, V2 }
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2221/d2715c07e9dd9839c0c7675ecfa18bec539a6ee9..d5687ff08d346e6712a4e3ff8447db62dfc02a08)*

tshepang:
should this be placed in `1.96` milestone given that the restriction first appeared in Rust 1.96

traviscross:
It should correspond to the version of the Reference in which we shipped the text.  If we shipped it later than the language did, then the milestone will differ from that of the stabilization.

tshepang:
does shipping Reference mean adding it to /rust monorepo

traviscross:
> does shipping Reference mean adding it to /rust monorepo

Right; the commit of the Reference pointed at by the submodule pointer when a version of Rust was released.

PR #2222, opened 2026-04-04
Fix misuse of *usage* throughout the Reference

traviscross:
*Usage* means a customary pattern of use (as in "regional usage varies") or a rate or quantity of use (as in "disk usage").  It doesn't mean *use*.  Whenever *use* works, *usage* shouldn't appear.

Let's fix each incorrect use of *usage* -- i.e., adopt better *usage* of *usage* -- throughout the Reference.

cc @ehuss

### DIFF
--- a/src/attributes.md
+++ b/src/attributes.md
@@ -223,7 +223,7 @@ The following is an index of all built-in attributes.
   - [`automatically_derived`] --- Marker for implementations created by `derive`.
 
 - Macros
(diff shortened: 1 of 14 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2222/d2715c07e9dd9839c0c7675ecfa18bec539a6ee9..f98bfa3f4f46be45e80f000dfd0dbede9024ef6d)*

PR #2223, opened 2026-04-10
Update `derive` export locations

ehuss:
The `derive` export locations were changed in https://github.com/rust-lang/rust/pull/154442.

I'm a little on the fence listing them out this way, but I think I would prefer to be explicit.


traviscross (approved):

traviscross:
Thanks.

PR #2224, opened 2026-04-11
parens are not optional in guard chains that have || operators

tshepang:


src/expressions/if-expr.md
```diff
@@ -186,7 +186,7 @@ fn nested() {
 ```
 
 r[expr.if.chains.or]
-If any condition operand is a `let` pattern, then none of the condition operands can be a `||` [lazy boolean operator expression][expr.bool-logic] due to ambiguity and precedence with the `let` scrutinee. If a `||` expression is needed, then parentheses can be used. For example:
+If any condition operand is a `let` pattern, then none of the condition operands can be a `||` [lazy boolean operator expression][expr.bool-logic] due to ambiguity and precedence with the `let` scrutinee. If a `||` expression is needed, then parentheses must be used. For example:
```
ehuss: I don't think it is correct to use "must" here. This is saying that this is an option for how to work around the limitation. For example, you could sometimes use curly braces, or lift parts out to a variable, or don't use chains.

I think the correct change would be to move this statement out to an example block like the other one, since it isn't a language rule.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2224/8d1af73c0e144e5faf3f52ece5dfa38b2b710710..e81dd7cd671e929fd5e914d7358d7a37419b5838)*

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2224/581920f9109f141b88b860b3e1e8359e3896a150..d1fb175e68072d131eefa9381da1adcd71e2d306)*

PR #2226, opened 2026-04-13
Document diagostic attribute duplicates

mejrs:
You can stack these attributes, which I think is a cool little known fact. For example:
```
#[diagnostic::on_unimplemented(
    message = "pyclass `{Self}` cannot be subclassed",
    label = "required for `#[pyclass(extends={Self})]`",
    note = "`{Self}` must have `#[pyclass(subclass)]` to be eligible for subclassing"
)]
#[cfg_attr(
    Py_LIMITED_API,
    diagnostic::on_unimplemented(
        note = "with the `abi3` feature enabled, PyO3 does not support subclassing native types",
    )
)]
pub trait PyClassBaseType: Sized {
```

See https://github.com/rust-lang/rust/pull/155294 for the test this rule will link to.

### DIFF
--- a/src/attributes/diagnostics.md
+++ b/src/attributes/diagnostics.md
@@ -575,6 +575,9 @@ The `note` option can appear several times, which results in several note messag
 r[attributes.diagnostic.on_unimplemented.repetition]
 If any of the other options appears several times the first occurrence of the relevant option specifies the actually used value. Subsequent occurrences generates a warning.
 
+r[attributes.diagnostic.on_unimplemented.duplicates]
+The `diagnostic::on_unimplemented` attribute may be used any number of times. All keys from all uses of the attribute will be used and are subject to the repetition rules.
+
 r[attributes.diagnostic.on_unimplemented.unknown-keys]
 A warning is generated for any unknown keys.
 


src/attributes/diagnostics.md
```diff
@@ -575,6 +575,9 @@ The `note` option can appear several times, which results in several note messag
 r[attributes.diagnostic.on_unimplemented.repetition]
 If any of the other options appears several times the first occurrence of the relevant option specifies the actually used value. Subsequent occurrences generates a warning.
 
+r[attributes.diagnostic.on_unimplemented.duplicates]
+The `diagnostic::on_unimplemented` attribute may be used any number of times. All keys from all uses of the attribute will be used and are subject to the repetition rules.
```
mejrs, line 579: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2226/e7c00218c937fb3a0df4fc3cbb2fbfda73675777..a432709ef2943c2d96f249649f9d2057cd8c9550)*

Existing prose looks like
```
The `cfg` attribute may be used any number of times on a form.
```
and
```
The `derive` attribute may be used any number of times on an item.
```

I'm not sure what I should use, or what a form is for that matter (it's not in the glossary).

ehuss:
Thanks for the PR! We have an open PR at https://github.com/rust-lang/reference/pull/1893 that updates this rule (among a few other things).

mejrs:
Oh shoot, I didn't see that at all. Great minds think alike, I suppose. Would you like me to merge the linked test?

ehuss:
The test looks good. I'll try to remember to add the reference link later once the other PR is merged.

PR #2227, opened 2026-04-15
enums with fields: clarify that repr(C, int) is quite different from repr(int)

RalfJung:
This is sufficiently subtle and seems worth pointing out.


RalfJung:
@rust-lang/lang-docs I hope this is a fairly easy one since it does not change anything normative :)

JohnTitor (approved):
Looks good!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2227/a65a85110a50d312de798be5da6c5d1624a49a40..7b708f0af4780523553a4bc3e44010425838a5b0)*

PR #2229, opened 2026-04-17
Clarify items.fn.generics.param-bounds

DanielEScherzer:
"methods with that trait" does not really make sense when talking about methods that a trait provides - use "methods from that trait" instead.


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2229/18d42468b25d69a223aaee237616265aa6ec5be7..49135a588b880107fb9ebd3747811d9b30baaca9)*

PR #2231, opened 2026-04-18
place "place projection" definition in glossary

tshepang:
The term is defined in 2 separate places, and the definitions do not match.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -28,8 +28,7 @@ r[undefined.pointer-access]
   a misaligned pointer].
 
 r[undefined.place-projection]
-* Performing a [place projection][glossary.place-projection] that violates the requirements of [in-bounds
-  pointer arithmetic](pointer#method.offset).
+* Performing a [place projection][glossary.place-projection] that violates the requirements of [in-bounds pointer arithmetic](pointer#method.offset).
 
 r[undefined.alias]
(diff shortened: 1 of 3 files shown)


ehuss:
I'm not sure these are intended to be the same.

@RalfJung Is the definition of *place projection* intended to be the same for these two scenarios (closure capture and UB pointer arithmetic)?

RalfJung:
Yeah this changes the opsem page to suddenly include pointer deref's and that doesn't make sense in that context (since it is talking about pointer arithmetic, and a deref isn't pointer arithmetic).

tshepang:
since this is about pointer arithmetic, should *field expression* be removed from the more narrow definiton of *place expression*

RalfJung:
We should probably reword the UB page to use a different term, since it only refers to a few specific place projections.

tshepang:
would this work
> A tuple index expression or an array/slice index expression that violates the (safety) requirements of in-bounds pointer arithmetic.

RalfJung:
You forgot "field expression".

I would suggest something like https://github.com/rust-lang/reference/pull/2253.

tshepang:
> You forgot "field expression".

I actually thought it was accidental, since I assumed that, given offset works only with integers, then only APIs taking integers were connected to this rule.

RalfJung:
Please don't make silent assumptions. Instead, ask for clarification. :)

It was very much deliberate. `(*ptr).field` is UB if `field` has offset N and that offset goes beyond the bounds of the allocation `ptr` is derived from.

PR #2236, opened 2026-04-19
Document `#[link_section = "..."]` format

madsmtm:
The values accepted by the `link_section` attribute is platform-specific, it depends on [the binary format](https://github.com/rust-lang/rust/issues/152586).

On most platforms, it just forwards to the linker to do any special stuff with the name IIUC, but for Mach-O, it's a bit different, there LLVM parses the name and sets various flags depending on what the user specified.

For example, the following:
```rust
#[unsafe(link_section = "__DATA,my_cstring,cstring_literals,no_toc+no_dead_strip")]
static MY_CSTRING: [u8; 1] = *b"\0";
```

Results in a binary containing this data:
```rust
object::macho::Section64 {
    sectname: "__DATA",
    segname: "my_cstring",
    align: 1,
    reloff: 0,
    nreloc: 0,
    flags: object::macho::S_CSTRING_LITERALS | object::macho::S_ATTR_NO_TOC | object::macho::S_ATTR_NO_DEAD_STRIP,
    // ...
}
```

That is, the name used in `#[link_section]` on Mach-O is not inherent to the file format. Therefore, I think we should document what the expected format actually is!

(We should perhaps have had something like `#[link_section(segment = "__DATA", segment = "my_cstring", type = "cstring_literals", attributes = ["no_dead_strip", "no_toc"])]` on Mach-O instead, but that ship has probably sailed a while ago, I don't think it makes sense to change this).

I discovered this while working on https://github.com/rust-lang/rustc_codegen_cranelift/pull/1648, it turns out that `rustc_codegen_cranelift` must parse these in the same manner as LLVM to support things like `ctor`.

Also closely related to https://github.com/rust-lang/rust/pull/155065, it might make sense to expand that parsing to the entire string (to have better control over diagnostics / not be reliant on LLVM changing it).

---

An alternative here would be to:
- Document this [on the Clang attribute](https://clang.llvm.org/docs/AttributeReference.html#section-declspec-allocate) instead. (Or possibly somewhere in LLVM's [LangRef](https://llvm.org/docs/LangRef.html)).
- Document this in Cranelift.
- Link to the support in each of those.

I'd be fine with that too, if you think that's better?

### DIFF
--- a/src/abi.md
+++ b/src/abi.md
@@ -121,6 +121,26 @@ r[abi.link_section.format]
 The format of the `link_section` attribute is platform-specific, and the
 supported options varies depending on the underlying object file format.
 
+r[abi.link_section.elf]
+### Link section ELF format
+
+TODO, maybe document `.hash` -> `SHT_HASH` etc.?
+
+Or perhaps just link to <https://www.man7.org/linux/man-pages/man5/elf.5.html#:~:text=Various%20sections%20hold%20program%20and%20control%20information>?
+
+r[abi.link_section.coff]
+### Link section COFF format
+
+TODO. Something like:
+
+r[abi.link_section.coff.syntax]
+```grammar,attributes
+COFFLinkSection -> IDENTIFIER (`$` COFFOrder)?
+COFFOrder -> IDENTIFIER
+```
+
+See <https://learn.microsoft.com/en-us/windows/win32/debug/pe-format>.
+
 r[abi.link_section.mach-o]
 ### Link section Mach-O format
 
@@ -151,6 +171,16 @@ size (if the `symbol_stubs` section type is used).
 
 r[abi.link_section.mach-o.default-type]
 The section type defaults to `regular` if not set.
+
+r[abi.link_section.wasm]
+### Link section WebAssembly format
+
+TODO.
+
+r[abi.link_section.coff]
+### Link section XCOFF format
+
+TODO.
 
 r[abi.export_name]
 ## The `export_name` attribute
--- a/src/abi.md
+++ b/src/abi.md
@@ -117,6 +117,41 @@ r[abi.link_section.edition2024]
 > [!EDITION-2024]
 > Before the 2024 edition it is allowed to use the `link_section` attribute without the `unsafe` qualification.
 
+r[abi.link_section.format]
+The format of the `link_section` attribute is platform-specific, and the
+supported options varies depending on the underlying object file format.
+
+r[abi.link_section.mach-o]
+### Link section Mach-O format
+
+<!-- Parsing done by LLVM in https://github.com/llvm/llvm-project/blob/llvmorg-22.1.3/llvm/lib/MC/MCSectionMachO.cpp -->
+r[abi.link_section.mach-o.syntax]
+```grammar,attributes
+MachOLinkSection -> MachOSegment `,` MachOSection (`,` (MachOSectionType (`,` MachOSectionAttributes?)?)?)?
+
+MachOLinkSectionStub -> MachOSegment `,` MachOSection `,` `symbol_stubs` `,` MachOSectionAttributes? `,` MachOSectionStubSize
+
+MachOSegment -> <0 to 16 bytes>
+
+MachOSection -> <1 to 16 bytes>
+
+MachOSectionType -> `regular` | `zerofill` | `cstring_literals` | `4byte_literals` | `8byte_literals` | `literal_pointers` | `non_lazy_symbol_pointers` | `lazy_symbol_pointers` | `mod_init_funcs` | `mod_term_funcs` | `coalesced` | `interposing` | `16byte_literals` | `thread_local_regular` | `thread_local_zerofill` | `thread_local_variables` | `thread_local_variable_pointers` | `thread_local_init_function_pointers`
+
+MachOSectionAttributes -> MachOSectionAttribute (`+` MachOSectionAttribute)*
+
+MachOSectionAttribute -> `pure_instructions` | `no_toc` | `strip_static_syms` | `no_dead_strip` | `live_support` | `self_modifying_code` | `debug`
+
+MachOSectionStubSize -> DEC_DIGIT (DEC_DIGIT)*
+```
+
+r[abi.link_section.mach-o.intro]
+The link section format string on Mach-O has five configurable fields: the
+segment, the section, the section type, the section attributes and the stub
+size (if the `symbol_stubs` section type is used).
+
+r[abi.link_section.mach-o.default-type]
+The section type defaults to `regular` if not set.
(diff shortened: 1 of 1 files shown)


src/abi.md
```diff
@@ -117,6 +117,65 @@ r[abi.link_section.edition2024]
...
+
+TODO. Something like:
+
+r[abi.link_section.coff.syntax]
+```grammar,attributes
+COFFLinkSection -> IDENTIFIER (`$` COFFOrder)?
+COFFOrder -> IDENTIFIER
+```
+
+See <https://learn.microsoft.com/en-us/windows/win32/debug/pe-format>.
```
madsmtm, line 142: I'll fill these in properly if you think it makes sense to document this in the reference too?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2236/18d42468b25d69a223aaee237616265aa6ec5be7..2fa4226637a77715ae8eb93d36fb15640b99eec7)*

src/abi.md
```diff
@@ -117,6 +117,71 @@ r[abi.link_section.edition2024]
...
+```grammar,attributes
+MachOLinkSection -> MachOSegment `,` MachOSection (`,` (MachOSectionType (`,` MachOSectionAttributes?)?)?)?
+
+MachOLinkSectionStub -> MachOSegment `,` MachOSection `,` `symbol_stubs` `,` MachOSectionAttributes? `,` MachOSectionStubSize
+
+MachOSegment -> <0 to 16 bytes>
+
+MachOSection -> <1 to 16 bytes>
+
+MachOSectionType -> `regular` | `zerofill` | `cstring_literals` | `4byte_literals` | `8byte_literals` | `literal_pointers` | `non_lazy_symbol_pointers` | `lazy_symbol_pointers` | `mod_init_funcs` | `mod_term_funcs` | `coalesced` | `interposing` | `16byte_literals` | `thread_local_regular` | `thread_local_zerofill` | `thread_local_variables` | `thread_local_variable_pointers` | `thread_local_init_function_pointers`
```
madsmtm, line 158: We can leave out the exact names here if you prefer? Though I tend to think it makes sense to have, for example `mod_init_funcs` is slightly different from what the attribute is documented as in the header (`S_MOD_INIT_FUNC_POINTERS`).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2236/18d42468b25d69a223aaee237616265aa6ec5be7..8446a1904e835642000ecb4cc8ac01b9e358e958)*

src/abi.md
```diff
@@ -117,6 +117,71 @@ r[abi.link_section.edition2024]
...
+COFFLinkSection -> IDENTIFIER (`$` COFFOrder)?
+COFFOrder -> IDENTIFIER
+```
+
+See <https://learn.microsoft.com/en-us/windows/win32/debug/pe-format>.
+
+r[abi.link_section.mach-o]
+### Link section Mach-O format
+
+<!-- Parsing done by LLVM in https://github.com/llvm/llvm-project/blob/llvmorg-22.1.3/llvm/lib/MC/MCSectionMachO.cpp -->
```
madsmtm, line 147: I checked what GCC does here, it seems to pass the section onwards to the assembler unchanged (i.e. forwards the problem to the assembler). On most GCC installations on macOS, the default assembler is LLVM's (so it supports the same syntax as documented here).

A few other assemblers:
- Apple's old, deleted assembler `cctools/as` supported a subset of these in the same format: https://github.com/apple-oss-distributions/cctools/blob/cctools-1010.6/as/read.c#L232-L279
- NASM has it's own (more limited) syntax: https://github.com/netwide-assembler/nasm/blob/aa12d3381a1fd08e72f594147f2486c1b5ab7813/output/outmacho.c#L739-L748 (though they still use the same comma-separated format for segments and sections)


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2236/18d42468b25d69a223aaee237616265aa6ec5be7..8446a1904e835642000ecb4cc8ac01b9e358e958)*

src/abi.md
```diff
@@ -117,6 +117,71 @@ r[abi.link_section.edition2024]
...
+segment, the section, the section type, the section attributes and the stub
+size (if the `symbol_stubs` section type is used).
+
+r[abi.link_section.mach-o.default-type]
+The section type defaults to `regular` if not set.
+
+r[abi.link_section.wasm]
+### Link section WebAssembly format
+
+TODO.
```
daxpedda, line 178:
```suggestion
Link sections must be `static`s with a simple list of bytes with no extra levels of indirection such as references. They are ultimately encoded as Wasm custom sections.

The exception is the `.init_array` link section, which serves as a list for static constructors.
```

I'm not sure if we should mention `.init_array` here at all, considering how its not mentioned for any other targets.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2236/18d42468b25d69a223aaee237616265aa6ec5be7..8446a1904e835642000ecb4cc8ac01b9e358e958)*

madsmtm:
I've opened https://github.com/llvm/llvm-project/pull/209167 to document this in LLVM instead, since I tend to think it makes more sense there. If that lands, I'll update this to be a link to that.

src/abi.md
```diff
@@ -117,6 +117,71 @@ r[abi.link_section.edition2024]
 > [!EDITION-2024]
 > Before the 2024 edition it is allowed to use the `link_section` attribute without the `unsafe` qualification.
 
+r[abi.link_section.format]
+The format of the `link_section` attribute is platform-specific, and the
+supported options varies depending on the underlying object file format.
```
Mark-Simulacrum, line 122: What level of validation does rustc do here and to what extent do we want that to be part of the *language* definition? It seems plausible that we would want to move the details here out of the reference and into (for example) a section in the rustc book (e.g., https://doc.rust-lang.org/nightly/rustc/exploit-mitigations.html).

It seems reasonable to me that a Rust implementation would be complete without the extra validation, though it's highly likely validation *does* provide better diagnostics and so is useful.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2236/18d42468b25d69a223aaee237616265aa6ec5be7..8446a1904e835642000ecb4cc8ac01b9e358e958)*

madsmtm: The only validation I'm aware of is https://github.com/rust-lang/rust/pull/155065, which is useful because Mach-O sections require two components, which often trips people up.

I agree that a Rust implementation would be complete without doing anything extra - the attribute is `unsafe` for a reason, and it is intended to just be forwarded to the assembler / linker / whatever lower-level mechanism.

PR #2237, opened 2026-04-20
Document trailing `self` in paths

traviscross:
The Reference says that `self` can only appear as the first segment of a path.  Further, the `use` chapter describes `{self}` brace syntax as the way to bind the parent entity of a `use` path under its own name.

In rust-lang/rust#155137, we're relaxing these restrictions: `self` may now also appear as the last segment of a path (preceded by `::`) as long as the preceding path resolves to a module, enumeration, or trait.  In a `use` path, `use P::self [as name]` is equivalent to `use P::{self [as name]}`.  In non-`use` paths, forms such as `type Ty = P::self` and `pub(in P::self)` are now accepted.

Let's update the Reference to reflect this.

cc @ehuss @mu001999

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -203,6 +203,21 @@ mod example {
 > [!NOTE]
 > `self` may also be used as the first segment of a path. The use of `self` as the first segment and inside a `use` brace is logically the same; it means the current module of the parent segment, or the current module if there is no parent segment. See [`self`] in the paths chapter for more information on the meaning of a leading `self`.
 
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2237/18d42468b25d69a223aaee237616265aa6ec5be7..cf92e58fef3d6368c63de56d9f503e9a05cd4829)*

PR #1536, opened 2024-07-18
Changes for unsafe extern blocks (RFC 3484)

spastorino:
This goes after this is merged:

- https://github.com/rust-lang/rust/pull/127921

...and CI should be happy after that.

I'd say that also this is kind of WIP until someone wouldn't mind fixing my broken english :).

r? @traviscross 

Tracking:

- https://github.com/rust-lang/rust/issues/123743

PR #927, opened 2021-01-11
Update grammar for parser unification.

ehuss:
Updates the grammar for function and associated item unification/simplification. Over a series of PRs, various parts of the grammar were unified so that some the same grammar is reused in various places, with restrictions and validation applied after parsing.

Some of the PRs:
* https://github.com/rust-lang/rust/pull/67131: Merge `TraitItem` & `ImplItem` into `AssocItem`
* https://github.com/rust-lang/rust/pull/68728: merge `fn` syntax + cleanup item parsing
* https://github.com/rust-lang/rust/pull/68764: syntactically allow `self` in all `fn` contexts
* https://github.com/rust-lang/rust/pull/68788: towards unified `fn` grammar
* https://github.com/rust-lang/rust/pull/69023: unify function front matter parsing
* https://github.com/rust-lang/rust/pull/69194: fuse associated and extern items up to defaultness

This also includes a few other fixes:
* Function front matter fixes (like `async const`).
* Fix function pointer front matter (which does not syntactically allow `async` or `const`).
* Add some notes about various restrictions.

I can provide links to rustc parser functions if you want, but I think they are pretty easy to find.

Closes #750
Closes #751

PR #2240, opened 2026-04-24
Fix the grammar of function pointer types

fmease:
The grammar of function pointer types on master drastically diverges from reality.
Please read the individual commit messages for details (changes, motivation).

For example, according to the grammar on master, the following snippets are syntactically illegal but that is in conflict with rustc which deems them syntactically well-formed:

1. `safe fn()`
2. `fn(...)`, `fn(...,)`, `fn(_: ...)`, `fn(v: ...)`, `fn(..., ())`
3. `fn(&self)`, `fn(&'static mut self, ())`
4. `fn(mut x: ())`, `fn(false: bool)`, `fn(&_: ())`, `fn(&&true: ())`

Context: [#t-lang > Naming arguments in Fn traits](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/Naming.20arguments.20in.20Fn.20traits/with/590761828) and https://github.com/rust-lang/rfcs/pull/3955 where this was quite relevant.

### DIFF
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -9,9 +9,9 @@ Function ->
         FunctionReturnType? WhereClause?
         ( BlockExpression | `;` )
 
-FunctionQualifiers -> `const`? `async`?[^async-edition] ItemSafety?[^extern-qualifiers] (`extern` Abi?)?
+FunctionQualifiers -> `const`? `async`?[^async-edition] Safety?[^extern-qualifiers] (`extern` Abi?)?
 
-ItemSafety -> `safe`[^extern-safe] | `unsafe`
+Safety -> `safe`[^safe-semantics] | `unsafe`
 
 Abi -> STRING_LITERAL | RAW_STRING_LITERAL
 
@@ -34,11 +34,11 @@ FunctionReturnType -> `->` Type
 
 [^async-edition]: The `async` qualifier is not allowed in the 2015 edition.
 
-[^extern-safe]: The `safe` function qualifier is only allowed semantically within `extern` blocks.
-
-[^extern-qualifiers]: *Relevant to editions earlier than Rust 2024*: Within `extern` blocks, the `safe` or `unsafe` function qualifier is only allowed when the `extern` is qualified as `unsafe`.
+[^extern-qualifiers]: The `safe` or `unsafe` qualifiers are only allowed semantically within `extern` blocks. *Relevant to editions earlier than Rust 2024*: More specifically within `extern` blocks that are qualified as `unsafe`.
 
 [^fn-param-2015]: Function parameters with only a type are only allowed in an associated function of a [trait item] in the 2015 edition.
+
+[^safe-semantics]: The `safe` qualifier is only allowed semantically on functions within `extern` blocks.
 
 r[items.fn.intro]
 A _function_ consists of a [block] (that's the _body_ of the function), along with a name, a set of parameters, and an output type. Other than a name, all these are optional.
--- a/src/types/function-pointer.md
+++ b/src/types/function-pointer.md
@@ -7,7 +7,7 @@ BareFunctionType ->
     ForLifetimes? FunctionTypeQualifiers `fn`
        `(` FunctionParametersMaybeNamedVariadic? `)` BareFunctionReturnType?
 
-FunctionTypeQualifiers -> `unsafe`? (`extern` Abi?)?
+FunctionTypeQualifiers -> Safety? (`extern` Abi?)?
 
 BareFunctionReturnType -> `->` TypeNoBounds
 
--- a/src/types/function-pointer.md
+++ b/src/types/function-pointer.md
@@ -5,23 +5,17 @@ r[type.fn-pointer.syntax]
 ```grammar,types
 BareFunctionType ->
     ForLifetimes? FunctionTypeQualifiers `fn`
-       `(` FunctionParametersMaybeNamedVariadic? `)` BareFunctionReturnType?
+       `(` MaybeNamedFunctionParameters? `)` BareFunctionReturnType?
 
 FunctionTypeQualifiers -> Safety? (`extern` Abi?)?
 
 BareFunctionReturnType -> `->` TypeNoBounds
 
-FunctionParametersMaybeNamedVariadic ->
-    MaybeNamedFunctionParameters | MaybeNamedFunctionParametersVariadic
-
 MaybeNamedFunctionParameters ->
     MaybeNamedParam ( `,` MaybeNamedParam )* `,`?
 
 MaybeNamedParam ->
-    OuterAttribute* ( ( IDENTIFIER | `_` ) `:` )? Type
-
-MaybeNamedFunctionParametersVariadic ->
-    ( MaybeNamedParam `,` )* MaybeNamedParam `,` OuterAttribute* `...`
+    OuterAttribute* ( ( IDENTIFIER | `_` ) `:` )? ( Type | `...` )
 ```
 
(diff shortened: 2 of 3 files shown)


src/types/function-pointer.md
```diff
@@ -5,23 +5,21 @@ r[type.fn-pointer.syntax]
...
+    | (SelfParam `,`)? MaybeNamedParam ( `,` MaybeNamedParam )* `,`?
 
 MaybeNamedParam ->
-    OuterAttribute* ( ( IDENTIFIER | `_` ) `:` )? Type
+    OuterAttribute* ( MaybeNamedPattern `:` )? ( Type | `...` )
 
-MaybeNamedFunctionParametersVariadic ->
-    ( MaybeNamedParam `,` )* MaybeNamedParam `,` OuterAttribute* `...`
+MaybeNamedPattern ->
+    `mut`? IDENTIFIER | ( `&` | `&&` )? ( `_` | `false` | `true` )
```
fmease: Yes, indeed, that's the grammar! It's a tiny subset of patterns.

Note that this subgrammar is also used for associated functions in traits in Rust 2015. I didn't know how I am meant to branch on editions, so I didn't update the corresponding function item grammar that currently only has `FunctionParam -> … Type[^fn-param-2015] …` which is not sufficient strictly speaking.

I guess I could do sth. akin to "`ModernGrammar[^1] | LegacyGrammar[^2]` // `[^1]: Rust 2018–2024` // `[^2]: Rust 2015 only`" (pseudo). Let me know what you think.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2240/8c88f9d0bdd75ffdc0691676d83212ae22a18cee..400cb4b933a7c748f0f776441a6c55bcd23f2d8f)*

fmease:
r? ehuss or reassign

src/items/functions.md
```diff
@@ -34,12 +34,12 @@ FunctionReturnType -> `->` Type
...
 [^async-edition]: The `async` qualifier is not allowed in the 2015 edition.
 
-[^extern-safe]: The `safe` function qualifier is only allowed semantically within `extern` blocks.
-
-[^extern-qualifiers]: *Relevant to editions earlier than Rust 2024*: Within `extern` blocks, the `safe` or `unsafe` function qualifier is only allowed when the `extern` is qualified as `unsafe`.
+[^extern-qualifiers]: The `safe` or `unsafe` qualifiers are only allowed semantically within `extern` blocks. *Relevant to editions earlier than Rust 2024*: More specifically within `extern` blocks that are qualified as `unsafe`.
 
 [^fn-param-2015]: Function parameters with only a type are only allowed in an associated function of a [trait item] in the 2015 edition.
 
+[^safe-semantics]: The `safe` qualifier is only allowed semantically on functions and function pointer types within `extern` blocks.
```
fmease:
```suggestion
[^safe-semantics]: The `safe` qualifier is only allowed semantically on functions within `extern` blocks.
```

oops

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2240/8c88f9d0bdd75ffdc0691676d83212ae22a18cee..400cb4b933a7c748f0f776441a6c55bcd23f2d8f)*

src/types/function-pointer.md
fmease: TODO: Note somewhere (maybe in a footnote that semantically only identifiers and underscores are valid)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2240/8c88f9d0bdd75ffdc0691676d83212ae22a18cee..400cb4b933a7c748f0f776441a6c55bcd23f2d8f)*

ehuss: This doesn't look quite right. My understanding is that this logic comes from [`is_named_param`](https://github.com/rust-lang/rust/blob/9838411cb723b60dc62b1625751075c4d933b992/compiler/rustc_parse/src/parser/item.rs#L3591-L3606).

I think it would look something closer to:

```suggestion
    ((`mut`? | `&` | `&&`)? IDENTIFIER) | ((`&` | `&&`)? (`_` | `false` | `true`))
```

because ident can be preceded by & or &&.

However, this seems to be getting close to the territory of unintentional behavior. I'm wondering if this really is a bug in the compiler, and the grammar should be something else.

ehuss (commented):
Thanks!

This is just an initial review. I think we'll want to add rules that express each of the semantic restrictions. Whether it's conditional (like [`items.fn.safety-qualifiers`](https://doc.rust-lang.org/nightly/reference/items/functions.html#r-items.fn.safety-qualifiers)) or unconditional (like [`items.traits.associated-visibility`](https://doc.rust-lang.org/nightly/reference/items/traits.html#r-items.traits.associated-visibility)).


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2240/8c88f9d0bdd75ffdc0691676d83212ae22a18cee..400cb4b933a7c748f0f776441a6c55bcd23f2d8f)*

ehuss:
I did some research to see how these got missed. Some of them just weren't communicated during implementation/stabilization. Some were just missed when reverse-engineering the compiler. Some were missed when trying to handle too many things at once.

`safe` syntactically allowed on a function pointer -- Missed in https://github.com/rust-lang/reference/pull/1536 (not communicated, missed in verification)

`...` allowed as a standalone argument -- Missed in https://github.com/rust-lang/reference/pull/1494 (not communicated, missed in verification)

`...` as last argument switch to a semantic restriction -- Missed in https://github.com/rust-lang/reference/pull/927 (oversight, overwhelmed in large number of changes)

trailing `...` can be followed by a comma -- Missed in https://github.com/rust-lang/reference/pull/927 (oversight, overwhelmed in large number of changes)

`self` allowed in function pointer -- Missed in https://github.com/rust-lang/reference/pull/927 (from https://github.com/rust-lang/rust/pull/68764) (oversight, overwhelmed in large number of changes)

`...` allows named argument -- Added in https://github.com/rust-lang/rust/pull/57760 which should have been unstable. No notification of this change.

function pointer parameters have limited patterns -- Missed in initial introduction in https://github.com/rust-lang/reference/pull/433. I don't remember what happened there. The current behavior does not look intentional to me, though.

src/types/function-pointer.md
fmease: Good catch, I messed that up when copying over & adjusting [#t-lang > Naming arguments in Fn traits @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/Naming.20arguments.20in.20Fn.20traits/near/590690410) (which had its own problems).

Yeah, so the gnarly grammar stems from ``INTERSECT( PatternNoTopAlt , ( `mut` | `&` | `&&` ) IDENTIFIER_OR_KEYWORD )``.

Right, the `&` and `&&` are ancient remnants of argument modes (~2012/2013) and should be banished. The rest should be removed, too. I'm going to spin up a crater run soon and add more context with regards to "argument modes" etc. Slightly more context: https://github.com/rust-lang/rfcs/pull/3955#discussion_r3147491234.

fmease: ``INTERSECT( PatternNoTopAlt , ( `mut` | `&` | `&&` ) IDENTIFIER_OR_KEYWORD )`` is actually `` `mut`? IDENTIFIER | ( `&` | `&&` )? ( `_` | `false` | `self` | `true` | IDENTIFIER ) `` -- I was missing `self`, `fn(&&self: ())` is a valid type.

Need to update the PR.

PR #2243, opened 2026-04-28
type-layout: rewrite `#[repr(C)]` struct layout algorithm

DanielEScherzer:
Make the code actual compile by defining types, not using a variable named `struct`, adding `mut` where needed, etc.

Properly handle the `max()` alignment call returning `Option<usize>` rather than `usize` by using 1 for the alignment if there are no fields.

Add comments with the parts of the algorithm that are represented by the statements.

Also include a runnable example hidden by default.


DanielEScherzer:
This should wait until after the new guarantee is added in #2264

traviscross (approved):

traviscross:
Thanks @DanielEScherzer.

PR #2248, opened 2026-05-02
Make definition of fragment specifier `path` more precise

fmease:
The wording "a `TypePath` style path" is confusing and strictly speaking meaningless since the document never defines "`Rule`-style" where `Rule` is a grammar rule. Obviously, the original author meant "a type-style path (as defined by `TypePath`)". I've replaced it with "a `TypePath`" which is consistent with the other definitions above & below.

To prove that `path` truly is `TypePath` and not something more than that, [see the definition in rustc](https://github.com/rust-lang/rust/blob/8275de8cdd126d39676519a45a90d63f5ed747f2/compiler/rustc_parse/src/parser/nonterminal.rs#L185).

Fixes rust-lang/reference#2219.

### DIFF
--- a/src/macros-by-example.md
+++ b/src/macros-by-example.md
@@ -116,7 +116,7 @@ Valid fragment specifiers are:
   * `meta`: an [Attr], the contents of an attribute
   * `pat`: a [Pattern] (see [macro.decl.meta.edition2021])
   * `pat_param`: a [PatternNoTopAlt]
-  * `path`: a [TypePath] style path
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2248/581920f9109f141b88b860b3e1e8359e3896a150..694bb9551f411de275dc86749516c67f701e5afa)*

PR #2252, opened 2026-05-02
gitignore linkcheck

tlyu:
Ignore linkcheck.sh, which is downloaded by the linkcheck test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /book
 /dev-guide/book
 /target
+/linkcheck.sh


DanielEScherzer (approved):
Makes sense to me, that file is not currently in the repo so should not be added unintentionally

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2252/581920f9109f141b88b860b3e1e8359e3896a150..2f11bf246942b23ec8ba4e6de39398bae0c3b320)*

ehuss (approved):
Thanks! I'm going to go ahead and approve for now. But I feel like there could be a better longer-term solution that doesn't leave the file lying around. Perhaps the xtask should just automatically delete it? 

Unfortunately deleting it would interrupt a workflow of doing iterative linkcheck testing, which can be much faster. Maybe that could somehow be natively handled in the xtask, and store the script somewhere out of the way? I just don't know where that would be.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2252/581920f9109f141b88b860b3e1e8359e3896a150..2f11bf246942b23ec8ba4e6de39398bae0c3b320)*

PR #2254, opened 2026-05-03
describe equality constraints

### LINKED ISSUE #931: Document associated type equality constraint
AFAIK, associated type equality constraints aren't documented at all (other than _GenericArgsBinding_ in the grammar).

I'm not sure where exactly that would go or where to start. Seems related to [trait bounds](https://github.com/rust-lang/reference/blob/master/src/trait-bounds.md) and the new [section on constrained impls](https://github.com/rust-lang/reference/blob/master/src/items/implementations.md#generic-implementations).

tlyu:
Equality constraints on associated types have been supported for a long time, but lacking in documentation.

Much of this is based on RFC 195.

fixes rust-lang/reference#931

### DIFF
--- a/src/trait-bounds.md
+++ b/src/trait-bounds.md
@@ -100,6 +100,63 @@ r[bound.sized]
 
 `?` is only used to relax the implicit [`Sized`] trait bound for [type parameters] or [associated types]. `?Sized` may not be used as a bound for other types.
 
+r[bound.equality]
+## Equality constraints
+A trait bound that includes a [GenericArgsBinding] expresses an equality constraint on an associated type of a trait.
+
+```rust
+trait Tr { type Output; }
+struct S {}
+impl Tr for S { type Output = S; }
+fn foo<T: Tr<Output=S>>() {}
+```
+
+Equality constraints cannot currently be expressed using a [WhereClause].
+
+```rust,compile_fail
+# trait Tr { type Output; }
+# struct S {}
+# impl Tr for S { type Output = S; }
+// error: equality constraints are not yet supported in `where` clauses
+fn uses_where<T>() where T::Output = S {}
+```
+
+Equality constraints must come after any input types (type parameters of a trait) in [GenericArgs].
+
+```rust,compile_fail
+trait Tr<T> { type Output; fn foo(x: T); }
+struct S {}
+impl<T> Tr<T> for S { type Output = S; fn foo(x: T) {} }
+// error: generic arguments must come before the first constraint
+fn wrong_order<T: Tr<Output=S, ()>>() {}
+```
+
+Equality constraints cannot be used in a [QualifiedPathType].
+
+```rust,compile_fail,E0229
+# trait Tr<T> { type Output; fn foo(x: T); }
+# struct S {}
+# impl<T> Tr<T> for S { type Output = S; fn foo(x: T) {} }
+// error[E0229]: associated item constraints are not allowed here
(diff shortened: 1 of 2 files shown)


src/trait-bounds.md
```diff
@@ -100,6 +100,63 @@ r[bound.sized]
...
+```rust,compile_fail,E0229
+# trait Tr<T> { type Output; fn foo(x: T); }
+# struct S {}
+# impl<T> Tr<T> for S { type Output = S; fn foo(x: T) {} }
+// error[E0229]: associated item constraints are not allowed here
+fn uses_qualified_path() { <S as Tr<(), Output=S>>::foo(()); }
+```
+
+r[bound.equality.trait-object]
+Equality constraints are *required* for generic [trait objects].
```
fmease, line 145: That's too imprecise, they aren't required per se.

They are required for associated types specifically or rephrased, every associated type from the trait itself or from any of its supertraits needs to be constrained in the trait object type ... unless it has a `where Self: Sized;` clause or was already constrained via supertrait bounds.

In any case, isn't this partially or fully duplicating facts from the *Dyn compatibility* section and thus redundant?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2254/581920f9109f141b88b860b3e1e8359e3896a150..e88aba2560140fe620e1fbd946d40d584d942f7f)*

src/trait-bounds.md
```diff
@@ -100,6 +100,63 @@ r[bound.sized]
...
+```
+
+Equality constraints cannot currently be expressed using a [WhereClause].
+
+```rust,compile_fail
+# trait Tr { type Output; }
+# struct S {}
+# impl Tr for S { type Output = S; }
+// error: equality constraints are not yet supported in `where` clauses
+fn uses_where<T>() where T::Output = S {}
```
fmease, line 121: I'm not sure whether it makes sense to document something that isn't supported, especially since the Reference grammar already implies that.

Technically speaking, equality predicates like this are still part of the language in a gray area since they're syntactically valid in rustc. However, that'll change soon since T-lang has accepted my proposal to remove them. See https://github.com/rust-lang/rust/pull/153513, I just need to tie up some loose ends.

That's why I initially clicked on this PR, I thought you were adding them to the Reference grammar.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2254/581920f9109f141b88b860b3e1e8359e3896a150..e88aba2560140fe620e1fbd946d40d584d942f7f)*

src/trait-bounds.md
tlyu: Equality constraints appear to currently only be valid for associated types, as stated earlier in the proposed text. Stating that they are required for trait objects implies that they apply only to associated types.

> In any case, isn't this partially or fully duplicating facts from the Dyn compatibility section and thus redundant?

Where is the requirement to fully specify associated types on trait objects in the existing Reference? It doesn't logically belong in the section on `dyn` compatibility, because a generic trait can be `dyn` compatible.

There is a case for describing this requirement in the section on trait objects, but I think it makes more sense to put it with the other trait bounds.

src/trait-bounds.md
tlyu: The lack of support for equality constraints in `where` clauses is specifically called out in RFC 195 (even though it is internally inconsistent on about it). I think it's useful to mention it anyway, because `where` clauses otherwise generally have superset of the capabilities of bounds attached inline to generic parameters.

PR #2255, opened 2026-05-04
add xtask doctest

### LINKED ISSUE #2251: dev-guide instructions for running tests are incomplete
It's not completely clear whether this is a Reference bug, or a bug in mdBook.

The provided build setup instructions in the dev-guide are incomplete. Context: I currently have my rustup default toolchain set to stable. Installing the nightly toolchain and setting a rustup override in the reference directory (as instructed) seems to be sufficient for building, but running tests doesn't always work. Setting `rustup default nightly` seems to be necessary to run the tests directly from `mdbook test`.

Running `cargo xtask test-all` succeeds for some reason, even without `rustup default nightly`. Directly running `mdbook test` from the reference directory succeeds if `rustup default nightly` has been set, or if `RUSTUP_TOOLCHAIN=nightly` is set in the environment.

As far as I can tell from its source code, `mdbook test` uses a temporary directory (based on the OS default), which in general won't be subject to the override set in the reference directory. I don't know what additional setup is being propagated either from `cargo run` or `xtask` that is causing `cargo xtask test-all` to succeed.

macOS 15.7.5 (24G624) in case that matters.

Sample failure output

```
test expressions/operator-expr.md - Operator_expressions::Type_cast_expressions::Semantics::Pointer_to_pointer_cast (line 2074) - compile fail ... ok

failures:

---- expressions/operator-expr.md - Operator_expressions::The_try_propagation_expression (line 785) stdout ----
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> expressions/operator-expr.md:785:1
    |
785 | #![ feature(try_trait_v2) ]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0554`.
Couldn't compile the test.
```

tlyu:
Add a new `xtask`` command `doctest` that runs `mdbook test`. This ensures that `rustdoc`, when run from `mdbook`, will use the toolchain override.

fixes rust-lang/reference#2251

### DIFF
--- a/tools/xtask/src/main.rs
+++ b/tools/xtask/src/main.rs
@@ -8,7 +8,7 @@ type Result<T> = std::result::Result<T, Box<dyn Error>>;
 fn main() -> Result<()> {
     let mut args = std::env::args().skip(1);
     let cmd = args.next();
-    const OPTIONS: &str = "linkcheck, style-check, test-all";
+    const OPTIONS: &str = "mdbook-test, linkcheck, style-check, test-all";
     match cmd.as_deref() {
         Some("test-all") => {
             mdbook_test()?;
@@ -18,7 +18,7 @@ fn main() -> Result<()> {
             cargo_test()?;
             eprintln!("all tests passed!");
         }
-        Some("doctest") => mdbook_test()?,
+        Some("mdbook-test") => mdbook_test()?,
         Some("linkcheck") => linkcheck(args)?,
         Some("style-check") => style_check()?,
         Some("-h" | "--help") => eprintln!("valid options: {OPTIONS}"),
--- a/tools/xtask/src/main.rs
+++ b/tools/xtask/src/main.rs
@@ -18,6 +18,7 @@ fn main() -> Result<()> {
             cargo_test()?;
             eprintln!("all tests passed!");
         }
+        Some("doctest") => mdbook_test()?,
         Some("linkcheck") => linkcheck(args)?,
         Some("style-check") => style_check()?,
         Some("-h" | "--help") => eprintln!("valid options: {OPTIONS}"),
--- a/dev-guide/src/examples.md
+++ b/dev-guide/src/examples.md
@@ -29,9 +29,9 @@ When demonstrating success cases, multiple cases may be included in a single cod
 
 ## Testing examples
 
-The Rust code blocks are tested in CI. You can verify that the samples pass by running [`mdbook test`].
+The Rust code blocks are tested in CI. You can verify that the samples pass by running [`cargo xtask mdbook-test`].
 
-[`mdbook test`]: tests.md#inline-tests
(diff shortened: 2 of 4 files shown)


tools/xtask/src/main.rs
```diff
@@ -18,6 +18,7 @@ fn main() -> Result<()> {
             cargo_test()?;
             eprintln!("all tests passed!");
         }
+        Some("doctest") => mdbook_test()?,
```
ehuss: Just bikeshedding the name here, maybe it could be something like "mdbook-test"? The reason I say that is that I would prefer to not hide the abstraction here, and would like to see this option be able to forward arguments to `mdbook test`. For example, the `--chapter` argument is very useful.

(Supporting other arguments isn't needed here, I can follow up in a separate PR.)

tools/xtask/src/main.rs
ehuss: The OPTIONS list here will also need to be updated.

ehuss (commented):
If we add this, then I think we should also update the documentation to guide users towards the preferred command/workflow. So for example:

* https://github.com/rust-lang/reference/blob/e6b301082d8178b0378f9589d7ea0037cb918817/dev-guide/src/examples.md#L32
* https://github.com/rust-lang/reference/blob/e6b301082d8178b0378f9589d7ea0037cb918817/dev-guide/src/tests.md?plain=1#L6
* https://github.com/rust-lang/reference/blob/e6b301082d8178b0378f9589d7ea0037cb918817/dev-guide/src/tests.md?plain=1#L22-L28

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2255/581920f9109f141b88b860b3e1e8359e3896a150..c2ac77453ad12a50d09c61abf14579eb3d4048d4)*

tlyu:
Thanks for the review! Oops, I had some edits in progress before I saw your comments. Will incorporate them soon.

tools/xtask/src/main.rs
tlyu: I would prefer the shorter name, but I agree that having the name indicate the underlying command will help if we want to allow argument passing to `mdbook test` in the future.

tlyu:
I think the latest revision addresses your comments. Please let me know what you think!

tlyu:
I think the commit messages and PR title should be adjusted to reflect the changed command name. Please let me know if you're OK with me squashing and rewording accordingly.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2255/581920f9109f141b88b860b3e1e8359e3896a150..ffe4cfdbc88f1b3bf45ddb118d66b673b3434a44)*

PR #2257, opened 2026-05-05
Fix grammar rules containing or pertaining to bounds

fmease:
At least since PR https://github.com/rust-lang/rust/pull/39158 (2017) bounds are intentionally always optional after "bound heralds" (`:`, `impl`, `dyn`). However, the Reference didn't reflect this fact everywhere. This PR rectifies this.

For example reference@master considers the following snippets to be syntactically ill-formed which directly contradicts rustc:

* items `type T: ;`, `type T: where;`, `type T: = U;`
* type paths `T<U: >`, `T<U<V>: >` (as briefly mentioned in PR rust-lang/reference#2247)
* types `impl`, `fn() -> impl`
* types `dyn`, `fn() -> dyn`
* `struct T<'a:> where 'a:;`

Furthermore, the edition disclaimer for trait object types is incomplete / imprecise / inaccurate:

It states that token sequence `dyn::` will be interpreted as the start of a path in Rust 2015 which is correct in isolation but far from complete in context. Per argumentum e contrario, it would wrongly imply that in Rust 2015 the following snippets all contain (bare) trait object types:

* `type T = dyn;`, `type T = (dyn);`, `type T = [dyn];`
* `type T = dyn<>;`, `type T = dyn<()>;`, `type T = dyn<<T>::S>;`

To address this, I've changed the note to use an exhaustive and positive listing of tokens. The follow set is {*PathIdentSegment*, *LIFETIME_OR_LABEL*, `for`, `(`, `?`} as per rustc's [`can_begin_dyn_bound_in_edition_2015`](https://github.com/rust-lang/rust/blob/83adf7e080d0fe4f0970c00eac2976a767dbd042/compiler/rustc_parse/src/parser/ty.rs#L88-L103).

### DIFF
--- a/src/types/impl-trait.md
+++ b/src/types/impl-trait.md
@@ -23,6 +23,10 @@ fn foo(arg: impl Trait) {
 fn bar() -> impl Trait {
 }
 ```
+
+r[type.impl-trait.bounds]
+There must be at least one trait bound, no more than one `use<..>` bound, and no more than one opt-out bound (e.g., `?Sized`).
+
 r[type.impl-trait.param]
 ## Anonymous type parameters
 
--- a/src/types/impl-trait.md
+++ b/src/types/impl-trait.md
@@ -3,9 +3,9 @@ r[type.impl-trait]
 
 r[type.impl-trait.syntax]
 ```grammar,types
-ImplTraitType -> `impl` Bounds
+ImplTraitType -> `impl` Bounds?
 
-ImplTraitTypeOneBound -> `impl` TraitBound
+ImplTraitTypeOneBound -> `impl` TraitBound?
 ```
 
 r[type.impl-trait.intro]
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -3,10 +3,13 @@ r[type.trait-object]
 
 r[type.trait-object.syntax]
 ```grammar,types
-TraitObjectType -> Bounds | `dyn` Bounds?
+TraitObjectType -> Bounds[^bare-2021] | `dyn`[^dyn-2018] Bounds?
 
-TraitObjectTypeOneBound -> TraitBound | `dyn` TraitBound?
+TraitObjectTypeOneBound -> TraitBound[^bare-2021] | `dyn`[^dyn-2018] TraitBound?
 ```
+
+[^bare-2021]: See [type.trait-object.syntax-edition2021].
+[^dyn-2018]: See [type.trait-object.syntax-edition2018].
 
 r[type.trait-object.intro]
 A *trait object* is an opaque value of another type that implements a set of traits. The set of traits is made up of a [dyn compatible] *base trait* plus any number of [auto traits].
@@ -30,7 +33,7 @@ For example, given a trait `Trait`, the following are all trait objects:
 
 r[type.trait-object.syntax-edition2021]
 > [!EDITION-2021]
-> Before the 2021 edition, the `dyn` keyword may be omitted.
+> Before the 2021 edition, the `dyn` keyword may be omitted. In the 2021 edition and beyond, the `dyn` keyword is required semantically.
 
 r[type.trait-object.syntax-edition2018]
 > [!EDITION-2018]
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -37,7 +37,9 @@ r[type.trait-object.syntax-edition2021]
 
 r[type.trait-object.syntax-edition2018]
 > [!EDITION-2018]
-> In the 2015 edition, if the first bound of the trait object is a path that starts with `::`, then the `dyn` will be treated as a part of the path. The first path can be put in parenthesis to get around this. As such, if you want a trait object with the trait `::your_module::Trait`, you should write it as `dyn (::your_module::Trait)`.
+> In the 2015 edition, `dyn` must be followed by [PathIdentSegment], [LIFETIME_TOKEN], `for`, `(` or `?` to be interpreted as a keyword instead of a regular identifier.
+>
+> Most notably, `dyn`, `dyn::T` and `dyn<T>` will all be treated as type paths. As such, if you want a trait object type with the trait `::module::Trait`, you need to put the path in parentheses and write it as `dyn (::module::Trait)`.
 >
 > Beginning in the 2018 edition, `dyn` is a true keyword and is not allowed in paths, so the parentheses are not necessary.
 
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -3,9 +3,9 @@ r[type.trait-object]
 
 r[type.trait-object.syntax]
 ```grammar,types
-TraitObjectType -> `dyn`? Bounds
+TraitObjectType -> Bounds | `dyn` Bounds?
 
-TraitObjectTypeOneBound -> `dyn`? TraitBound
+TraitObjectTypeOneBound -> TraitBound | `dyn` TraitBound?
 ```
 
 r[type.trait-object.intro]
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -37,7 +37,9 @@ r[type.trait-object.syntax-edition2021]
 
 r[type.trait-object.syntax-edition2018]
 > [!EDITION-2018]
-> In the 2015 edition, if the first bound of the trait object is a path that starts with `::`, then the `dyn` will be treated as a part of the path. The first path can be put in parenthesis to get around this. As such, if you want a trait object with the trait `::your_module::Trait`, you should write it as `dyn (::your_module::Trait)`.
+> In the 2015 edition, `dyn` must be followed by [PathIdentSegment][grammar-PathIdentSegment], [LIFETIME_OR_LABEL][grammar-LIFETIME_OR_LABEL], `for`, `(` or `?` to be interpreted as the start of a trait object type. Otherwise, it will be interpreted as a regular identifier.
+>
+> Most notably, `dyn`, `dyn::T` and `dyn<T>` will all be treated as type paths. The first path can be put in parenthesis to get around this. As such, if you want a trait object with the trait `::module::Trait`, you should write it as `dyn (::module::Trait)`.
 >
 > Beginning in the 2018 edition, `dyn` is a true keyword and is not allowed in paths, so the parentheses are not necessary.
 
--- a/src/types/trait-object.md
+++ b/src/types/trait-object.md
@@ -14,11 +14,8 @@ A *trait object* is an opaque value of another type that implements a set of tra
 r[type.trait-object.impls]
 Trait objects implement the base trait, its auto traits, and any [supertraits] of the base trait.
 
-r[type.trait-object.name]
-Trait objects are written as the keyword `dyn` followed by a set of trait bounds, but with the following restrictions on the trait bounds.
-
-r[type.trait-object.constraint]
-There may not be more than one non-auto trait, no more than one lifetime, and opt-out bounds (e.g. `?Sized`) are not allowed. Furthermore, paths to traits may be parenthesized.
+r[type.trait-object.bounds]
+There must be at least one trait bound, there may not be more than one non-auto trait, no more than one lifetime, and opt-out bounds (e.g., `?Sized`) and `use<..>` bounds are not allowed.
 
 For example, given a trait `Trait`, the following are all trait objects:
 
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -7,7 +7,7 @@ GenericParams -> `<` ( GenericParam (`,` GenericParam)* `,`? )? `>`
 
 GenericParam -> OuterAttribute* ( LifetimeParam | TypeParam | ConstParam )
 
-LifetimeParam -> Lifetime ( `:` LifetimeBounds )?
+LifetimeParam -> Lifetime ( `:` LifetimeBounds? )?
 
 TypeParam -> IDENTIFIER ( `:` Bounds? )? ( `=` Type )?
 
@@ -234,7 +234,7 @@ WhereClauseItem ->
       LifetimeWhereClauseItem
     | TypeBoundWhereClauseItem
 
-LifetimeWhereClauseItem -> Lifetime `:` LifetimeBounds
+LifetimeWhereClauseItem -> Lifetime `:` LifetimeBounds?
 
 TypeBoundWhereClauseItem -> ForLifetimes? Type `:` Bounds?
 ```
--- a/src/items/type-aliases.md
+++ b/src/items/type-aliases.md
@@ -4,7 +4,7 @@ r[items.type]
 r[items.type.syntax]
 ```grammar,items
 TypeAlias ->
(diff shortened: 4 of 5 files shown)


src/types/trait-object.md
```diff
@@ -3,9 +3,9 @@ r[type.trait-object]
 
 r[type.trait-object.syntax]
 ```grammar,types
-TraitObjectType -> `dyn`? TypeParamBounds
+TraitObjectType -> TypeParamBounds | `dyn` TypeParamBounds?
```
fmease: Due to this grammar rule, *TypeParamBounds* (now: *Bounds*) must remain non-empty.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/4f86eeb5c25a7e5f1dd8a64f436caf8393ad1a3c..4200b2e53f4b5c606846eaddfe11239b5297d47d)*

src/types/trait-object.md
```diff
@@ -37,7 +37,9 @@ r[type.trait-object.syntax-edition2021]
 
 r[type.trait-object.syntax-edition2018]
 > [!EDITION-2018]
-> In the 2015 edition, if the first bound of the trait object is a path that starts with `::`, then the `dyn` will be treated as a part of the path. The first path can be put in parenthesis to get around this. As such, if you want a trait object with the trait `::your_module::Trait`, you should write it as `dyn (::your_module::Trait)`.
+> In the 2015 edition, `dyn` must be followed by [PathIdentSegment][grammar-PathIdentSegment], [LIFETIME_OR_LABEL][grammar-LIFETIME_OR_LABEL], `for`, `(` or `?` to be interpreted as the start of a trait object type. Otherwise, it will be interpreted as a regular identifier.
```
fmease: Originally I wrote *Otherwise, it will be interpreted as a regular identifier **and thus as the start of a type path**.* However the final clause wasn't correct strictly speaking: In Rust 2015 in a type context, `dyn` that's not followed by one of those tokens could start a *TypePath* (e.g., `dyn`, `dyn::X`), a *TraitObjectType* (e.g., `dyn+`, `dyn + X`) or a *MacroInvocation* (e.g., `dyn!()`, `dyn::m!()`), so I've retracted that statement again.

I don't feel like mentioning *TypePath*, *TraitObjectType* & *MacroInvocation* in this sentence since it can easily become outdated. The consequences should be self-evident anyway.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/4f86eeb5c25a7e5f1dd8a64f436caf8393ad1a3c..7b807ace8d2dc87e50fac6720be986e751c6b6c2)*

src/types/impl-trait.md
```diff
@@ -3,9 +3,9 @@ r[type.impl-trait]
 
 r[type.impl-trait.syntax]
 ```grammar,types
-ImplTraitType -> `impl` Bounds
+ImplTraitType -> `impl` Bounds?
```
ehuss, line 6: Can we add a rule somewhere in this chapter that specifies at least one trait must be listed in the bounds? I believe this corresponds to the `AtLeastOneTrait` error.

src/types/trait-object.md
```diff
@@ -3,9 +3,9 @@ r[type.trait-object]
 
 r[type.trait-object.syntax]
 ```grammar,types
-TraitObjectType -> `dyn`? Bounds
+TraitObjectType -> Bounds | `dyn` Bounds?
```
ehuss: Similar to my other comment, can we have a rule to say at least one trait is required?

I'm not sure if it makes sense to combine that into type.trait-object.constraint or to have a separate rule just below that.

src/types/trait-object.md
ehuss: The link destinations aren't required here, since definitions for the rule names are generated automatically.


```suggestion
> In the 2015 edition, `dyn` must be followed by [PathIdentSegment], [LIFETIME_OR_LABEL], `for`, `(` or `?` to be interpreted as the start of a trait object type. Otherwise, it will be interpreted as a regular identifier.
```

ehuss (commented):
Looks great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/ac7b991181147dbb499f462b2d837c2e964dbdc7..d43e9b5e3e92a30f6ce90570cb74098cabf436e3)*

src/types/trait-object.md
```diff
@@ -15,10 +15,10 @@ r[type.trait-object.impls]
 Trait objects implement the base trait, its auto traits, and any [supertraits] of the base trait.
 
 r[type.trait-object.name]
-Trait objects are written as the keyword `dyn` followed by a set of trait bounds, but with the following restrictions on the trait bounds.
+Trait object types are written as the keyword `dyn` followed by a set of bounds, but with the following restrictions on the bounds:
```
fmease: > followed by a set of […] bounds

(preexisting) Is *set* generally to be understood as a mathematical set in the Reference? Is it defined anywhere? Right now, it's unclear to me if this sentence makes any statements about duplicate bounds. As you know, `dyn Trait + 'a + 'a` and `dyn NonAutoTrait + NonAutoTrait` are illegal and `dyn Sync + Sync` is legal...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..5fd453fcf046173d17e9addb75093efdb32c1b76)*

src/types/impl-trait.md
```diff
@@ -23,6 +23,14 @@ fn foo(arg: impl Trait) {
 fn bar() -> impl Trait {
 }
 ```
+
```
fmease, line 26: I'm mirroring the structure of the section about trait object types here with the `name` vs. `constraint` split. I'm definitely not married to it. I actually find the rule name `name` a bit confusing and I feel like the rule sort of mixes syntax and semantics (just like its preexisting sibling for trait object types).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..5fd453fcf046173d17e9addb75093efdb32c1b76)*

src/types/impl-trait.md
```diff
@@ -23,6 +23,14 @@ fn foo(arg: impl Trait) {
...
+r[type.impl-trait.name]
+
+`impl Trait` types are written as the keyword `impl` followed by a set of bounds, but with the following restrictions on the bounds:
+
+r[type.impl-trait.constraint]
+There must be at least one trait bound, no more than one `use<..>` bound, and no more than one opt-out bound (e.g. `?Sized`).
```
fmease: > no more than one `use<...>` bound

This is already mentioned in rule `type.impl-trait.generic-capture.precise.constraint-single` but it feels logical to mentioned it here to be exhaustive.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..5fd453fcf046173d17e9addb75093efdb32c1b76)*

fmease: > no more than one opt-out bound (e.g. `?Sized`)

Copying the terminology from the trait object section. Personally speaking I call them *relaxed bounds* and that's what the compiler calls them now everywhere, too (it used to call them *relaxed bounds*, *maybe bounds* and *unbounds*).

Strictly speaking, *duplicate* relaxed bounds are forbidden but on stable that doesn't make a difference since there's only one trait that can be relaxed (`Sized`).

It felt natural to mention this here but of course, relaxed bounds (and `use<..>` bounds) come with a lot more semantic restrictions (specifically when it comes to legal locations) but the Reference doesn't have a central place to talk about them[^1].

[^1]: E.g., for the longest time `?Trait` was allowed even for `Trait` != `Sized` but two years ago or so that was finally made an error but the Reference was never updated.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..5fd453fcf046173d17e9addb75093efdb32c1b76)*

src/types/trait-object.md
```diff
@@ -15,10 +15,10 @@ r[type.trait-object.impls]
...
 
-r[type.trait-object.constraint]
-There may not be more than one non-auto trait, no more than one lifetime, and opt-out bounds (e.g. `?Sized`) are not allowed. Furthermore, paths to traits may be parenthesized.
+r[type.trait-object.bounds]
```
fmease, line 20: Changed it to `bounds` from `constraint` since the corresponding rule for impl-Trait had to be renamed to `bounds` not to conflict with the preexisting `types.impl-trait.constraint`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..1a547083ec582c9ef6682f6d5c22210184ec71ec)*

src/types/trait-object.md
```diff
@@ -37,7 +37,9 @@ r[type.trait-object.syntax-edition2021]
 
 r[type.trait-object.syntax-edition2018]
 > [!EDITION-2018]
-> In the 2015 edition, if the first bound of the trait object is a path that starts with `::`, then the `dyn` will be treated as a part of the path. The first path can be put in parenthesis to get around this. As such, if you want a trait object with the trait `::your_module::Trait`, you should write it as `dyn (::your_module::Trait)`.
+> In the 2015 edition, `dyn` must be followed by [PathIdentSegment], [LIFETIME_OR_LABEL], `for`, `(` or `?` to be interpreted as the start of a trait object type. Otherwise, it will be interpreted as a regular identifier.
```
fmease: FIXME: Imprecise language: In Rust 2015 `dyn+` is a legal trait object type where `dyn` indeed isn't a keyword but the name of the trait, so strictly speaking `dyn` is actually "interpreted as the start of a trait object type". This sentence needs to be reworded to say something about `dyn` being an "active keyword" if followed by XY // something of that nature.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..1a547083ec582c9ef6682f6d5c22210184ec71ec)*

src/types/trait-object.md
ehuss: If you want to clarify it, that seems reasonable. Lots of options, but it could just drop the "set" noun as in: "followed by the bounds".

src/types/impl-trait.md
ehuss: Yea, normally I don't like duplicating things since it makes it more difficult to maintain. But in this case it seems reasonable.

src/types/trait-object.md
```diff
@@ -3,9 +3,9 @@ r[type.trait-object]
 
 r[type.trait-object.syntax]
 ```grammar,types
-TraitObjectType -> `dyn`? Bounds
+TraitObjectType -> Bounds | `dyn` Bounds?
 
-TraitObjectTypeOneBound -> `dyn`? TraitBound
+TraitObjectTypeOneBound -> TraitBound | `dyn` TraitBound?
```
ehuss: Just wanted to double-check. The reason for adding the variants without `dyn` is because the requirement for `dyn` is done during validation and not during parsing, correct?

If so, I'm thinking of adding a footnote on these to point to `type.trait-object.syntax-edition2021` like we do in other places like [`ExternBlock`](https://github.com/rust-lang/reference/blob/6f2d1eac3b9b327536b7c481f7618f14ec83f314/src/items/external-blocks.md?plain=1#L7) where old edition syntax could be potentially confusing to readers.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/3fde038352015967da1e417115ff511c8f71baf5..1a547083ec582c9ef6682f6d5c22210184ec71ec)*

src/types/impl-trait.md
ehuss: Looking more closely at this, I agree the ".name" rules don't really make sense.

Generally we don't want to have an English description of the syntax (which is often very imprecise). I'm thinking we can just drop these two `.name` rules entirely.

src/types/trait-object.md
ehuss: (Though see my other comment where I'm thinking we can just drop this rule.)

fmease:
What changed since your last review:

1. Removed rule `type.trait-object.name` [as per your review comment](https://github.com/rust-lang/reference/pull/2257#discussion_r3392694710) & parts of `type.trait-object.bounds` [as per my own review comment](https://github.com/rust-lang/reference/pull/2257#discussion_r3424361390).
2. Clarified the section about `dyn` in Rust 2015 to address [my own concern](https://github.com/rust-lang/reference/pull/2257#discussion_r3369441283)
3. Added footnotes to the grammar of trait object types to highlight edition-sensitive parts to address [this review comment](https://github.com/rust-lang/reference/pull/2257#discussion_r3392651903)
4. Replaced *LIFETIME_OR_LABEL* with *LIFETIME_TOKEN* in the Rust 2015 `dyn` paragraph. I accidentally mixed them up and wrongly thought the "or label" one was the more general of the two. Obviously, `dyn 'static` (where we have a ticked keyword) is a syntactically valid trait object type in Rust 2015.

Ready for re-review.

src/types/trait-object.md
```diff
@@ -3,22 +3,22 @@ r[type.trait-object]
...
 r[type.trait-object.impls]
 Trait objects implement the base trait, its auto traits, and any [supertraits] of the base trait.
 
-r[type.trait-object.name]
-Trait objects are written as the keyword `dyn` followed by a set of trait bounds, but with the following restrictions on the trait bounds.
-
-r[type.trait-object.constraint]
-There may not be more than one non-auto trait, no more than one lifetime, and opt-out bounds (e.g. `?Sized`) are not allowed. Furthermore, paths to traits may be parenthesized.
+r[type.trait-object.bounds]
+There must be at least one trait bound, there may not be more than one non-auto trait, no more than one lifetime, and opt-out bounds (e.g., `?Sized`) and `use<..>` bounds are not allowed. Furthermore, paths to traits may be parenthesized.
```
fmease: Preexisting:

> Furthermore, paths to traits may be parenthesized.

I'm now also dropping this similar to the `[type.trait-object.name]`. It's merely restating parts of the grammar in an incomplete manner (not only may one parenthesize *paths* to traits, one may generally parenthesize *TraitBound*s which aren't just paths as they may be preceded by `?` or `for<...>`).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/f6f674f0319847057fce30cfe53c167a74690c91..eb086319c64a2230922e5790ded865a277723b97)*

ehuss (approved):
Looks great, thanks!

(Noting you still had a FIXME for Rust 2015, but I'm fine moving forward.)


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2257/f6f674f0319847057fce30cfe53c167a74690c91..362c0ee6661ee69002cf8a9995eb389cf73ae723)*

PR #2258, opened 2026-05-05
Rename grammar rule `TypeParamBounds` to just `Bounds`

fmease:
I take issue with the name of grammar rule *TypeParamBounds*. Bounds aren't inherently tied to type parameters, they can perfectly exist without them. This name stems from an ancient time when where-clauses didn't exist yet and bounds could only be put on type parameters. Today, that is no longer true.

Consider the following program which contains bounds that aren't associated with type parameters in any way:

```rs
fn f0() where i32: Copy {}
fn f1() where [&'static str; 10]: std::fmt::Debug {}
fn f2<const N: usize>() where [String; N]: Copy {}
fn f3<'a>() where &'a (): Sized {}
```

Bounds can be placed on any type. I'd also be fine with renaming it to *TypeBounds*, just anything but *TypeParamBounds* :)

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -9,7 +9,7 @@ GenericParam -> OuterAttribute* ( LifetimeParam | TypeParam | ConstParam )
 
 LifetimeParam -> Lifetime ( `:` LifetimeBounds )?
 
-TypeParam -> IDENTIFIER ( `:` TypeParamBounds? )? ( `=` Type )?
+TypeParam -> IDENTIFIER ( `:` Bounds? )? ( `=` Type )?
 
 ConstParam ->
     `const` IDENTIFIER `:` Type
@@ -236,7 +236,7 @@ WhereClauseItem ->
 
 LifetimeWhereClauseItem -> Lifetime `:` LifetimeBounds
 
-TypeBoundWhereClauseItem -> ForLifetimes? Type `:` TypeParamBounds?
+TypeBoundWhereClauseItem -> ForLifetimes? Type `:` Bounds?
 ```
 
 r[items.generics.where.intro]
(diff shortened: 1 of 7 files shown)


ehuss:
Sounds reasonable to me.

Can you also update the sentence in `paths.type.intro` that says "type parameter bounds" and replace that with something else ("type bounds" maybe?) if that makes sense.

fmease:
I've gone ahead and removed the mention entirely:

```patch
-Type paths are used within type definitions, trait bounds, type parameter bounds, and qualified paths.
+Type paths are used within type definitions, trait bounds, and qualified paths.
```

That's because it was redundant anyway. *Bound* (formerly *TypeParamBound*) is defined as *Lifetime | TraitBound | UseBound* but *TraitBound* is already listed in that sentence and obviously *Lifetime* and *UseBound* don't contain *TypePath*s.

fmease:
~~I've also taken the liberty to rename `trait-bounds.md` to `bounds.md` since it actually describes all three kinds of bounds: Trait bounds, lifetime bounds (outlives-bounds) and use-bounds (precise capturing).~~

~~Feel free to drop that commit if you'd like to see that split off into a separate PR.~~

**Edit**: Dropped again since I forgot about URL stability and since I can't be bothered to add redirects (if you even can; I only know about the JS-driven URL fragment redirects); the Book and the rustc book both reference `trait-bounds.md`.

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2258/081e3c9cb2e4aedec421df03f7c41eb9c82eee5f..d6f4cc427ddb064a047074ab94e1a6fdefacb709)*

PR #2261, opened 2026-05-05
Avoid direct Drop::drop method link in destructor docs

P8L1:
This avoids linking directly to `core::ops::Drop::drop` from the destructor docs.

That direct method link can resolve to a rustdoc method-fragment anchor such as `#tymethod.drop`. In rust-lang/rust#144537, `Drop::drop` becomes a provided/default method, so the old required-method anchor is no longer generated, and linkchecker fails for `reference/destructors.html` and `reference/print.html`.

The docs still mention `<T as core::ops::Drop>::drop`, but now link to the `Drop` trait page instead of relying on a method-specific fragment.

This is the Reference-side fix needed for the linkchecker failure in rust-lang/rust#144537.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -7,7 +7,7 @@ When an [initialized]&#32;[variable] or [temporary] goes out of [scope](#drop-sc
 r[destructors.operation]
 The destructor of a type `T` consists of:
 
-1. If `T: Drop`, calling [`<T as core::ops::Drop>::drop`](core::ops::Drop::drop)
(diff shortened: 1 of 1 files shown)


P8L1:
r? petrochenkov

ehuss:
Discussing this at https://github.com/rust-lang/rust/pull/144537#issuecomment-4382106221.

P8L1:
Thanks, that makes sense. I agree this is cleaner than trying to fix the generated Reference link for a stage1 rustdoc/std mismatch.

I’ll prepare a small PR with the `x86_64-gnu-llvm3.sh` linkchecker skip.

PR #2262, opened 2026-05-05
Structs with no fields or all-ZST fields are ZSTs

DanielEScherzer:
Guarantee that structs with Rust representation and with no fields, or where all fields are ZSTs, are themselves ZSTs.

Guarantee that enums with Rust representation with a single struct-like variant with no fields, or where all fields are ZSTs, are themselves ZSTs.

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -223,6 +223,7 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 - `repr(Rust)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.rust.struct-zst]).
 - `repr(C)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
 - `repr(transparent)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
+- `repr(Rust)` [enums] (without a primitive representation specified) with a single struct-like variant with no fields or where all fields are zero-sized (see [layout.repr.rust.enum-struct-like-zst])
 - [Arrays] of zero-sized types (see [layout.array]).
 - [Arrays] of length zero (see [layout.array]).
 - [Unions] of zero-sized types (see [items.union.common-storage]).
@@ -261,6 +262,9 @@ union U {
     f2: [(); 10],
     f3: [u8; 0],
 }
+enum E2 {
+    V1 { f1: (), f2: [(); 10 ] },
+}
 assert_eq!(0, size_of::<()>());
 assert_eq!(0, size_of_val(&f));
 assert_eq!(0, size_of_val(&S));
@@ -275,6 +279,7 @@ assert_eq!(0, size_of::<T2>());
 assert_eq!(0, size_of::<[(); 10]>());
 assert_eq!(0, size_of::<[u8; 0]>());
 assert_eq!(0, size_of::<U>());
+assert_eq!(0, size_of::<E2>());
 ```
 
 [`extern` blocks]: items.extern
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -220,6 +220,7 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 - [Function items] (see [type.fn-item.intro]).
 - The constructors of [tuple-like structs] (see [type.fn-item.intro]).
 - The constructors of [tuple-like enum variants] (see [type.fn-item.intro]).
+- `repr(Rust)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.rust.struct-zst]).
 - `repr(C)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
 - `repr(transparent)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
 - [Arrays] of zero-sized types (see [layout.array]).
@@ -231,6 +232,13 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 fn f() {}
 struct S(u8);
 enum E { V(u8) }
+struct UnitLike;
+struct NoFields {}
+struct OnlyZST {
+    f1: (),
+    f2: [(); 10],
+    f3: [u8; 0],
+}
 #[repr(C)]
 struct C1 {}
 #[repr(C)]
@@ -257,6 +265,9 @@ assert_eq!(0, size_of::<()>());
 assert_eq!(0, size_of_val(&f));
 assert_eq!(0, size_of_val(&S));
 assert_eq!(0, size_of_val(&E::V));
+assert_eq!(0, size_of::<UnitLike>());
+assert_eq!(0, size_of::<NoFields>());
+assert_eq!(0, size_of::<OnlyZST>());
 assert_eq!(0, size_of::<C1>());
 assert_eq!(0, size_of::<C2>());
 assert_eq!(0, size_of::<T1>());
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -220,8 +220,10 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 - [Function items] (see [type.fn-item.intro]).
 - The constructors of [tuple-like structs] (see [type.fn-item.intro]).
 - The constructors of [tuple-like enum variants] (see [type.fn-item.intro]).
+- `repr(Rust)` [structs] with no fields or where all fields are zero sized (see [layout.repr.rust.struct-zst]).
 - `repr(C)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
 - `repr(transparent)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
+- `repr(Rust)` [enums] (without a [primitive representation] specified) with a single [field-struct-like variant], a single [unit-struct-like variant], or a single [tuple-struct-like variant] and where the struct-like thing has no fields or where all of the fields are zero sized (see [layout.repr.rust.enum-struct-like-zst]).
 - [Arrays] of zero-sized types (see [layout.array]).
 - [Arrays] of length zero (see [layout.array]).
 - [Unions] of zero-sized types (see [items.union.common-storage]).
@@ -231,6 +233,13 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 fn f() {}
 struct S(u8);
 enum E { V(u8) }
+struct UnitLike;
+struct NoFields {}
+struct OnlyZST {
+    f1: (),
+    f2: [(); 10],
+    f3: [u8; 0],
+}
 #[repr(C)]
 struct C1 {}
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


traviscross:
cc @RalfJung @rust-lang/opsem

src/type-layout.md
```diff
@@ -180,6 +180,12 @@ For [structs], it is further guaranteed that the fields do not overlap. That is,
 
 Be aware that this guarantee does not imply that the fields have distinct addresses: [zero-sized types] may have the same address as other fields in the same struct.
 
+r[layout.repr.rust.struct-zst]
+For [structs] with no fields, or where all fields are [zero-sized], it is further guaranteed that the structs are themselves [zero-sized].
+
+r[layout.repr.rust.enum-struct-like-zst]
+For [enums] (without a primitive representation specified) with a single struct-like variant with no fields or where all fields are [zero-sized], the enum itself is [zero-sized].
```
theemathas: Seems bizarre to guarantee this for struct-like variants but not unit-like variants or tuple-like variants.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/4fd61400e123abc713b8b0b3b1f52bc2bfb56888..5011dcdb90f3b84599ace73a4ee0e05c3de7d286)*

RalfJung:
That seems reasonable to me (specifically for `repr(Rust)`; not necessarily for other reprs). Someone should check whether it is honored by `-Zrandomize-layout`.

traviscross:
We talked about this in the lang call today.  Let's propose to do this with the correction that, for enums, we mean those `repr(Rust)` enums without a primitive representation specified and that have a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant where the struct-like thing has no fields or where all of the fields are zero-sized.

@rfcbot fcp merge lang

traviscross:
@rfcbot concern check-randomize-layout

tmandry:
This seems stronger than what we guarantee for non-ZST structs, which require `#[repr(transparent)]` to have the same representation as the inner type. In the lang meeting, @steffahn pointed out that the non-ZST case is harder because certain types get passed in registers, etc.

I don't see a reason *not* to make this guarantee, and it seems like a useful one, so let's do it.

@rfcbot reviewed

zachs18:
Should we also guarantee that zero-variant `repr(Rust)` enums are ZSTs? They currently are in practice.

-----

> Someone should check whether it is honored by -Zrandomize-layout.

Yes, I believe it is. IIUC the current implementation of `-Zrandomize-layout` only reorderes fields, never introduces any "extra" padding[^1]. Shuffling the field order cannot make a ZST into a non-ZST (or vice versa), so the current implementation upholds the proposed guarantees.

<details> <summary>reasoning</summary>

When the `-Zrandomize-layout` flag is enabled, the compiler enables `ReprFlags::RANDOMIZE_LAYOUT` for types defined in that crate ([src](https://github.com/rust-lang/rust/blob/3daae5e42ec9ba435212987331af1b7b8634fa90/compiler/rustc_middle/src/ty/mod.rs#L1600)) (`unstable_opts.randomize_layout` is not mentioned anywhere else in the compiler).

`ReprFlags::RANDOMIZE_LAYOUT` is mentioned in four places in the compiler:


1. When it is declared ([src](https://github.com/rust-lang/rust/blob/3daae5e42ec9ba435212987331af1b7b8634fa90/compiler/rustc_abi/src/lib.rs#L92))
1. When it is enabled (see previous section)
1. In `SizeSkeleton`, which IIUC ignores it when computing the size of generic NPO-optimized enums and newtype structs/enums ([src](https://github.com/rust-lang/rust/blob/3daae5e42ec9ba435212987331af1b7b8634fa90/compiler/rustc_middle/src/ty/layout.rs#L463))
     1. NPO is irrelevant here (there are no ZSTs eligible for NPO optimization)
     2. For newtypes over ZSTs, this should correctly give a ZST size regardless of `RANDOMIZE_LAYOUT`
1. When it is used in `ReprOptions::can_randomize_type_layout` which returns `!self.inhibit_struct_field_reordering() && self.flags.contains(ReprFlags::RANDOMIZE_LAYOUT)` ([src](https://github.com/rust-lang/rust/blob/3daae5e42ec9ba435212987331af1b7b8634fa90/compiler/rustc_abi/src/lib.rs#L225))

`ReprOptions::can_randomize_type_layout` is used in one place in the compiler: in layout computation (`fn univariant_biased`) where the field list (excluding the unsized tail) is shuffled if it returns true (instead of being optimized by sorting by alignment) ([src](https://github.com/rust-lang/rust/blob/3daae5e42ec9ba435212987331af1b7b8634fa90/compiler/rustc_abi/src/layout.rs#L1121)). Shuffling (or sorting) the field order cannot make a ZST into a non-ZST (or vice versa), so the current implementation upholds the proposed guarantees.

</details>

I opened a PR https://github.com/rust-lang/rust/pull/157905 to update the comments in some of the places where `-Zrandomize-layout` is implemented referring back to this FCP to ensure that future changes to `-Zrandomize-layout` do not break these guarantees.

~~@/rustbot label I-lang-nominated~~

~~Nominating for possibly resolving the `-Zrandomize-layout` concern (and maybe considering zero-variant `repr(Rust)` enums) ~~

[^1]: other than the padding necessary to make fields aligned, and since `0` is a multiple of any alignment, this padding is never added for struct-like layouts if all fields are zero-sized

traviscross:
@rfcbot resolve check-randomize-layout

Thanks @zachs18.  I looked into this and concur.

> Should we also guarantee that zero-variant `repr(Rust)` enums are ZSTs? They currently are in practice.

Probably.  Let's do that in a separate Reference PR.

DanielEScherzer:
> We talked about this in the lang call today. Let's propose to do this with the correction that, for enums, we mean those `repr(Rust)` enums without a primitive representation specified and that have a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant where the struct-like thing has no fields or where all of the fields are zero-sized.

If/when FCP passes I'll apply this correction

chorman0773:
> without a primitive representation specified

This wording doesn't seem super clear on whether or not repr(C) enums are excluded (which they should be, because they are implicitly repr(C, Int), where Int is a target-specific integer type, but that isn't "specified" in the repr attribute).

DanielEScherzer:
> > We talked about this in the lang call today. Let's propose to do this with the correction that, for enums, we mean those `repr(Rust)` enums without a primitive representation specified and that have a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant where the struct-like thing has no fields or where all of the fields are zero-sized.
> 
> If/when FCP passes I'll apply this correction

Done, https://github.com/rust-lang/reference/pull/2262/changes/fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc

@rustbot ready

src/glossary.md
```diff
@@ -253,17 +262,46 @@ union U {
     f2: [(); 10],
     f3: [u8; 0],
 }
+# /// An enum with a single field-struct-like variant with all fields being ZSTs.
+enum E2 {
+    V1 { f1: (), f2: [(); 10 ] },
```
traviscross:
```suggestion
    V1 { f1: (), f2: [(); 10] },
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/86635e30bf861a038dc197d7e16fd09e7e514e7a..fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc)*

src/type-layout.md
```diff
@@ -180,6 +180,12 @@ For [structs], it is further guaranteed that the fields do not overlap. That is,
 
 Be aware that this guarantee does not imply that the fields have distinct addresses: [zero-sized types] may have the same address as other fields in the same struct.
 
+r[layout.repr.rust.struct-zst]
+For [structs] with no fields, or where all fields are [zero-sized], it is further guaranteed that the structs are themselves [zero-sized].
```
traviscross:
```suggestion
For [structs] with no fields, or where all fields are [zero sized], it is further guaranteed that the structs are themselves [zero sized].
```

When used as a phrasal adjective, let's hyphenate this.  In predicate position, let's not (I know the Reference isn't yet consistent about this today — we're moving in that direction).


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/86635e30bf861a038dc197d7e16fd09e7e514e7a..fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc)*

traviscross:
```suggestion
For [structs] with no fields or where all fields are [zero-sized], it is further guaranteed that the structs are themselves [zero-sized].
```

(Remove the comma there.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/86635e30bf861a038dc197d7e16fd09e7e514e7a..fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc)*

DanielEScherzer:
> > without a primitive representation specified
> 
> This wording doesn't seem super clear on whether or not repr(C) enums are excluded (which they should be, because they are implicitly repr(C, Int), where Int is a target-specific integer type, but that isn't "specified" in the repr attribute).

In the glossary, the new entries start with ``- `repr(Rust)` [structs]`` and ``- `repr(Rust)` [enums]`` s so I think it is clear there

On the type layout page, the entries are under the ``### The `Rust` representation`` section heading; the other layout rules there don't explicitly note that they are for the rust representation because that is clear from the location and context

src/type-layout.md
```diff
@@ -180,6 +180,12 @@ For [structs], it is further guaranteed that the fields do not overlap. That is,
...
+
+r[layout.repr.rust.enum-struct-like-zst]
+For [enums] (without a primitive representation specified) with a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant where the struct-like thing has no fields or where all of the fields are [zero-sized], the enum itself is [zero-sized].
```
traviscross:
```suggestion
For [enums] (without a primitive representation specified) with a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant and where the struct-like thing has no fields or where all of the fields are [zero sized], the enum itself is [zero sized].
```

(This came directly from my FCP text.  Let's add an "and" to help disambiguate the scope of the "where".


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/86635e30bf861a038dc197d7e16fd09e7e514e7a..fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc)*

traviscross:
```suggestion
For [enums] (without a primitive representation specified) with a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant where the struct-like thing has no fields or where all of the fields are [zero-sized], the enums themselves are [zero sized].
```

Let's match the plurality pattern of the struct case above.  (Again, I'm correcting my earlier verbiage here.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/86635e30bf861a038dc197d7e16fd09e7e514e7a..fa61cbdf2a146d3f7a16cbe0d10648f81e3626dc)*

src/type-layout.md
```diff
@@ -180,6 +180,12 @@ For [structs], it is further guaranteed that the fields do not overlap. That is,
 
 Be aware that this guarantee does not imply that the fields have distinct addresses: [zero-sized types] may have the same address as other fields in the same struct.
 
+r[layout.repr.rust.struct-zst]
+For [structs] with no fields or where all fields are [zero sized], it is further guaranteed that the structs are themselves [zero sized].
+
+r[layout.repr.rust.enum-struct-like-zst]
+For [enums] (without a primitive representation specified) with a single field-struct-like variant, a single unit-struct-like variant, or a single tuple-struct-like variant and where the struct-like thing has no fields or where all of the fields are [zero sized], the enums themselves are [zero sized].
```
traviscross: Let's add links to cross-reference each of these, here and in the glossary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/61c35c046e32955fae60107dd0b9c58de788aadc..034e3cf6cbc086cd9a3f5a302d627f8a91f34a77)*

traviscross: Let's add links to cross-reference each of these, here and in the glossary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/61c35c046e32955fae60107dd0b9c58de788aadc..034e3cf6cbc086cd9a3f5a302d627f8a91f34a77)*

traviscross (approved):

traviscross:
Thanks @DanielEScherzer.

src/type-layout.md
Darksonn, line 97: Now that structs guarantee the same, the exception for unit tuples is no longer needed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2262/61c35c046e32955fae60107dd0b9c58de788aadc..08c156565eb9fc1741d8bd97fd95116d9b66f4f8)*

zachs18: I think technically we don't guarantee that empty `repr(Rust)` structs are 1-aligned (yet), only that they are 0-sized.

Also, I think for clarity it'd be better to keep this section regardless, but maybe rephrase it to not say it's an exception, just a note or something.

PR #2263, opened 2026-05-05
default type parameters

### LINKED ISSUE #24: Default type parameters are not in the language reference

tlyu:
Describe default types for generic parameters, as currently implemented. There are some differences from RFC 213, but I couldn't find any real specification, so relied on observed behavior.

fixes rust-lang/reference#24

### DIFF
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -9,7 +9,9 @@ GenericParam -> OuterAttribute* ( LifetimeParam | TypeParam | ConstParam )
 
 LifetimeParam -> Lifetime ( `:` LifetimeBounds )?
 
-TypeParam -> IDENTIFIER ( `:` Bounds? )? ( `=` Type )?
+TypeParam -> IDENTIFIER ( `:` Bounds? )? DefaultType?
+
+DefaultType -> `=` Type
 
 ConstParam ->
     `const` IDENTIFIER `:` Type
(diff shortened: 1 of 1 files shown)


src/items/generics.md
```diff
@@ -257,6 +259,110 @@ where
 }
 ```
 
+r[items.generic.defaulttypes]
+## Default types
+
+r[items.generic.defaulttypes.intro]
+Generic type parameters may have default types, using the [DefaultType] syntax. Only type parameters on nominal types and traits may have default types.
```
theemathas, line 266: Type aliases may have default types.

```rust
type Foo<T = i32> = Vec<T>;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2263/cf0e3915cb84e9903ef1294fc64a50b70a1da82c..fb4977645052f379e2878c4acbad7696dd95613c)*

src/items/generics.md
```diff
@@ -257,6 +259,110 @@ where
...
+let x = <S1>::default();
+// T defaults to u8: let Type is type context
+let y: S1 = S1::default();
+// U defaults to u8: T is inferred from z2
+let z1: S2<_> = S2::default();
+let z2: S2<u8, _> = z1;
+```
+
+r[items.generic.defaulttypes.exprctx]
+In expression contexts, all omitted type arguments are treated as the inferred type `_`. [RFC 213] states that the provided default type will be used as a fallback during inference, but this seems work inconsistently in practice.
```
theemathas, line 288: Could you provide an example for when the default type is used when a generic type is specified in an expression context? The example snippets seem to give evidence that the "work inconsistently in practice" part actually never works.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2263/cf0e3915cb84e9903ef1294fc64a50b70a1da82c..fb4977645052f379e2878c4acbad7696dd95613c)*

PR #2264, opened 2026-05-05
Guarantee alignment of `repr(C)` structs with no fields

DanielEScherzer:


src/type-layout.md
```diff
@@ -200,6 +200,9 @@ r[layout.repr.c.struct]
 r[layout.repr.c.struct.align]
 The alignment of the struct is the alignment of the most-aligned field in it.
 
+r[layout.repr.c.struct.align-empty]
+The alignment of a struct with no fields is 1.
```
RalfJung: We are moving towards `repr(C)` being defined entirely by the platform ABI (https://github.com/rust-lang/rfcs/pull/3845). So I'm not sure we should add any such promises.

joshtriplett: We talked about this in today's @rust-lang/lang call. Our understanding of the migration plan is that that we'll be adding a today's-edition name for `repr(C)` (whatever that ends up being named), adding `repr(ordered_fields)` which will be the same (but indicates that the user explicitly wants that rather than being auto-migrated), and then adding a *new* `repr(C)` for the new edition.

Given that transition plan, our understanding is that all the current documentation of `repr(C)` will move to be documentation of the today's-edition `repr(C)` and `repr(ordered_fields)`, and then we'll add *new* documentation for the new `repr(C)` that largely says we do whatever the platform C ABI does.

Given that transition plan, do you see an issue with this proposal as ongoing documentation of what will end up being the old `repr(C)` rather than the new one, assuming that all this documentation moves to being for the old one?

RalfJung: For what might eventually be called `repr(ordered_fields)`, this new clause is indeed correct.

It's also redundant though, it follows from the definition of the layout algorithm.

DanielEScherzer: This rule was proposed in the context of discussing #2243 on the t-lang-docs call yesterday where it was concluded that it does *not* actually follow from the definition of the layout algorithm - the size and offsets algorithm doesn't describe how to identify the alignment of the type, and the `layout.repr.c.struct.align` rule says

> The alignment of the struct is the alignment of the most-aligned field in it.

but if there are no fields then there is no most-aligned field

Jules-Bertholet: > Our understanding of the migration plan is that that we'll be adding a today's-edition name for `repr(C)` (whatever that ends up being named), adding `repr(ordered_fields)` which will be the same

FWIW, I don't think that end state is set in stone. E.g. to resolve https://github.com/rust-lang/rust/issues/112480, we might potentially end up at a state where all 3 of old `repr(C)`, new `repr(C)`, and `repr(ordered_fields)` are different.

RalfJung: > the layout.repr.c.struct.align rule says
>
>  The alignment of the struct is the alignment of the most-aligned field in it.

Then I would suggest that it makes more sense to edit that rule than add a new one. A single rule should suffice to define the alignment of the type.

Something lie:
The alignment of the struct is the maximum of the alignments of all fields, or 1 if there are no fields.

DanielEScherzer: We can combine the rules, no objection to that, just need T-lang to sign off on the new guarantee being added

traviscross:
@rfcbot fcp merge lang

tmandry:
I'm fine with adding this as a guarantee. The suggestion to edit the existing rules makes sense, but I think we should leave that as an editorial question for the reference.

@rfcbot reviewed

nikomatsakis:
@rfcbot reviewed

I agree with making this guarantee. I don't have an opinion about whether we add it as a separate clause or as part of another clause -- @RalfJung's [suggestion here](https://github.com/rust-lang/reference/pull/2264#discussion_r3197184005) made sense to me.

src/type-layout.md
steffahn, line 244: you might also want to update this algorithm. Currently the line
```rs
struct.alignment = struct.fields().map(|field| field.alignment).max();
```
is kinda bugged anyway, since the real *max()* method returns an `Option`. So:
```rs
struct.alignment = struct.fields().map(|field| field.alignment).max().unwrap_or(1);
```
would be better ^^

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2264/4fd61400e123abc713b8b0b3b1f52bc2bfb56888..8d2c69aad62d72897ee9f1edd54f6a80cef777a5)*

DanielEScherzer: already working on that in #2243 but we couldn't updated the algorithm with unwrap_or(1) since that adds a new guarantee that isn't in the text, after this is approved I'll rebase that PR

DanielEScherzer:
> @rfcbot reviewed
> 
> I agree with making this guarantee. I don't have an opinion about whether we add it as a separate clause or as part of another clause -- @RalfJung's [suggestion here](https://github.com/rust-lang/reference/pull/2264#discussion_r3197184005) made sense to me.

Given that T-lang is okay with this new rule, can we merge this as-is and then I'll send a follow-up commit to combine the rules?

ehuss:
Can we just fix it here to avoid the churn? Is there some reason to need to do this now?

DanielEScherzer:
I figured that it would be simpler to merge the exact version that was voted on, and then simplify, rather than updating the version that was FCP'ed, but I can tweak it here too

traviscross (approved):

traviscross:
Thanks @DanielEScherzer.

traviscross:
cc @rust-lang/release

This should go in the release notes.

PR #2270, opened 2026-05-11
Fix the grammar of const param defaults by reusing rule for const args

fmease:
`struct T<const N: U = crate>;` is syntactically valid but the previous definition of rule *ConstParam* didn't capture that fact since it only listed *IDENTIFIER* while it needed to be *SimplePathSegment*.

### DIFF
--- a/src/paths.md
+++ b/src/paths.md
@@ -62,12 +62,11 @@ TypeList ->
     ( Type `,` )* Type `,`?
 
 GenericArg ->
-    Lifetime | Type | GenericArgsConst | GenericArgsBinding | GenericArgsBounds
+    Lifetime | Type | ConstArg | GenericArgsBinding | GenericArgsBounds
 
-GenericArgsConst ->
+ConstArg ->
       BlockExpression
-    | LiteralExpression
-    | `-` LiteralExpression
+    | `-`? LiteralExpression
     | SimplePathSegment
 
 GenericArgsBinding ->
--- a/src/items/generics.md
+++ b/src/items/generics.md
@@ -11,9 +11,7 @@ LifetimeParam -> Lifetime ( `:` LifetimeBounds )?
 
 TypeParam -> IDENTIFIER ( `:` Bounds? )? ( `=` Type )?
 
-ConstParam ->
-    `const` IDENTIFIER `:` Type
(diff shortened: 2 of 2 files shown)


src/paths.md
```diff
@@ -62,12 +62,11 @@ TypeList ->
     ( Type `,` )* Type `,`?
 
 GenericArg ->
-    Lifetime | Type | GenericArgsConst | GenericArgsBinding | GenericArgsBounds
+    Lifetime | Type | ConstArg | GenericArgsBinding | GenericArgsBounds
 
-GenericArgsConst ->
+ConstArg ->
```
fmease, line 67: *const arg* / *constant argument* is the term we use in rustc (internally and in diagnostics).

Moreover, in rustc's parser, *const arg* doesn't just refer to const arguments literally but also to const param defaults (and to the RHS of associated const bindings as in `TraitRef<CT = { 1 + 1 }>` but that's unstable syntax). So "reusing" *ConstArg* in rule *ConstParam* doesn't come out of thin air. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2270/ac7b991181147dbb499f462b2d837c2e964dbdc7..7aa59dad7ffd17359388c9b0e9855f8752b07841)*

fmease:
I found an issue: The "const arg" used for the param default should include `Self` while the "const arg proper" should not contain `Self` as that would conflict with type arguments. Let me think about how to solve this most nicely.

Edit: Well, as a matter of fact, *GenericArgsConst* already conflicts with *Type* on master ... I think remember a Zulip or GH discussion where ehuss (?) stated that it's still unclear how we want to represent identifier const args that can refer to either type or const in the grammar.

ehuss:
I don't remember if or where I may have said that.

One way to handle that would do the same as what was done in https://github.com/rust-lang/reference/pull/2022 for patterns. That is, to add a note indicating that there is an ambiguity for single-segment identifiers (with no generic args themselves), and that it is only resolved after name resolution. My understanding is that the disambiguation is done [here](https://github.com/rust-lang/rust/blob/1d59f669f9c699b982f112ebbc81e8b1eb550d68/compiler/rustc_resolve/src/late.rs#L1281-L1312), and it is based on which namespace it is from.

PR #2272, opened 2026-05-19
Clarify where `global_asm!` may be placed

ehuss:
The docs here were not clear about where `global_asm!` may be placed. It says the assembly is "emitted...outside a function", but doesn't exactly say where it is allowed.

The rules for this just changed in https://github.com/rust-lang/rust/pull/156582 which added the ability to place it in statement position.


folkertdev:
Note that we intend to revert this change in https://github.com/rust-lang/rust/pull/156884 because it did not go through FCP

PR #2274, opened 2026-05-24
Document RISC-V d, e, and f extensions

ehuss:
These are being stabilized in https://github.com/rust-lang/rust/pull/156188 as cfg-only features. It is a little odd to list them here in the target_feature attribute section since they can't be used with the attribute, but I think it makes sense to keep things together. Perhaps in the future we may consider breaking these lists out into a separate section and rearrange things.

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -270,6 +270,9 @@ r[attributes.codegen.target_feature.availability]
 
 The following is a list of the available feature names.
 
+r[attributes.codegen.target_feature.cfg-only]
+Target feature names marked as "(cfg only)" in this list may only be used with the [`target_feature`][cfg.target_feature] conditional compilation option, not with the `target_feature` attribute.
+
 r[attributes.codegen.target_feature.x86]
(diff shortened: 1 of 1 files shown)


ehuss:
@romancardenas can you review and confirm everything looks right here? In particular, I was uncertain about the E feature, since it has a slightly different way of being documented in the manual.

Also, I noticed that we are linking to the 20250508 tag, which is a bit old now. Would it make sense to change the links to a newer version? I've noticed these features have had some updates. I think the latest ratified version is v20260120.

romancardenas (approved):
LGTM! Regarding updating the links to the RISC-V specification, I think It should be OK. At least these extensions have not experienced too many changes.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2274/5d5be50fa6c77f2d31f6b1f24ad921f3261d9036..a61431c116d6d050613a6dd6b80997f3d27e8fd2)*

traviscross (approved):

PR #2278, opened 2026-06-01
Stabilize #[optimize] attribute

veluca93:


src/attributes/codegen.md
```diff
@@ -220,6 +220,51 @@ Only the first use of the `no_builtins` attribute has effect.
 > [!NOTE]
 > `rustc` lints against any use following the first.
 
+<!-- template:attributes -->
+r[attributes.codegen.optimize]
+## The `optimize` attribute
+
+r[attributes.codegen.optimize.intro]
+The *`optimize` [attribute]* suggests whether a function, inherent method, trait method, or closure should be optimized for size or speed.
```
ehuss: I'd like to find wording here that avoids attempting to duplicate what's specified in `attributes.codegen.optimize.allowed-positions`. I think in the intro we can be a little looser on the specifics, and that's what we seem to do in the other attributes. 

Additionally, I think it could mention the third mode.

So this could be something like:


```suggestion
The *`optimize` [attribute]* suggests whether a function should be optimized for size, speed, or with optimizations disabled.
```

src/attributes/codegen.md
```diff
@@ -220,6 +220,51 @@ Only the first use of the `no_builtins` attribute has effect.
...
+
+```grammar,attributes
+@root OptimizeAttribute ->
+      `optimize` `(` `size` `)`
+    | `optimize` `(` `speed` `)`
+    | `optimize` `(` `none` `)`
+```
+
+r[attributes.codegen.optimize.allowed-positions]
+The `optimize` attribute may only be applied to [closures], [async blocks], [free functions], [associated functions] in an [inherent impl] or [trait impl], and associated functions in a [trait definition].
```
ehuss:
```suggestion
The `optimize` attribute may only be applied to [closures], [async blocks], [free functions], [associated functions] in an [inherent impl] or [trait impl], and associated functions in a [trait definition] with a default body.
```

AIUI, this is not allowed:

```rust
trait T {
    #[optimize(speed)]
    fn f();
}
```

ehuss (commented):
There are a few other things that need updating:

- Update the list of allowed attributes on functions in [`items/functions.md`](https://github.com/rust-lang/reference/blob/6f2d1eac3b9b327536b7c481f7618f14ec83f314/src/items/functions.md?plain=1#L352-L367).
- Add to the list of [builtin attributes](https://github.com/rust-lang/reference/blob/6f2d1eac3b9b327536b7c481f7618f14ec83f314/src/attributes.md?plain=1#L253-L260).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2278/01b0ee707f4571e803c8b2c471d8335a448f5d60..59b5d8d771b5ccd9b7f4c4fc8e3f3e03897e4062)*

src/attributes/codegen.md
```diff
@@ -220,6 +220,51 @@ Only the first use of the `no_builtins` attribute has effect.
...
+
+r[attributes.codegen.optimize.intro]
+The *`optimize` [attribute]* suggests whether a function should be optimized for size, speed, or with optimizations disabled.
+
+> [!EXAMPLE]
+> ```rust
+> #[optimize(size)]
+> pub fn example1() {}
+>
+> #[optimize(none)]
```
DanielEScherzer, line 235: I would suggest putting `none` last, since it is probably going to be the least used (just a guess), and to match the order in the introduction

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2278/01b0ee707f4571e803c8b2c471d8335a448f5d60..e449c642c00918e383c0a71367d6afe401a069e7)*

PR #2279, opened 2026-06-04
Add more example target_arch values

patchunwrap:
These have been collected from
https://doc.rust-lang.org/stable/core/arch/index.html

Since mips was already on the list I figured I would add all target arch options have been added (even ones not available on stable Rust) and add a note to help the reader have the right expectations.

I found this when trying to figure out what the target_arch for riscv would be. I guess I kinda assumed it would be `riscv64gc`. If there is a different authoritative document somewhere, maybe this should link to that instead. But I figured the reference _was_ that authoritative document.

### DIFF
--- a/src/conditional-compilation.md
+++ b/src/conditional-compilation.md
@@ -97,15 +97,25 @@ r[cfg.target_arch.def]
 Key-value option set once with the target's CPU architecture. The value is similar to the first element of the platform's target triple, but not identical.
 
 r[cfg.target_arch.values]
-Example values:
+Example values (Note that not all of these are available on stable rust):
 
 * `"x86"`
 * `"x86_64"`
+* `"arm"`
+* `"aarch64"`
+* `"amdgpu"`
+* `"hexagon"`
+* `"riscv32"`
+* `"riscv64"`
 * `"mips"`
 * `"powerpc"`
 * `"powerpc64"`
-* `"arm"`
-* `"aarch64"`
+* `"nvptx"`
+* `"wasm"`
+* `"loongarch32"`
+* `"loongarch64"`
+* `"s390x"`
+
 
 r[cfg.target_feature]
 ### `target_feature`


folkertdev:
That is not even the full list, rust will print the full list when you do give it something it doesn't recognize

```
warning: unexpected `cfg` condition value: `x85`
 --> src/lib.rs:2:7
  |
2 | #[cfg(target_arch = "x85")]
  |       ^^^^^^^^^^^^^^-----
  |                     |
  |                     help: there is a expected value with a similar name: `"x86"`
  |
  = note: expected values for `target_arch` are: `aarch64`, `amdgpu`, `arm`, `arm64ec`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch32`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`, and `xtensa`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default
```

I'm not sure that the reference is the right place for this though.

patchunwrap:
> That is not even the full list, rust will print the full list when you do give it something it doesn't recognize

I was not aware there were so many.

> I'm not sure that the reference is the right place for this though.

I'm not sure either. I'm happy enough if you or another maintainer vetos this and it doesn't merge. I will say that when faced with this problem I came here first. The only other document that sorta specifies this information is the [rustc book, under platform support](https://doc.rust-lang.org/nightly/rustc/platform-support.html). But it only sorta specifies it indirectly.

At the very least if the goal is a _representative_ example set, rather than a _comprehensive_ one. I think it should include `aarch64`, `riscv64` and `wasm` simply because of how popular they are.

I would like to have a reference document somewhere listing it, if you agree and think somewhere else is a better location. I'm happy to make a pull request there. It's all up to you :)

patchunwrap:
This is a minor change and I don't want to waste the maintainers time. I'm going to close this now to reduce the list of pull requests :)

PR #2282, opened 2026-06-09
Document when `Box<T>` to `Box<U>` transmute is sound

traviscross:
We hadn't documented when it's sound to transmute a `Box<T>` to a `Box<U>`.  Now that we've documented when `*mut T` and `*mut U` can be assumed to have the same layout, let's use that to document when this transmute is sound.

Related to:

- https://github.com/rust-lang/rust/pull/155597
- https://github.com/rust-lang/rust/pull/157572

---

This is stacked on https://github.com/rust-lang/reference/pull/2292 (and on the ones it's stacked on).  Before being broken out, this PR originally covered what is now in PRs:

Not new lang guarantees:

- https://github.com/rust-lang/reference/pull/2286
- https://github.com/rust-lang/reference/pull/2287
- https://github.com/rust-lang/reference/pull/2288
- https://github.com/rust-lang/reference/pull/2290
- https://github.com/rust-lang/reference/pull/2289

New lang guarantees:

- https://github.com/rust-lang/reference/pull/2291
- https://github.com/rust-lang/reference/pull/2292
- https://github.com/rust-lang/reference/pull/2282 (This PR.)

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -147,7 +147,12 @@ r[undefined.validity.reference-box]
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
-  * Slice (`[T]`) and `str` metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], this metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
+  * Slice (`[T]`) and `str` metadata must be a valid `usize`.
+
+  In addition, for a wide reference or [`Box<T>`], the metadata is invalid if it makes the total size of the pointed-to value (as determined by `size_of_val`) bigger than `isize::MAX`.
+
+  > [!NOTE]
+  > This bound is on the size of the entire pointed-to value, not just its unsized tail, and it constrains `dyn Trait` metadata just as it does a slice or `str` length. A valid vtable describes an erased type no larger than `isize::MAX`, but a sized prefix can still carry the total past the limit.
 
 r[undefined.validity.valid-range]
 * If a type has a custom range of a valid values, then a valid value must be in that range. In the standard library, this affects [`NonNull<T>`] and [`NonZero<T>`].
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -147,7 +147,7 @@ r[undefined.validity.reference-box]
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
-  * Slice (`[T]`) metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], slice metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
+  * Slice (`[T]`) and `str` metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], this metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
 
 r[undefined.validity.valid-range]
 * If a type has a custom range of a valid values, then a valid value must be in that range. In the standard library, this affects [`NonNull<T>`] and [`NonZero<T>`].
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -100,7 +100,7 @@ r[undefined.dangling.zero-size]
 If the [size is 0][zero-sized], then the pointer is trivially never "dangling" (even if it is a null pointer).
 
 r[undefined.dangling.dynamic-size]
-Note that dynamically sized types (such as slices and strings) point to their entire range, so it is important that the length metadata is never too large.
+Note that dynamically sized types (such as slices and strings) point to their entire range, so it is important that the length [metadata] is never too large.
 
 r[undefined.dangling.alloc-limit]
 In particular, the dynamic size of a Rust value (as determined by `size_of_val`) must never exceed `isize::MAX`, since it is impossible for a single allocation to be larger than `isize::MAX`.
@@ -142,10 +142,10 @@ r[undefined.validity.union]
 * For a `union`, the exact validity requirements are not decided yet. Obviously, all values that can be created entirely in safe code are valid. If the union has a [zero-sized] field, then every possible value is valid. Further details are [still being debated](https://github.com/rust-lang/unsafe-code-guidelines/issues/438).
 
 r[undefined.validity.reference-box]
-* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the metadata). Note that the last point (about pointing to a valid value) remains a subject of some debate.
+* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the [metadata]). Note that the last point (about pointing to a valid value) remains a subject of some debate.
 
 r[undefined.validity.wide]
-* The metadata of a wide reference, [`Box<T>`], or raw pointer must match the type of the unsized tail:
+* The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the unsized tail:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
   * Slice (`[T]`) metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], slice metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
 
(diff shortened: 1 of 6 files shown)


src/dynamically-sized-types.md
```diff
@@ -8,9 +8,9 @@ r[dynamic-sized.restriction]
 Such types can only be used in certain cases:
 
 r[dynamic-sized.pointer-types]
-* [Pointer types] to <abbr title="dynamically sized types">DSTs</abbr> are sized but have twice the size of pointers to sized types
-    * Pointers to slices and `str` also store the number of elements.
-    * Pointers to trait objects also store a pointer to a vtable.
+* [Pointer types] to <abbr title="dynamically sized types">DSTs</abbr> are sized but have twice the size of pointers to sized types, since they also store *metadata*:
+    * Pointers to slices store the number of elements; pointers to `str` store the length in bytes.
+    * Pointers to trait objects store a pointer to a vtable.
```
Mark-Simulacrum, line 13: Comparing notes with https://doc.rust-lang.org/nightly/std/ptr/trait.Pointee.html#pointer-metadata, I think the only missing piece is the "unsized / last field" indirection for indirectly unsized types.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..b624dd7cb201c51f9237d0f5efc6424a74dc464d)*

src/behavior-considered-undefined.md
```diff
@@ -145,7 +145,7 @@ r[undefined.validity.reference-box]
 * A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the [metadata]). Note that the last point (about pointing to a valid value) remains a subject of some debate.
 
 r[undefined.validity.wide]
-* The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the unsized tail:
+* The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
   * Slice (`[T]`) metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], slice metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
```
Mark-Simulacrum: * This is missing the `str` case.
* The "invalid if total size" seems like it applies more generally to all metadata, including indirectly stored size metadata in the vtable? Though that becomes complicated since it depends on a deref to determine that the original thing is valid.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..b624dd7cb201c51f9237d0f5efc6424a74dc464d)*

src/special-types-and-traits.md
```diff
@@ -21,6 +21,31 @@ r[lang-types.box.fundamental]
...
+let boxed: Box<i8> = Box::new(1);
+let addr = (&raw const *boxed).addr();
+// SAFETY: `i8` and `u8` have the same size and the same alignment, so
+// `boxed`'s allocation is valid for `u8`.
+let reboxed: Box<u8> = unsafe { transmute(boxed) };
+// The transmute reuses the same allocation.
+assert_eq!((&raw const *reboxed).addr(), addr);
+assert_eq!(*reboxed, 1);
+
+// The pointee may also be unsized: `[u32]` and `[f32]` share a size and
```
Mark-Simulacrum: I don't think we currently guarantee align(u32) == align(f32): https://doc.rust-lang.org/nightly/reference/type-layout.html#r-layout.primitive

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..b624dd7cb201c51f9237d0f5efc6424a74dc464d)*

src/type-layout.md
```diff
@@ -69,6 +69,36 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
...
+    };
+}
+trait Tr1 {}
+trait Tr2 {}
+struct W<T: ?Sized> {
+    head: u8,
+    tail: T,
+}
+// Pointers to two types share a layout when the types are both sized,
+// both have a slice or `str` tail, or both have a trait object tail.
```
Mark-Simulacrum, line 93: Hm, I guess the pointers would share a layout even under -Zrandomize-layout? Though that may not guarantee transmutes are sound (since alignment could change).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..b624dd7cb201c51f9237d0f5efc6424a74dc464d)*

src/type-layout.md
```diff
@@ -69,6 +69,36 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
 > [!NOTE]
 > Though you should not rely on this, all pointers to <abbr title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of the size of `usize` and have the same alignment.
 
+r[layout.pointer.parametric]
+The raw pointer types `*const T` and `*const U` have the same layout, as do `*mut T` and `*mut U`, the shared references `&T` and `&U`, and the mutable references `&mut T` and `&mut U`, in each of the following cases: `T` and `U` are both sized types; `T` and `U` both have a [slice] or [`str`] as their [unsized tail]; or `T` and `U` both have a [trait object] as their unsized tail.
```
ehuss: Can the cases be broken out into a list instead of a semicolon?

Also, I'm unclear why the last two are separate. That is, could it be:

`T` and `U` both have a [slice], [`str`], or [trait object] as their [unsized tail]


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..5a667b161eb73e6e8c3f1371d29bc52d16b56ff7)*

src/special-types-and-traits.md
```diff
@@ -21,6 +21,31 @@ r[lang-types.box.fundamental]
 
 <!-- Editor Note: This is nowhere close to an exhaustive list -->
 
+r[lang-types.box.transmute]
+For types `T` and `U` such that `*mut T` and `*mut U` have the same layout (per [layout.pointer.parametric]) and a [pointer-to-pointer cast] from `*mut T` to `*mut U` is permitted, transmuting a `Box<T>` to a `Box<U>`, where both boxes use the global allocator, is sound when both of the following hold:
```
ehuss, line 25: Just curious, why "global" here?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..5a667b161eb73e6e8c3f1371d29bc52d16b56ff7)*

src/special-types-and-traits.md
traviscross: Thanks, yes.  Meant to assert this.  Done now.

src/special-types-and-traits.md
Mark-Simulacrum: Box with a custom allocator is not necessarily going to put the allocator in the same place -- e.g. under -Zrandomize-layout -- for different value types, so I don't think we can make this guarantee. Maybe it can be made for 1-ZST allocators? Not sure...

src/behavior-considered-undefined.md
traviscross: (As context for others, this is preexisting text.)

On the first point, thanks; I've added that.

On the second, the text requires that, for `dyn Trait`, the "metadata must be a pointer to a compiler-generated vtable for `Trait`".  Given that, think we still need to place further requirements on its contents?

src/type-layout.md
traviscross: Thanks; I've broken these out into bullet points.  These are stated separately as the *each of* is load bearing.

traviscross:
cc @RalfJung

src/behavior-considered-undefined.md
Mark-Simulacrum: For the second: yes, because the requirement here is placed on the _total size_ including the unsized tail. E.g., if my `dyn Trait` is `isize::MAX - 1` bytes that is only valid when combined with a 1-byte sized "head". The requirement for slice metadata and for dyn Trait always applies to both.

src/type-layout.md
```diff
@@ -69,6 +69,40 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
 > [!NOTE]
 > Though you should not rely on this, all pointers to <abbr title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of the size of `usize` and have the same alignment.
 
+r[layout.pointer.parametric]
+The raw pointer types `*const T` and `*const U` have the same layout, as do `*mut T` and `*mut U`, the shared references `&T` and `&U`, and the mutable references `&mut T` and `&mut U`, in each of the following cases:
+
+- `T` and `U` are both sized types.
+- `T` and `U` both have a [slice] or [`str`] as their [unsized tail].
```
nikomatsakis, line 76: I find this wording confusing (e.g., are you intending to guarantee that `&[u8]` and `&str` have the same layout)?

src/type-layout.md
Mark-Simulacrum: Discussed a bit offline -- we agreed that the only randomization that is theoretically possible is highly unlikely in practice: e.g., in theory, we could have the compiler have 3-usize-wide pointers or flip the metadata and pointer halves depending on the type of the pointee. So I'll resolve this thread.

src/type-layout.md
nikomatsakis: I guess you are, based on the test below.

nikomatsakis (commented):
My vibe is that the final decision on what types can be converted safely to one another is ideally a T-opsem call. 

However, the text itself made a lot of sense to me at a high level!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/6f2d1eac3b9b327536b7c481f7618f14ec83f314..6ca17b8fc1f17a3e6d8b6635d2a3a2f2609d77be)*

src/special-types-and-traits.md
scottmcm: Big fan of sticking to just `Global` here for now.  We can talk about the more complex cases later.

(See the similar discussion about how we might be fine in https://github.com/rust-lang/rfcs/pull/3775 to define the field order of `*const [T]` but are not ready to define the field order of `Vec<T>`.)

src/behavior-considered-undefined.md
traviscross: Yes, thanks for that... now I remember.  I've fixed this and added an admonition with the explanation.

src/type-layout.md
traviscross: We already [guarantee](https://doc.rust-lang.org/nightly/reference/type-layout.html#r-layout.str) that:

> A reference `&str` has the same layout as a reference `&[u8]`.

This PR further guarantees that pointers with any slice-type metadata have the same layout as pointers with `str` metadata.

src/dynamically-sized-types.md
traviscross: Thanks.  Added.

src/special-types-and-traits.md
```diff
@@ -21,6 +21,33 @@ r[lang-types.box.fundamental]
...
```
RalfJung, line 25: > and a [pointer-to-pointer cast] from `*mut T` to `*mut U` is permitted

Shouldn't this refer to the ptr-to-ptr transmute?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2282/2c27905c15a51983b54d84f050d3bda096194d27..fe3b466c5b0ab7bee4a19337b472a02e6bfbe42b)*

PR #2287, opened 2026-06-16
Define an unsized tail

traviscross:
We refer to the unsized tail of a type, but we hadn't defined it.  Let's do that and link to the definition.

---

This is stacked on https://github.com/rust-lang/reference/pull/2286 and that should be merged first.

### DIFF
--- a/src/dynamically-sized-types.md
+++ b/src/dynamically-sized-types.md
@@ -25,10 +25,17 @@ r[dynamic-sized.struct-field]
 > [!NOTE]
 > [Variables], function parameters, [const] items, and [static] items must be `Sized`.
 
+r[dynamic-sized.tail]
+The *unsized tail* of a type is the dynamically sized component that the [metadata] of a pointer to the type describes. A [slice] (`[T]`) and a [`str`] are each their own unsized tail, described by a length; a [trait object] (`dyn Trait`) is its own unsized tail, described by a pointer to a vtable. When a struct (per [dynamic-sized.struct-field]) or a tuple has an unsized last field, its unsized tail is the unsized tail of that field. A sized type has no unsized tail.
+
+[metadata]: dynamic-sized.pointer-types
 [sized]: special-types-and-traits.md#sized
 [Slices]: types/slice.md
+[slice]: types/slice.md
 [str]: types/str.md
+[`str`]: types/str.md
 [trait objects]: types/trait-object.md
+[trait object]: types/trait-object.md
 [Pointer types]: types/pointer.md
 [Variables]: variables.md
 [const]: items/constant-items.md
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2287/2c27905c15a51983b54d84f050d3bda096194d27..e3cc399c8d17739bbe1228b6099ad7e03e918d57)*

src/dynamically-sized-types.md
```diff
@@ -25,10 +25,17 @@ r[dynamic-sized.struct-field]
 > [!NOTE]
 > [Variables], function parameters, [const] items, and [static] items must be `Sized`.
 
+r[dynamic-sized.tail]
+The *unsized tail* of a type is the dynamically sized component that the [metadata] of a pointer to the type describes. A [slice] (`[T]`) and a [`str`] are each their own unsized tail, described by a length; a [trait object] (`dyn Trait`) is its own unsized tail, described by a pointer to a vtable. When a struct (per [dynamic-sized.struct-field]) or a tuple has an unsized last field, its unsized tail is the unsized tail of that field. A sized type has no unsized tail.
```
RalfJung, line 29: This definition contains redundant information with the definition of "metadata" from https://github.com/rust-lang/reference/pull/2286. That doesn't seem great, we should define these things only in one place and then reference them elsewhere.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2287/be573a7ea3a0444c0fc98c7460f0587f801f313b..39fa8df6318c878f917537dbd48a4ec8be7ac435)*

traviscross: Agreed.  I had the same thought.  I didn't have any immediate good ideas for how to resolve it, though (other than ensuring they were cross-linked, as I did).  Suggestions welcome.

RalfJung: We could define "unsized tail" without repeating the metadata definition. For instance:

The *unsized tail* of an unsized type is the unsized type at the "end" of the type, after recursing through compound types:
-  A [slice] (`[T]`), a [`str`], and a [trait object] (`dyn Trait`) are each their own unsized tail.
- When a struct (per [dynamic-sized.struct-field]) or a tuple has an unsized last field, its unsized tail is the unsized tail of that field.
- A sized type has no unsized tail.

And then for metadata I'd say something like:
[Pointer types] to <abbr title="dynamically sized types">DSTs</abbr> are sized but have twice the size of pointers to sized types, since they also store *metadata*.
The metadata carries by a pointer is determined by the [unsized tail] of its pointee type:
- If the unsized tail is a slice or `str`, ...
- If the unsized tail is a trait object, ...

PR #2288, opened 2026-06-16
Cover `str` in the wide-pointer metadata rule

traviscross:
The validity rule for the metadata of a wide reference, `Box<T>`, or raw pointer mentions `dyn Trait` and slice but had omitted `str`.  Let's fix that.

---

I'm breaking this out from https://github.com/rust-lang/reference/pull/2282 so that we can merge the prerequisites before  considering the new lang guarantees.

This is stacked on https://github.com/rust-lang/reference/pull/2286 and https://github.com/rust-lang/reference/pull/2287 and those should be merged first.

cc @ehuss @RalfJung @Mark-Simulacrum

PR #2289, opened 2026-06-16
Document metadata of ptrs to indirectly unsized types

traviscross:
We say that pointers to DSTs store metadata and what that metadata is for pointers to slices, `str`, and trait objects.  But a struct or tuple with an unsized tail is itself a DST, and we hadn't said what the metadata is for pointers to these unsized types.

Now that we've defined *metadata* and *unsized tail*, let's complete this enumeration.

---

This is stacked on https://github.com/rust-lang/reference/pull/2286 and https://github.com/rust-lang/reference/pull/2287 and https://github.com/rust-lang/reference/pull/2288 and https://github.com/rust-lang/reference/pull/2290 (yes, that one is out of order) and those should be merged first.


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2289/2c27905c15a51983b54d84f050d3bda096194d27..f0c3d8cceee0e61e37cad4cdf70d07e5bd67dc8d)*

PR #2290, opened 2026-06-16
Generalize the wide-pointer total-size bound

traviscross:
We document that, for references and `Box<T>`, pointed-to values with slice or `str` metadata must
be no larger than `isize::MAX`.  We hadn't required this for pointed-to values with `dyn` metadata.
It's tempting to think this isn't necessary since we separately require that the metadata point to
a vtable generated by the compiler, which ensures the encoded size of the erased type is OK.

But the bound is on the total size of the pointed-to value, including any sized prefix of a type
with an unsized tail.  Since the prefix combined with the size in the vtable can push us past the
limit, we need the separate restriction.

Let's apply the rule to both cases and add an admonition to remind ourselves of why this is needed.

---

This is stacked on https://github.com/rust-lang/reference/pull/2286 and https://github.com/rust-lang/reference/pull/2287 and https://github.com/rust-lang/reference/pull/2288 and those should be merged first.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -147,7 +147,12 @@ r[undefined.validity.reference-box]
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
-  * Slice (`[T]`) and `str` metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], this metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
+  * Slice (`[T]`) and `str` metadata must be a valid `usize`.
+
+  In addition, for a wide reference or [`Box<T>`], the metadata is invalid if it makes the total size of the pointed-to value (as determined by `size_of_val`) bigger than `isize::MAX`.
+
+  > [!NOTE]
+  > This bound is on the size of the entire pointed-to value, not just its unsized tail, and it constrains `dyn Trait` metadata just as it does a slice or `str` length. A valid vtable describes an erased type no larger than `isize::MAX`, but a sized prefix can still carry the total past the limit.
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2290/2c27905c15a51983b54d84f050d3bda096194d27..df82d18fd945544b9e357846fb31f032e8ad4da0)*

src/behavior-considered-undefined.md
```diff
@@ -147,7 +147,12 @@ r[undefined.validity.reference-box]
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
   * `dyn Trait` metadata must be a pointer to a compiler-generated vtable for `Trait`. (For raw pointers, this requirement remains a subject of some debate.)
-  * Slice (`[T]`) and `str` metadata must be a valid `usize`. Furthermore, for wide references and [`Box<T>`], this metadata is invalid if it makes the total size of the pointed-to value bigger than `isize::MAX`.
+  * Slice (`[T]`) and `str` metadata must be a valid `usize`.
+
+  In addition, for a wide reference or [`Box<T>`], the metadata is invalid if it makes the total size of the pointed-to value (as determined by `size_of_val`) bigger than `isize::MAX`.
```
RalfJung, line 152: FWIW I would consider all of this a derived property. It follows from the fact that wide references and `Box` must be *dereferenceable* for the actual dynamic size given by their pointee type and metadata, and it is impossible to be dereferenceable for more than `isize::MAX` bytes as that is the largest an allocation can be.

So, if this is explicitly stated at all, IMO it should only be done *non-normatively*, as the normative part should avoid such redundancies.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2290/9a1026011f07cd0a2be08f2e7dd4c3ceff98543b..4a5f81c16ff071420a9acca11f406d773d246fb2)*

RalfJung: Opened https://github.com/rust-lang/reference/pull/2352 to fix this

PR #2291, opened 2026-06-16
Document layout equivalences for pointer types

traviscross:
We guarantee various things about the layout of pointers and references, but we'd made no guarantees of equivalence between two pointers to distinct unsized types that differ in the unsized tail.  Let's make some useful guarantees about this.

---

I'm breaking this out from https://github.com/rust-lang/reference/pull/2282 so that these may be considered in smaller pieces.  This is stacked on https://github.com/rust-lang/reference/pull/2289 (and on the ones it's stacked on).

### DIFF
--- a/src/type-layout.md
+++ b/src/type-layout.md
@@ -69,6 +69,40 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
 > [!NOTE]
 > Though you should not rely on this, all pointers to <abbr title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of the size of `usize` and have the same alignment.
 
+r[layout.pointer.parametric]
+The raw pointer types `*const T` and `*const U` have the same layout, as do `*mut T` and `*mut U`, the shared references `&T` and `&U`, and the mutable references `&mut T` and `&mut U`, in each of the following cases:
+
+- `T` and `U` are both sized types.
+- `T` and `U` both have a [slice] or [`str`] as their [unsized tail].
+- `T` and `U` both have a [trait object] as their unsized tail.
+
+```rust
+# use core::alloc::Layout;
+macro_rules! assert_layout_eq {
+    ($a:ty, $b:ty) => {
+        assert_eq!(Layout::new::<$a>(), Layout::new::<$b>())
+    };
+}
+trait Tr1 {}
+trait Tr2 {}
+struct W<T: ?Sized> {
+    head: u8,
+    tail: T,
+}
+// Pointers to two types share a layout when the types are both sized,
+// both have a slice or `str` tail, or both have a trait object tail.
+assert_layout_eq!(*const u8, *const u64);
+assert_layout_eq!(*const [u8], *const [u64]);
+assert_layout_eq!(*const str, *const [u8]);
+assert_layout_eq!(*const dyn Tr1, *const dyn Tr2);
+assert_layout_eq!(*const W<[u8]>, *const W<[u64]>);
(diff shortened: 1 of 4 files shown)


src/type-layout.md
```diff
@@ -69,6 +69,40 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
 > [!NOTE]
 > Though you should not rely on this, all pointers to <abbr title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of the size of `usize` and have the same alignment.
 
+r[layout.pointer.parametric]
+The raw pointer types `*const T` and `*const U` have the same layout, as do `*mut T` and `*mut U`, the shared references `&T` and `&U`, and the mutable references `&mut T` and `&mut U`, in each of the following cases:
+
+- `T` and `U` are both sized types.
+- `T` and `U` both have a [slice] or [`str`] as their [unsized tail].
+- `T` and `U` both have a [trait object] as their unsized tail.
```
weiznich, line 77: Doesn't this also require that the VTable is the same between both traits? Otherwise you would end up having exactly [this](https://github.com/rust-lang/rust/pull/155597#discussion_r3362093603) problem? Or does this just refer to the fact that the internal layout of the vtable is so that you could convert between both, but you get the "wrong" behaviour in this cases?



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2291/2c27905c15a51983b54d84f050d3bda096194d27..d3d70fbd5d2ad1afcdb63a4bb30aa788b541a6af)*

traviscross: This is only a guarantee about the layout of wide pointers.  It doesn't guarantee that you can transmute, and we exclude trait object unsized tails in https://github.com/rust-lang/reference/pull/2292/.

RalfJung:
Given that the PR this is stacked on landed, it'd be great if you could rebase this PR to make the diff view reflect just what is new.

src/type-layout.md
```diff
@@ -69,6 +69,40 @@ Pointers to unsized types are sized. The size and alignment of a pointer to an u
 > [!NOTE]
 > Though you should not rely on this, all pointers to <abbr title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of the size of `usize` and have the same alignment.
 
+r[layout.pointer.parametric]
+The raw pointer types `*const T` and `*const U` have the same layout, as do `*mut T` and `*mut U`, the shared references `&T` and `&U`, and the mutable references `&mut T` and `&mut U`, in each of the following cases:
```
RalfJung, line 73: "Same layout" here seems to mean "same size and alignment", based on the assertions below? That's not really useful on its own though. And it's really hard to define "same layout" in general for two arbitrary types. 

So, I am honestly not sure what this is supposed to guarantee. It is much easier to precisely say "you may transmute from A to B under precondition P, and then X will happen with the value".

Unless you really just mean "same size and alignment", in which case I think we should say that explicitly rather than saying using undefined terms such as "same layout".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2291/2c27905c15a51983b54d84f050d3bda096194d27..d3d70fbd5d2ad1afcdb63a4bb30aa788b541a6af)*

src/type-layout.md
RalfJung: > It doesn't guarantee that you can transmute,

That might be worth making explicit, since people often take "same layout" as meaning "I may transmute".

PR #2283, opened 2026-06-12
Update for stabilization of the never type

ehuss:
This updates for the stabilization of the never type proposed in https://github.com/rust-lang/rust/pull/155499.

cc @WaffleLapkin

PR #2293, opened 2026-06-17
Empty repr(Rust) enums are ZSTs

zachs18:
Guarantee that enums with Rust representation with no variants are ZSTs.

Followup to https://github.com/rust-lang/reference/pull/2262 , (which guaranteed the same for enums with Rust representation with a single variant whose fields are all ZSTs).

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -211,7 +211,8 @@ r[glossary.uninhabited]
 
 A type is uninhabited if it has no constructors and therefore can never be instantiated. An uninhabited type is "empty" in the sense that there are no values of the type. The canonical example of an uninhabited type is the [never type] `!`, or an enum with no variants `enum Never { }`. Opposite of [Inhabited](#inhabited).
 
-> [!NOTE] Uninhabited types are not necessarily [zero-sized][glossary.zst]. For example, `enum Never { }` is uninhabited and zero-sized, but `(u8, Never)` is uninhabited and not zero-sized.
+> [!NOTE]
+> Uninhabited types are not necessarily [zero-sized][glossary.zst]. For example, `enum Never { }` is uninhabited and zero-sized, but `(u8, Never)` is uninhabited and not zero-sized.
 
 r[glossary.zst]
 ### Zero-sized type (ZST)
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -223,6 +223,7 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 - `repr(Rust)` [structs] with no fields or where all fields are zero sized (see [layout.repr.rust.struct-zst]).
 - `repr(C)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
 - `repr(transparent)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
+- `repr(Rust)` [enums] (without a [primitive representation] specified) with no variants (see [layout.repr.rust.enum-empty-zst])
 - `repr(Rust)` [enums] (without a [primitive representation] specified) with a single [field-struct-like variant], a single [unit-struct-like variant], or a single [tuple-struct-like variant] and where the struct-like thing has no fields or where all of the fields are zero sized (see [layout.repr.rust.enum-struct-like-zst]).
 - [Arrays] of zero-sized types (see [layout.array]).
 - [Arrays] of length zero (see [layout.array]).
@@ -284,6 +285,8 @@ enum E5 {
 enum E6 {
     V1 (),
 }
+# /// An enum with no variants.
+enum E7 {}
 
 assert_eq!(0, size_of::<()>());
 assert_eq!(0, size_of_val(&f));
@@ -304,6 +307,7 @@ assert_eq!(0, size_of::<E3>());
 assert_eq!(0, size_of::<E4>());
 assert_eq!(0, size_of::<E5>());
 assert_eq!(0, size_of::<E6>());
+assert_eq!(0, size_of::<E7>());
 ```
 
 [`extern` blocks]: items.extern
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -211,6 +211,8 @@ r[glossary.uninhabited]
 
 A type is uninhabited if it has no constructors and therefore can never be instantiated. An uninhabited type is "empty" in the sense that there are no values of the type. The canonical example of an uninhabited type is the [never type] `!`, or an enum with no variants `enum Never { }`. Opposite of [Inhabited](#inhabited).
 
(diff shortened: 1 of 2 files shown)


zachs18:
@rustbot ready

Followup to #2262 to guarantee that zero-variant `repr(Rust)` enums are zero-sized. Rebased since #2262 merged, using similar phrasing as in that PR.

zachs18:
cc @traviscross As per https://github.com/rust-lang/reference/pull/2262#issuecomment-4727106611 , here's a followup PR to extend the ZST guarantee to zero-variant `repr(Rust)` enums.

scottmcm:
This seems obvious to me: of course an uninhabitable type with no variants and no fields needn't take up any space, so might as well be guaranteed a ZST.  Especially since this doesn't say anything about alignment it seems fine.

@rfcbot merge lang

(Since 1-ZSTs are extra meaningful in various ways it'd probably also be fine to make these 1-ZSTs specifically, not just ZSTs, but let's accept the proposal here for just "size ≡ 0" for now since we have it in front of us.)

traviscross:
@rfcbot reviewed

joshtriplett:
These types are uninhabited, but for the purposes of layout, yes, they should not affect the size (or alignment) of whatever they're put inside.

@rfcbot reviewed

nikomatsakis:
@rfcbot reviewed

steffahn:
I would appreciate if there’s *some* (at least minor) indication about this case being an *uninhabited* type, too; because those feel different from the more typical case of a “ZST”. Also, looking up at the previous section about uninhabited types, which lists `!` and variant-less enums, it feels inconsistent not to list `!` here as well? Or is the size of `!` not guaranteed?

Maybe the (or part of the) callout needed would just be in that previous section about uninhabited types, noting the fact that types can be uninhabited and zero-sized at the same time. (And that uninhabited types *often* - but not always - are zero-sized.)

zachs18:
> I would appreciate if there’s some (at least minor) indication about this case being an uninhabited type, too

I added a note to `layout.repr.rust.enum-empty-zst` with a link to the "uninhabited" glossary section.

> it feels inconsistent not to list ! here as well? Or is the size of ! not guaranteed?

I think `!` should be listed in the primitives section, not the `repr(Rust)` section. I'll make a separate PR for that, since this one is already in FCP.

> Maybe the (or part of the) callout needed would just be in that previous section about uninhabited types, noting the fact that types can be uninhabited and zero-sized at the same time. (And that uninhabited types often - but not always - are zero-sized.)

~~I'll do that in the `!` PR~~

zachs18:
Hmm, maybe the `!` layout guarantees should be in or after https://github.com/rust-lang/reference/pull/2283 ?

I'll put the note that that types can be uninhabited and zero-sized at the same time in this PR then.

src/type-layout.md
```diff
@@ -183,6 +183,11 @@ Be aware that this guarantee does not imply that the fields have distinct addres
 r[layout.repr.rust.struct-zst]
 For [structs] with no fields or where all fields are [zero sized], it is further guaranteed that the structs are themselves [zero sized].
 
+r[layout.repr.rust.enum-empty-zst]
+For [enums] (without a [primitive representation] specified) with no variants, the enums themselves are [zero sized].
+
+Note that such enums are [uninhabited].
```
traviscross:
```suggestion
> [!NOTE]
> Such enums are [uninhabited].
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2293/1358a7722595cce68ac003616c62ec17bd890e00..5c793073e6b3cdadc2ba97b347f77e28b6d0f6d4)*

src/glossary.md
```diff
@@ -211,6 +211,8 @@ r[glossary.uninhabited]
 
 A type is uninhabited if it has no constructors and therefore can never be instantiated. An uninhabited type is "empty" in the sense that there are no values of the type. The canonical example of an uninhabited type is the [never type] `!`, or an enum with no variants `enum Never { }`. Opposite of [Inhabited](#inhabited).
 
+> [!NOTE] Uninhabited types are not necessarily [zero-sized][glossary.zst]. For example, `enum Never { }` is uninhabited and zero-sized, but `(u8, Never)` is uninhabited and not zero-sized.
```
zachs18:
```suggestion
> [!NOTE]
> Uninhabited types are not necessarily [zero-sized][glossary.zst]. For example, `enum Never { }` is uninhabited and zero-sized, but `(u8, Never)` is uninhabited and not zero-sized.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2293/1358a7722595cce68ac003616c62ec17bd890e00..5c793073e6b3cdadc2ba97b347f77e28b6d0f6d4)*

src/glossary.md
```diff
@@ -223,6 +226,7 @@ A type is zero sized (a ZST) if its size is 0. Such types have at most one possi
 - `repr(Rust)` [structs] with no fields or where all fields are zero sized (see [layout.repr.rust.struct-zst]).
 - `repr(C)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.c.struct.size-field-offset]).
 - `repr(transparent)` [structs] with no fields or where all fields are zero-sized (see [layout.repr.transparent.layout-abi]).
+- `repr(Rust)` [enums] (without a [primitive representation] specified) with no variants (see [layout.repr.rust.enum-empty-zst])
```
steffahn: @zachs18 

> > it feels inconsistent not to list ! here as well? Or is the size of ! not guaranteed?
> 
> I think `!` should be listed in the primitives section, not the `repr(Rust)` section. I'll make a separate PR for that, since this one is already in FCP.

I was looking at the list in the glossary, actually ^^ - sorry for any potential confusion by not including a more precise reference to what I was talking about.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2293/1358a7722595cce68ac003616c62ec17bd890e00..19075c89d116ec6603679e5898e60c5f31b0a219)*

steffahn: By having `!` in the previous glossary entry, but not this section, one *could* be made to wonder whether it’s intentionally not included here as any sort of meaningful distinction. (I know these are just lists of examples, technically.)

zachs18: Added in https://github.com/rust-lang/reference/pull/2309

DanielEScherzer (approved):
makes sense to me, language looks good to me assuming lang sign off

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2293/1358a7722595cce68ac003616c62ec17bd890e00..19075c89d116ec6603679e5898e60c5f31b0a219)*

DanielEScherzer:
@rustbot label -S-waiting-on-author
Seems to have been present from the start when this was a draft PR, no longer waiting

traviscross (approved):

traviscross:
Thanks @zachs18.

PR #2296, opened 2026-06-24
Add LoongArch LSX/LASX inline asm documentation

heiher:
Update the inline assembly reference to document LoongArch LSX (`vreg`) and LASX (`xreg`) register classes, their supported operand types, register aliases with FPU registers, and the `w`/`u` template modifiers.

Tracking issue: https://github.com/rust-lang/rust/issues/133416


heiher:
Since `vreg` and `xreg` are now gated behind `asm_experimental_reg`, I don't think this change is needed yet. We can update the reference when they are stabilized. Thanks!

PR #2297, opened 2026-06-25
remove broken text

tshepang:
It had a typo, but it should also not be needed since it should be clear from context what it explains


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2297/e097e1b29dc300f1979ba4b9f7cf39115d6ada12..160551d8f70732dc3adb7b91eebe9f80b6ac4dbd)*

PR #2299, opened 2026-06-25
fix link target

tshepang:


ehuss (approved):
Thanks! In a sense, the path patterns section is also correct, since from a pattern standpoint, it is a path pattern. But it also seems fine to link to the more-specific section, it's just that the constant patterns section doesn't stand on its own very well.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2299/e097e1b29dc300f1979ba4b9f7cf39115d6ada12..86eaede6f374147a02f5cf905e84ef8ddb2ef9d6)*

PR #2300, opened 2026-06-27
Add `extern "custom"`

folkertdev:
stabilization PR: https://github.com/rust-lang/rust/pull/158504

This is a tricky one that is an exception to some rules. I've tried to put the raw information in, but this'll probably require some refinement.

### DIFF
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -48,7 +48,7 @@ r[items.extern.fn.param-patterns]
 Patterns are not allowed in parameters, only [IDENTIFIER] or `_` may be used.
 
 r[items.extern.fn.qualifiers]
-The `safe` and `unsafe` function qualifiers are allowed, but other function qualifiers (e.g. `const`, `async`, `extern`) are not.
+The `safe` and `unsafe` function qualifiers are allowed, but other function qualifiers (e.g. `const`, `async`, `extern`) are not. The `safe` qualifier is rejected in `extern "custom"` blocks.
 
 r[items.extern.fn.foreign-abi]
 Functions within external blocks may be called by Rust code, just like functions defined in Rust. The Rust compiler automatically translates between the Rust ABI and the foreign ABI.
@@ -111,6 +111,9 @@ r[items.extern.abi.system]
 
 r[items.extern.abi.unwind]
 * `extern "C-unwind"` and `extern "system-unwind"` --- Identical to `"C"` and `"system"`, respectively, but with [different behavior][unwind-behavior] when the callee unwinds (by panicking or throwing a C++ style exception).
+
+r[items.extern.abi.custom]
+* `unsafe extern "custom"` --- A custom ABI that is not known to the rust compiler.
 
 r[items.extern.abi.platform]
 There are also some platform-specific ABI strings:
--- a/src/items/external-blocks.md
+++ b/src/items/external-blocks.md
@@ -113,7 +113,7 @@ r[items.extern.abi.unwind]
 * `extern "C-unwind"` and `extern "system-unwind"` --- Identical to `"C"` and `"system"`, respectively, but with [different behavior][unwind-behavior] when the callee unwinds (by panicking or throwing a C++ style exception).
 
 r[items.extern.abi.custom]
-* `unsafe extern "custom"` --- A custom ABI that is not known to the rust compiler.
+* `unsafe extern "custom"` --- A custom ABI that is not known to the compiler.
 
 r[items.extern.abi.platform]
 There are also some platform-specific ABI strings:
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -239,6 +239,15 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
 
 For other considerations and limitations regarding unwinding across FFI boundaries, see the [relevant section in the Panic documentation][panic-ffi].
 
+r[items.fn.extern.custom]
+An `extern "custom"` function has an unknown, custom ABI. The only way to call such a function is via [inline assembly].
+
+r[items.fn.extern.custom.safety]
+An `extern "custom"` function must be `unsafe`.
+
+r[items.fn.extern.custom.naked]
+An `extern "custom"` function definition must be a [naked function].
+
 [forced-unwinding]: https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html#forced-unwinding
 [panic handler]: ../panic.md#the-panic_handler-attribute
 [panic-ffi]: ../panic.md#unwinding-across-ffi-boundaries
@@ -411,6 +420,7 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [testing attributes]: ../attributes/testing.md
 [`cold`]: ../attributes/codegen.md#the-cold-attribute
 [`inline`]: ../attributes/codegen.md#the-inline-attribute
+[naked function]: ../attributes/codegen.md#the-naked-attribute
 [`deprecated`]: ../attributes/diagnostics.md#the-deprecated-attribute
 [`doc`]: ../../rustdoc/the-doc-attribute.html
 [`must_use`]: ../attributes/diagnostics.md#the-must_use-attribute
@@ -427,3 +437,4 @@ fn foo_oof(#[some_inert_attribute] arg: u8) {
 [variadic function]: external-blocks.md#variadic-functions
 [`extern` block]: external-blocks.md
 [zero-sized]: glossary.zst
+[inline assembly]: ../inline-assembly.md
--- a/src/items/functions.md
+++ b/src/items/functions.md
@@ -245,6 +245,39 @@ r[items.fn.extern.custom]
 r[items.fn.extern.custom.intro]
 An `extern "custom"` function has an unknown, custom ABI. The only way to call such a function is via [inline assembly].
 
+> [!EXAMPLE]
+> ```rust
+> # #[cfg(target_arch = "x86_64")] {
+> # use core::arch::{asm, naked_asm};
+> #
+> /// Adds 1 to `rax`.
+> ///
+> /// This function uses a custom calling convention: the argument is
+> /// passed in `rax`, the result is returned in `rax`, the flags may
+> /// be clobbered, and all other registers are preserved.
+> #[unsafe(naked)]
+> unsafe extern "custom" fn increment() {
+>     naked_asm!(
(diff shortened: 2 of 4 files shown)


src/items/external-blocks.md
tgross35, line 1: Maybe it would be good to have an example at `src/inline-assembly.md` alongside naked functions?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/86635e30bf861a038dc197d7e16fd09e7e514e7a..0a83cc3e74c047aa6a27a5823ec291c032fa072d)*

src/items/functions.md
```diff
@@ -239,6 +239,15 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
 
 For other considerations and limitations regarding unwinding across FFI boundaries, see the [relevant section in the Panic documentation][panic-ffi].
 
+r[items.fn.extern.custom]
+An `extern "custom"` function has an unknown, custom ABI. The only way to call such a function is via [inline assembly].
```
tgross35, line 246: Maybe a usage hint:

> ... unknown, custom ABI. This can be used to implement calling conventions unsupported by `rustc`. The only way to call such a function is via [inline assembly].

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/86635e30bf861a038dc197d7e16fd09e7e514e7a..0a83cc3e74c047aa6a27a5823ec291c032fa072d)*

src/items/external-blocks.md
tgross35, line 1: Somewhere in the reference it would probably be good to clarify the purpose of parameters and return types in `extern "custom"` definitions, since they don't directly serve any purpose. I feel like we discussed this somewhere but can't seem to find it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/86635e30bf861a038dc197d7e16fd09e7e514e7a..0a83cc3e74c047aa6a27a5823ec291c032fa072d)*

folkertdev: We actually disallow them entirely

```
error: invalid signature for `extern "custom"` function
 --> <source>:6:31
  |
6 | unsafe extern "custom" fn foo(a: i32) -> i32 {
  |                               ^^^^^^     ^^^
  |
  = note: functions with the "custom" ABI cannot have any parameters or return type
help: remove the parameters and return type
  |
6 - unsafe extern "custom" fn foo(a: i32) -> i32 {
6 + unsafe extern "custom" fn foo() {
  |
```

src/items/external-blocks.md
```diff
@@ -112,6 +112,9 @@ r[items.extern.abi.system]
 r[items.extern.abi.unwind]
 * `extern "C-unwind"` and `extern "system-unwind"` --- Identical to `"C"` and `"system"`, respectively, but with [different behavior][unwind-behavior] when the callee unwinds (by panicking or throwing a C++ style exception).
 
+r[items.extern.abi.custom]
+* `unsafe extern "custom"` --- A custom ABI that is not known to the rust compiler.
```
traviscross:
```suggestion
* `unsafe extern "custom"` --- A custom ABI that is not known to the compiler.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/86635e30bf861a038dc197d7e16fd09e7e514e7a..0a83cc3e74c047aa6a27a5823ec291c032fa072d)*

src/items/functions.md
```diff
@@ -239,6 +239,15 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
...
+
+r[items.fn.extern.custom.safety]
+An `extern "custom"` function must be `unsafe`.
+
+r[items.fn.extern.custom.naked]
+An `extern "custom"` function definition must be a [naked function].
+
```
traviscross, line 302: We need to add the rules here about what's allowed or not allowed in terms of parameters and return types.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/86635e30bf861a038dc197d7e16fd09e7e514e7a..0a83cc3e74c047aa6a27a5823ec291c032fa072d)*

traviscross: @rustbot author

traviscross: And we need to add text about the function pointer restrictions as well.

folkertdev: where should that function pointer text go? Here on the item and/or in `external-blocks.md`, or somewhere else?

traviscross: I've pushed a commit with it.  I consolidated the signature rules into one rule on function items, then cite that normatively from the function pointer types chapter.

src/items/functions.md
```diff
@@ -239,6 +239,22 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
...
 For other considerations and limitations regarding unwinding across FFI boundaries, see the [relevant section in the Panic documentation][panic-ffi].
 
+r[items.fn.extern.custom]
+### Extern "custom"
+
+r[items.fn.extern.custom.intro]
+An `extern "custom"` function has an unknown, custom ABI. The only way to call such a function is via [inline assembly].
+
+r[items.fn.extern.custom.signature]
+The signature of an `extern "custom"` function must:
```
DanielEScherzer: Are these really properties of the "signature", rather than the function itself? I.e. the signature specifies that the function is `unsafe`, but the "signature" isn't unsafe, the function is unsafe


```suggestion
An `extern "custom"` function must:
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/2ed1cb8b19d372ee586bd8bc138aa0b43fa6ad15..112dfa29520a41b19ce7e6839a992c3210a39cd0)*

src/items/external-blocks.md
traviscross: Agreed.  Pushed a commit to add one.

src/items/functions.md
traviscross: I pushed a commit adding an admonition to naked functions that suggests when `extern "custom"` functions are useful.

src/items/functions.md
traviscross: Thanks.  Changed, and made the parallel change in the function pointer types chapter.

traviscross (approved):

src/items/functions.md
```diff
@@ -239,6 +239,55 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
...
+> assert_eq!(x, 42);
+> # }
+> ```
+
+r[items.fn.extern.custom.signature]
+An `extern "custom"` function must:
+
+- Be `unsafe`.
+- Not have any parameters.
+- Return the [unit type].
```
tgross35: Needs some consideration for generics, https://github.com/rust-lang/rust/pull/158504#issuecomment-5229649454 and the comment below

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/2ed1cb8b19d372ee586bd8bc138aa0b43fa6ad15..66b3bc5b868839206f675b2c54416f89c1098a60)*

src/items/functions.md
tgross35, line 1: Looking through the list at https://doc.rust-lang.org/reference/items/functions.html#attributes-on-functions, also noticed that `#[cold]` may make no sense. Everything else seems fine

 https://github.com/rust-lang/rust/pull/158504#issuecomment-5229702188

tgross35 (changes requested):
Two things that I think should be resolved before merge

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/2ed1cb8b19d372ee586bd8bc138aa0b43fa6ad15..66b3bc5b868839206f675b2c54416f89c1098a60)*

src/items/functions.md
tgross35: I guess that generics are probably okay and allowed by default if not disallowed?

src/items/functions.md
folkertdev: It doesn't do harm either, so should we really make an exception here? I'm not against it per se, just wondering if we should make this calling convention even more special.

src/items/functions.md
```diff
@@ -239,6 +239,58 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
...
+> assert_eq!(x, 42);
+> # }
+> ```
+
+r[items.fn.extern.custom.signature]
+An `extern "custom"` function must:
+
+- Be `unsafe`.
+- Not have any parameters.
+- Return the [unit type], either by omitting a return type, or by explicitly returning `()`.
```
folkertdev: I'm not sure if we should actually use the word syntactic here. 

The main idea is that a type alias that would resolve to unit is not accepted. It really has to be a literal `()`.

src/items/functions.md
```diff
@@ -239,6 +239,58 @@ With `panic=unwind`, when a `panic` is turned into an abort by a non-unwinding A
...
+
+r[items.fn.extern.custom.naked]
+An `extern "custom"` function definition must be a [naked function].
+
+r[attributes.codegen.naked.inline]
+The [`cold` attribute] cannot be applied to an `extern "custom"` function.
```
folkertdev: added this in the same style of how attributes are rejected in combination with `#[naked]`.

folkertdev (commented):
I've added two updates based on the T-lang meeting ([notes](https://hackmd.io/nPO0wh-4QhSwVU-1f_QfJg#Stabilize-extern-custom-rust158504)). 

- we decided that the current implementation of only allowing unit as a return type syntactically (i.e. type aliases that resolve to unit are rejected) is acceptable.
- https://github.com/rust-lang/rust/pull/161002

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2300/6a5392a7d2e429766a9e128d3bebcd981514f29e..217edc5ab288e5a93a89ac657a8b55ebd2b2605c)*

src/items/functions.md
traviscross: Thanks.  I revised and used the word *syntactic* in a paired admonition.

src/items/functions.md
traviscross: Thanks.  I moved it over to the section on the `cold` attribute, since we usually pair these placement restrictions with the attribute text.

traviscross (approved):

traviscross:
Thanks @folkertdev (and @tgross35).

PR #2302, opened 2026-06-30
Do not promote `extern` statics

sjwang05:
Reference update for rust-lang/rust#157641: a borrow of an extern static cannot be promoted, even when the reference is never read from.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -394,7 +394,17 @@ static mut S: *const &u8 = [unsafe { &X }].as_ptr(); // ERROR.
 ```
 
 > [!NOTE]
-> Only expressions that cannot fail to evaluate can be promoted. The value of an `extern` static cannot be known at compile time, so any expression that uses an `extern` static is rejected by promotion, even if the code never reads the static's value. To borrow from an `extern` static in a const context, use a [const block] instead.
+> Only expressions that cannot fail to evaluate can be promoted. The value of an `extern` static cannot be known at compile time, so any expression that uses an `extern` static is rejected by promotion, even if the code never reads the static's value. To borrow from an `extern` static in a const context, use a [const block] instead:
+>
+> ```rust
+> unsafe extern "C" {
+>     static X: u8;
+> }
+>
+> // The array is evaluated as a constant, giving it a `'static`
+> // slot without promotion.
+> static mut S: *const &u8 = const { [unsafe { &X }] }.as_ptr();
+> ```
 
 r[destructors.scope.lifetime-extension]
 ### Temporary lifetime extension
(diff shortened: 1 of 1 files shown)


traviscross (approved):

src/destructors.md
```diff
@@ -378,6 +378,17 @@ r[destructors.scope.const-promotion]
...
+A borrow of an [`extern` static] cannot be promoted, even if the resulting reference is never read from.
+
+```rust,compile_fail
+unsafe extern "C" {
+    static X: i32;
+}
+
+static mut FOO: *const &i32 = [unsafe { &X }].as_ptr();
+```
+
```
traviscross, line 395: If the error produced has an error code, we'll want to add that to the infostring line.

Probably we also want to add some comments to this example to help make clear what's going on.  And maybe it's worth adding an admonition that mentions the theory behind why this is rejected (see, e.g., what I wrote in https://github.com/rust-lang/rust/pull/157641#issuecomment-4775681359).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2302/86635e30bf861a038dc197d7e16fd09e7e514e7a..9d66d5f5a1d24a960f8f597b9d13633b88a67bfe)*

sjwang05: Only a plain `error: encountered dangling pointer in final value of mutable static` gets shown, so there's no error code.

I added comments to the code example, along with an admonition that adds a little more detail.

src/destructors.md
```diff
@@ -378,6 +378,24 @@ r[destructors.scope.const-promotion]
...
+// The array is a temporary, so it would have to be promoted to a
+// `'static` slot in order to be used in a static initializer.
+// However, since it borrows an `extern` static, it is not eligible
+// for promotion.
+static mut FOO: *const &i32 = [unsafe { &X }].as_ptr(); // ERROR
+```
+
+> [!NOTE]
+> Only expressions that cannot fail to evaluate can be promoted. The value of an `extern` static cannot be known at compile time, so any expression that uses an `extern` static is rejected, even if it never reads the static's value. To borrow an `extern` static in a const context, use a [const block] instead.
```
DanielEScherzer: "use a const block instead" - maybe give an example of this specific for extern statics? That page doesn't have the word "extern" at all

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2302/bec6b5e6631be1e45ed905c9e058ffa7ecd9eb62..68746b96e90062e6f943e93dd051bfdcb5296646)*

sjwang05: Thanks, I added an example directly to the admonition

sjwang05:
Thanks @traviscross for the revisions

PR #2303, opened 2026-07-02
Fix the tagged-union pattern-matching recommendation

traviscross:
The union chapter recommends matching a tag and its corresponding union field together in a single pattern when implementing a C-style tagged union.  This suggests the idiom is safe.

But it's not.  We don't guarantee the order that subpatterns are checked -- this could only be safe if we could be sure the tag is checked first.  And, in fact, if an or-pattern is used when matching the tag, then `rustc` defers the check, causing actual UB (see rust-lang/rust#158387).

Let's fix that.  We can remove the `items.union.pattern.subpattern` rule entirely -- it's trivially true that a pattern can appear as a subpattern.  In its place, let's add a warning admonition that highlights what isn't guaranteed and describes what people should do instead.

cc @theemathas

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -124,35 +124,36 @@ fn f(u: MyUnion) {
 }
 ```
 
-r[items.union.pattern.subpattern]
-Pattern matching may match a union as a field of a larger structure. In particular, when using a Rust union to implement a C tagged union via FFI, this allows matching on the tag and the corresponding field simultaneously:
-
-```rust
-#[repr(u32)]
-enum Tag { I, F }
-
-#[repr(C)]
-union U {
-    i: i32,
-    f: f32,
-}
-
-#[repr(C)]
-struct Value {
-    tag: Tag,
-    u: U,
-}
-
-fn is_zero(v: Value) -> bool {
-    unsafe {
-        match v {
-            Value { tag: Tag::I, u: U { i: 0 } } => true,
-            Value { tag: Tag::F, u: U { f: num } } if num == 0.0 => true,
-            _ => false,
-        }
-    }
-}
-```
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


RalfJung:
LGTM, thanks!

ehuss (approved):
Thanks! We discussed how this change fits in procedurally. I felt like the existing rule was making a specific statement ("you can safely match these simultaneously"). This was copied directly from [RFC 1444](https://rust-lang.github.io/rfcs/1444-union.html). Is it a language change to remove that?

There seems to have been a series of things as I understand it:
- The RFC probably never should have specified this, as the order of pattern matching was not intended to be specified.
- The implementation phase never addressed this.
- The stabilization never highlighted or addressed this.
- The text was copied from the RFC without considering this ordering requirement.

I'm also not sure for how long everyone has understood that the order of pattern matching is not specified. Nested or-patterns came long after unions. Was the unspecified nature a well-understood decision going back to the beginning of the language, or was this something that evolved over time?

In the end, this seems like it is fixing an error. There is concern that code in the wild is using this pattern perhaps as a result of our documentation. My understanding is that in the simple case it does actually match in order, but that isn't something people want to guarantee (and it doesn't work with more complex patterns). Perhaps a lint as suggested in https://github.com/rust-lang/rust/issues/158387 is the answer. Or maybe the order should be guaranteed under strict circumstances.

Regardless, this seems like a safer step for now. I think the teams should consider what to do about this, as it seems like an easy footgun. If at some point it is decided to specify some specific ordering requirements, then we'll add those rules then.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2303/86635e30bf861a038dc197d7e16fd09e7e514e7a..4fd8e8151aace293c28d063f67d2b51e58851fc8)*

PR #2304, opened 2026-07-03
aliasing rules also apply inside private fields

RalfJung:
People might reasonably assume that a reference hidden inside private fields isn't really a reference. Let's make it clear that that's not the case.
This is based on feedback by @jethrogb.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -24,7 +24,9 @@ r[undefined.place-projection]
 r[undefined.alias]
 * Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is an outline of the general principles:
 
-  `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]), and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live. `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules. The exact liveness duration is not specified, but some bounds exist:
+  `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]), and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live (with no exceptions). `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules. These rules apply to *all* references and `Box<T>`, including those stored inside private fields (e.g., if your type has a private field of type `&mut`, that reference must be unique in the sense described above for as long as values of your type are live).
+
+  The exact liveness duration is not specified, but some bounds exist:
 
   * For references, the liveness duration is upper-bounded by the syntactic lifetime assigned by the borrow checker; it cannot be live any *longer* than that lifetime.
   * Each time a reference or box is dereferenced or reborrowed, it is considered live.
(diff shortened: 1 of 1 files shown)


src/behavior-considered-undefined.md
```diff
@@ -24,7 +24,9 @@ r[undefined.place-projection]
 r[undefined.alias]
 * Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is an outline of the general principles:
 
-  `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]), and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live. `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules. The exact liveness duration is not specified, but some bounds exist:
+  `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]), and `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live. `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules. These rules apply to *all* references and `Box<T>`, including those inside private fields.
```
RalfJung: FWIW all I did is add a single sentence, and then break the liveness part into a separate paragraph. But the newline style makes that very hard to tell from the diff.

EDIT: The "rich" diff makes it fairly clear though.
EDIT2: Now that I did a few more minor edits, the "rich" diff has also fallen apart. In one-sentence-per-line style, the diff would still be very easy to interpret.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2304/86635e30bf861a038dc197d7e16fd09e7e514e7a..960da81e017cc36f58fca51b07d158b8e73e8474)*

jethrogb: > including those inside private fields

Hmm this could also be read as "a private field that is a `&T`", while the intended reading (I believe) is "given &T, fields of T". But I don't know that this addition is particularly useful. I'd rather add a parenthetical for `&mut T`:

> (with no exception for [`UnsafeCell<U>`])

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2304/86635e30bf861a038dc197d7e16fd09e7e514e7a..960da81e017cc36f58fca51b07d158b8e73e8474)*

RalfJung: >  this could also be read as "a private field that is a &T"

Yes, that is what is meant.

> I'd rather add a parenthetical for &mut T:

If we list all the types that are *no* exception, we'll never be done.

Note that this edit is about your "I thought this was fine because the reference was never visible to the user". That's what I thought you said anyway.

RalfJung: We can add "no exceptions" without mentioning any particular type. But that's something we could add to *every single sentence* in the Reference. If people go about randomly assuming exceptions, there's very little we can do.

traviscross: Try [delta](https://github.com/dandavison/delta) (e.g., with `side-by-side = true`), or `git diff --word-diff`.  Not sure why GitHub's diff is weak.

traviscross (approved):

traviscross:
Thanks @RalfJung; thanks @jethrogb.

PR #2284, opened 2026-06-13
Document modules in attribute & derive macro input

TimNN:
To stabilize that part of the `proc_macro_hygiene` feature (https://github.com/rust-lang/rust/issues/54727).

Stabilization:

- https://github.com/rust-lang/rust/pull/157857

Tracking:

- https://github.com/rust-lang/rust/issues/54727

---

r? @petrochenkov for an initial review (the docs said to have the reference PR ready before the stabilization PR, I'll have that one up in a moment).

PR #2310, opened 2026-07-09
Add a rule for attribute macro outline modules

ehuss:
https://github.com/rust-lang/rust/pull/157857 stabilized the ability to use `#[my macro] mod foo;`. The reference never specified that this was not allowed, so it doesn't absolutely need to be changed for this. However, since this was something that the lang team considered deeply enough on its own, it seems worthy to be explicit about this behavior.

This is an alternative to https://github.com/rust-lang/reference/pull/2284. Since we already have a rule that specifies where the attributes are allowed, this follows more closely to that to specify the behavior when applied to one of these module items.

### DIFF
--- a/src/procedural-macros.md
+++ b/src/procedural-macros.md
@@ -317,6 +317,9 @@ Attribute macros can only be used on:
 - Inherent and trait [implementations]
 - [Trait definitions]
 
+r[macro.proc.attribute.outline-mod]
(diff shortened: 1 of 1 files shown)


traviscross (approved):

traviscross:
Makes sense to me.  Thanks @ehuss.

PR #2311, opened 2026-07-09
Document prohibition on inner attribute macros

ehuss:
Proc macro attributes as inner attributes are currently unstable via the `custom_inner_attributes` feature gate (https://github.com/rust-lang/rust/issues/54726).


traviscross (approved):

traviscross:
Thanks @ehuss.

PR #2312, opened 2026-07-13
Document `import_trait_associated_functions` feature

camsteffen:
[Tracking issue](https://github.com/rust-lang/rust/issues/134691)

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -96,12 +96,13 @@ The [paths] that are allowed in a `use` item follow the [SimplePath] grammar and
 * Nameable [items]
 * [Enum variants]
 * [Built-in types]
+* [Associated items] of traits
 * [Attributes]
 * [Derive macros]
 * [`macro_rules`]
 
 r[items.use.path.disallowed]
-They cannot import [associated items], [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
+They cannot import associated items of structs or enums, [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
 
 r[items.use.path.namespace]
 `use` will create bindings for all [namespaces] from the imported entities, with the exception that a `self` import will only import from the type namespace (as described below). For example, the following illustrates creating bindings for the same name in two namespaces:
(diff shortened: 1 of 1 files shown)


src/items/use-declarations.md
```diff
@@ -96,12 +96,13 @@ The [paths] that are allowed in a `use` item follow the [SimplePath] grammar and
...
 * [Enum variants]
 * [Built-in types]
+* [Associated items] of traits
 * [Attributes]
 * [Derive macros]
 * [`macro_rules`]
 
 r[items.use.path.disallowed]
-They cannot import [associated items], [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
+They cannot import associated items of structs or enums, [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
```
DanielEScherzer, line 105: also not from unions, right?

maybe instead 
```suggestion
They cannot import associated items (other than associated items of traits), [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
```

so that if some other thing with associated items is added later we don't need to update the list here again?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2312/afdc77bab886d4455c11247cdd32391bfab636ae..c91676579ea9275a5fc6b85edaed08a7f961866a)*

camsteffen: Good catch. I think we may as well list them out. But the sentence structure gets pretty unwieldy so we could split it into two sentences.

```suggestion
They cannot import associated items of structs, enums or unions. They cannot import [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
```

DanielEScherzer: What about

```suggestion
They cannot import associated items of non-traits, [generic parameters], [local variables], paths with [`Self`], or [tool attributes]. More restrictions are described below.
```

?

camsteffen: That works but I think it's more straightforward to just be explicit. I don't feel strongly though. Maybe we can get a third opinion.

PR #2314, opened 2026-07-19
Document attributes on closures and method calls

Unique-Usman:
Update the expression attribute documentation to cover attributes directly on closure and method-call expressions.

Document that an attribute before a method-call expression applies to the complete method-call chain.

This accompanies the rust-lang/rust stabilization of these expression attribute positions.

Stabilization PR: https://github.com/rust-lang/rust/pull/159581


Unique-Usman:
@estebank

PR #2317, opened 2026-07-27
S390X stabilize backchain

fneddy:
Document the s390x `backchain` target feature
Stabilized in https://github.com/rust-lang/rust/pull/158612

### DIFF
--- a/src/attributes/codegen.md
+++ b/src/attributes/codegen.md
@@ -715,6 +715,7 @@ Feature                                | Implicitly Enables                    |
 `miscellaneous-extensions-2`           |                                       | miscellaneous extensions 2
 `miscellaneous-extensions-3`           |                                       | miscellaneous extensions 3
 `miscellaneous-extensions-4`           |                                       | miscellaneous extensions 4
+`backchain`                            |                                       | store the caller's frame address in the callee's stack frame backchain slot
 
 [z/Architecture Principles of Operation]: https://publibfp.dhe.ibm.com/epubs/pdf/a227832d.pdf
 


src/attributes/codegen.md
```diff
@@ -715,6 +715,7 @@ Feature                                | Implicitly Enables                    |
 `miscellaneous-extensions-2`           |                                       | miscellaneous extensions 2
 `miscellaneous-extensions-3`           |                                       | miscellaneous extensions 3
 `miscellaneous-extensions-4`           |                                       | miscellaneous extensions 4
+`backchain`                            |                                       | store the caller's frame address in the callee's stack frame backchain slot
```
DanielEScherzer, line 718: a quick PDF search of the `[z/Architecture Principles of Operation]` PDF doesn't have any hits for "backchain" - is there some other documentation that might be worth linking to?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2317/afdc77bab886d4455c11247cdd32391bfab636ae..d61f918b969533b07c1b2da775a324256f4d09ff)*

fneddy: oh yes the PoP does only describe the ISA. the ABI is described here: https://ibm.github.io/s390x-abi/ and i think due to "historic reasons" its called "back chain"(with a space). Should I add a reference to the md?

DanielEScherzer: I would suggest yes (and maybe with a note that it has a space in that document)

PR #2321, opened 2026-08-08
document naked c-variadic functions

folkertdev:
A first draft. I'm not sure what the right way is to draw a distinction between function definition and naked function definition. 

Stabilization:

- https://github.com/rust-lang/rust/pull/159746


traviscross (approved):

traviscross:
Thanks @folkertdev.

PR #2322, opened 2026-08-09
aliasing rule summary: use a bulleted list

RalfJung:
@tgross35  suggested this may be easier to read.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -24,7 +24,11 @@ r[undefined.place-projection]
 r[undefined.alias]
 * Breaking the pointer aliasing rules. The exact aliasing rules are not determined yet, but here is an outline of the general principles:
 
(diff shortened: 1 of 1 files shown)


DanielEScherzer (approved):
looks like this is just rearranging things and not changing the semantics, looks good to me

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2322/82da570bafd9efed421bdbae0f8603ede8dd308b..1701957af94bd61cf448bc942de3159555bf9376)*

src/behavior-considered-undefined.md
```diff
@@ -24,7 +24,11 @@ r[undefined.place-projection]
...
+  * `&T` must point to memory that is not mutated while they are live (except for data inside an [`UnsafeCell<U>`]).
+  * `&mut T` must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live (with no exceptions).
+  * `Box<T>` is treated similar to `&'static mut T` for the purpose of these rules.
+
+  These rules apply to *all* references and `Box<T>`, including those stored inside private fields (e.g., if your type has a private field of type `&mut T`, that reference must be unique in the sense described above for as long as values of your type are live).
```
DanielEScherzer, line 31:
```suggestion
These rules apply to *all* references and `Box<T>`, including those stored inside private fields (e.g., if your type has a private field of type `&mut T`, that reference must be unique in the sense described above for as long as values of your type are live).
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2322/82da570bafd9efed421bdbae0f8603ede8dd308b..1701957af94bd61cf448bc942de3159555bf9376)*

DanielEScherzer: nevermind I misunderstood how this would be rendered

traviscross (approved):

PR #2325, opened 2026-08-12
Add grammar validation tool

ehuss:
This adds a tool that validates the grammar by testing it against other parsers such as rustc and proc-macro2.

There are two new crates:

- `parser` is an interpreter that takes a source input and generates a tree of nodes based on the Reference grammar.
- `grammar-check` is a CLI program for checking the grammar against other parsers. It includes several different subcommands and utilities.

## Incomplete

Various parts of this are incomplete. Lexing is generally working, but AST-based parsing needs some work (primarily with token splitting and support for comparisons). The permutation-based testing doesn't cover as much as I'd like and needs more work. There should be notes here and there throughout the source and READMEs to explain things in more detail. Generally this should be enough for validating the lexer.

## Daily workflow

This adds a daily GitHub Actions workflow to validate the tokenizer against rustc and proc-macro2. If it fails it should open an issue. This can fail if the API for rustc changes, or if there is a legitimate disconnect between the Reference and rustc. This is done as a scheduled job instead of a PR-check since breaking API changes can be very disruptive to contributors (who would see errors that are unrelated to their work). We could potentially add a PR-check using a hard-coded dated nightly release, but then we'd need some process for regularly updating to a new date.

It uses all of the `.rs` files in `rust-lang/rust` as its corpus as well as algorithmically generated inputs.

## Walkthrough

A general walkthrough for this PR (in order):

The first several commits are in preparation for the last main commit.

- `parser`
    - `parser/README.md` — Introduction to the parser.
    - `parser/src/lib.rs` — See that there are some basic definitions like a `Node`.
    - `parser/src/parser.rs` — This is the main parser that takes the Reference grammar and interprets some input. See that it has a `Source` abstraction that is used to handle byte inputs (for tokenization) or token inputs (for tree parsing).
    - `parser/src/lexer.rs` — The interface for tokenization. Notice that it handles all the things that are embedded in English rules (like CRLF normalization). It also handles shebang and frontmatter. There's also some various normalization that helps with working with the tokens.
    - `parser/src/tree.rs` — The interface for tree-based parsing using tokens as an input.
    - `parser/src/coverage.rs` — Handling for coverage analysis of the Reference grammar when validating it. It keeps track of coverage hits for each sub-expression, and emitting an HTML report. Coverage analysis is emitted with the `--coverage` CLI option to an HTML report.
- `grammar-check`
    - `grammar-check/README.md` — Introduction to the grammar-check CLI tool.
    - `grammar-check/src/main.rs` — The CLI definition (and a few misc helpers).
    - `grammar-check/src/test_cases.rs` — These are some hard-coded sample Rust inputs for validating the grammar. This is incomplete, see its comment for more.
    - `grammar-check/src/permute.rs` — The permutation-based Rust source generator. This attempts to exhaustively generate all possible inputs (including errors) for the grammar. This is incomplete, see its comment for more.
    - `grammar-check/src/tools/` — The different tools that this supports (proc-macro2, `rustc_parse` and `rustc_lexer`).
    - `grammar-check/src/commands/` — The various CLI sub-commands.
        - `grammar-check/src/commands/print_grammar.rs` — Very simple, emits the grammar to the output.
        - `grammar-check/src/commands/tokenize.rs` — Takes an input and tokenizes it and prints out the tokens in human-readable format. Can take various different kinds of inputs, and supports different tools.
        - `grammar-check/src/commands/tree.rs` — Same as `tokenize.rs`, except for emitting a tree for the AST grammar.
        - `grammar-check/src/commands/split_check.rs` — This was an experimental subcommand for finding token splits. Not sure if this will ever be useful, but it's what I used to find the token split points. Hopefully this demonstrates that `grammar-check` can potentially be used for all sorts of analysis (I created several other subcommands for that purpose, but I did not include them).
        - `grammar-check/src/commands/lex_compare.rs` — Compares tokenization between multiple tools. This is the key of this entire PR. It takes Rust input and tokenizes it with the Reference grammar and some other tool and compares the output. It runs in parallel because running millions of tests can be slow. Note that it needs to do some normalization to compare against different tools because not all tools have the same concept of what is a token or what is the exact span of a token.
- `tools/grammar/src/frontmatter.rs` — This is a temporary addition to support frontmatter.
- `.github/workflows/daily-grammar-check.yml` — The daily workflow.

I recommend just playing around with the CLI for a while to get a sense of using it. There are a bunch of commands and options. See the README for getting started, and then just look at the help pages and try some things out.


tools/grammar-check/src/commands/lex_compare.rs
```diff
@@ -0,0 +1,378 @@
...
+        let location = info
+            .location()
+            .map(|l| format!("{}:{}:{}", l.file(), l.line(), l.column()))
+            .unwrap_or_else(|| "unknown".to_string());
+        let thread = std::thread::current();
+        let name = thread.name().unwrap_or("<unnamed>");
+        let output = format!("thread '{name}' panicked at {location}:\n{msg}");
+
+        // We print it here as well to ensure it is seen if the thread dies unexpectedly.
+        // eprintln!("{}", output);
```
DanielEScherzer, line 65: Was this meant to be commented out here?

tools/grammar-check/src/commands/split_check.rs
```diff
@@ -0,0 +1,441 @@
...
+    let right_starts = get_possible_starts(grammar, right);
+
+    // Get descriptions of the left and right elements
+    let left_desc = describe_expression(left);
+    let right_desc = describe_expression(right);
+
+    // Check if any combination could form a multi-char token
+    for left_end in &left_endings {
+        for right_start in &right_starts {
+            let combined = format!("{}{}", left_end, right_start);
```
DanielEScherzer, line 195: formatting seems a bit excessive here to just add two strings together, maybe

```suggestion
            let combined = left_end.to_owned() + right_start;
```
?

tools/grammar-check/src/tools/rustc.rs
```diff
@@ -0,0 +1,267 @@
...
+    spans: Vec<DiagSpan>,
+}
+
+#[derive(serde::Deserialize)]
+struct DiagSpan {
+    is_primary: bool,
+    byte_start: u32,
+}
+
+pub fn normalize(tokens: &[Node]) -> Result<Vec<Node>, ParseError> {
```
DanielEScherzer, line 259: it doesn't look like this can ever return an error, why not just have it return a `Vec<Node>` directly?

tools/grammar-check/src/tools/rustc_lexer.rs
```diff
@@ -0,0 +1,30 @@
...
+//! macros, and those are not the same as the ones as generated by
+//! `rustc_lexer`.
+
+extern crate rustc_lexer;
+
+use parser::{Node, ParseError};
+use rustc_lexer::{FrontmatterAllowed, TokenKind};
+use std::ops::Range;
+
+pub fn tokenize(src: &str) -> Result<Vec<Node>, ParseError> {
```
DanielEScherzer, line 15: this also never returns an error and can just return `Vec<Node>`

tools/grammar-check/src/main.rs
```diff
@@ -0,0 +1,427 @@
...
+    channel: Sender<Message>,
+    use_spinner: bool,
+}
+
+impl CommonOptions {
+    fn new(
+        matches: &clap::ArgMatches,
+        default_tools: &[Tool],
+    ) -> (CommonOptions, Receiver<Message>) {
+        fn map_case(case: &String) -> Vec<(String, String)> {
```
DanielEScherzer, line 96: can I suggest moving these function definitions outside of the impl to avoid the deep nesting?

tools/grammar-check/src/main.rs
```diff
@@ -0,0 +1,427 @@
...
+                    .flat_map(|(name, cases)| {
+                        cases.iter().map(|c| (name.to_string(), c.to_string()))
+                    })
+                    .collect(),
+                case_pattern => {
+                    let cs: Vec<_> = test_cases::LEX_CASES
+                        .iter()
+                        .filter(|(name, _)| {
+                            let name_parts: Vec<_> = name.split("::").collect();
+                            let pattern_parts: Vec<_> = case_pattern.split("::").collect();
```
DanielEScherzer, line 109: it seems that pattern_parts can be created outside of the iteration rather than re-splitting the case_pattern every time though

tools/grammar-check/src/main.rs
```diff
@@ -0,0 +1,427 @@
...
+        let permute_iter = matches.get_one::<String>("permute").map(|permute_name| {
+            if permute_name == "three" {
+                return Mutex::new(Box::new(permute::ThreeIterator::new())
+                    as Box<dyn Iterator<Item = String> + Send>);
+            }
+            let mut diag = Diagnostics::new();
+            let grammar = grammar::load_grammar(&mut diag);
+
+            // Leak the grammar to get a 'static reference for the iterator
+            let grammar_ref: &'static grammar::Grammar = Box::leak(Box::new(grammar));
```
DanielEScherzer, line 190: is this ever going to be cleaned up, or is this just a memory leak?

tools/grammar-check/src/permute.rs
```diff
@@ -0,0 +1,831 @@
+//! Permutation-based tests.
+//!
+//! This attempts to generate exhaustive coverage of the grammar by using
+//! permutations of all of the possible inputs to the grammar. This includes
+//! both valid and invalid inputs (particularly those that are truncated).
+//!
+//! It generates representative inputs for some of the expressions. For
+//! example, a a repetition generates an output that includes 0, 1, or 2
```
DanielEScherzer:
```suggestion
//! example, a repetition generates an output that includes 0, 1, or 2
```

tools/grammar-check/src/permute.rs
```diff
@@ -0,0 +1,831 @@
...
+        current_index: usize,
+    },
+    Optional {
+        iterator: Box<PermutationIterator<'g>>,
+        emitted_empty: bool,
+    },
+    Repeat {
+        expr: &'g Expression,
+        include_empty: bool,
+        current_stage: usize, // 0 = empty (if include_empty), 1 = single, 2 = double
```
DanielEScherzer, line 77: current_stage is only ever compared to literals, so it doesn't need to match the type of some other value; suggest shrinking to u8 which is all that is needed


```suggestion
        current_stage: u8, // 0 = empty (if include_empty), 1 = single, 2 = double
```

tools/parser/src/parser.rs
```diff
@@ -0,0 +1,630 @@
...
+/// Returns `Ok(None)` if the expression does not match. Otherwise, it
+/// returns the [`Nodes`] that match, along with the new index pointing
+/// just after the matched nodes.
+///
+/// Note that some expressions match zero elements (like `e*` when `e` doesn't
+/// match), and those are treated as a successful match where `Nodes` is
+/// empty.
+///
+/// Returns `Err` if there is some kind of syntax error.
+#[instrument(level = "debug", skip(grammar, e, src, coverage), ret)]
```
DanielEScherzer, line 138: suggest matching the order in the function signature

```suggestion
#[instrument(level = "debug", skip(grammar, coverage, e, src), ret)]
```

DanielEScherzer (commented):
not going to pretend like I understand this, but some notes

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2325/6a5392a7d2e429766a9e128d3bebcd981514f29e..7a13fb65c041bbb9a9917f9ea3fb05e4f4ed3e8e)*

mattheww:
I tried this out by running `lex-compare` against lexeywan's list of testcases and that shows three sources of discrepancies with `rustc_parse`.

Using bash (so the `$'...'` strings below use C-like escapes):

```bash
cargo run -q -p grammar-check -r -- lex-compare --tool rustc_parse --edition 2024 --string $'/**\r CR starting block doc comment */'
cargo run -q -p grammar-check -r -- lex-compare --tool rustc_parse --edition 2024 --string $' "string\\\n\n\r\tcontinuation"'
cargo run -q -p grammar-check -r -- lex-compare --tool rustc_parse --edition 2024 --string $'#! /** doc */ [attr]\n'
```

I think the first is rust-lang/fls#610.

I think the second is rust-lang/fls#605.

It looks to me like in both those cases the Reference text is basically right but the lexer rules are wrong.

mattheww:
I also tried bodging lexeywan's proptest-based randomised testing into grammar-check, and it's found one more discrepancy:

```bash
cargo run -q -p grammar-check -r -- lex-compare --tool rustc_parse --edition 2024 --string '#!/*'
```

tools/grammar-check/src/permute.rs
traviscross: Thanks.  Fixed.

traviscross (approved):

traviscross:
Thanks @ehuss.

traviscross:
And thanks @mattheww for flagging the interesting divergences here.  @ehuss and I are working through those separately.

ehuss:
Thanks @DanielEScherzer for the reviews. For some context, TC asked for me to put this up "as-is". This is something that I've been working on since last year, but never really "finished". I probably won't be working on it much moving forward. He wanted to at least get it up to have some testing coverage.

And thanks @mattheww for taking a look! I opened https://github.com/rust-lang/reference/pull/2330 to fix the first ones, but the last one you pointed out is quite a bit more difficult to fix. 

In the past TC and I had talked about using a different approach of using code generation instead of an interpreter to be able to do fuzzing more easily. That might ultimately be better, but I never got around to trying that.

traviscross:
We've now fixed the `#!/*` case as well in https://github.com/rust-lang/reference/pull/2331.

PR #2327, opened 2026-08-15
items.use.restrictions.extern-prelude: be more clear example succeeds only in 2015 edition

tshepang:
Seeing `Ok` in example made me expect the example to succeed in later editions

### DIFF
--- a/src/items/use-declarations.md
+++ b/src/items/use-declarations.md
@@ -448,7 +448,7 @@ r[items.use.restrictions.extern-prelude]
 > ```rust,edition2015
 > use ::{self as root}; //~ OK in the 2015 edition.
 > ```
-> 
+>
 > Starting with the 2018 edition, the `::` prefix refers to the extern prelude, which cannot be directly imported.
 
 r[items.use.restrictions.duplicate-name]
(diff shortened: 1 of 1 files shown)


src/items/use-declarations.md
```diff
@@ -443,11 +443,13 @@ r[items.use.restrictions.extern-prelude]
 > ```
 
 > [!EDITION-2018]
-> In the 2015 edition, the prefix `::` refers to the crate root, so `use ::{self as root};` is allowed because it is same as `use crate::{self as root};`. Starting with the 2018 edition the `::` prefix refers to the extern prelude, which cannot be directly imported.
+> In the 2015 edition, the prefix `::` refers to the crate root, so `use ::{self as root};` is allowed because it is same as `use crate::{self as root};`.
 >
 > ```rust,edition2015
-> use ::{self as root}; //~ Ok
+> use ::{self as root}; //~ Ok in 2015 edition
```
traviscross:
```suggestion
> use ::{self as root}; //~ OK in the 2015 edition.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2327/6a5392a7d2e429766a9e128d3bebcd981514f29e..7b937b9d6d3b1ddde53bb37b7748d7e844c15156)*

PR #2328, opened 2026-08-16
Bump railroad-dependency; themed syntax diagrams

lukaslueg:
This bumps the `railroad`-dependency from 0.3.7 to 0.3.9, which fixes some minor layout issues. It also introduces support for the themes used by `mdBook`, instead of relying on just two global `light` / `dark` themes.

Some global CSS rules were (re-) moved to support multiple themes. The `railroad`-specific stylesheets are no longer put into the page's content if that page doesn't include a diagram, shaving some bytes off. The test was added to ensure that the stylesheets' selectors (which are defined upstream) don't ever leak outside the `svg.railroad`-namespace; this wasn't a problem before, just for future-compatibility.

I built and visually inspected the reference locally, and it seems fine to me.

Please notice that I couldn't exactly figure out the licensing status of the `Rust`, `Coal`, `Navy` and `Ayu` themes; out of an abundance of caution, the stylesheets include [a reference to the original definitions](https://github.com/lukaslueg/railroad/blob/master/src/stylesheet_rust.css) from which these themes were derived, which is included in the final CSS. De facto the palettes seem to be ubiquitous around the web.

Please notice that an LLM was used to assist in these changes.

## Light
light

## Rust
rust

## Coal
coal

## Navy
navy

## Ayu
ayu

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,9 +263,9 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "railroad"
-version = "0.3.7"
+version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "813b53dbe6d583f1ac0c3fb394e0ce3b613530adff8265c274fcdfcd82cc6da5"
+checksum = "7286d8b4d9fc00078819e06ff24cefa276f39461ff53b74d63d24736be27192f"
 dependencies = [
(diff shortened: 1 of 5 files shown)


ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2328/9df7eb8878ce5cc82d4da172e2b39e9d8d4f7f65..aa11837e57a6b5f87df5ba6a6ce4d60031f95ac3)*

PR #2331, opened 2026-08-18
Tolerate malformed comments in shebang lookahead

traviscross:
Our `SHEBANG` rule did not accept `#!/*`, even though that's a valid shebang per `rustc`.  The trouble is the interaction between the negative lookahead and the cut within `BLOCK_COMMENT`: the cut forces an immediate error even within the context of a negative lookahead. We rely on that semantic elsewhere.

To fix this, let's add a `SHEBANG_BLOCK_COMMENT` rule that accepts what `rustc` does.

cc @ehuss

### DIFF
--- a/src/shebang.md
+++ b/src/shebang.md
@@ -17,18 +17,31 @@ A *[shebang]* is an optional line that is typically used in Unix-like systems to
 r[shebang.syntax]
 ```grammar,lexer
 @root SHEBANG ->
-    `#!` !((WHITESPACE | LINE_COMMENT | BLOCK_COMMENT)* `[`)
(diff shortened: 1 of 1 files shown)


traviscross:
cc @mattheww

ehuss (approved):

traviscross:
cc https://github.com/rust-lang/reference/pull/2325

traviscross:
Thanks again, @mattheww, for flagging the divergences here.

mattheww:
I think given this change, commit 5b328bcf7e5e56c5cd3c660655348c2d85dea5cf from #2330 is unnecessary.

PR #2336, opened 2026-08-23
dangling pointers: turn some consequences of the definition into notes

RalfJung:
It seems to me that we have unnecessarily many normative clauses here. Most of this follows from the definitions of "points to" and allocations. Allocations are currently not defined in the Reference so I added a link to the standard library where we have a definition.

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -198,7 +198,7 @@ r[undefined.validity.undef]
 [`bool`]: types/boolean.md
 [`const`]: items/constant-items.md
 [abi]: items/external-blocks.md#abi
-[allocation]: ../std/ptr/index.html#allocation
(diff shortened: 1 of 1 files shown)


traviscross (approved):

traviscross:
Thanks @RalfJung.

src/behavior-considered-undefined.md
```diff
@@ -198,7 +198,7 @@ r[undefined.validity.undef]
 [`bool`]: types/boolean.md
 [`const`]: items/constant-items.md
 [abi]: items/external-blocks.md#abi
-[allocation]: ../std/ptr/index.html#allocation
+[allocation]: core::ptr#allocation
```
RalfJung, line 201: Oh I had no idea these work here, nice :)

PR #2337, opened 2026-08-23
nail down the final validity rules: references and unions

### LINKED ISSUE #413: Can references to uninhabited types ever be valid?
This discussion presupposes that "Do we have full recursive validity for references" (https://github.com/rust-lang/unsafe-code-guidelines/issues/412) is answered "no".

Is the following UB?
```rust
let x: &! = transmute(&());
```
The reason we could consider making this UB is that all of the arguments in #412 do not apply: to determine whether an `&!` is valid, we do *not* have to check memory, do any potentially racy reads, or anything like that. We can just answer the question with "no".

Basically, we can think of `!` having an impossible-to-satisfy alignment requirement, and therefore `&!` is invalid since it is never properly aligned. (I don't really want to fold this into the alignment check, but the point is that we already have the validity of `&T` depend on `T`, namely via alignment -- I see no fundamental reason why we couldn't do the same with inhabitedness.)

### LINKED ISSUE #414: Should validity of a reference depend on the *contents* of memory in any way?
This is one of the successors to https://github.com/rust-lang/unsafe-code-guidelines/issues/77. The question is: for a reference to be valid, what (if anything) do we require about the data it points to in memory?

Note that the Rust reference currently answers this question with "things must be valid fully recursively", but in my view this is mostly because we haven't yet figured out what exactly the weaker requirement is that we actually want to impose.
This maximalist position of requiring full recursive validity is almost certainly untenable:
- Full recursive validity is super fragile: to reason about why the reference you are creating does not cause UB, you have to reason about all code that might mutate any memory transitively reachable (via references).
- The benefit is questionable: having deeply nested references cause UB is very unlikely to ever benefit compiler optimizations.
- Allowing mutable references to uninit data (between consenting APIs) is (unfortunately) how the `io::Read` trait often works, and while better solutions are being developed, there's a long tail of copies of this API and a lot of old code to port, that we should have good reasons to consider UB.
- For shared references with interior mutability specifically, even determining whether an `&Mutex<bool>` is valid would cause a conceptual data race with another thread holding the lock and mutating that `bool`.
- Also see [this post](https://github.com/rust-lang/unsafe-code-guidelines/issues/77#issuecomment-519997799).
- There is [safe code](https://github.com/rust-lang/unsafe-code-guidelines/issues/412#issuecomment-2379009718) violating this principle.

However, many weaker positions exist. See https://github.com/rust-lang/unsafe-code-guidelines/issues/346 for a broader list of arguments for why validity should not depend on the contents of memory in any way.

### LINKED ISSUE #438: What are the values of a union type? (in particular, what is the validity invariant of a union)
When loading data at union type from one place and storing it in another -- what exactly is being required about the bytes, and what is preserved? In other words, what is the representation relation of a union type, and what is even the set of [values](https://github.com/RalfJung/minirust/blob/master/spec/lang/values.md) that can be generated by a union-typed load?

Originally I intended the answer to be "the values are just lists of raw bytes as large as the union". IOW, a union has no validity requirements whatsoever, and all data inside it is perfectly preserved. However, [reality disagrees](https://github.com/rust-lang/unsafe-code-guidelines/issues/156). With the C ABI, some of the bytes in a union value do *not* always get preserved -- there can be padding. Also there is some desire to make unions have validity invariants, so that e.g. a `union { usize, *const T }` is `noundef`.

On the other hand, it is certainly *not* the case that the value of a union is "a value of one of its fields". Apart from the problem about writing `decode` for such a specification, it can be [violated in safe code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=81617567297c690076415104e593eb31).

Note that for the purpose of this, we do assume that the LLVM type `iN` will perfectly preserve N bytes of memory. LLVM code generated from Rust can never put poison in memory (as that would not be preserved on a per-byte level, the entire `iN` gets "infected" if any byte is poison), which is currently accurate (Rust code has UB for any situation where the generated LLVM code might produce poison). LLVM `iN` values carrying provenance is [incoherent](https://www.ralfj.de/blog/2020/12/14/provenance.html) but mostly works in practice and LLVM does not offer us a better option. We need a "byte" type to properly express our intent to LLVM; until that exist, we make do with what we have. If you want to discuss LLVM issues, please open a new thread -- this one here is solely about the Rust AM semantics.

### Possible designs

MiniRust decides that a value of union type is a list of "chunks" where the bytes are perfectly preserved, but there can be gaps between those chunks where data is lost of a copy. The extent of these chunks can be [computed from a Rust union type layout](https://github.com/RalfJung/minirust/blob/master/tooling/minimize/src/chunks.rs).  That is sufficient to model the reality of union padding. The effective validity invariant here is still "any list of bytes allowed".

The `noundef` property could be achieved if we refine this "chunk" idea a bit further: every byte in a union is either "discarded", "preserved", or "initialized". Bytes which are padding in all variants get "discarded"; bytes which are padding only in *some* variants get "preserved"; bytes which are "in the data part" for *all* variants get "initialized". A value of union type is a list of bytes the size of the union with the constraint that "discarded" bytes must be `Uninit` and "initialized" bytes must *not* be `Uninit`. The validity invariant requires that all "initialized" bytes must be initialized -- and that's it; "discarded" bytes get flushed to `Uninit` on a `decode`. Then the Rust type `union { usize, *const T }` would translate to a union type where all bytes are "initialized", and we could emit `noundef` attributes for such a type.
(I don't want multiple distinct AM values to be equivalent, hence the 'value of union type' predicate demands that all "uninit" bytes are truly `Uninit`, but of course the validity invariant allows those bytes to be anything. Note that the validity invariant is a predicate over "list of bytes", whereas "what are the values of this type" is a predicate over `Value`. The chunks idea achieves this by just not having the padding bytes in the `Value`; we could do the same here but that would be formally more awkward to express I think. Maybe I should just link to a MiniRust patch here, that might be more clear than trying to say this in English.^^)

We could in principle go further and tro to achieve something along the lines of "if a byte is non-null in each variant, it is also non-null in the union", but that becomes increasingly complicated -- in particular it requires the "generate union type description from Rust type" logic to encode more and more knowledge about validity invariants, when the entire intent of this "value representation" business was that the validity invariant is [implicitly encoded by the representation relation](https://github.com/RalfJung/minirust/blob/master/spec/lang/representation.md#relation-to-validity-invariant).

What other designs should we consider? And are they worth the complexity?

RalfJung:
This translates the following t-opsem FCPs into Reference text:
- https://github.com/rust-lang/unsafe-code-guidelines/issues/414#issuecomment-4090976382
- https://github.com/rust-lang/unsafe-code-guidelines/issues/413#issuecomment-4529063212
- https://github.com/rust-lang/unsafe-code-guidelines/issues/438#issuecomment-4529070534

Cc @rust-lang/opsem 

@rust-lang/lang you have not FCP'd this yet; please let us know if you want this to go through your own decision process as well.

Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/413
Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/414
Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/438

### DIFF
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -145,10 +145,13 @@ r[undefined.validity.struct]
 * A `struct`, tuple, and array requires all fields/elements to be valid at their respective type.
 
 r[undefined.validity.union]
-* For a `union`, the exact validity requirements are not decided yet. Obviously, all values that can be created entirely in safe code are valid. If the union has a [zero-sized] field, then every possible value is valid. Further details are [still being debated](https://github.com/rust-lang/unsafe-code-guidelines/issues/438).
+* For a `union`, there are no validity requirements. All byte sequences are valid union values.
 
 r[undefined.validity.reference-box]
-* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the [metadata]). Note that the last point (about pointing to a valid value) remains a subject of some debate.
+* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and the pointee type `T` must be *inhabited*.
+
+  The exact classification of inhabited types is unspecified, similar to the size and alignment of types.
+  However, types that can be constructed from safe code are definitely inhabited.
 
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:
@@ -195,8 +198,11 @@ r[undefined.validity.const-provenance]
   > };
   > ```
 
-r[undefined.validity.undef]
-**Note:** Uninitialized memory is also implicitly invalid for any type that has a restricted set of valid values. In other words, the only cases in which reading uninitialized memory is permitted are inside `union`s and in "padding" (the gaps between the fields of a type).
+> [!NOTE]
+> The definition above implies that uninitialized memory is invalid everywhere except inside `union`s and in "padding" (the gaps between the fields of a type).
+
+> [!WARNING]
+> Just because a value is *valid* does not mean that it is *safe to use*. A value being *valid* merely means that creating the value does not cause immediate undefined behavior; such UB can still be caused later, even by safe operations. For instance, consider that `&str` pointing to initialized non-UTF8 data is *valid* as defined above, but passing such a value to a safe function can cause undefined behavior. As a more extreme example, consider that `&[u8]` pointing to allocated but uninitialized memory is *valid*, but one can trivially cause UB simply by accessing an element of the slice. Generally, only values you can construct in safe code are *safe to use*.
 
 [`bool`]: types/boolean.md
 [`const`]: items/constant-items.md
--- a/src/behavior-considered-undefined.md
+++ b/src/behavior-considered-undefined.md
@@ -152,6 +152,8 @@ r[undefined.validity.reference-box]
 
   The exact classification of inhabited types is unspecified, similar to the size and alignment of types.
   However, types that can be constructed from safe code are definitely inhabited.
+  For unsized types, this check considers dynamic information from the metadata:
+  In particular, if `T` has an unsized tail of slice type `[U]`, and if `U` is uninhabited, and if the length encoded in the metadata is non-zero, then the pointee is uninhabited.
 
 r[undefined.validity.wide]
 * The [metadata] of a wide reference, [`Box<T>`], or raw pointer must match the type of the [unsized tail]:


RalfJung:
A proposal came up in https://github.com/rust-lang/unsafe-code-guidelines/issues/620 to slightly strengthen the requirements for references / `Box` to types with a slice tail.

@scottmcm do you have suggestions for how that should be worded?

src/behavior-considered-undefined.md
```diff
@@ -145,10 +145,15 @@ r[undefined.validity.struct]
...
+* For a `union`, there are no validity requirements. All byte sequences are valid union values.
 
 r[undefined.validity.reference-box]
-* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the [metadata]). Note that the last point (about pointing to a valid value) remains a subject of some debate.
+* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and the pointee type `T` must be *inhabited*.
+
+  The exact classification of inhabited types is unspecified, similar to the size and alignment of types.
+  However, types that can be constructed from safe code are definitely inhabited.
+  For unsized types, this check considers dynamic information from the metadata:
+  If `T` has an unsized tail of slice type (`[U]`), it is inhabited of the length encoded in the metadata is 0 or if `U` is inhabited.
```
RalfJung: These two lines are for https://github.com/rust-lang/unsafe-code-guidelines/issues/620. We didn't explicitly discuss this case in the opsem FCP.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2337/3b38834b39f732c64686f7c64aa29dcf3cd83ba5..5dcbf411c8308f28017ed3dfa2f97296cf61aa72)*

scottmcm:
> do you have suggestions for how that should be worded?

I don't have anything concrete.

At least as a mental model -- whether it can be phrased well for a reference I'm less sure -- I wonder if there's a general rule here like saying that the type in the reference must, including looking at the metadata, represent a well-formed and inhabited type.  (Part of me wants to say it comes from an unsizing, but there's nothing to unsize to a `struct Foo(u32, [f64]);` kind of thing, so I don't think that works, but maybe someone can come up with a cleverer phrasing.)

Basically I wish the "length has to be zero for uninhabited elements" and "length bound for non-ZST elements is `usize::MAX / sizeof(T)`" kinds of things were just corollaries of the general rule, to avoid things like the https://github.com/rust-lang/unsafe-code-guidelines/issues/620#issuecomment-5390173045 concern.

RalfJung:
So what do you think about the wording I put in this PR?

traviscross:
Thanks @RalfJung.  We talked about this in the lang meeting and agreed we'd expect a lang FCP here if we're changing the language.

@rfcbot fcp merge lang

(Eliding opsem here as it's already been FCPed by the opsem team.)

joshtriplett:
I reviewed the proposed rules both for unions and for references, and this looks reasonable to me.

@rfcbot reviewed

scottmcm:
This makes me want a separate language construct for "no, I'm not doing partial overwrite silliness" because this rule has meant that unions are not usable for the cases I wanted, but if this is the right answer for FFI unions and such then so be it.

@rfcbot reviewed

RalfJung:
> (Eliding opsem here as it's already been FCPed by the opsem team.)

FWIW the unsized type part hasn't been FCP'd by us yet, [as mentioned above](https://github.com/rust-lang/reference/pull/2337#discussion_r3842179202). We can do that in a separate issue or a joint FCP here.

traviscross:
@rfcbot fcp cancel

traviscross:
Thanks @RalfJung.

@rfcbot fcp merge lang,opsem

(In addition to what was previously FCPed by opsem, opsem is signing off here on what RalfJ mentions in https://github.com/rust-lang/reference/pull/2337#discussion_r3842179202.)

RalfJung:
@rfcbot reviewed

RalfJung:
FWIW @saethlin has ticked their box in https://github.com/rust-lang/unsafe-code-guidelines/issues/620, that should be carried over.

saethlin:
@rfcbot reviewed

src/behavior-considered-undefined.md
```diff
@@ -145,10 +145,15 @@ r[undefined.validity.struct]
...
+* For a `union`, there are no validity requirements. All byte sequences are valid union values.
 
 r[undefined.validity.reference-box]
-* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and it must point to a valid value (in case of dynamically sized types, using the actual dynamic type of the pointee as determined by the [metadata]). Note that the last point (about pointing to a valid value) remains a subject of some debate.
+* A reference or [`Box<T>`] must be aligned and non-null, it cannot be [dangling], and the pointee type `T` must be *inhabited*.
+
+  The exact classification of inhabited types is unspecified, similar to the size and alignment of types.
+  However, types that can be constructed from safe code are definitely inhabited.
+  For unsized types, this check considers dynamic information from the metadata:
+  In particular, if `T` has an unsized tail of slice type `[U]`, and if `U` is uninhabited, and if the length encoded in the metadata is non-zero, then the pointee is uninhabited.
```
WaffleLapkin, line 156: I feel like making inhabitedness depend on metadata values is not right / clean.

I would rather make this a separate restriction on the value of the metadata. That is, I would say something along the lines of "A reference or `Box<T>` must ... have valid metadata" and "slice metadata for a slice type with uninhabited elements, it must be 0".

This allows adding other restrictions in the future, such as the one mentioned by @scottmcm that slice metadata can at most be `usize::MAX / size_of::<T>()` (for non-ZSTs).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2337/3b38834b39f732c64686f7c64aa29dcf3cd83ba5..c8ce01e71a71c7a64ea3c6918e652a69d6d72d1d)*

RalfJung: We generally pretty consistently treat the metadata as telling us the "actual type" that we use to inspect the pointee. I think it would be odd do do something else here. It would look like an arbitrary set of rules, rather than just something that falls out of a more general principle.

> This allows adding other restrictions in the future, such as the one mentioned by @scottmcm that slice metadata can at most be usize::MAX / size_of::<T>() (for non-ZSTs).

As I mentioned in reply to Scott, that restriction is already present. It follows from the fact that the reference has to be dereferenceable for its actual dynamic size.

WaffleLapkin: Hm. It still feels weird to me to do it this way for slices, but if we are consistently doing it this way I suppose that is better.

I suppose the reason it feels weird to me is that in the compiler "inhabitedness" is a property of types, and does not concern the value in any way.  If we would like exploit this UB in the compiler, we _would_ have to add an arbitrary rule in the form of "if the tail is a slice of an uninhabited type, `assume(len == 0)`".

In a way, considering the "actual type" feels _more_ arbitrary to me.

That being said, I imagine we have to consider the "actual type" for trait objects, at which point it's better do the same thing for slices too.

(I'd resolve this conversation, but it appears I don't have permissions for that)

RalfJung: The compiler can only statically approximate all these properties using the types, yes. But that's expected, everything on this page is written assuming full knowledge of the current *dynamic* state (think: Miri, MiniRust). The compiler might get better at approximating things and that should not change the spec!

IOW, I think you're thinking too much like a compiler writer and not enough like a language specifier. :)

WaffleLapkin: guilty as charged I'm afraid ^^'

RalfJung: > If we would like exploit this UB in the compiler, we would have to add an arbitrary rule in the form of "if the tail is a slice of an uninhabited type, assume(len == 0)".

To also reply to this specifically -- the compiler does not have to exploit this UB if we think it's too weird or doesn't fit the current compiler architecture. But I would like to keep the door open to the compiler potentially exploiting this UB in the future, as the compiler architecture may change and make this less weird. Maybe one day it becomes trivial to make it so that the length metadata of `&[!]` is annotated with range information saying it has to be 0. This seems quite plausible, don't we already encode that the length is at most `isize::MAX` for non-zero-sized types?

asquared31415: in many contexts we do encode that the length is `0..=floor(isize::MAX/size_of::<pointee>)`, in particular function arguments. in a few context we do not (due to perf regressions and dubious benefit), and we never do for unsized tails, it's only `&[T]`.

for example, `fn meow(x: &[u16])` is `i64 noundef range(i64 0, 4611686018427387904)` for the length arg, but `struct Meow { val: [u16] }` `fn meow(x: &Meow)` is a plain `i64 noundef`.

WaffleLapkin: @RalfJung I was not arguing against adding arbitrary rules to the compiler (it's full of them!), I was more so arguing that it should be fine to have them here too (which I don't believe anymore); thanks for explaining your view though :)

~~@asquared31415 do you know _why_ we are only adding range metadata for `&[T]` and not `&Meow`? Seems like an oversight...~~ Turns out this metadata is calculated with a special case in cg_llvm: https://github.com/rust-lang/rust/blob/c4c4a576936e9e67717d0deb8e74e02dd5dd10de/compiler/rustc_codegen_llvm/src/abi.rs#L561-L579 As the comment suggests, I don't think this is a good way to handle this, hm...

asquared31415: I looked into it as a part of another PR, and it ended up regressing codegen in most cases, with minimal benefits: https://github.com/rust-lang/rust/pull/159921#issuecomment-5097037972 unfortunately I don't have numbers on hand, but llvm was making some **bad** assembly involving 64 bit immediates and sometimes even using a whole extra register (in a contrived test, but I'm unsure if it would go away even after inlining in normal usage)

The attributes are set [here](https://github.com/rust-lang/rust/blob/c4c4a576936e9e67717d0deb8e74e02dd5dd10de/compiler/rustc_codegen_llvm/src/abi.rs#L561-L578) in rustc_codegen_llvm, where it matches only `Ref(Slice(T))` types, and there's even a note that it's suboptimal.

During the aforementioned PR, I even hit the layout cycles that the comment alludes to, so it's definitely still an issue as of july-ish.

asquared31415: @scottmcm might have some additional context as the original author of that hack.  They mentioned https://github.com/rust-lang/rust/pull/152843 could help get *some* of the layout info we would like on DSTs, but IIRC I also touched that code in similar ways (with nuw and/or nsw on that math) and ended up getting similar "llvm suddenly emits 64 bit literals and extra registers".

Jules-Bertholet:
In https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/No.20reference.20validity.20vs.20improvements.20to.20closure.20captures/with/624981987, I describe a potential improvement to closure capture inference which *may* conflict with not having memory-dependent reference validity. (I'm hopeful that we can resolve the conflict.)

RalfJung:
@CAD97  @digama0 @joshlf friendly checkbox reminder -- from the opsem side the only part we didn't already FCP is the part about unsized types, where we are conservatively saying that we are evaluating inhabitedness of the pointee type using the information about the "actual" type as represented by the reference metadata.

CAD97:
@rfcbot reviewed

src/behavior-considered-undefined.md
digama0: @RalfJung  I agree that this is not stated correctly. It's not about uninhabitedness, this is a validity clause. Inhabitedness is a property of a type, validity is a property of a value of the type.

```suggestion
  For unsized types, this check considers dynamic information from the metadata:
  In particular, if `T` has an unsized tail of slice type `[U]`, and if `U` is uninhabited, then the only valid value of the length encoded in the metadata is zero.
```

RalfJung: I think it is part of inhabitedness. I view inhabitedness as a property of a potentially unsized type and metadata. We use the metadata to figure out what the actual dynamic type is, and then evaluate inhabitedness of that.

That's what it's going to look like in Miri and MiniRust, and IMO it's the structure we should also follow here. I thought it was okay to leave that implicit, but I see that it is too confusing, so I will try to reword.

digama0: No, that much I agree with; the part I disagree with is that the consequence is about inhabitedness (of a type). For a type like `&dyn NeverImplements`, obviously it's going to be inhabited since any types could be implementing `NeverImplements`. What we want to say is that the value `[ptr, never_vtable]: &dyn NeverImplements` is *invalid*, for the reason that the vtable is for a type that is *uninhabited*. For the unsized type `dyn NeverImplements` itself, that's a whole separate question, I'm not really sure what validity/inhabitedness means for unsized types since they don't exist on their own but the natural reading is that it is still inhabited, and it has "values" only after fixing the metadata, and there are no valid values which have `never_vtable` as the metadata.

The reasoning for the present case is similar: if `U` is *uninhabited*, then `[U; 1]` is *uninhabited* and `&[U; 1]` is *uninhabited* and `&[U]` is *inhabited* but has only the *valid* value `len = 0`; the value `[ptr, 1]: &[U]` is not valid for the type regardless of `ptr`. If you would like to say that the type `[U]` is "uninhabited at len=1" then that's fine but a new concept we haven't discussed before, and also apparently the same thing as just saying `[U; 1]` is uninhabited.

RalfJung: What is being defined here is whether `dyn NeverImplements` is inhabited. Which depends on the metadata, just like inhabitedness of `[!]`.

RalfJung: > If you would like to say that the type [U] is "uninhabited at len=1" then that's fine but a new concept we haven't discussed before, 

That's exactly what I mean and what I thought we had discussed: for unsized types, inhabitedness depends on the metadata. Slices are inhabited iff the length is 0 or the element type is inhabited. `dyn Trait` is inhabited iff the actual dynamic type represented in the metadata is inhabited.

RalfJung: I have reworded this part to hopefully make it more clear.

PR #2340, opened 2026-08-25
Specify the values of safely-constructed unions

djkoloski:
This specifies that the operations for safely constructing and modifying union values are more limited than those for unsafely doing so. It adds the guarantee that safe contexts cannot create certain kinds of union values which are invalid at every field.

## Summary

This formalizes some very narrow guarantees about union initialization by updating the reference. My motivation is re-enabling some trait derives in zerocopy related to transmuting bytes to and from union types. I did some background research on the opsem stance towards union validity, and read through several discussions about union safety semantics here. Following the lang team change process flowchart, I am opening this PR and nominating it for lang team discussion. Given that there has been a lot of discussion about related topics in the past, I'd like to summarize some background information.

## Links

1. [Zerocopy: "Restrict support for #[derive(IntoBytes)] on unions, work to guarantee forwards-compatible soundness"](https://github.com/rust-lang/unsafe-code-guidelines/issues/533)
2. [UCG: "Minimum guarantees for union construction and typed copies?"](https://github.com/rust-lang/unsafe-code-guidelines/issues/533)
3. [t-lang: "Union field semantics and safety"](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/union.20field.20semantics.20and.20safety/near/248632442)
4. [UCG: "What is the safety invariant, if any, for unions?"](https://github.com/rust-lang/unsafe-code-guidelines/issues/352)

## Motivation

Right now, safe Rust allows two operations on unions:
1. A value of a union type can be created with the value of a single field (i.e. `let u = MyUnion { f1: 1 }`)
2. Fields and subfields of a union value can be written (i.e. `u.f1 = 2`)

For some unions, this forms a guarantee about whether safe Rust can produce values of that union type with uninitialized bytes. Under these conditions:

- The union is `#[repr(C)]`
- The fields of the union are the same size (i.e. no trailing padding in the union)
- The field types of the union have no uninitialized bytes (e.g. padding)

... safe Rust cannot currently produce a value of the union that has any uninitialized bytes. One of the fields must be initialized to start (thus initializing all bytes of the union), and no subsequent subfield writes can de-initialize any bytes.

This set of guarantees allows us to [safely transmute](https://github.com/google/zerocopy/issues/1792) union values that meet these conditions. We'd like to be confident that the code we write that relies on this reasoning remains sound in the future as well.

### Is this a validity / safety requirement?

**No.** Per rust-lang/reference#2337, unions have no validity requirements. Safety requirements apply to all values of a type that are outside of an unsafe encapsulation boundary (e.g. returned from a public API). This change only applies to values that are created and modified exclusively within a safe context. It is 100% OK for those values to pass through an unsafe context and stop following this rule.

### Should this be a safety requirement?

There is no consensus as to whether unions should have safety requirements. This change avoids adding a safety requirement while also leaving the door open to safety requirements in the future.

### Why do we need to specify this if it's already true?

Even though it's true right now, we need this to remain true going forward.

### Why/how might we break this rule?

The only situation I could find was allowing wholly-uninitialized unions, like:

```rust
union Foo {
    a: u32,
}

fn main() {
    let x: Foo = unsafe { core::mem::MaybeUninit::<Foo>::uninit().assume_init() };
}
```

This specific example would still be allowed because it requires `unsafe`. I could not find any proposed uses for a safe version of this behavior. So I was unable to find examples of why we might want to break this rule.

### Why should we do this if reading any field of a union is `unsafe`?

This rule gives us permission to safely transmute values of certain kinds of unions. This is especially useful for generated bindings, where unions that fit the criteria outlined here are used extensively. Being able to guarantee this is sound with a derive macro prevents bespoke unsafe code from being written to do the same thing.

### Should this be more general? What if a byte can never be 0?

This change is intentionally narrow. Saying "safe Rust cannot write uninit bytes here" is much less restrictive than saying "safe Rust cannot write 0 here". Whether bytes are uninit or not are a purely compile-time property, but whether bytes are 0 or not may be a runtime property. This change does not preclude strengthening this guarantee in the future.

### DIFF
--- a/src/types/union.md
+++ b/src/types/union.md
@@ -16,5 +16,8 @@ Union field types are also restricted to a subset of types which ensures that th
 r[type.union.layout]
 The memory layout of a `union` is undefined by default (in particular, fields do *not* have to be at offset 0), but the `#[repr(...)]` attribute can be used to fix a layout.
 
+r[type.union.safe-construction]
+Although unions have no validity requirements, the values of union types that can be created exclusively in safe contexts are limited. Specifically, safe contexts cannot create values of union types that have uninitialized bytes at an offset where none of the fields permit uninitialized bytes.
+
 [`Copy`]: ../special-types-and-traits.md#copy
 [item]: ../items/unions.md


djkoloski:
@rustbot label +I-lang-nominated

I am nominating this for lang-team attention. At the all-hands earlier this year, several questions surrounding union validity and invariants were discussed and settled. There appeared to be mild consensus that this kind of a guarantee was OK to make, and so I'd appreciate it if the lang team could take a look.

Is this a guarantee we are comfortable making?

RalfJung:
As currently written in this PR, it sounds like you are saying that the following library is unsound:
```rust
pub union Union1 { a: i32, b: f32 }

impl Union1 {
  pub fn new() -> Self {
    // SAFETY: all bit patterns are valid for unions
    unsafe { std::mem::MaybeUninit::uninit().assume_init() }
  }
}
```
I think that library is entirely sound, so at the very least the wording will have to be improved.

I think what you are asking for is a commitment from the language that the language itself will never offer ways to de-initialize the fields of unions? In other words, you are concerned that the following code might *not* be sound:
```rust
#[repr(C)] // load-bearing to ensure the fields are at the same offset
// Note the public fields! Just to make it a bit more spicy.
pub union Union2 { pub a: i32, pub b: f32 }

impl Union2 {
  pub fn as_i32(&self) -> i32 { unsafe { *self.a } }
  pub fn as_f32(&self) -> f32 { unsafe { *self.b } }
}
```
I agree this code should be sound. In fact I would have said that this is already sound today, ;)
This code would indeed be unsound if we added e.g. a `Union {}` constructor that just creates a union value without initializing any of the fields.

I don't know what is the best way to promise that we will not add such a constructor. I don't think we have precedent for this, though it reminds me of https://github.com/rust-lang/rust/issues/161168 where also the t-lang ask was "let us officially declare that this pattern is sound".

Cc @rust-lang/opsem

djkoloski:
> As currently written in this PR, it sounds like you are saying that the following library is unsound:

It is not my intention to declare that code unsound. The wording "exclusively in safe contexts" is intended to say "without using `unsafe`". I'll try to clarify this a little more.

> I think what you are asking for is a commitment from the language that the language itself will never offer ways to de-initialize the fields of unions?

It is a little narrower than that. It's a commitment that, without using `unsafe`:

1. Union values are always initialized on creation, and
2. Any writes to fields/subfields/sub-subfields are typed. Some other ways to phrase this are:
  - You can't write data in such a way that you cannot soundly read it immediately afterward (equivalent?)
  - The value of each byte of a union is one allowed by at least one of its fields (weaker/more permissive?)

So it's still OK to de-initialize a field of a union, but only by writing another field of that same union. Like if you had a `union Foo { a: u32, b: (u8, u16) }`, initialized it with `a`, then wrote to `b`.

> I don't know what is the best way to promise that we will not add such a constructor.

I also found this somewhat difficult to articulate. Phrasing this as a limitation on the operations we can do (e.g. initialization and mutation) is pretty restrictive, since we may want to leave the door open for more kinds of safe operations on unions. So I tried to find the least-awkward way to make this a property that we can commit to upholding with exclusively safe code instead. I think it's still pretty awkward. :)

joshlf:
> I think what you are asking for is a commitment from the language that the language itself will never offer ways to de-initialize the fields of unions? In other words, you are concerned that the following code might _not_ be sound:
> 
> ```rust
> #[repr(C)] // load-bearing to ensure the fields are at the same offset
> // Note the public fields! Just to make it a bit more spicy.
> pub union Union { pub a: i32, pub b: f32 }
> 
> impl Union {
>   pub fn as_i32(&self) -> i32 { unsafe { *self.a } }
>   pub fn as_f32(&self) -> f32 { unsafe { *self.b } }
> }
> ```
> 
> I agree this code should be sound. In fact I would have said that this is already sound today, ;) This code would indeed be unsound if we added e.g. a `Union {}` constructor that just creates a union value without initializing any of the fields.

IIUC, for the intended zerocopy use, this is exactly the thing we need. Namely, if someone puts `#[derive(IntoBytes)]` on their *own* union declaration, then the [`IntoBytes::as_bytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#method.as_bytes) function (with signature `fn as_bytes(&self) -> &[u8]`) can justify that `&Self -> [u8]` is sound because the union's safety invariant disallows writing uninit bytes.

The "on their *own* union declaration" part is important because in theory someone could write a union which documents in prose that its safety invariant is weaker than the "default" one, in which case it would be unsound – and the author's fault – for the author to add `#[derive(IntoBytes)]`:

```rust
/// # Safety
///
/// The safety invariant of `Union` is equivalent to that of `MaybeUninit<Union>`.
#[repr(C)]
// #[derive(IntoBytes)] // THIS WOULD BE UNSOUND
pub union Union { pub a: i32, pub b: f32 }

impl Union {
    pub fn uninit() -> Union { unsafe { MaybeUninit::uninit().assume_init() } }
}
```

RalfJung:
So one alternative here would be to literally just promise that the library examples I gave are both sound, i.e., we will not add a language feature that makes them unsound. That is at least very precise. But is there a rigorous argument from there to what you need?

> IIUC, for the intended zerocopy use, this is exactly the thing we need. Namely, if someone puts #[derive(IntoBytes)] on their own union declaration, then the [IntoBytes::as_bytes](https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#method.as_bytes) function (with signature fn as_bytes(&self) -> &[u8]) can justify that &Self -> [u8] is sound because the union's safety invariant disallows writing uninit bytes.

Right, so basically that derive is a promise "my safety invariant is that all values of this union are obtained by initializing one of the fields, and then arbitrarily overwriting any (nested) fields with values satisfying the safety invariant of their respective types".

joshlf:
> So one alternative here would be to literally just promise that the library examples I gave are both sound, i.e., we will not add a language feature that makes them unsound. That is at least very precise. But is there a rigorous argument from there to what you need?

Hmm I'd have to think more carefully to convince myself that we could go from that to the proof we need. In particular, how do you generalize to *arbitrary* `#[rep(C)]` unions? I wonder whether such a generalization would just *be* equivalent to a more formal statement – either the one that David is proposing here, or something more restrictive like "the safety invariant is equivalent to the set of safely-constructible values using initialization and field assignment".

> > IIUC, for the intended zerocopy use, this is exactly the thing we need. Namely, if someone puts #[derive(IntoBytes)] on their own union declaration, then the [IntoBytes::as_bytes](https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#method.as_bytes) function (with signature fn as_bytes(&self) -> &[u8]) can justify that &Self -> [u8] is sound because the union's safety invariant disallows writing uninit bytes.
> 
> Right, so basically that derive is a promise "my safety invariant is that all values of this union are obtained by initializing one of the fields, and then arbitrarily overwriting any (nested) fields with values satisfying the safety invariant of their respective types".

Basically, yeah. I'm going back and forth on whether we want to say that the *act of writing a derive* is equivalent to making such a promise, but it may be nearly a moot point since writing that derive exposes an API whose mere existence implies a weaker form of that promise (namely, `T: IntoBytes` implies that `T`'s safety invariant does *not* admit uninitialized bytes at any byte offset).

jswrenn:
I'm not convinced the guarantee we're seeking here to unblock a safe, maximal `derive(IntoBytes)` on unions is actually sufficient. It's not enough for only the *language* to guarantee it never exposes a safe uninitializer on unions — the whole ecosystem must never guarantee it never provides such a thing.

I encountered this in https://github.com/google/zerocopy/pull/3492, where I'm attempting to provide a safe, maximal `derive(Project)` for field projection. Presently, it's sound for me to provide safe projections from an `&mut MyUnion` to `&mut MaybeUninit<MyUnionFieldTy>`, because Rust already has a safety onus on union projections that assume fields are in a valid state.

But if I implement such a `derive(Project)` in this manner, then I'm in conflict with us also providing a safe `derive(IntoBytes)` on unions. Obviously, us maintainers of zerocopy can look at this dilemma and conclude "well don't do that", but we can't impose that same conclusion on the entirety of the ecosystem. For all I know, someone else out there has *already* implemented such a union projection mechanism, and adopting the guarantee proposed by this PR would invalidate their abstraction.

Alternatively, if Rust had a way to spell "unsafe derive", we could instead make deriving `IntoBytes` on unions unsafe, with the user providing a safety proof they don't place union fields in uninitialized states. This would resolve https://github.com/google/zerocopy/issues/1792 without imposing a new ecosystem-wide restriction on what abstractions folks are allowed to build for unions.

RalfJung:
> But if I implement such a derive(Project) in this manner, then I'm in conflict with us also providing a safe derive(IntoBytes) on unions. Obviously, us maintainers of zerocopy can look at this dilemma and conclude "well don't do that", but we can't impose that same conclusion on the entirety of the ecosystem. For all I know, someone else out there has already implemented such a union projection mechanism, and adopting the guarantee proposed by this PR would invalidate their abstraction.

I think that's fine? It just means those two `derive` make conflicting promises about the safety invariant of the union.

I don't think we should disallow a `derive` that generates the equivalent of my first example above. We already agree that that's sound code.

But yes this does mean that the derive needs to be unsafe. Good point.

djkoloski:
Is there really a point in providing a derive from zerocopy if it always requires `unsafe`?

Sorry for the noise, it looks like this is something that lang team guidance won't be able to help with after all.

RalfJung:
> Is there really a point in providing a derive from zerocopy if it always requires unsafe?

Hm yeah that is unfortunate.
This isn't noise, I think this indicates a problem we should solve somehow. I'm just not sure how to solve it... maybe an attribute that tells the compiler about the safety invariant, so that the compiler can enforce that it is used only once per type?

PR #2350, opened 2026-09-02
Document which union patterns are unsafe

Jules-Bertholet:
Documents https://github.com/rust-lang/rust/pull/161771.

### DIFF
--- a/src/items/unions.md
+++ b/src/items/unions.md
@@ -124,6 +124,15 @@ fn f(u: MyUnion) {
 }
 ```
 
+`unsafe` is only required if the union field is accessed, in whole or in part, by the pattern, including any of its sub-patterns. For the purpose of this requirement, the following patterns are considered to perform an access:
+
+- [Literal patterns](../patterns.md#literal-patterns)
+- [Identifier patterns](../patterns.md#identifier-patterns)
+- [Range patterns](../patterns.md#range-patterns)
+- [Reference patterns](../patterns.md#reference-patterns)
+- [Struct](../patterns.md#struct-patterns) and [tuple struct](../patterns.md#tuple-struct-patterns) patterns which correspond to an enum variant
+- [Path patterns](../patterns.md#path-patterns), if the result of expanding the constant into a pattern contains one of the above.
+
 > [!WARNING]
 > The order in which the subpatterns of a pattern are tested is not specified. A union field named in a pattern may be read even when the pattern as a whole does not match. Reading a union field is undefined behavior unless it holds a valid value of its type (see [items.union.fields.validity]). Nothing else in the pattern can be relied on to prevent the read.
 >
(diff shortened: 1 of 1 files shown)


theemathas:
I think this kind of definition of what patterns "access" a thing could also be useful for implementing a lint for https://github.com/rust-lang/rust/issues/158387#issuecomment-5469275677. cc @Nadrieril

src/items/unions.md
```diff
@@ -124,6 +124,15 @@ fn f(u: MyUnion) {
 }
 ```
 
+`unsafe` is only required if the union field is accessed, in whole or in part, by the pattern, including any of its sub-patterns. For the purpose of this requirement, the following patterns are considered to perform an access:
+
+- [Literal patterns](../patterns.md#literal-patterns)
+- [Identifier patterns](../patterns.md#identifier-patterns)
+- [Range patterns](../patterns.md#range-patterns)
+- [Reference patterns](../patterns.md#reference-patterns)
```
dianne: Maybe worth noting that this includes implicit reference patterns inserted by match ergonomics?

dianne, line 132: I think this should explicitly state that it includes implicit derefs inserted by match ergonomics

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2350/e24eecf97b0c9a6dbac67191098204dc8a190aaa..c903ed60625eb5e33d3ef9c843ea2832948366da)*

src/items/unions.md
dianne, line 1: Maybe it'd be worth having an admonition acknowledging slice patterns matching on slices? Not sure on this (maybe it's immediately obvious that they'd need to be under references anyway?), but it feels a bit suspicious that they're not a list of patterns that are considered to perform an access

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2350/e24eecf97b0c9a6dbac67191098204dc8a190aaa..c903ed60625eb5e33d3ef9c843ea2832948366da)*

src/items/unions.md
```diff
@@ -124,6 +124,19 @@ fn f(u: MyUnion) {
...
+- [Non-reference patterns](../patterns.md#r-patterns.ident.binding.non-reference) matching reference values
+- [Struct](../patterns.md#struct-patterns) and [tuple struct](../patterns.md#tuple-struct-patterns) patterns which correspond to an enum variant
+- [Path patterns](../patterns.md#path-patterns), if the result of expanding the constant into a pattern contains one of the above, or if the expanded pattern could not have been written directly at the location where it is used due to field privacy or `#[non_exhaustive]`.
```
Nadrieril, line 135: I'd like to add constant patterns there, because it should be allowed to turn a constant patter nmatch into a call to `PartialEq::eq` for performance. This conflicts with https://doc.rust-lang.org/reference/patterns.html?highlight=patterns#r-patterns.const.translation a bit so we conceptually are saying "unsafety checking is done before we expand constant patterns"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2350/e24eecf97b0c9a6dbac67191098204dc8a190aaa..d663ce881a82dbbebfa96e1a089ebb7e0ab896c2)*

PR #2354, opened 2026-09-11
RISC-V: add rule about target features vs. misa extensions

TechnoPorg:
As per the discussion in https://github.com/rust-lang/rust/pull/162552 and [#t-compiler/risc-v > RISC-V extensions vs. Rust target features](https://rust-lang.zulipchat.com/#narrow/channel/250483-t-compiler.2Frisc-v/topic/RISC-V.20extensions.20vs.2E.20Rust.20target.20features/with/623171684), this clarifies that enabling a target feature means the `misa` bit of the same name (if any) is assumed to always be set. Although this originally came up in the context of the `M` extension, it applies to many others as well, which are listed in the linked Machine ISA manual page.

cc @RalfJung


src/attributes/codegen.md
```diff
@@ -563,6 +563,13 @@ specification. Many specifications are described in the [RISC-V ISA Manual],
 [version 20250508]: https://github.com/riscv/riscv-isa-manual/tree/20250508
 [RISC-V GitHub Account]: https://github.com/riscv
 
+> [!NOTE]
+> Some RISC-V standard extensions can be enabled (1) or disabled (0) via their corresponding bits in the [Machine ISA][rv-machine] (`misa`) register.
+> For example, the [A][rv-a] and [M][rv-m] bits being cleared means atomic instructions and integer multiplication/division instructions are unimplemented.
+>
+> Rust code compiled with a RISC-V target feature of the same name as an extension assumes the extension will always be available.
+> It is undefined behaviour to execute it in an environment where the extension is disabled (bit is 0 in `misa`).
```
RalfJung, line 571: I think a note (which is not normative) is not enough here. We need to say, normatively, that code compiled with a "single-letter target feature" (or whatever they are called) is UB to invoke unless the corresponding bit is actually set in the `misa` register.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/reference/2354/e24eecf97b0c9a6dbac67191098204dc8a190aaa..ac18b5c13c60e8dc0ddab87b80a50c89218167ec)*

TechnoPorg: Would that go in the same part of the document, just not as a note?

RalfJung: I'm not sure -- @traviscross @ehuss

traviscross: I'd expect the rule, or a generalization of it, to go in [Behavior Considered Undefined](https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html); then we'd add admonitions (notes) in other places where relevant (such as in codegen) pointing to that.

RalfJung: It's [already there](https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html#r-undefined.target-feature).

The point here is that we should clarify what "the current platform does not support" means specifically for RISCV target features that are impacted by the `misa` register. I would not expect such target-specific definitions to be listed on the general UB page.

traviscross: Right.  I was looking at the rule as written in the quote above, "it is undefined behavior to execute it in an environment where ...".  Rules of that form are what I'd expect to go where mentioned.

As you say, probably what we need here instead is a normative statement about how what we're discussing affects what's supported on the platform.  Then it (or perhaps an accompanying admonition) can cite [`undefined.target-feature`](https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html#r-undefined.target-feature) rather than reiterating that substance.

As for where that rule should go, I'm open to ideas.  I'd probably expect to see it in the codegen chapter, given the current organization of things.

TechnoPorg: Thanks! I've reworded it and made it a rule.

PR #2355, opened 2026-09-12
Document RISC-V d and f extensions

RalfJung:
This re-lands https://github.com/rust-lang/reference/pull/2274, except that `e` is not included this time and `d` and `f` are fully stabilized, not just as cfg-only features. (The features are restricted, they cannot be disabled/enabled on some targets. But the same is true for e.g. SSE and neon and I don't think it is reflected here.)

The corresponding rustc PR is https://github.com/rust-lang/rust/pull/161385.


traviscross (approved):

traviscross:
Thanks @RalfJung.

PR #2357, opened 2026-09-15
Glossary: add entry and examples for `n-zst`

DanielEScherzer:


DanielEScherzer:
I didn't add it in the first batch since it takes a bit more understanding, but

* unions cannot have 0 fields per https://doc.rust-lang.org/reference/items/unions.html#r-items.union.fieldless
* combined with https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.union.size-align a `repr(C)` union of only 1-ZSTs will also be a 1-ZST

traviscross (approved):

traviscross:
Thanks @DanielEScherzer.

